PR #11240, opened 2024-09-25
forgerock: fix handling of query time ranges

### LINKED ISSUE #11220: [forgerock]: agent does not keep time stamp ranges within API requirements
### Integration Name

ForgeRock [forgerock]

### Dataset Name

all

### Integration Version

1.18.0

### Agent Version

na

### Agent Output Type

elasticsearch

### Elasticsearch Version

na

na

### Software/API Version

na

### Error Message

```
{
  "log.level": "error",
  "@timestamp": "2024-09-16T09:25:15.008Z",
  "message": "Error while processing http request: failed to collect first response: failed to execute http GET: server responded with status code 400: {\"errors\":[\"24.57 Days Worth Of Data Requested (Query Start Date 2024-08-22 19:45:58 +0000 UTC, Query End Date 2024-09-16 09:25:14 +0000 UTC). Please Limit The Scope Of Your Query To Within A Day: Error: Cannot Request More Than One Days Worth Of Logs\"]}",
  "component": {
    "binary": "filebeat",
    "dataset": "elastic_agent.filebeat",
    "id": "httpjson-default",
    "type": "httpjson"
  },
  "log": {
    "source": "httpjson-default"
  },
  "log.logger": "input.httpjson-cursor",
  "log.origin": {
    "file.line": 172,
    "file.name": "httpjson/input.go",
    "function": "github.com/elastic/beats/v7/x-pack/filebeat/input/httpjson.run.func1"
  },
  …
}
```

### Event Original

_No response_

### What did you do?

Install a new policy with a greater-than-one-day look-back.

### What did you see?

Error above and no ingestion.

Happy agent.

### Anything else?

According to the Forgerock documentation [here](https://backstage.forgerock.com/docs/idcloud/latest/tenants/audit-debug-logs.html#get-log-results-for-a-time-period), it looks like we are trying to be too helpful and shooting ourselves in the foot.

> ### Get log results for a time period
> 
> Use the `beginTime` and `endTime` query parameters to return entries created between two ISO 8601 formatted times.
> 
> Example request:
> 
> ```
> $ curl -G \
> --request GET 'https://<tenant-env-fqdn>/monitoring/logs' \
> --header 'x-api-key: <api-key>' \
> --header 'x-api-secret: <api-secret>' \
> --data 'source=am-authentication' \
> --data 'beginTime=2023-03-01T12:45:00Z' \
> --data 'endTime=2023-03-01T12:50:00Z'
> ```
> 
> The `beginTime` and `endTime` query parameters are subject to these rules:
> 
> 1. If `endTime` is not specified, it defaults to the current time.
> 2. If `startTime` is not specified, it defaults to 24 hours before `endTime`.
> 3. If `startTime` is specified, it must be 24 hours or less before `endTime`. **← We are hitting this.**

Note that a possible interpretation of the documentation suggests that not providing the `endTime` parameter would allow the API to intelligently fill the `endTime` in a way that is consistent with the API's requirements. An experiment has been performed and it does not behave in a helpfully intelligent way; it adds the `endTime` as now and then baulks because now is to far into the future from `startTime`.

The fix will be to synthesise a correct `endTime` for all of the data streams.

efd6:
## Proposed commit message

```
The Forgerock API requires that the time range query satisfy three rules[1]. We
were not satisfying the last rule, that the end of the range must not be more
than 24 hours after the start of the range.

The API does not require that the end of the range must not be in the future, so
change the end time template to always render as the start of the range plus 24
hours.

The current tests do not correctly set the pagination token in some cases, and
in all cases the stream mock does not restrict queries that do not contain a
token from re-initiating a pagination cycle. Fix both of these and apply event
count assertions.

[1]https://backstage.forgerock.com/docs/idcloud/latest/tenants/audit-debug-logs.html#get-log-results-for-a-time-period
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #11220

### DIFF
--- a/packages/forgerock/_dev/deploy/docker/files/config.yml
+++ b/packages/forgerock/_dev/deploy/docker/files/config.yml
@@ -5,6 +5,7 @@ rules:
       source: "am-access"
       beginTime: "{beginTime:.*}"
       endTime: "{endTime:.*}"
+      _pagedResultsCookie: null
     responses:
       - status_code: 200
         body: |-
@@ -26,6 +27,7 @@ rules:
       source: "am-activity"
       beginTime: "{beginTime:.*}"
       endTime: "{endTime:.*}"
+      _pagedResultsCookie: null
     responses:
       - status_code: 200
         body: |-
@@ -47,6 +49,7 @@ rules:
       source: "am-authentication"
       beginTime: "{beginTime:.*}"
       endTime: "{endTime:.*}"
+      _pagedResultsCookie: null
     responses:
       - status_code: 200
         body: |-
@@ -68,6 +71,7 @@ rules:
       source: "am-config"
       beginTime: "{beginTime:.*}"
       endTime: "{endTime:.*}"
+      _pagedResultsCookie: null
     responses:
       - status_code: 200
         body: |-
@@ -89,6 +93,7 @@ rules:
       source: "am-core"
       beginTime: "{beginTime:.*}"
       endTime: "{endTime:.*}"
+      _pagedResultsCookie: null
     responses:
       - status_code: 200
         body: |-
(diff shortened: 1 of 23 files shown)


andrewkroh (approved):
LGTM.

It might be worthwhile to take what we know about the API, and codify that into a mock / emulator.

PR #11246, opened 2024-09-25
[Elastic Agent] Add data retention policy of 30d to all data streams

kpollich:
## Proposed commit message

Add data retention policy of 30d to all Elastic Agent data streams

## Checklist

### DIFF
--- a/packages/elastic_agent/changelog.yml
+++ b/packages/elastic_agent/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.1.0"
+  changes:
+    - description: Add data retention policy of 30d to all data streams
+      type: enhancement
+      link: "https://github.com/elastic/integrations/pull/10773"
 - version: "2.0.3"
   changes:
     - description: Restore Agent errors visualisation to Elastic-Agent agent info dashboard
--- a/packages/elastic_agent/data_stream/apm_server_logs/lifecycle.yml
+++ b/packages/elastic_agent/data_stream/apm_server_logs/lifecycle.yml
@@ -1,0 +1,1 @@
+data_retention: "30d"
--- a/packages/elastic_agent/data_stream/apm_server_metrics/lifecycle.yml
+++ b/packages/elastic_agent/data_stream/apm_server_metrics/lifecycle.yml
@@ -1,0 +1,1 @@
+data_retention: "30d"
--- a/packages/elastic_agent/data_stream/auditbeat_logs/lifecycle.yml
+++ b/packages/elastic_agent/data_stream/auditbeat_logs/lifecycle.yml
@@ -1,0 +1,1 @@
+data_retention: "30d"
--- a/packages/elastic_agent/data_stream/auditbeat_metrics/lifecycle.yml
+++ b/packages/elastic_agent/data_stream/auditbeat_metrics/lifecycle.yml
@@ -1,0 +1,1 @@
+data_retention: "30d"
--- a/packages/elastic_agent/data_stream/cloud_defend_logs/lifecycle.yml
+++ b/packages/elastic_agent/data_stream/cloud_defend_logs/lifecycle.yml
@@ -1,0 +1,1 @@
+data_retention: "30d"
--- a/packages/elastic_agent/data_stream/cloudbeat_logs/lifecycle.yml
+++ b/packages/elastic_agent/data_stream/cloudbeat_logs/lifecycle.yml
@@ -1,0 +1,1 @@
+data_retention: "30d"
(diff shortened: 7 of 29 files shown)


cmacknz:
What happens if someone has manually configured ILM or DSL on these separately? What takes precendence?

30d is a good default but I wonder if this is potentially a breaking change for people, especially because sometimes data retention can fall under compliance requirements for some areas.

kpollich:
> What happens if someone has manually configured ILM or DSL on these separately? What takes precedence?

The behavior is controlled by the `index.lifecycle_prefer_ilm` setting on the backing index:

https://www.elastic.co/guide/en/elasticsearch/reference/current/data-stream-lifecycle-settings.html#index-lifecycle-prefer-ilm

This default to true, so unless the user takes action to change this behavior (e.g. by following https://www.elastic.co/guide/en/elasticsearch/reference/current/tutorial-migrate-data-stream-from-ilm-to-dsl.html), their existing ILM policy would continue to take precedence. 

> 30d is a good default but I wonder if this is potentially a breaking change for people, especially because sometimes data retention can fall under compliance requirements for some areas.

We could potentially de-risk this with a longer default retention policy. If the goal is "prevent infinite growth" we can meet that goal even with a year long retention policy for example. If the goal is to cap the growth at a more reasonable size on disk, then obviously a shorter policy would help us there.

felixbarny:
I, too, have concerns about this being a potentially breaking change. But putting that aside, we could consider setting the lifecycle at a central location, like the [`logs@settings`](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/logs%40settings.json) component template. This would impact more than just integration data streams, however.

mrodm:
> Weirdly the tests failed here during the Docker cleanup lifecycle of the integration tests:
> 
> ```
> Take down the Elastic stack
> --
>   | Error: tearing down the stack failed: stopping docker containers failed: could not
> create docker compose project: could not find Docker Compose configuration file:
> /opt/buildkite-agent/.elastic-package/profiles/default/stack/docker-compose.yml: stat
> /opt/buildkite-agent/.elastic-package/profiles/default/stack/docker-compose.yml: no such file
> or directory
> ```
> 
> Not sure what would cause this. Maybe @mrodm knows?

The error is due to it's missing to update the version in the manifest.yml file (at the root of the package) to be the same version as in the changelog file:

```
Error: checking package failed: linting package failed: found 1 validation error:
   1. current manifest version doesn't have changelog entry
[elastic_agent] run_tests_package failed
```

kpollich:
> But putting that aside, we could consider setting the lifecycle at a central location, like the [logs@settings](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/logs%40settings.json) component template.

I thought the same, and I think we even used to do this via `logs@lifecycle.json` - https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/logs%40lifecycle.json. I tried to trace through the git blame for the deprecation here and didn't come away with a great sense of why this is deprecated or if this is something we should bring back in some way. For serverless especially it seems like infinitely growing log storage is going to be problematic, and we should consider our options for defaults here.

cmacknz (approved):

felixbarny:
>  For serverless especially it seems like infinitely growing log storage is going to be problematic, and we should consider our options for defaults here.

I was under the impression that on serverless, there will be a project-wide default/max retention setting.

felixbarny:
See [`data_streams.lifecycle.retention.max`](https://www.elastic.co/guide/en/elasticsearch/reference/8.16/data-stream-lifecycle-settings.html#data-streams-lifecycle-retention-max) and [`data_streams.lifecycle.retention.default`](https://www.elastic.co/guide/en/elasticsearch/reference/8.16/data-stream-lifecycle-settings.html#data-streams-lifecycle-retention-default).

kpollich:
> I was under the impression that on serverless, there will be a project-wide default/max retention setting.
> 
> See [data_streams.lifecycle.retention.max](https://www.elastic.co/guide/en/elasticsearch/reference/8.16/data-stream-lifecycle-settings.html#data-streams-lifecycle-retention-max) and [data_streams.lifecycle.retention.default](https://www.elastic.co/guide/en/elasticsearch/reference/8.16/data-stream-lifecycle-settings.html#data-streams-lifecycle-retention-default).

This would make a lot of sense, and I think this would be vastly preferable for us as opposed to making sure all of "installed by default" integrations like `elastic_agent` explicitly specify some retention policy that can't be changed by the user in the same way these ES settings can. 

Searching through Git history a bit I found the PR where these were introduced here: https://github.com/elastic/elasticsearch/pull/111972. 

I'll reach out offline about serverless stuff here.

kpollich:
There's work tracked elsewhere to set a default + max retention period for logs data on observability projects, but there are concerns with hidden indices. Either way, this is tracked by the o11y team and I'm closing this PR.

PR #11248, opened 2024-09-25
qualys_vmdr: retain event.original as json

efd6:
## Proposed commit message

In real use of the API, the XML document that is sent may be hundreds of megabytes resulting in inability to ingest the resulting event documents. Even with compression, the XML is on the order of tens of megabytes. So do not retain the full XML, but only the original generated JSON when preserve_original_event is enabled.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/qualys_vmdr/changelog.yml
+++ b/packages/qualys_vmdr/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "5.2.0"
+  changes:
+    - description: Retain event.original for asset_host_detection and knowledge_base as JSON.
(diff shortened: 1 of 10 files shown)


andrewkroh (commented):
This approach looks good.

To put some concrete numbers on it, in one example we saw the initial response from the knowledge_base API be 886 MB of XML. That compresses down to 64 MB of gzip or 89 MB after base64 encoding. So as a matter of practice, we are including JSON here instead of the full XML response.

andrewkroh (approved):

PR #11249, opened 2024-09-25
[windows] Windows Defender Data stream overhaul to GA

nicpenning:
- Enhancement

## Proposed commit message

Overhaul Windows Defender data stream in the Windows integration to make it GA.

Added many ECS fields and removed un-needed fields/processors

## Checklist

## Screenshots
image


nicpenning:
Ready for review and tests.

nicpenning:
Made a few corrections. Please review and test now.

intxgo (commented):
changes LGTM, but I guess this will fully process only a subset of Windows Defender events. Are the other IDs unimportant or there's just no documentation for them?

packages/windows/data_stream/windows_defender/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -74,38 +146,89 @@ processors:
...
+      if: ctx.process?.executable == null && ctx.event?.code != "1126"
+  - rename:
+      field: winlog.event_data.Process_Name
+      target_field: process.name
+      ignore_missing: true
+      if: ctx.process?.name == null && ctx.event?.code == "1126"
+  - grok:
+      field: process.executable
+      patterns:
+        - '(?<process.name>([^\\]*$))'
```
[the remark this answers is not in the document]
nicpenning: Is this extra \\ because it lives in a YML file? This is grabbing everything after the last \ in a process.executable such as:
c:\data\helloworld.exe ---> helloworld.exe

I think my tests worked for this? :) Please advise!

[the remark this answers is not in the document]
intxgo: It caught my eye for the sake of consistency, windows paths will be always with double slashes to parse, but indeed there's no functional change.

intxgo (approved):

packages/windows/data_stream/windows_defender/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -27,36 +27,108 @@ processors:
...
+      value: file-uploaded
+      if: ctx.event?.code == "2050"
+  - set:
+      field: event.action
+      value: malware-detected
+      if: ctx.event?.code == "1116" || ctx.event?.code == "1117"
+  - set:
+      field: event.action
+      value: malware-quarantined
+      if: ctx.event?.code == "1117" && ctx.winlog?.event_data?.Action_Name == "Quarantine"
```
taylor-swanson, line 56: Will we have any issues with case sensitivity here, or is it guaranteed to always be `Quarantine`? If not, you can use  something like this to do a case insensitive comparision:

```
ctx.winlog?.event_data?.Action_Name.equalsIgnoreCase("Quarantine") == true
```

(The `== true` is needed so the expression results in a bool and not a null.

packages/windows/data_stream/windows_defender/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -27,36 +27,108 @@ processors:
...
+      if: ctx.event?.category == null  
   - set:
-      field: event.type
-      value: ["start"]
+      field: event.kind
+      value: event
+  - set:
+      field: event.outcome
+      value: success
+      if: ctx.winlog?.event_data?.Error_Description == "The operation completed successfully. "
```
taylor-swanson, line 80: Is the trailing whitespace intended?

packages/windows/data_stream/windows_defender/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -74,38 +146,89 @@ processors:
...
-      field: winlog.user_data.FullFilePath
+      field: winlog.event_data.Filename
+      patterns:
+        - ':_%{GREEDYDATA:file.path}'
+      ignore_missing: true
+      on_failure:
+        - grok:
+            field: winlog.event_data.Filename
+            patterns:
+              - '%{GREEDYDATA:file.path}'
```
taylor-swanson, line 169: Unless I'm missing something, we can add the second pattern to the first list.
```suggestion
        - ':_%{GREEDYDATA:file.path}'
        - '%{GREEDYDATA:file.path}'
      ignore_missing: true
```

Alternatively, [gsub](https://www.elastic.co/guide/en/elasticsearch/reference/current/gsub-processor.html) could be used to remove this prefix.

Same thing applies to grok processor below.

taylor-swanson (commented):
Couple of minor comments, otherwise seems good to me

packages/windows/data_stream/windows_defender/elasticsearch/ingest_pipeline/default.yml
nicpenning: Good question! According to documentation, yes. I am good with keeping it specific until proven otherwise.

I will keep this in mind for any future updates, thank you!

image

packages/windows/data_stream/windows_defender/elasticsearch/ingest_pipeline/default.yml
nicpenning: Sadly, yes. This is how I see it in the raw events the Elastic Agent/Filebeat send to the stack. I had to triple check this.

I am interested in somebody's else's environment if this seems to be the case.

packages/windows/data_stream/windows_defender/elasticsearch/ingest_pipeline/default.yml
nicpenning: I struggled with this. It seems that order does not matter for the GROK expressions. If this is my example data:
`"winlog.event_data.Path": "file:_C:\Users\itpr12009\Desktop\eat_more_yams.exe"`

It will always end up: 
`"file.path": "file:_C:\Users\itpr12009\Desktop\eat_more_yams.exe"`

Is there any reason why the order doesn't work?
I tried both of these:
image
image


GSUB might do the trick here. I will see what I can do here.

nicpenning: @taylor-swanson thanks for the review! I went ahead and used GSUB and SET processors to handle this. I also added some missed event_data blocks in the test data for improved testing. Please let me know what you think!

taylor-swanson: Yeah `GREEDYDATA` has some nasty side effects (and to be honest, I don't fully understand how grok handles wildcards like this). If the intent is to just remove some characters, gsub is more or less meant for that.

nicpenning: Yes, very GREEDY it is :D

fearful-symmetry (approved):
Logic looks fine, can't comment on the windows-specific fields though.

nicpenning:
Looks like pipeline tests failed. I will test locally to see what is going on.

nicpenning:
Doesn't appear that thr generated test files were pushed. I will get those pushed as soon as I can.

nicpenning:
Ready for tests again!

rdner:
@taylor-swanson could you validate the changes in https://github.com/elastic/integrations/pull/11249/commits/3de2f48000c06f11a8cac57868f1c2ddca4b5c97 please?

taylor-swanson:
> @taylor-swanson could you validate the changes in [3de2f48](https://github.com/elastic/integrations/commit/3de2f48000c06f11a8cac57868f1c2ddca4b5c97) please?

I think the test results are good.

- At first I thought an event was removed, but it was the git diff that tripped me up. It looks like one of the test cases (event ID `1116`) was moved to the top and that can cause some misleading git diffs as a result. This includes extra lines showing up in the diff even though the results didn't really change. This is why I try to avoid reordering test cases, since the resulting git diffs tend to be misleading. Everything seems to check out, but feel free to point out something that's suspicious if I missed something.
- It looks like the `user_data` field was removed and was replaced with an enriched `event_data` field. The PR doesn't mention why this is the case, but I'll assume we have better sample data now and are including it in the tests.
- The rest of the changes to the expected test file are either from fields being sorted or updates from the changes in the pipeline.

nicpenning:
> > @taylor-swanson could you validate the changes in [3de2f48](https://github.com/elastic/integrations/commit/3de2f48000c06f11a8cac57868f1c2ddca4b5c97) please?
> 
> I think the test results are good.
> 
> * At first I thought an event was removed, but it was the git diff that tripped me up. It looks like one of the test cases (event ID `1116`) was moved to the top and that can cause some misleading git diffs as a result. This includes extra lines showing up in the diff even though the results didn't really change. This is why I try to avoid reordering test cases, since the resulting git diffs tend to be misleading. Everything seems to check out, but feel free to point out something that's suspicious if I missed something.
> * It looks like the `user_data` field was removed and was replaced with an enriched `event_data` field. The PR doesn't mention why this is the case, but I'll assume we have better sample data now and are including it in the tests.
> * The rest of the changes to the expected test file are either from fields being sorted or updates from the changes in the pipeline.

Good observations.

I did reorder because I wanted the system test to use this more practical event in the docs. Also when I generated improved json that maps closer to the real world, I used it as is and didn't try to put them back in the right order.

As for the user fields, those were likely residual fields from other windows integrations that this one didn't really use so I capitalized on what actual event fields are being used in the event IDs I was able to generate for my tests.

I will be cautious next time keeping these event IDs in the proper order. Thanks for the review!

PR #11258, opened 2024-09-26
[citrix_adc] Allow customer to define date/time format

jrmolin:
## Proposed commit message

[citrix_adc] Allow customer to define date/time format
* Customer was seeing errors when Citrix appliance sent DeLink logs with Day/Month/Year on days larger than 12 (so obviously not a month).
* Found a lot of places where the default date formatting chooses month first, which will cause inaccuracies in how the data is stored.
* Created a new integration option that will allow the customer to set the specific format (dd/MM/yyyy:HH:mm:ss is what this customer would use).
* Needed to find all the places where date information is parsed and make sure they are consistent.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/citrix_adc/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -88,8 +88,24 @@ processors:
         - yyyy MMMM d HH:mm:ss
         - yyyy MMMM  d HH:mm:ss
         - yyyy MMMM dd HH:mm:ss
+  - script:
+      description: Convert timestamp_native via custom format
+      tag: date_tmp_timestamp_native_custom_format
+      lang: painless
+      source: >-
+        def dateFormat = ctx["_conf"]["custom_date_format"];
```
aleksmaus, line 96: what hapens with `ctx["_conf"]["custom_date_format"]` in painless if `ctx["_conf"]`  is not set or null ?
the reason I'm asking because the later `if` condition seems to assume that this is possible
```
if: ctx._tmp?.timestamp_native != null &&
```

packages/citrix_adc/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.8.2"
+  changes:
+    - description: "Allow the customer to specify a custom date/time format."
+      type: bugfix
```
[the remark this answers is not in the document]
jrmolin: It feels like that, yeah.

packages/citrix_adc/data_stream/log/agent/stream/stream.yml.hbs
```diff
@@ -21,9 +21,14 @@ processors:
 {{#if processors}}
 {{processors}}
 {{/if}}
-{{#if tz_offset}}
+{{#if (or tz_offset custom_date_format) }}
```
jrmolin, line 24: I think this is wrong. I'm seeing errors in the local kibana when I'm trying to test it. I don't know how to do a logical or in handlebars.

packages/citrix_adc/data_stream/log/elasticsearch/ingest_pipeline/default.yml
aleksmaus: right. but what about `ctx["_conf"]["custom_date_format"]` above in the painless script itself. will it error out?

packages/citrix_adc/data_stream/log/agent/stream/stream.yml.hbs
aleksmaus: there is doesn't seem to be an "or" handlebars handler registered on the kibana side 
might have to do this with `if ... else if ... else` instead

packages/citrix_adc/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.9.0"
+  changes:
+    - description: "Allow the customer to specify a custom date/time format."
```
andrewkroh, line 4:
```suggestion
    - description: Make the data/time parsing format configurable.
```

packages/citrix_adc/data_stream/log/manifest.yml
```diff
@@ -180,6 +190,16 @@ streams:
...
+      - name: custom_date_format
+        type: text
+        title: Custom Date Format
+        multi: false
+        required: false
+        show_user: true
+        default: ''
+        description: |-
+          Format to use to parse the date/time fields in the data.
+          If not populated, this uses one of "yyyy/MM/dd:HH:mm:ss", "MM/dd/yyyy:HH:mm:ss", or the ISO8601 standard.
```
andrewkroh, line 202: As a user, how do I learn what the available format specifiers are? I think we need to link to the relevant documentation of explicitly state what we are using to parse so that the user can find the information.

packages/citrix_adc/data_stream/log/agent/stream/stream.yml.hbs
jrmolin: Yeah, I just decided to stop trying to patch kibana and do that.

packages/citrix_adc/changelog.yml
jrmolin: changed locally

packages/citrix_adc/data_stream/log/manifest.yml
jrmolin: So link to the ingest pipeline on `Date`? So https://www.elastic.co/guide/en/elasticsearch/reference/current/date-processor.html?

jrmolin: Oh, link to the formats allowed by the Java class I'm using. So https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/time/format/DateTimeFormatter.html ?

andrewkroh: The ES docs link out to `DateTimeFormatter` so I think that's what this should link users to as well.

jrmolin: Ah, found them in the Painless documentation: https://www.elastic.co/guide/en/elasticsearch/painless/current/painless-datetime.html. Looks exactly like what I'm trying to do. Should I fall back to the default parsing options?

packages/citrix_adc/data_stream/log/_dev/test/pipeline/test-citrix-waf-native.log-expected.json
```diff
@@ -1797,6 +1797,7 @@
                 "preserve_duplicate_custom_fields"
             ],
             "url": {
+                "extension": "com/page",
```
mjwolf, line 1800: This doesn't seem right, `url.extension` should be the file extension, if it exists. But I also don't see how your changes could cause this. Maybe it needs a new bug reported, so it can be investigated more

packages/citrix_adc/data_stream/log/manifest.yml
```diff
@@ -236,3 +258,14 @@ streams:
...
+        type: text
+        title: Custom Date Format
+        multi: false
+        required: false
+        show_user: true
+        default: ''
+        description: |-
+          Format to use to parse the date/time fields in the data.
+          This is used to build a Java DateTimeFormatter in [Painless](https://www.elastic.co/guide/en/elasticsearch/painless/current/painless-datetime.html) to parse the incoming date. For Day/Month/Year... parsing, this could be "dd/MM/yyyy:HH:mm:ss".
+          If not populated, this uses one of "yyyy/MM/dd:HH:mm:ss", "MM/dd/yyyy:HH:mm:ss", or the ISO8601 standard.
```
mjwolf, line 271: This wording is a bit confusing to me. It makes me think it will use only one of these, but doesn't specify which one. 

Something like this could be better `If not populated, this matches any of "yyyy/MM/dd:HH:mm:ss", "MM/dd/yyyy:HH:mm:ss", or the ISO8601 standard.`

packages/citrix_adc/data_stream/log/_dev/test/pipeline/test-citrix-waf-native.log-expected.json
jrmolin: Oh, interesting. I didn't even look at the generated output. It failed so consistently and for so long that I just automatically trusted when it worked.

packages/citrix_adc/data_stream/log/manifest.yml
jrmolin: Thanks! Yes, that is much clearer.

packages/citrix_adc/data_stream/log/_dev/test/pipeline/test-citrix-waf-native.log-expected.json
jrmolin: I think it's a change from the config file for one of the tests. I don't quite understand why it's there nor why it didn't cause problems before now. I didn't change that log.

taylor-swanson: If it's unrelated, we can ignore it for this change (but file a separate issue so we can address it).

taylor-swanson (approved):
LGTM, ~~we'll need approval from @elastic/obs-infraobs-integrations since there is shared ownership on this integration.~~

Edit: Never mind, we can merge with just one approval.

PR #11227, opened 2024-09-24
[CI] Update find oldest supported script

mrodm:
## Proposed commit message

Add support in `find_oldest_supported_version.py` script to return the version as is when that version does not exist, but exist a next patch (newer) version.

Currently, `8.13.0` version is not part of the `versions`  reported in the artifacts-api response. But that Elastic stack version exists.

Before:
```
Manifest ^8.13.0 -> script 8.x-SNAPSHOT 	[keycloak]
Manifest ^8.13.0 -> script 8.x-SNAPSHOT 	[httpjson]
Manifest ^8.13.0 -> script 8.x-SNAPSHOT 	[vectra_detect]
Manifest ^8.3.0 -> script 8.3.0 	[fortinet_fortimail]
```

After:
```
Manifest ^8.13.0 -> script 8.13.0 	[keycloak]
Manifest ^8.13.0 -> script 8.13.0 	[httpjson]
Manifest ^8.13.0 -> script 8.13.0 	[vectra_detect]
Manifest ^8.3.0 -> script 8.3.0 	[fortinet_fortimail]
```

<details>

<summary>Current example of artifacts-api-response</summary>

```json
{
  "versions": [
    "7.17.20",
    "7.17.21",
    "7.17.22",
    "7.17.23",
    "7.17.24-SNAPSHOT",
    "7.17.24",
    "7.17.25-SNAPSHOT",
    "7.17.25",
    "8.12.3",
    "8.13.0+build202403281537",
    "8.13.0+build202403281758",
    "8.13.1+build202404121909",
    "8.13.1+build202404122010",
    "8.13.1",
    "8.13.2",
    "8.13.3",
    "8.13.4",
    "8.13.5",
    "8.14.0",
    "8.14.1",
    "8.14.2",
    "8.14.3",
    "8.14.4",
    "8.15.0",
    "8.15.1-SNAPSHOT",
    "8.15.1",
    "8.15.2-SNAPSHOT",
    "8.15.2",
    "8.16.0-SNAPSHOT",
    "9.0.0-SNAPSHOT"
  ],
  "aliases": [
    "7.17-SNAPSHOT",
    "7.17",
    "8.x-SNAPSHOT",
    "8.12",
    "8.13",
    "8.14",
    "8.15-SNAPSHOT",
    "8.15",
    "8.16-SNAPSHOT",
    "9.0-SNAPSHOT"
  ],
  "manifests": {
    "last-update-time": "Tue, 24 Sep 2024 07:41:14 UTC",
[...]

PR #11265, opened 2024-09-27
Remove major snapshots check find oldest script

mrodm:
## Proposed commit message

Simplify `find_oldest_supported_script.py` script used in CI pipelines in charge of retrieving the Elastic stack version to be used in `elastic-package stack up` command. With these changes, it will not output a version like `<major>-x.SNAPSHOT`.

Given the minimum version `X.Y.Z` supported in the package defined in the package manifest (`.conditions.kibana.version`), the oldest Elastic stack supported follows this logic:
- If the next version does not exist (`X.Y.Z+1` or `X.Y.Z+1-SNAPSHOT`) and it does exist `X.Y.Z-SNAPSHOT`, then the `X.Y.Z-SNAPSHOT` version is returned.
- In any other case, the version from the package manifest is returned.

In this PR, the process to check if there is any newer version in elastic-package is disabled. This is disabled via the environment variable `ELASTIC_PACKAGE_CHECK_UPDATE_DISABLED`

## Checklist

- ~[ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.~
- ~[ ] I have verified that all data streams collect metrics or logs.~
- ~[ ] I have added an entry to my package's `changelog.yml` file.~
- ~[ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).~

## Author's Checklist

- [ ] Test stacks in current packages

```shell
python3 .buildkite/scripts/find_oldest_supported_version.py --manifest-path <manifest_path>
# results from last week where 8.15.3-SNAPSHOT did not exist in the API response
python3 .buildkite/scripts/find_oldest_supported_version.py --manifest-path packages/vsphere/manifest.yml
8.15.2-SNAPSHOT

# as 8.15.3-SNAPSHOT is already reported by the API, the stack version is now 8.15.2
python3 .buildkite/scripts/find_oldest_supported_version.py --manifest-path packages/vsphere/manifest.yml
8.15.2

python3 .buildkite/scripts/find_oldest_supported_version.py --manifest-path packages/zoom/manifest.yml
8.13.0
```

It could be checked stack version reported for all packages with this script:
```shell
#!/bin/bash
set -euo pipefail

for manifest in $(find . -maxdepth 3 -mindepth 3 -name manifest.yml | LC_ALL=C sort) ; do
    package=$(echo $manifest |awk -F '/' '{print $3}')
    version_manifest=$(cat $manifest |yq -r '.conditions.kibana.version')
    if [[ $version_manifest == null ]]; then
        version_manifest=$(cat $manifest |yq -r '.conditions."kibana.version"')
    fi
    
    script_version=$(python3 .buildkite/scripts/find_oldest_supported_version.py --manifest-path $manifest)
    
    echo -e "Manifest ${version_manifest} -> script ${script_version} \t[$package]"
done
```

## Related issues

- Follows https://github.com/elastic/integrations/pull/11227


.buildkite/scripts/find_oldest_supported_version.py
```diff
@@ -26,8 +26,6 @@ def fetch_version():
 
 
 def find_oldest_supported_version(kibana_version_condition: str) -> str:
-    # The logic of this function is copied from https://github.com/elastic/apm-pipeline-library/blob/main/vars/findOldestSupportedVersion.groovy
```
mrodm, line 29: This function is not anymore based on that code

mrodm:
Created issue for the `mysql` package that is failing in CI: https://github.com/elastic/integrations/issues/11278

.buildkite/scripts/find_oldest_supported_version.py
jsoriano: Yes :+1:

jsoriano (approved):

PR #11267, opened 2024-09-27
[Elastic Connectors] Add index name as input var

jedrazb:
## Add connector_name as input variable

Update connector package definition to pass `connector_name` as input variable for the package logic. 

## Checklist

## Author's Checklist

- [x] Test the final version of this package def e2e (ping @jedrazb for exact steps if needed)

- Since agentless package is not public yet, ping @jedrazb for steps

## Screenshots

Screenshot 2024-10-03 at 17 14 09

### DIFF
--- a/packages/elastic_connectors/changelog.yml
+++ b/packages/elastic_connectors/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: 0.0.2
   changes:
-    - description: Add index_name as the connector input variable
+    - description: Add connector_name as the connector input variable
       type: enhancement
       link: https://github.com/elastic/integrations/pull/11267
 - version: 0.0.1
--- a/packages/elastic_connectors/changelog.yml
+++ b/packages/elastic_connectors/changelog.yml
@@ -1,9 +1,9 @@
 # newer versions go on top
 - version: 0.0.2
   changes:
-    - description: WIP....
+    - description: Add index_name as the connector input variable
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/1234
+      link: https://github.com/elastic/integrations/pull/11267
 - version: 0.0.1
   changes:
     - description: Initial draft of the package
--- a/packages/elastic_connectors/changelog.yml
+++ b/packages/elastic_connectors/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: 0.0.2
+  changes:
+    - description: WIP....
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1234
 - version: 0.0.1
   changes:
     - description: Initial draft of the package
--- a/packages/elastic_connectors/manifest.yml
+++ b/packages/elastic_connectors/manifest.yml
@@ -42,7 +42,8 @@ policy_templates:
         title: GitHub & GitHub Enterprise Server Connector
         description: Connectors syncs data from an original data source to an Elasticsearch index.
         vars:
-          - name: connector_name
+          - &connector_name
+            name: connector_name
             type: text
             title: Connector Name
             required: true
@@ -68,12 +69,7 @@ policy_templates:
         title: Google Drive Connector
         description: Connectors syncs data from an original data source to an Elasticsearch index.
         vars:
-          - name: connector_name
-            type: text
-            title: Connector Name
-            required: true
-            show_user: false
-            description: Connector name to identify the connector in the UI
+          - <<: *connector_name
             default: "[Elastic-managed] Google Drive Connector"
         template_path: google_drive.yml.hbs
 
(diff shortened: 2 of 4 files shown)


packages/elastic_connectors/manifest.yml
```diff
@@ -37,11 +37,19 @@ policy_templates:
...
       - type: connectors-py
         title: GitHub & GitHub Enterprise Server Connector
         description: Connectors syncs data from an original data source to an Elasticsearch index.
+        vars:
+          - &connector_name
+            name: connector_name
+            type: text
+            title: Connector Name
+            required: true
+            show_user: false
```
jedrazb, line 50: If `show_user: true` this won't render in UI, see bug report: https://github.com/elastic/kibana/issues/194310

packages/elastic_connectors/manifest.yml
```diff
@@ -37,11 +37,19 @@ policy_templates:
         enabled: false
       agentless:
         enabled: true
-    multiple: false
```
jedrazb, line 40: Revert to default `multiple: true` given that now we can have multiple policies with the same input (each input corresponds to a single connector)

packages/elastic_connectors/manifest.yml
seanstory: Can you add a code comment with a link to that bug, so that we remember to change this later? Or file an issue for our backlog?

packages/elastic_connectors/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: 0.0.2
+  changes:
+    - description: Add index_name as the connector input variable
```
seanstory, line 4: Should this be about adding connector name? Not index name?

packages/elastic_connectors/manifest.yml
```diff
@@ -37,11 +37,19 @@ policy_templates:
...
         description: Connectors syncs data from an original data source to an Elasticsearch index.
+        vars:
+          - &connector_name
+            name: connector_name
+            type: text
+            title: Connector Name
+            required: true
+            show_user: false
+            description: Connector name to identify the connector in the UI
+            default: "[Elastic-managed] Github Connector"
```
seanstory, line 52: Can we add an auto-increment number in here, like we get in the policy and integration name defaults? That might help avoid collisions.

packages/elastic_connectors/manifest.yml
seanstory: Double checking - this would let you set up two github connectors at once? Or this shows you both github and gdrive at once? I'm hoping NOT the later.

packages/elastic_connectors/manifest.yml
jedrazb: Unfortunately not, we don't have much control over the logic here. Perhaps we could leave this blank and have a custom naming logic (with autoincrement or short randomised string) in connectors python upon creation

jedrazb: See, the approach I mentioned in the component PR https://github.com/elastic/connectors/pull/2861#discussion_r1787563818

I think it would be better to mark this field as optional, don't default to anything and have custom logic to generate the name in connector component. This makes this process independent on fleet logic :) for now I'm appending short id to make connectors unique in frontend, see:

Screenshot 2024-10-04 at 13 33 10

packages/elastic_connectors/changelog.yml
jedrazb: good spot, forgot to rename! Thanks

packages/elastic_connectors/manifest.yml
jedrazb: Actually we might want to keep `multiple: false` it prevents adding multiple integrations of type elastic connectors to the same policy. I think this is what we want. 

> Or this shows you both github and gdrive at once?

This doesn't affect this, if you click google drive tile it will just show you that tile. If you click "elastic connectors" you will see both, at some point we need to figure this out. I know it's possible to add custom UI for that parent integration with some effort. 

The good thing is that the component logic will detect if multiple inputs were passed to the component (e..g when enabling multiple inputs in parent integration tile UI) and will log a warning.

jedrazb: Added multiple: false back in https://github.com/elastic/integrations/pull/11267/commits/c635362f9ee791af462fc8b8aaf8d92838450fd8

packages/elastic_connectors/manifest.yml
seanstory: I think that makes sense, to control this in the backend for now. CC @kpollich, in case the idea of auto-increment vars is something worth us filing an issue for, somewhere.

packages/elastic_connectors/manifest.yml
seanstory: 👍 sounds good

> if you click "elastic connectors" (parent package) you will see bothgdrive and github... at some point we need to figure this out.

Proposal to remove that parent tile is being tracked in https://github.com/elastic/package-spec/issues/802

seanstory (approved):

PR #11239, opened 2024-09-24
Add aws.firehose.arn, aws.firehose.request_id and aws.metrics_names_fingerprint fields

kaiyan-sheng:
## Proposed commit message

This PR adds `aws.firehose.arn` and `aws.firehose.request_id` field definitions for the firehose integration.

This PR also adds `aws.metrics_names_fingerprint` which is a hash of the list of metric names exist in each document. This way we don't have to count `request_id` as a dimension.
 
## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).

## How to test this PR locally?
1. Install firehose integration
2. Install aws integration
3. POST two documents using the dev tool into the `metrics-aws.cloudwatch-default` data stream to mimic metrics coming in from Firehose: ` POST metrics-aws.cloudwatch-default/_doc/`
<details>
```
POST metrics-aws.cloudwatch-default/_doc/
{
  "@timestamp": "2024-09-25T23:12:00.000Z",
    "agent.type": "firehose",
    "cloud.provider": "aws",
    "cloud.account.id": "627286350134",
    "cloud.region": "eu-central-1",
    "aws.exporter.arn": "arn:aws:cloudwatch:eu-central-1:627286350134:metric-stream/CustomFull-KefpMG",
    "aws.cloudwatch.namespace": "AWS/Kinesis",
    "aws.firehose.arn": "arn:aws:firehose:eu-central-1:627286350134:deliverystream/KS-PUT-ELI-2",
    "aws.firehose.request_id": "7628c293-ba54-4ea4-9abe-d256976c1dbd",
    "aws.dimensions": {
      "StreamName": "test-esf-encrypted"
    },
    "aws.kinesis.metrics.GetRecords_Success": {
      "count": 240,
      "sum": 240,
      "avg": 1,
      "max": 1,
      "min": 1
    },
    "aws.kinesis.metrics.GetRecords_Bytes": {
      "count": 240,
      "sum": 0,
      "avg": 0,
      "max": 0,
      "min": 0
    },
    "aws.kinesis.metrics.GetRecords_Latency": {
      "count": 240,
[...]

PR #11281, opened 2024-09-30

kaiyan-sheng:
## Proposed commit message

This PR adds `aws.metrics_names_fingerprint` which is a hash of the list of metric names exist in each document. For metrics coming in from Firehose, there can be cases two documents have the same timestamp, dimension, namespace, accountID, exportARN and region BUT from two different requests. With adding the `aws.metrics_names_fingerprint` as TSDB dimension, we are able to ingest these two documents into ES.

## Checklist

## Related issues

- Relates https://github.com/elastic/integrations/pull/11239


packages/aws/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.24.2"
+  changes:
+    - description: Add aws.metrics_names_fingerprint.
+      type: enhancement
```
efd6, line 5: Enhancement or bug? If enhancement, please bump the minor rather than the patch.

kaiyan-sheng: well i have mix feeling about it 😂 I will make it a bug fix so we can release for 2.24.2!

efd6 (approved):
Approving as codeowner.

axw (approved):

agithomas (approved):
LGTM!

PR #11282, opened 2024-10-01
crowdstrike: add shims to recover deprecated fields

efd6:
## Proposed commit message

```
The documentation for the deprecation of fields indicates the following
correspondences:

old	                              new
is_synthetic_quarantine_disposition   pattern_disposition* to identify quarantined files
has_script_or_module_ioc              ioc_context
ioc_values                            ioc_value

However, there is no other information relating to how these correspond
with each other.

By inspection of documents from an alerts stream, we can see that
pattern_disposition_details contains a quarantine_file boolean. This,
with the text in the deprecation notice, hints that we can use this
field to get the is_synthetic_quarantine_disposition. The ioc_context
field contains an array of object with a type property which in the
examples I have available include (only) "module", hinting that this can
be used to detect the state corresponding to has_script_or_module_ioc.
Finally, ioc_value fields are sprinkled around the documents, so collect them
into ioc_values.

The test case is derived from the first case, but with deprecated fields
removed.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/crowdstrike/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -844,6 +844,23 @@ processors:
...
+      tag: reconstruct_has_script_or_module_ioc_from_ioc_context
+      lang: painless
+      if: ctx.crowdstrike?.alert?.has_script_or_module_ioc == null && ctx.json?.ioc_context instanceof List
+      source: |
+        if (ctx.crowdstrike == null) {
+          ctx.crowdstrike = [:];
+        }
+        if (ctx.crowdstrike.alert == null) {
+          ctx.crowdstrike.alert = [:];
+        }
```
chrisberkhout, line 857: Either:
- move the `ctx.crowdstrike.alert` initalization into 
  `if (c.type == 'module' || c.type == 'script') { ...`, or
- set a false value.

packages/crowdstrike/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1045,6 +1079,23 @@ processors:
...
+      tag: reconstruct_is_synthetic_quarantine_disposition_from_pattern_disposition_details
+      lang: painless
+      if: ctx.crowdstrike?.alert?.is_synthetic_quarantine_disposition == null && ctx.json?.pattern_disposition_details instanceof Map
+      source: |
+        if (ctx.crowdstrike == null) {
+          ctx.crowdstrike = [:];
+        }
+        if (ctx.crowdstrike.alert == null) {
+          ctx.crowdstrike.alert = [:];
+        }
```
chrisberkhout, line 1092: As above

packages/crowdstrike/data_stream/alert/_dev/test/pipeline/test-alert.log
```diff
@@ -1,3 +1,4 @@
...
...9034413","product":"epp","quarantined_files":[{"filename":"\\Device\\Volume3\\Users\\yuvraj.mahajan\\AppData\\Local\\Temp\\Temp3cc4c329-2896-461f-9dea-88009eb2e8fb_pfSenseFirewallOpenVPNClients-20230823T120504Z-001.zip\\pfSenseFirewallOpenVPNClients\\Windows\\openvpn-cds-pfSense-UDP4-1194-pfsense-install-2.6.5-I001-amd64.exe","id":"2ce412d17b334ad4adc8c1c54dbfec4b_b26a6791b72753d2317efd5e1363d93fdd33e611c8b9e08a3b24ea4d755b81fd","sha256":"b26a6791b72753d2317efd5e1363d93fdd33e611c8b9e08a3b24ea4d755b81fd","state":"quarantined"}],"scenario":"NGAV","severity":30,"sha1":"0000000000000000000000000000000000000000","sha256":"b26a6791b72753d2317efd5e1363d93fdd33e611c8b9e08a3b24ea4d755b81fd","show_in_ui":true,"source_products":["FalconInsight"],"source_vendors":["CrowdStrike"],"status":"new","tactic":"MachineLearning","tactic_id":"CSTA0004","technique":"Adware/PUP","technique_id":"CST0000","timestamp":"2023-11-03T18:00:22.328Z","tree_id":"1931778","tree_root":"38687993","triggering_process_graph_id":"pid:2ce4124ad4adc8c1c54dbfec4b:399748687993","type":"ldt","updated_timestamp":"2023-11-03T19:00:23.985007341Z","user_id":"S-1-5-21-1909377054-3469629671-4104191496-4425","user_name":"mohit.jha"}
```
chrisberkhout, line 4: This new example contains
```
"ioc_values": [],
```
Does that exist post deprecation?

chrisberkhout (approved):
Looks good.
One comment in a couple of places.
Question about new input data.

packages/crowdstrike/data_stream/alert/_dev/test/pipeline/test-alert.log
efd6: We have no cases to test against. I left it in to ensure that we continue to function correctly with the current state. I have added another test case which is identical to this one, but with that field removed.

PR #11299, opened 2024-10-01
[Santa] Make the ingest pipeline less restrictive with unexpected log formats

chemamartinez:
## Proposed commit message

The current implementation of the ingest pipeline was based on a `grok` processor with only three patterns, so it fails every time that a field is missing or a new one is added to the incoming logs.

After several reports of pipeline errors, this PR changes the parsing of the fields relying on a `kv` processor that makes the parsing less restrictive. After the `kv` processor, known fields are mapped as they were before, and unknown fields will be added into the `santa` root field. IMO there is no risk of mapping explosion because as can be seen in the example logs, there are not many different fields.

## Checklist


packages/santa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -12,11 +12,147 @@ processors:
...
-        - '\[%{TIMESTAMP_ISO8601:timestamp}\] %{NOT_SEPARATOR:log.level} santad: action=%{NOT_SEPARATOR:santa.action}\|mount=%{NOT_SEPARATOR:santa.disk.mount}?\|volume=%{NOT_SEPARATOR:santa.disk.volume}\|bsdname=%{NOT_SEPARATOR:santa.disk.bsdname}?(\|fs=%{NOT_SEPARATOR:santa.disk.fs})?(\|model=%{NOT_SEPARATOR:santa.disk.model}?)?(\|serial=%{NOT_SEPARATOR:santa.disk.serial}?)?(\|bus=%{NOT_SEPARATOR:santa.disk.bus}?)?(\|dmgpath=%{NOT_SEPARATOR:santa.disk.dmgpath}?)?(\|appearance=%{TIMESTAMP_ISO8601:santa.disk.appearance})?'
+        - '\[%{TIMESTAMP_ISO8601:_tmp.timestamp}\] %{NOT_SEPARATOR:log.level} santad: %{GREEDYDATA:_tmp.message}'
       pattern_definitions:
         NOT_SEPARATOR: '[^\|]+'
+
+  - kv:
+      field: _tmp.message
+      target_field: santa
+      tag: kv_santa_fields
+      field_split: "\\|"
```
efd6, line 23:
```suggestion
      field_split: '\|'
```

packages/santa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -12,11 +12,147 @@ processors:
...
+      value_split: "="
+      ignore_missing: true
+      on_failure:
+        - append:
+            field: error.message
+            value: "fail-{{{ _ingest.on_failure_processor_tag }}}"
```
efd6, line 29: Do we need this as well as the message that we will get from the `fail` below?

efd6 (approved):

PR #11309, opened 2024-10-02
[ti_anomali] Support the ThreatStream API

### LINKED ISSUE #9610: [Anomali] Support ThreatStream API
Our Anomali integration currently relies on their Integrator SDK via an Elastic challenge. Based on user feedback, this can be a challenge to configure and maintain. 

Anomali Threatstream now provides the ability to share threat indicators via their TAXII service or a Direct API integration, both avoiding the need to deploy and rely on their integrator SDK. 

Research and collaboration with Anomali is needed to determine a path towards making our integration easier to configure and. maintain.

chrisberkhout:
## Proposed commit message

```
[ti_anomali] Support the ThreatStream API

A new data stream is added that uses the CEL input to fetch data from
the Anomali ThreatStream API's intelligence endpoint.

This new data stream should be preferred over the existing data stream
for new users, because the existing data stream relies on the
additional Elastic Extension software, which sends data to the
integration via the HTTP Endpoint input.

The two data sources are roughly equivalent, although there are
significant differences in the data format and some other details.

Support for fetching remote indicators is added as an advanced option.
Using this on the test instance shows no differences in results, but
it may be useful for on-prem users.

The `date_first` and `date_last` (indicator first and last seen) fields
are not available from the API endpoint and the time fields differ.

Variables and fields related to the IOC expiry transform have been
changed to be more clear.

The old pipeline handles a smaller number of `itype` values. The new
pipeline extends this to list all documented values, but they remain
unhandled in some cases where the mapping to a STIX indicator type is
unclear. All indicator types for which example data exists have been
covered in pipeline tests.

For fields that aren't mapped to ECS, the new data stream uses field
names that match the API response rather than matching them to the
legacy field names.

There was some system test setup and dashboard filtering for a
datastream removed earlier (#4220), and that is now removed.

Dashboard filters were updated and deprecated controls were switched to
current equivalents. Kibana tags were removed. Screenshots have been
added.
```

## Checklist

Ask me for test credentials and API documentation.
There is a [spreadsheet](https://docs.google.com/spreadsheets/d/1s0QMHCDZ97ZrFyOuT0n7lLdBXvAf605vwiblIrdeZHI/edit?usp=sharing) with some mapping details.

## Related issues

- Closes #9610

## Screenshots

Please see the screenshots added to the integration.


packages/ti_anomali/_dev/build/docs/README.md
```diff
@@ -1,42 +1,53 @@
...
 
 ### Destination index versioning and deleting older versions
-The destination indices created by the transform are versioned with an integer suffix such as `-1`, `-2`. Example index name - `logs-ti_anomali_latest.threatstream-1`. 
 
-Due to schema changes on destination index, the versioning on it could be bumped. For example, in integration version `1.15.1`, the destination index  is changed to `logs-ti_anomali_latest.threatstream-2` from `logs-ti_anomali_latest.threatstream-1`. 
+The destination indices created by the transform are versioned with an integer suffix such as `-1`, `-2`, for example, `logs-ti_anomali_latest.intelligence-1`.
+
+Due to schema changes in the destination index, its version number may be incremented.
+
+When this happens, the transform does not have the functionality to auto-delete the old index, so users must to delete this old index manually. This is to ensure that duplicates are not present when using wildcard queries such as `logs-ti_anomali_latest.intelligence-*`. To delete an old index Please follow the steps below (either for `intelligence` as below, or for the older `threatstream` equivalents):
```
efd6, line 26:
```suggestion
When this happens, the transform does not have the functionality to auto-delete the old index, so users must delete this old index manually. This is to ensure that duplicates are not present when using wildcard queries such as `logs-ti_anomali_latest.intelligence-*`. To delete an old index, follow the steps below (either for `intelligence` as below, or for the older `threatstream` equivalents):
```

packages/ti_anomali/data_stream/intelligence/_dev/test/pipeline/test-common-config.yml
```diff
@@ -0,0 +1,9 @@
+fields:
+  tags:
+    - preserve_original_event
+  _conf:
+    ioc_duration_before_deletion: 85d
+dynamic_fields:
+  "event.ingested": "^.*$"
```
efd6, line 7: Do we need anchors here?

packages/ti_anomali/data_stream/intelligence/_dev/test/pipeline/test-common-config.yml
```diff
@@ -0,0 +1,9 @@
+fields:
+  tags:
+    - preserve_original_event
```
efd6, line 3: ~Does this do anything?~

Ah, I see, it only has an effect in system tests. Maybe remove it from here?

packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,105 @@
...
+  					"cursor": (has(state.cursor) ? state.cursor : {}).with(
+  						{
+  							?"last_update_id": (body.objects.size() > 0) ? optional.of(body.objects[body.objects.size() - 1].update_id) : optional.none(),
+  						}
+  					),
+  				}
+  			)
+  		)
+  	:
+  		state.with(
```
efd6, line 73: Suggest hoisting this to the top of the expression.

packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,105 @@
...
+  fields:
+    - api_key
+program: |-
+  request(
+  	"GET",
+  	state.url.trim_right("/") + "/api/v2/intelligence/?" + {
+  		"order_by": ["update_id"],
+  		"modified_ts__lt": [(now - duration(state.delay)).format(time_layout.RFC3339)],
+  		"limit": [string(state.page_size)],
+  		?"search_after": (has(state.cursor) && has(state.cursor.last_update_id)) ? optional.of([string(int(state.cursor.last_update_id))]) : optional.none(),
```
efd6, line 41:
```suggestion
  		?"search_after": has(state.?cursor.last_update_id) ? optional.of([string(int(state.cursor.last_update_id))]) : optional.none(),
```
From your finding, this could also be rewritten as `?"search_after": state.?cursor.last_update_id.optFlatMap(optional.of([string(int(state.cursor.last_update_id))])),` I think.

packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,105 @@
...
+  		bytes(resp.Body).decode_json().as(body,
+  			state.with(
+  				{
+  					"events": body.objects.map(obj,
+  						{"json": obj}.with(
+  							(has(state.preserve_original_event) && state.preserve_original_event) ?
+  								{"event": {"original": obj.encode_json()}}
+  							:
+  								{}
+  						)
```
efd6, line 61:
```suggestion
  						{
  							"json": obj,
  							?"event": (state.?preserve_original_event.orValue(false)) ?
  								optional.of({"original": obj.encode_json()})
  							:
  								optional.none(),
  						}
```
The removing the `with` saves map allocations.

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,552 @@
+---
+description: Pipeline for parsing Anomali ThreatStream Intelligence
+processors:
+
+  - set:
+      field: ecs.version
+      value: '8.11.0'
+  - set:
+      field: event.kind
+      value: enrichment
```
efd6, line 10: Constant keywords?

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,552 @@
...
+        trojan_ipv6: ipv6-addr
+        trojan_url: url
+        vpn_domain: domain-name
+        vps_ip: ipv4-addr
+        vps_ipv6: ipv6-addr
+        whois_bulk_reg_email: email-addr
+        whois_privacy_domain: domain-name
+        whois_privacy_email: email-addr
+      source: >
+        if (params.containsKey(ctx.json.itype) && params[ctx.json.itype] != null) {
```
efd6, line 262: Is there a reason to have a present null being meanigful? Is this for detecting an unknown itype?

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,552 @@
...
+            ctx.threat.indicator = [:];
+          }
+          ctx.threat.indicator.type = params[ctx.json.itype];
+        } else {
+          throw new IllegalArgumentException("indicator type not found from itype");
+        }
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Unable to determine STIX 2.0 indicator type from itype "{{{ json.itype }}}": {{{ _ingest.on_failure_message }}}'
```
efd6, line 276: I was wondering if this could be written easily to avoid the use of exceptions. It cannot, and this is likely to be infrequent.

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,552 @@
...
+
+  # Detect ipv6 for ipv4-addr types
+  - set:
+      field: threat.indicator.type
+      value: ipv6-addr
+      if: 'ctx.threat?.indicator?.type == "ipv4-addr" && ctx.json.ip != null && ctx.json.ip.contains(":")'
```
efd6, line 282:
```suggestion
      if: ctx.threat?.indicator?.type == "ipv4-addr" && ctx.json.ip != null && ctx.json.ip.contains(":")
```
(similar below)

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
efd6, line 1: Automagically add tags?

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,552 @@
...
+
+  # Map IP geolocation fields.
+  - convert:
+      field: json.latitude
+      target_field: threat.indicator.geo.location.lat
+      type: double
+      if: "ctx.json?.latitude != null && ctx.json?.longitude != null"
```
efd6, line 289:
```suggestion
      if: ctx.json?.latitude != null && ctx.json.longitude != null
```

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,552 @@
...
+      field: json.tlp
+      target_field: threat.indicator.marking.tlp
+      ignore_missing: true
+  - uppercase:
+      field: threat.indicator.marking.tlp
+      ignore_missing: true
+  - set:
+      field: threat.indicator.marking.tlp
+      value: AMBER
+      if: 'ctx.threat?.indicator?.marking?.tlp == null && !ctx.json.is_public'
```
efd6, line 319: Do we know that `json.is_public` is always a boolean?

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,552 @@
...
+        if (value == null) {
+          ctx.threat.indicator.confidence = "Not Specified";
+        } else if (value <= 0.0 || value > 100.0) {
+          ctx.threat.indicator.confidence = "None";
+        } else if (value >= 1.0 && value <= 29.0) {
+          ctx.threat.indicator.confidence = "Low";
+        } else if (value >= 30.0 && value < 70) {
+          ctx.threat.indicator.confidence = "Medium";
+        } else {
+          ctx.threat.indicator.confidence = "High";
```
efd6, line 346:
```suggestion
        } else if (value <= 0.0 || 100.0 < value) {
          ctx.threat.indicator.confidence = "None";
        } else if (value < 30.0) {
          ctx.threat.indicator.confidence = "Low";
        } else if (value < 70.0) {
          ctx.threat.indicator.confidence = "Medium";
        } else {
          ctx.threat.indicator.confidence = "High";
```
Currently this will give a confidence of "High" for 0 < values < 1 or 29 < values < 30.

Reordering of first guard to align with number line (read, "value is outside the range [0,100)").

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,552 @@
...
+        }
+
+  # Convert asn field.
+  - convert:
+      field: json.asn
+      target_field: threat.indicator.as.number
+      type: long
+      ignore_missing: true
+      if: "ctx.json.asn != null && ctx.json.asn != ''"
```
efd6, line 355: Previously `ctx.json` has been null-guarded. From here down it's not. We know that it is always present, so maybe remove that from above and add a comment to the effect that this is an invariant.

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: I wanted to have the map contain every `itype` value that's in the API documentation, even though some of them aren't currently mapped to STIX indicator types and we don't have example documents for them.

packages/ti_anomali/data_stream/intelligence/_dev/test/pipeline/test-common-config.yml
chrisberkhout: No. Removed.

packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
chrisberkhout: I personally prefer it this way, but I think it's basically a matter of style and I think it's more important is to be consistent so I've hoisted it to the top.

The reason I prefer the nested style is because I like to be able to look at small blocks of the expression and see that it returns _this_ or _that_, where each small block includes everything. In hoisted style I need to remember that a far-away line means the block's result is combined with something else.

packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
chrisberkhout: Good idea. I changed it to
```
?"search_after": state.?cursor.last_update_id.optMap(id, [string(int(id))]),
```
It can be `optMap` rather than `optFlatMap`, and the value if present gets another name inside `optMap`.

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Yeah, I try to avoid them. Here it would need to check and create `error`, and `error.message`, then append to it. I think it's okay as is. It's not silencing failures that could have been handled.

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Yep. That's what it is in the documentation, and in the responses.

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Yes for `ecs.version`. That's done.

The `event.kind` value changes in the `on_failure` handler. The other categorization fields must have array values, which is a problem:
```
Error: error running package pipeline tests: could not complete test run: creating fields validator for data stream failed (path: /home/chrisberkhout/Code/integrations/packages/ti_anomali/data_stream/intelligence, test case file: test-itype-apt-domain.json): can't load fields from directory (path: /home/chrisberkhout/Code/integrations/packages/ti_anomali/data_stream/intelligence/fields): unmarshalling field body failed: yaml: unmarshal errors:
  line 18: cannot unmarshal !!seq into string
  line 31: cannot unmarshal !!seq into string
```

packages/ti_anomali/_dev/build/docs/README.md
chrisberkhout: Oh, I did this then accidentally undid it. Fixed now.

efd6 (approved):

packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
andrewkroh: For this field we would need to replace our current `set` processors with a `remove` processor to ensure that the Agent isn't adding its own `ecs.version` to the `_source`.

```
- remove:
    field: ecs.version
    ignore_missing: true
```

```
# base-fields.yml
- field: ecs.version
  external: ecs
  type: constant_keyword
  value: 8.11.0
```

chrisberkhout:
@chemamartinez asked:
> @chrisberkhout do you think is a good opportunity to completely remove the deprecated data stream? or still too soon?

@jamiehynds said:
> Anomali: TBD. Pinged Anomali to ask about adoption of the Anomali Integrator for SIEM integrations. I think we may need to revert the deprecation, and continue to support both the Integrator SDK and API, but will update once we hear back from Anomali.

It makes sense to coordinate with Anomali to ensure we don't remove the legacy data stream while the Elastic Extension software is still being offered to users. In general the new data stream is better and easier to set up, so I look forward to removing the legacy one if Anomali gives the go-ahead.

PR #10910, opened 2024-08-28
[Draft] K8s otel overview dashboard

tetianakravchenko:
## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [ ] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).

## Author's Checklist

- [ ] 

## Related issues

- 

## Screenshots

Current state:
Screenshot 2024-08-29 at 18 59 45

TODO:
- [x] add total numbers: nodes, namespaces, workloads (? - for later maybe)
    - [ ] workload is missing for now
    - [ ] add container visualisation : 
      - [x] number of restarts
      - [x] unhealthy containers (? maybe)
- [x] cluster-wide - keep colors, but use defaults 
- [x] should be tested with multi nodes
- [x] add the same panels we have with ad-hoc thing for deploy/ds/sts/pods
- [x] add workload + nodes events warning + errors
- [x] add grouping
- [x] add filtering panel
- [ ] To add a tag : OpenTelemetry sth
- [ ] Document somewhere the final dashboard
- [ ] Use the api command to import it and test it

decisions:
- not change manifests: pods allocatable and node conditions
- cronjob/jobs are not included in the workload (for now)

PR #11310, opened 2024-10-02
[Kubernetest OTEL] Follow up enhancements

tetianakravchenko:
## Proposed commit message

Please explain:

- WHAT: 
    - change logo
    - fix package description
    - add consistency to the terminology: usage vs utilisation, remove pct
    - do not rely on generic.otel datastream name: changed to filter `data_stream.dataset: *.otel`, use `metrics-*otel*` for adhoc dataviews, for events use `logs-*otel*`
- WHY:  Address review comments of the first version of the kubernetes content package  

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


tetianakravchenko:
Hey @ChrsMark in this PR I think I've addressed some of the comments you've left in https://github.com/elastic/integrations/pull/10910#pullrequestreview-2299392318, mainly in the second commit - https://github.com/elastic/integrations/pull/11310/commits/071f1efa122aa39a326bd20d358da7b8d4885703

> Could we try to be consistent with the terminology, i.e. usage vs utilization vs pct? Our panel titles etc should be aligned with the [semantic convention definitions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/metrics.md#instrument-naming). Hence we should only use usage or utilization where it applies (no need for pct, utilization is a ratio anyways).

1) chnged name of those vizualisation to `Usage`
Screenshot 2024-10-02 at 19 24 03

2) for `Node overview` vizualisations switched to utilization and changed formula for cpu to address https://github.com/elastic/integrations/pull/10910#discussion_r1756257787

> Could we consider leveraging the limits' utilization metrics as well? There was strong push-back by the community when we introduced the *.node.utilization metrics so relying only these might be risky in case the community decides to deprecate them. From my perspective these are nice to haves but we can rely to *.cpu.usage ones and the limit based utilizations.

to double-check: is this comment already updated? Now there is used both:
Screenshot 2024-10-02 at 19 27 58

> Also in the panels it would be nice if we explicitly mention what utilization we display (ratio against the node's capacity VS ratio against the limits).

done

gizas:
Just a reminder to update the id of the dashborad :`kubernetes-opentelemetry-cluster-overview`

packages/kubernetes_otel/kibana/dashboard/kubernetes_otel-70c709ff-59b9-43b7-8ef4-4c34d8890bde.json
```diff
@@ -1162,25 +1182,83 @@
...
+                                                    "label": "CPU Utilization",
+                                                    "operationType": "formula",
                                                     "params": {
                                                         "format": {
                                                             "id": "percent",
                                                             "params": {
                                                                 "decimals": 2
                                                             }
                                                         },
+                                                        "formula": "last_value(metrics.k8s.node.cpu.usage)/last_value(metrics.k8s.node.allocatable_cpu)",
```
ChrsMark, line 1195: FYI there is a [plan](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/27885#issuecomment-2242780898) to add back the `k8s.node.cpu.utilization` but it will take some time since it's happening through a deprecation process which will take a while. In that case the `k8s.node.cpu.utilization` will be a ratio against the Node's capacity not the Node's allocatable cpu (capacity>allocatable), but it can be decided later if it will be fine to switch to that metric directly or not.

ChrsMark (approved):
Thank's for addressing the comments!
LGTM

tetianakravchenko:
@gizas 
> Just a reminder to update the id of the dashborad :kubernetes-opentelemetry-cluster-overview

Note: the name of the dashboard must follow this pattern: https://github.com/elastic/package-spec/blob/v3.2.2/spec/content/kibana/spec.yml#L12, so renamed to `kubernetes_otel-cluster-overview`
https://github.com/elastic/integrations/pull/11310/commits/82c295a4480246be04ae170ad462bc4cbe280bc4

gizas (approved):

PR #11313, opened 2024-10-02
[cloudflare] Add health check to mock service

### LINKED ISSUE #10619: [Stack 8.16.0-SNAPSHOT] [cloudflare] Failing test daily: system test: default in cloudflare.logpull
- Stack version: 8.16.0-SNAPSHOT
- Package: cloudflare
- Failing test: system test: default
- DataStream: logpull
- Owners:
    - @elastic/security-service-integrations

Failure:
```
test case failed: could not find hits in logs-cloudflare.logpull-87857 data stream
```

First build failed: https://buildkite.com/elastic/integrations/builds/14060

Latest 5 failed builds:
- https://buildkite.com/elastic/integrations/builds/15425
- https://buildkite.com/elastic/integrations/builds/15615
- https://buildkite.com/elastic/integrations/builds/15904
- https://buildkite.com/elastic/integrations/builds/16100
- https://buildkite.com/elastic/integrations/builds/16554

chrisberkhout:
## Proposed commit message

```
[cloudflare] Add health check to mock service

Adds a health check to the mock service used in the `logpull` system
test. This will let docker check when the mock service is ready and
avoid starting the test too early.

The system test `interval` setting is lowered from `1h` to `2m`, which
may also help avoid test failures.
```

## Checklist

## Related issues

- Closes #10619

### DIFF
--- a/packages/cloudflare/changelog.yml
+++ b/packages/cloudflare/changelog.yml
@@ -1,9 +1,4 @@
 # newer versions go on top
-- version: "2.28.1"
-  changes:
-    - description: "Add health check to mock service."
-      type: bugfix
-      link: https://github.com/elastic/integrations/pull/11313
 - version: "2.28.0"
   changes:
     - description: "Allow @custom pipeline access to event.original without setting preserve_original_event."
--- a/packages/cloudflare/changelog.yml
+++ b/packages/cloudflare/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.28.1"
+  changes:
+    - description: "Add health check to mock service."
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/11313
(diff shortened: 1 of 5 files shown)


packages/cloudflare/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.28.1"
```
efd6, line 2: I'm wondering if this needs a released version. This is a dev-only-significant change.

efd6 (approved):
Thank you for fixing this. One query that needs to be decided.

packages/cloudflare/changelog.yml
andrewkroh: With no user facing changes we can skip releasing. So no changelog is necessary IMO.

PR #11322, opened 2024-10-03
Fix changelog in network traffic

jsoriano:
1.31.1 was released after 1.32.0, including more changes than 1.32.0. Upgrading to 1.32.0 actually causes regression as there is one change that is not included.

Adjust the changelog to represent the current state of packages, and release a new 1.32.1 version that includes the latest changes.

We detected this issue in https://github.com/elastic/package-spec/pull/808, adding validations to try to prevent this kind of problems.

### DIFF
--- a/packages/network_traffic/changelog.yml
+++ b/packages/network_traffic/changelog.yml
@@ -1,12 +1,20 @@
 # newer versions go on top
-- version: "1.31.1"
+- version: "1.32.1"
   changes:
-    - description: Add `event.module` to datastreams
+    - description: Add `event.module` to datastreams.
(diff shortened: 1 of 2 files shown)


jsoriano:
@andrewkroh @efd6 can I get a review on this? Another change was merged on the incorrect version.

In latest elastic-package we are adding a check to prevent this, network traffic is failing because of this in https://github.com/elastic/integrations/pull/11400.

nfritts (approved):

Tacklebox (approved):

jsoriano:
Thanks!

PR #11326, opened 2024-10-04
packages/salesforce: preparation for GA release

shmsr:
## Proposed commit message

Preparation for GA release.

Note: We have made the minimum stack version 8.16.0 as we have [fixed a bug](https://github.com/elastic/beats/pull/41015) when responses with non-200 status codes should be discarded and not processed as the response might not be as expected. This will help in correct debugging. Along with this change, we have added some logs that could be helpful in debugging. Currently, this will go as 1.0.0-next and only be released when the 8.16.0 stack is released to the public.

More info on `-next` tag is here: https://github.com/elastic/integrations/blob/927d060c0d1d394c1c05f84520891eee4d780d6c/docs/developer_workflow_design_build_test_integration.md#remember-to-bump-up-the-version

> We will bump the version in the manifest to 1.0.0 and remove `-next` suffix when 8.16.0 is available to the public. 

## Checklist

### DIFF
--- a/packages/salesforce/changelog.yml
+++ b/packages/salesforce/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.0.0-next"
+  changes:
+    - description: GA release of the Salesforce integration.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/<TODO>
 - version: "0.15.1"
   changes:
     - description: Improve the Salesforce integration documentation.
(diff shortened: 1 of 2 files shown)


packages/salesforce/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.0.0-next"
+  changes:
+    - description: GA release of the Salesforce integration.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/<TODO>
```
shmsr, line 6:
```suggestion
      link: https://github.com/elastic/integrations/pull/11326
```

lalit-satapathy (approved):

shmsr:
So test is failing because 8.16 is not available.

kush-elastic (approved):

PR #11332, opened 2024-10-04
Add Filestream integration

belimawr:
## Proposed commit message

This commit adds the Filestream integration

## Checklist

## Author's Checklist
 - [x] Use fingerprint as default file identity

1. Build the integration
2. Start a stack with Elastic-Package or upload the integration to an existing stack
3. Go to the "Add integrations page"
4. Make sure "Display beta integrations" is selected
5. Add the "Custom Filestream integration", configure `paths` to read from any text file
6. Deploy the Elastic-Agent enrolled in this policy
7. Ensure the data from the file is ingested.

The Custom Filestream integration uses the fingerprint file identity by default, so make sure the file used for testing has a size of at least `1024` bytes or change the fingerprint length in the integration settings.

## Related issues
- Closes https://github.com/elastic/elastic-agent/issues/5370


packages/filestream/changelog.yml
```diff
@@ -0,0 +1,5 @@
+- version: "0.0.1"
+  changes:
+    - description: Initial Release
+      type: enhancement
+      link: TODO
```
jlind23, line 5:
```suggestion
      link: https://github.com/elastic/integrations/pull/11332
```

belimawr: Thanks!

packages/filestream/data_stream/generic/agent/stream/filestream.yml.hbs
```diff
@@ -0,0 +1,146 @@
...
+{{/if}}
+
+{{#if harvester_limit }}
+harvester_limit: {{harvester_limit}}
+{{/if}}
+
+{{#if fingerprint }}
+prospector.scanner.fingerprint.enabled: true
+file_identity.fingerprint.enabled: true
+file_identity.fingerprint.offset: {{ fingerprint_offset }}
```
leehinman, line 144: do we need to protect this with `{{#if fingerprint_offset}}`, I know a user *should* set the offset if they set fingerprint, just wondering if the UI allows them to make this mistake.  Same for `fingerprint_length` below.

packages/filestream/data_stream/generic/fields/filestream.yml
```diff
@@ -0,0 +1,13 @@
+- name: log.file.inode
+  type: integer
+  description: |
+    inode of the ingested file.
+- name: log.file.device_id
+  type: integer
+  description: |
+    device ID from the device where the file is.
+- name: log.file.fingerprint
+  type: text
```
leehinman, line 10: I'm thinking this should be keyword

packages/filestream/data_stream/generic/fields/filestream.yml
```diff
@@ -0,0 +1,13 @@
+- name: log.file.inode
+  type: integer
```
leehinman, line 2: I'm thinking this should be keyword

packages/filestream/data_stream/generic/fields/filestream.yml
```diff
@@ -0,0 +1,13 @@
...
+  description: |
+    inode of the ingested file.
+- name: log.file.device_id
+  type: integer
```
leehinman, line 6: I'm thinking this should be keyword

packages/filestream/manifest.yml
```diff
@@ -0,0 +1,26 @@
+format_version: 3.1.5
+name: filestream
+title: Custom Filestream Logs
+description: Collect log data using filestream with Elastic Agent.
+type: integration
+version: 0.0.1
+conditions:
+  kibana:
+    version: ^8.15.0
```
leehinman, line 9: Any reason not to go farther back?

leehinman (commented):
Looking good.

Couple of questions in-line.  The biggest is on field types.  We probably need to double check other integrations to make sure they are defined the same across all of them.  We don't want any mismatches.

packages/filestream/data_stream/generic/agent/stream/filestream.yml.hbs
belimawr: There are default values set in the integration, they will be automatically populated with the correct value.

I'll test what happens if the user sets an empty string (aka delete the value from the text box)

packages/filestream/data_stream/generic/fields/filestream.yml
belimawr: Well, it's a number, but now you're making me think if there is any FS that reports it in a way that's not a number.

I'm not opposed to set it as keyword, after all nodoby is going to be making numeric operations with it any ways.

packages/filestream/manifest.yml
belimawr: Yes, the default for `clean_inactive` changed in `8.15.0` to `-1`, older versions do not support the negative number, and currently `0` means "clean as soon as possible", which generates plenty of data re-ingestion issues.

So to have the integration listing a default value, we need to go with `8.15.0`.

belimawr:
> Looking good.
> 
> Couple of questions in-line. The biggest is on field types. We probably need to double check other integrations to make sure they are defined the same across all of them. We don't want any mismatches.

I checked them and they're all keyword in other integrations, thanks for spotting that!

packages/filestream/manifest.yml
leehinman: 👍

leehinman (approved):

strawgate:
Can we document that users migrating from a log integration can temporarily use `ignore_inactive: since_first_start` to set the offset in the registry to the end of discovered files?

belimawr:
> Can we document that users migrating from a log integration can temporarily use `ignore_inactive: since_first_start` to set the offset in the registry to the end of discovered files?

That's a good idea @strawgate ! Where do you think we should add it? To the overview or directly to the configuration option? Or should we have a migration guide (for an ideal migration we'd have to implement some extra fetures)?

packages/filestream/data_stream/generic/agent/stream/filestream.yml.hbs
belimawr: I checked it, if the user remove both settings (offset and length), the're rendered as `null` and the default values are used.

**computed-config.yml**
```yaml
          file_identity.fingerprint.enabled: true
          file_identity.fingerprint.length: null
          file_identity.fingerprint.offset: null
```

**beat-rendered-config.yml**
```yaml
      file_identity:
        fingerprint:
            enabled: true
            length: null
            offset: null
```

Log showing the default values:
```json
{
  "log.level": "debug",
  "@timestamp": "2024-11-18T16:23:33.846Z",
  "message": "fingerprint mode enabled: offset 0, length 1024",
  "component": {
    "binary": "filebeat",
    "dataset": "elastic_agent.filebeat",
    "id": "filestream-default",
    "type": "filestream"
  },
  "log": {
    "source": "filestream-default"
  },
  "ecs.version": "1.6.0",
  "log.logger": "scanner",
  "log.origin": {
    "file.line": 311,
    "file.name": "filestream/fswatch.go",
    "function": "github.com/elastic/beats/v7/filebeat/input/filestream.newFileScanner"
  },
  "service.name": "filebeat"
}
```

strawgate:
I think there's some prior art in having migration information in the overview so that makes sense to me.

belimawr:
> I think there's some prior art in having migration information in the overview/readme so that makes sense to me.

I like the idea, I'm thinking if we should involve the docs team and questioning what is the best way to communicate it.

I tried to write something, however we're not at a point where we can provide provide a migration guide without data loss, while those options can help with preventing, or reducing, data duplication, there is still the possibility from data loss between the last line ingested by the Log input and when Filestream actually starts. So I fear it could cause confiscation among our users.

We could merge the PR as is and work on a better documentation for migration that will be released when it goes GA or close to it. What do you think?

strawgate:
When I made a version of this in the past I landed on including language like:
> As `Filestream` configures a new input, configuring it to collect from a file that was previously collected by `Custom Logs (Legacy)` will result in duplicate data. You may wish to configure `ignore_older` or temporarily set `ignore_inactive: since_first_start` to limit the amount of duplicate data collected.

I think we could:
1. Start with something like that and refine later with docs links, or
2. At least include a warning about duplicate data

Do we already have an issue created for improving the migration information?

belimawr:
I love the wording @strawgate ! I'll add it to the overview page and create an issue to follow up on improving it.

belimawr:
@strawgate, I've added this bit at the end of the README (https://github.com/elastic/integrations/pull/11332/commits/89429cec4e3a528fb8fca588ef2f6f9544a124a9):
> As Filestream configures a new input, configuring it to collect data
from a file that was previously collected by Custom Logs integration
will result in duplicate data. You may wish to configure
`ignore_older` or temporarily set `ignore_inactive: since_first_start`
to limit the amount of duplicate data collected.
>
>If the Custom Logs integration is removed and the Custom Filestream
Logs is added in the same policy change, there risk of data being
missed between the last entry ingested by the Custom Logs and the
first one ingested by the Custom Filestream Logs.

What do you think?

strawgate:
That seems good to me. Though I think we should rename `Custom Logs` to `Custom Logs (Legacy)` or something similar -- @kpollich ? If we are going to rename it we should reference the updated name in the readme here too.

belimawr:
My understanding is that we'll rename the Custom Logs integration once the Custom Filestream is GA, this PR releases it in technical preview.

kpollich:
+1 on renaming to `Custom Logs (Legacy)` once the filestream integration is GA

PR #11366, opened 2024-10-08

kaiyan-sheng:
## Proposed commit message

This PR adds aws.metrics_names_fingerprint which is a hash of the list of metric names exist in each document. For metrics coming in from Firehose, there can be cases two documents have the same timestamp, dimension, namespace, accountID, exportARN and region BUT from two different requests. With adding the aws.metrics_names_fingerprint as TSDB dimension, we are able to ingest these two documents into ES.

## Checklist

## Related issues

This PR is to add the change from https://github.com/elastic/integrations/pull/11281 into main branch.


agithomas (approved):
LGTM!

PR #11367, opened 2024-10-08
Problem child ingest pipeline bug

r0ot:
## Proposed commit message

The painless script `if` block serving as the entry point into the problem child inference pipeline has a bug in its logic checking for the appropriate host operating system. It separates its checks for various sub-fields of `host.os` and it's actual accessing of those fields from the `ctx` object, potentially leading to errors trying to access keys that don't exist.

For example, if a document contains `host.os.family` but doesn't contain `host.os.type`, this painless script will raise an error instead of simply failing.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/problemchild/changelog.yml
+++ b/packages/problemchild/changelog.yml
@@ -1,5 +1,5 @@
 # newer versions go on top
-- version: "2.1.6"
+- version: "2.2.0"
   changes:
     - description: Improved notation of painless script if block in ingest pipeline
       type: enhancement
--- a/packages/problemchild/changelog.yml
+++ b/packages/problemchild/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.1.6"
+  changes:
+    - description: Fixed bug in ingest pipeline
+      type: bugfix
+      link: 
 - version: "2.1.5"
   changes:
     - description: Add fields for integration package testing
--- a/packages/problemchild/changelog.yml
+++ b/packages/problemchild/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Fixed bug in ingest pipeline
       type: bugfix
-      link: 
+      link: https://github.com/elastic/integrations/pull/11367
 - version: "2.1.5"
   changes:
     - description: Add fields for integration package testing
--- a/packages/problemchild/changelog.yml
+++ b/packages/problemchild/changelog.yml
@@ -1,8 +1,8 @@
 # newer versions go on top
 - version: "2.1.6"
   changes:
-    - description: Fixed bug in ingest pipeline
-      type: bugfix
+    - description: Improved notation of painless script if block in ingest pipeline
+      type: enhancement
       link: https://github.com/elastic/integrations/pull/11367
 - version: "2.1.5"
   changes:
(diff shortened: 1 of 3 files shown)


sodhikirti07:
@r0ot Could you provide an example document that is causing the pipeline errors?

r0ot:
> @r0ot Could you provide an example document that is causing the pipeline errors?

I don't have a real-world document I could provide but the example I gave in my description should be sufficient. If the user's datasource isn't elastic endpoint and only provided `host.os.family` or `host.os.platform` without `host.os.type` it would trigger this bug.

susan-shu-c:
Hi, @r0ot we are testing this, can you also share what Elastic stack version you are using, as well as the ProblemChild/Living off the Land Detection package you are on?

r0ot:
Upon testing, this doesn't raise an error like I thought, it rather just returns false. The notation used here could be cleaned up with `?.` accesses to make the logic more concise and avoid confusion in the future. I've updated this PR with the updated notation.

So not a bug but an improvement.

packages/problemchild/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.1.6"
+  changes:
+    - description: Improved notation of painless script if block in ingest pipeline
+      type: enhancement
```
andrewkroh, line 5: `bugfix`?

That would align to your version choice and the bugfix label on the PR.

packages/problemchild/elasticsearch/ingest_pipeline/problem_child_ingest_pipeline.yml
```diff
@@ -2,7 +2,7 @@
 description: "Pipelines for ProblemChild detection"
 processors:
   - pipeline:
-      if: ctx.containsKey('event') && ctx['event'].containsKey('kind')  && ctx['event'].containsKey('category') && ctx['event']['kind'] == 'event' && ctx['event']['category'].contains('process') && ctx.containsKey('host') && ctx['host'].containsKey('os') && (ctx['host']['os'].containsKey('type') || ctx['host']['os'].containsKey('family') || ctx['host']['os'].containsKey('platform')) && (ctx['host']['os']['type'] == 'windows' || ctx['host']['os']['type'] == 'Windows' || ctx['host']['os']['family'] == 'windows' || ctx['host']['os']['family'] == 'Windows' || ctx['host']['os']['platform'] == 'windows' || ctx['host']['os']['platform'] == 'Windows')
+      if: ctx?.event?.kind == 'event' && ctx?.event?.category?.contains('process') && (ctx?.host?.os?.type?.toLowerCase() == 'windows' || ctx?.host?.os?.family?.toLowerCase() == 'windows' || ctx?.host?.os?.platform?.toLowerCase() == 'windows')
```
andrewkroh, line 5: `ctx` can never be null in this context so omit the null-safe operator (i.e. `ctx.event` instead of `ctx?.event`)

qcorporation:
@r0ot, would it be possible to email me (my profile has my contact information), Elastic has an ask regarding your fork

andrewkroh (approved):
LGTM

mgarzon (approved):

PR #11368, opened 2024-10-08

kaiyan-sheng:
## Proposed commit message

This PR adds aws.metrics_names_fingerprint which is a hash of the list of metric names exist in each document. For metrics coming in from Firehose, there can be cases two documents have the same timestamp, dimension, namespace, accountID, exportARN and region BUT from two different requests. With adding the aws.metrics_names_fingerprint as TSDB dimension, we are able to ingest these two documents into ES.

## Checklist

## Related issues

backport https://github.com/elastic/integrations/pull/11281 to 2.25


Kavindu-Dodan (approved):

kcreddy (approved):

agithomas (approved):
LGTM!

packages/aws/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.25.1"
+  changes:
+    - description: Add aws.metrics_names_fingerprint field and mark it as a dimension.
```
constanca-m, line 4: @kaiyan-sheng I would include in the description this part you added in the issue:
> For metrics coming in from Firehose, there can be cases two documents have the same timestamp, dimension, namespace, accountID, exportARN and region BUT from two different requests.

Does it cause document loss because of TSDB? I would include it.

I think it would help to navigate better when looking at the changelog in case this appears in a SDH in the future.

kaiyan-sheng: @constanca-m Yes it does cause document loss because of TSDB. Let me add it in the aws firehose integration since that's where the actual fingerprint processor is added: https://github.com/elastic/integrations/blob/main/packages/awsfirehose/data_stream/metrics/elasticsearch/ingest_pipeline/default.yml#L29 
changelog: https://github.com/elastic/integrations/blob/main/packages/awsfirehose/changelog.yml#L7

PR #11372, opened 2024-10-09
entityanalytics_{ad,entra_id,okta}: don't include input-private details to index

efd6:
## Proposed commit message

The event.action field is an implementation detail that has an unfortunate name that could mislead users; the values held for entities do not relate to security details, but only to internal accounting. So remove them.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/entityanalytics_ad/changelog.yml
+++ b/packages/entityanalytics_ad/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.3.2"
+  changes:
+    - description: Don't leak collection implementation details.
+      type: bugfix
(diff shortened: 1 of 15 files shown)


kcreddy (approved):
LGTM, but looking for some clarification.

Does it make sense to remove these values from the input itself, if its not going to be used in integration? Or maybe use a different field instead of `event.action` to store these internal information?

efd6:
There are two reasons I didn't do that: the details of the documents are published in the filebeat docs, and include those fields (I think this is unfortunate, but given that they are there removing the fields would be a breaking change — we can consider this), and I wanted to make sure that the fields are not used in the integration docs now rather than waiting for a new stack release.

PR #11409, opened 2024-10-15
forgerock: fix endTime query parameter handling

### LINKED ISSUE #11331: [forgerock]: Logs being pulled for wrong date time range
### Integration Name

ForgeRock [forgerock]

### Dataset Name

_No response_

### Integration Version

1.18.2

### Agent Version

8.15.2

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.15.2

Elastic agent POD

### Software/API Version

_No response_

### Error Message

There are no error messages however we see that elastic integration with ForgeRock is building request params with wrong values.

### Event Original

_No response_

### What did you do?

We upgraded ForgeRock integration from 1.18.0 to 1.18.2

### What did you see?

we see in elastic logs that no events were published with message "request finished: 0 events published" but when we directly query ForgeRock we can see that logs are being published and are available.

In elastic logs we also see that we have request from 2. October with query string parameters in request begintime 3.October and endtime 4.OctoberImage

Logs from ForgeRock being ingested.

### Anything else?

We think it could be related to this pull request https://github.com/elastic/integrations/pull/11240/commits/b4e121092fa95c86bb941e089c2cb60f4142e19a

We think it is building endtime parameter incorrectly.

Before this PR endtime was built as 
value: '[[formatDate (now) "2006-01-02T15:04:05-07:00"]]'

so let's say last request was on October 1st at 12:00. This date and time will be used as begintime on next request and if we are running this every 10 minutes endtime in new request will be October 1st at 12:10 and it will pull only logs from these 10 minutes.

However after the change endtime is built as
value: '[[formatDate ((parseDate .cursor.last_timestamp "2006-01-02T15:04:05-07:00").Add (parseDuration "24h"))  "2006-01-02T15:04:05-07:00" ]]'

this means that for same example next request will have begintime on October 1st 12:00 and instead of trying to pull logs for next ten minutes it will put endtime as October 2nd at 12:00. This would also pull logs only for 10 minutes. but next request would use this date and time as begintime. This would cause request on October 1st at 12:10 to create request with begintime October 2nd at 12:00 and endtime as October 3rd at 12:00 which is in the future and no logs would be retrieved. This is what we see is happening.

As a proposal to fix this issue we think it should utilize both old and new logic (which is trying to address different issue). If current date and time (old logic) is less then 24 hours before begintime then use that as endtime. Otherwise use new logic where endtime will be calculated as time span of 24 hours from begintime to avoid ForgeRock rules violation of span bigger then 24 hours.

efd6:
## Proposed commit message

```
In #11240 we attempted to fix preparation of endTime where it could be more than
24 hours after the startTime parameter, a case that is invalid for the API. To
do this, we just always added 24 hours to the startTime on the basis that the
API would crop at the present moment, and — notably — under the assumption that
the returned document would be used to construct the next startTime. Clearly
this is not the case, since we use the actual endTime parameter to construct the
next startTime; the API does not provide an HTTPJSON-easy way to get the last
timestamp. So set the endTime query parameter to the earliest of now and 24
hours after startTime.

Queries with an end time of now may not get all documents up to now. Since we
are using the query parameter to define the end of the range of documents that
have definitively been retrieved, this means that we will never try to get these
missed documents. We could us the maximum timestamp of the collected documents
to define this range, but that less easy to do here, so just set the end of the
time range to be some short period before now. Ten seconds is chosen as a
conservative value (the report claimed milliseconds of documents were lost).
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #11331


chrisberkhout (approved):

igrujic-ap:
Hi @efd6 I see that this PR is aiming to solve issue that I reported. However I noticed one additional thing that I noticed so not sure if this PR is going to solve it or should I create a new issue. When we use "now" as endtime, events that happened few millisecond before it are not immediately exposed in ForgeRock audit logs and they get lost as next request starts with that end date and these logs belong to previous request. My suggestion would be to reduce "now" (when used) by few milliseconds or even seconds. In this case those logs would be picked up with next request. Thanks

efd6:
@igrujic-ap Thanks for the heads up. I will include that in the change here. There are two options:
1. set the `endTime` back a few seconds when it is constructed as you propose — this moves our endpoint to before what looks like a refractory period.
2. set the `startTime` back a few seconds when it is derived from the stored cursor when it is used — this repeats collection to before the refractory period. (I think this one is safer at the cost of potentially collecting some data twice; we would need to add a fingerprinting processor to handle this)

igrujic-ap:
Thank you @efd6. Do you know release schedule for this fix? We would like to test and let you know if everything is working as expected.

chrisberkhout (approved):
Good

PR #11414, opened 2024-10-15
[Cloud Security] release Wiz version 2.0.0

maxcold:
## Proposed commit message

Changes: 
- removing `rule.references` non-ecs as this field is not required anymore by Kibana https://github.com/elastic/security-team/issues/10793 . The field `rule.reference` should be used instead
- mappings for rule.* fields which were `text` changed to `keyword` to match ecs
- `resource.id` is provided from `id` when `provider_id` is missing on Wiz
- `external: esc` is added for ECS fields
- add `cloud.account.name` mapping to vulnerability cdr transform destination

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- relates to https://github.com/elastic/security-team/issues/10794
- fixes https://github.com/elastic/security-team/issues/10595
- fixes https://github.com/elastic/security-team/issues/10811
- fixes  https://github.com/elastic/security-team/issues/10819
- fixes https://github.com/elastic/security-team/issues/10875


packages/wiz/changelog.yml
```diff
@@ -1,26 +1,21 @@
 # newer versions go on top
-- version: "1.9.0-preview05"
+- version: "2.0.0"
   changes:
+    - description: Remove rule.references field and adjust rule.* mappings.
+      type: enhancement
```
andrewkroh, line 5: > and adjust rule.* mappings.

When I read this I'm wonder what adjustments were made.

I also think the changelog should tell user how to deal with the breaking change if they were using the removed field.

```suggestion
    - description: Remove rule.references field and its mapping. Please use the ECS rule.reference field instead.
      type: breaking-change
```

packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml
```diff
@@ -11,7 +11,5 @@
       type: text
     - name: remediation
       type: text
-    - name: references
-      type: text
     - name: reference
       type: text
```
andrewkroh, line 15: The `rule` fields are part of ECS. Anywhere we have an ECS field we should depend on the ECS definition through the use of `external: ecs` to ensure that we always have consistent mappings. This ensures that there are never data type conflicts between data streams for ECS fields.

This is what it would look like to use `external: ecs` for the ECS fields in wiz -- https://github.com/elastic/integrations/commit/d3d335c73bd32df269c71719a2f77bce03a57281.

However, after doing that clean-up it becomes easier to spot other problems.

1) There are two problematic field definitions that are conflicting with ECS.

    ```
    packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml:10:7 rule.description field declared as type text conflicts with the ECS data type keyword (conflict)
    packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml:14:7 rule.reference field declared as type text conflicts with the ECS data type keyword (conflict)
    ```

2) rule.remediation has multiple data types (keyword, text)

    ```
    packages/wiz/elasticsearch/transform/latest_cdr_misconfigurations/fields/rule.yml:12:7 keyword (conflict)
    packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml:12:7 text (conflict)
    ```

3) duplicate definitions of the same field in a data stream

    ```
    packages/wiz/elasticsearch/transform/latest_cdr_misconfigurations/fields/ecs.yml:1:3 event.created is declared 2 times (duplicate)
      packages/wiz/elasticsearch/transform/latest_cdr_misconfigurations/fields/ecs.yml:19:3 additional definition (duplicate)
    packages/wiz/elasticsearch/transform/latest_cdr_vulnerabilities/fields/fields.yml:5:3 package.name is declared 2 times (duplicate)
      packages/wiz/elasticsearch/transform/latest_cdr_vulnerabilities/fields/fields.yml:23:3 additional definition (duplicate)
    ```

packages/wiz/changelog.yml
maxcold: thanks, makes sense!

packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml
maxcold: @andrewkroh thanks for the detailed feedback! I updated the mapping according to your suggestion. A couple of questions:
1. how do I check for conflicts with ECS? neither `lint` nor `check` produce the same output as you have in your comment
2. I changed `rule.description` and `rule.reference` to `ecs` and that means the mapping changed from `text` to `keyword` which is a breaking change as far as I understand. What should be the guidance for users in the changelog in that case?

andrewkroh: > how do I check for conflicts with ECS? 

I don't know that elastic-package has this capability yet. I have an unofficial [means](https://github.com/andrewkroh/fydler) that I occasionally use to audit the mappings.

> I changed rule.description and rule.reference to ecs and that means the mapping changed from text to keyword which is a breaking change as far as I understand. What should be the guidance for users in the changelog in that case?

It's hard to prescribe one solution because it depends on how the data is being used and what problem they encounter (types of queries or aggregations they are performing). One solution for conflicts mentioned in this [blog post](https://www.elastic.co/blog/troubleshooting-guide-common-issues-kibana-discover-load) is to apply a runtime field to the old index to make the field a keyword.

packages/wiz/data_stream/cloud_configuration_finding/fields/base-fields.yml
```diff
@@ -16,5 +13,4 @@
   description: Event dataset.
```
andrewkroh, line 13: This can use `external: ecs` too. Leave the `value` and `type` which will override the ECS `keyword` which is completely fine because constant_keyword and keyword are the same field type family.

maxcold: I'm assuming the same works for `data_stream.namespace` when we want to redefine it as `keyword` for the transform destination index, correct?

andrewkroh: Correct.

packages/wiz/changelog.yml
```diff
@@ -1,26 +1,30 @@
 # newer versions go on top
-- version: "1.9.0-preview05"
+- version: "2.0.0"
   changes:
+    - description: Add cloud.account.name mapping to latest_cdr_vulnerabilities transform destination,
```
kcreddy, line 4:
```suggestion
    - description: Add cloud.account.name mapping to latest_cdr_vulnerabilities transform destination.
```

packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml
kcreddy: Since we are adhering to ECS, should we even define the rule fields that are of `external: ecs`? The `ecs@mappings` component template should take care of them for the datastream.

packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml
```diff
@@ -2,16 +2,14 @@
   type: group
   fields:
     - name: uuid
-      type: keyword
+      external: ecs
```
kcreddy, line 5: Added comment: https://github.com/elastic/integrations/pull/11414#discussion_r1812325626 to clarify if this is even required.

kcreddy (commented):
Change `external: esc` to `external: ecs` in PR commit message.

packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml
maxcold: @kcreddy @andrewkroh Looking at the value of the `ecs@mappings` in my stack I don't see it as a replacement of `external: ecs` . Actually, I would like to understand the use of `ecs@mappings` better, as in our team we understood that it is a replacement of manual mapping, but then not everything that is available in ECS also mapped in this component template. Eg. there are no mappings for the `rule.*` fields, except the ones falling into broader mappings such as `*.name` 
```
{
  "dynamic_templates": [
    {
      "ecs_timestamp": {
        "mapping": {
          "ignore_malformed": false,
          "type": "date"
        },
        "match": "@timestamp"
      }
    },
    {
      "ecs_message_match_only_text": {
        "path_match": [
          "message",
          "*.message"
        ],
        "mapping": {
          "type": "match_only_text"
        },
        "unmatch_mapping_type": "object"
      }
    },
    {
      "ecs_non_indexed_keyword": {
        "path_match": [
          "*event.original"
        ],
        "mapping": {
          "index": false,
          "type": "keyword",
          "doc_values": false
        }
      }
    },
    {
      "ecs_non_indexed_long": {
        "path_match": [
          "*.x509.public_key_exponent"
        ],
        "mapping": {
          "index": false,
          "type": "long",
          "doc_values": false
        }
      }
    },
    {
      "ecs_ip": {
        "path_match": [
          "ip",
          "*.ip",
          "*_ip"
        ],
        "mapping": {
          "type": "ip"
        },
        "match_mapping_type": "string"
      }
    },
    {
      "ecs_wildcard": {
        "path_match": [
          "*.io.text",
          "*.message_id",
          "*registry.data.strings",
          "*url.path"
        ],
        "mapping": {
          "type": "wildcard"
        },
        "unmatch_mapping_type": "object"
      }
    },
    {
      "ecs_path_match_wildcard_and_match_only_text": {
        "path_match": [
          "*.body.content",
          "*url.full",
          "*url.original"
        ],
        "mapping": {
          "fields": {
            "text": {
              "type": "match_only_text"
            }
          },
          "type": "wildcard"
        },
        "unmatch_mapping_type": "object"
      }
    },
    {
      "ecs_match_wildcard_and_match_only_text": {
        "mapping": {
          "fields": {
            "text": {
              "type": "match_only_text"
            }
          },
          "type": "wildcard"
        },
        "unmatch_mapping_type": "object",
        "match": [
          "*command_line",
          "*stack_trace"
        ]
      }
    },
    {
      "ecs_path_match_keyword_and_match_only_text": {
        "path_match": [
          "*.title",
          "*.executable",
          "*.name",
          "*.working_directory",
          "*.full_name",
          "*file.path",
          "*file.target_path",
          "*os.full",
          "*email.subject",
          "*vulnerability.description",
          "*user_agent.original"
        ],
        "mapping": {
          "fields": {
            "text": {
              "type": "match_only_text"
            }
          },
          "type": "keyword"
        },
        "unmatch_mapping_type": "object"
      }
    },
    {
      "ecs_date": {
        "path_match": [
          "*.timestamp",
          "*_timestamp",
          "*.not_after",
          "*.not_before",
          "*.accessed",
          "created",
          "*.created",
          "*.installed",
          "*.creation_date",
          "*.ctime",
          "*.mtime",
          "ingested",
          "*.ingested",
          "*.start",
          "*.end",
          "*.indicator.first_seen",
          "*.indicator.last_seen",
          "*.indicator.modified_at",
          "*threat.enrichments.matched.occurred"
        ],
        "mapping": {
          "type": "date"
        },
        "unmatch_mapping_type": "object"
      }
    },
    {
      "ecs_path_match_float": {
        "path_match": [
          "*.score.*",
          "*_score*"
        ],
        "mapping": {
          "type": "float"
        },
        "path_unmatch": "*.version",
        "unmatch_mapping_type": "object"
      }
    },
    {
      "ecs_usage_double_scaled_float": {
        "path_match": "*.usage",
        "mapping": {
          "scaling_factor": 1000,
          "type": "scaled_float"
        },
        "match_mapping_type": [
          "double",
          "long",
          "string"
        ]
      }
    },
    {
      "ecs_geo_point": {
        "path_match": [
          "*.geo.location"
        ],
        "mapping": {
          "type": "geo_point"
        }
      }
    },
    {
      "ecs_flattened": {
        "path_match": [
          "*structured_data",
          "*exports",
          "*imports"
        ],
        "mapping": {
          "type": "flattened"
        },
        "match_mapping_type": "object"
      }
    },
    {
      "all_strings_to_keywords": {
        "mapping": {
          "ignore_above": 1024,
          "type": "keyword"
        },
        "match_mapping_type": "string"
      }
    }
  ]
}
```

packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml
maxcold: I have some concerns about it, answered in the linked comment

packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml
kcreddy: AFAIU the idea of `ecs@mappings` component template is to have a generic set of dynamic templates that works across all ECS entities, instead of defining mappings for specific entities like `rule.*`, `vulnerability.*` etc. 

In our case, following are `rule.*` fields that are defined as `external: ecs` and the corresponding dynamic template inside [`ecs@mappings`](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/ecs%40mappings.json) that is handling it:
| Field in `rule.yml`   |      Handled by dynamic template     |
|----------|:-------------:|
| rule.uuid |  `all_strings_to_keywords` |
| rule.id |    `all_strings_to_keywords`   |
| rule.name |    `ecs_path_match_keyword_and_match_only_text`   |
| rule.description |    `all_strings_to_keywords`   |
| rule.reference |    `all_strings_to_keywords`   |

Comparing the above dynamic mappings with [ECS rule fields](https://www.elastic.co/guide/en/ecs/current/ecs-rule.html), it matches correctly. 

`rule.remediation` is anyway manually defined as it is not in ECS.

kcreddy:
@maxcold, should this PR also contain a documentation update suggesting users to update `Initial Interval` according to their needs, like we discussed? or is it going to be part of Cloud Security documentation and we just link it here?

packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml
maxcold: @kcreddy now I see how `keyword` mapping is being solved in the template, thanks. Then yes, I agree, it makes sense to remove these explicit ecs mappings

packages/wiz/data_stream/cloud_configuration_finding/fields/rule.yml
maxcold: updated

maxcold:
@kcreddy regarding the docs, the work on them is in progress, when product with docs team come up with a final version we can incorporate these changes in our PRs

packages/wiz/changelog.yml
```diff
@@ -1,34 +1,36 @@
 # newer versions go on top
-- version: "1.9.0-preview07"
+- version: "2.0.0"
```
efd6, line 2: I don't think this is the correct way to do this; the versions that have been released should continue to exist.

Also, what is it that makes this a 2.0.0? What is breaking here?

efd6 (changes requested):

kcreddy (approved):
LGTM for my comments. 
Please merge after Dan's comments are addressed regarding version.

packages/wiz/changelog.yml
maxcold: what would be the correct way to do that? From a user's point of view if they never enabled "show beta integrations", they would upgrade from 1.8.1 (latest released non-preview version) to 2.0.0 and would get all the changes from `1.9.0-preview*` with the change, and as a user I would expect to see this as a part of 2.0.0 changelog

> Also, what is it that makes this a 2.0.0? What is breaking here?

this is reflected in the changelog
```
    - description: Rely on external ecs for ESC fields. rule.reference, rule.descipriton and rule.remediation changed from text to keyword.
      type: breaking-change
      link: https://github.com/elastic/integrations/pull/11414
    - description: Remove rule.references field and its mapping. Please use the ECS rule.reference field instead.
      type: breaking-change
      link: https://github.com/elastic/integrations/pull/11414
```

efd6: The change here inserts a bunch of breaking changes before a bugfix that had already been released as a 1.9.0 preview. If correct chronological order had been used and they were added as a 1.9.0, the full set of 1.9.0 changes would not have been breaking; there would have been a single new version entry looking like this:
```
- version: "1.9.0"
  changes:
    - description: Add cloud.account.name mapping to latest_cdr_vulnerabilities transform destination.
      type: bugfix
      link: https://github.com/elastic/integrations/pull/<SOMEOTHERPRNUMBER>
    - description: Default to resource id when provider_id is missing for resource.id field in cloud_configuration_finding data stream.
      type: enhancement
      link: https://github.com/elastic/integrations/pull/<SOMEOTHERPRNUMBER>
    - description: Add cloud.account.name mapping to latest_cdr_vulnerabilities transform destination.
      type: enhancement
      link: https://github.com/elastic/integrations/pull/11414
```
(leaving all the other already published versions in tact)

then you could have added this breaking enhancement:
```
- version: "2.0.0"
  changes:
    - description: Rely on external ecs for ESC fields. rule.reference, rule.descipriton and rule.remediation changed from text to keyword.
      type: breaking-change
      link: https://github.com/elastic/integrations/pull/11414
    - description: Remove rule.references field and its mapping. Please use the ECS rule.reference field instead.
      type: breaking-change
      link: https://github.com/elastic/integrations/pull/11414
```
This is not an academic issue; there is a bugfix that is being hidden here from people who may not want to upgrade across a break, and also blocks them from getting non-breaking enhancements. Admittedly, there is a fix in 1.8.something hidden in a backport branch, so the bugfix is less of a concern, but the non-breaking enhancements are still an issue.

maxcold: The problem is that 1.9.0-preview1 and 1.9.0-preview3 introduce the latest transforms, which we plan to release as part of the CDR initiative. These changes require the 8.16.x version of the stack; that's why we were asked to hold them in an open PR until closer to the 8.16 release date, as the flow with preview is not advised to be used. So, I can't release 1.9.0 just yet. Additionally, there are problems that we found while testing the CDR approach, together with the product folks, which require breaking changes before we can release transforms. That's why I'm skipping the 1.9.0 release and moving to 2.0.0, as all the changes there are bundled under CDR. We don't want users to be able to install the version with the transforms before the breaking changes are made. I can move the breaking change entries to the top of the changelog if you think that this would help.

efd6: This is deeply dissatisfying. I have removed my block.

packages/wiz/data_stream/vulnerability/fields/base-fields.yml
```diff
@@ -1,20 +1,16 @@
...
-  type: constant_keyword
-  description: Data stream dataset.
+  external: ecs
 - name: data_stream.namespace
-  type: constant_keyword
-  description: Data stream namespace.
+  external: ecs
 - name: event.module
+  external: ecs
   type: constant_keyword
```
CohenIdo, line 9: Shouldn’t we remove type: constant_keyword? 
The type is inherited from ecs.

CohenIdo (changes requested):
There are several fields configured as ecs that also have a specified type. Is there a reason for this configuration?

packages/wiz/data_stream/issue/fields/base-fields.yml
```diff
@@ -1,20 +1,16 @@
...
-  type: constant_keyword
-  description: Data stream dataset.
+  external: ecs
 - name: data_stream.namespace
-  type: constant_keyword
-  description: Data stream namespace.
+  external: ecs
 - name: event.module
+  external: ecs
   type: constant_keyword
```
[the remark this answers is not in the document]
maxcold: the type in ECS is `keyword` but we set the value `wiz`, that's why we change the type to the one which is better for performance as per recommendation from Service Integration team

packages/wiz/data_stream/issue/fields/base-fields.yml
```diff
@@ -1,20 +1,16 @@
...
-  description: Event module.
   value: wiz
 - name: event.dataset
+  external: ecs
   type: constant_keyword
```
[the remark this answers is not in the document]
maxcold: the type in ECS is `keyword` but we set the value `wiz.issue`, that's why we change the type to the one which is better for performance as per recommendation from Service Integration team

packages/wiz/data_stream/vulnerability/fields/base-fields.yml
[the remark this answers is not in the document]
maxcold: the type in ECS is keyword but we set the value wiz, that's why we change the type to the one which is better for performance as per recommendation from Service Integration team

CohenIdo (approved):

PR #11336, opened 2024-10-05
[Auditd Manager] Add Session Data option

opauloh:
## Proposed commit message

- Updating Auditd Manager manifest to include Session Data option starting from Kibana version 8.16.0.
- Updated logic in the `hbs` file to append Session Data Audit Rules and Session Data Processors when Session Data is selected.
- Updated documentation to include both manual and Toggle Switch options

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).

## How to test this PR locally

To test these changes, go to Kibana -> Integrations -> Create new integration -> upload it as a .zip; and upload the following package:

[auditd_manager-1.18.0.zip](https://github.com/user-attachments/files/17333356/auditd_manager-1.18.0.zip)

**Please note**. This Kibana [PR](https://github.com/elastic/kibana/pull/195159) should first be merged to test this integration.

## Related issues

- Closes https://github.com/elastic/security-team/issues/10330

## Screenshots

Session Data Switcher

image

elastic-agent .yml

image

Docs

image

PR #11420, opened 2024-10-15
[AuditdManager] Reverting Session Data option

opauloh:
## Proposed commit message

This PR reverts the Session Data option introduced on [this](https://github.com/elastic/integrations/pull/11336) PR.

The revert is needed because the `1.8.0` version relies on the `escape_multiline_string` helper that was introduced in Kibana in [this](https://github.com/elastic/kibana/pull/195159) PR. 

It fixes the error below when attempting to save the Auditd Manager integration version `1.8.0` in Serverless environments that don't contain the [Kibana](https://github.com/elastic/kibana/pull/195159) changes:

image

Reasoning:

1.8.0 manifest was set to be enabled Starting on Kibana 8.16+, but serverless release cadences follow a different cadence and it's on version 9.0.0 already, therefore  [this Kibana PR](https://github.com/elastic/integrations/pull/11336) must be deployed on serverless in order to `escape_multiline_string` be available.

## Checklist

### DIFF
--- a/packages/auditd_manager/data_stream/auditd/agent/stream/auditd.yml.hbs
+++ b/packages/auditd_manager/data_stream/auditd/agent/stream/auditd.yml.hbs
@@ -7,17 +7,8 @@ socket_type: '{{socket_type}}'
 immutable: {{immutable}}
 resolve_ids: {{resolve_ids}}
 failure_mode: {{failure_mode}}
-{{#if session_data}}
-audit_rules: "{{escape_multiline_string audit_rules}}
-{{escape_multiline_string "
-# Session data audit rules
--a always,exit -F arch=b64 -S execve,execveat -k exec
--a always,exit -F arch=b64 -S exit_group
--a always,exit -F arch=b64 -S setsid"}}"
-{{else}}
 {{#if audit_rules}}
 audit_rules: {{escape_string audit_rules}}
-{{/if}}
 {{/if}}
 {{#if audit_rule_files.length}}
 audit_rule_files:
@@ -42,12 +33,4 @@ publisher_pipeline.disable_host: true
 {{#if processors}}
 processors:
 {{processors}}
-{{#if session_data}}
-{{session_data_processors}}
-{{/if}}
-{{else}}
-{{#if session_data}}
-processors:
-{{session_data_processors}}
-{{/if}}
 {{/if}}
--- a/packages/auditd_manager/manifest.yml
+++ b/packages/auditd_manager/manifest.yml
@@ -1,7 +1,7 @@
 format_version: "3.0.0"
 name: auditd_manager
 title: "Auditd Manager"
-version: "1.18.0"
+version: "1.18.1"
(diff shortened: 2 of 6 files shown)


acorretti (approved):

mjwolf (approved):

packages/auditd_manager/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: "3.0.0"
 name: auditd_manager
 title: "Auditd Manager"
-version: "1.18.0"
+version: "1.18.1"
 description: "The Auditd Manager Integration receives audit events from the Linux Audit Framework that is a part of the Linux kernel."
 type: integration
 categories:
```
andrewkroh, line 7: The change to include `^9.0.0` in the kibana condition should be backed out too. As per the email to the `elastic-package` mailing list, 

"TLDR; Please don't update the kibana constraints in packages for 9.0.0 yet. We will communicate a more definitive plan in the following weeks."

packages/auditd_manager/data_stream/auditd/agent/stream/auditd.yml.hbs
```diff
@@ -7,18 +7,9 @@ socket_type: '{{socket_type}}'
 immutable: {{immutable}}
 resolve_ids: {{resolve_ids}}
 failure_mode: {{failure_mode}}
-{{#if session_data}}
-audit_rules: "{{escape_multiline_string audit_rules}}
-{{escape_multiline_string "
```
andrewkroh, line 12: Regarding `escape_multiline_string`, if we could get a `to_json` function (that doesn't [exempt string](https://github.com/elastic/kibana/blob/b4c3ab55a0680db2ec1a9d2f01051266f599e172/x-pack/plugins/fleet/server/services/epm/agent/agent.ts#L148-L151) types) this would cover many use cases of properly encoding data for use inside of YAML. The fact that the current `to_json` help exempts strings makes is almost useless (in fact, it's never used in this repo). I would love to see the string exemption removed, then we could use `to_json` here to form a properly escaped string.

For example, this would become `audit_rules: {{ to_json audit_rules }}`, and we would not have to think about escaping anything because JSON encoding is entirely valid within YAML.

packages/auditd_manager/manifest.yml
opauloh: Reverted, thanks for the info

packages/auditd_manager/data_stream/auditd/agent/stream/auditd.yml.hbs
opauloh: yes, it is similar to what I faced when I noticed that `escape_string` always wraps the string with single quotes, it doesn't allow for flexibility with concatenating YAMLs. 

I will check the possibility of enhancing the `to_json` in Kibana to remove the string exempt instead if that can provide more value, I didn't notice it wasn't being used due to this reason

PR #11441, opened 2024-10-16
[cisco_duo] Set request rate limits

chrisberkhout:
## Proposed commit message

```
[cisco_duo] Set request rate limits

The Duo Admin API has rate limiting. It doesn't return rate limit
headers, but it does enforce limits with HTTP 429 responses.

For some endpoints, the API documentation specifies "a rate limit of 50
calls per minute". This as also been observed on the authentication
logs endpoint.

This changes sets a limit of 0.5 calls/second or 30 calls per minute
for all data streams and inputs.

HTTP 429 responses continue to be treated as errors.

API documentation: https://duo.com/docs/adminapi
```

I considered returning `{ "events": [], "want_more": false }` when a 429 response is received, however, with the new settings we expect to never exceed the limit, and if the limit is repeatedly exceeded, treating it as an error will help with troubleshooting.

## Checklist

### DIFF
--- a/packages/cisco_duo/changelog.yml
+++ b/packages/cisco_duo/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.0.3"
+  changes:
+    - description: Set request rate limits
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/11441
 - version: "2.0.2"
   changes:
     - description: Return response body when a non-200 HTTP status is returned by the API.
--- a/packages/cisco_duo/changelog.yml
+++ b/packages/cisco_duo/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "2.0.3"
   changes:
-    - description: Set request rate limits
+    - description: Set request rate limits.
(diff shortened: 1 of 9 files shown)


packages/cisco_duo/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.0.3"
+  changes:
+    - description: Set request rate limits
```
efd6, line 4:
```suggestion
    - description: Set request rate limits.
```

efd6 (approved):

efd6 (commented):
It may be worthwhile assessing for a 429 in the CEL code and returning the state unmodified, but with an empty `events` array.

chrisberkhout:
> It may be worthwhile assessing for a 429 in the CEL code and returning the state unmodified, but with an empty `events` array.

That was my original plan, but I changed my mind. If it happens occasionally that would be good because it would just wait and try again after the interval. But we're only going to use 60% of the expected allowance, so if we do get 429s, especially a lot of them, it'll be good to see them as errors.

PR #11462, opened 2024-10-18
proofpoint_tap: ensure that queries satisfy API restrictions

### LINKED ISSUE #11461: [proofpoint_tap] Recover from or prevent HTTP 400 Bad Request (for requests >= 7.00d)
If your agent is offline for more than 7 days, when it comes back online the cursor will point to a time that is older than the maximum allowed time range for API requests. It would be nice if the integration prevented this from happening (such as capping the max look back at 7 days).

```json
{
  "log.level": "error",
  "@timestamp": "2024-10-16T02:00:59.319Z",
  "message": "Error while processing http request: failed to collect first response: failed to execute http GET: server responded with status code 400: The requested start time is too far into the past. Requests for information up to 7.00d are accepted"
  ...
}
```

efd6:
## Proposed commit message

Queries may not extend more than 7 days into the past and queried intervals must be at least 30 seconds. If these restrictions are not met, the API returns a 400 status code.

We already document that integration loop intervals should be at least 2m30s, so we do not need to check in the cursor-based iterations, but we do check in pagination to ensure that we do not attempt to collect a trailing set of events that are less than 30 seconds in length. Always check for over 7 day limit.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #11461


packages/proofpoint_tap/data_stream/clicks_blocked/agent/stream/httpjson.yml.hbs
```diff
@@ -23,47 +23,66 @@ request.transforms:
       value: |-
         [[- $last := (parseDate .cursor.last_received_time "RFC3339") -]]
         [[- $hour := (parseDuration "1h") -]]
+        [[- $week_ago := (now (parseDuration "-167h55m")) -]]
+        [[- with $last -]]
+          [[- if .Before $week_ago -]]
```
kcreddy, line 28: I am a bit confused with prefix notation. Does this block:
```
[[- with $last -]]
    [[- if .Before $week_ago -]]
```

 evaluate to `if ($last < $week_ago)` or `if ($week_ago < $last)`?
If it evaluates to the former, then maybe we shouldn't assign `[[- $last = $week_ago -]]` because `$last` is not 7 days ago and API returns success?

kcreddy (commented):
Needed clarification. LGTM otherwise.

packages/proofpoint_tap/data_stream/clicks_blocked/agent/stream/httpjson.yml.hbs
efd6: `with` assigns the value of the pipleline to `.`, so this is equivalent to `if $last before $week_ago { $last = $ week_ago }`. The logic here is intended to clamp last at no more than one week in the past. https://go.dev/play/p/4Ih0PjvMDgn

kcreddy: Got it, my interpretation of what `$last` stands for caused the confusion. 👍🏼

kcreddy (approved):

Issue #11467, opened 2024-10-18
[Elasticsearch integration] timeout option


lucabelluccini:
The `timeout` option, available in metricbeat stack monitoring (https://www.elastic.co/guide/en/beats/metricbeat/current/configuration-metricbeat.html#_timeout) is not available in the Elasticsearch Integration https://www.elastic.co/docs/current/integrations/elasticsearch

consulthys:
The latest 1.15.2 `elasticsearch` integration package has been released: https://epr.elastic.co/search?package=elasticsearch

The HTTP timeout is now available to be configured
![Image](https://github.com/user-attachments/assets/a1740caa-129a-4825-b366-696d245edb6a)

sumnsmnt:
# Integration Page not loading (1.15.3)

I installed Elasticsearch and Kibana locally. When I open the integration page, it's loading infinitely.

I find this log on the browser console.
``` Console Log
core.entry.js:16 
 GET http://localhost:5601/api/fleet/epm/packages?prerelease=false 403 (Forbidden)
fetchResponse	@	core.entry.js:16
(anonymous)	@	core.entry.js:16
await in (anonymous)		
(anonymous)	@	core.entry.js:16
(anonymous)	@	core.entry.js:16
Ot	@	esUiShared.plugin.js:1
u	@	fleet.plugin.js:2
queryFn	@	fleet.plugin.js:2
fetchFn	@	kbn-ui-shared-deps-npm.dll.js:388
O	@	kbn-ui-shared-deps-npm.dll.js:360
(anonymous)	@	kbn-ui-shared-deps-npm.dll.js:360
```

I'm not sure where to ask about this issue, so I'm adding a comment on this closed issue.
I'll appreciate any kind of suggestion to fix the issue.

consulthys:
@sumnsmnt thanks for reporting this.
This repository is about specific integrations, and your issue seems more related to the Integrations page in Kibana.
So in order to have a better chance of this getting solved, you should first ask your question on [Discuss in the Kibana/Fleet category](https://discuss.elastic.co/tags/c/elastic-stack/kibana/7/fleet).
Thanks for your understanding

PR #11479, opened 2024-10-21
[elasticsearch] Add timeout configuration support

consulthys:
## Proposed commit message

The [`timeout` option](https://www.elastic.co/guide/en/beats/metricbeat/current/configuration-metricbeat.html#_timeout) is available when configuring the Metricbeat `elasticsearch` module. This PR adds the possibility of configuring the `timeout` option for the corresponding `elasticsearch` integration.

## Checklist

## Related issues

Closes https://github.com/elastic/integrations/issues/11467


packages/elasticsearch/manifest.yml
```diff
@@ -78,6 +78,14 @@ policy_templates:
             required: true
             show_user: false
             default: node
+          - name: timeout
+            type: text
```
sachin-frayne, line 82: why is this of type `text`?

sachin-frayne (approved):
Other than my understanding of why you need `text` for a timeout, this LGTM!

packages/elasticsearch/manifest.yml
consulthys: Because the values are like "10s"

PR #11486, opened 2024-10-22
Cisco ASA - missing message id has esoteric error statement

jrmolin:
## Proposed commit message

[cisco_asa] Missing message_id field has esoteric error

- If a cisco_asa integration receives unexpected data, there are steps to continue processing it. This can lead to some parsing failures that throw data-processing errors, instead of pipeline-failed-here errors.
- In this case, the missing_id is actually set to `""` and the message is assumed to be a `security` event. Further processing tries to split another expected field on a delimiter that is absent, leading to the error.
- This PR adds a check for this specific case and then fails the pipeline with a more helpful error message:
  >  The message_id ("") indicates this to be a security event, but the message field is not formatted appropriately. I am expecting a comma-separated sequence of 'key: value' pairs."

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/cisco_asa/data_stream/log/_dev/test/pipeline/test-asa-fix.log
```diff
@@ -12,3 +12,4 @@ Aug  6 2020 11:01:38: %ASA-1-106103: access-list filter denied icmp for user joe
 Jun 21 2022 11:47:08: %ASA-6-302015: Built inbound UDP connection 7 for outside:81.2.69.142/3424 (81.2.69.142/3424)(LOCAL\alice, 123) to inside:89.160.20.112/9803 (89.160.20.112/9803) (bob)
 Jun 21 2022 11:47:08: %ASA-6-302015: Built inbound UDP connection 7 for outside:81.2.69.142/3424 (81.2.69.142/3424)(LOCAL\alice) to inside:89.160.20.112/9803 (89.160.20.112/9803) (bob)
 Jun 21 2022 11:47:09: %ASA-6-302015: Built inbound UDP connection 7 for outside:81.2.69.142/3424 (81.2.69.142/3424)(LOCAL\alice, 123) to inside:89.160.20.112/9803 (89.160.20.112/9803)(LOCAL\dave, 246) (bob)
+Jun 21 2022 11:47:09: %ASA-6-8675309666: if anything is here, then the heat death of the universe is imminent
```
jrmolin, line 15: TODO: Remove this test when the other make-it-fail condition is removed. Before merging!!

packages/cisco_asa/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.38.2"
+  changes:
+    - description: "Update tags to make erros states clearer when logs are missing message_id field."
```
taylor-swanson, line 4: typo

```suggestion
    - description: "Update tags to make errors states clearer when logs are missing message_id field."
```

packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -71,6 +71,19 @@ processors:
         # Before version 6.3, FTD used ASA prefix in syslog messages
         FTD_PREFIX: "%{DATA}%(?:[A-Z]+)"
 
+  #
+  # If there is no message at all to process, then fail.
+  #
+  # This has been seen in misconfigured integration scenarios.
+  - remove:
+      field: _temp_.full_message
+      if: ctx._temp_.cisco.message_id == "8675309666"
```
taylor-swanson, line 80: This song is stuck in my head now 🎵 🎵 🎵

packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1192,6 +1205,19 @@ processors:
...
+  # In some cases, the ctx._temp_.cisco.message_id field is created as an empty string. This generally
+  # indicates a security event in an older FTD version.
+  #
+  # However, if misconfigured and incorrect data is sent through this pipeline, it is important to
+  # throw an error that is explanatory to the failure condition.
+  - fail:
+      if: '["430001", "430002", "430003", "430004", "430005", ""].contains(ctx._temp_.cisco.message_id) && !ctx.message.contains(":")'
+      message:  >
+          The message_id ( "{{{ctx._temp_.cisco.message_id}}}" ) indicates this to be a security event, but the message field is not formatted appropriately.
+          I am expecting a comma-separated sequence of 'key: value' pairs.
```
taylor-swanson, line 1218: nit: Prefer not to say "I" in message.
```suggestion
         Expected a comma-separated sequence of 'key: value' pairs.
```

taylor-swanson (commented):
I'm still a bit confused about this change...

> In some cases, the ctx._temp_.cisco.message_id field is created as an empty string. This generally indicates a security event in an older FTD version.

Okay, so it is valid to have an empty message ID, but if the data in message is "bad", then that's bad. Do we know under what conditions this happens? I've been wanting to get us away from "validating" messages and just perform best-effort parsing, without throwing errors (if possible). We could raise an error here, but that just usually leads to more SDHs/bugs. 🤷

packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
jrmolin: I think I might also want to just remove the actual message ids, since the error happens when it's a `""`

jrmolin:
@taylor-swanson AFAICT the only way this happens is if non-ASA data flows into this integration. However, because of how the integration is structured to handle current and very old devices, there is some forgiveness for missing fields. Thus, when invalid data comes in and has a SYSLOG header, I think that makes it get far enough to hit a key/value parse error. The error message for that is rather opaque, because it's doing a CSV split and expecting each item to be `key: value`. 🤷

packages/cisco_asa/data_stream/log/_dev/test/pipeline/test-additional-messages.log-expected.json
```diff
@@ -11746,4 +11746,4 @@
             ]
         }
     ]
-}
```
jrmolin, line 11749: I guess the elastic-package generator changed end-of-file newlines between whenever this was created and the latest release?

packages/cisco_asa/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.38.2"
+  changes:
+    - description: "Add a check before a processor that consistently fails when invalid data transits the integration."
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/11191
```
aleksmaus, line 6: 11191 -> 11486

aleksmaus (approved):
LGTM. Just fix the PR number in the changelog.

taylor-swanson (commented):
PR number needs to be fixed in the changelog, otherwise this looks fine to me

taylor-swanson (approved):

PR #11489, opened 2024-10-23
wiz: fix missing score handling and retain cloud_configuration_rule_description

efd6:
## Proposed commit message

See title.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Ref #11488


packages/wiz/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "1.8.2"
+  changes:
+    - description: Relax requirement for vulnerability score to be present.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/11489
+    - description: Retain `cloudConfigurationRuleDescription` from `sourceRule`.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/11489
```
efd6, line 9: This is unfortunately required to be a patch bump despite being an enhancement due to constraints that exist due to the use of preview marks in the 1.9 version of the integration published in main.

packages/wiz/data_stream/issue/fields/fields.yml
```diff
@@ -103,6 +103,8 @@
           type: keyword
         - name: control_description
           type: keyword
+        - name: cloud_configuration_rule_description
+          type: keyword
```
efd6, line 107: This should really be a `match_only_text`, but the other fields here, `control_description` and `resolution_recommendation` should also, so maintaining alignment and not breaking things by fixing those fields.

packages/wiz/changelog.yml
kcreddy: Thanks for the clarification.

kcreddy (approved):

PR #11492, opened 2024-10-23
[IBM MQ integration package] Update the compatibility section

alaudazzi:
This PR updates the Compatibility section for IBM MQ integration package with the supported official container images.

Closes https://github.com/elastic/integration-docs/issues/559

### DIFF
--- a/packages/ibmmq/_dev/build/docs/README.md
+++ b/packages/ibmmq/_dev/build/docs/README.md
@@ -28,7 +28,7 @@ Note:
 
 ## Compatibility
 
-This integration has been tested against `IBM MQ v9.1` and `IBM MQ v9.2`. The ibmmq `qmgr` data stream is compatible with a containerized distribution of IBM MQ (since version 9.1.0).
+This integration has been tested against IBM MQ v9.1 and IBM MQ v9.2. The ibmmq qmgr data stream is currently only compatible with a [containerized distribution of IBM MQ](https://hub.docker.com/r/ibmcom/mq). 
 
 ## Prerequisites
 
--- a/packages/ibmmq/_dev/build/docs/README.md
+++ b/packages/ibmmq/_dev/build/docs/README.md
@@ -28,7 +28,7 @@ Note:
 
 ## Compatibility
 
-This integration has been tested against IBM MQ v9.1 and IBM MQ v9.2. Currently, the `ibmmq qmgr` data stream is only compatible with a [containerized distribution of IBM MQ](https://hub.docker.com/r/ibmcom/mq). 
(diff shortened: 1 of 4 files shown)


packages/ibmmq/_dev/build/docs/README.md
```diff
@@ -28,7 +28,7 @@ Note:
 
 ## Compatibility
 
-This integration has been tested against `IBM MQ v9.1` and `IBM MQ v9.2`. The ibmmq `qmgr` data stream is compatible with a containerized distribution of IBM MQ (since version 9.1.0).
+This integration has been tested against IBM MQ v9.1 and IBM MQ v9.2. Currently, the `ibmmq qmgr` data stream is only compatible with a [containerized distribution of IBM MQ](https://hub.docker.com/r/ibmcom/mq).
```
tommyers-elastic, line 31: does this suggest we do not support the image from the ibm container registry?

i.e. icr.io/ibm-messaging/mq

alaudazzi:
@tommyers-elastic 
Thank you for your feedback. I'll let @harnish-elastic chime in.

packages/ibmmq/_dev/build/docs/README.md
harnish-elastic: > does this suggest we do not support the image from the ibm container registry?
> 
> i.e. icr.io/ibm-messaging/mq

I just run the integration for the mentioned image and I am able to collect the metrics. So we do support the image from the ibm container registry. So we can mention that we do support container registry images in the compatibility section.

> i think it's a bit ambiguous, the text says "a containerized distribution", but doesn't specify if it must be a specific containerized distribution (even though the link suggests so). there are many others on docker hub that are not the 'official' image. does it work with these? can we make it more specific?

I have also run the integration for the unofficial images and collected metrics successfully. Let me know your thoughts on the below statement for the compatibility section.

```suggestion
This integration has been tested against IBM MQ v9.1 and IBM MQ v9.2. Currently, the `ibmmq qmgr` data stream is only with the containerized versions of IBM MQ, such as those available from [IBM Cloud Container Registry](https://icr.io/) or [Docker Hub](https://hub.docker.com/r/ibmcom/mq). 
```

alaudazzi: Thank you, @harnish-elastic!
I would only complete the sentence with the missing "compatible" word:

> Currently, the `ibmmq qmgr` data stream is only compatible with the containerized versions of IBM MQ

@tommyers-elastic 
Are you happy with these changes? Please let us know if anything else is required.

harnish-elastic: > I would only complete the sentence with the missing "compatible" word:

Mb. I have updated the suggestion with missing word. Thanks!

tommyers-elastic: 👍 LGTM

harnish-elastic (approved):
LGTM! Just need to run `elastic-package build` command for CI pass!

muthu-mps (approved):

Issue #9920, opened 2024-05-20
[Microsoft 365] Pipeline failure resulting in incorrect root-level fields


jvalente-salemstate:
This looks like it was introduced in v2.0.0 or v2.1.0 as this only began showing in mid January. See #8571 and #8803 for the changes in the specific PR.

Almost entirely, this occurs to `o365.audit.record.type: 64`, for Automated Investigation & Response (AIR) events. When the pipeline fails, it does not continue to rename fields nested under `o365audit` to `o365.audit` and these events will not match under any queries for those fields.

Many events do end up parsing correctly, but I've got roundly 9200 since January 22nd with an `error.message` that initially contained:

> Duplicate field 'QueryTime' at [Source: (org.elasticsearch.common.io.stream.ByteBufferStreamInput); line: x, column: y] 

At some point after, this became: 

```json
//Duplicate field 'QueryTime' at [Source (String)"
{
  "Version": "3.0",
  "VendorName": "Microsoft",
  "ProviderName": "OATP",
  "AlertType": "8e6ba277-ef39-404e-aaf1-294f6d9a2b88",
  "StartTimeUtc": "2024-05-18T18:48:51.2937397Z",
  "EndTimeUtc": "2024-05-18T18:48:51.2937397Z",
  "TimeGenerated": "2024-05-18T18:51:31.44Z",
"ProcessingEndTime": "2024-05-18T19:18:50.6990119Z",
"Status": "Resolved",
  "DetectionTechnology": "UrlReputation",
"Severity":"Informational",
"ConfidenceLevel":"Unknown",
"ConfidenceScore":1.0,
"IsIncident":false,
"ProviderAlertId":"01234567-ab10-1337-b000-abcd..."

// [truncated 10638 chars]; line: 1, column: 4841]
```

Formatting edited for readability

The AIR events contain an array of entities related to an alert. An alert might have multiple instances of one entity type and it seems like most,if not all of these, are when the alert has more than one mail cluster that may be sender+IP+subject, sender+attachments, etc (with the ID being analogous for the `fingerprint` processor in elastic, based on those values). Since something like `QueryTime` has exists in each, the above error is thrown. 

Sample data:

```json
{
  "Version": "3.0",
  "VendorName": "Microsoft",
  "ProviderName": "OATP",
  "AlertType": "8e6ba277-ef39-404e-aaf1-294f6d9a2b88",
  "StartTimeUtc": "2024-05-18T18:19:54.5065341Z",
[...]

PR #11499, opened 2024-10-23
o365: Fix the handling of duplicated field QueryTime

chemamartinez:
## Proposed commit message

For some unknown reason we couldn't find out, some events from O365 introduces a malformed JSON object by adding a duplicated `QueryTime` field. The pattern this issue follows is:
- It only affects events with `RecordType == 64`
- The undesired `QueryTime` field doesn't follow ISO8601 format.

So the pipeline fails with:
```
"error": {
      "message": [
        "Duplicate field 'QueryTime'\n at [Source: (String)\"{\"Version\":\"3.0\",\"VendorName\":\"Microsoft\"...\"[truncated 11497 chars]; line: 1, column: 5099]"
      ]
}
```

To fix the error it was added a `gsub` processor that attempts to find the undesired `QueryTime` field and removed from the whole `Data` object. In addition, to prevent other similar issues, added a `on_failure` case when decoding the `Data` field which generally contains unknown fields.

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/9920

### DIFF
--- a/packages/o365/changelog.yml
+++ b/packages/o365/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.6.5"
+  changes:
+    - description: Fix the processing of duplicated QueryTime in Data field.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/11499
 - version: "2.6.4"
   changes:
     - description: Remove in-program template snippets and format CEL code.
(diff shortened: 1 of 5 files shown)


efd6 (approved):
They are not technically wrong; RFC 8259 only says that "The names within an object SHOULD be unique." (section 4). Not nice though. (RFC 7493 _does_ prohibit this.)

PR #11554, opened 2024-10-28
1password - fix system test preserve_original_event

andrewkroh:
## Proposed commit message

```
The system tests for 1password place the preserve_original_event
configuration option into the the top-level vars instead of the
per data_stream vars where the variable is declared. The sample
events generated by the system tests did not contain the
'event.original' field as a result.
```

### DIFF
--- a/packages/1password/data_stream/audit_events/_dev/test/system/test-default-config.yml
+++ b/packages/1password/data_stream/audit_events/_dev/test/system/test-default-config.yml
@@ -3,9 +3,9 @@ service: 1password_eventsapi_mock
 vars:
   url: http://{{Hostname}}:{{Port}}
   token: --token--
-  preserve_original_event: true
   enable_request_tracer: true
 data_stream:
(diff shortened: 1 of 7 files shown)


efd6 (approved):
ISTM that this should be something that's caught by ep.

andrewkroh:
The fact that this is accepted probably means the `POST /api/fleet/package_policies` fleet API is not strict about validating the specified parameters exist.

It is something that elastic-package could validate. Surprisingly, there are only two levels of variables listed in the code[^1]  when there are four levels of specifying variables in packages (integration, policy template, policy template input, data stream). I'll open an issue.

[^1]: https://github.com/elastic/elastic-package/blob/700dd69a9df381475fbdc3440d01db093ec233f9/internal/testrunner/runners/policy/testconfig.go#L20-L23

andrewkroh:
https://github.com/elastic/elastic-package/issues/2186

PR #11583, opened 2024-10-29
[osquery_manager] Add mappings for ECS email fields

aleksmaus:
## Proposed commit message

Add mappings for ECS email fields

This addresses one of the issues with email subfields mapping specifically `email.attachments` ECS field was not specified in the package before, resulting with invalid auto generated object mapping instead of the nested field. 
Adding all of the ECS email fields mappings with this change. 

## Checklist

## Screenshots

Screenshot 2024-10-29 at 6 15 41 PM

Screenshot 2024-10-29 at 6 15 29 PM

### DIFF
--- a/packages/osquery_manager/changelog.yml
+++ b/packages/osquery_manager/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.14.1"
+  changes:
+    - description: Add mappings for ECS email fields
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/
 - version: "1.14.0"
   changes:
     - description: Update schema for osquery 5.13.1
--- a/packages/osquery_manager/changelog.yml
+++ b/packages/osquery_manager/changelog.yml
@@ -1,5 +1,5 @@
 # newer versions go on top
-- version: "1.14.1"
+- version: "1.15.0"
   changes:
     - description: Add mappings for ECS email fields
       type: enhancement
--- a/packages/osquery_manager/changelog.yml
+++ b/packages/osquery_manager/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Add mappings for ECS email fields
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/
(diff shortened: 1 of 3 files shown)


packages/osquery_manager/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.14.1"
```
andrewkroh, line 2: I notice a discrepancy between the proposed version number and the changelog `type`. According to the [Semantic Versioning](https://semver.org) rules, if you’ve categorized this change as an `enhancement`, the minor version should be incremented, not the patch. Conversely, if you've categorized the change as a `bugfix`, then the patch version should be incremented.

Additionally, please ensure that the GitHub label matches the change type — either "bugfix" or "enhancement."

andrewkroh (commented):
I'm surprised to find that `email.attachments` is not covered by [ecs@mappings](https://github.com/elastic/elasticsearch/blob/c6f7827105ffa8b986c830e80ab3b230f5860b5a/x-pack/plugin/core/template-resources/src/main/resources/ecs%40mappings.json).

packages/osquery_manager/changelog.yml
aleksmaus: hmm, I kind of an enhancement, because this mapping was never in the package in the first place, but it also kind of fixes the issue that the user reported. tough call :-)
will feeeex

aleksmaus: Made it an enhancement

aleksmaus:
> I'm surprised to find that `email.attachments` is not covered by [ecs@mappings](https://github.com/elastic/elasticsearch/blob/c6f7827105ffa8b986c830e80ab3b230f5860b5a/x-pack/plugin/core/template-resources/src/main/resources/ecs%40mappings.json).

I was not aware before of the use case where these fields are used with osquery results data, so we didn't have it in the package either.

tomsonpl (approved):
Thanks Aleks for jumping to it so quickly ❤️

andrewkroh (approved):

PR #11589, opened 2024-10-30
Add related.entity to okta integration

romulets:
**tldr; Adding `related.entity` to okta is part of the initial steps of CDR ([epic overview](https://github.com/elastic/security-team/issues/10152)). That enables our customers to better correlate events based on entities and enhances the investigation workflows for the Cloud Security use case.**

[ECS `related.entity` PR](https://github.com/elastic/ecs/pull/2360)

### Background

Elastic Cloud Security Team has been focusing, this past year, on Cloud Detection and Response (CDR). One of the first steps towards the CDR vision is [to enhance investigation workflows for the Cloud Security use-case in SIEM](https://github.com/elastic/security-team/issues/9979).

As part of enhancing investigation workflows it's necessary to be able to correlate events and entities. Meaning, if an alert is triggered on the user `xxxx@xxxx.com`, it is of great value to easily be able to search all the events related to that entity, across multiple indices, with one query. [Therefore we are working on extracting entities and enabling them to be correlated](https://github.com/elastic/security-team/issues/9426).

### Why `related.entity`

With this background, we've researched a few options on what would be the best approach to enable such feature (discussions [https://github.com/elastic/security-team/issues/10026](https://github.com/elastic/security-team/issues/10026) and [https://github.com/elastic/security-team/issues/9798](https://github.com/elastic/security-team/issues/9798), and outcomes [https://github.com/elastic/security-team/issues/10152](https://github.com/elastic/security-team/issues/10152)), and arrived at the ecs field `related`.

Based on the `related` description:

> This field set is meant to facilitate pivoting around a piece of data.
> 
> Some pieces of information can be seen in many places in an ECS event.  
> To facilitate searching for them, store an array of all seen values to their  
> corresponding field in `related.`.

To add a broad `related.entity` field that can hold any needed identifier to pivot data on seems to be well fitted. This would enable customers to simply run `related.entity: "i-000000000"` and get all the hits to that specific cloud resource.

### What is an `entity`?

The term "entity" is broader than the current set of available fields under `related`. Although `ip`, `user` and `hosts` can be identities, there is a lack of space to represent messaging queues, load balancers, storage systems, databases and others. Therefore the proposal to add a new field.

### Related issues

- Closes https://github.com/elastic/security-team/issues/10393


packages/okta/data_stream/system/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -36,6 +36,30 @@ processors:
           return false;
         }
         drop(ctx);
+
+  - script:
+      description: Populate  related.entity
+      lang: painless
+      source: |
+        boolean addValue(Set entities, String value) {
+          if (value == null || value == "" || value == "unknown") {
```
romulets, line 45: @tinnytintin10 I'm ignoring the value `unknown` because if the alternative id is unknown to okta, the value in the event is `unknown`, e.g. 

image


Do you agree with that decision?

packages/okta/data_stream/system/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -36,6 +36,30 @@ processors:
...
+            return false;
+          }
+          return entities.add(value);
+        }
+
+        // Using tree set to ensure a sorting is kept (testing purposes)
+        TreeSet related = new TreeSet();
+
+        addValue(related, field("json.actor.id").get(null));
+        addValue(related, field("json.actor.alternateId").get(null));
```
romulets, line 55: @tinnytintin10 

Do we also capture `alternateId`?  I deemed it relevant because it's where we have the `email` present, which should be able to pivot on. But I also kept the `id` because it's the Okta id.

Any thoughts on if I should capture both indeed? Or maybe just one of the two? Same goes for target.

packages/okta/data_stream/system/elasticsearch/ingest_pipeline/default.yml
terrancedejesus: I'd caution on ignoring `unknown` for Display Name and Alternate ID, specifically when it comes to custom client's that use Okta's API. Typically any custom client that uses the API will not be recognized by Okta and thus be considered `unknown`. In your example, the authentication enrollment is user specific. As long as we favor the existing Alternate ID over unknown, then we should be fine here.

When testing Okta, I'd test via 3 ways to have a good showcase of the workflows:

- Console Admin --> You logging in and doing stuff
- Custom Client --> Create App integration > API Services and then use the API key and client secret to interact with Okta's API.
- Third-Party App --> Register a third-party application that is available via Browse App Catalog and setup auth workflows. Then assign this to an existing user or yourself and interact with your registered applications via your Okta portal or browser plugin.

Each of these will have different workflows and potentially different related entities, targets, etc. From what I have found from past research is `okta.actor.alternate_id` (source), `okta.target` (target) and `okta.debug_context.debug_data` (context) will be your go to for most of this.

packages/okta/data_stream/system/elasticsearch/ingest_pipeline/default.yml
terrancedejesus: Im in favor of both. Specifically for directory sync purposes. The `Id` does capture what Okta has registered as an IdP, but when acting as an IdP on behalf of other SaaS applications, alternateId is often pulled by syncing directory services. For example, if an environment sync'd Okta directory services with an Active Directory tree in a Windows Server, the Alternate ID will be more reliable than the Okta ID based on the functionality of the Universal Directory in Okta.

terrancedejesus:
cc @efd6 --> Adding Dan here who has been making some Okta adjustments requested by TRADE where we have discussed `okta.target` and a few other fields as well. Want to make sure we are all in sync here.

packages/okta/data_stream/system/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -36,6 +36,30 @@ processors:
...
+
+        field("json.target").get(new ArrayList()).stream().forEach(target -> {
```
terrancedejesus, line 57: I've seen `okta.target` as an array of JSON objects but also a single JSON object. Just want to confirm that the code here addresses that or maybe that is inherit regardless.

Screenshot 2024-10-30 at 9 03 51 AM
Screenshot 2024-10-30 at 9 03 46 AM

romulets: Good catch! I covered it in the painless script.

Interesting, once I tried to test with it we got a different error. The event is generated, but if an error. I wonder if that happens on pro.

Anyway it seems to work. Also with what I tested

efd6:
I think this is putting the cart before the horse. There is no `related.entity` field defined in the ECS. That should be proposed and worked through before we add a field use here that will prevent a proper exploration of how the semantics of this should be best represented. That is elastic/ecs#2360 noted in the PR description. I think that change is well posed, but let's wait until it's merged.

romulets:
Hey @efd6 given you approved the `related.entity` [implementation in GCP](https://github.com/elastic/integrations/pull/11762), without the need of ECS, could you have a look on this one again?

packages/okta/data_stream/system/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -36,6 +36,44 @@ processors:
...
+          return entities.add(value);
+        }
+
+        // Using tree set to ensure a sorting is kept (testing purposes)
+        TreeSet related = new TreeSet();
+
+        addValue(related, ctx?.json?.actor?.id);
+        addValue(related, ctx?.json?.actor?.alternateId);
+
+        def oktaTarget = ctx?.json?.target;
```
efd6, line 62:
```suggestion
        addValue(related, ctx.json?.actor?.id);
        addValue(related, ctx.json?.actor?.alternateId);

        def oktaTarget = ctx.json?.target;
```

packages/okta/data_stream/system/fields/fields.yml
```diff
@@ -343,3 +343,10 @@
   fields:
     - name: ip_chain
       type: flattened
+- name: related.entity
+  description: |
+   A collection of all entity identifiers associated with the document. 
+   If the document  contains multiple entities, identifiers for each will be included.
+   Example identifiers include(but not limited to) cloud resource IDs, ARNs,  email addresses,
+   and hostnames. 
+  type: keyword
```
efd6, line 352: Final new line.

efd6 (commented):
Well played. What is the progress of the ECS change?

romulets:
The status of the ECS is awaiting product discussions. @tinnytintin10 is the best one to give context. On that

packages/okta/data_stream/system/elasticsearch/ingest_pipeline/default.yml
romulets: I'm getting:

- `okta.actor.id`
- `okta.actor.alternate_id`
- `okta.target.id`
- `okta.target.alternate_id`

And ignoring if any of those values is empty or contains the string `unknown`. Does that make sense?

packages/okta/data_stream/system/elasticsearch/ingest_pipeline/default.yml
romulets: Fixed

PR #11591, opened 2024-10-30
Otel k8s events added back to kubernetes_otel-cluster-overview Dashboard

gizas:
- Enhancement

## Proposed commit message

- WHAT: Reverting back the events in the Otel General Dashboard of the relevant integration
- WHY:  The events mapping was not correct and relavant fields were not available. Latest 8.16.0-SNAPSHOT includes correct mappings. See [issue](https://github.com/elastic/opentelemetry-dev/issues/469)

  - Changed colours of warning/ errors to red pallete
  - Applied filter datastream.dataset: logs-generic-otel to Events
  - For the `Latest Events` visualsation the Creation Timestampt added, the kind of event and removed the reason

## Checklist

- Clone repo
- Build package locally `elastic-package build`
- Create ES elastic-package stack up -d -v --version=8.16.0-SNAPSHOT
- Apply EDOT collector

## Related issues

- Closes https://github.com/elastic/opentelemetry-dev/issues/469 

## Screenshots

Screenshot 2024-10-30 at 2 44 54 PM

Screenshot 2024-10-30 at 12 56 30 PM


MichaelKatsoulis (approved):

tetianakravchenko:
I've installed this dashboard, few comments:
Viz with title the `Occurrences of Warning/Error  Events`
1. It doesn't make sense to have a breakdown by the `object.type` - it will be always `Warning` (since we have filter for the NOT `object.type: Normal`)
2. instead please use breakdown by kind (Pod/Node) and reason instead:
Screenshot 2024-09-06 at 13 25 23
3. `Occurrences of Warning/Error  Events` - please rename, there is no Error type
4. filter can be set to `body.structured.object.kind: Event` instead of exists to be more specific


Viz `Latest Warning/Error Events`
1. please rename
2. move kind as a second column
3. instead of using for name `body.structured.object.metadata.name`, use `body.structured.object.regarding.name` (is does not include unique id in the end: `coredns-7db6d8ff4d-4jnns` instead of `coredns-7db6d8ff4d-4jnns.1804bc4924e782cf`) and maybe rename to the `Resource Name`
4. add `Event Reason` to make it easier to correlate with another viz


both vizualisation:
- filtering by the namespace and node is not working

gizas:
@tetianakravchenko thank you so much for the above. I think all addressed.

One minor the title of `body.structured.object.regarding.name` made it just `Resource` (as all other titles consist of one word as well)
Screenshot 2024-11-04 at 3 42 02 PM

gizas:
> filtering by the namespace and node is not working

Events dont have the node info
For the namespace we would need to map : object.regarding.namespace -> kubernetes.namespace

tetianakravchenko:
@gizas 

> One minor the title of body.structured.object.regarding.name made it just Resource (as all other titles consist of one word as well)

As I see you renamed `Kind` -> `Resource`, not the `Name` -> `Resource` it is intentional, right?

> Events dont have the node info

there are 2 kind of events - `kind: Pod` and `kind: Node`, Node is hard to reproduce on the 1 node cluster

>  For the namespace we would need to map : object.regarding.namespace -> resource.attributes.k8s.namespace.name

there are few options: either the mapping (I think we can use alias for that) or the ad-hoc dataview with the script or we just agree on the fact that the filtering is not going to work for the events vizualisations

gizas:
> As I see you renamed Kind -> Resource, not the Name -> Resource it is intentional, right?

Yes. Do you like it? Can we keep it like that?

tetianakravchenko:
> Yes. Do you like it? Can we keep it like that?

Yes, lets keep it this way 👍

tetianakravchenko:
what is the reason to have timestamp hiden, and created another `creation timestamp`?
Screenshot 2024-11-04 at 16 39 59

gizas:
> what is the reason to have timestamp hiden, and created another creation timestamp?

Wanted only to show the creation timestamp as I believe is helpful for users. Some times the creation timestamp have the impression can be old and shows an old event. 

Timestamp can be deleted I guess

gizas:
@tetianakravchenko latest update inlcudes the removal of timestamp

For the filtering issue I tried to use a runtime field but seems that they can not replace/update existing fields
Screenshot 2024-11-05 at 12 11 08 PM

So I would say for the time being to merge the PR with no filtering support (cc @MichaelKatsoulis do u agree?) 
and I have opened also a PR for the alias [here](https://github.com/elastic/elasticsearch/pull/116238)

packages/kubernetes_otel/kibana/tags.yml
```diff
@@ -1,3 +0,0 @@
-- text: OpenTelemetry
```
tetianakravchenko, line 1: was it deleted on purpose?

gizas: accidentaly, seems when I  did the export
thanks

PR #11594, opened 2024-10-30
[HAProxy] Fix source.ip condition in append processors

Bearloggs:
Bug

## Proposed commit message

* Minor bug fix to source.ip conditions in append processors to avoid following error :  
```
cannot access method/field [ip] from a null def reference
```

## Checklist


packages/haproxy/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -157,11 +157,11 @@ processors:
   - append:
       field: event.category
       value: network
-      if: "ctx?.source.ip != null && ctx?.destination?.ip != null"
+      if: "ctx?.source?.ip != null && ctx?.destination?.ip != null"
```
andrewkroh, line 160:
```suggestion
      if: ctx.source?.ip != null && ctx.destination?.ip != null
```

packages/haproxy/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -157,11 +157,11 @@ processors:
...
   - append:
       field: event.type
       value: connection
-      if: "ctx?.source.ip != null && ctx?.destination?.ip != null"
+      if: "ctx?.source?.ip != null && ctx?.destination?.ip != null"
```
andrewkroh, line 164:
```suggestion
      if: ctx.source?.ip != null && ctx.destination?.ip != null
```

`ctx` can never be null in this context so we can remove the null-safe operators.

andrewkroh (commented):
LGTM. Just a minor cleanup suggestion.

andrewkroh (approved):

packages/haproxy/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -157,11 +157,11 @@ processors:
   - append:
       field: event.category
       value: network
-      if: "ctx?.source.ip != null && ctx?.destination?.ip != null"
+      if: ctx.source?.ip != null && ctx.destination?.ip != null
```
ishleenk17, line 160:
```suggestion
      if: ctx?.source?.ip != null && ctx?.destination?.ip != null
```

Slightly more fail safe. Adding null check for both ctx and source/destination

packages/haproxy/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -157,11 +157,11 @@ processors:
...
   - append:
       field: event.type
       value: connection
-      if: "ctx?.source.ip != null && ctx?.destination?.ip != null"
+      if: ctx.source?.ip != null && ctx.destination?.ip != null
```
ishleenk17, line 164:
```suggestion
      if: ctx?.source?.ip != null && ctx?.destination?.ip != null
```
Same as above.

ishleenk17 (approved):
Please take care of the comment!
Otherwise looks good

packages/haproxy/data_stream/log/elasticsearch/ingest_pipeline/default.yml
andrewkroh: @ishleenk17 `ctx` can never be null in this context, therefore it is safe to remove the null-safe operators.

This removal was done at my ask in https://github.com/elastic/integrations/pull/11594#discussion_r1823237862

PR #11596, opened 2024-10-30
proofpoint_tap: fix interval end clamping

efd6:
## Proposed commit message

The logic was previously incorrectly stretching all end points to now. Fix this by inverting the clamp condition.

I do not see a way to test the full set of pagination conditions now that this has been fixed; elastic-package does not provide a way to set a testing time, which would be required to test this since the only communication between the agent and the endpoint is via time ranges. Since stream is stateless, we cannot know what we should expect the next request to be. So just do a single request test, leaving the configuration of stream the same on the basis that at some point in the future we may figure out a way to test this.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


kcreddy (approved):
Makes sense

PR #11609, opened 2024-10-31
[ProxySG] Support 'bcreportermain_v1', 'bcreporterssl_v1', and 'ssl' formats

mjwolf:
## Proposed commit message

In the ProxySG integration, add support for 'bcreportermain_v1', 'bcreporterssl_v1', and 'ssl' log formats. These formats are defined here: https://techdocs.broadcom.com/us/en/symantec-security-software/web-and-network-security/edge-swg/7-3/getting-started/page-help-administration/page-help-logging/log-formats/default-formats.html

As it's not possible to automatically detect what log format is being used, users of the integration will need to select the log format to be processed beforehand when setting up the input (this selector dropdown already exists, but previously only had one option).

## Checklist


taylor-swanson:
I'm still looking through this, but had one question (apologies in advance if I already asked this on the first proxysg review).

With regard to the access log format, is it expected that the stream of events we receive will only be in one format? Would we ever need to detect which format it is and handle accordingly?

packages/proxysg/data_stream/log/elasticsearch/ingest_pipeline/bcreportermain_v1.yml
```diff
@@ -0,0 +1,73 @@
...
+                if (!lastWasSpace) {
+                    sb.append(currentChar);
+                }
+                lastWasSpace = true;
+            } else {
+                sb.append(currentChar);
+                lastWasSpace = false;
+            }
+        }
+        ctx._temp_.message = sb.toString();
```
taylor-swanson, line 25: Would a [gsub processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/gsub-processor.html) work here?

taylor-swanson (approved):

PR #11622, opened 2024-11-03
carbon_black_cloud: fix handling of source.address to be dependent on direction

### LINKED ISSUE #11616: [VMware Carbon Black Cloud]: Destination.address not set when netconn_inbound is false
### Integration Name

VMware Carbon Black Cloud [carbon_black_cloud]

### Dataset Name

endpoint_event

### Integration Version

2.5.4

### Agent Version

8.15.2

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.15.1

Ubuntu 20.04 x86_64

### Software/API Version

_No response_

### Error Message

_No response_

### Event Original

_No response_

### What did you do?

I configured the integration to pull data from aws s3 or sqs.

### What did you see?

I am observing that the source.address field is not mapped properly. source.address is always set regardless of netconn_inbound being true or false.

I expect destination.address to be set when netconn_inbound is false and source.address to be set when it is true. Similar to how source.ip and destination.ip are set

### Anything else?

_No response_

efd6:
## Proposed commit message

The netconn_domain is the DNS domain of the remote end of the
connection, which means that it should be assigned to either
source.domain or destination.domain depending on the value of
netconn_inbound. The documentation indicates that it is a domain
so we do not check whether it is an IP for assigning to the
corresponding .ip field. The *.address field is left empty.

ref:https://developer.carbonblack.com/reference/carbon-black-cloud/data-forwarder/schema/deprecated/endpoint.event-1.0.0/

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #11616


efd6:
> [!NOTE]
> AUTOMERGE ENABLED: IF YOU NEED CHANGES DO NOT APPROVE

packages/carbon_black_cloud/data_stream/endpoint_event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -178,22 +178,20 @@ processors:
           ctx.destination.port = ctx.json?.local_port;
           ctx.source.ip = ctx.json?.remote_ip;
           ctx.source.port = ctx.json?.remote_port;
+          ctx.source.domain = ctx.json?.netconn_domain;
         } else {
           ctx.source.ip = ctx.json?.local_ip;
           ctx.source.port = ctx.json?.local_port;
           ctx.destination.ip = ctx.json?.remote_ip;
           ctx.destination.port = ctx.json?.remote_port;
+          ctx.destination.domain = ctx.json?.netconn_domain;
```
kcreddy, line 187: Irrelevant to this PR, but instead of performing null-checks in every line, `ctx.json?.`, the script processor could have global null-check

kcreddy (approved):
LGTM

PR #11623, opened 2024-11-04
sublime_security: add high-water mark timestamp to prevent duplicate event ingestion

### LINKED ISSUE #11363: [Sublime Security]: Duplicate Data Ingestion when using API
### Integration Name

Sublime Security [sublime_security]

### Dataset Name

logs-sublime_security.audit-default

### Integration Version

1.0.0

### Agent Version

8.15.2

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.15.2

GCP Kubernetes Engine (Standalone agent)

### Software/API Version

_No response_

### Error Message

_No response_

### Event Original

_No response_

### What did you do?

Analyzing the agent config, we could not understand how the agent stores the state to avoid data duplication. Looking at the ingest pipeline, it was noticed that the `_id` of the document is computed at runtime.

Image

We triggered a roll over task to the datastream using `POST logs-sublime_security.audit-default/_rollover`

Rollover response:

```
{
  "acknowledged": true,
  "shards_acknowledged": true,
  "old_index": ".ds-logs-sublime_security.audit-default-2024.10.04-000001",
  "new_index": ".ds-logs-sublime_security.audit-default-2024.10.07-000002",
  "rolled_over": true,
  "dry_run": false,
  "lazy": false,
  "conditions": {}
}
```

After the datastream was rolled-over, we restated the agent and checked the data.

Now it can be seen that the same documents ingested in the previous index, were re-ingested in the new index

Image

### What did you see?

Data ingested multiple times after an index roll-over.

Agent would keep the state of the data already ingested and would not re-ingest all the data.

### Anything else?

Policy Configuration:

Image

efd6:
## Proposed commit message

The API does not appear to offer any way to filter events by timestamp and while it offers an offset into a set of events, it does not document that this offset is valid over arbitrarily long periods, so this adds a timestamp to a cursor to keep an indication of events that have already been seen and so which can be dropped.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #11363

### DIFF
--- a/packages/sublime_security/data_stream/audit/agent/stream/cel.yml.hbs
+++ b/packages/sublime_security/data_stream/audit/agent/stream/cel.yml.hbs
@@ -34,13 +34,31 @@ program: |
         "Authorization": ["Bearer " + state.api_key],
       },
     }).do_request().as(resp, resp.StatusCode == 200 ?
-      bytes(resp.Body).decode_json().as(body,{
-        "events": body.events.map(e,{
-          "message": e.encode_json()
-        }),
-        "offset": size(body.events) > 0 ? int(state.offset) + int(state.page_size) : 0,
-        "want_more": size(body.events) > 0,
-      })
+      state.?latest.optMap(t, timestamp(t)).as(latest,
+        bytes(resp.Body).decode_json().as(body, {
+          "events": body.events.map(e,
+            // Filter on timestamp to reduce the risk of duplicate events
+            // being ingested. We allow equality to prevent loss of events
+            // at the slight risk of duplication. The implementation here
+            // depends on the ordering of events being returned by the API
+            // being ascending chronological. If that is not the case this
+            // will result in data loss.
+            !latest.hasValue() || timestamp(e.created_at) >= latest.value(),
+            {
+              "message": e.encode_json()
+            }
+          ),
+          "offset": size(body.events) > 0 ? int(state.offset) + int(state.page_size) : 0,
+          ?"cursor": size(body.events) > 0 ?
+            optional.of({
+              "latest": body.events.map(e, timestamp(e.created_at)).max(),
+            })
+          :
+            // Leave the existing cursor in place if there is no event.
+            optional.none(),
+          "want_more": size(body.events) > 0,
+        })
+      )
     :
       {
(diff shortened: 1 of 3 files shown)


efd6:
> [!NOTE]
> AUTOMERGE ENABLED: IF YOU NEED CHANGES DO NOT APPROVE

packages/sublime_security/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -34,13 +34,31 @@ program: |
...
       },
     }).do_request().as(resp, resp.StatusCode == 200 ?
-      bytes(resp.Body).decode_json().as(body,{
-        "events": body.events.map(e,{
-          "message": e.encode_json()
-        }),
-        "offset": size(body.events) > 0 ? int(state.offset) + int(state.page_size) : 0,
-        "want_more": size(body.events) > 0,
-      })
+      state.?latest.optMap(t, timestamp(t)).as(latest,
```
kcreddy, line 37: Shouldn't this be: `state.?cursor.latest.optMap(t, timestamp(t)).as(latest,` because `latest` is being defined inside the `cursor` and not at the `state` root in L#54.

kcreddy (approved):
Clarification only. LGTM

packages/sublime_security/data_stream/audit/agent/stream/cel.yml.hbs
kcreddy: Only clarification for me. System tests passed so logic has to be correct.

efd6: Your suspicion is correct. I have sent #11683 to fix this.

> "Program testing can be used to show the presence of bugs, but never to show their absence!"
>
> ― [Edsger W. Dijkstra](https://www.cs.utexas.edu/~EWD/transcriptions/EWD02xx/EWD249/EWD249.html)

kcreddy: Sorry for approving this without seeing your `NOTE:` about auto-merge

PR #11630, opened 2024-11-04
Default to `vulnerableAsset.id` when `vulnerableAsset.provider_unique_id` is missing for `resource.id`

maxcold:
## Proposed commit message

`resource.id` is the field used for wiz vulnerability latest transform as a unique key, if it's missing, multiple findings might be merged into one. In many cases (in 10k out of 300k results with dev API key) provider id for a resource is missing, but wiz generated id seems to be always present. Defaulting to this id when provider id is missing

## Checklist

## Author's Checklist

- [ ] 

Test case added to the pipeline test

## Related issues

- fixes https://github.com/elastic/security-team/issues/11030


packages/wiz/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -516,53 +516,6 @@ processors:
       tag: rename_vulnerableAsset_type
       target_field: wiz.vulnerability.vulnerable_asset.type
       ignore_missing: true
-  - remove:
```
maxcold, line 519: moving these processors to the end, to where they were before new processors were added for CDR. In my case the `wiz.vulnerability.vulnerable_asset.id` was removed before I could use it, but it makes sense in general as well

packages/wiz/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -585,9 +538,15 @@ processors:
       ignore_empty_value: true
   - set:
       field: resource.id
-      tag: set_resource_id
+      tag: set_resource_id_from_provider_unique_id
       copy_from: wiz.vulnerability.vulnerable_asset.provider_unique_id
       ignore_empty_value: true
+  - set:
+      field: resource.id
+      tag: set_resource_id_from_resource_id
```
kcreddy, line 546:
```suggestion
      tag: set_resource_id_from_vulnerable_asset_id
```
To give better context.

packages/wiz/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -636,6 +595,60 @@ processors:
...
+        - wiz.vulnerability.vulnerable_asset.subscription.name
+        - wiz.vulnerability.detailed_name
+      tag: remove_custom_duplicate_fields
+      ignore_missing: true
+      if: ctx.tags == null || !(ctx.tags.contains('preserve_duplicate_custom_fields'))
+  - remove:
+      field: event.original
+      tag: remove_event_original
+      ignore_missing: true
+      if: ctx.tags == null || !(ctx.tags.contains('preserve_original_event'))
```
kcreddy, line 628: This block is not required as this is handled by [fleet's final pipeline](https://github.com/kpollich/kibana/blob/main/x-pack/plugins/fleet/server/constants/fleet_es_assets.ts#L126-L131), added in `8.13.0`.

maxcold: fixed

kcreddy (approved):

PR #11641, opened 2024-11-05
ti_misp: add request rate limit

chemamartinez:
## Proposed commit message

Adding the possibility to set the request trace limit for the TI MISP integration.

The nature of API responses that require nested splits to decode events may lead to high resource utilization in the agent, which cannot be avoided by the filters that MISP currently provides. Setting a request trace limit may help in the processing of the responses to avoid the input being overwhelmed.

## Checklist

### DIFF
--- a/packages/ti_misp/changelog.yml
+++ b/packages/ti_misp/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Set request rate limits.
       type: bugfix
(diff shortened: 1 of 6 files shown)


chrisberkhout (approved):
Looks good.

In the proposed commit message, when you say "request trace limit", I think it should be "request rate limit".

PR #11643, opened 2024-11-05
Repairing logs for the aruba integration within the feature branch

qcorporation:
Change Log:
- removing all information from the logs
- regenerating pipeline tests

Tests:
validated that pipeline tests pass locally `elastic-package test pipeline -v`
validated that system tests pass locally `elastic-package test system -v`


packages/hpe_aruba_cx/data_stream/log/_dev/test/pipeline/test-aruba-cx.log
```diff
@@ -0,0 +1,166 @@
+2024-07-12T08:01:16.228730-04:00 6000RCSE4802 lldpd[2407]: Event|104|LOG_INFO|AMM|1/1|LLDP neighbor XX:XX:XX:XX:XX:XX added on 1/1/38
```
dwhyrock, line 1: Is `XX:XX:XX:XX:XX:XX` the normal way to "anonymize" MAC addresses?  Or can we use something that has the valid hex characters, like `ab:cd:ef:12:34:56`?

qcorporation: This is what I got through an LLM - but you are right, Mac addresses should have a range of acceptable characters e.g. a-f, A-F, colon and numerical values.
Let me refactor the logs as I don't want to run into problems when we parse out the `message` field in our subsequent work

qcorporation: changed mac, ip and ipv6-locallink

taylor-swanson (approved):

PR #11663, opened 2024-11-07
CSPM: add cloud connectors support

moukoublen:
## Proposed commit message
Add new template URL for cloudformation creates the cloud connectors remote role. 
Add a new bool variable indicating that cloud connectors are used in the integration. Cloudbeat should parse this boolean. 

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/security-team/issues/11071

### DIFF
--- a/packages/cloud_security_posture/changelog.yml
+++ b/packages/cloud_security_posture/changelog.yml
@@ -1,5 +1,7 @@
 # newer versions go on top
 # version map:
+# 1.12.x - 8.17.x
+# 1.11.x - 8.16.x
 # 1.10.x - 8.15.x
 # 1.9.x - 8.14.x
 # 1.8.x - 8.13.x
@@ -9,6 +11,11 @@
 # 1.4.x - 8.9.x
 # 1.3.x - 8.8.x
 # 1.2.x - 8.7.x
+- version: "1.12.0-preview01"
+  changes:
+    - description: Add cloud connectors support
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/11663
 - version: "1.11.0"
   changes:
     - description: Promote integration
--- a/packages/cloud_security_posture/data_stream/findings/agent/stream/aws.yml.hbs
+++ b/packages/cloud_security_posture/data_stream/findings/agent/stream/aws.yml.hbs
@@ -16,6 +16,9 @@ config:
       {{#if aws.account_type}}
       account_type: {{aws.account_type}}
       {{/if}}
+      {{#if supports_cloud_connectors}}
+      supports_cloud_connectors: {{supports_cloud_connectors}}
+      {{/if}}
       credentials:
         {{#if access_key_id}}
         access_key_id: {{access_key_id}}
--- a/packages/cloud_security_posture/data_stream/findings/agent/stream/aws.yml.hbs
+++ b/packages/cloud_security_posture/data_stream/findings/agent/stream/aws.yml.hbs
@@ -16,8 +16,8 @@ config:
       {{#if aws.account_type}}
       account_type: {{aws.account_type}}
       {{/if}}
-      {{#if supports_cloud_connectors}}
-      supports_cloud_connectors: {{supports_cloud_connectors}}
+      {{#if aws.supports_cloud_connectors}}
+      supports_cloud_connectors: {{aws.supports_cloud_connectors}}
       {{/if}}
       credentials:
         {{#if access_key_id}}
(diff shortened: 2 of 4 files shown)


packages/cloud_security_posture/changelog.yml
```diff
@@ -9,6 +9,11 @@
 # 1.4.x - 8.9.x
 # 1.3.x - 8.8.x
 # 1.2.x - 8.7.x
+- version: "1.11.0-preview14"
```
maxcold, line 12: do we want this change in 8.16 ? as we didn't test 8.16 with this change, maybe it would be safer to have in in 1.12.x?

moukoublen: You are right; I forgot we still need to bump to 8.17. 
This means I should merge after we bump the stable `1.11.0` to create the `1.12.0-preview1.` (otherwise, it would be a bit more complicated to produce the stable `1.11.0` release with a backport branch).

packages/cloud_security_posture/data_stream/findings/agent/stream/aws.yml.hbs
```diff
@@ -16,6 +16,9 @@ config:
       {{#if aws.account_type}}
       account_type: {{aws.account_type}}
       {{/if}}
+      {{#if supports_cloud_connectors}}
+      supports_cloud_connectors: {{supports_cloud_connectors}}
```
Omolola-Akinleye, line 20: @moukoublen I thought  we're going to add the `support_cloud_connectors`  config in `agentPolicy.agent_features` . Why do we need the `supports_cloud_connectors` field here?  Will cloudbeat use this field `support_cloud_connectors`?

packages/cloud_security_posture/manifest.yml
```diff
@@ -133,6 +133,16 @@ policy_templates:
...
             default: https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateURL=https://elastic-cspm-cft.s3.eu-central-1.amazonaws.com/cloudformation-cspm-direct-access-key-ACCOUNT_TYPE-8.16.0.yml
+          - name: cloud_formation_cloud_connectors_template
+            type: text
+            title: CloudFormation Cloud Connectors Template
+            multi: false
+            required: true
+            show_user: false
+            description: Template URL to Cloud Formation Cloud Connectors Stack
+            # ACCOUNT_TYPE value should be either "single-account" or "organization-account"
+            # PROJECT_ID project id for serverless / deployment id for ess
```
Omolola-Akinleye, line 144: PROJECT_ID refers to Serverless Projects this maybe be confusing toinclude the deployment id. Perhaps another slightly more generic name `ELASTIC_SERVICE_ID`?

packages/cloud_security_posture/data_stream/findings/agent/stream/aws.yml.hbs
moukoublen: @Omolola-Akinleye yes we need it in both places currently. This flag will be used on the cloudbeat (or beat in general) side to perform double role chaining towards the remote role. 

So we need to be able to indicate that this double chaining needs to happen.

packages/cloud_security_posture/manifest.yml
moukoublen: Sure I will update accordingly.

Omolola-Akinleye (approved):

packages/cloud_security_posture/data_stream/findings/manifest.yml
```diff
@@ -148,6 +148,13 @@ streams:
         multi: false
         required: false
         show_user: false
+      - name: supports_cloud_connectors
```
Omolola-Akinleye, line 151: should we prefix `aws.support_cloud_connectors` here?

moukoublen: Thats a good question, I am not entirely sure. I thought that since, hierarchy wise, we are under the  `input: cloudbeat/cis_aws` object that its not really needed. But I will try to find out why do we have the `aws.` prefix on the `aws.credentials.type` and in `aws.account_type`.

orestisfl (changes requested):

packages/cloud_security_posture/data_stream/findings/manifest.yml
moukoublen: Thanks @kubasobon! I added `aws.` prefix in `supports_cloud_connectors` as well.

orestisfl (approved):

kubasobon (approved):
Thanks for adding the prefixes

PR #11640, opened 2024-11-05
[Cisco Duo] - Added fix for 400 and 401 errors in telephony v2 data stream

ShourieG:
## Type of change
- Bug

## Proposed commit message

[Cisco API documentation](https://duo.com/docs/adminapi#authentication) mentions that the `Date` header should be equal to the current timestamp. In the current CEL program, the previous execution's `state.date` is being captured and used in pagination and subsequent requests leading to possible `40103` errors. In a related issue users also observed the error `400`: `mintime must be within the past 180 days`.

This PR:
- Fixes `40103` by creating `Date` header during each CEL request.
- Preemptively fixes the `400` error by checking `duration` between `mintime` and `maxtime` doesn't cross
  `180d`, based on a similar issue & [PR](https://github.com/elastic/integrations/pull/11621).

## Note: 
 - This PR was derived from a similar PR [here](https://github.com/elastic/integrations/pull/11621)
 - The version bump here is from `2.2.0` to `2.2.2` because this will only be merged after the PR mentioned above is 
    merged, and that already increments the version to `2.2.1`.

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## Related issues
- Relates #11621

PR #11621, opened 2024-11-03
cisco_duo.trust_monitor: Fix 400 and 401 errors in Trust Monitor

kcreddy:
## Proposed commit message
[cisco_duo] Fix `400` and `401` errors in Trust Monitor.

[Cisco API documentation](https://duo.com/docs/adminapi#authentication) mentions that the `Date` header should equal current timestamp. 
In current CEL program, the previous execution's `state.date` is being captured and 
used in pagination and subsequent requests leading to possible `40103` errors. 
Also, although the API doc doesn't provide the information, users observed error
`400`: `mintime must be within the past 180 days`.

This PR:
- Fixes `40103` by creating `Date` header at each CEL request preparation.
- Fixes `400` by checking `duration` between `mintime` and `maxtime` doesn't cross
  `180d`.

-----

Note to reviewers: `dyn(state)` was used in few occasions in CEL program to avoid errors namely: 
- `type 'string' does not support field selection`
- `no such overload string(double)`
See https://github.com/elastic/integrations/pull/11621#discussion_r1827378901 and https://github.com/elastic/integrations/pull/11621#discussion_r1829191166

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally

**Current tracer logs, 1st pagination request:**
The `timestamp` and `Date` header are not same.
`{"log.level":"debug","@timestamp":"2024-11-05T11:14:07.296Z","message":"HTTP request",.........\r\nDate: Tue, 05 Nov 2024 11:14:05 +0000\r\nAccept......}`

**After the change: tracer logs, 1st pagination request:**
The `timestamp` and `Date` header are same.
`{"log.level":"debug","@timestamp":"2024-11-05T11:05:47.546Z","message":"HTTP request","......\r\nDate: Tue, 05 Nov 2024 11:05:47 +0000\r\nAccept......}`

PR #11670, opened 2024-11-08
cisco_duo: fix 400 and 401 errors in Auth and Activity

chemamartinez:
## Proposed commit message

[Cisco API documentation](https://duo.com/docs/adminapi#authentication) mentions that the `Date` header should be equal to the current timestamp. In the current CEL program, the previous execution's `state.date` is being captured and used in pagination and subsequent requests leading to possible `40103` errors. In a related issue users also observed the error `400`: `mintime must be within the past 180 days`.

This PR:
- For Auth and Activity, fixes `40103` by creating `Date` header during each CEL request.
- For Auth and Activity, preemptively fixes the `400` error by checking `duration` between `mintime` and `maxtime` doesn't cross
  `180d`.
- For Auth, enable it by default to avoid it getting disabled during upgrades.

## Checklist

## Related issues

 - This PR was derived from similar PRs for other data streams:
   - https://github.com/elastic/integrations/pull/11621
   - https://github.com/elastic/integrations/pull/11640
 - The version bump here is from `2.2.1` to `2.2.3` because this will only be merged after the PR mentioned above is 
    merged, and that already increments the version to `2.2.2`.


chrisberkhout (approved):
after merge of #11640

packages/cisco_duo/data_stream/activity/agent/stream/cel.yml.hbs
```diff
@@ -31,113 +28,112 @@ program: |
...
-                            optional.of(body.response.metadata.next_offset.re_find("next_offset_timestamp"))
+            }).do_request().as(resp, (resp.StatusCode == 200) ?
+                bytes(resp.Body).decode_json().as(body, has(body.?response.items) && size(body.response.items) > 0 ?
+                    {
+                        "events": body.response.items.map(item,
+                            {
+                                "message": item.encode_json(),
+                            }
+                        ),
+                        "want_more": has(body.?response.?metadata.next_offset),
```
efd6, line 91:
```suggestion
                        "want_more": has(body.?response.metadata.next_offset),
```

packages/cisco_duo/data_stream/auth/agent/stream/cel.yml.hbs
```diff
@@ -27,20 +28,29 @@ program: |
...
-    ).as(state, state.with(
+    ).as(state, state.with({
+        // the duration between maxtime and mintime should be less than 180d to avoid error_code 400 from API.
+        "maxtime": state.want_more ?
+            string(state.maxtime)
+        :
+            string(int([
+                int(timestamp(int(state.mintime)/1000) + duration("4320h")),
+                int(now - duration("2m"))
+            ].min()) * 1000)
```
efd6, line 41: This allocates an ephemeral array. Ideally we would not need to do this, but the expansion of code that would be required to not do this I think, on balance, would be worse. Maybe we need a `min(a, b)`/`max(a, b)` as well as the overloads that we currently have.

packages/cisco_duo/data_stream/auth/agent/stream/cel.yml.hbs
```diff
@@ -27,20 +28,29 @@ program: |
...
+    })).as(state, state.with(
+        // prepare request data
         {
-            // prepare request data
-            "date": now.format(time_layout.RFC1123Z),
+            // data should be calculated for each request to avoid error_code 40103.
+            "date": now().format(time_layout.RFC1123Z),
```
efd6, line 46: Why are we using the exact execution time rather than the eval start time?

chemamartinez: Just trying to add as recent a date as possible, there should be no problem as the call to `now()` is made once and stored in the `date` variable for use in both the header and the signature.

efd6: From experimentation, this would make about a microsecond different. I think adding a defined grace period would be better.

efd6: Do you want to add a few seconds grace?

efd6 (approved):

PR #11672, opened 2024-11-08
[docker] use ecs definition of the 'event.dataset' field for container_logs

tetianakravchenko:
## Proposed commit message

- WHAT: use ecs definition of the 'event.dataset' field for container_logs
- WHY:  There is no need to set event.dataset explicitly - this should be done on the elastic-agent side https://github.com/elastic/beats/pull/20076. This PR is similar to https://github.com/elastic/integrations/pull/7667

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/docker/data_stream/container_logs/agent/stream/stream.yml.hbs
+++ b/packages/docker/data_stream/container_logs/agent/stream/stream.yml.hbs
@@ -16,3 +16,5 @@ parsers:
 processors:
 {{processors}}
 {{/if}}
+data_stream:
+  dataset: {{data_stream.dataset}}
--- a/packages/docker/changelog.yml
+++ b/packages/docker/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: 2.12.0
+  changes:
+    - description: Use ecs definition of the 'event.dataset' field for container_logs.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/11196
 - version: 2.11.0
   changes:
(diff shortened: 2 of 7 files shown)


packages/docker/data_stream/container_logs/agent/stream/stream.yml.hbs
```diff
@@ -3,6 +3,8 @@ paths:
 {{#each paths}}
   - {{this}}
 {{/each}}
+data_stream:
```
MichaelKatsoulis, line 6: Why do we need to se this? This is different that the event.dataset

tetianakravchenko: it is not needed, it is added anyway to the policy:
```streams:
      - id: 'docker-container-logs-${docker.container.name}-${docker.container.id}'
        data_stream:
          dataset: docker.container_logs
          type: logs
          elasticsearch:
            dynamic_dataset: true
            dynamic_namespace: true
        paths:
          - '/var/lib/docker/containers/${docker.container.id}/*-json.log'
        parsers:
          - container:
              stream: all
              format: docker
```
the same for other datasets, so if we are not providing a way to set a default value it can be removed. I've removed in https://github.com/elastic/integrations/pull/11672/commits/e7051df89cf519d9ec927afdfa18f43e3ea10839, also did the same change for `event.dataset` in all datasreams

tetianakravchenko:
@MichaelKatsoulis @gizas can you please review this PR?

MichaelKatsoulis (approved):

Issue #10746, opened 2024-08-08
[Network Packet Capture]: Should show up in search when searching for Packetbeat


strawgate:
### Integration Name

Network Packet Capture [network_traffic]

### Integration Version

1.29.0

### Agent Version

8.15.0

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.15.0

### OS Version and Architecture

Ubuntu

### What did you do?

Searched for Packetbeat

### What did you see?

Nothing

### What did you expect to see?

Network Packet capture

PR #11680, opened 2024-11-08
Add mention of "Packetbeat" to network packet capture

fearful-symmetry:
## Proposed commit message

Closes https://github.com/elastic/integrations/issues/10746

I'm not sure if we want to mention the beats, or if we have some kind of metadata field that we can use instead of adding it to the package description, but I figured it would be easier to add a PR and ask for opinions instead of asking for ideas on Slack.

## Checklist


efd6:
I'm wondering why we would want to do this. We don't say "Filebeat" for integrations that use that as the collector.

fearful-symmetry:
@efd6 we did get an issue for it, and I think that compared to filebeat or metricbeat this is a tad non-obvious. I've actually hit this before, where I just wanted to turn on packetbeat but I had no idea what it was called in the integrations UI.

efd6:
Yeah, I saw the issue, but it did not look compelling. I just tried looking for packetbeat and while typing, I got a hit because of the prefix substring, though this went away as I continued typing. But in general it seems to be placing a particular bit of technology on a pedestal purely due to a historical accident. ISTM that a better solution (though not one that's available today AFAIUI) is your idea of using a metadata tag to make this possible.

fearful-symmetry:
@efd6 yeah, I agree this isn't an entirely optimal solution. I honestly assumed that some kind of metadata field already existed.

nfritts:
How does the categories field for the integrations work?  Would it be viable to add "tags" as categories?  and would search find and filter on those?  
And following if those answers seem favorable... 
Does putting "tags" in as categories, does it make anything more awkward for customers?  (like does it populate some filter-by-category box that will end up with extra data in it)

fearful-symmetry:
@nfritts , yeah, using categories would make sense? Maybe? I assume that someone on the Kibana side would need to change the search behavior.

andrewkroh:
@kpollich, In effect this PR is trying to do SEO for the package such that the term "Packetbeat" surfaces the integration in Fleet. However, the underlying Beat is an implementation detail that could change at some future time, and we would like to refrain from making explicit references to it in the description.

1. Does the current Fleet package search have some way of accomplishing this?
2. If not, would you entertain a proposal to add some additional terms to the manifest for search purposes? I can see this being useful for Auditbeat related functionality too that has been split into different packages.

kpollich:
Thanks for flagging this, @andrewkroh. We have been working to brainstorm some UX improvements around integrations search, and fully fleshing out the search experience with improved metadata is first on our list of "things that need to be done before we can make other meaningful improvements." 

I added an internal backlink above to our meta issue. We're still in the concept/UX phase but hopefully we'll have some movement on this soon.

The only fields we expose in Fleet's search today are the `name` `title` and `description` fields on the package manifest. So, if we put some mention of packetbeat in the package description, we could surface it in search today to unblock.

nfritts (approved):

PR #11703, opened 2024-11-11
ti_crowdstrike: fix mapping type for ioc.value field

### LINKED ISSUE #11688: [CrowdStrike Falcon Intelligence]: IOC enrichment failed for custom IOCs except IP
### Integration Name

CrowdStrike [crowdstrike]

### Dataset Name

ti_crowdstrike.ioc

### Integration Version

v1.1.7

### Agent Version

8.13.3

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.13.3

Oracle Linux Server 9.4

### Software/API Version

_No response_

### Error Message

<html>
<body>

Ignored value (The value in this field is malformed and can't be searched or filtered)
--

</body>
</html>

### Event Original

{"action":"no_action","applied_globally":false,"created_by":"me@mydomain.com","created_on":"2024-11-11T09:17:03.445826361Z","deleted":false,"description":"Test hash IOC","expired":false,"from_parent":false,"host_groups":["my_host_group"],"id":"a3redactedredactedredactedredactedredactedef21","metadata":{},"modified_by":"me@mydomain.com","modified_on":"2024-11-11T09:17:03.445826361Z","platforms":["windows"],"severity":"medium","type":"md5","value":"65efdcbd4bc64e6e48d82bfa31f710fd"}

### What did you do?

I added custom IOCs of different types (hashes, domains, IPs) in crowdstrike portal
Image

### What did you see?

In kibana, I found the IOC values are just enriched to IPs regardless of the original type which caused this issue if the IOC type is not IP address
Image

Just IP address IOC types are enriched as intended.
Image

All the IOC types should enriched as intended.

### Anything else?

To search, filter and correlate with ti_crowdstrike.ioc.value field, we need that field to be well enriched regardless of the IOC type.

efd6:
## Proposed commit message

The ingest pipeline handles both IP and hash IOCs, but the fields definitions state that the field is an ip type. This prevents users from being able to make use of this field for hash IOCs. So change the type to a keyword. This is a breaking change.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #11688

### DIFF
--- a/packages/ti_crowdstrike/data_stream/ioc/fields/fields.yml
+++ b/packages/ti_crowdstrike/data_stream/ioc/fields/fields.yml
@@ -55,5 +55,5 @@
       type: keyword
       description: The type of indicator.
     - name: value
-      type: ip
+      type: keyword
       description: The specific value of the indicator.
--- a/packages/ti_crowdstrike/elasticsearch/transform/latest_ioc/transform.yml
+++ b/packages/ti_crowdstrike/elasticsearch/transform/latest_ioc/transform.yml
@@ -9,7 +9,7 @@ source:
 # us that ability in order to prevent having duplicate IoC data and prevent query
 # time field type conflicts.
 dest:
-  index: "logs-ti_crowdstrike_latest.dest_ioc-4"
+  index: "logs-ti_crowdstrike_latest.dest_ioc-5"
   aliases:
     - alias: "logs-ti_crowdstrike_latest.ioc"
       move_on_creation: true
--- a/packages/ti_crowdstrike/elasticsearch/transform/latest_ioc/transform.yml
+++ b/packages/ti_crowdstrike/elasticsearch/transform/latest_ioc/transform.yml
@@ -35,4 +35,4 @@ _meta:
   managed: true
   # Bump this version to delete, reinstall, and restart the transform during package.
   # Version bump is needed if there is any code change in transform.
-  fleet_transform_version: 0.4.0
+  fleet_transform_version: 0.5.0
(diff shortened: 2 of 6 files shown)


packages/ti_crowdstrike/elasticsearch/transform/latest_ioc/transform.yml
```diff
@@ -35,4 +35,4 @@ _meta:
   managed: true
   # Bump this version to delete, reinstall, and restart the transform during package.
   # Version bump is needed if there is any code change in transform.
-  fleet_transform_version: 0.4.0
+  fleet_transform_version: 0.5.0
```
kcreddy, line 38: The destination index suffix should also be incremented to create new index, otherwise leads to mapping conflict.

```yaml
dest:
  index: "logs-ti_crowdstrike_latest.dest_ioc-5"
```

packages/ti_crowdstrike/data_stream/ioc/fields/fields.yml
```diff
@@ -55,5 +55,5 @@
       type: keyword
       description: The type of indicator.
     - name: value
-      type: ip
+      type: keyword
```
kcreddy, line 58: This value is also being set to `threat.indicator.name` (`keyword`) in https://github.com/elastic/integrations/blob/main/packages/ti_crowdstrike/data_stream/ioc/elasticsearch/ingest_pipeline/default.yml#L294-L298. But I agree this needs to be fixed. Just FYI.

efd6: Sorry, is there anything that needs to be done here?

kcreddy: No all good. I added to the issue comment. An alternate field to use in case users dont want this breaking change.

kcreddy (approved):

PR #11707, opened 2024-11-12
Universal Profiling Agent: add project_id

florianl:
## Proposed commit message

Add project ID as optional configuration option.

## Checklist

### DIFF
--- a/packages/universal_profiling_agent/manifest.yml
+++ b/packages/universal_profiling_agent/manifest.yml
@@ -1,11 +1,11 @@
 name: profiler_agent
 title: Universal Profiling Agent
-version: 8.14.0
+version: 8.17.0
 categories: ["elastic_stack", "monitoring"]
 description: Fleet-wide, whole-system, continuous profiling with zero instrumentation.
 conditions:
   kibana:
-    version: ^8.13.2
+    version: ^8.17.0
   elastic:
     subscription: basic
 format_version: 3.0.2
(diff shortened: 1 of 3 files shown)


packages/universal_profiling_agent/manifest.yml
```diff
@@ -62,6 +62,12 @@ policy_templates:
             show_user: true
             type: bool
             default: false
+          - name: profiler.project_id
+            title: Project ID
+            description: Splits profiling data into logical groups that you control.
+            show_user: true
+            type: integer
```
rockdaboot, line 69: The field `profiling-project.id` in ES is already a `keyword` (indexed string).
Should we accept a string in the agent as well instead of limiting it to an integer?
The required change in the collector is tiny, as projectID is already used as a string except for Metrics.ProjectID.

rockdaboot (approved):

packages/universal_profiling_agent/manifest.yml
florianl: Accepting a string for a numeric value sounds like a overhead to me, that should be avoided. From a usability point of view, if the accepted string is not a numeric value, where and when does it let the user know about the expectation and what is the expected failing state and how can a user recover form this workflow? How the value is used internally in the backend should not matter to the user.

florianl: this is just my personal opinion, but the use of project ID is an internal detail that users should not care about. for logical separation, there are tags (strings) that users can apply and are used to with every other concept of deployment.
Project ID is based on a concept, that is SaaS centric but doesn't fit well in the current integration anymore.

PR #11718, opened 2024-11-12
Add log format option to all input configs

mjwolf:
## Proposed commit message

In ProxySG integration, add new log format config options to all inputs. Previously, the new formats were added to only one input.  These config values need to exist on all inputs, so all can be configured to all supported log formats.

## Checklist

### DIFF
--- a/packages/proxysg/changelog.yml
+++ b/packages/proxysg/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Add format config to all inputs
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/9999999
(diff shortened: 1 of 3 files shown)


packages/proxysg/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: 0.3.1
+  changes:
+    - description: Add format config to all inputs
+      type: enhancement
```
aleksmaus, line 5: I've been schooled on this before. The enhancement should increment the minor 0.4.0. The bugfix should increment the patch number.

PR #11724, opened 2024-11-13
[crowdstrike] Map additional fields observed in alert data

chrisberkhout:
## Proposed commit message

```
[crowdstrike] Map additional fields observed in alert data

The previously unmapped fields are:
- aggregate_id
- command_line
- detect_type
- host_name
- host_type
- image_file_name
- operating_system
```

## Discussion

I looked at the documentation and other sources but didn't find a comprehensive definition of the alert objects that can be returned by `/alerts/queries/alerts/v2`. The example added here has been seen in actual use.

## Checklist


packages/crowdstrike/data_stream/alert/fields/fields.yml
```diff
@@ -51,6 +51,8 @@
           type: boolean
         - name: cmdline
           type: keyword
+        - name: command_line
```
ShourieG, line 54: This seems a bit redundant at first glance. We already have **cmdline**. Is **command_line** actually different from the existing **cmdline** ?

kcreddy (approved):

packages/crowdstrike/data_stream/alert/fields/fields.yml
chrisberkhout: In the examples we have, `cmdline` is the full path an filename of an executable, and `command_line` is a command without path and with command line parameters.

It's hard to say if they're actually different. I look for documentation but didn't find that kind of detail.

I think matching the vendor prefixed field to what we receive from upstream is good. If they've renamed the field, having both will give us a smooth transition.

PR #11727, opened 2024-11-13
[syslog_router] Add syslog router integration

taylor-swanson:
## Proposed commit message

- Add initial implementation of the syslog router integration which will identify and route syslog-based events to security integrations.

## Checklist

```
cd packages/syslog_router
elastic-package test
```

> [!WARNING]
> Due to limitations in `elastic-package`, automated tests cannot verify routing behavior to data streams external to this package. Verification will have to be performed manually. To verify, follow the instructions in the integration on installing assets for another integration (Cisco ASA, for example), configure the integration, and send a relevant log to the Agent.

## Related issues

- Closes (TODO: integrations issue)
- Relates elastic/security-team#10081


packages/syslog_router/_dev/build/docs/README.md
```diff
@@ -0,0 +1,172 @@
...
+more relaxed and could potentially match against multiple integrations should be
+run last and stricter patterns should be run first. The next priority should be
+given to integrations which will see the most traffic.
+
+Pattern definitions may be reordered by moving the entire `if/then` block up or
+down in the list. For example, moving **Imperva SecureSphere** above **Cisco ASA**:
+
+**Before:**
+
+```yaml
```
mjwolf, line 52: As a future improvment idea, the Defend for Containers integration uses UI components (buttons, dropdowns, etc) to create somewhat similar rules, and then generates a yaml file based on the GUI.

It would be easier for users to have something like that here. I didn't work on that part myself, so I'm not too sure how much work it would be to add here, but it's probably worth investigating. I think setting up this yaml could cause difficulties for a lot of users.

packages/syslog_router/_dev/build/docs/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Setup
+
+Install the relevant integration assets in Kibana.
+
+1. In order for the forwarded event to be properly handled, the target integration's assets (data stream, ingest pipeline, index template, etc.) need to be installed. In Kibana, navigate to **Management** > **Integrations** in the sidebar.
+
+2. Find the relevant integration(s) by searching or browsing the catalog. For example, the Cisco ASA integration. 
+
+![Cisco ASA Integration](../img/catalog-cisco-asa.png)
```
mjwolf, line 30: The image links don't work for me when the integration is installed in Kibana

packages/syslog_router/_dev/build/docs/README.md
taylor-swanson: Phase 2 of this project does involve UI work, but I'll take a look at that integration to see if I can apply any of that here.

I agree the yaml is convoluted, but this is how beats is designed. I'm not sure what to do here, though. One of my original approaches to this integration was developing a new processor, which allowed me to have a much "nicer" looking yaml configuration. I abandoned that in favor of the existing conditionals and processors in beats. Unfortunately, you can't do anything special with yaml blocks from the agent configuration in handlebars, so I wasn't able to take a nicer looking yaml from the agent config and produce the correct filebeat config from it.

I'll take a look at the Defend for Containers integration and see what I can use from that.

taylor-swanson: Yeah looking at Defend for Containers, that's what we're aiming for in Phase 2.

This was the "nicer" looking yaml config I came up with (one of the reroute definitions). I feel like this would be easier for the UI to work with than the beats config that's currently being used. The beats config would be fairly easy for the UI to emit, but not read back in.

```yaml
- target: citrix_waf.log
  patterns:
    - "CEF:0\\|Citrix\\|NetScaler"
  processors:
    - add_fields:
        target: ''
        fields:
          _conf.tz_offset: "UTC"
    - append:
        target_field: tags
        values:
          - citrix_waf-log
```

packages/syslog_router/_dev/build/docs/README.md
taylor-swanson: Odd, this works on my end. What URL is it trying to resolve to?

Screenshot 2024-11-22 at 8 25 24 AM

packages/syslog_router/data_stream/log/_dev/test/pipeline/test-syslog-events.json-expected.json
```diff
@@ -0,0 +1,32 @@
+{
+    "expected": [
+        {
+            "@timestamp": "2024-04-23T09:16:40.645Z",
+            "_conf": {
```
andrewkroh, line 5: How can we get rid of this field? IIUC it is not needed after routing is complete.

taylor-swanson: Correct, it's not needed after routing. In theory, it will be removed by the target integration's pipeline. I agree this isn't ideal, but I haven't found a way to invoke any actions after routing. If you know of a way, I'll happily add it.

packages/syslog_router/_dev/build/docs/README.md
```diff
@@ -0,0 +1,172 @@
+# Syslog Router Integration
```
qcorporation, line 1: There would be a lot of benefits in a compatibility section.
There are already a certain number of built-in routing rules through the `data_stream/log/manifest.yml` file. There are 21 routing rules to begin with. It would help to list those within this documentation to help the user.

packages/syslog_router/data_stream/log/_dev/test/pipeline/test-syslog-events.json-expected.json
taylor-swanson: I ran some additional tests this morning:

- We cannot use `data_stream.dataset` or `event.dataset`, since these will be overwritten by Filebeat/Agent (these fields will be set to `syslog_router.log`).
- The syslog router pipeline does not run when a document is re-routed.
- There is no way to inject any processors (that I know of) after the routing rules are applied, but before the document is actually rerouted.

Unless I'm missing some other method, not sure how to solve this. Using `_conf` is the best solution I know of at the moment since all of our integrations will remove `_conf` fields before indexing.

packages/syslog_router/_dev/build/docs/README.md
```diff
@@ -0,0 +1,207 @@
...
+In the `then` statement, a list of processors can be given. At minimum, an 
+`add_fields` processor needs to be added with the following fields:
+
+**Required fields:**
+
+- `_conf.dataset`: The dataset (`integration.data_stream`) to forward to. This field is used by the routing rules in the integration to route documents to the correct pipeline.
+
+**Optional fields:**
+- `_conf.use_decode_cef`: If set to `true`, the `decode_cef` processor will be used. If the desired value is `false`, this field can be omitted. This field will be automatically removed if set to `true`.
+- `_conf.use_syslog`: If set to `true`, the `syslog` processor will be used. If the desired value is `false`, this field can be omitted. This field will be automatically removed if set to `true`.
```
dwhyrock, line 170: I'm a bit confused by the statements here: `This field will be automatically removed if set to true`

The value can be omitted if you want it false, but if you want it to be `true`, it will be removed.  That seems confusingly similar.  What does it really mean?

packages/syslog_router/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,14 @@
+---
+description: Pipeline for unmatched syslog events.
+processors:
+  - set:
+      field: ecs.version
+      value: 8.11.0
```
dwhyrock, line 6: Do you want to set this to `8.16.0` like you were in another PR for some integrations?

dwhyrock (commented):
A couple comments and questions

packages/syslog_router/_dev/build/docs/README.md
taylor-swanson: Hmm yeah I can see how that's confusing. I didn't think of the case where you set it to false, since it won't remove the field. I can add a `drop_fields` processor to remove those in any case, and add some clarification that is only a temporary field used for running the processor.

packages/syslog_router/data_stream/log/manifest.yml
```diff
@@ -0,0 +1,853 @@
...
+        required: true
+        show_user: true
+        title: Preserve original event
+        description: Preserves a raw copy of the original event, added to the field `event.original`
+        type: bool
+        multi: false
+        default: false
+      - name: reroute_config
+        type: yaml
+        title: Reroute configuration
```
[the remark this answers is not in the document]
taylor-swanson: Yep, exactly. Technically, we really don't even need a pipeline here, but I added it so we can have pipeline tests.

packages/syslog_router/_dev/build/docs/README.md
mjwolf: they're working for me now

mjwolf (approved):
I tested this with Arista NG and Check Point logs, and both were routed to the correct datastream.

dwhyrock (approved):
Great work on this!

PR #11744, opened 2024-11-14
cisco_duo: fix the handling of details field for Activity logs

chemamartinez:
## Proposed commit message

For the Activity data stream, logs that include the following fields `actor.details`, `old_target.details` and `target.details` are not being indexed due to this error:
```
Failed to parse object: expecting token of type [START_OBJECT] but found [VALUE_STRING]
```

These fields contain valid JSON objects so they are defined as flattened, however, the API returns these fields as an escaped string.

## Checklist


packages/cisco_duo/data_stream/activity/_dev/test/pipeline/test-activity.log
```diff
@@ -1,4 +1,4 @@
...
-{"access_device":{"browser":"Edge","browser_version":"110.0.1587.46","epkey":"EP564738291027364891","ip":{"address":"10.0.0.50"},"location":{"city":"Sydney","country":"Australia","state":"New South Wales"},"os":"Linux","os_version":"Ubuntu 20.04"},"action":{"details":null,"name":"mfa_challenge"},"activity_id":"c4e5f678-9012-3456-7890-abcdef012345","actor":{"details":{"created":"2020-05-30T10:55:10.000000+00:00","last_login":"2023-06-15T14:45:22.000000+00:00","status":"Active","groups":[{"name": "HR_Team", "key": "DGHRE987TSDF87ER9X"}]},"key":"DU12MJOPJKLFPKO4PMNBV","name":"rpatel","type":"user"},"akey":"DAAR7EFGTV9TQZA5NSH6","application":{"key":"DIJOTYVER32KXUBMEZ8G","name":"Employee Portal","type":"websdk"},"old_target":null,"outcome":{"result":"SUCCESS"},"target":{"details":{"authenticator_type":"MFA","user_agent": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:110.0) Gecko/20100101 Edge/110.0.1587.46"},"key":"WAZM76EKL9LOHGJPTYG4N","name":"WAZM76EKL4PLJRTYGK3N9","type":"mfa_challenge"},"ts":"2023-06-15T14:46:22.123456+00:00"}
```
kcreddy, line 4: Were the existing logs collected from the real API responses or they are simulated/synthetically generated? 
If they are from real data, maybe its the different API version that returns `details` as escaped string. So, we may have to handle both cases (as an object and also as an escaped string). 
WDYT?

kcreddy: Perfect, thanks !

kcreddy (approved):

PR #11751, opened 2024-11-14
Adding Hardware Health Monitor Events

qcorporation:
Change Log:
- refactored parsed_xxxx tags to be aligned with the rest of the `default.yml` document of `<event_type>_event_xxxx`
- added mock logs for events 3001 - 3013
- generated the expected events within *-expected.json
- corrected some document mistakes from a previous commit
- added documentation for mapping the hardware health monitor events


packages/hpe_aruba_cx/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -182,139 +182,191 @@ processors:
...
+  # https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/HW-HEALTH-MONITOR.htm
+  - grok:
+      if: "['3001', '3002', '3003', '3004', '3005'].contains(ctx.event?.code)"
+      tag: hwhm_event_3001_3002_3003_3004_3005
+      field: "message"
+      description: "Event raised when hardware diagnostics detects error in management module | line card | fabric card | fan tray | rear display card"
+      patterns:
+        - "^Diagnostic %{DATA:aruba.hardware.test_name} failed with error code %{DATA:error.code} on (management module|line card|fabric card|fan tray|rear display card) %{GREEDYDATA:aruba.slot:long}"
+  - dissect:
+      if: "ctx.event.code == '3006'"
```
dwhyrock, line 305: General question here.  In other ones, I've been doing an if like this:
```
if: "ctx.event?.code == '1805'"
```

which I believes fails/skips gracefully if `ctx.event` doesn't exist.  If that is the case here (`if: "ctx.event.code == '3006'"`), would this fail in a different way?

In other words, should we be doing the `ctx.event?.code` reference or not?

mjwolf: I think it should be using `?.`, because it's possible that `event.code` is not set, since it's optionally set in the grok processor here: https://github.com/elastic/integrations/blob/3161df094e8a135392f1d545d474f90069dd959c/packages/hpe_aruba_cx/data_stream/log/elasticsearch/ingest_pipeline/default.yml#L30-L36

When using `event.code`, if code is not set, there will be an error when evaluating it. With `?.`, it evaluates it to `null`, so there's a value to use in the comparison.

So if you know the reference exists, you can use `.`. If it might not exist, use `?.`, and it evaluates to null if it's not set.  At least that's my understanding of it.

gogochan (approved):

gogochan:
Some test failure, other than that, lgtm

packages/hpe_aruba_cx/data_stream/log/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: `ctx.event?.` would only apply to `event`, specifically. If `code` is null, but `event` is defined, then everything still works since the `==` operator can work with null values.

Essentially, it simplifies down to this (assuming `ctx.event` exists, but `ctx.event.code` does not):

```
null == '1805'
```

Which evaluates to false. Where this becomes more apparent is if you try to use a method on `code`, such as `equalsIgnoreCase`:

```
ctx.event?.code?.equalsIgnoreCase('Some text goes here')
```

If you just had `ctx.event?.code.equalsIgnoreCase('Some text goes here')`, then it would blow up if `code` is null.

BUT WAIT! There's more...

The `if` condition really wants a bool as the result, so you have to make sure you have to do another `==` with the expected boolean result:

```
ctx.event?.code?.equalsIgnoreCase('Some text goes here') == true
```

If the left-hand side evaluates to `null`, then it becomes `null == true` which evaluates to `false`.

Remember, we call this language "painless" for a reason 🙄

dwhyrock: Thanks for the in-depth explanation.  Sounds like using `==` would be fine w/o using `?` since it will default to `false`; is that fair to say?

Running methods on the potentially `null` objects sounds like we definitely need to use `?`.

taylor-swanson: In this case, yeah it will be fine.

The main risk here is if `event` is null, which is impossible as long as we don't remove/change the `set event.original from message` processor at the very top, or the `set event.category` processor. Whether or not `code` is null doesn't matter for the reasons already mentioned.

taylor-swanson (approved):
Looks like the README needs to be refreshed, otherwise LGTM

PR #11115, opened 2024-09-12
Add related.entity to cloudtrail integration

romulets:
**tldr; Adding `related.entity` to cloudtrail is part of the initial steps of CDR ([epic overview](https://github.com/elastic/security-team/issues/10152)). That enables our customers to better correlate events based on entities and enhances the investigation workflows for the Cloud Security use case.**

[ECS `related.entity` PR](https://github.com/elastic/ecs/pull/2360)

### Background

Elastic Cloud Security Team has been focusing, this past year, on Cloud Detection and Response (CDR). One of the first steps towards the CDR vision is [to enhance investigation workflows for the Cloud Security use-case in SIEM](https://github.com/elastic/security-team/issues/9979).

As part of enhancing investigation workflows it's necessary to be able to correlate events and entities. Meaning, if an alert is triggered on the ec2 instance `i-000000000`, it is of great value to easily be able to search all the events related to that entity, across multiple indices, with one query. [Therefore we are working on extracting entities and enabling them to be correlated](https://github.com/elastic/security-team/issues/9426).

### Why `related.entity`

With this background, we've researched a few options on what would be the best approach to enable such feature (discussions [https://github.com/elastic/security-team/issues/10026](https://github.com/elastic/security-team/issues/10026) and [https://github.com/elastic/security-team/issues/9798](https://github.com/elastic/security-team/issues/9798), and outcomes [https://github.com/elastic/security-team/issues/10152](https://github.com/elastic/security-team/issues/10152)), and arrived at the ecs field `related`.

Based on the `related` description:

> This field set is meant to facilitate pivoting around a piece of data.
> 
> Some pieces of information can be seen in many places in an ECS event.  
> To facilitate searching for them, store an array of all seen values to their  
> corresponding field in `related.`.

To add a broad `related.entity` field that can hold any needed identifier to pivot data on seems to be well fitted. This would enable customers to simply run `related.entity: "i-000000000"` and get all the hits to that specific cloud resource.

### What is an `entity`?

An "entity" in our context refers to any discrete component within an IT environment that can be uniquely identified and monitored. This broad term encompasses both managed and unmanaged elements.

The term "entity" is broader than the current set of available fields under `related`. Although `ip`, `user` and `hosts` can be identities, there is a lack of space to represent messaging queues, load balancers, storage systems, databases and others. Therefore the proposal to add a new field.

### Related issues

- Closes https://github.com/elastic/security-team/issues/9867

PR #11762, opened 2024-11-18
[CDR][GCP] Add related.entity to GCP Audit Logs

kubasobon:
## Proposed commit message

This PR introduces an ECS field called `related.entity` containing IDs of GCP resources tied to any audit log event.

Closes https://github.com/elastic/security-team/issues/10392

## Background and reasoning

Quoting the original explanation @romulets gave in https://github.com/elastic/integrations/pull/11115:

---

**tldr; Adding `related.entity` to cloudtrail is part of the initial steps of CDR ([epic overview](https://github.com/elastic/security-team/issues/10152)). That enables our customers to better correlate events based on entities and enhances the investigation workflows for the Cloud Security use case.**

[ECS `related.entity` PR](https://github.com/elastic/ecs/pull/2360)

### Background

Elastic Cloud Security Team has been focusing, this past year, on Cloud Detection and Response (CDR). One of the first steps towards the CDR vision is [to enhance investigation workflows for the Cloud Security use-case in SIEM](https://github.com/elastic/security-team/issues/9979).

As part of enhancing investigation workflows it's necessary to be able to correlate events and entities. Meaning, if an alert is triggered on the ec2 instance `i-000000000`, it is of great value to easily be able to search all the events related to that entity, across multiple indices, with one query. [Therefore we are working on extracting entities and enabling them to be correlated](https://github.com/elastic/security-team/issues/9426).

### Why `related.entity`

With this background, we've researched a few options on what would be the best approach to enable such feature (discussions [https://github.com/elastic/security-team/issues/10026](https://github.com/elastic/security-team/issues/10026) and [https://github.com/elastic/security-team/issues/9798](https://github.com/elastic/security-team/issues/9798), and outcomes [https://github.com/elastic/security-team/issues/10152](https://github.com/elastic/security-team/issues/10152)), and arrived at the ecs field `related`.

Based on the `related` description:

> This field set is meant to facilitate pivoting around a piece of data.
> 
> Some pieces of information can be seen in many places in an ECS event.  
> To facilitate searching for them, store an array of all seen values to their  
> corresponding field in `related.`.

To add a broad `related.entity` field that can hold any needed identifier to pivot data on seems to be well fitted. This would enable customers to simply run `related.entity: "i-000000000"` and get all the hits to that specific cloud resource.

### What is an `entity`?

The term "entity" is broader than the current set of available fields under `related`. Although `ip`, `user` and `hosts` can be identities, there is a lack of space to represent messaging queues, load balancers, storage systems, databases and others. Therefore the proposal to add a new field.

---


kubasobon:
@tinnytintin10 I'd appreciate you taking a look at test cases. You can find examples of how `related.entity` field is going to look when this is merged.

packages/gcp/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -105,6 +105,118 @@ processors:
...
+        void addField(Set entities, String fieldName) {
+          addValue(entities, field(fieldName).get(null));
+        }
+
+        boolean addValues(Set entities, String[] values) {
+            boolean addedAll = true;
+            for (String value : values) {
+              addedAll = addedAll && addValue(entities, value);
+            }
+
```
efd6, line 132:
```suggestion
```
The return seems to me like it should live in the stanza as the decl of the var that's being returned.

packages/gcp/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -105,6 +105,118 @@ processors:
...
+            return addedAll;
+        }
+
+        boolean addValue(Set entities, String value) {
+          if (value == null || value == "") {
+            return false;
+          }
+
+         return entities.add(value);
```
efd6, line 141:
```suggestion
          return entities.add(value);
```
Though same query here; while the `addValues` caller uses the return, no other caller does, and `addValues`'s return value is never used, so this seems unnecessary. I would not be surprised if all of this could be elided and done inline more concisely when the return values are removed.

packages/gcp/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -105,6 +105,118 @@ processors:
...
+            value: "{{{ _ingest.on_failure_message }}}"
+      source: |
+        void addFields(Set entities, String[] fields) {
+            for (String field : fields) {
+              addField(entities, field);
+            }
+        }
+
+        void addField(Set entities, String fieldName) {
+          addValue(entities, field(fieldName).get(null));
```
efd6, line 124: What is this doing? I think this needs explanation for maintainers.

packages/gcp/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -105,6 +105,118 @@ processors:
...
+        }
+
+        boolean addValues(Set entities, String[] values) {
+            boolean addedAll = true;
+            for (String value : values) {
+              addedAll = addedAll && addValue(entities, value);
+            }
```
efd6, line 131: This will return true for the case that `values.length` is zero. This does not seem right to me.

However, the function is never used in an expression context. Why is it returning anything at all?

packages/gcp/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -105,6 +105,118 @@ processors:
...
+        HashMap authInfo = field("json.protoPayload.authenticationInfo").get(new HashMap());
+        if (!isKubernetes) {
+          addValue(entities, authInfo.principalEmail);
+        }
+        addValues(entities, new String[]{
+          authInfo.principalSubject,
+          authInfo.serviceAccountKeyName
+        });
+        if (authInfo.serviceAccountDelegationInfo != null) {
+          if (!(authInfo.serviceAccountDelegationInfo instanceof List)) {
```
efd6, line 173:
```suggestion
        if (authInfo.serviceAccountDelegationInfo != null && !(authInfo.serviceAccountDelegationInfo instanceof List)) {
```
with associated cleanup.

efd6 (commented):
Please add a proposed commit message.

packages/gcp/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
romulets: Once I implemented this on the previous PR, it was initially void. But once I called `addValues` (or `addValue`) inside of a `stream().forEach()` the loop was not scanning all the list members unless I returned true.

That, however doesn't make any sense, given the interface of stream ([docs](https://docs.oracle.com/javase/8/docs/api/?java/util/stream/Stream.html))

We could test more, maybe something I did was wrong, and not the return.

But I remember at the time being just wanting to solve the issue and thinking "Well, I don't really know the underlying implementation of painless, could be that lambdas have a different behaviour than plain java".

If it turns out I was wrong, I can fix here too: https://github.com/elastic/integrations/pull/11245

packages/gcp/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
romulets: Essentially this is syntax sugar for adding a value to a set. There a few layers to it because of what we use.

1. The most basic is `addValue`, where we add one specific value to the set. That's is the foundation because it checks for empty and null values, removing the need to clean the set afterwards.
2. `AddValues` is a layer on top, syntax sugar to not call multiple `addValue`, but one `addValues` with all the values you know you want to add.
3. But then, a very common situation is to extract a value from a field inside of the document. That what the `addField` is there for. It extracts a field value with `field(fieldName).get(null)` and adds its value using `addValue`.
4. And on the same reasoning of `addValues`, an `addFields` was added.

With that said, I would appreciate feedback on how would you approach this! I first did it because it felt better to code and read throughout the script. Specially AWS one that is huge.

But if you believe we should add comments, might be a smell of unclarity and unreadable code. How would you approach it? Do you have suggestions of better naming?

I personally am a bit against adding a comment here (just a bit, if we need we do it). This is a function made to make life easier. If we need to add comments to make it clear, might not be worth to have it all. Specially in a script language inside of a yaml without proper highlighting. So I rather have more readable code if that's possible.

packages/gcp/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -105,6 +105,118 @@ processors:
...
+            authInfo.serviceAccountDelegationInfo.get(new ArrayList()).stream().forEach(
+              delegInfo -> {
+                addValues(entities, new String[]{
+                  delegInfo.principalSubject,
+                  delegInfo.firstPartyPrincipal.principalEmail,
```
romulets, line 178: Have you tested with `firstPartyPrincipal` being null? If that's a possibility, I think this could break.

Painless has a nice syntax sugar for null accessing like Kotlin. you could do
```suggestion
                  delegInfo.firstPartyPrincipal?.principalEmail,
```

romulets (commented):
Good stuff! Left some improvement comments and typo, but overall is looking good!

packages/gcp/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
romulets: of course that applies for other non null checked chains

packages/gcp/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
efd6: The patterns that are used here are not common in the remainder of the integrations code base. This means that people need to spend time looking at the code to understand what it's doing. The absence of commentary exacerbates this, as does the indirection of arising from the helpers. I'd be happy for the code to be replaces with explicit iteration rather than the streaming iterator approach that's used here. But I would also like to have an explanation of what the intention of the code is.

romulets: I don't disagree with you. But this is really taste based, right. Previously, the code was on loops and add values. I received the [comment suggesting](https://github.com/elastic/integrations/pull/11115#discussion_r1776017658) to add another function to sugar coat it.

If we do it, we gotta update [this PR](https://github.com/elastic/integrations/pull/11245), which is a lot of code, and refactoring painless is very painful.

Could you please double check with your team what is your preference?

I personally don't mind too much either. I do believe that consistency is the key here, and both have its readability ok.

Let us know what is the preferred way to go from your team's standpoint please.

kubasobon: @efd6 I appreciate your concerns about clarity of code's purpose. Thank you for including a neat rework with iterative approach.
It would be great if we could establish what is the preferred approach for this and future PRs. Given Cloud Security's roadmap and plans to enrich `related`, `target`, and `source` for Cloud Provider logs, it would help avoid us doing the guesswork and introducing inconsistent code. It would also help answer your questions and alleviate any concerns.

efd6: It's partially taste, but the rest of the integration does not use the fluent™ style and the additional code that was required to make it work (the bool returns for example) indicate that the code is fighting against the intention. The previous code was also allocating in some paths in order get the behaviour that a `if` statement gives. The fact that a but was hiding in the original code and was not apparent until the rewrite further argues against it.

Sugar is good in limited quantities, when it becomes a complete glaze, you end up with Dunkin Donuts, and I for one don't want that. With `for` loops and `if` statements, we have a commonly understood language, the fluent™ approach is another set of things that need to be understood.

kubasobon: Thank you for taking the time to refactor the code. I've applied the changes and fixed the tests where new code changed related entities' order.

efd6 (approved):
LGTM, but pleas wait for at least one other since I'm essentially reviewing my own code.

kubasobon:
Thank you @efd6. @elastic/obs-infraobs-integrations & @elastic/obs-ds-hosted-services, I'm kindly requesting a review.

kubasobon:
I'm going to go ahead and press the button, since two weeks have passed.

Issue #11553, opened 2024-10-28
[Aruba] Support 20-30 event types for Aruba integration


qcorporation:
Support Event Types:

1. DCBx events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/DCBX.htm
2. DNS client events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/DNS_CLIENT.htm
3. DPSE events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/DPSE.htm
4. ECMP events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/ECMP.htm
5. ERPS events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/ERPS.htm
6. EVPN events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/EVPN.htm
7. External Storage events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/EXTERNAL-STORAGE.htm
8. Fan events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/FAN.htm
9. Fault monitor events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/FAULT_MONITOR.htm
10. Firmware update events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/UPDATE.htm

PR #11779, opened 2024-11-19
[Aruba] Add Fan Events

gogochan:
https://github.com/elastic/integrations/issues/11553

This PR adds pipeline process for Fan events documented in https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/FAN.htm

Some log samples are manually generated based on the documentation.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/hpe_aruba_cx/data_stream/log/_dev/test/pipeline/test-aruba-fabricated.log
```diff
@@ -1,3 +1,24 @@
...
+2024-06-18T12:45:38.182641-05:00 8360-Primaire fand[1094]: Event|202|LOG_INFO|AMM|1/1|Subsystem A setting fan speed control register to 3: 1500.
+2024-06-18T12:46:38.182641-05:00 8360-Primaire fand[1094]: Event|203|LOG_INFO|AMM|1/1|Air flow direction: Front-to-Back.
+2024-06-18T12:47:38.182641-05:00 8360-Primaire fand[1094]: Event|204|LOG_WARN|AMM|1/1|Fan tray 1 was removed.
+2024-06-18T12:48:38.182641-05:00 8360-Primaire fand[1094]: Event|205|LOG_INFO|AMM|1/1|Fan tray 1 was inserted.
+2024-06-18T12:49:38.182641-05:00 8360-Primaire fand[1094]: Event|206|LOG_INFO|AMM|1/1|Fan module PSU-1/1/2 in fan tray 1 was removed.
+2024-06-18T12:51:38.182641-05:00 8360-Primaire fand[1094]: Event|208|LOG_ERR|AMM|1/1|Unsupported fan tray 2 detected. Please insert a standard fan tray.
+2024-06-18T12:52:38.182641-05:00 8360-Primaire fand[1094]: Event|209|LOG_ERR|AMM|1/1|Shutting down system now because 3 fan_failures exceeded limit of 2.
+2024-06-18T12:53:38.182641-05:00 8360-Primaire fand[1094]: Event|211|LOG_WARN|AMM|1/1|Shutting down system in 60 seconds because 3 fan_failures exceeded limit of 2.
```
qcorporation, line 9: I don't think there's a period at the end of this log line (not in the docs)

packages/hpe_aruba_cx/data_stream/log/_dev/test/pipeline/test-aruba-fabricated.log
```diff
@@ -1,3 +1,24 @@
...
+2024-06-18T12:54:38.182641-05:00 8360-Primaire fand[1094]: Event|212|LOG_WARN|AMM|1/1|System shutdown timer is cancelled.': yes
```
qcorporation, line 10: In my PRs, I've not included anything after the `'` - I remember there was an example in the real log that didn't include anything after the `'`

packages/hpe_aruba_cx/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -46,6 +46,174 @@ processors:
...
+  # Fan Events (2xx)
+  # https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/FAN.htm
+  # * 212 is not included as it does not require further processing.
+  - grok:
+      field: message
+      tag: fan_event_201
+      description: "Log the total number of fans in the subsystem"
+      if: "ctx.event?.code == '201'"
+      patterns:
+        - "^There are %{INT:event.count} total fans in subsystem %{ARUBA_SUBSYSTEM:aruba.subsystem}"
```
qcorporation, line 58: within the docs `<count>` was mapped to `aruba.fan.count`

packages/hpe_aruba_cx/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -46,6 +46,174 @@ processors:
...
+      pattern_definitions:
+        ARUBA_SUBSYSTEM: "[-_0-9a-zA-Z]+"
```
qcorporation, line 60: from https://github.com/elastic/elasticsearch/blob/main/libs/grok/src/main/resources/patterns/ecs-v1/grok-patterns - this is very similar to %{USERNAME} with the addition of the `.` did you want to use that instead?

This also applies to the other FAN_VALUE `pattern_definitions` below

Taking a step back, though, is it documented that these are the only valid characters? I know that Aruba actually does log internationalization so I'm surprised by this restriction

packages/hpe_aruba_cx/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -46,6 +46,174 @@ processors:
...
+        - "^Fan tray %{DATA:aruba.fan.ft_num} powered %{FAN_STATUS:aruba.fan.status}"
+      pattern_definitions:
+        FAN_STATUS: "[-_0-9a-zA-Z]+"
+  - grok:
+      field: message
+      tag: fan_event_220
+      description: "Log event when a fan tray is inserted specifying its airflow direction"
+      if: "ctx.event?.code == '220'"
+      patterns:
+        - "Fan tray %{DATA:aruba.fan.ft_num} airflow is %{FAN_DIRECTION:aruba.fan.ft_dir}"
```
qcorporation, line 179: Any reason to not have the `^` here?

note: same for 221, 223, 224

qcorporation:
I also noticed that we do not include the `.` (period) at the end of the patterns - should I stop doing that within my code submissions for this Aruba integration?

packages/hpe_aruba_cx/_dev/build/docs/README.md
```diff
@@ -293,35 +293,38 @@ Note: Descriptions have not been filled out
...
+| <FanStatus>              | aruba.fan.status             |
+| <FanModuleIdx>           | aruba.fan.module_idx         |
+| <FanTrayIdx>             | aruba.fan.tray_idx           |
+| <OldStatus>              | aruba.fan.old_status         |
+| <NewStatus>              | aruba.fan.status             |
+| <FanCount>               | aruba.fan.count              |
+| <FanMinimum>             | aruba.fan.minimum            |
+| <ZoneIdx>                | aruba.fan.zone_idx           |
+| <FanSpdIdxStatus>        | aruba.fan.speed_idx_status   |
+| <Status>                 | aruba.fan.status             |
```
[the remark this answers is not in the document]
gogochan: I thought about it, and there is `old_status` and I suppose the status may be specific to fan, which we may not want to mix up with `aruba.status`. Hence, falling back to `aruba.fan.status`

packages/hpe_aruba_cx/data_stream/log/_dev/test/pipeline/test-aruba-fabricated.log
gogochan: Thanks, I will remove those too.

packages/hpe_aruba_cx/data_stream/log/elasticsearch/ingest_pipeline/default.yml
gogochan: The reason I came up with a specific pattern is because of handling `.` at the end, which isn't part of the subsystem. I have seen cases where doc didn't have period at the end, but the sample including it.

packages/hpe_aruba_cx/data_stream/log/elasticsearch/ingest_pipeline/default.yml
gogochan: yup, valid for 200.
221, 223, 224 are dissect. I don't think we need them.

packages/hpe_aruba_cx/data_stream/log/elasticsearch/ingest_pipeline/default.yml
qcorporation: Let me clarify the comment about stepping back a bit.
I'm glad we are being restrictive as long as values for the subsystem are bound to `ARUBA_SUBSYSTEM` - I'm questioning if there will be special characters outside of a(A)-z(Z) and numbers. Possibly a character within extended ascii

dwhyrock: "There are 8 total fans in subsystem 🐩 "

taylor-swanson: My two cents: we restrict patterns to aid in our extraction of values from the message, not to validate against any kind of standard/spec, known or otherwise. Our job is not to validate correctness of the logs, only to extract information where we can.

Reason being: the docs lie. Not intentionally, of course, but I've seen it way too often that the docs say one thing, but then the user produces a contradictory log.

PR #11781, opened 2024-11-20
[citrix_adc] Support addition log message types and ECS mappings

mjwolf:
## Proposed commit message

Improve Citrix ADC integration log parsing and ECS mappings.

Changes are:

- Support "Mapped Ip" as value for `Nat_Ip` in all patterns in the sslvpn pipeline
- Add support for additional "Message" subtypes, and add a "DATA" wildcard that will capture all patterns. All valid "Message" patterns are not known, so it's better to capture all without parsing individual fields than to cause an error.
- Add addition ECS mappings for `event.kind`, `event.outcome`, `observer.hostname`
- Calculate `event.duration` as the difference from `event.start` and `event.end` 

## Checklist


jrmolin (approved):
looks better and better!

PR #11794, opened 2024-11-20
aws_inspector: add note about AWS Permissions and Role ARN support

chemamartinez:
## Proposed commit message

`inspector2:ListFindings` is a required AWS permission for IAM users in order to avoid this error when enabling the Inspector data stream:
```
{\"message\":\"User: arn:aws:iam::123456789:user/service/elastic is not authorized to perform: inspector2:ListFindings on resource: arn:aws:inspector2:eu-west-1: 123456789:/findings/list\"}"
```
See https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazoninspector2.html.

It also clarifies that `Role ARN`, which is a global setting for the AWS integration, is not supported for the Inspector data stream as it make requests to the Inspector API through HTTPJSON, while `Role ARN` is a setting for AWS-based inputs.
 
## Checklist

### DIFF
--- a/packages/aws/_dev/build/docs/README.md
+++ b/packages/aws/_dev/build/docs/README.md
@@ -168,6 +168,7 @@ make sure these permissions are given:
 * `ec2:DescribeInstances`
 * `ec2:DescribeRegions`
 * `iam:ListAccountAliases`
+* `inspector2:ListFindings`
 * `logs:DescribeLogGroups`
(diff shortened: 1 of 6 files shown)


efd6 (commented):
This does need a changelog entry, otherwise it won't be presented to users in the visible documentation.

ShourieG (approved):

PR #11804, opened 2024-11-21
[crowdstrike] Extract user and host names from the name field

chrisberkhout:
## Proposed commit message

```
[crowdstrike] Extract user and host names from the name field (#)

Some alerts data from Crowdstrike has a `name` field of the form
`USERNAME on HOSTNAME`. Extract those values when present, add them to
`related.user` and `related.hosts`, and use them for `user.name` and
`host.name` if those fields aren't already populated.

Also corrects one use of `related.hash` to be `related.hosts`.
```

## Checklist


packages/crowdstrike/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "1.46.0"
+  changes:
+    - description: Extract user and host names from the name field.
```
kcreddy, line 4:
```suggestion
    - description: Extract user and host names from the alert.name field.
```

kcreddy (approved):
nit. LGTM

packages/crowdstrike/changelog.yml
chrisberkhout: It's a top level field in the incoming data. Renamed from `json.name` before the new logic.

kcreddy: Yeah, I deduced that (a bit late).

PR #11822, opened 2024-11-22
cisco_duo.auth: Make retry options configurable

kcreddy:
## Proposed commit message
Authentication logs from Cisco Duo Admin API has very low API request limit (not more than 1 request per minute). Reaching this limit would lead to `42901` from the API. The response also doesn't include a `Retry-After` header, which leads to a default back-off retry attempts. This subsequently leads to further `42901` as the limit is already reached. 

This PR allows users to set retry configuration options allowed by CEL input. This helps in reducing the number retry requests being made when already reached `429`.

Reference: 
https://help.duo.com/s/article/1338?language=en_US
https://duo.com/docs/adminapi#authentication-logs

> [!NOTE]
> This only reduces `42901` Too Many Requests from the retry attempts. The users have to tune `Interval` and `Rate limit` to avoid the `42901` from the initial request (non-retry).

## Checklist

## Screenshots

Screenshot 2024-11-22 at 11 52 29 AM

### DIFF
--- a/packages/cisco_duo/changelog.yml
+++ b/packages/cisco_duo/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.2.6"
+  changes:
+    - description: Make retry options configurable.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/1
 - version: "2.2.5"
   changes:
     - description: Fix the parsing of last_published timestamp in Activity and Telephony data streams.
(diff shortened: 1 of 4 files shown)


packages/cisco_duo/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.2.6"
+  changes:
+    - description: Make retry options configurable.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/1
```
chemamartinez, line 6: Update PR link

chemamartinez (commented):
It may be worthy to include the same options for the rest of the data streams using CEL

kcreddy:
> It may be worthy to include the same options for the rest of the data streams using CEL

Agreed. Will do

packages/cisco_duo/changelog.yml
kcreddy: Updated PR number

chemamartinez (approved):
LGTM

PR #11825, opened 2024-11-22
1password: unify the use for user.full_name and user.name for all data streams

chemamartinez:
## Proposed commit message

This PR unifies the use of `user.full_name` for the `audit_events` data stream, as the other both ones already use that field.

In addition, based on real data and the current definition of these fields in [ECS](https://www.elastic.co/guide/en/ecs/current/ecs-user.html), `full_name` is a better match.

Updated the `audit_events` data stream to also populate `user.full_name`. Same way it has been added `user.name` to the other two data streams.

## Checklist


chemamartinez:
> Looks like a breaking-change that needs incrementing major?
> 
> The [ECS doc](https://www.elastic.co/guide/en/ecs/current/ecs-user-usage.html#ecs-user-identifiers) says we could populate `user.name`. Maybe we could avoid breaking change that way?
> 
> > In cases where a purpose-specific user field such as url.username is populated, user.name should also be populated with the same user name.

Agreed. I will keep populating `user.name` to avoid any breaking change, do you think it is ok to do the same for the other two data streams for consistency?

kcreddy:
> Agreed. I will keep populating user.name to avoid any breaking change, do you think it is ok to do the same for the other two data streams for consistency?

@chemamartinez , sounds good 👍🏼

kcreddy (approved):
LGTM. Please update PR title and description accordingly.

PR #11863, opened 2024-11-25
[Zscaler ZIA] Add custom input options to http input

SimonKoetting:
Adding the ability to specify custom options for the HTTP endpoint input. That's very useful in some situations or for debugging


efd6 (approved):
This kind of thing makes me feel uncomfortable; particularly given that we do not have the capacity to control where the documentation link points to so that it matches the stack version that a user has installed. However, I see that this is uses in a multitude of other integrations, so approving.

PR #11878, opened 2024-11-26
[Synthetics] Fix mapping for decimal fields !!

shahzad31:
Fix mapping for decimal fields !!

### DIFF
--- a/packages/synthetics/changelog.yml
+++ b/packages/synthetics/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.2.3"
+  changes:
+    - description: Fixed cls field type to float
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/10497
 - version: "1.2.2"
   changes:
     - description: Added mappings for journey.duration fields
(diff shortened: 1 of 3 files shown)


packages/synthetics/manifest.yml
dominiqueclarke, line 1: The safest thing to do here would be to also bump the kibana version constraint. That way we can reduce the scope of any potential unintended consequences.

packages/synthetics/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.2.3"
+  changes:
+    - description: Fixed cls field type to float
+      type: enhancement
```
dominiqueclarke, line 5: Technically a fix

packages/synthetics/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.2.3"
+  changes:
+    - description: Fixed cls field type to float
+      type: fix
```
dominiqueclarke, line 5:
```suggestion
      type: bugfix
```

packages/synthetics/manifest.yml
```diff
@@ -30,7 +30,7 @@ conditions:
       - "uptime"
     subscription: "basic"
   kibana:
-    version: "^8.11.0"
+    version: "^8.17.0"
```
shahzad31, line 33:
```suggestion
    version: "^8.11.0"
```

we should fix this as back as possible

Issue #11511, opened 2024-10-24
[Falco]: Dashboard seems to be broken


nick-alayil:
### Integration Name

Falco [falco]

### Dataset Name

falco.alerts

### Integration Version

1.0.2

### Agent Version

8.16

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.16

### OS Version and Architecture

Amazon Linux 2023

### What did you do?

Ingest falco data in to Elastic using Falcosidekick but it could be done using falco integration as well. In the context of the issue, I assume how data is ingested have less relevance. I may be wrong here. Irrespective of the ingestion methodology it run through the same pipelines and would end up populated in the dashboard.

### What did you see?

Falco dashboard seems to be broken by looking at `log-*` index pattern, which ends up showing broken widgets / cards in the dashboard. Also, in `Alerts by Host - Top 10 [Logs Falco]` widget / card, table is expecting `host.hostname` field, which is in fact not present in the falco.alerts datastream. I tried to duplicate the dashboard and edited it to look specifically at `logs-falco* ` index pattern by creating a new data view and made most of the widgets / cards look at the new data view, which fixed many broken experiences. The details of this explanation is visible in the attached recording. Sorry for the lengthy 2.10mints recording.

https://upload.elastic.co/d/9a180431a47a43b55c42881fc6cfef3c1c79f6944bdf9427cf500313f1820108
AuthZ token: f92143b2d591efab

### What did you expect to see?

Expectation is to see only data from logs-falco* reflected in all the widgets / cards inside the falco dashboard.

nick-alayil:
Summary of sync w/ @chemamartinez 

Confirmed that dashboard filter by default to `data_stream.dataset : falco.alerts` showing only falco content. However, if an user remove the filter, of course it'll populate data from all `logs-*` indices. The real problem is dashboard `reset` button seems to showcase only in some cases to revert to default / original state.

Alerts by Host widet is broken because of changes made in [this PR](https://github.com/elastic/integrations/pull/11051). We need to modify this widge to look for `host.name` instead of `host.hostname`

`event.original` column of Alert Stream widget/card seems to be showing no value since the file data stream is missing it. We need to make sure we populate that field as it's value to see the raw event.

chemamartinez:
I've found several issues when forwarding directly from Falco Sidekick to Elastic without using the integration. The incoming events don't follow the expected format for the ingest pipeline, as the ingest pipeline expects to parse events that are read by the integration.

[...]

PR #11881, opened 2024-11-26
falco: fix dashboard visualisation and handling of event.original

chemamartinez:
## Proposed commit message

Falco assets are not only used for the integration itself. They are also used by users that want to ingest Falco alerts directly into Elasticsearch through Falco Sidekick (more information at https://www.elastic.co/blog/cncf-open-source-security-tools).

This PR adds `event.original` to events directly ingested from Falco Sidekick.

It also updates the `Alerts by Host` visualisation to rely on the `host.name` instead of `host.hostname`.

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/11511

## Screenshots

image

image


packages/falco/data_stream/alerts/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -36,7 +38,14 @@ processors:
...
         }
-        ctx['falco'] = m
+        ctx['falco'] = m;
+        def original = ctx.falco.toString();
+        if (original != null) {
+            if (ctx.event == null) {
+                ctx.event = new HashMap();
+            }
+            ctx.event.original = original;
+        }
```
efd6, line 48: Make this depend on `preserve_original_event`?

packages/falco/data_stream/alerts/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -36,7 +38,14 @@ processors:
                 ctx.remove(v);
             }
         }
-        ctx['falco'] = m
+        ctx['falco'] = m;
+        def original = ctx.falco.toString();
```
efd6, line 42: I don't think this is the right thing to do; `toString` gives a Java syntax representation of the value:
```
def a = [:];
a.a = 1;
a.b = 2;
a.c = "three";
a.toString();
```
```
{a=1, b=2, c=three}
```
We want something that renders the JSON. `Json.dump` works, though note that it does not canonicalise key order in my experience, so tests will be brittle.
```
def a = [:];
a.a = 1;
a.b = 2;
a.c = "three";
Json.dump(a);
```
```
{"a":1,"b":2,"c":"three"}
```

efd6 (commented):
Can we add a test case for this?

packages/falco/data_stream/alerts/elasticsearch/ingest_pipeline/default.yml
efd6: Alternatively, just retaining the actual original data.

chemamartinez: Thanks for the suggestions. I've tried using `Json.dump()` and it works as you mentioned. I think it is ok that result doesn't follow canonical order as tests were not added for this use case.

packages/falco/data_stream/alerts/elasticsearch/ingest_pipeline/default.yml
chemamartinez: The use case of this processor is users that want to ingest directly from Falco without the need of a Elastic agent, so they'd just install the integration assets. In this scenario, incoming events would not include the `preserve_original_event` tag so that's why I didn't add it as a dependency.

Matching `ctx?.output_fields != null` means that the event doesn't follow the format that Filebeat delivers, wrapping the entire event into `message`, and it follows the format that Falco Sidekick provides.

This is also why I didn't add test cases, in the context of the integration, I couldn't find a way to reproduce this format.

efd6 (approved):

Issue #11806, opened 2024-11-21
Bring uniformity in `aws.dimensions.*` fields mapping


zmoog:
## Situation

Elastic offers multiple integrations to collect metrics from CloudWatch in the AWS integration package. 

Some are "specialized" and only collect metrics from one AWS service,  like the EC2, ECS, Redshift, and others. However, one "generic" integration allows users to write the integration configs to collect metrics from any AWS service. The generic integration is excellent for collecting metrics from services we have not supported.

All integrations (specialized and generic), collect metric dimensions as `aws.dimentions.*` fields.

## Problem

Unfortunately, not all metrics integrations use the exact mapping for the `aws.dimentions` field.

| Integration                      | Type        | Mapping for aws.dimentions |
| -------------------------------- | ----------- | ---------------------------- |
| Collect EC2 logs from CloudWatch | Specialized | `object`                     |
| Collect metrics from CloudWatch  | Generic     | `flattened`                  |

With different mappings, it's hard for users to correlate dimensions and build dashboards.

## Solution

Work in progress.

We need to identify:

- What is the best mapping to represent AWS metric dimensions
- Describe a custom mapping to work around the issue in the short term (if it exists).
- Identify a mid-long term solution to bring uniformity to dimensions field mapping.

## References

Related:

- A large migration to flattened: https://github.com/elastic/integrations/pull/8115/files

zmoog:
For a short-term solution for users who want to map `aws.dimensions` as an `object` of `keyword`s, I tried defining a custom dynamic template in the generic integration. 

```yaml
  - name: dimensions.*
    type: object
    object_type: keyword
    object_type_mapping_type: "*"
```

However, the `aws.dimensions` takes precedence.

zmoog:
All `*_metrics` data streams use the `object` mapping. 

For example, here is the `dimensions` definition for EC2 metrics:

```yaml
# packages/aws/data_stream/ec2_metrics/fields/fields.yml
- name: aws
  type: group
  fields:
    - name: dimensions
      type: group
      fields:
        - name: AutoScalingGroupName
          type: keyword
          dimension: true
          description: An Auto Scaling group is a collection of instances you define if you're using Auto Scaling.
        - name: ImageId
          type: keyword
          dimension: true
          description: This dimension filters the data you request for all instances running this Amazon EC2 Amazon Machine Image (AMI)
        - name: InstanceId
          type: keyword
          dimension: true
          description: Amazon EC2 instance ID
[...]

Issue #9910, opened 2024-05-17
[Meta] [TSDB] Eliminate fingerprinting of dimensions and labels added as part of TSDB Enablement


agithomas:
The Elasticsearch issues 
-  https://github.com/elastic/elasticsearch/issues/93564
- https://github.com/elastic/elasticsearch/pull/98023 
are resolved.

These changes help to overcome the dimension limit and the need of hashing the labels and dimensions that are part of the Elastic Integration package field. 

The changes are presently [evaluated & tested](https://github.com/elastic/integrations/pull/9785) for the Prometheus package. 

This meta issue captures the scope of integration package which could take the benefit of above mentioned changes added in Elasticsearch. 

**Requirements**
- 8.14 or above.

**Pass Criteria**
- No document drop should happen after applying the changes 

**Testing Methodology.**
- Manual verification
- [TSDB Test toolkit evaluation](https://github.com/elastic/TSDB-migration-test-kit)

PR #11883, opened 2024-11-26
[aws] [cloudwatch_metrics] Map aws.dimensions field as object

zmoog:
## Proposed commit message

Change the mapping type for the `aws.dimensions` field from `flattened` to `object`.

Currently, all `*_metrics` data streams but one use the `object` mapping. The `cloudwatch_metrics` data stream uses the `flattened` type instead.

We need to unify the mapping of `aws.dimensions` across all metrics-related data streams in the AWS integration.
 If all data streams use the exact mapping for `aws.dimensions`, users will be able to query and build a dashboard that correlates data across different data streams.

## Checklist

## Author's Checklist

- [ ] #11893 (to get `subobjects: false` support)
- [ ] Add a case for "How to test this PR locally" to prove the new mapping makes a difference for query and dashboard purposes.

- Start the local stack (use 8.17.0)
- Install the latest released integration (using AWS 2.36.2)
- Check the component template
- Post the test document
- Try to visualize the metric in Kibana
- Upgrade the integration to this PR version
- Delete the data stream and re-index the same test document
- Re-try to visualize the metric in Kibana using the new mapping

### Check the component template

With 2.36.2, the component template should be something similar to:

```text
// GET _component_template/metrics-aws.cloudwatch_metrics@package
{
  "aws": {
    "dynamic": true,
    "type": "object",
    "properties": {
      "dimensions_fingerprint": {
        "time_series_dimension": true,
        "type": "keyword"
      },
      "cloudwatch": {
        "properties": {
          "namespace": {
            "ignore_above": 1024,
            "type": "keyword"
          }
        }
      },
      "tags": {
        "type": "flattened"
      },
      "dimensions": {
        "type": "flattened"
      }
    }
  }
}
```

### Post the test document

Using the Dev Tools, post the following test document:

```text
// POST metrics-aws.cloudwatch_metrics-sdh5390/_doc
{
  "@timestamp": "2024-12-30T14:57:07+01:00",
    "aws": {
      "dimensions": {
        "name": "Maurizio Branca",
        "AutoScalingGroupName": "whatever"
        // "ImageId": "whatever",
        // "InstanceId": "whatever"
        //"InstanceType": "whatever"
      },
      "metric": {
        "cpu": 10
      }
  }
}
```

### Try to visualize the metric in Kibana

If I try to visualize the CPU metric in Kibana, the dimension subfields in `aws.dimensions.*` are not available:

CleanShot 2024-12-30 at 16 53 42@2x

### Upgrade the integration to this PR version

Bump the integration version (for example, 2.37.0), and build and update the integration:

```sh
elastic-package build && elastic-package stack up -d -v --services package-registry
```

### Delete the data stream and re-index the same test document

```text
// DELETE _data_stream/metrics-aws.cloudwatch_metrics-sdh5390

// POST metrics-aws.cloudwatch_metrics-sdh5390/_doc
{
  "@timestamp": "2024-12-30T14:57:07+01:00",
    "aws": {
      "dimensions": {
        "name": "Maurizio Branca",
        "AutoScalingGroupName": "whatever"
        // "ImageId": "whatever",
        // "InstanceId": "whatever"
        //"InstanceType": "whatever"
      },
      "metric": {
        "cpu": 10
      }
  }
}
```

If we check the mapping, we should now see a dynamic template:

CleanShot 2024-12-30 at 16 59 59@2x

### Re-try to visualize the metric in Kibana using the new mapping

With the updated mapping, we can now use `aws.dimensions.*` fields in Kibana to break down values by a dimension:

CleanShot 2024-12-30 at 17 01 32@2x

## Related issues

- Closes https://github.com/elastic/integrations/issues/11806


zmoog:
All `*_metrics` data streams use the `object` mapping. 

For example, here is the `dimensions` definition for EC2 metrics:

```yaml
# packages/aws/data_stream/ec2_metrics/fields/fields.yml
- name: aws
  type: group
  fields:
    - name: dimensions
      type: group
      fields:
        - name: AutoScalingGroupName
          type: keyword
          dimension: true
          description: An Auto Scaling group is a collection of instances you define if you're using Auto Scaling.
        - name: ImageId
          type: keyword
          dimension: true
          description: This dimension filters the data you request for all instances running this Amazon EC2 Amazon Machine Image (AMI)
        - name: InstanceId
          type: keyword
          dimension: true
          description: Amazon EC2 instance ID
        - name: InstanceType
          type: keyword
          dimension: true
          description: This dimension filters the data you request for all instances running with this specified instance type.
```

Since ALL metrics data streams use this mapping, it probably makes sense to use the `object` type in the `cloudwatch_metrics` as well instead of the `flattened` type. We can't do the opposite since TSDB doesn't allow `dimension: true` on `flattened` type.

packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml
```diff
@@ -5,8 +5,10 @@
       type: flattened
       description: |
         Tag key value pairs from aws resources.
-    - name: dimensions
-      type: flattened
+    - name: dimensions.*
```
tetianakravchenko, line 8: should be added explicitly `subobjects: false` ? for custom dimension keys

tetianakravchenko (commented):
please add a change log and the version change

packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml
zmoog: Yeah, but this requires a package-spec update, which needs several changes. We need to make this happen in a dedicated PR.

zmoog: I added a task to upgrade the package-spec to 3.1.0+

packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml
```diff
@@ -5,8 +5,10 @@
...
+      type: object
+      object_type: keyword
+      object_type_mapping_type: "*"
```
felixbarny, line 11: Isn't `*` the default? If so, this can be removed.

packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml
```diff
@@ -5,8 +5,10 @@
...
       description: |
         Metric dimensions.
     - name: dimensions_fingerprint
```
felixbarny, line 14: I think we could remove the fingerprint in favor of mapping `aws.dimensions.*` as dimensions. If we decide to do that, we should apply some of the learnings from the compatibility issues introduced in the prometheus integration when we removed the fingerprint and also perform dedicated upgrade tests.

packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml
```diff
@@ -5,8 +5,10 @@
       type: flattened
       description: |
         Tag key value pairs from aws resources.
-    - name: dimensions
-      type: flattened
+    - name: dimensions.*
+      type: object
+      object_type: keyword
```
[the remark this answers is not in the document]
tetianakravchenko: based on the definition of dimension https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_Dimension.html - for both `Name` and `Value`: `Type: String`

packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml
zmoog: > There is a meta issue - https://github.com/elastic/integrations/issues/9910, I think such change should be part of this meta issue

Yeah, I'm also in favour of leaving this change to the meta issue https://github.com/elastic/integrations/issues/9910 and keep the PR focus to have the same `aws.dimensions` mapping across the AWS integrations.

WDYT @felixbarny ?

packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml
zmoog: The PR to bump the package-spec version to 3.3.0 is https://github.com/elastic/integrations/pull/11893

@tetianakravchenko, are you okay to bump package-spec to 3.3.0 and introduce `subobjects: false` in dedicated PRs?

packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml
zmoog: If I am not looking in the wrong place, the [object_type_mapping_type](https://github.com/elastic/package-spec/blob/b10a15dac78de3a41c155d0a22eff58589851f49/spec/integration/data_stream/fields/fields.spec.yml#L357-L374) definition does not have a default value.

felixbarny: What I meant is that I think you can remove `object_type_mapping_type` completely which will also match all mapping types, as no explicit subset of mapping types have been specified.

packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml
felixbarny: I'd split this into two parts. What I think we should do in this PR is to do the same what we ended up doing with the Prometheus integration - map `dimensions.*` as TSDB dimensions, keep the fingerprint processing and field but not map it as a dimension. As a second step, in the context of https://github.com/elastic/integrations/issues/9910, we can then work on removing the fingerprinting entirely.

> keep the PR focus to have the same aws.dimensions mapping across the AWS integrations.

Sounds like what I'm proposing would be aligned with that? However, I think we shouldn't add `dimensions_fingerprint` to the more specific aws integrations that previously didn't have that field.

zmoog: Based on previous comments and an offline conversation with @felixbarny, I'm changing the `aws.dimensions` definition as follows:

```yaml
- name: aws
  type: group
  fields:
    - name: dimensions.*
      type: keyword
      dimension: true
      description: |
        Metric dimensions.
```

This creates a dynamic template that can map all dimensions as TSDB dimension. With this, we no longer need to keep `dimensions_fingerprint`, so I am removing this field from the mapping and the related pipeline processor (and, since it was the only processor, I removed the entire pipeline).

After installing the integration on my local stack, here are the results.

The component template:

CleanShot 2024-12-30 at 16 26 13@2x

With these settings, when I post a test document (1), each field under `aws.dimensions` (2) becomes a TSDB dimension:

CleanShot 2024-12-30 at 16 27 49@2x

felixbarny: > I am removing this field from the mapping and the related pipeline processor

Unfortunately, this will lead to issues when users upgrade from the previous version of the integration to the new one, leading to the same data loss issues we had with the Prometheus remote write integration. 

The problem is that when rolling over a TSDS, the rollover doesn't happen immediately, but only after the [look-ahead time](https://www.elastic.co/guide/en/elasticsearch/reference/current/tsds.html#tsds-look-ahead-time) has passed. However, the change to the ingest pipeline happens immediately. This leads to a situation where the pipeline isn't creating the fingerprint anymore, and the `dimensions.*` fields aren't mapped as `time_series_dimension`s. As a result, there will be no field that's mapped as a dimension, causing ingestion to fail.

In order to avoid that, I'd suggest keeping the pipeline that is doing the fingerprinting, and keeping the mapping for the dimensions fingerprint. But you can remove mapping the fingerprint field as a dimension.

zmoog:
> Could you add a section in the description on how to test this locally. A use case that would prove that this uniformity solves the problem of data correlation?

I guess the main issue for users is (quoting @felixbarny) "[they] can’t build visualizations with flattened fields in Lens, because Kibana doesn’t recognize fields within flattened objects."

I updated the "how to test this locally" section accordingly.

packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml
zmoog: Ouch. 

Thanks for the heads up, I'm restoring the fingerprint then.

zmoog: `dimensions_fingerprint` field restored!

felixbarny: Yes, we need to keep mapping the field. But we don't need to map it as a dimension, as all `dimension.*` fields are mapped as a dimension.

zmoog: The current index will continue to map field using the "old" mappings and when ES performs the rollover, and then we'll get the "new" mappings. Got it.

felixbarny (approved):
LGTM!
Please create follow-ups for removing `dimensions_fingerprint` (https://github.com/elastic/integrations/issues/9910), and using `subobjects: false` for `dimensions.*`.

zmoog:
I tested the upgrade from AWS integration 2.36.2 to 2.37.0.

- Installed AWS integration 2.36.2 (with `aws.dimensions` as `flattened`)
- Started sending 1 document every 5 secs with a field containing a sequence number
- Upgraded the AWS integration to 2.37.0 (unreleased, this PR)
- Waited for the rollout to take effect (checking `settings.index.time_series | .start_time, .end_time`)
- Checked that the the data stream didn't lose any sequence number

More details on selected steps.

## Started sending 1 document every 5 secs with a field containing a sequence number

I sed the following shell script:

```sh
sequence=0
while true
do
cat > metrics.json <<EOF
{
  "@timestamp": "$(date '+%Y-%m-%dT%H:%M:%S%z')",
    "aws": {
      "dimensions": {
        "name": "Maurizio Branca",
        "AutoScalingGroupName": "whatever"
      },
      "metric": {
        "cpu": 10,
        "sequence": $sequence
      }
  }
}  
EOF

((sequence++))

cat metrics.json | jq -c | es docs bulk -f - -i metrics-aws.cloudwatch_metrics-sdh5390
sleep 5
done
```

The scripts sends a document like the following every 5 secs:

```json
{
  "@timestamp": "2024-12-31T00:14:58+0100",
  "aws": {
    "dimensions": {
      "name": "Maurizio Branca",
      "AutoScalingGroupName": "whatever"
    },
    "metric": {
      "cpu": 10,
      "sequence": 270
    }
  }
}
```

## Waited for the rollout to take effect

Right after the upgrade, Fleet/ES start a new index with contiguous `settings.index.time_series | .start_time, .end_time`

Old index:

```
{
  ".ds-metrics-aws.cloudwatch_metrics-sdh5390-2024.12.30-000001": {
    "settings": {
      "index": {
        "mapping": {
          "total_fields": {
            "limit": "1000",
            "ignore_dynamic_beyond_limit": "true"
          }
        },
        "hidden": "true",
        "time_series": {
          "end_time": "2024-12-30T23:05:37.000Z",
          "start_time": "2024-12-30T20:35:37.000Z"
        },
```

New index:

```
{
  ".ds-metrics-aws.cloudwatch_metrics-sdh5390-2024.12.30-000002": {
    "settings": {
      "index": {
        "mapping": {
          "total_fields": {
            "limit": "1000",
            "ignore_dynamic_beyond_limit": "true"
          }
        },
        "hidden": "true",
        "time_series": {
          "end_time": "2024-12-30T23:35:37.000Z",
          "start_time": "2024-12-30T23:05:37.000Z"
        },
```

## Checked that the the data stream didn't lose any sequence number

At `2024-12-30T23:05:37.000Z`, ES started to write to the new index without losing a sequence number:

CleanShot 2024-12-31 at 00 06 09@2x

felixbarny:
The testing methodology looks great!

Longer-term, I wish we had something similar to this for all integrations as automated upgrade tests. The only thing that should be required in an integration is to provide a templated sample document. The rest could be a generic infrastructure. WDYT @jsoriano?

jsoriano:
> Longer-term, I wish we had something similar to this for all integrations as automated upgrade tests. 

We have an open issue about testing integration upgrades: https://github.com/elastic/elastic-package/issues/1831

> The only thing that should be required in an integration is to provide a templated sample document. The rest could be a generic infrastructure.

Some integrations start dummy services with docker compose to generate events. But for this case we'd be missing the validation on the sequence numbers.

jsoriano (approved):

packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml
tetianakravchenko: @zmoog sure, lets add it in dedicated PR 👍

tetianakravchenko (commented):
@zmoog please add changelog entry and change the version of the integration

tetianakravchenko:
@zmoog great approach to test this PR - https://github.com/elastic/integrations/pull/11883#issuecomment-2566001814! And thank you for sharing!

ritalwar (approved):
LGTM! Just ensure the changelog entry and version update as suggested already.

chemamartinez (approved):
LGTM

zmoog:
@tetianakravchenko, changelog updated!

tetianakravchenko (approved):

Issue #11603, opened 2024-10-31
[AWS docs] Required permissions under AWS custom logs


octavioranieri:
Currently we only mention the required permissions under the parent AWS integration, which would cover all the AWS integrations:
https://www.elastic.co/docs/current/integrations/aws#aws-permissions

Users may need more granular of what’s _required_ for the specifics instead of having all the permissions.

[AWS custom logs docs](https://www.elastic.co/docs/current/integrations/aws_logs) should have the specific required permissions, similar to the Filebeat module:
https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-aws-s3.html#_aws_permissions_2

lucabelluccini:
The request here could be expanded further, where we should have a way to include the documentation associated to a specific input in all integrations using it.

E.g. `aws-s3` input  is used in several integrations (O11y & Security owned) and in all of them generally the settings and prerequisites are similar.

lucabelluccini:
FYI @alaudazzi / @eedugon do you know what might be the best way to put this under the radar?

alaudazzi:
@lucabelluccini Thanks for raising this! It makes sense to provide more granular info where needed and consolidate similar settings to facilitate discoverability. I can follow up on this ticket.

alaudazzi:
@lucabelluccini 
I'm working on this ticket and I try to recap below what is needed:
- be more specific about the required permissions (check the [Filbeat module](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-aws-s3.html#_aws_permissions_2) as a good example)
- replicate/reference them in all the integrations that apply

Can you confirm this is correct?
Who can provide the input to update the docs?

alaudazzi:
@lucabelluccini 
Thank you for your feedback. It makes sense to track the second part in a separate issue: https://github.com/elastic/integrations/issues/11793.

alaudazzi:
CC @kaiyan-sheng

Issue #11793, opened 2024-11-20
[Integrations docs] Add permissions set for specific input types


alaudazzi:
As a follow-up on https://github.com/elastic/integrations/issues/11603 and https://github.com/elastic/integrations/issues/6771, this issue proposes to provide more granular information where needed, and consolidate similar settings for the various input types. A few points to consider:
- the integrations that use the same input (for ex. all integrations that use AWS S3) should refer to the same permission set
- locate the permissions set in one place to facilitate content standardization and maintenance
- engage with integrations devs & doc team as this effort might require some sort of include/import docs in each integration
- this doc enhancement affects both the O11y and the Security organization

lucabelluccini:
Ideally, this should be done per datastream and not per integration.
Each data stream can have a different input and therefore different requirements/permissions.

Example with `AWS Integration`.
It contains a huge amount of data streams / sub-integrations.
Some of them rely on `aws-s3` input (polling or SQS), others on `aws-cloudwatch` input, other `httpjson` (e.g. CloudTrail logs from third party REST API). For each one, there's a separate set of requirements and credentials.

alaudazzi:
CC @haetamoudi

PR #11891, opened 2024-11-27
[AWS docs] Add required permissions for Custom AWS Logs

alaudazzi:
This PR adds the specific permissions required to use the [Custom AWS Logs](https://www.elastic.co/docs/current/integrations/aws_logs) integration, both for `aws-s3` input and `aws-cloudwatch` input.

Closes https://github.com/elastic/integrations/issues/11603.


lucabelluccini (commented):
Hello @alaudazzi 
I would involve the owners / developer of the `aws_logs` integration for a double check.

To me, the pointers to the Filebeat `aws-s3` doc and `cloudwatch` doc make sense,  but I would have expected that the whole `Requirement` section of https://www.elastic.co/docs/current/integrations/aws#requirements would have been ported to https://www.elastic.co/docs/current/integrations/aws_logs.
IIUC, the permissions should be similar (just the AWS integration Requirement section mixes up both the permissions required for S3 and Cloudwatch).
Obviously the "cost" of this is we're duplicating the content.

If instead we want to tackle this problem properly in the follow up issue https://github.com/elastic/integrations/issues/11793, then this PR should be ok to allow users to understand which permissions are required.

I would ask @octavioranieri to review.

octavioranieri:
+1 to @lucabelluccini suggestions above. 

It's better to have it within the page (and product) instead of referencing to an external page.
we already do that under the parent page for https://www.elastic.co/docs/current/integrations/aws#aws-permissions

alaudazzi:
Duplicating content often leads to maintenance issues that are difficult to track.

kaiyan-sheng (approved):

kaiyan-sheng:
@alaudazzi I agree with Luca on `the whole Requirement section` should be included as well. This is not necessarily permissions but should be mentioned as required for using the integration.

alaudazzi:
@kaiyan-sheng @octavioranieri 
Thank you for your feedback.

@lucabelluccini 

> If instead we want to tackle this problem properly in the follow up issue https://github.com/elastic/integrations/issues/11793, then this PR should be ok to allow users to understand which permissions are required.

Yes, let's do that. 

Please note that one of the points highlighted in the follow-up issue is `locate the permissions set in one place to facilitate content standardization and maintenance`.

Issue #11880, opened 2024-11-26
[Aruba] Support the first 40-50 event types for the Aruba integration


qcorporation:
Support the below event types:

1. IRDP events : https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/IRDP.htm
2. L3 Encap capacity events : https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/L3_ENCAP.htm
3. L3 Resource Manager events : https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/L3_RESMGR.htm
4. LACP events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/LACP.htm
5. LAG Events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/LAG.htm
6. Layer 3 Interface Events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/L3INTERFACE.htm
7. LED Events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/LED.htm
8. LLDP Events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/LLDP.htm
9. Loop Protection Events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/LOOP-PROTECT.htm
10. Loopback Events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/LOOPBACK.htm

PR #11905, opened 2024-11-27
Support the first 40-60 event types for the Aruba integration

qcorporation:
Github Tracking Issue:
https://github.com/elastic/integrations/issues/11880 and https://github.com/elastic/integrations/issues/11906

Change Log:
Added parsing logic, doc, pipeline tests for:

- IRDP events
- L3 Encap capacity events
- L3 Resource Manager events
- LACP events
- LAG Events
- Layer 3 Interface Events
- LED Events
- LLDP Events
- Loop Protection Events
- Loopback Events
- MAC Address mode configuration events
- MAC Learning events
- MACsec events
- Management events
- MDNS events
- MGMD events
- Mirroring events
- Module events
- MSDP events
- Multicast Traffic Manager events

Test Cases:
- reviewed each log message added was processed properly within the pipeline tests

### DIFF
--- a/packages/hpe_aruba_cx/_dev/build/docs/README.md
+++ b/packages/hpe_aruba_cx/_dev/build/docs/README.md
@@ -562,91 +562,96 @@ Note: Descriptions have not been filled out
 
 
 #### [MAC Address mode configuration events](https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/L3_MAC_ADDRESS_CONFIGURATION.htm)
-| Field               | Description | Type | Common                       |
-|---------------------|-------------|------|------------------------------|
-| aruba.mac.mac       |             |      | server.mac                   |
-| aruba.mac.new_mode  |             |      |                              |
-| aruba.mac.old_mode  |             |      |                              |
-| aruba.mac.vlan      |             |      | network.vlan.id              |
+| Docs Field | Schema Mapping       |
+|------------|----------------------|
+| <mac>      | server.mac           |
+| <new_mode> | aruba.mac.new_mode   |
+| <old_mode> | aruba.mac.old_mode   |
+| <vlan>     | network.vlan.id      |
 
 #### [MAC Learning events](https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/MAC-LEARN.htm)
-| Field               | Description | Type | Common                       |
-|---------------------|-------------|------|------------------------------|
-| aruba.mac.from-intf |             |      | observer.ingress.interface.name |
-| aruba.mac.mac       |             |      | server.mac                   |
-| aruba.mac.to-intf   |             |      | observer.egress.interface.name |
-| aruba.mac.vlan      |             |      | network.vlan.id              |
+| Docs Field  | Schema Mapping               |
+|-------------|------------------------------|
+| <from-intf> | aruba.interface.prev_id      |
+| <intf>      | aruba.interface.id           |
+| <mac>       | server.mac                   |
+| <to-intf>   | aruba.mac.interface.id       |
+| <vlan>      | network.vlan.id              |
 
 #### [MACsec events](https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/MACSEC.htm)
-| Field               | Description | Type | Common                       |
-|---------------------|-------------|------|------------------------------|
-| aruba.mac.ckn       |             |      |                              |
-| aruba.mac.ifname    |             |      | observer.ingress.interface.name |
-| aruba.mac.latest_an |             |      |                              |
-| aruba.mac.latest_kn |             |      |                              |
-| aruba.mac.old_an    |             |      |                              |
-| aruba.mac.old_kn    |             |      |                              |
-| aruba.mac.sci       |             |      |                              |
+| Docs Field   | Schema Mapping               |
+|--------------|------------------------------|
+| <ckn>        | aruba.mac.ckn                |
+| <ifname>     | aruba.interface.name         |
+| <latest_an>  | aruba.mac.latest_an          |
+| <latest_kn>  | aruba.mac.latest_kn          |
+| <old_an>     | aruba.mac.old_an             |
+| <old_kn>     | aruba.mac.old_kn             |
+| <sci>        | aruba.mac.sci                |
 
 #### [Management events](https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/MGMT.htm)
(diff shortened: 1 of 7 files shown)


packages/hpe_aruba_cx/docs/README.md
```diff
@@ -454,108 +454,111 @@ Note: Descriptions have not been filled out
...
 
 #### [Loopback events](https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/LOOPBACK.htm)
-| Field                  | Description | Type | Common                       |
-|------------------------|-------------|------|------------------------------|
-| aruba.loopback.interface |           |      | observer.ingress.interface.name |
-| aruba.loopback.status  |             |      | aruba.status                 |
+| Docs Field | Schema Mapping       |
+|------------|----------------------|
+| <interface> | aruba.interface.id   |
+| <state>     | aruba.status         |
```
gogochan, line 561: this should go to `aruba.interface.state`

packages/hpe_aruba_cx/docs/README.md
```diff
@@ -454,108 +454,111 @@ Note: Descriptions have not been filled out
...
-| aruba.lldp.pvid        |             |      | network.vlan.id              |
-| aruba.lldp.reinit_delay|             |      |                              |
-| aruba.lldp.tx_delay    |             |      |                              |
-| aruba.lldp.tx_hold     |             |      |                              |
-| aruba.lldp.tx_timer    |             |      |                              |
+| Docs Field   | Schema Mapping               |
+|--------------|------------------------------|
+| <chassisid>  | aruba.instance.id            |
+| <interface>  | aruba.interface.id           |
+| <ninterface> | aruba.lldp.ninterface        |
```
gogochan, line 539: I'm wondering if this also needs to be aruba.interface.id

qcorporation: we need `ninterface` because it's being using within the same line where `aruba.interface.id` is being used. It needs two fields to parse the event type out properly

packages/hpe_aruba_cx/docs/README.md
qcorporation: added changes to reference `aruba.interface.state`

qcorporation:
While there was time this week lumped in support for https://github.com/elastic/integrations/issues/11906 into this PR: @dwhyrock @gogochan

packages/hpe_aruba_cx/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -238,9 +255,93 @@ processors:
...
+      patterns:
+        - "^Loopback Interface %{DATA:aruba.interface.id}, (created|deleted|configured administratively %{GREEDYDATA:aruba.interface.state})"
+
+    # LACP events (13xx)
+    # https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/LACP.htm
+  - grok:
+      field: message
+      tag: lacp_event_1301_through_1323
+      description: "Dynamic LAG has event has been raised"
+      if: "['1301','1302','1303','1304','1305','1306','1307','1308','1309','1310','1311','1312','1313','1314','1315','1316','1317','1318','1319','1320','1321','1322','1323'].contains(ctx.event?.code)"
```
dwhyrock, line 284: I don't think we'll really know this until we get benchbuilder in a more advanced state, but I feel like this would be pretty inefficient.

I would think that for every matching Event ID (`1301`, `1302`, etc), it will try and match it against the `patterns` list, one by one, until it either gets a match or runs out of patterns.  Is that correct, @taylor-swanson ?

If so, it seems like it would be much more efficient (though larger in `default.yml` size) to add a separate `grok` entry for most of these so that the `if:` block is the main thing being compared instead of a regex pattern.  The exception may be for messages that are very similar in structure.

dwhyrock (approved):
Approved.

I did note one concern in the combining of a long list of fairly divergent regexes into one `patterns` grouping.  We probably won't know the performance impact until benchbuilder is more mature, but it feels like separating them out and letting the `if:` block do the quicker short-circuiting would improve performance.

packages/hpe_aruba_cx/_dev/build/docs/README.md
```diff
@@ -454,196 +454,204 @@ Note: Descriptions have not been filled out
...
-| aruba.mac.ckn       |             |      |                              |
-| aruba.mac.ifname    |             |      | observer.ingress.interface.name |
-| aruba.mac.latest_an |             |      |                              |
-| aruba.mac.latest_kn |             |      |                              |
-| aruba.mac.old_an    |             |      |                              |
-| aruba.mac.old_kn    |             |      |                              |
-| aruba.mac.sci       |             |      |                              |
+| Docs Field   | Schema Mapping               |
+|--------------|------------------------------|
+| <ckn>        | aruba.mac.ckn                |
```
gogochan, line 584: should we use `aruba.macsec` to separate it from `aruba.mac` from the previous event?

packages/hpe_aruba_cx/data_stream/log/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: That's generally where I lean on this as well, keeping them separate. If there are overlapping event IDs and grok patterns, those can be kept together, but orthogonal ones should be separated out.

It's hard to say for sure without actual numbers, but generally, a simple if statement should be faster than not matching grok pattern.

qcorporation: Thanks, @taylor-swanson and @dwhyrock, for the call out - I've updated the PR to break up some of the larger groks, mainly:
- split up LACP events (13xx)
- split up Layer 3 Interface events (17xx)
I think it is more readable and it will be more performant based upon the if logic

packages/hpe_aruba_cx/_dev/build/docs/README.md
qcorporation: @gogochan thanks for the suggestion, I lumped in Mac address mode config, Mac learning and Mac sec events into `aruba.mac.*` events. Would you like for me to separate them all out?

gogochan: I was asking for an opinion. if you feel that mac name space is sufficient, that is fine.

qcorporation:
> > While there was time this week lumped in support for #11906 into this PR: @dwhyrock @gogochan
> 
> This PR has too many changes to review and will take very long for the reviewers, and the quality of the review will suffer too. It would make things a lot easier if the PR only had a section at a time.

Thank @gogochan, for the call out - I'll make them smaller next time.

PR #11914, opened 2024-11-28
[Synthetics] fix mapping from long to float

shahzad31:
we made a mistake instead of using float we are using long for decimal values.

for PR https://github.com/elastic/integrations/pull/11878


packages/synthetics/data_stream/browser/fields/synthetics.yml
```diff
@@ -81,29 +81,29 @@
...
           fields:
             - name: us
-              type: long
+              type: float
         - name: load
           type: group
           description: duration of Load end event
           fields:
             - name: duration
-              type: long
```
shahzad31, line 102: this we reverted to original

packages/synthetics/data_stream/browser/fields/synthetics.yml
```diff
@@ -81,29 +81,29 @@
...
+              type: integer
             - name: us
-              type: long
```
shahzad31, line 104: same here

emilioalvap (approved):
LGTM

maryam-saeidi (approved):
Rubber stamp from my side, as Emilio reviewed this PR.

PR #11929, opened 2024-11-29
Fix broken link for Prebuilt Security Detection Rules 

alaudazzi:
This PR fixes a broken link to https://www.elastic.co/guide/en/security/master/rules-ui-management.html#import-export-rules-ui

Relates https://github.com/elastic/integration-docs/issues/586


packages/security_detection_engine/changelog.yml
```diff
@@ -1,5 +1,10 @@
 # newer versions go on top
 # NOTE: please use pre-release versions (e.g. -beta.0) until a package is ready for production
+- version: 8.16.3
```
shashank-elastic, line 3: I don't see why a version bump is being done to fix readme docs here, which will not include any rule updates. Which means the rule packages will be as good as 8.16.2

We could combine this fix in our upcoming release schedule along with rule package changes, if this is not a critical fix that needs be published right away.

We have a release scheduled on Dec 9 for 8.16.3 version, along with other supported versions, which can accommodate this change for 8.16.3 and future versions.
cc @Mikaayenson

shashank-elastic:
Apologies, this was missed in the last release train, we ran into issues of packages, which made me err and forget we had to piggy back this change. 

The change will fix the broken documentation, here where there will be no updates to the package. 

image

My only suggestion would be make this beta 8.17.2-beta.1 and we will GA it together with package updates ? will that work @alaudazzi

packages/security_detection_engine/changelog.yml
```diff
@@ -1,5 +1,10 @@
 # newer versions go on top
 # NOTE: please use pre-release versions (e.g. -beta.0) until a package is ready for production
+- version: 8.17.2
```
shashank-elastic, line 3:
```suggestion
- version: 8.17.2-beta.1
```

shashank-elastic (approved):
Approving the beta change to GA this along with packages updates in Jan 2025.

packages/security_detection_engine/manifest.yml
```diff
@@ -21,4 +21,4 @@ source:
   license: Elastic-2.0
 title: Prebuilt Security Detection Rules
 type: integration
-version: 8.17.1
+version: 8.17.2
```
shashank-elastic, line 24:
```suggestion
version: 8.17.2-beta.1
```

PR #11944, opened 2024-12-02
Restore Elastic-Agent Filbeat beat.stats metrics

alexsapran:
## Proposed commit message

This commit restores the mapping for Filebeat's `beat.stats` metrics when running inside Elastic-Agent, which was dropped due to TSDB.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Some context https://github.com/elastic/ingest-dev/issues/4403#issuecomment-2507704967


packages/elastic_agent/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.0.4"
+  changes:
+    - description: Add back elastic-agent Filbeat beat.stats metrics that dropped due to TSDB
```
cmacknz, line 4: It is safe to do the same change for Metricbeat as well https://github.com/elastic/integrations/tree/main/packages/elastic_agent/data_stream/metricbeat_metrics

Probably all the beat metrics data streams could have these mappings.

cmacknz (approved):
LGTM, mappings are the same as the ones in https://github.com/elastic/integrations/blob/main/packages/elastic_agent/data_stream/elastic_agent_metrics/fields/beat-stats-fields.yml

If you want to save future us another trip in here in the near future, you can add the same mappings to Metricbeat. Probably all the Beats from the Beats repository are safe to have this change applied.

alexsapran:
>If you want to save future us another trip in here in the near future, you can add the same mappings to Metricbeat. Probably all the Beats from the Beats repository are safe to have this change applied.

I was not sure if I should update everything in one go. But if you think the same then it might make sense to do it in this PR as you said. Thanks for double-checking it!

AndersonQ (approved):

PR #11946, opened 2024-12-02
[Juniper SRX] Populate event fields for system log

### LINKED ISSUE #11834: [Juniper SRX] Add event.category and event.outcome to failed SSH login attempt
We have added Juniper switches to this integration, but we see no SIEM alerts on failed SSH attempts to this switch.
Looking at this integration, it seems like this can be added easily by doing the following steps.

Original event:
```
<37>1 2024-11-22T12:45:00.207+01:00 ams-edge-sw-003 sshd - SSHD_LOGIN_FAILED [junos@2636.1.1.1.4.82.30 username="user1" source-address="192.168.0.3"]
```
This are the relevant fields as parsed by the integration:

juniper.srx.tag: SSHD_LOGIN_FAILED
source.user.name: user1

It seems to me that by filtering for "SSHD_LOGIN_FAILED" and then setting event.category (authentication) and event.outcome (failed ?) this can be easily added to existing SIEM alerts.

### LINKED ISSUE #11835: [Juniper SRX] Add host.name to system logs
We would like to see the field `host.name` filled in with the value from field `observer.name` when the logs are from the device itself (system log), at the moment host.name is not filled in in that case.

Oddly:
## Proposed commit message
The purpose of this code change is to make failed SSH login attempts to Juniper equipment visible in Elastic Security.
To this end, we populate the following fields on a failed SSH login attempt in the system logging pipeline.

(if SSH login failed)
- event.outcome: failure
- event.category: authentication (append)

(no conditions)
- observer.name -> host.name (copy)

## Checklist

- `elastic-package test` in the package directory.
## Related issues

- Closes #11834
- Closes #11835

### DIFF
--- a/packages/juniper_srx/changelog.yml
+++ b/packages/juniper_srx/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.21.3"
+  changes:
+    - description: Populated event.outcome and event.category to system log on SSH login failure.
+      type: enhancement
+      link: https://github.com/elastic/integrations/issues/11834
 - version: "1.21.2"
   changes:
     - description: Use triple-brace Mustache templating when referencing variables in ingest pipelines.
--- a/packages/juniper_srx/changelog.yml
+++ b/packages/juniper_srx/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "1.21.3"
   changes:
-    - description: Populated event.outcome and event.category to system log on SSH login failure.
+    - description: Populated event.outcome and event.category to system log on SSH login failure and added host.name field to system log.
       type: enhancement
       link: https://github.com/elastic/integrations/issues/11834
 - version: "1.21.2"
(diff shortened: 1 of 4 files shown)


Oddly:
As an addition, to make the ML job “Failed logon” work, I think we need to populate the host.name field as well.
Some options:
- We can just copy the observer.name field to the host.name field.
- We can do the above and remove the observer.name field.

The second option seems the most logical since system log doesn’t really have an observer.

taylor-swanson:
Changes so far LGTM.

> As an addition, to make the ML job “Failed logon” work, I think we need to populate the host.name field as well. Some options:
> 
> * We can just copy the observer.name field to the host.name field.
> * We can do the above and remove the observer.name field.
> 
> The second option seems the most logical since system log doesn’t really have an observer.

Since this failed login refers to a login to the equipment itself, setting `host.name` makes sense here. I'm hesitant about dropping the observer.name field. If we did that, then we'd probably want to drop the other observer fields. We'd lose some metadata information in the log (vendor/product/type) and there's also the risk that we may break rules/dashboards other users have set up. For that reason, I would have to consider this a breaking change. I don't think it's worth the hassle, and I would recommend just copying the value to `host.name`.

Oddly:
Hi Taylor,

Agreed, I will add these changes in the coming days.
Do you agree that `observer.name` can be copied to `host.name` in the `system.yml` pipeline without condition?

taylor-swanson:
> Hi Taylor,
> 
> Agreed, I will add these changes in the coming days. Do you agree that `observer.name` can be copied to `host.name` in the `system.yml` pipeline without condition?

The only condition I would add is `ignore_empty_value: true`. Otherwise, I don't see `host.name` being set anywhere in the default or system pipeline. Since this is limited to the system pipeline, I think this should be fine (I would otherwise suggest `override: false` if we were worried about overwriting an existing value).

Oddly:
That should be it, I think we can run the test and if successful, merge. Thanks for looking into it.

taylor-swanson (approved):
LGTM!

PR #11949, opened 2024-12-02
[tanium] Extra field presence checks, truncation of excessive depth

chrisberkhout:
## Proposed commit message

```
[tanium] Extra field presence checks, truncation of excessive depth

Fixes for some new test data that was showing:
- Errors attempting to run conversions on absent fields.
- Events being dropped because they are too deep.

Checking field presence before conversion was difficult because those
processors were run in `foreach` loops, and Painless doesn't get access
to `_ingest._value`. This was fixed by extracting separate pipelines,
which could make the current value accessible at the start, then set it
back after processing.

To avoid dropping events with fields that are too deeply nested (more
than a depth of 20), some depth truncation logic was added.

There were also a number of field definitions added in order to make
tests pass for the new test data.
```

Can be read commit-by-commit.

## Checklist

Increasing `maxDepth` by `1` will make the system test fail due to an event being dropped for excessive depth.


packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -333,10 +330,49 @@ processors:
...
+      ignore_missing: true
+
+  # Add the data from the Match Details column to any other match details data
+  - script:
+      tag: script_handle_match_details_content_the_same_as_when_it_appears_in_an_encoded_payload
+      description: Handle 'Match Details' content the same as when it appears in an encoded payload.
+      lang: painless
+      if: ctx.json['Match Details'] instanceof Map
+      source: |
+        if (ctx.tanium.threat_response.match_details == null) ctx.tanium.threat_response.match_details = [:];
```
efd6, line 367:
```
ctx.tanium.threat_response.match_details = ctx.tanium.threat_response.match_details ?: [:];
```
?

packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-threat-response.log-expected.json
```diff
@@ -116,226 +116,222 @@
...
+                                                    "instance_hash": "9667017345150968342"
+                                                },
+                                                "timestamp": "2023-01-18T10:37:18.000Z",
+                                                "verb": 6
                                             }
+                                        ]
+                                    },
+                                    "intel_intra_ids": [
+                                        {
+                                            "id": "850394582"
```
efd6, line 255: Can these be retained as numeric keywords?

packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/match_details_whats.yml
```diff
@@ -0,0 +1,260 @@
+---
+description: Pipeline for processing tanium.threat_response.match_details.finding.whats[]
+processors:
+
+  - rename:
+      tag: rename_ingest_value
+      field: _ingest._value
+      target_field: _what
```
efd6, line 8: This is at the same time horrifying and very clever.

packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -403,6 +403,56 @@ processors:
...
+                }
+            }
+            return node;
+        }
+        int maxDepth = 20;
+        List truncations = [];
+        truncateMap(ctx, 1, maxDepth, truncations);
+        if (truncations.size() > 0) {
+          if (ctx.labels == null) ctx.labels = [:];
+          ctx.labels.truncations = truncations.size().toString();
```
efd6, line 454: Rather than just reporting the count of truncations, perhaps this should also be a list of JSON paths. This would need to be stored under `tanium`,

chrisberkhout: I did it, listing the truncation points. That means with
```
key1.key2.key3: "value4"
```
and `maxDepth` of `3`, the value `"value4"` will be set to null (then later `key3` will be removed because it has a null value) and this will be set:
```
tanium.truncations: ["key1.key2.key3"]
```
The `maxDepth` of `20` is the correct one, because that's the greatest number that avoids dropped events in the system test.

packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -316,15 +316,12 @@ processors:
...
+  # Parse 'Other Parameters', decode any payload value and store it as match details data
   - kv:
-      tag: kv_tanium_threat_response_other_parameters_original
-      field: tanium.threat_response.other_parameters.original
-      target_field: tanium.threat_response.other_parameters.log_details
+      tag: kv_json_other_parameters
+      field: json.Other Parameters
+      target_field: tanium.threat_response.other_parameters
       field_split: "\\|"
       value_split: "="
```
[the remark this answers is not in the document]
chrisberkhout: Done.

packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Yep. It would be a problem if the left hand side of the assignment was primitive type variable.

I'm not sure it's clearer. It is a bit shorter, so I did change this one.

packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/match_details_whats.yml
chrisberkhout: Agreed. It's worse at the bottom:

```
  - set:
      tag: set_back_what
      field: _ingest._value
      copy_from: _what
  - remove:
      tag: remove_what
      field: _what

```

packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-threat-response.log-expected.json
chrisberkhout: I'm a bit confused about this. Do you mean set `numeric_keyword_fields` in the test config?
I see the old field name was there and I updated it now.

That test config seems to be a way to say "don't worry, this number will be mapped as a keyword".

In the original and new versions, the `intel_intra_ids.id` values are converted to strings, and are mapped as keywords. In the old data (unchanged) they arrive as numbers, in the new data they arrive as strings. Now, with the updated test config, the generated expected output is still always strings. When I re-generate on the original version, I do get numbers. I'm not sure why that's behaving differently.

Do you know what I'm missing here?

efd6: You're converting the integers to strings in the ingest pipeline. For them to be rendered as numbers, they need to be either converted to a number type, or as is the case here, not converted since they already come in as numbers.

So if I do this,
```diff
diff --git a/packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-common-config.yml b/packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-comm
on-config.yml
index c20f747cf7..fda3f0c3dc 100644
--- a/packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-common-config.yml
+++ b/packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-common-config.yml
@@ -3,5 +3,6 @@ fields:
     - preserve_original_event
     - preserve_duplicate_custom_fields
 numeric_keyword_fields:
+  - tanium.threat_response.id
   - tanium.threat_response.match_details.finding.whats.intel_intra_ids.id
   - tanium.threat_response.match_details.finding.whats.intel_intra_ids.id_v2
diff --git a/packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml b/packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
index 32d067d0a7..4fbec3f047 100644
--- a/packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
@@ -47,23 +47,17 @@ processors:
             field: error.message
             value: '{{{_ingest.on_failure_message}}}'
             allow_duplicates: false
+  - set:
+      tag: set_tanium_threat_response_id
+      field: tanium.threat_response.id
+      copy_from: json.id
+      ignore_empty_value: true
   - convert:
-      tag: convert_json_id
-      field: json.id
-      target_field: tanium.threat_response.id
+      tag: convert_event_id
+      field: tanium.threat_response.id
+      target_field: event.id
       type: string
       ignore_missing: true
-      if: ctx.json?.id != ''
-      on_failure:
-        - append:
-            field: error.message
-            value: '{{{_ingest.on_failure_message}}}'
-            allow_duplicates: false
-  - set:
-      tag: set_event_id
-      field: event.id
-      copy_from: tanium.threat_response.id
-      ignore_empty_value: true
   - rename:
       tag: rename_json_action
       field: json.action
```
I get the expected behaviour; the numeric keywords are numeric.

Could also do this
```diff
diff --git a/packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-common-config.yml b/packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-common-config.yml
index c20f747cf7..4f79f57fa3 100644
--- a/packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-common-config.yml
+++ b/packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-common-config.yml
@@ -3,5 +3,7 @@ fields:
     - preserve_original_event
     - preserve_duplicate_custom_fields
 numeric_keyword_fields:
+  - event.id
+  - tanium.threat_response.id
   - tanium.threat_response.match_details.finding.whats.intel_intra_ids.id
   - tanium.threat_response.match_details.finding.whats.intel_intra_ids.id_v2
diff --git a/packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml b/packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
index 32d067d0a7..a0bfd26f3d 100644
--- a/packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
@@ -47,22 +47,15 @@ processors:
             field: error.message
             value: '{{{_ingest.on_failure_message}}}'
             allow_duplicates: false
-  - convert:
-      tag: convert_json_id
-      field: json.id
-      target_field: tanium.threat_response.id
-      type: string
-      ignore_missing: true
-      if: ctx.json?.id != ''
-      on_failure:
-        - append:
-            field: error.message
-            value: '{{{_ingest.on_failure_message}}}'
-            allow_duplicates: false
   - set:
-      tag: set_event_id
+      tag: set_tanium_threat_response_id
+      field: tanium.threat_response.id
+      copy_from: json.id
+      ignore_empty_value: true
+  - set:
+      tag: set_tanium_threat_response_id
       field: event.id
-      copy_from: tanium.threat_response.id
+      copy_from: json.id
       ignore_empty_value: true
   - rename:
       tag: rename_json_action
```
Note that `convert` to `string` cannot fail and so does not need an `on_failure`.

As far as why this particular field was numeric before, I don't know. My suspicion is that this
```
  - foreach:
      tag: foreach_tanium_threat_response_other_parameters_log_details_payload_decoded_finding_whats_19
      field: tanium.threat_response.other_parameters.log_details.payload_decoded.finding.whats
      if: ctx.tanium?.threat_response?.other_parameters?.log_details?.payload_decoded?.finding?.whats instanceof List
      processor:
        convert:
          tag: convert_ingest_value_intel_intra_ids_id
          field: _ingest._value.intel_intra_ids.id
          type: string
          ignore_missing: true
          on_failure:
            - append:
                field: error.message
                value: '{{{_ingest.on_failure_message}}}'
                allow_duplicates: false
```
in the original never modifies the document. I removed it and it had no effect.

packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/match_details_whats.yml
efd6: Maybe call it `_wat`? :grin:

packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
efd6: I think the truncation point is probably a good compromise.

packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-threat-response.log-expected.json
chrisberkhout: Okay. I think your suspicion is right. I removed that conversion.

I added the `tanium.threat_response.id` in `numeric_keyword_fields` settings.

I did the simplification in the 2nd diff, except that I kept the conversion of `event.id`. The reason for that is that `ecs@mappings` will only make `event.id` a keyword if the incoming value is a string. If an integer value comes in first, that field will not be a keyword field, and later string values will be ignored values (and tests will fail).

So, for those fields we explicitly map as keywords, I set `numeric_keyword_fields` and don't convert them. The rest stays the same.

chrisberkhout:
@efd6 Last commit adds some extra logic to handle empty values see in new data. Includes a test case for that.

packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/match_details.yml
```diff
@@ -0,0 +1,331 @@
...
+      processor:
+        rename:
+          tag: rename_ingest_value_event_filecreate_path
+          field: _ingest._value.event.fileCreate.path
+          target_field: _ingest._value.event.file_create.path
+          ignore_missing: true
+  - convert:
+      tag: convert_tanium_threat_response_match_details_intel_id
+      field: tanium.threat_response.match_details.intel_id
+      type: string
```
efd6, line 143:
```suggestion
      type: long
```
(also below)

packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/match_details.yml
```diff
@@ -0,0 +1,331 @@
...
+      tag: convert_tanium_threat_response_match_details_config_rev_id
+      field: tanium.threat_response.match_details.config_rev_id
+      type: string
+      ignore_missing: true
+      if: ctx.tanium?.threat_response?.match_details?.config_rev_id != ''
+      on_failure:
+        - append:
+            field: error.message
+            value: '{{{_ingest.on_failure_message}}}'
+            allow_duplicates: false
```
efd6, line 107: convert to string can't fail, so this is not necessary.

packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-deep.log-expected.json
```diff
@@ -0,0 +1,256 @@
...
+            ],
+            "tanium": {
+                "threat_response": {
+                    "computer": {
+                        "ip": "67.43.156.65",
+                        "name": "hostname.example.com"
+                    },
+                    "match_details": {
+                        "config_id": "1000111",
+                        "config_rev_id": "1",
```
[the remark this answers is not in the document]
chrisberkhout: Done.

packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-threat-response.log-expected.json
```diff
@@ -116,226 +116,222 @@
...
                                                                     "domain": "root",
                                                                     "id": "6018826305876723269",
                                                                     "name": "root"
                                                                 }
                                                             }
                                                         },
-                                                        "pid": "43181",
-                                                        "start_time": "2023-01-18T10:37:18.000Z",
-                                                        "tanium_unique_id": "14156042576993495213",
+                                                        "pid": "2058",
```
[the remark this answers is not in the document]
chrisberkhout: Changed all `pid` and `ppid` fields to `long`. Confirmed all incoming values (in the test examples) for those fields are numbers. Removed redundant conversions.

packages/tanium/data_stream/threat_response/_dev/test/pipeline/test-threat-response.log-expected.json
```diff
@@ -116,226 +116,222 @@
...
-                                                                    "sha256": "e3998dbe02b51dada33de87ae43d18a93ab6915b9e34f5a751bf2b9b25a55492"
-                                                                },
-                                                                "instance_hash_salt": "3014767",
-                                                                "magic_number_hex": "3132372e",
-                                                                "modification_time": "2022-09-11T20:08:26.000Z",
-                                                                "path": "/var/lib/docker/overlay2/2bcffb7f0d6a3f37a6199ca691649ea49358a2c2e87f39202562e50eb5cd2805/diff/etc/hosts",
-                                                                "size_bytes": 79
-                                                            },
-                                                            "instance_hash": "9667017345150968342"
+                                                "pid": "43181",
```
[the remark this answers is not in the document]
chrisberkhout: Done above.

packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/match_details.yml
chrisberkhout: I added this field to ` numeric_keyword_fields` as suggested above, then I removed this conversion.

packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/match_details.yml
chrisberkhout: As above, this processor was removed.

I removed the `on_failure` from all other conversions to string.

efd6 (approved):

PR #11950, opened 2024-12-02
Update k8s dashboards to use links panel for navigation

tommyers-elastic:
This PR replaces the markdown navigation section at the top of most K8s dashboards with a native 'Links' panel which maintains filters between navigation.

I was not able to include the panel as a reference due to 'links' not being allowed in package-spec, so the json is repeated in every dashboard. I know we generally prefer inline visualisations, but in this case it seems like a reference would be significantly better from a maintanence POV. We should potentially resolve this by allowing links JSON in packages under `kibana/links`. 

Another sucky part of this change is that you have to scroll the links panel to get to the links 'further down' (on the right). If we want to be able to have the links fill more vertical space we need to take this up with the dashboards team. 

Fixes https://github.com/elastic/kibana/issues/202471


gizas:
@tommyers-elastic this should be kibana.version: "^8.16.0"
The feature is avaialble in 8.16 version right?

gizas:
In 8.16.1 works and not in 8.16.0 !!!

8.16.1 test:
Screenshot 2024-12-04 at 2 51 57 PM

tommyers-elastic:
@roshan-elastic @gizas 

sorry i let this go stale - Andreas was hitting an issue when testing this but i couldn't repro it on any version. 

not sure of the best path forward here. which team actually owns this functionality? in codeowners it's the old cloudnative team, which maps to hosted services now? can we get that team to take over on this? cc @bturquet

gizas:
I retested with latest elastic-package and in 8.17.0 version
All seem to work Tom. 

I still dont know why I was hitting this issue , (maybe old elastic-package?)

Approving
With 8.17.0. version
Screenshot 2025-01-08 at 2 56 20 PM

roshan-elastic (approved):
LGTM!

Issue #11251, opened 2024-09-26
Azure Logs: use one input per agent policy


zmoog:
## Situation

When users install the Azure Logs integration, they can specify only one event hub and enable multiple data streams.

![Image](https://github.com/user-attachments/assets/af19b858-3734-4381-a535-b5b6098d0999)

When users enable a data stream, the Azure Logs integration creates one input:

![Image](https://github.com/user-attachments/assets/ae20d572-db70-45dc-95c8-260449baa980)

## Problem

Each input created for a data stream (e.g., the input 1 from the previous screenshot) is unaware of other inputs. When multiple inputs access an event hub in an uncoordinated way, they create contention for the event hub partitions.

Each input assumes it is the only one accessing the event hub. When the number of inputs reading data from an event hub is higher than 5, they disconnect each other.

### Solution 1: one input with routing

Each Azure Logs integration install must spawn only one input and route log event to the intended destination using routing (routing rule or reroute processor).

Do you want to try this solution on the current version of Azure Logs?

Please follow the steps at https://github.com/zmoog/public-notes/issues/92 to set up log event ingestion using one input + routing.

PR #11984, opened 2024-12-03
[azure logs] add routing integration to use only one azure-eventhub input

zmoog:
## Proposed commit message

Switch the integration package from the one-input-per-data-stream model to the one-input model.

One input per data stream model:

image

One input model:

image

In the one-input model, there is only one azure-eventhub input running and sending events to the `events` data stream. In the `events` data stream, the ingest pipeline performs these tasks:

- discover and set the `event.dataset` field using the `category` field in the event.
- use the `event.dataset` field to reroute the event to the target data stream.

The discover process uses the following logic:

- if the event doesn't have a category, it sets `event.dataset` to `azure.eventhub` (the generic integration)
- if the event does have a category, it sets `event.dataset` to `azure.platformlogs` (it's probably an Azure log)
- if the event category is supported, it sets `event.dataset` to specific one like `azure.activitylogs` or `azure.signinlogs`.

After the discovery step, the routing rules use the `event.dataset` value to forward the events to the best available target data stream.

## Checklist

## Author's Checklist

- [ ] 

- Bump the integration version
- Build and start a local stack
- Install the Azure Logs package (turn on v2 and turn off all v1s)
- Send test documents and check they are redirected to the data stream

### Bump the integration version

I'll update the version and changelog later to avoid conflicts. In the meantime, a simple bump would do:

```diff
diff --git a/packages/azure/changelog.yml b/packages/azure/changelog.yml
index bd20d10043..6da52b0424 100644
--- a/packages/azure/changelog.yml
+++ b/packages/azure/changelog.yml
@@ -1,3 +1,8 @@
+- version: "1.20.0+build18"
+  changes:
+    - description: Add Azure Logs integration v2 (preview)
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/11432
 - version: "1.19.4"
   changes:
     - description: Fix destination.geo.region_name mapping.
diff --git a/packages/azure/manifest.yml b/packages/azure/manifest.yml
index be7dccf294..895d9a8b6e 100644
--- a/packages/azure/manifest.yml
+++ b/packages/azure/manifest.yml
@@ -1,6 +1,6 @@
 name: azure
 title: Azure Logs
-version: 1.19.4
+version: "1.20.0+build18"
 description: This Elastic integration collects logs from Azure
 type: integration
 icons:
```

### Build and start a local stack

```sh
elastic-package build && elastic-package stack up -d -v --version 8.16.1
```

### Install the Azure Logs package

- Turn on the v2 preview integration
- Turn off all the v1 integrations

### Send test documents and check they are redirected to the data stream

I use the [eventhubs](https://github.com/zmoog/eventhubs) CLI tool to send JSON documents to an event hub.

Set up some environment variables:

```sh
export EVENTHUB_CONNECTION_STRING="[redacted]"
export EVENTHUB_NAMESPACE="[redacted]"
export EVENTHUB_NAME="logs"
```

We can then `cat` the test document and pipe them into an event hub using `eh -v eventdata send-batch`:

```sh
cat data_stream/provisioning/_dev/test/pipeline/test-provisioninglogs-raw.log | eh -v eventdata send-batch 
```

All supported documents should land into the expected data stream. Any unsupported document should land into the platform logs data stream.

## Related issues

- Closes https://github.com/elastic/integrations/issues/11251

### Integration setup

CleanShot 2024-12-03 at 15 46 55@2x

### Sample from the integration docs

CleanShot 2024-12-03 at 23 25 35@2x

### DIFF
--- a/packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
@@ -76,8 +76,7 @@ processors:
   - set:
       field: event.dataset
       value: azure.springcloudlogs
-      # TODO: search more categories
-      if: 'ctx.tmp_json?.category == "ApplicationConsole"'
+      if: 'ctx.tmp_json?.category == "ApplicationConsole" || ctx.tmp_json?.category == "SystemLogs" || ctx.tmp_json?.category == "IngressLogs" || ctx.tmp_json?.category == "BuildLogs" || ctx.tmp_json?.category == "ContainerEventLogs"'
   
   # Remove the temporary field used to identify the event dataset.
   - remove:
--- a/packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
@@ -16,8 +16,12 @@ processors:
       value: event
   
   #
-  # Set the event dataset based on the event category.
-  # --------------------------------------------------
+  # Set `event.dataset` value based on the event category.
+  # ------------------------------------------------------
+  #
+  # In the `routing_rules.yml` file, we use the
+  # `event.dataset` value to route the event to
+  # the appropriate data stream.
   #
   
   - json:
@@ -25,12 +29,13 @@ processors:
       target_field: tmp_json
       description: 'Parses the message field as JSON and stores it in a temporary field to identify the event dataset.'
   
-  # Default to azure.eventhub if the category is not recognized.
+  # Defaults to azure.events if the `category` field is not present.
   - set:
       field: event.dataset
-      value: azure.eventhub
+      value: azure.events
       description: 'Sets the default event dataset.'
   
+  # Sets the event dataset based on the `category` field.
   - set:
       field: event.dataset
       value: azure.platformlogs
@@ -80,54 +85,7 @@ processors:
       ignore_missing: true
       description: 'Removes the temporary field used to identify the event dataset.'
 
-#   #
-#   # Route the event to the appropriate ingest pipeline based on the event dataset.
-#   # ------------------------------------------------------------------------------
-#   #
-#   - pipeline:
-#       name: '{{ IngestPipeline "pipeline_eventhub" }}'
-#       if: ctx.event?.dataset == 'azure.eventhub'
-#       ignore_missing_pipeline: true
-#   - pipeline:
-#       name: '{{ IngestPipeline "pipeline_activitylogs" }}'
-#       if: ctx.event?.dataset == 'azure.activitylogs'
-#       ignore_missing_pipeline: true
-#   - pipeline:
-#       name: '{{ IngestPipeline "pipeline_application_gateway" }}'
-#       if: ctx.event?.dataset == 'azure.application_gateway'
-#   - pipeline:
-#       name: '{{ IngestPipeline "pipeline_auditlogs" }}'
-#       if: ctx.event?.dataset == 'azure.auditlogs'
-#       ignore_missing_pipeline: true
-#   - pipeline:
-#       name: '{{ IngestPipeline "pipeline_firewall_logs" }}'
-#       if: ctx.event?.dataset == 'azure.firewall_logs'
-#       ignore_missing_pipeline: true
-#   - pipeline:
-#       name: '{{ IngestPipeline "pipeline_graphactivitylogss" }}'
-#       if: ctx.event?.dataset == 'azure.graphactivitylogss'
-#       ignore_missing_pipeline: true
-#   - pipeline:
-#       name: '{{ IngestPipeline "pipeline_identity_protection" }}'
-#       if: ctx.event?.dataset == 'azure.identity_protection'
-#       ignore_missing_pipeline: true
-#   - pipeline:
-#       name: '{{ IngestPipeline "pipeline_platformlogs" }}'
-#       if: ctx.event?.dataset == 'azure.platformlogs'
-#       ignore_missing_pipeline: true
-#   - pipeline:
-#       name: '{{ IngestPipeline "pipeline_provisioning" }}'
-#       if: ctx.event?.dataset == 'azure.provisioning'
-#       ignore_missing_pipeline: true
-#   - pipeline:
-#       name: '{{ IngestPipeline "pipeline_signinlogs" }}'
-#       if: ctx.event?.dataset == 'azure.signinlogs'
-#       ignore_missing_pipeline: true
-#   - pipeline:
-#       name: '{{ IngestPipeline "pipeline_springcloudlogs" }}'
-#       if: ctx.event?.dataset == 'azure.springcloudlogs'
-#       ignore_missing_pipeline: true
-
+#
 #
 # Error handling
 #
--- a/packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
@@ -77,6 +77,7 @@ processors:
       field: event.dataset
       value: azure.springcloudlogs
       if: 'ctx.tmp_json?.category == "ApplicationConsole" || ctx.tmp_json?.category == "SystemLogs" || ctx.tmp_json?.category == "IngressLogs" || ctx.tmp_json?.category == "BuildLogs" || ctx.tmp_json?.category == "ContainerEventLogs"'
+      description: 'Azure Spring Apps log categories (refs: https://learn.microsoft.com/en-us/azure/azure-monitor/reference/supported-logs/microsoft-appplatform-spring-logs)'
   
   # Remove the temporary field used to identify the event dataset.
   - remove:
--- a/packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
@@ -1,0 +1,138 @@
+---
+description: Pipeline for parsing Azure logs.
+processors:
+  - set:
+      field: ecs.version
+      value: '8.11.0'
+  # TODO: we can remove this processor when https://github.com/elastic/beats/issues/40561
+  # is fixed and released.
+  - rename:
+      field: azure
+      target_field: azure-eventhub
+      if: 'ctx.azure?.eventhub != null'
+      ignore_missing: true
+  - set:
+      field: event.kind
+      value: event
+  
+  #
+  # Set the event dataset based on the event category.
+  # --------------------------------------------------
+  #
+  
+  - json:
+      field: message
+      target_field: tmp_json
+      description: 'Parses the message field as JSON and stores it in a temporary field to identify the event dataset.'
+  
+  # Default to azure.eventhub if the category is not recognized.
+  - set:
+      field: event.dataset
+      value: azure.eventhub
+      description: 'Sets the default event dataset.'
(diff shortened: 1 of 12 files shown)


zmoog:
As the first step, this PR introduces a v2 integration that can handle all the Azure Logs using one input only.

- We mark the new integration as "v2" while all the existing integrations are marked as "v1". 
- We renamed the generic event hub integration "Collect raw events (v1)"  and recommend using the [Custom Azure Logs integration](https://www.elastic.co/docs/current/integrations/azure_logs) for more flexibility.

CleanShot 2024-12-03 at 15 46 55@2x

packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,96 @@
...
+  # Sets the event dataset based on the `category` field.
+  - set:
+      field: event.dataset
+      value: azure.platformlogs
+      if: 'ctx.tmp_json?.category != null'
+      description: 'If the event has a category field, we consider it a platform log.'
+  - set:
+      field: event.dataset
+      value: azure.activitylogs
+      if: 'ctx.tmp_json?.category == "Administrative" || ctx.tmp_json?.category == "Security" || ctx.tmp_json?.category == "ServiceHealth" || ctx.tmp_json?.category == "Alert" || ctx.tmp_json?.category == "Recommendation" || ctx.tmp_json?.category == "Policy" || ctx.tmp_json?.category == "Autoscale" || ctx.tmp_json?.category == "ResourceHealth"'
```
gizas, line 47: I guess those values come from the [schema](https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/activity-log-schema#categories) , so wondering how prone are those checks to any update or to  wrong capitals/lowercase of values. Do you think first apply a lowercase to ctx.tmp_json?.category and then check for string match makes sense?

packages/azure/docs/events.md
```diff
@@ -0,0 +1,504 @@
...
+│ <<event hub>> │──▶│    Agent    │─▶│ <<data stream>> │──┼─▶│ <<data stream>> │
+└───────────────┘   └─────────────┘  └─────────────────┘  │  └─────────────────┘
+                                                          │                     
+                                                          │  ┌─────────────────┐
+                                                          │  │   signin logs   │
+                                                          └─▶│ <<data stream>> │
+                                                             └─────────────────┘
+```
+
+The integration will automatically detect the log category and forward the logs to the appropriate data stream. When the integration v2 preview cannot find a matching data stream for a log category, it forward the logs to the platform logs data stream.
```
alaudazzi, line 32:
```suggestion
The integration will automatically detect the log category and forward the logs to the appropriate data stream. When the integration v2 preview cannot find a matching data stream for a log category, it forwards the logs to the platform logs data stream.
```

packages/azure/docs/events.md
```diff
@@ -0,0 +1,504 @@
...
+
+To use the v2 preview, you must turn off all the existing v1 integrations and turn on only the v2 preview integration.
```
alaudazzi, line 34:
```suggestion
IMPORTANT: To use the v2 preview, you must turn off all the existing v1 integrations and turn on only the v2 preview integration.
```
I would emphasize this with an admonition.

packages/azure/docs/events.md
```diff
@@ -0,0 +1,504 @@
...
+
+Under the hood, the v2 preview uses only one `azure-eventhub` input per event hub. The v2 preview avoids the contention and inefficiencies from using multiple inputs with the same event hub that happens the v1 architecture. With the v2 preview, you can still assign the agent policy to multiple Elastic Agents to scale out the logs processing.
```
alaudazzi, line 36:
```suggestion
Under the hood, the v2 preview uses only one `azure-eventhub` input per event hub. The v2 preview avoids contention and inefficiencies from using multiple inputs with the same event hub, which is typical of the v1 architecture. With the v2 preview, you can still assign the agent policy to multiple Elastic Agents to scale out the logs processing.
```

alaudazzi (approved):
I reviewed the packages/azure/docs/events.md file and left a few minor editing suggestions.

packages/azure/data_stream/events/manifest.yml
```diff
@@ -0,0 +1,93 @@
...
+      - Microsoft Graph Activity Logs
+      - Spring Apps logs
+      - Firewall logs
+      - Application Gateway logs
+
+      **You only need to enable this v2 integration to collect and route all the supported Azure logs**. You can turn off the equivalent v1 integrations.
+
+      If you need to collect raw events from Azure Event Hub, we recommend using the [Custom Azure Logs integration](https://www.elastic.co/docs/current/integrations/azure_logs) which provides more flexibility.
+
+      To learn more about the efficiency and routing enhancements introduced in version 1.20.0, please read the [Azure Logs (v2 preview)](https://www.elastic.co/docs/current/integrations/azure/events) documentation.
```
gizas, line 31: This is the events.md file that will be rendered correct? Just the word efficiency made me read again the evetns.md and was trying to find a comparison between v1 vs v2.
Maybe just general, `To learn more about the enhancements `

packages/azure/data_stream/events/routing_rules.yml
```diff
@@ -0,0 +1,57 @@
...
+        - default
+    - target_dataset: azure.platformlogs
+      if: ctx.event?.dataset == 'azure.platformlogs'
+      namespace:
+        - "{{data_stream.namespace}}"
+        - default
+    - target_dataset: azure.provisioning
+      if: ctx.event?.dataset == 'azure.provisioning'
+      namespace:
+        - "{{data_stream.namespace}}"
```
gizas, line 46: Does this mean that you want to have also the additonal namespaces defined in the policy, along with default?

Should not then have a field namespace in manifest like [this](https://github.com/elastic/integrations/blob/main/packages/kubernetes/manifest.yml#L49)?

packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
gizas: I think with the default values you apply we are safe here in scenarios of udpates of categories. You can neglect this comment above

gizas (approved):

packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
zmoog: From my experience and some research it seems Azure use the category names pretty consistently. I would try to stay with the documented names and adjust if necessary.

packages/azure/data_stream/events/manifest.yml
zmoog: > This is the events.md file that will be rendered correct? 

Yes, the `events.md` file will eventually be available at https://www.elastic.co/guide/en/integrations/current/azure-events.html (now it's a 404). 

> Maybe just general, "To learn more about the enhancements"

You mean adding a link to the GitHub issues with more details, or add a sections that explains it? Should we extend the "What's new in v2 preview?" section?

muthu-mps:
@zmoog - What if a user enables V2 and did not disable the V1? Do you think whats new added in the config description can go in the document and we say more explicitly to disable V1 while enabling V2 on top of the config description. WDYT?

packages/azure/data_stream/events/routing_rules.yml
zmoog: Uhm, good question. I don't know how it works under the hood, but the integration seems to manage `data_stream.*` its own.

zmoog:
> Probably you can delete such reference from dashboards if not used

Yeah, this is something we need to address. I'll open an issue if I can't fix them quickly in this PR.

zmoog:
> What if a user enables V2 and did not disable the V1? 

They may get duplicate events and increased contention among consumers.

> Do you think `whats new` added in the config description can go in the document and we say more explicitly to disable V1 while enabling V2 on top of the config description. WDYT?

Yeah, I agree. While reading the text today, I noticed the text is not clear and we should be more explicit. Let me share the updated version...

zmoog:
> Do you think `whats new` added in the config description can go in the document and we say more explicitly to disable V1 while enabling V2 on top of the config description. WDYT?

What about this version?

CleanShot 2024-12-06 at 12 46 10@2x

zmoog:
> Some minors:
> 
> * format_version: "3.0.2" --> to 3.3.0 probably and retest?

Oh, great point, I missed it! 🤦 — thank you for the heads up, I'm updating the package-spec version and I'll re-run the tests.

> * conditions.kibana.version: "^8.13.0" ---> Not sure when routing_rules introduced, is that accurate?

Stack version 8.13.0 is okay:

- Elasticsearch introduced the [reroute processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/reroute-processor.html) in release 8.8.0.
- package-spec [added](https://github.com/elastic/package-spec/pull/535) support for the routing rules in version v2.9.0 

We finally found the time to update the integration with these long awaited changes.

packages/azure/data_stream/events/fields/fields.yml
```diff
@@ -0,0 +1,27 @@
+- name: azure-eventhub
+  type: group
+  fields:
+    - name: eventhub
+      type: keyword
+      description: |
+        Event hub name.
```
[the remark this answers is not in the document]
zmoog: Removed!

packages/azure/data_stream/events/manifest.yml
gizas: Both can work. I would add some more details why v2 is better (if you have not done it already in last updates)

packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,96 @@
...
+      value: azure.provisioning
+      if: 'ctx.tmp_json?.category == "ProvisioningLogs"'
+  - set:
+      field: event.dataset
+      value: azure.signinlogs
+      if: 'ctx.tmp_json?.category == "SignInLogs" || ctx.tmp_json?.category == "NonInteractiveUserSignInLogs" || ctx.tmp_json?.category == "ServicePrincipalSignInLogs" || ctx.tmp_json?.category == "ManagedIdentitySignInLogs"'
+  - set:
+      field: event.dataset
+      value: azure.springcloudlogs
+      # TODO: search more categories
```
muthu-mps, line 79: Does this mean that we will support `ApplicatonConsole` for `springcloudlogs` and add more categories later?

zmoog: I was mostly a reminder to double-check if there are additional log categories to include in the list. 

Microsoft [announced](https://learn.microsoft.com/en-gb/azure/spring-apps/basic-standard/retirement-announcement) the Azure Spring Apps retirement earlier this year. However, we need to support existing deployments.

The integration supports `ApplicationConsole` for sure because we have tests in the `_dev/test` folder:

```sh
$ cat data_stream/springcloudlogs/_dev/test/pipeline/*.log | jq .category  
"ApplicationConsole"
"ApplicationConsole"
"ApplicationConsole"
```

But personally I never tested it end to end. 

@muthu-mps, can I also include the other log categories?

```text
SystemLogs
IngressLogs
BuildLogs
ContainerEventLogs
```

muthu-mps: Yes, we can include the other categories as well @zmoog.

muthu-mps (approved):
LGTM!

PR #11990, opened 2024-12-03
[citrix_adc] Ensure time zone setting is applied properly

taylor-swanson:
## Proposed commit message

- Reworked how time zones are applied to ensure time zones values are applied in the proper order.
- The order is: from log, from config, from add_locale, and fallback to UTC.
- If tz_offset is null/empty string, the `add_locale` processor is not run.
- Add test cases

## Checklist

```
cd packages/citrix_adc
elastic-package test
```

### DIFF
--- a/packages/citrix_adc/data_stream/log/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/citrix_adc/data_stream/log/elasticsearch/ingest_pipeline/default.yml
@@ -59,7 +59,7 @@ processors:
       field: _tmp.tz
       copy_from: _conf.tz_offset
       override: false
-      if: ctx._conf?.tz_offset != null && ctx._conf.tz_offset != 'local'
+      ignore_empty_value: true
   - set:
       field: _tmp.tz
       copy_from: event.timezone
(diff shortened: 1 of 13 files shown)


packages/citrix_adc/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -49,18 +49,29 @@ processors:
       type: long
       ignore_missing: true
 
-# Time zone
+# Time zone (in order: log, config, locale, default to UTC).
 
   - set:
-      field: _conf.tz_offset
-      tag: set_tz_offset_utc
+      field: _tmp.tz
```
jrmolin, line 55: This is exactly what I was confused about! I'm glad you picked this up, @taylor-swanson !!

jrmolin (approved):
I didn't understand how the timezone information was being processed in this one. I'm glad you're on the case!

packages/citrix_adc/data_stream/log/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: I ended up reusing some logic from the Cisco ASA integration. Using the temporary variable allows us to apply the precedence a bit easier.

PR #11995, opened 2024-12-04
Revert: prisma_cloud: fix login endpoint path (#11885)

efd6:
## Proposed commit message

Change was incorrect[1]. Also improve documentation in the config UI.

[1]https://pan.dev/prisma-cloud/api/cwpp/post-authenticate/

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Ref #11885

### DIFF
--- a/packages/prisma_cloud/changelog.yml
+++ b/packages/prisma_cloud/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.4.2"
+  changes:
+    - description: 'Revert: Fix path to API login for host data sources.'
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/11995
 - version: "1.4.1"
   changes:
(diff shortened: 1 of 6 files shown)


chrisberkhout (approved):
Does revert as it says, so I assume it's good to go.

Looking at the documentation, I see a URL format like this:
```
https://<CONSOLE>/api/v<VERSION>/authenticate
```
So it looks like the `url` variable should be set to `https://<CONSOLE>/api/v<VERSION>`. In the manifest files, `url` is described as:

> Base URL of the Prisma Cloud Server API.

Maybe an example value should be added there to avoid setting `https://<CONSOLE>` by mistake.

efd6:
Waiting for input from @mohitjha-elastic

mohitjha-elastic:
The changes look good to me, and including an example format in the base URL makes it clearer for the user.

Issue #11673, opened 2024-11-08
[meta] Qualys VMDR: Enhancement to leverage cloud workflows


kcreddy:
As part of effort to support Cloud Security features with the data from these 3rd party integrations, the vulnerabilities findings data from Qualys VMDR should be enriched.

The vulnerability data resides in 2 datastreams in Qualys VMDR integration.

1. [Asset Host Detection](https://qualysguard.qg2.apps.qualys.com/qwebhelp/fo_portal/api_doc/assets/index.htm#t=host_lists%2Fhost_detection.htm): Provides some vulnerability information for hosts in users’s account.
2. [Knowledge Base](https://qualysguard.qg2.apps.qualys.com/qwebhelp/fo_portal/api_doc/scans/index.htm#t=kbase%2Fknowledgebase.htm): Provides detailed vulnerability data from the Qualys KnowledgeBase. Doesn’t associate with any user’s resources.

We will need both datastreams and use it to enrich vulnerability findings workflow, as Asset Host Detection is missing CVE and description which would be important to stitch in the vulnerability findings coming from Qualys. The integration should be performing the join as it queries the data. For example, as it queries the assets, it could fetch the related knowledge base entries related by the QID field.

Tasks:
- [x] Create Qualys user with access to asset host detections and knowledge base: https://github.com/elastic/infosec/issues/19399
- [x] Setup Qualys Asset(s) to use with the asset host detections API.
- [x] Setup `Asset Host Detection` data in demo environment.
- [x] Analyse and get clarifications on mapping changes with Cloud team.
   - [x] Fix any existing bugs: 
      - [x] https://github.com/elastic/integrations/pull/11877
      - [x] https://github.com/elastic/integrations/pull/12006
      - [x] https://github.com/elastic/integrations/pull/12009
   - [x] Cleanup https://github.com/elastic/integrations/pull/11927
- [x] Implement mapping changes: https://github.com/elastic/integrations/pull/11999
- [x] Implement transform: https://github.com/elastic/integrations/issues/12718

kcreddy:
**Update:**
With @clement-fouque's assistance, the asset is now available with vulnerability data. Also got access to the relevant API for fetching the asset vulnerability data.

kcreddy:
> We will need both datastreams and use it to enrich vulnerability findings workflow, as Asset Host Detection is missing CVE and description which would be important to stitch in the vulnerability findings coming from Qualys. The integration should be performing the join as it queries the data

[...]

PR #11999, opened 2024-12-04
qualys_vmdr: Improve support for Cloud Detection and Response (CDR) workflow

kcreddy:
## Proposed commit message
Ref: https://github.com/elastic/integrations/issues/11673

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/qualys_vmdr/changelog.yml
+++ b/packages/qualys_vmdr/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "5.4.0"
+  changes:
+    - description: Leverage Cloud Detection and Response (CDR) workflows in asset_host_detection data-stream.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
 - version: "5.3.0"
   changes:
     - description: Document required user role permissions to each API.
--- a/packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
+++ b/packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
@@ -25,48 +25,118 @@ redact:
   fields:
     - password
 program: |
-  state.with(
-    request("GET",
-      state.?want_more.orValue(false) ?
-        state.pagination_url
-      :
-        state.url.trim_right("/") + "/api/2.0/fo/asset/host/vm/detection/?" +
-        state.?params.orValue("").parse_query().with({
-          "action": ["list"],
-          "truncation_limit": [string(state.batch_size)],
-        }).format_query()
-    ).with({
-        "Header":{
+  (
+    has(state.worklist) && size(state.worklist) > 0 ?
+      state
+    :
+      state.with(
+        request("GET",
+          state.?want_more.orValue(false) ?
+            state.next_page.link
+          :
+            state.url.trim_right("/") + "/api/2.0/fo/asset/host/vm/detection/?" +
+            state.?params.orValue("").parse_query().with({
+              "action": ["list"],
+              "truncation_limit": [string(state.batch_size)],
+            }).format_query()
+        ).with({
+            "Header":{
+                "X-Requested-With": ["curl"],
+                "Authorization": ["Basic "+(state.user+":"+state.password).base64()],
+            }
+        }).do_request().as(resp, resp.StatusCode == 200 ?
+            resp.Body.as(xml, bytes(xml).decode_xml('qualys_api_2_0_ahd').as(ahd_body, {
+              "worklist": (
+                ahd_body.doc.HOST_LIST_VM_DETECTION_OUTPUT.RESPONSE.HOST_LIST.HOST.map(h,
+                  h.DETECTION_LIST.DETECTION.map(v,
+                    h.with({
+                      "DETECTION_LIST": v,
+                    })
+                  )).flatten()
+              ),
+              "next_page": { 
+                ?"link": ahd_body.?doc.HOST_LIST_VM_DETECTION_OUTPUT.RESPONSE.WARNING.URL
+              },
+              "next": 0,
+            }))
+          :
+            {
+              "events": {
+                "error": {
+                  "code": string(resp.StatusCode),
+                  "id": string(resp.Status),
+                  "message": "GET:"+(
+                    size(resp.Body) != 0 ?
+                      string(resp.Body)
+                    :
+                      string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
+                  ),
+                },
+              },
+              "want_more": false,
+            }
+        )
+      )
+  ).as(state, state.with(
+    !has(state.worklist) ? state : // Exit early due to GET failure.
+      state.next < size(state.worklist) ?
+        request("GET", state.url.trim_right("/") + "/api/2.0/fo/knowledge_base/vuln/?" +
+          state.?params.orValue("").parse_query().with({
+            "ids": [string(state.worklist[state.next].DETECTION_LIST.QID)],
+            "action": ["list"],
+          }).format_query()
+        ).with({
+          "Header":{
             "X-Requested-With": ["curl"],
             "Authorization": ["Basic "+(state.user+":"+state.password).base64()],
-        }
-    }).do_request().as(resp,
-      resp.StatusCode == 200
-      ?
-        resp.Body.as(xml, bytes(xml).decode_xml('qualys_api_2_0').as(body, {
-            "events": body.doc.HOST_LIST_VM_DETECTION_OUTPUT.RESPONSE.HOST_LIST.HOST.map(h,
-              h.DETECTION_LIST.DETECTION.map(v, {
-                "message": h.with({"DETECTION_LIST": v}).encode_json(),
-              })
-            ).flatten(),
-            ?"pagination_url": body.?doc.HOST_LIST_VM_DETECTION_OUTPUT.RESPONSE.WARNING.URL,
-            "want_more": body.?doc.HOST_LIST_VM_DETECTION_OUTPUT.RESPONSE.WARNING.URL.hasValue(),
-        }))
+          }
+        }).do_request().as(resp, resp.StatusCode == 200 ?
+          resp.Body.as(xml, bytes(xml).decode_xml('qualys_api_2_0_kb').as(kb_body, {
+            "events": (
+              has(kb_body.doc.KNOWLEDGE_BASE_VULN_LIST_OUTPUT.RESPONSE.VULN_LIST)
+              ?
+                kb_body.doc.KNOWLEDGE_BASE_VULN_LIST_OUTPUT.RESPONSE.VULN_LIST.VULN.map(e, e.with({
+                    "CVE_LIST": e.?CVE_LIST.CVE.orValue([]).map(c, c.ID),
+                })).as(kb, {
+                  "message": state.worklist[state.next].with({
+                    "knowledge_base": kb,
+                  }).encode_json()
+                })
+              :
+                {
+                  "message": state.worklist[state.next].with({
+                    "knowledge_base": {
+                      "response_error": "response IDs values are missing: " + string(kb_body.?doc.KNOWLEDGE_BASE_VULN_LIST_OUTPUT.RESPONSE.ID_SET.orValue([])),
+                    }
+                  }).encode_json()
+                }
+            ),
+            "worklist": int(state.next) + 1 < size(state.worklist) ? state.worklist : [],
+            "next": int(state.next) + 1 < size(state.worklist) ? int(state.next) + 1 : 0,
+            "want_more": int(state.next) + 1 < size(state.worklist) || has(state.?next_page.link),
+          }))
+        :
+          {
+            "events": {
+              "error": {
+                "code": string(resp.StatusCode),
+                "id": string(resp.Status),
+                "message": "GET:"+(
+                  size(resp.Body) != 0 ?
+                    string(resp.Body)
+                  :
+                    string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
+                ),
+              },
+            },
+            "want_more": false,
+          }
+        )
       :
         {
-          "events": {
-            "error": {
-              "code": string(resp.StatusCode),
-              "id": string(resp.Status),
-              "message": "GET:"+(
-                size(resp.Body) != 0 ?
-                  string(resp.Body)
-                :
-                  string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
-              ),
-            },
-          },
+          "events": [],
           "want_more": false,
+          "next_page": {},
         }
     )
   )
@@ -91,7 +161,7 @@ processors:
 {{processors}}
 {{/if}}
 xsd:
-  qualys_api_2_0: |
+  qualys_api_2_0_ahd: |
     <xsd:schema xmlns:xsd='http://www.w3.org/2001/XMLSchema'>
 
       <xsd:element name='HOST_LIST_VM_DETECTION_OUTPUT'>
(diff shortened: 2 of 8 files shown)


packages/qualys_vmdr/elasticsearch/transform/latest_cdr_vulnerabilities/transform.yml
```diff
@@ -0,0 +1,37 @@
...
+dest:
+  index: "security_solution-qualys_vmdr.vulnerability_latest-v1"
+  aliases:
+    - alias: "security_solution-qualys_vmdr.vulnerability_latest"
+      move_on_creation: true
+latest:
+  unique_key:
+    - resource.id
+    - event.id
+    - data_stream.namespace
```
kcreddy, line 20: This list of `unique_key` fields is different from the proposed guide due to the issue described in [here](https://github.com/elastic/security-team/issues/9987#issuecomment-2530636202) and applying suggestion from [here](https://github.com/elastic/security-team/issues/9987#issuecomment-2531035582)

packages/qualys_vmdr/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "6.0.0"
+  changes:
+    - description: Leverage Cloud Detection and Response (CDR) workflows in asset_host_detection data-stream.
```
efd6, line 4:
```suggestion
    - description: Use Cloud Detection and Response (CDR) workflows in asset_host_detection data-stream.
```
?

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1066,6 +1142,70 @@ processors:
       tag: rename_qualys_vmdr_asset_host_detection_vulnerability_RESULTS_1
       target_field: qualys_vmdr.asset_host_detection.vulnerability.results
       ignore_missing: true
+  # The below 2 gsub processors facilitate `splitOnToken` in subsequent script processor which otherwise cannot split \n or \t.
```
efd6, line 1145:
```suggestion
  # The next two gsub processors facilitate `splitOnToken` in the subsequent script processor, which otherwise cannot split \n or \t.
```
Do we know that `;` and `|` do not otherwise appear in the text?

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/pipeline_knowledge_base.yml
```diff
@@ -0,0 +1,873 @@
...
+      source: >-
+        if (!(ctx.qualys_vmdr?.asset_host_detection?.knowledge_base?.SEVERITY_LEVEL instanceof String)) {
+          return;
+        }
+        def vuln_type = ctx.qualys_vmdr?.asset_host_detection?.knowledge_base?.vuln_type;
+        if (!(vuln_type instanceof String)) {
+          return;
+        }
+        def level = Long.parseLong(ctx.qualys_vmdr.asset_host_detection.knowledge_base.SEVERITY_LEVEL);
+        if (['Potential Vulnerability', 'Vulnerability', 'Vulnerability or Potential Vulnerability'].contains(vuln_type)) {
```
efd6, line 790: Could this array be put in a `params` array to avoid allocs?

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/pipeline_knowledge_base.yml
```diff
@@ -0,0 +1,873 @@
...
+              ctx.qualys_vmdr.asset_host_detection.knowledge_base.SEVERITY_LEVEL = "Minimal";
+          } else if (level == 2) {
+              ctx.qualys_vmdr.asset_host_detection.knowledge_base.SEVERITY_LEVEL = "Medium";
+          } else if (level == 3) {
+              ctx.qualys_vmdr.asset_host_detection.knowledge_base.SEVERITY_LEVEL = "Serious";
+          } else if (level == 4) {
+              ctx.qualys_vmdr.asset_host_detection.knowledge_base.SEVERITY_LEVEL = "Critical";
+          } else if (level == 5) {
+              ctx.qualys_vmdr.asset_host_detection.knowledge_base.SEVERITY_LEVEL = "Urgent";
+          }
```
efd6, line 801: These are dense with a numeric index, so the assignment could be done from a `params` array after bounds checking. Same below.

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml
efd6, line 1: Add a
```
  - append:
      field: tags
      allow_duplicates: false
```
to the final `on_failure`.

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
```diff
@@ -25,43 +25,145 @@ redact:
...
+              "action": ["list"],
+              "truncation_limit": [string(state.batch_size)],
+            }).format_query()
+        ).with({
+            "Header":{
+                "X-Requested-With": ["curl"],
+                "Authorization": ["Basic "+(state.user+":"+state.password).base64()],
+            }
+        }).do_request().as(resp, resp.StatusCode == 200 ?
+            resp.Body.as(xml, try(bytes(xml).decode_xml('qualys_api_2_0_ahd'), "decode_xml_error_ahd").as(ahd_body,
```
efd6, line 48: Since we are now on v8.16.0 this can be
```suggestion
            resp.Body.as(xml, try(xml.decode_xml('qualys_api_2_0_ahd'), "decode_xml_error_ahd").as(ahd_body, 
```

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
```diff
@@ -25,43 +25,145 @@ redact:
...
-    }).do_request().as(resp,
-      resp.StatusCode == 200
-      ?
-        resp.Body.as(xml, try(bytes(xml).decode_xml('qualys_api_2_0'), "decode_xml_error").as(body, 
-          !has(body.decode_xml_error) 
-          ?
-            (body.?doc.HOST_LIST_VM_DETECTION_OUTPUT.RESPONSE.HOST_LIST.HOST.hasValue() 
+          }
+        }).do_request().as(resp, resp.StatusCode == 200 ?
+          resp.Body.as(xml, try(bytes(xml).decode_xml('qualys_api_2_0_kb'), "decode_xml_error_kb").as(kb_body,
```
efd6, line 129:
```suggestion
          resp.Body.as(xml, try(xml.decode_xml('qualys_api_2_0_kb'), "decode_xml_error_kb").as(kb_body, 
```

packages/qualys_vmdr/elasticsearch/transform/latest_cdr_vulnerabilities/fields/vulnerability.yml
```diff
@@ -0,0 +1,14 @@
+- name: vulnerability
```
maxcold, line 1: we need to also map the fields that are a part of `ecs@mappings`, for example, `vulnerability.severity`. As we use "security_solution-*.vulnerability_latest*" index pattern , by default it doesn't get `ecs@mappings` in contrast to `logs-*` data streams
Right now the findings page doesn't show qualys findings as sort by `vulnerability.severity` fails due to the missing mapping

maxcold: simples would be probably to check what fields are mapped in `wiz vulnerability` data stream

kcreddy: Makes sense. Will do that.

packages/qualys_vmdr/changelog.yml
kcreddy: Updated in https://github.com/elastic/integrations/pull/11999/commits/7d316538c752e8e035f59abb592f78a1e76de14f

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml
kcreddy: I checked most values and couldn't find them, hence I chose these delimiters. However to be safe, I changed it to 2 delimiters in  https://github.com/elastic/integrations/pull/11999/commits/7d316538c752e8e035f59abb592f78a1e76de14f.

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/pipeline_knowledge_base.yml
kcreddy: Updated to array in https://github.com/elastic/integrations/pull/11999/commits/7d316538c752e8e035f59abb592f78a1e76de14f

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/pipeline_knowledge_base.yml
kcreddy: Updated the logic using `params` in https://github.com/elastic/integrations/pull/11999/commits/7d316538c752e8e035f59abb592f78a1e76de14f

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml
kcreddy: Added in https://github.com/elastic/integrations/pull/11999/commits/7d316538c752e8e035f59abb592f78a1e76de14f

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
kcreddy: Updated as suggested in https://github.com/elastic/integrations/pull/11999/commits/7d316538c752e8e035f59abb592f78a1e76de14f

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
kcreddy: Updated as suggested in https://github.com/elastic/integrations/pull/11999/commits/7d316538c752e8e035f59abb592f78a1e76de14f

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/pipeline_knowledge_base.yml
```diff
@@ -865,6 +860,10 @@ processors:
       value: pipeline_error
       if: ctx.error?.message != null
 on_failure:
+  - append:
+      field: tags
+      value: preserve_original_event
+      allow_duplicates: false
```
efd6, line 866: Also above the `on_failure`, conditional on `ctx.error?.message != null`. It should also be happening in the default.yml.

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/pipeline_knowledge_base.yml
```diff
@@ -786,27 +799,9 @@ processors:
...
-          if (level == 1) {
-              ctx.qualys_vmdr.asset_host_detection.knowledge_base.SEVERITY_LEVEL = "Minimal";
-          } else if (level == 2) {
-              ctx.qualys_vmdr.asset_host_detection.knowledge_base.SEVERITY_LEVEL = "Medium";
-          } else if (level == 3) {
-              ctx.qualys_vmdr.asset_host_detection.knowledge_base.SEVERITY_LEVEL = "Serious";
-          }
+        String level = ctx.qualys_vmdr.asset_host_detection.knowledge_base.SEVERITY_LEVEL;
+        if (params.vuln_types.contains(vuln_type)) {
+          ctx.qualys_vmdr.asset_host_detection.knowledge_base.SEVERITY_LEVEL = params.vuln_level.getOrDefault(level, params.vuln_level["0"]);
```
efd6, line 804: The logic here is not quite the same in the case of vuln type "information gathered". Is that OK?

kcreddy: @efd6, I also got confused here, but it seems to be the same. 
The case when `vuln_type` is one of `['Potential Vulnerability', 'Vulnerability', 'Vulnerability or Potential Vulnerability']` or is `Information Gathered`, the values getting assigned are the same. Not sure why there were 2 cases to begin with in the Knowledge base pipeline. 
Let me know if I am missing something obviously wrong.

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/pipeline_knowledge_base.yml
kcreddy: Addressed in https://github.com/elastic/integrations/pull/11999/commits/ecb914bc5b9f7e43f6c503114be0ccabe030faed

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/pipeline_knowledge_base.yml
efd6: The previous code looked like it was doing something special. If that special behaviour is not necessary, then this is fine (simpler is better).

packages/qualys_vmdr/elasticsearch/transform/latest_cdr_vulnerabilities/transform.yml
kcreddy: Not adding transform to this PR.

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
```diff
@@ -25,87 +25,187 @@ redact:
...
+            "want_more": false,
+          })
+      )
+    )
+  ).as(state, state.with(
+    !has(state.worklist) ? state : 
+      size(state.worklist) > 0 ?
+        request("GET", state.url.trim_right("/") + "/api/2.0/fo/knowledge_base/vuln/?" +
+          {
+            "ids": [string(state.worklist[0].DETECTION_LIST.QID)],
```
kcreddy, line 130: There is a significant ingestion delay taking this approach (joining knowledge base using QID). 
A comparison between `5.8.0` (current GA) and `6.0` (this PR):

Screenshot 2025-01-02 at 9 10 55 AM

`5.8.0` version has ingestion rate of ~20k records per minute, while `6.0` has ~20 records per minute.

@andrewkroh, any suggestions how to handle this?

kcreddy: Few options:

For each host, we can accumulate all QIDs and pass onto single knowledge base API call using the same `ids` parameter, which would considerably reduce the number of API calls. With this we can reduce the number of API calls to number of hosts. This is straight-forward to achieve.

An addition to above approach, further reducing number of API calls, we can get all unique QIDs from all current hosts `worklist` and call knowledge base API for these unique QIDs. Then join the data between the hosts `worklist` and these unique QIDs. This join could be tricky.

andrewkroh: I like your first idea. I think it would be worth implementing that change and seeing how that performs under your testing.

kcreddy: @andrewkroh, 
The first option i.e., make 1 API call per host (sending all QIDs per host) results in ~400 events/minute significant improvement over ~20 events/minute:
Screenshot 2025-01-21 at 4 14 37 PM

maxcold (approved):
looks good to me. Tested with the reverted transform, there are indeed places in kibana where we will need to adjust, like multiple CVEs/no CVEs or severity being lowercase (right now we rely on the UPPERCASE severity values but I think in the UI we should support any case and cast to lowercase ourselves) But as we discussed, the transform is not part of this PR and can be added when Kibana is ready to support all these differences in the data
one thing though I'm concerned about is the regression in the ingestion speed, smth that might affect even our own InfoSec. But I'm leaving this discussion to integration owners, as I don't have a better solution to get CVE data without joining with knowledge base

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
kcreddy: >  single digit millions of events every day

We noticed that Qualys integration could be ingesting duplicates based on the interval period that user sets i.e., we could be ingesting multiple events for the same Qualys scan, host, qid.
Adding a UI option to de-duplicate the data in this PR: https://github.com/elastic/integrations/pull/12428

But I think when the number of hosts are really high, even with the [improved approach of making 1 Knowledge Base API call per host](https://github.com/elastic/integrations/pull/11999#discussion_r1923494184), we will have challenge because Qualys hourly API limits are quite low: https://cdn2.qualys.com/docs/qualys-api-limits.pdf.

smriti0321:
@kcreddy After an internal sync with @maxcold, it is concerning to see this update has caused the regression in ingestion speed. Thanks for trying out options for increasing the ingestion speed, what is the latest on this, have we optimized to get back to pre-CDR standards for ingestion?
Based on the data shared by our infosec team, any impact on ingestion speed will impact basic SIEM use cases as well. We would not like to introduce CDR at cost of this. Lets discuss this and brainstorm the options together. cc @narph @clement-fouque

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
clement-fouque: I don't think we should deduplicate things as:
1. Gain seems minimal (we'll only gain on vulnerabilities that have changed or are new).
2. We'll not be able to build out of box trends

If the ingestion speed is higher than now, it'll make it difficult to use the integration.

kcreddy: This PR has 1 Knowledge Base API call per host per QID. Gives ~20 events/min.

Variation 1: https://github.com/elastic/integrations/commit/2a5a231e9b77b4686fb59f4ec5fb1d727446f92a has 1 Knowledge Base API per host. Gives ~400 events/min.

The other variation I am currently working on is to make 1(few) Knowledge base API call(s) for each batch of hosts [(default batch size: 1000)](https://github.com/elastic/integrations/blob/main/packages/qualys_vmdr/data_stream/asset_host_detection/manifest.yml#L29). I am hoping to mitigate several thousand Knowledge base API calls through this approach.

kcreddy: Quick update: 

During my tests with KB API, I noticed I cant send lot of QIDs to KB API in single call. Upto 1000 QIDs worked using `ids=` parameter. So, in below screenshot, I applied an asset host API parameter: `id_min=932401809`, which has ~1000 hosts and total QIDs in these hosts are ~100 (mostly repeated). I sent these ~100 QIDs to KB API using `ids=` parameter.

So, 1 asset host API call fetching ~1000 hosts followed by 1 KB API call fetching ~100 QIDs for these hosts, total 2 API calls. 

With this I got positive outcome. ~18k events/minute.
Screenshot 2025-01-30 at 4 50 49 PM

Commit: https://github.com/elastic/integrations/commit/dc12c149058ad4726c9e4be33b8d0650ea234ff2 (still WIP). 
The commit still doesn't take take into account cases where if QIDs > 1000 need to be sent. This still needs to be tested along cases where we reach API limits.

kcreddy: More testing on this variation 2: `1-2 KB API calls per batch of hosts`

https://github.com/kcreddy/integrations/compare/qualys-cdr...kcreddy:integrations:qualys-vmdr-var2

In this, I am making 1-2 API calls based on number of unique QIDs in that batch of hosts. Ideally, it would be nice to split these unique QIDs into batches of 1000 (as per my tests that KB API could allow). But CEL doesn't seem to support slicing of lists. So, right now I am splitting the QIDs into [2 lists/groups of even and odd](https://github.com/kcreddy/integrations/compare/qualys-cdr...kcreddy:integrations:qualys-vmdr-var2#diff-5dd9dd821f7c8c44b8330ae9c7a90b4085fc4f0bf979be89922691e7bda5d9ccR66-R81) if unique QIDs are > 1000. With each list/group, I make 1 KB API call. So, at most 2 KB API calls for a batch of hosts.

With this, I was able to ingest all our events inside ~1 hr 20 min. with ingestion rate of ~12k/minute.

Screenshot 2025-02-01 at 1 05 52 AM
Screenshot 2025-02-01 at 1 06 31 AM

We could also split QIDs up further into groups of 3, 4, 5.., but each group increases the KB API call by 1.

@andrewkroh, let me know if this sounds feasible to proceed further.
@clement-fouque, @maxcold let me know your thoughts as well.

clement-fouque: Another benefit from this approach where both Asset Host Detection and KB calls are made together is that we don't have vulnerabilities without missing information. 

Sometimes we have detection that don't have any information from the knowledge base as it didn't have time to ingest the knowledge base update and to regenerate the enrich index. It's solved automatically on the next Asset Host Detection. 

With this new approach, we'll no longer have this temporary issue.

andrewkroh: > 5.8.0 version has ingestion rate of ~20k

> With this, I was able to ingest all our events ... with ingestion rate of ~12k/minute

That sounds like a pretty good trade-off for fully enriched events (asset vulns + KB). And if we find a way to have CEL split the array of QIDs by length, then we can avoid unnecessarily making two KB API calls when one would suffice (I have asked about this in the `#cel` Slack channel.

kcreddy: Thanks for the suggestions. I will go ahead with this new approach, fix merge conflicts and request for approval.

kcreddy:
Fixed the PR again.

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1,6 +1,9 @@
 ---
 description: Pipeline for processing Asset Host Detection data.
 processors:
+  - drop:
```
andrewkroh, line 4: Could we drop this earlier? Like add a non-user-modifieable processor in the agent:

```yaml
- drop_event:
    when:
      equals:
        message: want_more
```

packages/qualys_vmdr/data_stream/asset_host_detection/fields/ecs.yml
```diff
@@ -0,0 +1,5 @@
+# Define ECS constant fields as constant_keyword
+- name: observer.vendor
+  type: constant_keyword
```
andrewkroh, line 3: Continue to use `external: ecs` while retaining the override for `type`. This way the description from ECS is "inherited".

packages/qualys_vmdr/docs/README.md
andrewkroh, line 1: I think the new KB enrichment behavior should be described in the readme so that users will know about the enrichment.

Should we disable the knowledge_base data stream by default for 6.0 (i.e `enabled: false` in the stream)? Is my assumption correct that this data is no longer needed because the host assets already contain that information?

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1066,6 +1145,77 @@ processors:
       tag: rename_qualys_vmdr_asset_host_detection_vulnerability_RESULTS_1
       target_field: qualys_vmdr.asset_host_detection.vulnerability.results
       ignore_missing: true
+  # The next two gsub processors facilitate `splitOnToken` in the subsequent script processor, which otherwise cannot split \n or \t.
```
andrewkroh, line 1148: Consider baking this comment into the pipeline definition itself by using the `description` field of the two processors. It gives the comment visibility outside of the source code.

andrewkroh (commented):
Looks good. I _did not_ review the CEL changes.

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml
kcreddy: Got it. Added in https://github.com/elastic/integrations/pull/11999/commits/b4d6599b1bc8aedd7088eb419f747bb0cdec3e1a
Removed processor from `default.yml`

packages/qualys_vmdr/data_stream/asset_host_detection/fields/ecs.yml
kcreddy: Done. README now has the field description. Thanks!

packages/qualys_vmdr/docs/README.md
kcreddy: Updated the README with:
```
Starting from Qualys VMDR integration version 6.0, the `Asset Host Detection` data stream includes enriched vulnerabilities data from Qualys Knowledge Base API.
```
Let me know if that suffices or should be elaborated more. 

> Should we disable the knowledge_base data stream by default for 6.0 (i.e enabled: false in the stream)?

That makes sense. I disabled it by default.

> Is my assumption correct that this data is no longer needed because the host assets already contain that information?

For the user's host assets, yes. If for some reason the users still want to consume all Knowledge, they could still enable and use knowledge_base data stream.

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml
kcreddy: Updated as suggested in https://github.com/elastic/integrations/commit/b4d6599b1bc8aedd7088eb419f747bb0cdec3e1a

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
```diff
@@ -606,3 +757,636 @@ xsd:
...
+        </xsd:sequence>
+        </xsd:complexType>
+      </xsd:element>
+
+      <xsd:element name='CODE' type='xsd:string'>
+      </xsd:element>
+
+      <xsd:element name='TEXT' type='xsd:string'>
+      </xsd:element>
+      </xsd:schema>
```
efd6, line 1392: Final new line.

efd6 (commented):
Looked at the CEL code only. It looks OK mod some nits.

packages/qualys_vmdr/data_stream/asset_host_detection/_dev/test/pipeline/test-asset-host-detection.log
```diff
@@ -1,11 +1,3 @@
...
...LAST_ERROR":"","LAST_ERROR_DATE":"","LAST_STATUS":"Success","LAST_SUCCESS_DATE":"2024-10-23T12:57:43Z","NAME":"ownerId","VALUE":"777217977769"},{"LAST_ERROR":"","LAST_ERROR_DATE":"","LAST_STATUS":"Success","LAST_SUCCESS_DATE":"2024-12-08T15:54:56Z","NAME":"placementGroupName","VALUE":""},{"LAST_ERROR":"","LAST_ERROR_DATE":"","LAST_STATUS":"Success","LAST_SUCCESS_DATE":"2024-12-08T15:54:56Z","NAME":"productCode","VALUE":""},{"LAST_ERROR":"","LAST_ERROR_DATE":"","LAST_STATUS":"Success","LAST_SUCCESS_DATE":"2024-12-08T15:54:56Z","NAME":"productCodeType","VALUE":""},{"LAST_ERROR":"","LAST_ERROR_DATE":"","LAST_STATUS":"Success","LAST_SUCCESS_DATE":"2024-12-08T15:54:56Z","NAME":"spotInstanceRequestId","VALUE":""},{"LAST_ERROR":"","LAST_ERROR_DATE":"","LAST_STATUS":"Success","LAST_SUCCESS_DATE":"2024-12-08T15:54:56Z","NAME":"subnetId","VALUE":"subnet-f2b3ef85"},{"LAST_ERROR":"","LAST_ERROR_DATE":"","LAST_STATUS":"Success","LAST_SUCCESS_DATE":"2024-12-08T15:54:56Z","NAME":"vpcId","VALUE":"vpc-08266f6c"}]}},"NETBIOS":"BASTION-0117d0c9586be67e1.org.US-WEST-2B.AWS.net.CO","NETWORK_ID":"0","OS":"Ubuntu Linux 18.04.6","QG_HOSTID":"34771bd3-0bff-4944-be1f-944b390260e0","TRACKING_METHOD":"AGENT"}
```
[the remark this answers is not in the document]
kcreddy: Addressed in [d82efd7](https://github.com/elastic/integrations/pull/11999/commits/d82efd717b411fe0b21fb25bc7d0ec558d2c6b85)

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
```diff
@@ -26,92 +26,243 @@ redact:
...
+                ?
+                  (ahd_body.?doc.HOST_LIST_VM_DETECTION_OUTPUT.RESPONSE.hasValue() 
+                  ?
+                    ({
+                      "worklist": {
+                        "AHD_LIST": (
+                          ahd_body.?doc.HOST_LIST_VM_DETECTION_OUTPUT.RESPONSE.HOST_LIST.HOST.orValue([]).flatten()
+                        ),
+                        // HOST_QID_LIST has unique QIDs from this batch of hosts.
+                        // When unique QIDs > 1000, it is split into even/odd to avoid "414 Request-URI Too Long" in KB API.
```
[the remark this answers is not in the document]
kcreddy: Updated the comment in [d82efd7](https://github.com/elastic/integrations/pull/11999/commits/d82efd717b411fe0b21fb25bc7d0ec558d2c6b85)

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
```diff
@@ -26,92 +26,243 @@ redact:
...
           "want_more": false,
+          "next_page": {},
         }
     )
   )
 
+- drop_event:
+    when:
+      equals:
+        message: want_more
```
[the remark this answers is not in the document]
kcreddy: Updated text to `retry` in [d82efd7](https://github.com/elastic/integrations/pull/11999/commits/d82efd717b411fe0b21fb25bc7d0ec558d2c6b85)

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
kcreddy: Updated

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
```diff
@@ -26,88 +26,236 @@ redact:
...
+                "want_more": false,
+              }
+          )
+        :
+          ({
+            "events": {
+              "error": {
+                "code": string(resp.StatusCode),
+                "id": string(resp.Status),
+                "message": "GET:"+(
```
efd6, line 133: Sorry, I should have mentioned this in the last review; what I've been doing for GET/POST request is including the URL in the message, e.g. [here](https://github.com/elastic/integrations/blob/951e0aa7091a651b03ed528eb078ae0ed1941b5b/packages/mimecast/data_stream/siem_logs/agent/stream/cel.yml.hbs#L59) and [here](https://github.com/elastic/integrations/blob/951e0aa7091a651b03ed528eb078ae0ed1941b5b/packages/mimecast/data_stream/siem_logs/agent/stream/cel.yml.hbs#L105). This improves the information content of the error.
```suggestion
                "message": "GET "+state.url.trim_right("/") + "/api/2.0/fo/asset/host/vm/detection/: "+(
```
We don't need to include the full request, just enough to identify it.

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
```diff
@@ -26,88 +26,236 @@ redact:
...
         :
-          // Qualys sends 200 success but with no response body. 
-          // Handling this case as okay.
           {
-            "events": [],
+            "events": {
+              "error": {
+                "code": string(resp.StatusCode),
+                "id": string(resp.Status),
+                "message": "GET:"+(
```
efd6, line 243:
```suggestion
                "message": "GET "+state.url.trim_right("/") + "/api/2.0/fo/knowledge_base/vuln/:"+(
```

efd6 (commented):
Minor nits.

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
kcreddy: Updated in [0a39917](https://github.com/elastic/integrations/pull/11999/commits/0a399174f25eea51a9a2de11a3b086bb86adca71)

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
kcreddy: Thanks for catching! Updated in [0a39917](https://github.com/elastic/integrations/pull/11999/commits/0a399174f25eea51a9a2de11a3b086bb86adca71)

packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
```diff
@@ -240,7 +240,7 @@ program: |
               "error": {
                 "code": string(resp.StatusCode),
                 "id": string(resp.Status),
-                "message": "GET:"+(
+                "message": "GET "+state.url.trim_right("/") + "/api/2.0/fo/knowledge_base/vuln/:"+(
```
efd6, line 243: `"GET "+state.url.trim_right("/") + "/api/2.0/fo/knowledge_base/vuln/: "+(` then LGTM

kcreddy: Oh there's another.

kcreddy: Updated as suggested in [f0c8cfd](https://github.com/elastic/integrations/pull/11999/commits/f0c8cfdf60122376366f4c8157b5a1be3da202a3)

packages/qualys_vmdr/data_stream/asset_host_detection/fields/ecs.yml
kcreddy: Fixed the position of the processor leading to build failure in [e2e0e74](https://github.com/elastic/integrations/pull/11999/commits/e2e0e74797759285412cd2fbd6c40e02a93ee3ba)

efd6 (approved):

packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml
kcreddy: Fixed the position of the processor leading to build failure in [e2e0e74](https://github.com/elastic/integrations/pull/11999/commits/e2e0e74797759285412cd2fbd6c40e02a93ee3ba)

kcreddy:
Merging the PR as I got approval from Dan and addressed Andrew's comments who mentioned it otherwise looks good.

PR #12005, opened 2024-12-05
[cef] Add Pre-processor config option to CEF manifest

mjwolf:
## Proposed commit message

Add a config option to add "pre-processors" which will run before the "decode_cef" processor on the agent. CEF messages from some sources may be malformed by not following the CEF spec exactly, which can then cause problems in the "decode_cef" processor. Having a pre-processor will allow running scripts that can correct the malformed CEF messages, and allow decode_cef to process them correctly.

## Checklist

Add a processor to the "pre-processor" config option. For example, a script processor that modifies the CEF message by correcting improper escape sequences.  Install the integration with a log input, and feed the logfile incorrect CEF messages. Verify the pre-processor runs and corrects the messages,and decode_cef has no errors


andrewkroh (commented):
I would like to see if there is a better way that we can handle the pre-processing such that the `event.original` truly reflects what was received off the wire, and does not contain the mutations added by the pre-processors. I think it's important to retain the integrity of the original data.

Perhaps we make a copy into some obscure field like `@metadata.event_original` before the pre-processors, and then reset `event.original` after decode_cef. I'm open to ideas.

packages/cef/data_stream/log/agent/stream/log.yml.hbs
```diff
@@ -17,6 +17,9 @@ publisher_pipeline.disable_host: true
 {{/contains}}
 processors:
 {{#if preprocessors}}
+- set:
```
andrewkroh, line 20: The `set` processor is not valid for Beats. I would be surprised with the system tests passed with this. Here's an example that should work. I used `convert` to perform copy/rename because it will happily overwrite the target field if it exists.

```yaml
# Save an unmodifiable copy of message.
- convert:
    mode: copy
    fields:
      - from: message
        to: '@metadata.event_original'

#
# Pre-processors here.
#

- rename:
    fields:
      - {from: "message", to: "event.original"}
- decode_cef:
    field: event.original

# Overwrite event.original with unmodified value.
- convert:
    mode: rename
    fields:
      - from: '@metadata.event_original'
        to: event.original
```

packages/cef/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.18.0"
+  changes:
+    - description: Add preprocessors config option, to allow running processors before CEF messages are decoded
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/999999
```
andrewkroh, line 6:
```suggestion
      link: https://github.com/elastic/integrations/pull/12005
```

packages/cef/data_stream/log/agent/stream/log.yml.hbs
mjwolf: Thanks, I've changed to use this convert processor. And added a link to the filebeat processors in the manifest description, so no one else is confused about the types of processors supported

packages/cef/data_stream/log/agent/stream/log.yml.hbs
```diff
@@ -16,6 +16,13 @@ tags:
 publisher_pipeline.disable_host: true
 {{/contains}}
 processors:
+{{#if preprocessors}}
+- copy_fields:
+    fields:
+      - from: "message"
+        to: "@manifest.event_original"
```
andrewkroh, line 23: Did you mean `@metadata` which is an internal field that won't get shipped to Elasticsearch if it is accidentally left in the beat event.

packages/cef/data_stream/log/agent/stream/log.yml.hbs
```diff
@@ -24,6 +31,13 @@ processors:
 {{#if decode_cef_timezone}}
     timezone: "{{ decode_cef_timezone }}"
 {{/if}}
+{{#if preprocessors}}
+- convert:
+    mode: rename
+    fields:
+      - from: "@manifest.event_original"
```
andrewkroh, line 38: `@metadata` instead of `@manifest`?

andrewkroh (commented):
Looks good, but I think `@manifest` needs changed to `@metadata`.

packages/cef/data_stream/log/agent/stream/log.yml.hbs
mjwolf: I've changed to `@metadata`

taylor-swanson (approved):

andrewkroh (approved):

PR #12010, opened 2024-12-05
update codeowners for aws bedrock data streams

tommyers-elastic:


.github/CODEOWNERS
```diff
@@ -71,7 +71,9 @@
 /packages/aws/data_stream/waf @elastic/security-service-integrations
 /packages/aws/kibana @elastic/obs-ds-hosted-services @elastic/obs-infraobs-integrations
 /packages/aws/manifest.yml @elastic/obs-ds-hosted-services @elastic/security-service-integrations @elastic/obs-infraobs-integrations
```
tommyers-elastic, line 73: do i need to add a line for the manifest too? i only see this on aws and system integration, and i'm not sure why

lalit-satapathy: In which case all PRs will be routed to both teams.

lalit-satapathy (approved):
LGTM.

.github/CODEOWNERS
tommyers-elastic: isn't this handled by having both teams on the root package path?

lalit-satapathy: yes, thats my understanding.

lalit-satapathy: There may be some redundant entries there.

PR #12027, opened 2024-12-09
tines: work around multimorphic tines.audit_log.inputs.inputs.options…

efd6:
… fields

## Proposed commit message

See title.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/tines/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/tines/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
@@ -159,6 +159,21 @@ processors:
         - remove:
             field: tines.audit_log.inputs.inputs.required
             ignore_failure: true
+  - script:
+      tag: promote_scalar_to_object
+      description: Tines stores both scalar and object values in tines.audit_log.inputs.inputs.options, so make all values fit.
+      lang: painless
+      if: ctx.tines?.audit_log?.inputs?.inputs?.options instanceof List
+      source: |
+        for (int i = 0; i < ctx.tines.audit_log.inputs.inputs.options.length; i++) {
+          def v = ctx.tines.audit_log.inputs.inputs.options[i];
+          if (v instanceof Map) {
+            continue;
+          }
+          Map m = [:];
+          m.value = v;
+          ctx.tines.audit_log.inputs.inputs.options[i] = m;
+        }
   ### cleanup
   - script:
       description: Drops null/empty values recursively.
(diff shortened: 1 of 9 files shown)


packages/tines/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -159,6 +159,21 @@ processors:
         - remove:
             field: tines.audit_log.inputs.inputs.required
             ignore_failure: true
+  - script:
```
efd6, line 162: I'm not wildly happy about having to mutate the document to be able to make it fit, but the previous approach of using subobjects:false did not work. Any suggestions appreciated.

kcreddy: In [subobjects](https://www.elastic.co/guide/en/elasticsearch/reference/current/subobjects.html) documentation, it says: 
> The subobjects setting, which can be applied only to the top-level mapping definition

and the subsequent example shows the `subobjects: false` on parent field whose child can be scalar or object. In our case since `tines.audit_log.inputs.inputs.options` can be both scalar and object, shouldn't we set `subobjects: false` on its parent i.e., `tines.audit_log.inputs.inputs`? Currently we seem to be setting on the [child options](https://github.com/elastic/integrations/blob/main/packages/tines/data_stream/audit_logs/fields/fields.yml#L145-L149) itself.

efd6: I think you might be right. I'll give that a go.

kcreddy: Hope it wont cause any mapping conflicts

packages/tines/data_stream/audit_logs/manifest.yml
```diff
@@ -97,3 +97,7 @@ streams:
         show_user: false
         description: >-
           Processors are used to reduce the number of fields in the exported event or to enhance the event with metadata.  This executes in the agent before the logs are parsed. See [Processors](https://www.elastic.co/guide/en/beats/filebeat/current/filtering-and-enhancing-data.html) for details.
+elasticsearch:
+  index_template:
+    mappings:
+      subobjects: false
```
kcreddy, line 103: Even though the fields are not defined with the dotted names, the [auto-flattening](https://www.elastic.co/guide/en/elasticsearch/reference/current/subobjects.html#subobjects-auto-flattening) should ensure the mapping is automatically flattened. The caveats noted for auto-flattening are not applicable for this package.

kcreddy (approved):
LGTM

PR #12047, opened 2024-12-10
ssi_all: fix string literal syntax in painless scripts

efd6:
## Proposed commit message

See title.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/m365_defender/changelog.yml
+++ b/packages/m365_defender/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.15.2"
+  changes:
+    - description: Fix string literals in painless scripts.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/12047
 - version: "2.15.1"
   changes:
     - description: Add caseless fields to process events.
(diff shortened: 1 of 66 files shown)


kcreddy (commented):
There are few data-streams missed that are owned by our team that have similar problem:

```
azure.firewall_logs
m365_defender.alert
microsoft_sentinel.alert
microsoft_sentinel.event
```

Also, while looking at the CI system test errors, the only data-stream that is not complaining in `google_workspace` package is `admin`. It doesn't seem to have a cleanup null script [at the end](https://github.com/elastic/integrations/blob/main/packages/google_workspace/data_stream/admin/elasticsearch/ingest_pipeline/default.yml#L804-L815). We could add one.

efd6:
I'm unable to reproduce the google_workspace failures locally, so I am backing them out of this change.

packages/m365_defender/changelog.yml
```diff
@@ -1,4 +1,14 @@
 # newer versions go on top
+- version: "2.15.3"
```
andrewkroh, line 2: This package has two versions/changelogs for the change.

kcreddy (approved):

PR #12055, opened 2024-12-10
[tanium] Map extra fields, set event.kind for alerts

chrisberkhout:
## Proposed commit message

```
[tanium] Map extra fields, set event.kind for alerts

- Map extra fields from the new default `threat_response` format.
- Set `event.kind` as a scalar not an array.
- Set `event.kind` to `alert` if an alert ID is present.
```

## Checklist


packages/tanium/data_stream/threat_response/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -33,6 +32,46 @@ processors:
...
+  - rename:
+      tag: rename_json_alert_id
+      field: json.Alert Id
+      target_field: tanium.threat_response.alert_id
+      ignore_missing: true
+  - set:
+      tag: set_event_kind_alert
+      field: event.kind
+      value: alert
+      if: "![null, ''].contains(ctx.tanium?.threat_response?.alert_id)"
```
efd6, line 44: This allocates.

chrisberkhout: changed to
```
if: ctx.tanium?.threat_response?.alert_id != null && ctx.tanium.threat_response.alert_id != ''
```

efd6 (approved):

PR #12064, opened 2024-12-10
[digital_guardian] Map extra timestamps and other fields

chrisberkhout:
## Proposed commit message

```
[digital_guardian] Map extra timestamps and other fields

Based on a list of field names from a live system.
Hasn't been tested with full data.
```

## Discussion

It sets `@timestamp` to `dg_time` (preferred) or `dg_processed_time` or `inc_mtime`.

In the available field information, `dg_time` field has type `timestamp`, so I assume it's in the same format as `inc_mtime`.

The `dg_processed_time` field has a type of `long`, so I assumed it's in `epoch_seconds` format.

## Checklist


packages/digital_guardian/data_stream/arc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -97,11 +97,35 @@ processors:
...
       copy_from: digital_guardian.arc.inc_mtime
       ignore_empty_value: true
+  - date:
+      field: digital_guardian.arc.dg_processed_time
+      tag: date_dg_processed_time
+      target_field: digital_guardian.arc.dg_processed_time
+      formats:
+        - epoch_second
+      if: ctx.digital_guardian?.arc?.dg_processed_time != null
+  - set:
```
andrewkroh, line 112: This setup could overwrite the `@timestamp` value multiple times depending what fields are present. We should probably add an `overwrite: false` to the `set` processors. And if we keep them in the current order no changes will be observed for existing events that have `inc_mtime`

andrewkroh (commented):
Do you think we should document the timestamp fields and their priority order in the readme?

packages/digital_guardian/data_stream/arc/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: I did it so the last one wins, but yes, using `override: false` (and flipping the order) is much better.

I did that and noted the order in the README.

andrewkroh (approved):

PR #12084, opened 2024-12-12
ti_custom: Improve pagination options for TAXII servers

chemamartinez:
## Proposed commit message

This PR adds the following changes:

- Add a `limit` parameter, that can be used to control the size of responses from TAXII servers (see https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html#_Toc31107517)
- To avoid fetching duplicate indicators every interval, now the response header `X-Taxii-Date-Added-Last` is stored in the cursor and used to populate the `added_after` parameter every iteration (see https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html#_Toc31107519)

## Checklist


packages/ti_custom/data_stream/indicator/agent/stream/cel.yml.hbs
```diff
@@ -64,7 +69,16 @@ program: |
             : (has(state.initial_interval) && state.initial_interval != "") ?
                 (
                     state.url.trim_right("/") + "/?" + {
-                        "added_after": [(now() - duration(state.initial_interval)).format(time_layout.RFC3339)],
+                        ?"added_after": has(state.?cursor.last_timestamp) ?
+                                optional.of([state.cursor.last_timestamp])
+                            : has(state.initial_interval) ?
+                                optional.of([(now() - duration(state.initial_interval)).format(time_layout.RFC3339)])
+                            :
+                                optional.none(),
```
chrisberkhout, line 77:
```suggestion
                        ?"added_after": state.?cursor.last_timestamp.optMap(ts,
                            [ts]
                        ).or(
                            state.?initial_interval.optMap(i, [(now() - duration(i)).format(time_layout.RFC3339)])
                        ),

```
Using the the optional types more here works, but is less helpful than the other cases.

Alternative formatting:
```suggestion
                        ?"added_after": state.?cursor.last_timestamp.optMap(ts, [ts]).or(
                            state.?initial_interval.optMap(i, [(now() - duration(i)).format(time_layout.RFC3339)])
                        ),

```

chrisberkhout (changes requested):
The change is good.

You can use `optMap`, etc. to handle optional values in a more streamlined way.

I think that's better at least for the simpler examples. Maybe not worth it for the more complicated one.

packages/ti_custom/data_stream/indicator/agent/stream/cel.yml.hbs
```diff
@@ -64,7 +69,18 @@ program: |
...
+                            [ts]
+                        ).or(
+                            state.?initial_interval.optMap(i, [(now() - duration(i)).format(time_layout.RFC3339)])
+                        ),
+                        ?"added_after": has(state.?cursor.last_timestamp) ?
+                                optional.of([state.cursor.last_timestamp])
+                            : has(state.initial_interval) ?
+                                optional.of([(now() - duration(state.initial_interval)).format(time_layout.RFC3339)])
+                            :
+                                optional.none(),
```
chrisberkhout, line 82: For `added_after`, if you want to keep the `optMap` version then delete the original version.

chrisberkhout (changes requested):
one thing left in by mistake

packages/ti_custom/data_stream/indicator/agent/stream/cel.yml.hbs
chemamartinez: Yeah... left by mistake. Thank you for noticing it!

chrisberkhout (approved):

AkodoGarou:
Good Morning.

I found your integration, and it fits perfect into our TI feed for hashes with another server we have. But I'm running into an issue, where I'm getting a CEL error:

 failed eval: ERROR: <input>:32:22: no such key: objects
   |     ).do_request().as(resp, (resp.StatusCode == 200 || resp.StatusCode == 206) ?
   | .....................^,
  Processor "conditional" with tag "" in pipeline "logs-ti_custom.indicator-0.6.0" failed with message "Error during CEL program evaluation"

Is this where I need to set my CEL url location? And would that be changing the code that was commented out in the "The CEL program to be run for each polling." in settings? Any help would be appreciated.

efd6:
@AkodoGarou From the error you have posted, you are likely using the built-in CEL program. What is happening is that the document being returned by the API endpoint does not contain a field "objects", which the program expects (you can see that [here](https://github.com/elastic/integrations/blob/661a312c6be43b2d0fe81789c42b7cda0ac0b117/packages/ti_custom/data_stream/indicator/agent/stream/cel.yml.hbs#L98)).

There is not enough information in your post to figure out what is going wrong for you, but this really should be a question in community slack or at https://discuss.elastic.co/. If you believe that it is a bug, please file an issue here.

PR #12111, opened 2024-12-16
Add an integration for Rubrik with Filesets datastream

devamanv:
## Proposed commit message
We are developing a new integration package for monitoring Rubrik clusters. The PR adds the first data stream to collect information about Rubrik [Filesets](https://docs.rubrik.com/en-us/saas/saas/fs_filesets.html?hl=filesets). Following are the metrics collected:

| Metric                   | Description                                                                 |
|--------------------------|-----------------------------------------------------------------------------|
| Archive Storage          | Amount of archived storage data, in bytes.                                |
| Archived Snapshot Count  | Number of currently archived fileset snapshots.                           |
| Local Storage            | Amount of local storage used, in bytes, by the fileset.                   |
| Snapshot Count           | Current number of fileset snapshots.                                      |

### Note for reviewers:
Please ignore the fact the fact the changes don't contain the docs yet. That will be added for the entire package later once we have a few data streams merged. 

## Checklist

## Author's Checklist

- [ ] 

- Spin up the Elastic stack and point your `elastic-package` to the running Kibana instance
- Build the package locally and install the package using `elastic-package install --zip rubrik-0.1.0.zip`
- Observe the metrics in Kibana > Discovers section under the `rubrik_filesets` dataset

## Related issues
- Closes https://github.com/elastic/obs-integration-team/issues/115


packages/rubrik/manifest.yml
```diff
@@ -0,0 +1,71 @@
+format_version: 3.0.2
+name: rubrik
+title: "Rubrik "
+version: 0.1.0
+source:
+  license: "Elastic-2.0"
+description: "Collect Metrics from Rubrik RSC with Elastic Agent."
+type: integration
+categories:
+  - custom
```
muthu-mps, line 10:
```suggestion
  - observability
  - custom
```

packages/rubrik/data_stream/linux_filesets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,97 @@
+---
+description: Pipeline to parse rubrik.fileset workload data
+processors:
+  - set:
+      field: ecs.version
+      value: 8.11.0
```
muthu-mps, line 6:
```suggestion
      value: 8.16.0
```

packages/rubrik/data_stream/linux_filesets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,97 @@
...
+      ignore_missing: true
+      if: 'ctx.response?.node?.name != null'
+  - rename:
+      field: response.node.host.osType
+      target_field: rubrik.fileset.host.os_type
+      ignore_missing: true
+      if: 'ctx.response?.node?.host?.osType != null'
+  - rename:
+      field: response.node.reportWorkload.complianceStatus
+      target_field: rubrik.fileset.compliance_status
```
muthu-mps, line 45: Is it okay to ignore `node` and `report_workload` groups from the metric names? Are we not going to get `compliance_status` or the `storage metrics` other than node.reportworkload?

packages/rubrik/data_stream/linux_filesets/fields/fields.yml
```diff
@@ -0,0 +1,31 @@
+- name: rubrik.fileset
+  type: group
+  fields:
+    - name: name
+      type: keyword
+      description: The name of the Fileset.
+    - name: compliance_status
+      type: keyword
+      description: The current compliance status of the workload.
+    - name: archive_storage
```
muthu-mps, line 10: Can we apply the [naming conventions](https://www.elastic.co/guide/en/beats/devguide/7.17/event-conventions.html#:~:text=them%20as%20a-,dotted%20suffix,-when%20possible.%20For) for the `unit:byte` fields.
Add bytes as suffix, `archive_storage.bytes` and similarly for local_storage.

packages/rubrik/manifest.yml
```diff
@@ -0,0 +1,71 @@
+format_version: 3.0.2
+name: rubrik
+title: "Rubrik "
```
muthu-mps, line 3:
```suggestion
title: "Rubrik"
```

packages/rubrik/manifest.yml
```diff
@@ -0,0 +1,71 @@
...
+version: 0.1.0
```
muthu-mps, line 4: The initial version can go as `0.0.1`, If we plan to add the documentation later.

packages/rubrik/data_stream/windows_filesets/_dev/test/pipeline/test-fileset-metrics.log-expected.json
```diff
@@ -0,0 +1,26 @@
+{
+    "expected": [
+        {
+            "ecs": {
+                "version": "8.11.0"
+            },
+            "event": {
+                "kind": "metric",
+                "original": "{\"node\":{\"id\":\"a62483e3-b9ca-5e10-9c27-431ac7f86a47\",\"name\":\"TestFileset\",\"objectType\":\"WindowsFileset\",\"host\":{\"osType\":\"WINDOWS\"},\"reportWorkload\":{\"complianceStatus\":\"OUT_OF_COMPLIANCE\",\"archiveStorage\":0,\"localStorage\":384155,\"totalSnapshots\":2,\"archiveSnapshots\":0,\"location\":\"192.168.10.35\",\"cluster\":{\"id\":\"7f4f96b3-e777-4bde-ba08-47a40ecaa416\",\"name\":\"401-rubrik\",\"systemStatus\":\"OK\",\"type\":\"OnPrem\"},\"__typename\":\"Snappable\"},\"__typename\":\"WindowsFileset\"}}"
```
muthu-mps, line 9: Are we not processing the `location` and the `cluster metrics`? I don't see this fields in the expected data. If this information is specific to windows fileset then we need to process those as well.

muthu-mps:
@devamanv - Why do we have the datastream created per host? Are we expecting only Windows and Linux OS? 
Generally, we replicate the configuration for both the OS types and also the `hostname` input is set at integration level and not specific to the datastream. There is a metric differentiator based on the `node.host.os_type: LINUX or WINDOWS`. Can we use this to differentiate and have only single datastream to process different OS metrics?

packages/rubrik/data_stream/linux_filesets/_dev/test/pipeline/test-fileset-metrics.log-expected.json
```diff
@@ -0,0 +1,26 @@
+{
+    "expected": [
+        {
+            "ecs": {
+                "version": "8.11.0"
```
tommyers-elastic, line 5: shouldn't this be 8.16.0 to match the version specified in the build.yml?

packages/rubrik/data_stream/linux_filesets/agent/stream/httpjson.yml.hbs
```diff
@@ -0,0 +1,102 @@
...
+    query FilesetTemplateListQuery($hostRoot: HostRoot!, $first: Int, $after: String, $filter: [Filter!], $sortBy: HierarchySortByField, $sortOrder: SortOrder) {
+      filesetTemplates(hostRoot: $hostRoot, first: $first, after: $after, filter: $filter, sortBy: $sortBy, sortOrder: $sortOrder) {
+        edges {
+          node {
+            id
+            name
+            physicalChildConnection(filter: $filter) {
+              edges {
+                node {
+                  ... on LinuxFileset {
```
tommyers-elastic, line 33: does this only work for linux hosts?

packages/rubrik/data_stream/linux_filesets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,97 @@
...
+      on_failure:
+        - set:
+            field: error.message
+            value: 'Received invalid JSON. Unable to parse the source log message'
+            ignore_failure: true
+  - rename:
+      field: response.node.name
+      target_field: rubrik.fileset.name
+      ignore_missing: true
+      if: 'ctx.response?.node?.name != null'
```
tommyers-elastic, line 37: do we need the condition, as well as `ignore_missing`?

packages/rubrik/data_stream/linux_filesets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,97 @@
...
+      if: 'ctx.response?.node?.reportWorkload?.localStorage != null'
+  - rename:
+      field: response.node.reportWorkload.totalSnapshots
+      target_field: rubrik.fileset.total_snapshots
+      ignore_missing: true
+      if: 'ctx.response?.node?.reportWorkload?.totalSnapshots != null'
+  - remove:
+      field: response
+      ignore_missing: true
+      if: 'ctx.rubrik.fileset != null'
```
tommyers-elastic, line 71: what's the reason we don't remove `response` if no values were parsed out of it? 

shouldn't this be handled by `event.original` (which i think we also need to conditionally remove somewhere)?

packages/rubrik/data_stream/linux_filesets/fields/fields.yml
```diff
@@ -0,0 +1,31 @@
+- name: rubrik.fileset
+  type: group
+  fields:
+    - name: name
+      type: keyword
+      description: The name of the Fileset.
```
tommyers-elastic, line 6: is this not also a dimension?

packages/rubrik/data_stream/linux_filesets/manifest.yml
```diff
@@ -0,0 +1,62 @@
+type: logs
```
tommyers-elastic, line 1: logs or metrics?

packages/rubrik/data_stream/windows_filesets/agent/stream/httpjson.yml.hbs
```diff
@@ -0,0 +1,101 @@
...
+    query FilesetTemplateListQuery($hostRoot: HostRoot!, $first: Int!, $after: String, $filter: [Filter!], $sortBy: HierarchySortByField, $sortOrder: SortOrder) {
+      filesetTemplates(hostRoot: $hostRoot, first: $first, after: $after, filter: $filter, sortBy: $sortBy, sortOrder: $sortOrder) {
+        edges {
+          node {
+            id
+            name
+            physicalChildConnection(filter: $filter) {
+              edges {
+                node {
+                  ... on WindowsFileset {
```
tommyers-elastic, line 32: aha, my earlier question is answered

packages/rubrik/manifest.yml
```diff
@@ -0,0 +1,71 @@
...
+source:
+  license: "Elastic-2.0"
+description: "Collect Metrics from Rubrik RSC with Elastic Agent."
+type: integration
+categories:
+  - custom
+conditions:
+  kibana:
+    version: "^8.15.0"
```
tommyers-elastic, line 13: this doesn't match the ECS version, does that matter? (we should also add 9.0 here too ` || "^9.0.0"`)

tommyers-elastic (commented):
looking good! 

there so much repeated code between the two fileset datastreams, is there no way to combine them into one?

packages/rubrik/data_stream/linux_filesets/manifest.yml
lalit-satapathy: If we stick with [httpjson](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-httpjson.html), we may have choice but logs. 

Another example [here](https://github.com/elastic/integrations/blob/main/packages/citrix_adc/data_stream/system/manifest.yml#L2).

devamanv:
> there so much repeated code between the two fileset datastreams, is there no way to combine them into one?

@tommyers-elastic this is attributed to how the query field is defined in the schema 

```graphql
# Get a summary of all fileset templates.
filesetTemplates(
  first: Int,
  after: String,
  sortBy: HierarchySortByField,
  sortOrder: SortOrder,
  filter: [Filter!],
  hostRoot: HostRoot!
): FilesetTemplateConnection!
```
The thing to notice here is the hostRoot parameter, that is required and can only be one of `LINUX_HOST_ROOT` or `WINDOWS_HOST_ROOT` but NOT BOTH. 

I still went ahead and had tried to query both Filesets as part of the same query. Yes that's doable for less complex queries and has been done for other data streams. 
Since the query to fetch all the required details is already fairly complex, that didn't help either and I got the following error:

```json
{
  "code": 400,
  "uri": "/api/graphql",
  "traceSpan": {
    "traceId": "oWkhOC8ofr9i3EjX5wSqgg==",
    "operation": "/api/graphql",
    "spanId": "qXoKqEuj6gE="
  },
  "message": "[QueryAnalysisError] [Filesets Team] Encountered Client error (400) executing query with operations: [FilesetTemplateListQuery] and variables {}. Error: Query reducing error: Query is too expensive. Actual Complexity (1936) >= Threshold Complexity (1001)."
}
```
which clearly says the query got too expensive to execute and fetch results. I tried but couldn't find an alternative or a better approach to pull both Windows and Linux filesets together. Consequently, had to duplicate part of the queries, and hence two data streams. 

So, the TLDR; separate those out like I have done in this case, or somehow fire off another follow-up request to fetch for the other HostRoot value. 

In that regard, I had also thought if we could some how [chain](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-httpjson.html#_chain_step_request) the requests in the `httpjson` input. 
I could try that if you think that might work. Please let me know if you have any questions and or suggestions.

devamanv:
> Why do we have the datastream created per host? Are we expecting only Windows and Linux OS? 

@muthu-mps 
Yes, Filesets are usually of type Windows or Linux/Unix. There's one more called NAS type, but I couldn't find any workload or an object of that in the inventory either. The inventory in Rubrik's terminology is a place where can see all the workloads grouped together based on the type. Like so:

image

> Generally, we replicate the configuration for both the OS types and also the hostname input is set at integration level and not specific to the datastream. 

No, the hostname here is deliberately set at the integration level because the all the queries and datastreams are fetched from a single hostname(i.e. the Rubrik Cloud instance) and even the API URL remains the same i.e. `hostname.my.rubrik.com/api/graphql`. 
So, no matter how many data streams you enable, the configuration will be required only once, and can be used everywhere. I'm of the opinion that it should be like this, unlike other integrations where URIs change per datastream, and are usually REST based. 
 
> There is a metric differentiator based on the node.host.os_type: LINUX or WINDOWS. Can we use this to differentiate and have only single datastream to process different OS metrics?

Unless we expect the user to request only one of Windows or Linux fileset, that is the approach we need to follow. I tried to explain why I had to create two data streams in the [comment](https://github.com/elastic/integrations/pull/12111#issuecomment-2560275936) above. But with the current approach, we give user the flexibility to requests one or both types.

packages/rubrik/manifest.yml
ishleenk17: No, the ECS version and the Kibana versions need not be same. 
My suggestion, we should add th 9.0.0 support, while merging the final datastream PR to this Integration

packages/rubrik/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "0.1.0"
```
ishleenk17, line 2: Please use the -next tag for the follow up datastream versions, so that the release happens together for the user. 
This was done for vSphere as well

packages/rubrik/data_stream/windows_filesets/fields/fields.yml
```diff
@@ -0,0 +1,31 @@
+- name: rubrik.fileset
```
ishleenk17, line 1: Since there are 2 separate datastreams for linux and windows. 
IMO, its better to have fields as, rubrik.windows_fileset.* and
rubrik.linux_fileset.
More comprehensive for user.

packages/rubrik/data_stream/windows_filesets/manifest.yml
```diff
@@ -0,0 +1,62 @@
...
+        description: >
+          Specify filters for refining the data. Filters must be defined as an array of Filter objects. Refer to the [Rubrik API documentation](https://rubrikinc.github.io/rubrik-api-documentation/schema/reference/filter.doc.html) for valid fields.
+      - name: processors
+        type: yaml
+        title: Processors
+        multi: false
+        required: false
+        show_user: false
+        description: >
+          Processors are used to reduce the number of fields in the exported event or to enhance the event with metadata.  This executes in the agent before the logs are parsed.  See [Processors](https://www.elastic.co/guide/en/beats/filebeat/current/filtering-and-enhancing-data.html) for details.
```
ishleenk17, line 58: Please use the Agents link to RPeocessors

packages/rubrik/data_stream/windows_filesets/manifest.yml
```diff
@@ -0,0 +1,62 @@
...
+
+    template_path: httpjson.yml.hbs
+    title: Rubrik Windows Filesets
```
ishleenk17, line 61:
```suggestion
    title: Rubrik Windows Fileset
```

packages/rubrik/docs/README.md
```diff
@@ -0,0 +1 @@
+# Instructions to setup Rubrik Integration
```
ishleenk17, line 1: Is this TBD ?

packages/rubrik/manifest.yml
ishleenk17: Lets remove custom. Doesn't add any value to the category

packages/rubrik/docs/README.md
ishleenk17: Got it answered in the dwscription. Thanks

packages/rubrik/data_stream/windows_filesets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,97 @@
+---
+description: Pipeline to parse rubrik.fileset workload data
```
ishleenk17, line 2: We are having all the fields related to workload. 
But we are not using workload in any of the field names. 
IMO, we should have fields like
rubrik.windows_fileset.workload.compliance_status

packages/rubrik/data_stream/windows_filesets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,97 @@
...
+      on_failure:
+        - set:
+            field: error.message
+            value: 'Received invalid JSON. Unable to parse the source log message'
+            ignore_failure: true
+  - rename:
+      field: response.node.name
+      target_field: rubrik.fileset.name
+      ignore_missing: true
+      if: 'ctx.response?.node?.name != null'
```
ishleenk17, line 37: We don't need these if checks if we are using ignore_missing.

packages/rubrik/data_stream/windows_filesets/fields/fields.yml
tommyers-elastic: the two configurations return the same metrics though right? separating them makes it difficult to query for the same metrics on both windows and linux at the same time... i think the shared root namespace makes sense.

tommyers-elastic:
@devamanv if we could chain the calls it would be ideal. at least then we only have the httpjson config repeated, and not all the rest of the mappings and boilerplate.

packages/rubrik/data_stream/windows_filesets/fields/fields.yml
ishleenk17: We should ideally not have a metric with same name under 2 different datastreams then.
In this case, customer will see a metric named

rubrik.fileset.compliance_status in both windows and linux fileset datastreams.
We should then ideally combine these 2 datastreams or have the mapping once. This gets confusing from user perspective.

devamanv: @tommyers-elastic yes that's why I kept the same root namespace, the metric names remain the same, just the values that change. Even I'm of the opinion that we should be keep them as is, if we are to continue with two datastreams.

devamanv:
> if we could chain the calls it would be ideal. at least then we only have the httpjson config repeated, and not all the rest of the mappings and boilerplate.

I'll will give that a try to see If it works.

packages/rubrik/changelog.yml
```diff
@@ -7,6 +7,9 @@
     - description: Add `virtualmachines` data stream
       type: enhancement
       link: https://github.com/elastic/integrations/pull/12205
+    - description: Add fileset data stream
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/12205
```
andrewkroh, line 12:
```suggestion
      link: https://github.com/elastic/integrations/pull/12111
```

packages/rubrik/data_stream/filesets/manifest.yml
```diff
@@ -0,0 +1,72 @@
+type: metrics
+title: Rubrik Fileset Metrics
+streams:
+  - input: cel
+    title: Rubrik Fileset Metrics
+    description: Collect Rubrik Fileset data via the RSC GraphQL API
+    enabled: true
+    template_path: cel.yml.hbs
+    vars:
+      - name: http_client_timeout
```
gpop63, line 10: We can remove this

packages/rubrik/data_stream/filesets/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,190 @@
...
+auth.oauth2:
+  client.id: {{client_id}}
+  client.secret: {{client_secret}}
+  token_url: {{token_url}}
+
+resource.url: https://{{hostname}}.my.rubrik.com/api/graphql
+{{#if resource_ssl}}
+resource.ssl: 
+  {{resource_ssl}}
+{{/if}}
```
gpop63, line 12:
```suggestion
{{#if ssl}}
resource.ssl: 
  {{ssl}}
{{/if}}
```

packages/rubrik/data_stream/filesets/manifest.yml
```diff
@@ -0,0 +1,72 @@
...
+        multi: false
+        default: false
+      - name: first
+        type: integer
+        title: First
+        multi: false
+        required: false
+        show_user: false
+        description: Returns the first n elements from the list.
+      - name: after
```
gpop63, line 57: Can remove this, will be automatically set in cel pagination logic

packages/rubrik/data_stream/filesets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,105 @@
+---
+description: Pipeline to parse rubrik.fileset data
+processors:
+  - set:
+      field: ecs.version
+      value: 8.16.0
+      ignore_failure: true
+      ignore_empty_value: true
```
gpop63, line 8:
```suggestion
      field: ecs.version
      value: 8.16.0
```

packages/rubrik/data_stream/filesets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,105 @@
...
+  - set:
+      field: event.kind
+      value: metric
+      ignore_failure: true
+      ignore_empty_value: true
```
gpop63, line 13:
```suggestion
      field: event.kind
      value: metric
```

packages/rubrik/data_stream/filesets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,105 @@
...
+        if (ctx.containsKey('rubrik') && ctx.rubrik.containsKey('fileset')) {
+          drop(ctx.rubrik.fileset);
+        }         
+on_failure:
+  - set:
+      field: event.kind
+      value: pipeline_error
+  - set:
+      field: error.message
+      value: "{{{_ingest.on_failure_message}}}"
```
gpop63, line 105:
```suggestion
on_failure:
  - set:
      field: event.kind
      value: pipeline_error
  - append:
      field: error.message
      value: >-
        Processor '{{{ _ingest.on_failure_processor_type }}}'
        {{{#_ingest.on_failure_processor_tag}}}with tag '{{{ _ingest.on_failure_processor_tag }}}'
        {{{/_ingest.on_failure_processor_tag}}}failed with message '{{{ _ingest.on_failure_message }}}'
```

packages/rubrik/data_stream/filesets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,105 @@
...
+            return (((Map) o).size() == 0);
+          } else if (o instanceof List) {
+            ((List) o).removeIf(v -> drop(v));
+            return (((List) o).size() == 0);
+          }
+          return false;
+        }
+        if (ctx.containsKey('rubrik') && ctx.rubrik.containsKey('fileset')) {
+          drop(ctx.rubrik.fileset);
+        }
```
gpop63, line 98: I think you should add `ignore_failure: true` to this script. Also is this if check necessary?

packages/rubrik/data_stream/linux_filesets/fields/fields.yml
devamanv: Yes. Marked this as dimension.

gpop63 (approved):

packages/rubrik/data_stream/filesets/manifest.yml
```diff
@@ -0,0 +1,65 @@
...
+    template_path: cel.yml.hbs
+    vars:
+      - name: resource_timeout
+        type: text
+        title: Resource Timeout
+        description: Duration before declaring that the HTTP client connection has timed out. Valid time units are ns, us, ms, s, m, h. Default is "30"s.
+        show_user: false
+        multi: false
+        required: false
+        default: 60s
```
shmsr, line 17: Description says 30s. Also can we remove quotes from "30" in the description?

packages/rubrik/data_stream/filesets/manifest.yml
```diff
@@ -0,0 +1,65 @@
...
+        type: bool
+        multi: false
+        default: false
+      - name: first
+        type: integer
+        title: First
+        multi: false
+        required: false
+        show_user: false
+        description: Returns the first n elements from the list.
```
shmsr, line 56: This is weird config. Just the name is first? Also from description it does not seem clear what is it?

packages/rubrik/data_stream/filesets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,103 @@
...
+  - remove:
+      field: response
+      ignore_missing: true
+  - script:
+      description: Drops null/empty values recursively
+      lang: painless
+      ignore_failure: true
+      source: |
+        boolean drop(Object o) {
+          if (o == null || o == "") {
```
shmsr, line 82:
```suggestion
          if (o == null || (o instanceof String && o.isEmpty())) {
```

see if it works?

shmsr (approved):
Just few comments; rest looks good. Approving.

packages/rubrik/data_stream/filesets/elasticsearch/ingest_pipeline/default.yml
devamanv: Although seems correct and more robust, I'm just wondering if we should change it at all. Given the fact that we're only concerned with map and list types for this logic to work, for all other types it will simply be false, and move to other cases.  Since this script is used in almost all the integrations, so to be consistent, IMO we should keep it unchanged. Let me know if you still think it should be changed, and I shall add in the next PR.

packages/rubrik/data_stream/filesets/manifest.yml
devamanv: First is supposed to be a filter option. I will make the description a little more clear.

PR #12112, opened 2024-12-16
[M365 Defender] change connection_string to secret

SimonKoetting:
Change property connection_string to be a secret like in the other integrations like https://github.com/elastic/integrations/blob/main/packages/azure/manifest.yml

### DIFF
--- a/packages/m365_defender/changelog.yml
+++ b/packages/m365_defender/changelog.yml
@@ -1,5 +1,5 @@
 # newer versions go on top
-- version: "2.18.0"
+- version: "2.17.1"
   changes:
     - description: Change connection_string to be a secret
       type: bugfix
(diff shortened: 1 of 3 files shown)


packages/m365_defender/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.18.0"
```
efd6, line 2:
```suggestion
- version: "2.17.1"
```
We use semver, so bugfixes are patch-bumping.

packages/m365_defender/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: "3.0.2"
 name: m365_defender
 title: Microsoft M365 Defender
-version: "2.17.0"
+version: "2.18.0"
```
efd6, line 4:
```suggestion
version: "2.17.1"
```

efd6 (approved):

Issue #10419, opened 2024-07-08
tenable_sc: lastSeen filter is incorrectly formatted and uses incorrect time


efd6:
There is a syntax discordance with the [docs](https://docs.tenable.com/security-center/best-practices/api/Content/RetrieveVulnerabilityDataForSpecificTimeRange.htm) and the implementation in the integration; the docs have date ranges with the format "#:#" while we are constructing ["#-#"](https://github.com/elastic/integrations/blob/b99f669c8604f3239f5b0360f3786b40d6709b38/packages/tenable_sc/data_stream/vulnerability/agent/stream/httpjson.yml.hbs#L33) where # is a range endpoint.

In addition to this, the documentation explains that the start and end time points in the range are days in the past relative to today, while the integration presents the time values as absolute times in unix time stamps based on the `lastSeen` field in the returned documents from the API.

andrewkroh:
@cpascale43 Can you please check if we have any partnership with Tenable where we can get access to Tenable SC. I think that is best way to get a definitive answer to the what is the correct format needed in the request from our integration.

---------

Given that we likely already have a persisted cursor that is invalid, if we need to swap httpjson to CEL for the vulnerability data stream to do numeric operations then let's do it.

matthewabbott:
Hi @cpascale43 and @andrewkroh, have you been able to get access to Tenable SC and determine the necessary request format? Just wondering if you’ve had the chance to make progress on this issue.

PR #12114, opened 2024-12-16
Tenable_sc: Update API lastseen parameter format

kcreddy:
## Proposed commit message
The API documentation for `vulnerability` mentions the `lastSeen` parameter format should be `#:#` where `#` is in number of days.
Although the existing format `@-@` where `@` is epoch seconds is still working, this discrepancy is addressed to avoid any future issues.

Other changes:
- Updated request's `header.User-Agent` versions in input files.

Ref: https://github.com/elastic/integrations/issues/10419
API doc: https://docs.tenable.com/security-center/best-practices/api/Content/RetrieveVulnerabilityDataForSpecificTimeRange.htm

## Checklist

## Author's Checklist

- [x] Verify existing and new formats result in same number of documents.

1. Verified the number of documents matched between current version and new change version with 3 different values:
  a. Initial Interval as `24h`
  b. Initial Interval as `48h`
  c. Initial Interval as `4h`
1. Run pipeline tests for package. Should pass successfully.
`eval "$(elastic-package stack shellinit)" && elastic-package test pipeline --generate -v`
```
--- Test results for package: tenable_sc - START ---
╭────────────┬───────────────┬───────────┬───────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE    │ DATA STREAM   │ TEST TYPE │ TEST NAME                                         │ RESULT │ TIME ELAPSED │
├────────────┼───────────────┼───────────┼───────────────────────────────────────────────────┼────────┼──────────────┤
│ tenable_sc │ asset         │ pipeline  │ (ingest pipeline warnings test-asset.log)         │ PASS   │ 322.975375ms │
│ tenable_sc │ asset         │ pipeline  │ test-asset.log                                    │ PASS   │  65.882958ms │
│ tenable_sc │ plugin        │ pipeline  │ (ingest pipeline warnings test-plugin.log)        │ PASS   │   302.7675ms │
│ tenable_sc │ plugin        │ pipeline  │ test-plugin.log                                   │ PASS   │  58.123791ms │
│ tenable_sc │ vulnerability │ pipeline  │ (ingest pipeline warnings test-vulnerability.log) │ PASS   │ 323.123834ms │
│ tenable_sc │ vulnerability │ pipeline  │ test-vulnerability.log                            │ PASS   │  120.73975ms │
╰────────────┴───────────────┴───────────┴───────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: tenable_sc - END   ---
Done
```
3. Run system tests for package. Should pass successfully.
`eval "$(elastic-package stack shellinit)" && elastic-package test system --generate -v`
```
--- Test results for package: tenable_sc - START ---
╭────────────┬───────────────┬───────────┬───────────┬────────┬───────────────╮
│ PACKAGE    │ DATA STREAM   │ TEST TYPE │ TEST NAME │ RESULT │  TIME ELAPSED │
├────────────┼───────────────┼───────────┼───────────┼────────┼───────────────┤
│ tenable_sc │ asset         │ system    │ default   │ PASS   │   36.7143465s │
│ tenable_sc │ plugin        │ system    │ default   │ PASS   │ 39.125610417s │
│ tenable_sc │ vulnerability │ system    │ default   │ PASS   │  34.20195875s │
╰────────────┴───────────────┴───────────┴───────────┴────────┴───────────────╯
--- Test results for package: tenable_sc - END   ---
Done
```

## Related issues

- Closes https://github.com/elastic/integrations/issues/10419


packages/tenable_sc/changelog.yml
```diff
@@ -1,4 +1,15 @@
 # newer versions go on top
+- version: "2.0.0"
```
kcreddy, line 2: Incremented major version because `batch_size, interval, and initial_interval` are moved from package level to data-stream level.

packages/tenable_sc/data_stream/vulnerability/agent/stream/httpjson.yml.hbs
```diff
@@ -30,13 +31,13 @@ request.transforms:
...
             "operator":"=",
-            "value": "[[.cursor.last_event_ts]]-[[(now).Unix]]"
+            "value": "0:[[if .cursor.last_event_ts]][[div (add (toInt (now).Unix) (mul (toInt -1) (toInt .cursor.last_event_ts))) 86400]][[end]]"
         }
       default: |-
         {
             "filterName":"lastSeen",
             "operator":"=",
-            "value": "[[(now (parseDuration "-{{initial_interval}}")).Unix]]-[[(now).Unix]]"
+            "value": "0:{{initial_interval}}"
```
efd6, line 40: This doesn't look like it takes units into account. The API expects this to be in days, but we accept h/m/s (and others). Should we say the value is given in days or calculate days from the duration?

packages/tenable_sc/data_stream/vulnerability/agent/stream/httpjson.yml.hbs
```diff
@@ -30,13 +31,13 @@ request.transforms:
         {
             "filterName":"lastSeen",
             "operator":"=",
-            "value": "[[.cursor.last_event_ts]]-[[(now).Unix]]"
+            "value": "0:[[if .cursor.last_event_ts]][[div (add (toInt (now).Unix) (mul (toInt -1) (toInt .cursor.last_event_ts))) 86400]][[end]]"
```
efd6, line 34: I think this should have an `[[else]]`, probably reaching back `initial_interval`.

I'm not sure this calculation is right. What we have is equivalent to (now.Unix - .cursor.last_event_ts)/86400. This will truncate the value meaning that we will potentially lose data; if the cursor value indicates a non-integer day, the time interval calculated in the numerator will be the fraction less than a full day and the remainder of that day will be truncated by the division.

Also, is `.cursor.last_event_ts` a type that results in epoch seconds when converted to an integer? From the docs, and the code below this does not look like a time value. It looks like it is very hard to hold this API correctly.

packages/tenable_sc/data_stream/vulnerability/agent/stream/httpjson.yml.hbs
kcreddy: The intention was to change the `initial_interval` format. I updated it to default of `1` from existing `24h`. I also updated the description to mention it should be an integer in days.

efd6: Ah, I missed that because of the change in the package manifest. I think this is likely to cause confusion.

packages/tenable_sc/data_stream/vulnerability/agent/stream/httpjson.yml.hbs
kcreddy: > I think this should have an [[else]], probably reaching back initial_interval

We have the `default` block, but I also think it would make sense to avoid template error. I will add the `[[else]]` case.

> I'm not sure this calculation is right.
> Also, is .cursor.last_event_ts a type that results in epoch seconds when converted to an integer? 

Based on the samples ingested, the value of `.cursor.last_event_ts` which is derived from `.last_event.lastSeen` is always in epoch seconds. I verified that `lastSeen` in the response of each object is epoch seconds. I am not sure if this can make cursor value a non-integer day.  
I will remove the int conversion of `.cursor.last_event_ts` as it is not required.

efd6: > Based on the samples ingested, the value of `.cursor.last_event_ts` which is derived from `.last_event.lastSeen` is **always in epoch seconds**.

Is the documentation for this? 

> I am not sure if this can make cursor value a non-integer day.

If the value from the document has come from any time other than 00:00, it will result in a non-integer day before integer truncation, and the truncation will result in the issue that I raised.

kcreddy: Updated the logic as per ceil division to avoid data loss: https://github.com/elastic/integrations/pull/12114/commits/a7055b90e04d671cdbe32a92a5f3b7175f9626f4

Here are the results:
1. Re-verified documents counts between exiting package version `1.27.0` and new change version `2.0.0`. The counts match.
2. Verified from tracers the lastSeen is correctly getting applied (`Initial Interval: 2`):
   - First request: `{\"filterName\":\"lastSeen\",\"operator\":\"=\",\"value\":\"0:2\"}`
   - Second request (after interval): `{\"filterName\":\"lastSeen\",\"operator\":\"=\",\"value\":\"0:1\"}`

packages/tenable_sc/data_stream/vulnerability/agent/stream/httpjson.yml.hbs
efd6: I think it would be nice to be able to use the same syntax here, to avoid confusion and the conversion from a `time.Duration` to a day count is now a solved problem, so it should be easy to do.

packages/tenable_sc/changelog.yml
kcreddy: No longer relevant as the config is moved back to package level.

packages/tenable_sc/data_stream/vulnerability/agent/stream/httpjson.yml.hbs
kcreddy: Updated in https://github.com/elastic/integrations/pull/12114/commits/4a4f1ead9bfe4c126b0c2ab7cffa5b1ae7378cf4. Also moved `batch_size, interval, and initial_interval` back to package level. Removing the need for breaking change.

kcreddy: Tested with 2 scenarios:

1. `Initial Interval: 48h`:
   - First request: `{\"filterName\":\"lastSeen\",\"operator\":\"=\",\"value\":\"0:2\"}`. This returns last event's `lastSeen` as `1734500477`.
   - Second request (after interval): `{\"filterName\":\"lastSeen\",\"operator\":\"=\",\"value\":\"0:1\"}`

1. `Initial Interval: 4h`:
   - First request: `{\"filterName\":\"lastSeen\",\"operator\":\"=\",\"value\":\"0:1\"}`. This returns last event's `lastSeen` as `1734500477`.
   - Second request (after interval): `{\"filterName\":\"lastSeen\",\"operator\":\"=\",\"value\":\"0:1\"}`

efd6 (approved):

Issue #11952, opened 2024-12-02
[meta] Upgrade integrations to ECS 8.17


taylor-swanson:
## Guide

It is recommended to split the changes into smaller PRs to limit the number of changed files. Generally speaking, 10 integrations per PR is a good number to target.

### Automated method

Use the `ecs-update` tool which can be [found here](https://github.com/andrewkroh/go-examples/tree/main/ecs-update).

```
cd packages/
ecs-update -pr 999999 -ecs-git-ref v8.17.0 -ecs-version 8.17.0 -owner elastic/owner-name-here package_1 [package_2 ...]
```

Once PR is filed, the changelogs will need to be updated with the correct PR number.

### Manual method

Update ECS references in integrations to version 8.17.

- Update reference in `_dev/build/build.yml` to `v8.17.0`
- Update package changelog/manifest and regenerate README.

```diff
dependencies:
  ecs:
-    reference: "git@v8.11.0"
+    reference: "git@v8.17.0"
```

## ECS 8.17 Changes

https://github.com/elastic/ecs/releases/tag/v8.17.0

#### Bugfixes

* Fix link rendering issues and usage of http in links. #2423

#### Improvements

* Increase ignore_above value for url.query. #2424
* Set synthetic_source_keep = none on fields that represent sets. #2422

### @elastic/cloud-security-posture:
- [ ] cloud_security_posture

### @elastic/ecosystem:
- [ ] elastic_package_registry

### @elastic/elastic-agent:
- [ ] elastic_agent

### @elastic/elastic-agent-data-plane: 
- [ ] filestream
- [ ] journald
- [ ] linux
- [ ] log
- [ ] windows

### @elastic/fleet:
- [ ] fleet_server

### @elastic/obs-cloudnative-monitoring:
- [ ] containerd
- [ ] docker
- [ ] istio
- [ ] kubernetes
- [ ] kubernetes_otel
- [ ] nginx_ingress_controller

### @elastic/obs-ds-hosted-services: 
- [ ] aws
- [ ] aws_logs
- [ ] awsfirehose
- [ ] azure
- [ ] azure_metrics
- [ ] gcp
- [ ] gcp_metrics

### @elastic/obs-ds-intake-services: 
- [ ] apm
- [ ] profiler_collector
- [ ] profiler_symbolizer

### @elastic/obs-infraobs-integrations:
- [ ] activemq
- [x] airflow https://github.com/elastic/integrations/pull/15287
- [ ] apache
[...]

PR #12119, opened 2024-12-16
Update Deployment and Devices integrations to ECS 8.16.0 (part 1) 

taylor-swanson:
## Proposed commit message

Updates the following integrations to ECS 8.16.0:

- arista_ngfw
- bluecoat
- cef
- checkpoint
- cisco_aironet
- cisco_asa
- cisco_ftd
- cisco_ios
- cisco_ise
- cisco_nexus
- cisco_secure_email_gateway
- citrix_waf

## Checklist

## Related issues

- Relates #11952


packages/checkpoint/docs/README.md
```diff
@@ -793,5 +793,5 @@ An example event for `firewall` looks as following:
 | user_agent.name | Name of the user agent. | keyword |
 | user_agent.original | Unparsed user_agent string. | keyword |
 | user_agent.original.text | Multi-field of `user_agent.original`. | match_only_text |
-| vulnerability.id | The identification (ID) is the number portion of a vulnerability entry. It includes a unique identification number for the vulnerability. For example (https://cve.mitre.org/about/faqs.html#what_is_cve_id)[Common Vulnerabilities and Exposure CVE ID] | keyword |
+| vulnerability.id | The identification (ID) is the number portion of a vulnerability entry. It includes a unique identification number for the vulnerability. For example (https://cve.mitre.org/about/faqs.html#what_is_cve_id[Common Vulnerabilities and Exposure CVE ID]) | keyword |
```
dwhyrock, line 796: This doesn't look correct in the rich diff.

Shouldn't it be something like `[Common Vulnerabilities and Exposure CVE ID](https://cve.mitre.org/about/faqs.html#what_is_cve_id)` ?

dwhyrock (commented):
Looks like there's a hyperlink that needs some attention.  I will approve once it's fixed.

packages/checkpoint/docs/README.md
taylor-swanson: Oh good catch.

Let me go check ECS, I didn't hand write that, that's pulled directly from the ECS definition (but this could also be an issue with how elastic-package is rendering the markdown).

taylor-swanson: Here's the related PR: https://github.com/elastic/ecs/pull/2328

The links render correctly for the [ECS docs](https://www.elastic.co/guide/en/ecs/current/ecs-vulnerability.html#field-vulnerability-id), but the syntax used for the link is not valid markdown. The [integrations doc site](https://www.elastic.co/guide/en/integrations/current/checkpoint.html) seems to use the readme markdown more or less directly, so that will have problems, too.

For comparison, the [threat fields](https://www.elastic.co/guide/en/ecs/current/ecs-threat.html#field-threat-tactic-id) do not use a special markup, but rather use the link directly. The ECS docs and the integration readme page in Kibana (Suricata is an example), and the integrations doc site all enrich the URL into a link. Seems like we don't need any special markup here (unless you want to hide the URL behind friendly text, but that doesn't work if it doesn't render correctly).

Unfortunately, this would need to be fixed in ECS, first. Manually fixing it here won't help, since any re-render of the readme from here will just bring the issue back. It just so happens there's an ECS meeting today. I'll bring this issue up there and see where we go from here.

dwhyrock (approved):

PR #12139, opened 2024-12-17
update documentation to align with markup schema

qcorporation:
Change Log:
- Noticed that rendering of event fields were a problem due to `< >` (angle brackets)
- For all event id that have parsing logic updated the documentation


dwhyrock (approved):
Nice catch.

Interesting that it was fine with any underscore names and only filtered out the single words that were in angle brackets.

PR #12146, opened 2024-12-18
[vsphere] Add support for Shared Secrets /SSL Thumbprint Logins and add more granular filtering for file pipeline

SimonKoetting:
Vsphere also supports shared secrets & SSL thumbprints to login. When the login happens with a shared secret or ssl thumbprint, the log message looks slighty diffrent which currently caused the dissect to fail.

Also the file pipeline just filtered if the message contains "Upload", this causes some errors when there are logs containing the Upload string like in  `/LicensingStats/LicenseServiceProxy::UploadLicenseFile` which is a complete diffrent log.


kush-elastic (approved):
LGTM

packages/vsphere/data_stream/log/elasticsearch/ingest_pipeline/login.yml
```diff
@@ -9,8 +9,33 @@ processors:
...
+    description: logged in as User
+- dissect:
+    field: message
+    pattern: "%{} Shared secret from %{client.ip} %{_tmp.event} as %{user_agent.original}"
+    if: 'ctx?.message != null && (ctx.message?.contains(''logged in as'') && ctx.message?.contains(''Shared secret''))'
+    description: logged in as Shared Secret
+- dissect:
+    field: message
+    pattern: "%{} SSL thumbprint logged in as %{user_agent.original}"
+    if: 'ctx?.message != null && (ctx.message?.contains(''logged in as'') && ctx.message?.contains(''SSL thumbprint'')&& ctx.message?.contains(''User '') == false)'
```
muthu-mps, line 22: Does the **contains(''User '')** check is applicable only for SSL thumbprint and not for Shared Secret?

packages/vsphere/data_stream/log/_dev/test/pipeline/test-format-common.log-expected.json
```diff
@@ -5866,6 +5866,111 @@
...
+            },
+            "message": "info vsansystem[21254123] [vSAN@1234 sub=AccessChecker opId=011a11e2-7123] SSL thumbprint logged in as VMware-client/6.5.0 ",
+            "process": {
+                "name": "Hostd"
+            },
+            "tags": [
+                "preserve_original_event"
+            ],
+            "user": {
+                "name": "SSL thumbprint"
```
muthu-mps, line 5965: Is it okay to accept `SSL thumbprint` and `Shared Secret` as user names when this is from non user logins?

packages/vsphere/data_stream/log/elasticsearch/ingest_pipeline/login.yml
SimonKoetting: I guess that's not strictly necessary. I used it as a kind of "double" check because User should not be part of the message, as  otherwise the general grok should apply. But yes, as we already ask for "if SSL thumbprint" is in the text that should be enough, so we can remove the user check.

packages/vsphere/data_stream/log/_dev/test/pipeline/test-format-common.log-expected.json
SimonKoetting: Good point, I also asked me that :D TBH, I don't know where to put it else. I get that It's not technically a user, on the other hand, I thought about reports like "top 10 logged-in Users", where user.name is used to group the result. 
But I'm open to changing this if you have another idea

muthu-mps: - In that case, keep the names as `shared_secret_login` and `ssl_thumbprint_login` which gives more clarity on the login and then mention explicitly in the document about this log format. 
- or we can even keep the name as `non_user_login` but it is less meaningful for me and doesn't justify on the mode of login.
- If we consider this logins that we are `not treating as logged-in users` then we can exclude this two values from user.name in the visualisation for Top-10 logged-in users.
- WDYT?

packages/vsphere/data_stream/log/elasticsearch/ingest_pipeline/login.yml
SimonKoetting: updated this to not filter for "not contains User" as it would be a duplicate check

packages/vsphere/data_stream/log/_dev/test/pipeline/test-format-common.log-expected.json
SimonKoetting: I like the first idea, this way it seems clear to me that this is not a typical login. On the other side it's clearly visible to the user that these kinds of logins exist so he can decide whether to include or exclude it in reporting.
I updated the code to use the _ naming

muthu-mps (approved):
LGTM!

PR #12153, opened 2024-12-18
Update dga package to spec 3.0.4

jsoriano:
## Proposed commit message

Update DGA package to spec 3.0.4.

## Checklist


sodhikirti07 (approved):
Changes lgtm!

packages/dga/manifest.yml
```diff
@@ -1,7 +1,7 @@
-format_version: 2.2.0
+format_version: 3.0.4
```
sodhikirti07, line 1: Shouldn't this be `3.3.1`?

peteharverson (approved):
LGTM for ML

packages/dga/manifest.yml
jsoriano: I am being cautious as this package is intended to be used for versions of kibana since 8.9, see https://github.com/elastic/integrations/pull/12156#issuecomment-2553884820.

PR #12156, opened 2024-12-18
Update stack monitoring spec 3.x

jsoriano:
Migrate stack monitoring packages to spec 3.x.


consulthys:
@jsoriano Thanks for initiating this. Quick noob question: why did you use 3.0.4 instead of 3.3.1 as you commented [here](https://github.com/elastic/ingest-dev/issues/4484#issuecomment-2551112710)?

jsoriano:
> why did you use 3.0.4 instead of 3.3.1 as you commented [here](https://github.com/elastic/ingest-dev/issues/4484#issuecomment-2551112710)?

This package is intended for versions of kibana over 8.10.2. Kibana till 8.16 was filtering out packages up to 3.0 versions.
In summary, for packages supporting versions of kibana below 8.16, it is safer to use 3.0.

consulthys:
Thanks @jsoriano 
> Kibana till 8.16 was filtering out packages up to 3.0 versions.

Where can I learn about such Kibana-specific constraints?

packages/elasticsearch/data_stream/ccr/fields/fields.yml
```diff
@@ -32,7 +32,7 @@
       type: integer
       metric_type: gauge
     - name: read_exceptions.exception
-      type: object
+      type: flattened
```
consulthys, line 35: The reason for switching to `flattened` is because otherwise `object_type` would be required?

consulthys:
As a result of seeing the output of `elastic-package build`, all the changes make sense (except the `flattened` one I flagged)

jsoriano:
> > Kibana till 8.16 was filtering out packages up to 3.0 versions.
> 
> Where can I learn about such Kibana-specific constraints?

At the moment this is only in the default kibana configs. We should probably document this somewhere else, similar to what we have for kibana constraints in https://github.com/elastic/elastic-package/blob/main/docs/howto/stack_version_support.md#when-to-update-the-condition.

jsoriano:
Adding guidelines for spec version to https://github.com/elastic/elastic-package/pull/2291

packages/elasticsearch/data_stream/ccr/fields/fields.yml
consulthys: You can actually find a sample event containing some data for that field [here](https://github.com/elastic/beats/blob/main/metricbeat/module/elasticsearch/ccr/_meta/test/ccr_stats.700.json#L39-L56), which only [contains two subfields](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/monitoring-es-mb.json#L2108-L2129), one `keyword` and one `text`, which I guess rules out the `object_type` option altogether:
```
                      "read_exceptions": {
                        "type": "nested",
                        "properties": {
                          "exception": {
                            "type": "object",
                            "properties": {
                              "reason": {
                                "type": "text"
                              },
                              "type": {
                                "type": "keyword"
                              }
                            }
                          },
```

Also interesting is to see how that field is actually [used in Stack Monitoring](https://github.com/elastic/kibana/blob/main/x-pack/plugins/monitoring/server/lib/alerts/fetch_ccr_read_exceptions.ts#L41-L62), i.e. the query only checks for the existence of the `exceptions` object (which `flattened` also supports), and then the [`type` subfield is read](https://github.com/elastic/kibana/blob/main/x-pack/plugins/monitoring/server/routes/api/v1/elasticsearch/ccr.ts#L250).

I'll do some more checks to see if changing from `object` to `flattened` doesn't break anything (also in the case of monitoring indices not having the same mapping after rollover)

jsoriano: Well, these mappings with `nested` sound good too, I can try with them.

packages/elasticsearch/manifest.yml
```diff
@@ -1,19 +1,22 @@
...
-format_version: 2.6.0
+format_version: 3.0.4
 categories: ["elastic_stack", "datastore"]
 conditions:
-  elastic.subscription: basic
-  kibana.version: ^8.10.1
+  elastic:
+    subscription: basic
+  kibana:
+    version: ^8.11.2
```
jsoriano, line 17: This is the first safe version to use secrets. We can also delay the adoption of secrets, then I would revert the change in the kibana constraint, and set `secret: false` in the cases where it is set to true now.

consulthys: I think it should be fine bumping from 8.10.1 to 8.11.2

packages/elasticsearch/data_stream/ccr/fields/fields.yml
```diff
@@ -32,7 +32,19 @@
       type: integer
       metric_type: gauge
     - name: read_exceptions.exception
```
consulthys, line 34: Shouldn't this be just `read_exceptions` since `exception` is now a field of its own below?

packages/elasticsearch/docs/README.md
```diff
@@ -360,8 +360,11 @@ will not collect metrics. A DEBUG log message about this will be emitted in the
 | elasticsearch.ccr.leader.index | Name of leader index | keyword |  |
 | elasticsearch.ccr.leader.max_seq_no | Maximum sequence number of operation on the leader shard | long | gauge |
 | elasticsearch.ccr.read_exceptions |  | nested |  |
-| elasticsearch.ccr.read_exceptions.exception |  | object |  |
+| elasticsearch.ccr.read_exceptions.exception.exception.reason |  | text |  |
```
consulthys, line 363: I think the right field is already below `elasticsearch.ccr.read_exceptions.exception.reason`

packages/elasticsearch/docs/README.md
```diff
@@ -360,8 +360,11 @@ will not collect metrics. A DEBUG log message about this will be emitted in the
...
+| elasticsearch.ccr.read_exceptions.exception.exception.type |  | keyword |  |
```
consulthys, line 364: I think the right field is already below `elasticsearch.ccr.read_exceptions.exception.type`

packages/elasticsearch/docs/README.md
```diff
@@ -360,8 +360,11 @@ will not collect metrics. A DEBUG log message about this will be emitted in the
...
+| elasticsearch.ccr.read_exceptions.exception.from_seq_no |  | long |  |
```
consulthys, line 365: I think the right field is already below `elasticsearch.ccr.read_exceptions.from_seq_no`

packages/elasticsearch/docs/README.md
```diff
@@ -360,8 +360,11 @@ will not collect metrics. A DEBUG log message about this will be emitted in the
...
 | elasticsearch.ccr.read_exceptions.exception.reason |  | text |  |
+| elasticsearch.ccr.read_exceptions.exception.retries |  | integer |  |
```
consulthys, line 367: I think the right field is already below `elasticsearch.ccr.read_exceptions.retries`

packages/beat/manifest.yml
```diff
@@ -1,14 +1,17 @@
...
 title: Beat
-version: 1.0.0-beta1
+version: 1.0.0-beta2
 description: Beat Integration
 type: integration
 conditions:
-  kibana.version: ^8.10.2
-  elastic.subscription: basic
+  kibana:
+    version: ^8.10.2
```
consulthys, line 8: This should also probably be 8.11.2

jsoriano: I haven't updated the version here because this package doesn't make use of secrets, or afaik anything not available in 8.10.

packages/elasticsearch/data_stream/ccr/fields/fields.yml
jsoriano: You are right, good catch, thanks.

jsoriano: Ok, I hadn't noticed that there was already a definition from these fields. I think we only need to remove the definition of the `read_exceptions.exception` object.

jsoriano:
Elasticsearch migration to spec 3.0 is blocked by https://github.com/elastic/package-spec/issues/868 after the recent addition of transforms in the package.

PR #12158, opened 2024-12-18
[checkpoint_harmony_endpoint] Auth and pagination fixes

### LINKED ISSUE #12147: [Checkpoint Harmony Endpoint] - Add support for auto refresh of access tokens via CEL code
The Checkpoint Harmony Endpoint APIs do not support standard OAuth2, which is why once the access tokes expire after 30 mins, the integration starts getting 401 errors. To resolve this we need to rework the CEL code logic across all seven data streams to include a process of auto token refresh after a designated time period. 

Inspiration from a similar change linked [here](https://github.com/elastic/integrations/pull/11801/files#diff-cd4b9095da2b0bd3bbdaad1cd4c573bc036878094842c68149071f9c41f2fc86).

chrisberkhout:
## Proposed commit message

```
[checkpoint_harmony_endpoint] Auth and pagination fixes

Bug fixes:

- Select either a saved auth token (if it's not close to expiry) or a
  newly fetched one. Use it for any following request and save it in the
  cursor state. Clear the auth data after any data request that returns
  401. Don't clear the auth data at the end of a sequence of requests.
  Don't use the presence or absence of an auth token to decide which
  data request to make.

- Paginate correctly by advancing the `startTime` and `endTime` query
  parameters.

Other improvements and refactoring:

- Assume a successful response from the auth endpoint, rather explicitly
  trying to set the auth token from the body of an error response.
- Don't try to read `body.message` from an error response, since error
  responses don't have a `message` key.
- Remove the unused `last_page` key from cursor data.
- Remove the redundant `task_ready` key from cursor data and just use
  `page_token` instead. Don't override `page_token` to `null` when it's
  already `null`.
- Use optional access to cursor values for shorter conditions.
- Improve clarity in CEL comments and order of handling.

- Extend the default `interval` and improve the variable descriptions
  for `interval` and `initial_interval`.
- Add the missing definition for one set of rate limiting variables,
  give all definitions default values, and use those variables in the
  input config.
- Note the minimum `page_limit` value in that variable's description.

- System test just one data stream, but make it a better test that
  covers polling and pagination.

Notes:

- All data streams have identical `cel.yml.hbs` files.
- This has been manually tested against a live system.
- API doc: https://app.swaggerhub.com/apis-docs/Check-Point/infinity-events-api/1.0.0
```

## Tasks to complete the draft

- [x] Handle 401s for any request, not just the query submission
- [x] Save auth_token expiry information in the cursor, and use to decide when to get a new token
      Example auth response: `{"success":true,"data":{"token":"cGibheyJ...","csrf":"a69fe9ac-3764-436c-a254-812a443a6b8d","expires":"Wed, 18 Dec 2024 17:28:31 GMT","expiresIn":1800}}`
- [x] Check whether the system test data needs to be updated.
- [x] Update the proposed commit message with details of the additional changes.

## Review tips

Remember that all the `cel.yml.hbs` files are identical.

Can be read commit-by-commit, but it's probably best to focus on the bugfix-related logic in the final version.

Turning off whitespace changes in Github or another diff viewer can help a lot, as many lines have indentation changes.

## Checklist

## Related issues

- Closes #12147


chrisberkhout:
Coverage reduction is because I removed all but one of the system tests. Those tests weren't great and were just cut and pasted versions of each other, for data streams that all have identical input config.

ShourieG (approved):
LGTM, tested manually against live endpoint to cross-check proper operation.

PR #12161, opened 2024-12-19
Fix handling of lists in 'Set ECS categorization fields' scripts

joegallo:
Defensively copy list parameters in 'Set ECS categorization fields' scripts. See https://github.com/elastic/elasticsearch/issues/91977.

### DIFF
--- a/packages/authentik/changelog.yml
+++ b/packages/authentik/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.2.1"
+  changes:
+    - description: Defensively copy list parameters in 'Set ECS categorization fields' script.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/12161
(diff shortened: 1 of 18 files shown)


joegallo:
These `def hm = new HashMap(...)` bits are trying to work around the linked bug, but I think they're mistaken.

marc-gr (approved):

andrewkroh (approved):
@joegallo Is there any means of performing an automated escape analysis on these scripts? I'd love to be able to put an static check into our package test tooling.

joegallo:
> @joegallo Is there any means of performing an automated escape analysis on these scripts? I'd love to be able to put a static check into our package test tooling.

There's nothing built into Elasticsearch for that yet, but I'm going to think about it -- I realize that the current approach of 'just be careful/perfect all the time' is untenable.

Issue #12118, opened 2024-12-16
Support the first 80-90 event types for the Aruba integration


qcorporation:
## Parent Ticket:
https://github.com/elastic/integrations/issues/5255

## Description:

1. Proxy ARP events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/PROXY-ARP.htm
2. QoS ASIC Provider events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/QOS_ASIC.htm
3. Quality of Service events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/QOS.htm
4. Rapid per VLAN Spanning Tree Protocol events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/RPVST.htm
5. RBAC events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/RBACD.htm
6. Redundant Management events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/REDUNDANT_MANAGEMENT.htm
7. Replication Manager events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/REPLD.htm
8. REST events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/RESTD.htm
9. Self Test events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/SELFTEST.htm
10. sFlow events: https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-8214/Content/events/SFLOW.htm

PR #12184, opened 2024-12-23
#12118 8/9/10 Added 3 new events types (sFlow, Self Test and REST)

qcorporation:
## Parent Ticket:
https://github.com/elastic/integrations/issues/12118

## Change Log:
- sFlow events (10xx)
- Self Test events (450x)
- REST events (46xx)

Note: this is based on OS 10.15 message types different than 10.07. Will allow the rest of the already mapped fields in another PR

### DIFF
--- a/packages/hpe_aruba_cx/data_stream/log/_dev/test/pipeline/test-aruba-cx.log-expected.json
+++ b/packages/hpe_aruba_cx/data_stream/log/_dev/test/pipeline/test-aruba-cx.log-expected.json
@@ -400,6 +400,9 @@
                 "event_type": "Event",
                 "hardware": {
                     "device": "6000RCSE4802"
+                },
+                "rest": {
+                    "central_source": "activate"
                 }
             },
             "ecs": {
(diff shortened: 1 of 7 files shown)


packages/hpe_aruba_cx/data_stream/log/_dev/test/pipeline/test-aruba-cx.log-expected.json
```diff
@@ -4429,6 +4448,9 @@
                 "event_type": "Event",
                 "hardware": {
                     "device": "8360-Primaire"
+                },
+                "session": {
+                    "id": "-ZWxhc3RpYy5jby9kb2Nz"
```
dwhyrock, line 4453: This value seems a bit odd, especially since the same value is used in a subsequent entry, but without the leading `-`:

```
                    "id": "ZWxhc3RpYy5jby9kb2Nz"
```

dwhyrock (commented):
minor question about a result ID string that looks a bit odd

dwhyrock (approved):

dwhyrock:
Looks like some merge conflicts, FYI

PR #12202, opened 2024-12-26
Add support for fraction-of-second in Cisco-standard date/time strings.

jrmolin:
## Proposed commit message
[Cisco ASA/ISE] Time parsing did not handle fractions of a second
* Add support for Cisco ASA custom format parsing
* Add support for Cisco ISE custom format parsing
* Verify no existing dates are changed
* Verify with new example log samples

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/cisco_asa/data_stream/log/_dev/test/pipeline/test-asa-fix.log-expected.json
```diff
@@ -1173,6 +1173,158 @@
...
+                "address": "10.0.0.46",
+                "ip": "10.0.0.46",
+                "port": 49734
+            },
+            "tags": [
+                "preserve_original_event"
+            ]
+        },
+        {
+            "@timestamp": "2024-04-29T12:59:50.000Z",
```
mjwolf, line 1249: The fractions of seconds from the original log aren't represented here. Should they be?

mjwolf (approved):
LGTM, I'm not sure if it should parse the fractions of seconds into `@timestamp`, but since this fixes the parsing errors, I don't want to hold it up for that

packages/cisco_asa/data_stream/log/_dev/test/pipeline/test-asa-fix.log-expected.json
jrmolin: hrm, i missed that. i'll look into it. i think it should be represented there

taylor-swanson: I think what's happening here is `@timestamp` is a dynamic field (see test-common-config.yml). As a result, it won't update or validate the `@timestamp` field in the expected results. You can temporarily remove it as a dynamic field, regen the expected files, then re-add it to dynamic_fields. I think one of the reasons why we have it as a dynamic field is because some of the timestamps in the logs don't have a year on it.

jrmolin:
The issue was closed, and it is unclear if this is needed anywhere or it was a misconfiguration. Closing the PR.

PR #12206, opened 2024-12-27
entityanalytics_okta: Add `event.original` field when `preserve_original_event` tag is set.

kcreddy:
## Proposed commit message
Okta Entity Analytics doesn't capture full event inside raw `message` field. Add a painless script to capture the original event into `event.original` using `Preserve original event` UI option.

## Checklist

Running Pipeline tests should pass successfully.
Run -> `cd packages/entityanalytics_okta && elastic-package stack down && elastic-package build && elastic-package stack up --version=8.13.0 -d -v && eval "$(elastic-package stack shellinit)" && elastic-package test pipeline --generate -v`
Returns -> 
```
--- Test results for package: entityanalytics_okta - START ---
╭──────────────────────┬─────────────┬───────────┬───────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE              │ DATA STREAM │ TEST TYPE │ TEST NAME                                 │ RESULT │ TIME ELAPSED │
├──────────────────────┼─────────────┼───────────┼───────────────────────────────────────────┼────────┼──────────────┤
│ entityanalytics_okta │ user        │ pipeline  │ (ingest pipeline warnings test-user.json) │ PASS   │ 321.542792ms │
│ entityanalytics_okta │ user        │ pipeline  │ test-user.json                            │ PASS   │  51.258333ms │
╰──────────────────────┴─────────────┴───────────┴───────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: entityanalytics_okta - END   ---
Done
```


packages/entityanalytics_okta/data_stream/user/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -5,6 +5,21 @@ processors:
...
       tag: set_ecs_version
       value: 8.11.0
+  - script:
+      tag: script_add_event_original
+      lang: painless
+      description: Add "event.original"
+      if: ctx.tags != null && ctx.tags.contains('preserve_original_event')
+      source: |
+        def orig = ctx;
+        orig.entrySet().removeIf(entry -> ['_index', '_id'].contains(entry.getKey()));
```
chrisberkhout, line 15: After `def org = ctx;`, `org` has the same reference as `ctx`, so `_index` and `_id` get removed from `ctx` as well.

Pipeline tests won't accept data with `_id` or `_index` (the simulate API rejects it), if you run this logic by modifying the system test as follows, it will show index/publishing failures.

```diff
diff --git a/packages/entityanalytics_okta/data_stream/user/_dev/test/system/test-default-config.yml b/packages/entityanalytics_okta/data_stream/user/_dev/test/system/test-default-config.yml
index 8db4e54f64..0ec78b6d3c 100644
--- a/packages/entityanalytics_okta/data_stream/user/_dev/test/system/test-default-config.yml
+++ b/packages/entityanalytics_okta/data_stream/user/_dev/test/system/test-default-config.yml
@@ -5,6 +5,7 @@ data_stream:
     okta_domain: trial-xxxxxxx-admin.okta.com
     okta_token: xxxx
     preserve_duplicate_custom_fields: true
+    preserve_original_event: true
     enable_request_tracer: true
     ssl: |
       certificate_authorities:
@@ -42,3 +43,5 @@ data_stream:
           JE3WRcwNIONSRY/9seTyK6y9ed61m40Bx92kO2H0Ld+vKkYnFWAoZNYEM0WZ0zWz
           9EkWoxHOkvz1Pnp41Llpc7OPdbOQrx4=
           -----END CERTIFICATE-----
+assert:
+  hit_count: 4
```

chrisberkhout (changes requested):

packages/entityanalytics_okta/data_stream/user/elasticsearch/ingest_pipeline/default.yml
kcreddy: Maybe I was on a different stack version earlier. When I ran on `8.15` (kibana.version), I got them successful locally without having to delete `_id` and `_index`.

packages/entityanalytics_okta/data_stream/user/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -5,6 +5,20 @@ processors:
...
       tag: set_ecs_version
       value: 8.11.0
+  - script:
+      tag: script_add_event_original
+      lang: painless
+      description: Add "event.original"
+      if: ctx.tags != null && ctx.tags.contains('preserve_original_event')
+      source: |
+        def orig = ctx;
+        def stringified_orig = Json.dump(orig);
```
chrisberkhout, line 15:
```suggestion
        def stringified_orig = Json.dump(ctx);
```
`orig` refers to the same object, so we might as well use it directly.

chrisberkhout (approved):
Good, with a new `assert.hit_count` change.

packages/entityanalytics_okta/data_stream/user/_dev/test/system/test-default-config.yml
```diff
@@ -42,3 +43,5 @@ data_stream:
           JE3WRcwNIONSRY/9seTyK6y9ed61m40Bx92kO2H0Ld+vKkYnFWAoZNYEM0WZ0zWz
           9EkWoxHOkvz1Pnp41Llpc7OPdbOQrx4=
           -----END CERTIFICATE-----
+assert:
+  hit_count: 3 # 1 start marker + 1 user + 1 end marker
```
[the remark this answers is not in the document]
kcreddy: Got it, thanks for this analysis. I didn't run many test runs, and I was going off by system tests' mock records. 

I wonder if we shouldn't allow `incremental update` to start until 1 full sync is finished.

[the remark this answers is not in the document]
chrisberkhout: Yeah, probably.

At the moment when it starts up it sets the next times based on what's in the store. If it's a fresh start it will end up with both due and then choose randomly, but if its resuming it will continue on the previous schedule.

So the change might be: if times from the store aren't in the future, set do the full sync now and the incremental after its interval.

PR #12218, opened 2025-01-02
[O365_metrics] Add  teams_device_usage_user_counts data stream.

ritalwar:
- Enhancement

## Proposed commit message

Add teams_device_usage_user_counts data stream to O365_metrics package.

## Checklist


packages/o365_metrics/data_stream/teams_device_usage/manifest.yml
```diff
@@ -0,0 +1,48 @@
...
+        required: true
+        default: D7
+      - name: tags
+        type: text
+        title: Tags
+        multi: true
+        required: false
+        show_user: false
+        default:
+          - o365.metrics.outlook.activity
```
lucian-ioan, line 40: Rename tag.

packages/o365_metrics/data_stream/teams_device_usage/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,82 @@
+---
+description: Pipeline for renaming object
+processors:
+  - set:
+      field: ecs.version
+      value: "8.16.0"
+  - json:
+      field: teamsdeviceusage
+      target_field: o365.metrics.teams.device.usage
```
lucian-ioan, line 9: Maybe add `ignore_failure: true` here so the pipeline doesn't break.

packages/o365_metrics/data_stream/teams_device_usage/fields/fields.yml
```diff
@@ -0,0 +1,43 @@
+- name: o365.metrics.teams.device.usage
+  type: group
+  fields:
+    - name: android_phone.count
+      type: integer
```
agithomas, line 5: Can you please verify if the `type: integer` lead to any limitations in the value range. `2**31-1` is not a big value range. Please verify for the other `integer` fields.

packages/o365_metrics/_dev/build/docs/README.md
```diff
@@ -82,4 +82,14 @@ Please refer to the following [document](https://www.elastic.co/guide/en/ecs/cur
 
 Please refer to the following [document](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) for detailed information on ECS fields.
 
-{{fields "active_users"}}
\ No newline at end of file
+{{fields "active_users"}}
+
+### Teams Device Usage
+
```
agithomas, line 88: Suggestion: It will be good to add a line or two about what the dataset is for.

agithomas:
I assume the TSDS enablement will be done through a separate PR. Please remember to add metric_type mapping, TSDS enablement without fail.

packages/o365_metrics/data_stream/teams_device_usage/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,82 @@
...
+      ignore_missing: true
+  - rename:
+      field: o365.metrics.teams.device.usage.windows
+      target_field: o365.metrics.teams.device.usage.windows.count
+      ignore_missing: true
+  - rename:
+      field: o365.metrics.teams.device.usage.windows_phone
+      target_field: o365.metrics.teams.device.usage.windows_phone.count
+      ignore_missing: true      
+  - rename:
```
shmsr, line 56: for this and refresh_date, are we going to use the `date` processor to process?

ritalwar: From my understanding, we use the `date` processor to parse dates from fields and then use the parsed date or timestamp as the document's timestamp in Elasticsearch. However, we cannot use `report_date` or `report_refresh_date` as the document timestamp because these fields are generated by Microsoft and do not represent the actual ingestion time of the document in Elasticsearch. 

For example, if a report is fetched today (January 10) to retrieve the last 7 days of data, the `report_refresh_date` might be January 7, which would be the same for all 7 documents. Meanwhile, the `report_date` for these 7 documents would range from January 1 to January 7. This means neither `report_date` nor `report_refresh_date` can reliably represent the document's ingestion timestamp. 

Please let me know if I’ve misunderstood something or if your question was different.

shmsr: Okay I understand now. But suppose when you will make the visualization, what are you going to use as the timestamp? Suppose, the user wants to use the visualization (or for aggregation purpose) then we cannot use the ingestion time, right. Are you going to manually change the timestamp to these fields when plotting?

packages/o365_metrics/_dev/build/docs/README.md
ritalwar: I’ve added this to the [issue](https://github.com/elastic/obs-integration-team/issues/156) list to update it for all datasets collectively in a single documentation PR.

packages/o365_metrics/data_stream/teams_device_usage/fields/fields.yml
ritalwar: Since I couldn't find any specific details about these fields in the API documentation and we're only retrieving 7 days of data, it should not exceed the limit. However, I agree that I should verify this, not just for this field but for all fields across all data streams. Additionally, using `long` might indeed be a better choice here. I've added this to the [issue](https://github.com/elastic/obs-integration-team/issues/156) where we're tracking all review comments for verification across all data streams. I'll address this in a separate PR covering all the data streams. Does that sound okay?

packages/o365_metrics/data_stream/teams_device_usage/elasticsearch/ingest_pipeline/default.yml
ritalwar: Good question! I assume they can use `report_date` if they want the visualizations to reflect the actual date the data pertains to, or `report_refresh_date` if the visualizations should represent the date the report was last updated or refreshed. What are your thoughts?

agithomas: As part of the integration testing, do you see any latency in the availability of metrics? 

For example : If you run the integration for the current date, the data is available only till yesterday. 

If yes, it would be best to mention this latency in the README.

ritalwar: In the integration testing, I’ve set the API call interval to 3 minutes, and during this period, I do see data ingestion occurring every 3 min. However, this short interval is just for testing purposes. In a production environment, a 3-minute interval is not practical. Ideally, the interval should be at least 24 hours, as the Microsoft API only fetches the latest report, and reports are usually refreshed every 3-4 days. So, even if we call the API every 24 hours, the records might remain the same.

But I agree that updating these details in the README is important, especially since we're receiving data for like 7 days here, so users can configure accordingly for both period and interval. While I assume microsoft users should already be aware of how this works, I will ensure the README is detailed for clarity.

packages/o365_metrics/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.4.0-next"
+  changes:
+    - description: Add `teams_device_usage` data stream.
```
muthu-mps, line 4:
```suggestion
    - description: Add `teams_device_usage_user_counts` data stream.
```

packages/o365_metrics/data_stream/teams_device_usage_user_counts/manifest.yml
```diff
@@ -0,0 +1,56 @@
+title: Microsoft Office 365 Teams Device Usage metrics
+type: metrics
+streams:
+  - input: cel
+    title: Office 365 Teams Device Usage metrics.
```
muthu-mps, line 5:
```suggestion
    title: Office 365 Teams Device Usage User Counts metrics.
```

packages/o365_metrics/data_stream/teams_device_usage_user_counts/manifest.yml
```diff
@@ -0,0 +1,56 @@
...
+    enabled: true
+    description: Collect Office 365 Teams Device Usage metrics.
```
muthu-mps, line 7:
```suggestion
    description: Collect Office 365 Teams Device Usage User Counts metrics.
```

muthu-mps (approved):
LGTM!

packages/o365_metrics/data_stream/teams_device_usage_user_counts/manifest.yml
```diff
@@ -0,0 +1,56 @@
...
+  - input: cel
+    title: Office 365 Teams Device Usage User Counts metrics.
+    enabled: true
+    description: Collect Office 365 Teams Device Usage User Counts metrics.
+    template_path: cel.yml.hbs
+    vars:
+      - name: interval
+        type: text
+        title: Interval
+        description: The interval at which the API is polled, supported in seconds, minutes, and hours.
```
lucian-ioan, line 13: As per https://github.com/elastic/obs-integration-team/issues/156

```suggestion
        description: Interval of API polling. Interval supported in seconds, minutes and hours.
```

packages/o365_metrics/data_stream/teams_device_usage_user_counts/sample_event.json
```diff
@@ -0,0 +1,96 @@
...
+    "o365": {
+      "metrics": {
+        "teams": {
+          "device": {
+            "usage": {
+              "user": {
+                "counts": {
+                  "windows_phone": {
+                    "count": 0
+                  },
```
lucian-ioan, line 11: While we might not have real data for most of these, I would personally replace the zeros with some arbitrary values.

lucian-ioan (approved):
Nits, otherwise LGTM.

packages/o365_metrics/data_stream/teams_device_usage_user_counts/manifest.yml
ritalwar: I would prefer to stick with what's already in place.

PR #12219, opened 2025-01-02
[O365_metrics] Add teams_user_activity datastream

ritalwar:
- Enhancement

## Proposed commit message

Add teams_user_activity datastream to O365_metrics package.

## Checklist

### DIFF
--- a/packages/o365_metrics/_dev/build/docs/README.md
+++ b/packages/o365_metrics/_dev/build/docs/README.md
@@ -83,3 +83,13 @@ Please refer to the following [document](https://www.elastic.co/guide/en/ecs/cur
 Please refer to the following [document](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) for detailed information on ECS fields.
 
 {{fields "active_users"}}
+
+### Teams User Activity
+
+{{event "teams_user_activity"}}
+
+**ECS Field Reference**
+
+Please refer to the following [document](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) for detailed information on ECS fields.
+
+{{fields "teams_user_activity"}}
--- a/packages/o365_metrics/data_stream/teams_user_activity/_dev/test/pipeline/test-metrics.json-expected.json
+++ b/packages/o365_metrics/data_stream/teams_user_activity/_dev/test/pipeline/test-metrics.json-expected.json
@@ -1,0 +1,41 @@
+{
+    "expected": [
+        {
+            "ecs": {
+                "version": "8.16.0"
+            },
+            "o365": {
+                "metrics": {
+                    "teams": {
+                        "user": {
+                            "activity": {
+                                "calls": {
+                                    "count": "10"
+                                },
+                                "meetings": {
+                                    "count": "8"
+                                },
+                                "other_actions": {
+                                    "count": "7"
+                                },
+                                "private_chat_messages": {
+                                    "count": "15"
+                                },
+                                "report": {
+                                    "date": "2024-12-25",
+                                    "period": {
+                                        "day": "7"
+                                    },
+                                    "refresh_date": "2024-12-31"
+                                },
+                                "team_chat_messages": {
+                                    "count": "9"
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    ]
+}
--- a/packages/o365_metrics/data_stream/teams_user_activity/agent/stream/cel.yml.hbs
+++ b/packages/o365_metrics/data_stream/teams_user_activity/agent/stream/cel.yml.hbs
@@ -1,0 +1,84 @@
+config_version: 2
+interval: {{interval}}
+auth.oauth2:
+    client.id: {{client_id}}
+    client.secret: {{client_secret}}
+    provider: azure
+    scopes:
+{{#each token_scopes as |token_scope|}}
+      - {{token_scope}}
+{{/each}}
+    endpoint_params: 
+        grant_type: client_credentials
+{{#if token_url}}
+    token_url: {{token_url}}/{{azure_tenant_id}}/oauth2/v2.0/token
+{{else if azure_tenant_id}}
+    azure.tenant_id: {{azure_tenant_id}}
+{{/if}}
+
+resource.url: {{url}}
+{{#if resource_ssl}}
+resource.ssl: 
+  {{resource_ssl}}
+{{/if}}
+
+{{#if enable_request_tracer}}
+resource.tracer.filename: "../../logs/cel/http-request-trace-*.ndjson"
+{{/if}}
+
+tags:
+{{#if preserve_original_event}}
+  - preserve_original_event
+{{/if}}
+{{#each tags as |tag|}}
+  - {{tag}}
+{{/each}}
+{{#contains "forwarded" tags}}
+publisher_pipeline.disable_host: true
+{{/contains}}
+{{#if processors}}
+processors:
+{{processors}}
+{{/if}}
+
+state:
+  want_more: false
+  base:
+    tenant_id: "{{azure_tenant_id}}"
+    period: "{{period}}"
+
+redact:
+  fields:
+    - base.tenant_id
+
+
+program: |
+    state.with(
+      request(
+        "GET",
+        state.url + "/reports/getTeamsUserActivityUserCounts(period='" + state.base.period + "')"
+      ).do_request().as(resp,
+        resp.StatusCode == 200
+        ?
+          bytes(resp.Body).mime("text/csv; header=present").as(events, {
+                                     "events": events.map(e, {"teamsuseractivity": e.encode_json()}),                     
+        
+          })
+        :
+          {
+            "events": {
+              "error": {
+                "code": string(resp.StatusCode),
+                "id": string(resp.Status),
+                "message": "GET:"+(
+                  size(resp.Body) != 0 ?
+                    string(resp.Body)
+                  :
+                    string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
+                ),
+              },
+            },
+            "want_more": false,
+          }
+      )
+    )
--- a/packages/o365_metrics/data_stream/teams_user_activity/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/o365_metrics/data_stream/teams_user_activity/elasticsearch/ingest_pipeline/default.yml
@@ -1,0 +1,72 @@
+---
+description: Pipeline for renaming object
+processors:
+  - set:
+      field: ecs.version
+      value: "8.16.0"
+  - json:
+      field: teamsuseractivity
+      target_field: o365.metrics.teams.user.activity
+  - script:
+      lang: painless
+      description: Replace spaces and dashes in field names under o365.metrics.teams.user.activity.
+      tag: painless_purge_spaces_and_dashes
+      if: ctx.o365.metrics?.teams.user.activity instanceof Map
+      source: |
+        String underscore(String s) {
+          String result = /[ -]/.matcher(s).replaceAll('_').toLowerCase();
+          return /[\ufeff]/.matcher(result).replaceAll('')
+        }
+        
+        def out = [:];
+        for (def item : ctx.o365.metrics.teams.user.activity.entrySet()) {
+          out[underscore(item.getKey())] = item.getValue();
+        }
+        ctx.o365.metrics.teams.user.activity = out;
+
+  - remove:
+      if: ctx.teamsuseractivity != null
+      field: teamsuseractivity
+      ignore_missing: true
+
+  - rename:
+      field: o365.metrics.teams.user.activity.calls
+      target_field: o365.metrics.teams.user.activity.calls.count
+      ignore_missing: true
+  - rename:
+      field: o365.metrics.teams.user.activity.meetings
+      target_field: o365.metrics.teams.user.activity.meetings.count
+      ignore_missing: true
+  - rename:
+      field: o365.metrics.teams.user.activity.other_actions
+      target_field: o365.metrics.teams.user.activity.other_actions.count
+      ignore_missing: true
+  - rename:
+      field: o365.metrics.teams.user.activity.private_chat_messages
+      target_field: o365.metrics.teams.user.activity.private_chat_messages.count
+      ignore_missing: true
+  - rename:
+      field: o365.metrics.teams.user.activity.team_chat_messages
+      target_field: o365.metrics.teams.user.activity.team_chat_messages.count
+      ignore_missing: true
+  - rename:
+      field: o365.metrics.teams.user.activity.report_date
+      target_field: o365.metrics.teams.user.activity.report.date
+      ignore_missing: true
+  - rename:
+      field: o365.metrics.teams.user.activity.report_period
+      target_field: o365.metrics.teams.user.activity.report.period.day
+      ignore_missing: true
+  - rename:
+      field: o365.metrics.teams.user.activity.report_refresh_date
+      target_field: o365.metrics.teams.user.activity.report.refresh_date
+      ignore_missing: true
+
+on_failure:
+  - append:
+      field: error.message
+      value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - append:
+      field: event.kind
+      value: pipeline_error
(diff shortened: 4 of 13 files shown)


packages/o365_metrics/data_stream/teams_user_activity/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,72 @@
+---
+description: Pipeline for renaming object
+processors:
+  - set:
+      field: ecs.version
+      value: "8.16.0"
+  - json:
+      field: teamsuseractivity
+      target_field: o365.metrics.teams.user.activity
```
lucian-ioan, line 9: Maybe add `ignore_failure: true` here so the pipeline doesn't break.

packages/o365_metrics/_dev/build/docs/README.md
```diff
@@ -82,4 +82,14 @@ Please refer to the following [document](https://www.elastic.co/guide/en/ecs/cur
 
 Please refer to the following [document](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) for detailed information on ECS fields.
 
-{{fields "active_users"}}
\ No newline at end of file
+{{fields "active_users"}}
+
+### Teams User Activity
+
```
agithomas, line 88: Suggestion: Please add a one or two-line summary of what the dataset metrics are about.

packages/o365_metrics/data_stream/teams_user_activity/fields/fields.yml
```diff
@@ -0,0 +1,39 @@
...
+        The number of calls made by Teams users.
+    - name: meetings.count
+      type: integer
+      description: |
+        The number of meetings attended or organized by Teams users.
+    - name: other_actions.count
+      type: integer
+      description: |
+        The count of other user actions within Teams.
+    - name: private_chat_messages.count
```
agithomas, line 16: As with the other datasets, kindly verify if the 2**31 value length is enough to hold the value received? If not enough, kindly choose a different field type.

packages/o365_metrics/data_stream/teams_user_activity/fields/fields.yml
```diff
@@ -0,0 +1,39 @@
...
+        - name: period.day
+          unit: d
+          type: integer
+          description: |
+            The duration (e.g., 7 days) over which the report data is aggregated.
+        - name: refresh_date
+          type: date
+          description: |
+            The date when the report data was last updated.
+        - name: date
```
agithomas, line 36: Is this different from the `@timestamp` field value ?

packages/o365_metrics/data_stream/teams_user_activity/fields/fields.yml
```diff
@@ -0,0 +1,39 @@
...
+      description: |
+        The number of messages sent in private 1:1 or group chats.
+    - name: team_chat_messages.count
+      type: integer
+      description: |
+        The number of messages sent in Teams channels.
+    - name: report
+      type: group
+      fields:
+        - name: period.day
```
agithomas, line 27: Is this value same as the `period` configuration value?

packages/o365_metrics/data_stream/teams_user_activity/fields/fields.yml
shmsr: `long` would be a safe or an unsigned type would be a safe choice in general.

packages/o365_metrics/data_stream/teams_user_activity/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,72 @@
...
+      field: o365.metrics.teams.user.activity.private_chat_messages
+      target_field: o365.metrics.teams.user.activity.private_chat_messages.count
+      ignore_missing: true
+  - rename:
+      field: o365.metrics.teams.user.activity.team_chat_messages
+      target_field: o365.metrics.teams.user.activity.team_chat_messages.count
+      ignore_missing: true
+  - rename:
+      field: o365.metrics.teams.user.activity.report_date
+      target_field: o365.metrics.teams.user.activity.report.date
```
shmsr, line 54: should we do a explicit date handling based on type using the `date` processor?

packages/o365_metrics/data_stream/teams_user_activity/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,72 @@
...
+      target_field: o365.metrics.teams.user.activity
+  - script:
+      lang: painless
+      description: Replace spaces and dashes in field names under o365.metrics.teams.user.activity.
+      tag: painless_purge_spaces_and_dashes
+      if: ctx.o365.metrics?.teams.user.activity instanceof Map
+      source: |
+        String underscore(String s) {
+          String result = /[ -]/.matcher(s).replaceAll('_').toLowerCase();
+          return /[\ufeff]/.matcher(result).replaceAll('')
```
shmsr, line 18: is this bom char coming?

packages/o365_metrics/data_stream/teams_user_activity/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,72 @@
...
+processors:
+  - set:
+      field: ecs.version
+      value: "8.16.0"
+  - json:
+      field: teamsuseractivity
+      target_field: o365.metrics.teams.user.activity
+  - script:
+      lang: painless
+      description: Replace spaces and dashes in field names under o365.metrics.teams.user.activity.
```
shmsr, line 12: no mention of bom removal?

packages/o365_metrics/data_stream/teams_user_activity/fields/fields.yml
ritalwar: https://github.com/elastic/integrations/pull/12218#discussion_r1909907622

packages/o365_metrics/data_stream/teams_user_activity/fields/fields.yml
ritalwar: Yes, details in this [comment](https://github.com/elastic/integrations/pull/12218#discussion_r1909897911).

packages/o365_metrics/data_stream/teams_user_activity/fields/fields.yml
ritalwar: Yes, we can configure values such as D7, D30, etc., in the period, and it will display 7 and 30 accordingly.

packages/o365_metrics/data_stream/teams_user_activity/elasticsearch/ingest_pipeline/default.yml
ritalwar: Yes, there is a BOM character present in some fields, and I should update the pipeline comment to specify this as well.

packages/o365_metrics/data_stream/teams_user_activity/elasticsearch/ingest_pipeline/default.yml
ritalwar: I will mention, thanks!

packages/o365_metrics/_dev/build/docs/README.md
ritalwar: https://github.com/elastic/integrations/pull/12218#discussion_r1909907114

packages/o365_metrics/data_stream/teams_user_activity/sample_event.json
```diff
@@ -0,0 +1,83 @@
+{
+    "o365": {
+      "metrics": {
```
muthu-mps, line 3: @ritalwar - Query related to the metrics association. Is there a way the teams user activity metrics could be associated with the `user-id or name` to understand from where the request originates. If the user enables this specific data stream for user activity, how do they identify for which user this metrics is related to?

muthu-mps: Is this a limitation in API response itself?

ritalwar: Hi @muthu-mps , this data stream is primarily designed to retrieve the number of Microsoft Teams users categorized by activity type. The activity types include team chat messages, private chat messages, calls, and meetings. Refer to the documentation [here](https://learn.microsoft.com/en-us/graph/api/reportroot-getteamsuseractivityusercounts?view=graph-rest-1.0&tabs=http). 

I'll add such links to the README for each dataset to prevent similar confusion.

muthu-mps: When the API says user activity, I thought somehow they will link user to the activity type but this seems to be overall activity counts and not specific to a user.

kcreddy (commented):
@ritalwar, is this data-stream going to be renamed to `teams_user_activity_user_counts`? I am working on `teams_user_activity_user_detail`

I think there was an agreement on creating data-streams `viva_engage_device_usage_user_counts` and `viva_engage_groups_activity_group_details`. I think we should continue the same norm here.

ritalwar:
> @ritalwar, is this data-stream going to be renamed to `teams_user_activity_user_counts`? I am working on `teams_user_activity_user_detail` [here](https://github.com/elastic/integrations/pull/12256/files).

Yes, Krishna, this should be renamed as we discussed. I will take care of updating it.

> There was an agreement on creating data-streams `viva_engage_device_usage_user_counts` and `viva_engage_groups_activity_group_details`. I think we should continue the same norm here to avoid confusion for the users.

Agree! I missed renaming `yammer_device_usage`, but I'll update it in a separate PR.

packages/o365_metrics/data_stream/teams_user_activity/_dev/test/pipeline/test-metrics.json-expected.json
```diff
@@ -0,0 +1,41 @@
...
+                                    "count": "10"
+                                },
+                                "meetings": {
+                                    "count": "8"
+                                },
+                                "other_actions": {
+                                    "count": "7"
+                                },
+                                "private_chat_messages": {
+                                    "count": "15"
```
kcreddy, line 22: The `*.count` fields are `keyword` here. I think you could convert them explicitly to numbers.

packages/o365_metrics/data_stream/teams_user_activity/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,84 @@
...
+program: |
+    state.with(
+      request(
+        "GET",
+        state.url + "/reports/getTeamsUserActivityUserCounts(period='" + state.base.period + "')"
+      ).do_request().as(resp,
+        resp.StatusCode == 200
+        ?
+          bytes(resp.Body).mime("text/csv; header=present").as(events, {
+                                     "events": events.map(e, {"teamsuseractivity": e.encode_json()}),
```
kcreddy, line 64: You could use here [`message`](https://github.com/elastic/integrations/blob/main/packages/mimecast/data_stream/siem_logs/agent/stream/cel.yml.hbs#L97) field (that is more generic across other integrations so that ingest pipeline can be handle [`event.original`](https://github.com/elastic/integrations/blob/main/packages/apache/data_stream/access/elasticsearch/ingest_pipeline/default.yml#L13-L22). 

Right now, is there a way to preserve original event?

packages/o365_metrics/data_stream/teams_user_activity/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,84 @@
...
+resource.ssl: 
+  {{resource_ssl}}
+{{/if}}
+
+{{#if enable_request_tracer}}
+resource.tracer.filename: "../../logs/cel/http-request-trace-*.ndjson"
+{{/if}}
+
+tags:
+{{#if preserve_original_event}}
```
kcreddy, line 30: I don't see this option in `manifest.yml`

packages/o365_metrics/data_stream/teams_user_activity/_dev/test/pipeline/test-metrics.json-expected.json
```diff
@@ -0,0 +1,41 @@
+{
+    "expected": [
+        {
+            "ecs": {
```
kcreddy, line 4: The test result is missing `@timestamp`. I don't see it being set inside ingest pipeline. 
Shouldn't this be copied from `report.date`?

kcreddy (commented):
Sorry I haven't checked other reviews. 
Please ignore my comment if the same is discussed elsewhere.

packages/o365_metrics/data_stream/teams_user_activity/elasticsearch/ingest_pipeline/default.yml
ritalwar: https://github.com/elastic/integrations/pull/12218#discussion_r1909897911

packages/o365_metrics/data_stream/teams_user_activity/agent/stream/cel.yml.hbs
ritalwar: Did it in a few pipelines but will update for other data streams as well to preserve the original event using:
 ``` - set:
      copy_from: teamsuseractivity
      field: event.original
      ignore_empty_value: true
      if: ctx.event?.original == null && (ctx.tags != null && (ctx.tags.contains('preserve_original_event')))
```

File: packages/o365_metrics/data_stream/teams_user_activity/agent/stream/cel.yml.hbs (hunk 12 above)
Comment (ritalwar) [reply packages/o365_metrics/data_stream/teams_user_activity/agent/stream/cel.yml.hbs:30]: Yes, I will update this, missed it in a few datastreams and tracking it [here](https://github.com/elastic/obs-integration-team/issues/156) to review it again for the merged datastreams as well.

File: packages/o365_metrics/data_stream/teams_user_activity/_dev/test/pipeline/test-metrics.json-expected.json (hunk 13 above)
Comment (ritalwar) [reply packages/o365_metrics/data_stream/teams_user_activity/_dev/test/pipeline/test-metrics.json-expected.json]: Yes, this is a pipeline test, and since we're not processing the timestamp field in the pipeline, it's missing from the test results. However, the `sample_event` does contain the timestamp field, which records the time the event was ingested. I'm not sure whether we should use `report_date` or `report_refresh_date` as the timestamp. Let's discuss this in more detail today.

**Review by kcreddy — APPROVED** [review]
LGTM 👍🏼 (barring the timestamp and duplicates which needs further testing).

**Review by lucian-ioan — APPROVED** [review]

File: packages/o365_metrics/data_stream/teams_user_activity_user_counts/elasticsearch/ingest_pipeline/default.yml [hunk 14]
```diff
@@ -0,0 +1,166 @@
...
+      ignore_empty_value: true
+
+  - fingerprint:
+      fields:
+        - o365.metrics.teams.user.activity.user.counts.report.refresh_date
+        - o365.metrics.teams.user.activity.user.counts.report.date
+      target_field: _id
+  # Convert _Count fields to long
+  - convert:
+      field: o365.metrics.teams.user.activity.user.counts.calls
```
muthu-mps, line 77: - Can we perform `convert and rename in single step` by setting the target field, instead of using rename processor.
- As the [convert processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/convert-processor.html#:~:text=to%20be%20converted-,target_field,-no) has an option to set target field to assign the converted value.

packages/o365_metrics/data_stream/teams_user_activity_user_counts/sample_event.json
```diff
@@ -0,0 +1,83 @@
+{
+    "o365": {
+      "metrics": {
+        "teams": {
+          "user": {
+            "activity": {
+              "other_actions": {
```
muthu-mps, line 7: The `sample-events` field hierarchy doesn't match with the field definition. Please take a look.

muthu-mps: For more clarity, the fields.yml includes `user.counts` at top level which is not present here, https://github.com/elastic/integrations/blob/a2bf7c860e474e2b9b5214ea7cb1865519721de1/packages/o365_metrics/data_stream/teams_user_activity_user_counts/fields/fields.yml#L1

ritalwar: Updated, Thanks!

PR #12239, opened 2025-01-07
[microsoft-exchange-online-message-trace] - Added support for sliding window attributes and updated default interval values.

ShourieG:
## Type of change
- Enhancement

## Proposed commit message
Some scenarios demand for a configurable time window with an initial interval and an end interval. To allow this, a configurable time window has been introduced along with updates to necessary default interval values.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- 

## Screenshots

Screenshot 2025-01-10 at 6 53 22 PM

 ### Under Advanced Options

Screenshot 2025-01-10 at 6 53 54 PM


packages/microsoft_exchange_online_message_trace/_dev/build/docs/README.md
```diff
@@ -45,9 +45,13 @@ In order to continue using the Microsoft Exchange Online Message Trace you will
 - For configuring `Local Domains` you can check your [Microsoft Admin Exchange Center](https://admin.exchange.microsoft.com/) for the domains
 available in your organization. They are usually under the sections [Accepted Domains](https://admin.exchange.microsoft.com/#/accepteddomains) and [Remote Domains](https://admin.exchange.microsoft.com/#/remotedomains).
 
-- The default `Polling Interval` and `Initial Interval` values are configured to `1h`, you can however change these to your required values. The look-back 
+- The default `Polling Interval` is configured to `1h` and `Initial Interval` to `48h`, you can however change these to your required values. The look-back
```
chrisberkhout, line 48:
```suggestion
- The default `Interval` for polling is configured to `1h` and `Initial Interval` to `48h`, you can however change these to your required values. The look-back 
```
The title used for the `interval` variable is just "Interval"

packages/microsoft_exchange_online_message_trace/manifest.yml
```diff
@@ -123,7 +123,17 @@ policy_templates:
...
+            multi: false
+            required: true
+            show_user: true
+          - name: end_interval
+            type: text
+            title: End Interval
+            description: |
+              The time window starting from the initial interval up to when the logs would be pulled. This value should be always lesser than the interval.
+              Supported units for this parameter are h/m/s.
+            default: 24h
```
chrisberkhout, line 136: I think the default should be 1h, since I think I saw in the documentation or something that the status is usually available in an hour. If a user wants 24h to match what they've done elsewhere they can override the default.

packages/microsoft_exchange_online_message_trace/manifest.yml
```diff
@@ -123,7 +123,17 @@ policy_templates:
...
-            default: 1h
+            default: 48h
+            multi: false
+            required: true
+            show_user: true
+          - name: end_interval
+            type: text
+            title: End Interval
+            description: |
+              The time window starting from the initial interval up to when the logs would be pulled. This value should be always lesser than the interval.
```
chrisberkhout, line 134: > This value should be always lesser than the interval.

Also true of initial interval, to avoid request ranges where the end is before the start.

If we remove the lookback logic (in a different PR), this can be cleaned up. Then it will only be initial interval. And then we can probably just use initial interval +  end interval to avoid the complication entirely.

chrisberkhout (commented):
The logic of this looks good.

In a separate PR we should also remove the look-back logic and have it save the time of the last data seen rather than the time of the last execution.

As a name, "End Interval" is a bit confusing for someone who isn't looking at how the request is built. What do you think about this?
```
- name: min_age
  type: text
  title: Minimum Age
  description: |
    Logs will not be requested until they are at least this old.
    This value should...
    Supported units for this parameter are h/m/s.
```

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/httpjson.yml.hbs
```diff
@@ -29,8 +29,8 @@ request.transforms:
   - set:
       # Warning: Modification of the `$filter` param format can break the `last_execution_datetime` cursor
       target: url.params.$filter
-      value: StartDate eq datetime'[[(formatDate ((parseDate .cursor.last_execution_datetime "RFC3339").Add (parseDuration "-{{additional_look_back}}")) "RFC3339")]]' and EndDate eq datetime'[[(formatDate (now) "RFC3339")]]'
-      default: StartDate eq datetime'[[(formatDate (now (parseDuration "-{{initial_interval}}")) "RFC3339")]]' and EndDate eq datetime'[[(formatDate (now) "RFC3339")]]'
+      value: StartDate eq datetime'[[(formatDate ((parseDate .cursor.last_execution_datetime "RFC3339").Add (parseDuration "-{{additional_look_back}}")) "RFC3339")]]' and EndDate eq datetime'[[(formatDate (now (parseDuration "-{{end_interval}}")) "RFC3339")]]'
```
BenB196, line 32: I think you need to take into account the `additional_look_back` value here when doing the math for `end_interval`. If someone sets `additional_look_back` = `1h`, and then also sets `end_interval` = `1h`. You now have the chance of the `StartDate` & `EndDate` equaling each other, and causing the integration to deadlock on a timeframe of `0s`.

packages/microsoft_exchange_online_message_trace/_dev/build/docs/README.md
```diff
@@ -45,9 +45,13 @@ In order to continue using the Microsoft Exchange Online Message Trace you will
 - For configuring `Local Domains` you can check your [Microsoft Admin Exchange Center](https://admin.exchange.microsoft.com/) for the domains
 available in your organization. They are usually under the sections [Accepted Domains](https://admin.exchange.microsoft.com/#/accepteddomains) and [Remote Domains](https://admin.exchange.microsoft.com/#/remotedomains).
 
-- The default `Polling Interval` and `Initial Interval` values are configured to `1h`, you can however change these to your required values. The look-back 
+- The default `Polling Interval` is configured to `1h` and `Initial Interval` to `48h`, you can however change these to your required values. The look-back 
   value of `Initial Interval` should not exceed `200 hours` as this might cause unexpected errors with the API.
 
+- The default `End Interval` is configured to `24h`, you can however change these to your required values. The `End Interval` was introduced to allow a sliding
```
BenB196, line 51: Shouldn't this be a default of `0s` to keep existing functionality the same for current users? Else this would be a breaking change?

ShourieG: Yes, ideally that should be the case, still having some discussions on this internally.

ShourieG: @BenB196, so there's an issue happening with logs with status "GettingStatus" ingested, we are setting this to 1h by default to avoid pulling such logs and they seem to have their statuses updated by the end of 1h. Some users do not want these logs by default.

ShourieG: It's been reduced to 1h as a default in order to avoid logs with the "gettingStatus" state which a lot of users were complaining about.

ShourieG:
@chrisberkhout, could you review the updated PR once

chrisberkhout (approved):
Looks good.

My preference would be for a boolean setting to drop "Getting Status", but this way will work too.

ShourieG:
> Looks good.
> 
> My preference would be for a boolean setting to drop "Getting Status", but this way will work too.

Since I made a generic status field, I left it as a text type to its not only limited to "GettingStatus"

seang-es:
Just adding a note that 'end_interval' is represented in the UI as 'Minimum Age' to avoid confusion.

PR #12240, opened 2025-01-07
[google_secops] Initial release of the google secops

janvi-elastic:
## Proposed commit message
Create New integration package google_secops.
- Added data stream.
- Added data collection logic for alert data stream.
- Added the ingest pipeline for alert data stream.
- Mapped fields according to the ECS schema and added Fields metadata in the appropriate yml files.
- Added dashboards and visualizations.
- Added test for pipeline for alert data stream.
- Added system test cases for alert data stream.

## Checklist

- Clone integrations repo.
- Move to integrations/packages/google_secops directory.
> elastic-package test 

```
--- Test results for package: google_secops - START ---
╭───────────────┬─────────────┬───────────┬────────────────────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE       │ DATA STREAM │ TEST TYPE │ TEST NAME                                                              │ RESULT │ TIME ELAPSED │
├───────────────┼─────────────┼───────────┼────────────────────────────────────────────────────────────────────────┼────────┼──────────────┤
│ google_secops │             │ asset     │ dashboard google_secops-bfd74b15-5eff-45b8-b15f-01401f3deacc is loaded │ PASS   │      1.929µs │
│ google_secops │             │ asset     │ search google_secops-0e4e7181-a633-4044-af33-ab6b8e0784ec is loaded    │ PASS   │        258ns │
│ google_secops │ alert       │ asset     │ index_template logs-google_secops.alert is loaded                      │ PASS   │        209ns │
│ google_secops │ alert       │ asset     │ ingest_pipeline logs-google_secops.alert-0.1.0 is loaded               │ PASS   │        219ns │
╰───────────────┴─────────────┴───────────┴────────────────────────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: google_secops - END   ---
Done
--- Test results for package: google_secops - START ---
╭───────────────┬─────────────┬───────────┬───────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE       │ DATA STREAM │ TEST TYPE │ TEST NAME                                 │ RESULT │ TIME ELAPSED │
├───────────────┼─────────────┼───────────┼───────────────────────────────────────────┼────────┼──────────────┤
│ google_secops │ alert       │ pipeline  │ (ingest pipeline warnings test-alert.log) │ PASS   │ 288.418232ms │
│ google_secops │ alert       │ pipeline  │ test-alert.log                            │ PASS   │ 342.148692ms │
╰───────────────┴─────────────┴───────────┴───────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: google_secops - END   ---
Done
--- Test results for package: google_secops - START ---
╭───────────────┬─────────────┬───────────┬──────────────────────────┬────────┬──────────────╮
│ PACKAGE       │ DATA STREAM │ TEST TYPE │ TEST NAME                │ RESULT │ TIME ELAPSED │
├───────────────┼─────────────┼───────────┼──────────────────────────┼────────┼──────────────┤
│ google_secops │ alert       │ static    │ Verify sample_event.json │ PASS   │ 239.740171ms │
╰───────────────┴─────────────┴───────────┴──────────────────────────┴────────┴──────────────╯
--- Test results for package: google_secops - END   ---
Done
--- Test results for package: google_secops - START ---
╭───────────────┬─────────────┬───────────┬───────────┬────────┬────────────────╮
│ PACKAGE       │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT │   TIME ELAPSED │
├───────────────┼─────────────┼───────────┼───────────┼────────┼────────────────┤
│ google_secops │ alert       │ system    │ default   │ PASS   │ 2m8.093218093s │
╰───────────────┴─────────────┴───────────┴───────────┴────────┴────────────────╯
--- Test results for package: google_secops - END   ---
Done
```

## Screenshot

Screenshot 1
Screenshot 2


packages/google_secops/img/google-secops-alert-screenshot.png
jamiehynds, line 1: I think this dashboard may need to be reworked - focusing more on alerts rather than bytes and protocols. Many of the visualsations don't add value from a SecOps perspective. Can we remove visualiastions here such as source/destnation/network bytes, alerts by auth type, and alerts by ip protocol. Instead focus on areas such as Alerts by criticality, state (open, closed, acknowledeged, etc), rule name, category/category details. These areas of focus will be far more valuable to an analyst. Most of that data seems to be available in the alerts (going by the sample here- https://github.com/elastic/integrations/pull/12240/files#diff-ec4b289d2bf5a74070753b90e148fef78a98a2b9c6a44fb64ac925082158e90e)

janvi-elastic: @jamiehynds , The SecurityResult array includes fields like criticality, state, rule name, and category/category details. When creating visualizations based on these fields, it can lead to discrepancies in event counts.
For instance, we have 5 events in total and create a pie chart for the state field, you might see 5 "active" counts and 2 "alerting" counts. This could mislead users into thinking there are 7 events, whereas the actual count remains 5.

We can add one saved search which can contain all the above fields. Let me know your thoughts.

image

kcreddy: @janvi-elastic , can we split the `securityResult` inside CEL program and create 1 event for each element in `securityResult`? If there is any change of state for each of these elements in `securityResult`, I suggest you create a latest transform so that last(latest) state is preserved in destination index. Dashboards can be created on top of these destination indices.

janvi-elastic: @kcreddy , We have a [UDM schema](https://cloud.google.com/chronicle/docs/reference/udm-field-list#udm_event_data_model) with three objects containing arrays of objects: about, intermediary, and security_result. Do we need to split all three objects into separate entries?

Additionally, the metadata object (general information about a UDM event) lacks a modification time. If we split by security_result, it includes a last_updated_time. Should we use this last_updated_time as a replacement and apply a transformation?

kcreddy: @janvi-elastic,
The `about` field is also present under each security result, which is more relevant than the aggregated `event` level. Hence the split should not be using top-level `about`.

As per UDM schema `intermediary` is an entity between 'principal' and 'target'. Can you find any association inside each security event to intermediaries?
For example, if you have 3 security_results and 2 intermediaries, is it possible to say security_result1 is associated with intermediary2? 
If this is possible, then split by `security_results` and include only the associated `intermediary` of that security result.
If this is not possible, then split by `security_results` but include intermediary as an array as it is in the root `event` level.

kcreddy: @janvi-elastic @piyush-elastic , is this live data? Is it possible to give me access to the cluster?

janvi-elastic: @kcreddy I have checked the logs. We don't have any association between each security event to intermediaries. So we will split by security_results and include intermediary as an array.

jamiehynds: @janvi-elastic latest dashboard looks much better, thanks for working through it (and @kcreddy)

packages/google_secops/data_stream/alert/_dev/test/pipeline/test-alert.log-expected.json
```diff
@@ -0,0 +1,857 @@
...
+                },
+                "hostname": "sec-client01.example.com",
+                "ip": [
+                    "1.128.0.1",
+                    "10.0.0.1"
+                ]
+            },
+            "message": "Malware detected on endpoint. Immediate action required.",
+            "network": {
+                "bytes": 3072,
```
kcreddy, line 739: There are multiple values for bytes transferred as per sample event, they are not correlated.
Example: 
target.network.sentBytes: 15000 != principal.network.receivedBytes: 5120
network.receivedBytes = 2048, network.sentBytes: 1024 which are both less than what source and target are sending.

We can probably continue with your approach to populate from `network` and address any inconsistencies as bug.

packages/google_secops/data_stream/alert/_dev/test/pipeline/test-alert.log-expected.json
```diff
@@ -0,0 +1,857 @@
...
+                "transport": "tcp"
+            },
+            "observer": {
+                "product": "SecurityApp",
+                "vendor": "ExampleVendor",
+                "version": "1.0"
+            },
+            "process": {
```
kcreddy, line 747: Also add process.hash

packages/google_secops/data_stream/alert/_dev/test/pipeline/test-alert.log-expected.json
```diff
@@ -0,0 +1,857 @@
...
+                                "type": "authentication"
+                            },
+                            "vulns": {
+                                "vulnerabilities": [
+                                    {
+                                        "about": {
+                                            "labels": [
+                                                {
+                                                    "key": "vulnType",
+                                                    "value": "critical"
```
kcreddy, line 152: Inside raw data, if the key here is always `vulnType`, then we should derive `vulnerability.severity` from the `value`

packages/google_secops/img/google-secops-alert-screenshot.png
kcreddy, line 1: In the screenshot, alert severity `HIGH` and `high` are treated as different terms. I think we should lowercase or uppercase and stick to it.
Similar for priority, type, etc.

kcreddy: If I understand this correctly, we are not going to ingest if the alert is resolved? So, we don't know the active alerts (currently `ALERTING`), but only the total alerts created overall?

kcreddy: @janvi-elastic, I would like to know if we ingest all alerts states. 
https://github.com/elastic/integrations/pull/12240#discussion_r1923127933

packages/google_secops/data_stream/alert/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,120 @@
...
+    :
+      state.with({
+        "start_time": state.?cursor.last_timestamp.orValue((now - duration(state.initial_interval)).format(time_layout.RFC3339)),
+        "end_time": now.format(time_layout.RFC3339),
+      })
+  ).as(state, state.with(
+    request(
+      "GET",
+      state.url.trim_right("/") + "/v1/events:udmSearch?" + {
+        "query": ['security_result.alert_state="ALERTING"'],
```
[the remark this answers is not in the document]
janvi-elastic: ListAlerts endpoint is deprecated. So, alerts can be collected using the UDM Search API(https://cloud.google.com/chronicle/docs/reference/search-api#udmsearch) as described in [Use UDM Search](https://cloud.google.com/chronicle/docs/detection/migrate-cbn-alerts#use_udm_search). This involves querying with the condition security_result.alert_state = "ALERTING", leveraging UDM fields.

packages/google_secops/data_stream/alert/_dev/test/pipeline/test-alert.log-expected.json
janvi-elastic: No hash key-value pairs are being populated within the process object, so the process.hash mapping is not performed. All the hashes we receive are located within the file object.

packages/google_secops/img/google-secops-alert-screenshot.png
kcreddy: Based on the the `sample_event.json`, I see `alertState: active` i.e., inside response. 
But in request query you are setting `security_result.alert_state` as `ALERTING` (other possible values: `NOT_ALERTING`, `UNSPECIFIED`). 

With this, I believe that we will only ingest alerts that are active, but don't get updates when they are eventually resolved. 
Does the UDM search even provide resolved alerts? Can you check updating CEL program L#39 with query:
```
"query": ['security_result.alert_state="ALERTING" OR security_result.alert_state="NOT_ALERTING"'],
```
For UDM `query` syntax: https://cloud.google.com/chronicle/docs/investigation/udm-search

After this change, try resolving an alert. Does this CEL modification ingest your resolved alert?

My concern is that if we don't ingest alert updates (that are resolved), we keep stacking the alerts count. We maybe able to get stats such as total alerts created (by severity, type, priority, etc.) but we never know how many alerts are currently active that the users could act upon.

packages/google_secops/data_stream/alert/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,120 @@
...
+        "cursor": {
+          ?"last_timestamp": has(body.events) && size(body.events) == state.limit ? 
+            state.?cursor.last_timestamp 
+          : 
+            (
+              state.?want_more.orValue(false) ?
+                state.?first_end_time
+              :
+                state.?end_time
+            )
```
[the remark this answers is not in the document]
janvi-elastic: The second want_more check (L#75) will need to handle two scenarios:
If we have a page_size of 100 and receive 30 responses in the first call/page, the cursor should be set to state.end_time according to our logic.
If we have a page_size of 100 and receive 100 responses in the first call/page, followed by 30 responses in the second call/page, we should store state.first_end_time in the cursor instead of state.end_time (which is the current request's end time).
Note: We are retrieving the latest data in the first call or earlier. Thus, we are adjusting the request filter backwards to collect all events.

packages/google_secops/data_stream/alert/_dev/test/pipeline/test-alert.log-expected.json
kcreddy: The process itself has an executable file associated with it and these hashes are of that file. 
I think these hashes should belong with `process.hash.*` instead of `file.hash.*`

packages/google_secops/_dev/deploy/docker/files/config.yml
```diff
@@ -0,0 +1,30 @@
...
+      Content-Type:
+        - "application/x-www-form-urlencoded"
+    responses:
+      - status_code: 200
+        headers:
+          Content-Type:
+            - "application/json"
+        body: >
+          {"access_token": "1/fFAGRNJru1FTz70BzhT3Zg","expires_in": 3920,"token_type": "Bearer"}
+  - path: /v1/events:udmSearch
```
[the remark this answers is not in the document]
janvi-elastic: In system testing, adding a scenario to test pagination is not feasible as it will always match the first rule. This is because pagination in this integration is based on time filters, rather than the typical approaches like nextCursor, offsets, or nextLink.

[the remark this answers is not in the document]
kcreddy: Have you tried with regular expressions?

Example from mimecast: https://github.com/elastic/integrations/blob/main/packages/mimecast/_dev/deploy/docker/files/config.yml#L2026-L2037
In your config file, you can define your pagination rule (req/resp) before the existing one and get it matched using timestamp inside first request.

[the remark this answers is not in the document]
kcreddy: You can use [`gorilla/mux`](https://github.com/elastic/stream?tab=readme-ov-file#options) syntax for query_params as well.

packages/google_secops/img/google-secops-alert-screenshot.png
janvi-elastic: But we have just credentials json. We don't have access to google instance so we are not able to check that if we resolved/close the alert it will going to impact the security_result.alert_state or not.

kcreddy: > But we have just credentials json. We don't have access to google instance so we are not able to check that if we resolved/close the alert it will going to impact the security_result.alert_state or not.

Looks like we don't have a way to test if we can get alerts that are resolved. This means that we can get total counts of alerts created, but we can't know if or when the alert is resolved. So, all the alerts will continue to remain in `ALERTING` state.

@jamiehynds if this limitation sounds okay, we can move ahead. Let me know your thoughts.

packages/google_secops/data_stream/alert/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,122 @@
...
+        },
+        "start_time": state.start_time,
+        // As the records are received in descending order, we store the end_time from the first pagination call in first_end_time for use in subsequent interval API calls.
+        // During pagination calls, the end_time is continuously updated to represent the minimum value of the udm.metadata.eventTimestamp field.
+        "end_time": has(body.events) && body.events != null ? body.events.map(e, e.udm.metadata.eventTimestamp).map(t, timestamp(t)).min() : state.end_time,
```
kcreddy, line 84: > As the records are received in descending order, 

Can you confirm the descending order is correct?
If the events are in an order, why do we need to iterate and perform `min()` operation?
Isn't it going to be the timestamp from the first event? So, `body.events[0].udm.metadata.eventTimestamp` should be enough, right?

If there is no order, then a `max()` operation should be performed, right?

packages/google_secops/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1592 @@
...
+                field: error.message
+                value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - script:
+      lang: painless
+      description: Script to set vulnerability.severity
+      tag: set_vulnerability_severity
+      source: |-
+        def severity_v = "hye";
+        if (ctx.google_secops?.alert?.udm?.extensions?.vulns?.vulnerabilities instanceof List) {
+          def sv = "hello";
```
kcreddy, line 104:
```suggestion
        if (ctx.google_secops?.alert?.udm?.extensions?.vulns?.vulnerabilities instanceof List) {
```
?

packages/google_secops/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1592 @@
...
+          for (list in ctx.google_secops.alert.udm.extensions.vulns.vulnerabilities) {
+            for (label in list["about"]["labels"]) {
+              if (label["key"] == 'vulnType') {
+                ctx.vulnerability.severity = label["value"];
+              }
+            }
+          }
+        }
+      ignore_failure: true
```
kcreddy, line 113: Add error.message on_failure

packages/google_secops/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1592 @@
...
+        } else if (ctx.google_secops.alert.udm.securityResult.severity.toUpperCase() == 'INFORMATIONAL') {
+          ctx.event.severity = 100
+        }  else if (ctx.google_secops.alert.udm.securityResult.severity.toUpperCase() == 'LOW') {
+          ctx.event.severity = 200
+        } else if (ctx.google_secops.alert.udm.securityResult.severity.toUpperCase() == 'MEDIUM') {
+          ctx.event.severity = 300
+        } else if (ctx.google_secops.alert.udm.securityResult.severity.toUpperCase() == 'NONE') {
+          ctx.event.severity = 101
+        } else if (ctx.google_secops.alert.udm.securityResult.severity.toUpperCase() == 'UNKNOWN_SEVERITY') {
+          ctx.event.severity = 0
```
kcreddy, line 1386: Can you make them in an order? Also confirm how the values are provided (if there was any documentation).
Not sure how `NONE` is 101 and `INFORMATIONAL` is 100.

packages/google_secops/data_stream/alert/_dev/test/pipeline/test-alert.log-expected.json
```diff
@@ -0,0 +1,848 @@
...
+                                },
+                                {
+                                    "key": "notificationStatus",
+                                    "value": "sent"
+                                }
+                            ],
+                            "priority": "high",
+                            "priorityDetails": "Critical threat to organization security",
+                            "ruleName": "Critical Malware Detection",
+                            "severity": "critical",
```
kcreddy, line 535: Both `priority` and `severity` are lowercase here, and I don't see any processors that make them to uppercase.
So, how are the dashboards updated have them as uppercase?

kcreddy (commented):
Waiting for clarification on if just ingesting active alerts using 
` "query": ['security_result.alert_state="ALERTING"'],`
is sufficient.

packages/google_secops/data_stream/alert/agent/stream/cel.yml.hbs
kcreddy: Since we don't have system tests with pagination, I suggest you verify (using tracer and agent logs) against `limit` of `1`  that `start_time`, `end_time`, `first_end_time`, and `want_more` are accurately populated for first through last request.

packages/google_secops/data_stream/alert/_dev/test/pipeline/test-alert.log-expected.json
janvi-elastic: We are getting `priority` and `severity` in uppercase in live logs. We will update the pipeline test log.

packages/google_secops/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
janvi-elastic: Yes this are the values mentioned in [documentation](https://cloud.google.com/chronicle/docs/reference/udm-field-list#securityresultproductseverity).

kcreddy: Great, please add the link to processor description or as comment.

packages/google_secops/data_stream/alert/agent/stream/cel.yml.hbs
kcreddy: okay, just verify the pagination is working fine with live data. Thanks

PR #12246, opened 2025-01-07
[ Cloudflare Logpush ] Populate event.id from the Ray ID value for Firewall Events and HTTP Requests

### LINKED ISSUE #12234: [ Cloudflare Logpush ] Ray ID field should have the same name in all datasets.
Hello,

Cloudflare uses the Ray ID to identiy a request in its systems, if you are collecting logs from Firewall Events and HTTP Requests, you can identify a request in both datasets by the Ray ID.

Currently the Cloudflare Logpush integration uses different fields for each dataset, but the information is the same.

For example, the field woul be named `cloudflare_logpush.http_request.ray.id` in the HTTP Requests dataset and `cloudflare_logpush.firewall_event.ray.id` in the Firewall Events dataset.

The field with the ray id should have the same name in all datasets, so this could be renamed to `cloudflare_logpush.ray.id` or even `event.id`.

leandrojmp:
- Enhancement

## Proposed commit message

This PR adds a `set` processor on the Firewall Events and HTTP Requests ingest pipelines to populate the field `event.id` with the value of the fields `cloudflare_logpush.http_request.ray.id` in the HTTP Requests dataset and `cloudflare_logpush.firewall_event.ray.id` in the Firewall Events dataset.

This enables the user to filter for events on both datasets by the same `event.id`.

## Checklist

## Related issues
- Closes #12234


chemamartinez (approved):
LGTM

Issue #11746, opened 2024-11-14
[kubernetes] Make fingerprinting optional or make fingerprint length configurable


kaiyan-sheng:
For users who are trying to ingest small files using the kubernetes integration might run into this error with `file size is too small for ingestion`:
```
cannot start ingesting from file "/var/log/containers/test.log": filesize of "/var/log/containers/test.log" is 594 bytes, expected at least 1024 bytes for fingerprinting: file size is too small for ingestion
``` 

We should make fingerprinting optional or make fingerprint length configurable for this use case. 
```
prospector.scanner.fingerprint.length: 64
```

Original pr: https://github.com/elastic/integrations/pull/11212

PR #123, opened 2020-06-29
Add pull request template

mtojek:
This PR setups a pull-request template for the repository.

Preview link: https://github.com/mtojek/integrations/blob/pr_template/.github/PULL_REQUEST_TEMPLATE.md

PR #12274, opened 2025-01-08
Make prospector.scanner.fingerprint configurable

MichaelKatsoulis:
## Proposed commit message

- WHAT:  In kubernetes container logs, make usage of fingerprint optional and configurable
- WHY:  The default values of prospector.scanner.fingerprint do not suit all users

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-  Closes [#123](https://github.com/elastic/integrations/issues/11746)

## Screenshots

fingerprint


packages/kubernetes/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: 1.69.0
+  changes:
+    - description: Make fingerprint configurable
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/11243
```
andrewkroh, line 6:
```suggestion
      link: https://github.com/elastic/integrations/pull/12274
```

packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -69,6 +69,24 @@ streams:
...
+            enabled: true
+            offset: 0
+            length: 1024
+      - name: useFingerprint
+        type: bool
+        title: Use file_identity fingerprint
+        multi: false
+        required: true
+        show_user: false
+        description: If prospector.scanner.fingerprint.enabled is set to false, file_identity fingerprint must be switched off. If it set to true, file_identity fingerprint must be switched on.
```
tetianakravchenko, line 88: I think those conditions might be quite confusing, wondering if suggested option wouldn't be easier

packages/kubernetes/data_stream/container_logs/agent/stream/stream.yml.hbs
```diff
@@ -5,11 +5,13 @@ paths:
...
   dataset: {{data_stream.dataset}}
+prospector.scanner.symlinks: {{ symlinks }}
 prospector:
   scanner:
-    fingerprint.enabled: true
-    symlinks: {{ symlinks }}
+    {{ fingerprint }}
+{{#if useFingerprint}}
 file_identity.fingerprint: ~
+{{/if}}
```
tetianakravchenko, line 14: To avoid conditions in fields description, we could use smth like below instead:
```
{{#if useFingerprint}}
file_identity.fingerprint: ~
prospector:
  scanner:
    enabled: true
    {{#if fingerprintConfig}}
    {{fingerprintConfig}}
    {{/if}}
{{/if}}
```
(not tested)

MichaelKatsoulis: I thought about this but then it would be confusing to the users as to what options exactly they can set in fingerprintConfig yaml. For example they can set `offset` and `length` but not `enabled`. And if they set `enabled` as well then there will be an error. 
Maybe the`enabled` could also be part of the available configuration options, to avoid errors, but it won't have any effect as it will be under the `{{#if useFingerprint}}`

Like this:

```
{{#if useFingerprint}}
file_identity.fingerprint: ~
prospector:
  scanner:
    {{ fingerprint }}
{{/if}}
```

MichaelKatsoulis: @tetianakravchenko check the new [commit](https://github.com/elastic/integrations/pull/12274/commits/8287dab663ec0d037db6af93f9c4b4b52fc020c9)

tetianakravchenko: >  Maybe theenabled could also be part of the available configuration options, to avoid errors, but it won't have any effect as it will be under the {{#if useFingerprint}}

but does it change much from the previous option? I mean that we still allow config like:
```
useFingerprint: true
fingerprint:
  enabled: false
```
that is not supported, no?

As an option in `useFingerprint` description can be mentioned that `prospector.scanner.fingerprint` will be enabled automatically, for configuring additional `prospector.scanner.fingerprint` parameters please use `Fingerprint settings`. WDYT?

Another reason to put  `prospector.scanner.fingerprint` to the template - now we allow to configure all `prospector.scanner.*`, not just the `prospector.scanner.fingerprint` is it expected behavior?

MichaelKatsoulis: > Another reason to put  prospector.scanner.fingerprint to the template - now we allow to configure all prospector.scanner.*, not just the prospector.scanner.fingerprint is it expected behavior?

In theory they could configure all prospector.scanner.* but that was not the idea. The idea was to configure fingerprint only. It was just more clear if they also have to set the fingerprint key.  
We could change that and go one level deeper.

```
useFingerprint: true
fingerprint:
  enabled: false
```
That is not supported but can never happened as if the set useFingerprint false, then the `prospector.scanner.fingerprint` won't be include in the final config. If we do not include the `enabled` parameter, then it is error prone. First because users do not know exactly what are the options and second because if the add the `enabled` for whatever reason there will be an error about key duplication.

I suggest we just do that:

```
{{#if useFingerprint}}
file_identity.fingerprint: ~
prospector:
  scanner:
    fingerprint:
      {{ fingerprintConfig }}
{{/if}}
```

tetianakravchenko: > That is not supported but can never happened as if the set useFingerprint false, then the prospector.scanner.fingerprint won't be include in the final config.

why? it can happen
`useFingerprint: true`, but variable
`fingerprintYaml` is:
```
enabled: false
```

>  If we do not include the enabled parameter, then it is error prone.

Yes, it makes sense, I understand your point

> I suggest we just do that:
> ```
> {{#if useFingerprint}}
> file_identity.fingerprint: ~
> prospector:
>  scanner:
>    fingerprint:
>      {{ fingerprintConfig }}
> {{/if}}
> ```

I think it is better to prevent unexpected scanners configuration.
To the default value I would add a comment:
```
enabled: true    # This must be set to `true`
offset: 0
length: 1024
```
or smth like that, so to make it explicitly said. wdyt?

MichaelKatsoulis: if enabled is set to false but the file_identity.fingerprint is configured then there is an error. But I will add this comment to be more clear

MichaelKatsoulis: Comment added in https://github.com/elastic/integrations/pull/12274/commits/b04da0a28a1a8edf2d257764e47abbe2c9d686fc

tetianakravchenko (approved):

PR #12286, opened 2025-01-09
[GitLab] Update params mapping for nested object arrays

### LINKED ISSUE #11941: [GitLab]: gitlab.api logs with nested key-value pairs fail to index
### Integration Name

GitLab [gitlab]

### Dataset Name

gitlab.api

### Integration Version

0.3.1

### Agent Version

8.15.2

### Agent Output Type

logstash

### Elasticsearch Version

8.15.2

Rocky Linux 8.10 (x86_64)

### Software/API Version

GitLab 17.5

### Error Message

I can see the following indexing error from Elasticsearch in my Logstash DLQ:

```
response: {"create"=>{"status"=>400, "error"=>{"type"=>"document_parsing_exception", "reason"=>"[1:2202] failed to parse field [gitlab.api.params.value] of type [keyword] in document with id '7c4...'. Preview of field's value: '{features={vault_secrets=[FILTERED], service_variables=[FILTERED], variables=[FILTERED], trace_reset=[FILTERED], trace_size=[FILTERED], raw_variables=[FILTERED], trace_checksum=[FILTERED]}, shell=bash, executor=shell, name=gitlab-runner, version=17.4.0, platform=linux, revision=b92..., architecture=amd64}'", "caused_by"=>{"type"=>"illegal_argument_exception", "reason"=>"Expected text at 1:1850 but found START_OBJECT"}}}
```

### Event Original

```
{
    "time": "2024-12-02T09:03:21.162Z",
    "severity": "INFO",
    "duration_s": 0.11619,
    "db_duration_s": 0.09926,
    "view_duration_s": 0.01693,
    "status": 204,
    "method": "POST",
    "path": "/api/v4/jobs/request",
    "params": [
        {
            "key": "info",
            "value": {
                "name": "gitlab-runner",
                "version": "17.4.0",
                "revision": "b92...",
                "platform": "linux",
                "architecture": "amd64",
                "executor": "shell",
                "features": {
                    "variables": "[FILTERED]",
                    "image": null,
                    "services": null,
                    "artifacts": null,
                    "cache": null,
                    "fallback_cache_keys": null,
                    "shared": null,
                    ...
                    "raw_variables": "[FILTERED]",
                    "trace_reset": "[FILTERED]",
                    "trace_checksum": "[FILTERED]",
                    "vault_secrets": "[FILTERED]",
                    "service_variables": "[FILTERED]"
                },
                "config": {
                    "gpus": ""
                }
            }
        },
        {
            "key": "token",
            "value": "[FILTERED]"
        },
        {
            "key": "system_id",
            "value": "s_a9..."
        },
        {
            "key": "last_update",
            "value": "ede..."
        }
    ],
    ...
    "meta.caller.id": "POST /api/:version/jobs/request"
}
```

I've redacted some of the original, and removed some of the irrelevant contents since I'm having to type it out by hand. 

### What did you do?

GitLab integration with default configuration (no custom processors).

### What did you see?

Where the `gitlab_api.json` log has a gitlab.api.params.value that is a JSON object instead of a string, this conflicts with the keyword mapping in the gitlab.api index template and the log fails to index. Since we are ingesting our Elastic Agent integration logs via Logstash, we see these added to the DLQ. 

The gitlab.api ingest pipeline or index mapping should accommodate fields such as gitlab.api.params.value being a JSON object instead of a static string. 

### Anything else?

_No response_

tehbooom:
## Proposed commit message

For the GitLab API datastream the `params` field is an array of objects that are dynamic. The fields were broken into `key` and `value` fields but broke when the value was another object. 

This PR breaks each key value into their own field and adds field mappings for those fields.

For example this

```json
"params": [
    {
        "key": "private_token",
        "value": "[FILTERED]"
    }
]
```
is now this with correct field mappings.

```json
"params": [
    {
        "private_token": "[FILTERED]"
    }
]
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #11941


packages/gitlab/data_stream/api/fields/fields.yml
```diff
@@ -155,9 +155,43 @@
         - name: params
           type: group
           fields:
-            - name: value
+            - name: info
```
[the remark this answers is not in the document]
tehbooom: No we don't. GitLab does not provide a list of possible fields for `params`. The only way to know what they are is gathering as much data as possible to ensure we cover as much as we can.

I played around with nested groups objects 3 levels down with `object_mapping_type: '*'` but that doesnt seem like the best way IMO.

chemamartinez:
> > I have noticed that the `production` data stream already handles this issue with a different approach. In that case, `value` is renamed to `value_json` and mapped as an object in case it is not a string.
> > I do prefer the solution given in this PR, but it makes me wonder a couple of things:
> > 
> > * Shall we unify the criteria here? that would make the integration more consistent.
> > * If we follow this approach, we could cause a breaking change in the other data stream by stopping using the `value_json` field.
> 
> I do prefer the solution in this PR as I think it presents the fields in a cleaner way and easier to understand. However, that does require an adjustment to the other datastream to ensure they line up correctly.

Yes, me too. I have checked that these `value_json` params are not being used in any dashboards, and in this particular case I think it is unlikely that anyone is using these fields to create custom assets. I'd go with the change for both data streams.

packages/gitlab/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.4.1"
+  changes:
+    - description: Fix nested key-value pair fields for api and production data streams.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/12286
```
chemamartinez, line 6:
```suggestion
- version: "2.0.0"
  changes:
    - description: Add mapping for objects values in params field for api and production data streams.
      type: enhancement
      link: https://github.com/elastic/integrations/pull/12286
```

I just want to point out to users that this could be breaking by bumping the major version. Feel free to change the description if you consider that another one is better suited.

packages/gitlab/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 3.1.3
 name: gitlab
 title: GitLab
-version: 1.4.0
+version: 1.4.1
```
chemamartinez, line 4:
```suggestion
version: 2.0.0
```

chemamartinez (approved):
LGTM!

PR #12292, opened 2025-01-09
microsoft_exchange_online_message_trace - handle SenderAddress of '<>'

andrewkroh:
## Proposed commit message

```
In the microsoft_exchange_online_message_trace integration, JSON events
containing "SenderAddress":"<>" caused a pipeline error of

    Provided Grok expressions do no match field value: [<>]

This commit changes the pipeline to treat the value that same as it would an
empty string or no SenderAddress field present.

Improve the pipeline error.message by including more details about the
processor that is failing.
```

## Checklist


packages/microsoft_exchange_online_message_trace/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -335,4 +338,7 @@ on_failure:
       allow_duplicates: false
   - append:
       field: error.message
-      value: '{{{ _ingest.on_failure_message }}}'
+      value: >-
+        Processor '{{{ _ingest.on_failure_processor_type }}}'
+        {{#_ingest.on_failure_processor_tag}}with tag '{{{ _ingest.on_failure_processor_tag }}}'
```
chemamartinez, line 343:
```suggestion
        {{#_ingest.on_failure_processor_tag}} with tag '{{{ _ingest.on_failure_processor_tag }}}'
```

packages/microsoft_exchange_online_message_trace/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -335,4 +338,7 @@ on_failure:
...
+        {{/_ingest.on_failure_processor_tag}}failed with message '{{{ _ingest.on_failure_message }}}'
```
chemamartinez, line 344:
```suggestion
        {{/_ingest.on_failure_processor_tag}} failed with message '{{{ _ingest.on_failure_message }}}'
```

packages/microsoft_exchange_online_message_trace/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -37,6 +37,7 @@ processors:
       field: event.type
       value: [info]
   - set:
+      if: ctx.microsoft?.online_message_trace?.SenderAddress != "<>"
```
chemamartinez, line 40: Wondering if we should take the same precaution with `RecipientAddress` as they seem to be mirror fields.

andrewkroh: I'd like to leave it as is because I haven't seen any evidence that this is required.

packages/microsoft_exchange_online_message_trace/data_stream/log/elasticsearch/ingest_pipeline/default.yml
andrewkroh: That would introduce a double space when the processor has a tag. Perhaps it helps you to review the string in JSON encoding. Do you see any issues?

```
{
  "value": "Processor '{{{ _ingest.on_failure_processor_type }}}' {{#_ingest.on_failure_processor_tag}}with tag '{{{ _ingest.on_failure_processor_tag }}}' {{/_ingest.on_failure_processor_tag}}failed with message '{{{ _ingest.on_failure_message }}}'"
}
```

packages/microsoft_exchange_online_message_trace/data_stream/log/elasticsearch/ingest_pipeline/default.yml
andrewkroh: Same as above. The space should is not needed.

packages/microsoft_exchange_online_message_trace/data_stream/log/elasticsearch/ingest_pipeline/default.yml
chemamartinez: Thanks for the clarification.

chemamartinez (approved):

PR #3935, opened 2022-08-02
[github] Add support for ingesting Advanced Security alerts for an entire organization

kcreddy:
## What does this PR do?
Currently Github Code Scanning, Secret Scanning, and Dependabot alerts only support repo level endpoints. This PR adds support for organisation endpoints giving choice to user to choose.
 

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).

PR #12300, opened 2025-01-10
Test packages mappings - DO NOT MERGE

mrodm:
Relates #3935 

Do tests in the packages failed when using the validation based on mappings.


packages/box_events/data_stream/events/fields/ecs.yml
```diff
@@ -0,0 +1,7 @@
+# This definition could be removed once Kibana constraint is updated
+# to 8.15.2 or higher. "ecs@mappings" component template would define
+# the correct dynamic template for it.
+- external: ecs
+  name: threat.enrichments.indicator.first_seen
+- external: ecs
+  name: threat.enrichments.indicator.last_seen
```
mrodm, line 7: Previously they were mapped as `keyword`, with this change they would be `date`.

packages/claroty_ctd/data_stream/event/fields/ecs.yml
```diff
@@ -0,0 +1,6 @@
+# This definition could be removed once Kibana constraint is updated
+# to 8.15.2 or higher. "ecs@mappings" component template would define
+# the correct dynamic template for it.
+- external: ecs
+  name: threat.indicator.modified_at
+
```
mrodm, line 6: Previously they were mapped as keyword, with this change they would be date.

packages/crowdstrike/data_stream/fdr/fields/fields.yml
```diff
@@ -27,7 +27,7 @@
       type: long
     - name: AsepWrittenCount
       type: long
-    - name: assessments.*
+    - name: assessments
```
mrodm, line 30: This definition was generating a dynamic template to transform objects to `flattened`, and it was not recognizing the fields under assessments like `assessments.application_firewall_mac`.

packages/linux/data_stream/memory/fields/fields.yml
```diff
@@ -82,7 +82,7 @@
           metric_type: gauge
           description: |
             The percentage of used swap memory.
-    - name: vmstat.*
+    - name: vmstat
```
mrodm, line 85: This definition was generating a dynamic template to transform objects to `flattened`, and it was not recognizing the fields under vmstats like `vmstats.allocstall_device`.

PR #12369, opened 2025-01-16
[Defender Endpoint] fix null ref for description description

SimonKoetting:
There can be events without a description, which leads the if condition to fail with:
cannot access method/field [length] from a null def reference

### DIFF
--- a/packages/microsoft_defender_endpoint/data_stream/log/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/microsoft_defender_endpoint/data_stream/log/elasticsearch/ingest_pipeline/default.yml
@@ -188,7 +188,7 @@ processors:
       field: json.description
       target_field: rule.description
       ignore_missing: true
-      if: ctx?.json != null && (ctx.json?.description).length() < 1020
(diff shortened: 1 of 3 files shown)


packages/microsoft_defender_endpoint/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -188,7 +188,7 @@ processors:
       field: json.description
       target_field: rule.description
       ignore_missing: true
-      if: (ctx.json?.description).length() < 1020
+      if: ctx?.json?.description != null && (ctx.json?.description).length() < 1020
```
kcreddy, line 191:
```suggestion
      if: ctx.json?.description != null && ctx.json.description.length() < 1020
```

ctx is never null and we can remove `?` in second condition as null-check already done in first.

kcreddy (approved):

kcreddy:
@SimonKoetting, let me know if you don't have access to merge the PR.

SimonKoetting:
@kcreddy thanks, yes, thats the case :D would be nice if you could merge it. Thanks!

PR #12375, opened 2025-01-16
crowdstrike.fdr: avoid mapping conflict when host metadata is not an object 

chemamartinez:
## Proposed commit message

Found this error in a live scenario:
```
{\"type\":\"document_parsing_exception\",\"reason\":\"[1:2175] object mapping for [crowdstrike.info.host] tried to parse field [host] as object, but found a concrete value\"}, dropping event!
```

Where event being dropped contains metadata host info as follows:
```
{
  "metadata": {
    "host": "example_host_name"
  }
}
```

`metadata.host` is supposed to be an object with several host fields, so when it is renamed to `crowdstrike.info.host` it causes the error above, as this field is mapped as an object.

**This fix is being backported to skip the current minimum required Kibana version (8.16.0).**

## Checklist

### DIFF
--- a/packages/crowdstrike/changelog.yml
+++ b/packages/crowdstrike/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.46.1"
+  changes:
+    - description: Fix mapping error when host metadata is string instead of object.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/99999
 - version: "1.46.0"
   changes:
     - description: Extract user and host names from the name field.
--- a/packages/crowdstrike/changelog.yml
+++ b/packages/crowdstrike/changelog.yml
@@ -1,9 +1,9 @@
 # newer versions go on top
 - version: "1.46.1"
   changes:
-    - description: Fix mapping error when host metadata is string instead of object.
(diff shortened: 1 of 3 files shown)


andrewkroh (commented):
Can you please add a test case for this.

Another solution would be to adopt [subobjects: false](https://github.com/elastic/package-spec/blob/8f1292454250c378af0432606bb99418cc19c753/spec/integration/data_stream/manifest.spec.yml#L365) for the data stream.

andrewkroh:
> Isn't the subject approach conditioned by https://github.com/elastic/kibana/issues/193044? this was fixed for 8.17 and this fix needs to work in older versions.

Probably. I know the fix is required when you set `subobects` at the field level, but not sure about when you change it at the data stream level.

andrewkroh (commented):
>> Can you please add a test case for this.

> I will add a test case.

@chemamartinez Can you push in a pipeline test for this case please.

chemamartinez:
> @chemamartinez Can you push in a pipeline test for this case please.

I have not added it yet because I have the suspicion that the error was not realistic and that it is a case that cannot really happen, I am still doing analysis on this.

chemamartinez:
Closing the PR as the aid master sample was malformed for unrelated reasons.

PR #12379, opened 2025-01-16
[rubrik] Switch to cel and cleanup

ghost:
## Overview

This PR aims to improve consistency across all `rubrik` data streams by:
- Adding `cluster` and `effectiveSlaDomain` fields where necessary and using consistent field names.
- Switching entirely to the `cel` input, so we only depend on one input. This simplifies the integration and helps avoid future issues.
- Adding pagination support where needed.
  - For pagination, the `first` argument controls the page size. Should we set a default value of like `10` and allow users to customize it?

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/rubrik/docs/README.md
+++ b/packages/rubrik/docs/README.md
@@ -343,7 +343,7 @@ An example event for `virtual_machines` looks as following:
     "data_stream": {
         "namespace": "default",
         "type": "metrics",
-        "dataset": "rubrik.virtualmachines"
+        "dataset": "rubrik.virtual_machines"
     },
     "elastic_agent": {
         "id": "e74cda94-80b2-42d7-a508-21885a2614b5",
@@ -374,7 +374,7 @@ An example event for `virtual_machines` looks as following:
         "ingested": "2025-01-08T13:08:19Z",
         "created": "2025-01-08T13:08:18.698Z",
         "kind": "metric",
-        "dataset": "rubrik.virtualmachines"
+        "dataset": "rubrik.virtual_machines"
     }
 }
 ```


gpop63:
Should the `ssl`, `resource_timeout`, and `enable_request_tracer` settings be at the package level or the data stream level? Right now `ssl` is at the package level while `resource_timeout` and `enable_request_tracer` are at the data stream level.

tommyers-elastic:
@gpop63 what's the difference between effective SLA domain and configured SLA domain?

gpop63:
> @gpop63 what's the difference between effective SLA domain and configured SLA domain?

When getting SLA domain data, you can use either `effectiveSlaDomain` or `configuredSlaDomain`. However, their documentation is not very helpful: `Effective SLA Domain of the hierarchy object` and `SLA Domain configured for the hierarchy object`. Both fields contain the same values so we chose to use `effectiveSlaDomain`.  Do you think we should include both?

packages/rubrik/data_stream/drives/manifest.yml
```diff
@@ -33,15 +39,34 @@ streams:
...
-    template_path: httpjson.yml.hbs
-    title: Rubrik Node Drives
-    description: Collect Rubrik Node drives data via the RSC GraphQL API
+        default: "# filter:\n#   - field: \"location\"\n#     texts: \n#       - \"prod-lab.local\"\n"
+        description: >-
+          Specify filters for refining the data. Filters must be defined as an array of Filter objects. Refer to the [Rubrik API documentation](https://rubrikinc.github.io/rubrik-api-documentation/schema/reference/filter.doc.html) for valid fields.
+      - name: resource_timeout
+        type: text
+        title: Resource Timeout
+        description: Duration before declaring that the HTTP client connection has timed out. Valid time units are ns, us, ms, s, m, h. Default is "30"s.
```
devamanv, line 61:
```suggestion
        description: Duration before declaring that the HTTP client connection has timed out. Valid time units are ns, us, ms, s, m, h. Default is 60s.
```
I think we overlooked this. We need to change this for all data streams.

packages/rubrik/data_stream/managed_volumes/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,136 @@
...
+    - |
+      query getManagedVolumes($first: Int, $after: String, $filter: [Filter!]) {
+        managedVolumes(first: $first, after: $after, filter: $filter) {
+          pageInfo {
+            startCursor
+            endCursor
+            hasPreviousPage
+            hasNextPage
+          }
+          count
```
devamanv, line 48: we don't seem to be using this field. Should we keep it?

packages/rubrik/data_stream/virtualmachines/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -29,9 +29,21 @@ processors:
       field: json.name
       target_field: rubrik.virtualmachine.name
       ignore_missing: true
+  - rename:
+      field: json.cluster.id
+      target_field: rubrik.virtualmachine.cluster.id
```
devamanv, line 34:
```suggestion
      target_field: rubrik.virtual_machine.cluster.id
```
For consistency with other data streams, I think we should name it like this.

packages/rubrik/data_stream/virtualmachines/manifest.yml
```diff
@@ -1,30 +1,36 @@
...
+
       - name: tags
         type: text
         title: Tags
         multi: true
         required: false
-        show_user: true
+        show_user: false
+        default:
+          - rubrik.virtualmachines
```
devamanv, line 33: If we decide to change it above, please make the change here as well.

packages/rubrik/changelog.yml
```diff
@@ -22,6 +22,9 @@
     - description: Add `tasks` data stream
       type: enhancement
       link: https://github.com/elastic/integrations/pull/12396
+    - description: Switch to cel input and cleanup
```
tommyers-elastic, line 25: please can we be a little more descriptive/specific than 'cleanup' - e.g. add TSDB metric types and units, improve field descriptions etc

packages/rubrik/data_stream/filesets/manifest.yml
```diff
@@ -47,13 +47,13 @@ streams:
...
-      - name: first
+      - name: pageSize
         type: integer
-        title: First
+        title: Page Size
         multi: false
         required: false
         show_user: false
-        description: Specifies the maximum number of items to retrieve from the dataset.
+        description: The number of results to return per page.
```
tommyers-elastic, line 56: i would be a little more descriptive here - reading this without any context makes me think this has some impact on the resulting data in Elastic. we could say something like: the number of results returned per page from Rubrik GraphQL queries; decreasing page size will result in more API requests, but may be more efficient in some situations.

packages/rubrik/data_stream/managed_volumes/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -50,6 +50,22 @@ processors:
       field: response.numChannels
       target_field: rubrik.managed_volumes.num_channels.count
       ignore_missing: true
+  - rename:
+      field: response.cluster.id
+      target_field: rubrik.managed_volumes.cluster.id
```
tommyers-elastic, line 55: i know we typically use naming conventions like this, where the dataset is prefixed, but this means that customers can't run queries across all their rubrik data to filter by cluster ID, right? 

i think `rubrik.cluster.id` would be better and consistent across datasets, meaning queries can filter and aggregate per cluster regardless of data sets. 

same goes for SLA domain.

tommyers-elastic: same would go for any other rubrik object IDs which are not only associated with the current dataset.

packages/rubrik/data_stream/monitoring_jobs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -52,11 +52,11 @@ processors:
       ignore_missing: true
   - rename:
       field: response.name
-      target_field: rubrik.monitoring_jobs.cluster_name
+      target_field: rubrik.monitoring_jobs.cluster.name
```
tommyers-elastic, line 55: same goes for here (i don't think cluster.* fields should be prefixed with dataset name)

tommyers-elastic (approved):
thanks for this - looking great. 

only think i want to resolve is the naming of fields which are common across datasets

devamanv (approved):

efd6 (commented):
I've just taken a look at the CEL code in this change and I can see that you are using Handlebars templating in the CEL program. Please don't do that; it break tooling for ensuring consistent code formatting and, more significantly, hampers support cases. The preferred approach is to template into the state variable.

Issue #8435, opened 2023-11-08
[Custom Packages] Moving to Input packages


jamiehynds:
Our 'custom packages' currently state that changing the dataset will send the data to a different index. However, these packages create a template such as `logs-udp.generic` matching `logs-udp.generic-*` indices. This contradicts the text saying that changing the dataset will send the data to a different index.

Currently, the data itself will go into the new destination, however the initial component template is made to match the incorrect pattern.

In order to resolve this, our custom packages need to be convert to input packages (i.e. `type: input`)

We need to test the user experience when upgrading from a current custom package to the new input package. Are there breaking changes involved and to what extend? Depending on the experience, we may need to discuss deprecation of the current packages and shipping new v2 input packages (but hopefully that can be avoided). Fleet team will need to be engaged to discuss. 

List of packages to convert:

- [ ] CEL
- [ ] httpjson (consider depreaction here to avoid confusion over CEL vs httpjon)??
- [ ] AWS (S3 and Cloudwatch inputs)
- [ ] Azure Blob Storage
- [ ] Google Cloud Storage
- [ ] Google Pub/Sub
- [ ] HTTP Endpoint
- [ ] Journald
- [ ] Kafka
- [ ] Log
- [ ] TCP
- [ ] UDP
- [ ] Windows

andrewkroh:
> We need to test the user experience when upgrading from a current custom package to the new input package.

Agree. The most important thing is that the upgrade requires no user intervention and that the Fleet stream ID does not change so that input state is not lost. Please do a test for this using one of the stateful inputs and verify that the existing state continues to be used after the upgrade.

andrewkroh:
The blog post at https://www.elastic.co/blog/customize-data-ingestion-elastic-input-packages can be edited to include the converted packages when they are done. @ishleenk17 can probably aid in that edit if we let her know.

ishleenk17:
> The blog post at https://www.elastic.co/blog/customize-data-ingestion-elastic-input-packages can be edited to include the converted packages when they are done. @ishleenk17 can probably aid in that edit if we let her know.

Sure, let me know when these are migrated. We can add it to the blog.

taylor-swanson:
@ishleenk17, [per this comment](https://github.com/elastic/integrations/issues/8435#issuecomment-1945437359), we have a new package to add to the list, `Custom UDP Logs (udp)`.

PR #12385, opened 2025-01-17
[udp] Migrate to input package

taylor-swanson:
## Proposed commit message

- Migrate package from integration type to input.

## Checklist

```
cd packages/udp
elastic-package test
```

## Testing Notes

- I performed manual tests from the current version (`1.19.1`) and ensure that both the configuration from the previous version was properly migrated to `2.0.0`. I also verified that there was no disruption in handling incoming data.

Screenshot 2025-01-16 at 3 29 41 PM

## Related issues

- Relates #8435


packages/udp/_dev/test/system/test-syslog-config.yml
```diff
@@ -0,0 +1,8 @@
+service: test-syslog
+service_notify_signal: SIGHUP
+input: udp
+vars:
```
qcorporation, line 4: I see in other integrations it's more like the below. Is there a difference?
```
data_stream:
  vars:
    listen_address: 0.0.0.0
    listen_port: 9515
    preserve_original_event: true
```

packages/udp/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: "2.0.0"
```
qcorporation, line 1: This is a breaking change, hence the major version bump?

packages/udp/_dev/test/system/test-syslog-config.yml
taylor-swanson: Yes, switching to an input package means we no longer use data streams directly. An oddity with elastic-package system test is that it will assume the name of the dataset based on the package name and policy template (udp.udp in this case). However, since we use `udp.generic` by default, I have to override this in the system test config. One of them I use a custom name, just to test that. This is non-issue outside of the system tests (I did manual tests for upgrades as well).

packages/udp/changelog.yml
taylor-swanson: Major version bump for the package type change. I don't think it qualifies as a "breaking change" so much as the upgrade seems to pull configuration over correctly and there aren't any disruptions that I saw in my testing, hence just an "enhancement" as the changelog entry type. Other migrations for input types did the same thing (http_endpoint, for example).

taylor-swanson:
Added a note on my testing in the description, with screenshot of the config migration screen.

qcorporation (approved):
Thanks for the responses, LGTM

PR #12393, opened 2025-01-17
akamai - fingerprint event.original instead of requestId

andrewkroh:
## Proposed commit message

```
Fix the document _id calculation to allow events with the same requestId
to be indexed.

The httpMessage.requestId field in Akamai SIEM events is not unique to all
events based on observations of the data. It was observed that a given request
ID could have multiple records. Among three records with the same requestId,
one variant had httpMessage.bytes=0 and another had a different
httpMessage.responseHeaders with "Server: Akamai Image Server...".

I prefixed the document _id value with the timestamp based on recommendations
from the "Efficient Duplicate Prevention for Event-Based Data in Elasticsearch"
blog post.

References

- https://www.elastic.co/blog/efficient-duplicate-prevention-for-event-based-data-in-elasticsearch.
- https://github.com/akamai/akamai-apis/blob/b106ff167cb582efdaf4aa185c962ab33d3d6dee/apis/siem/v1/schemas/siem-event-200.yaml#L260-L263
```

### DIFF
--- a/packages/akamai/changelog.yml
+++ b/packages/akamai/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.27.2"
+  changes:
+    - description: Events with the same requestId are now properly indexed. Previously, multiple records with the same requestId could conflict (and be dropped) due to variations in other fields like httpMessage.bytes or httpMessage.responseHeaders.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/12393
 - version: "2.27.1"
   changes:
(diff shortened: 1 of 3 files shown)


kcreddy (approved):
LGTM.

We could possibly use only the [`httpMessage`](https://github.com/akamai/akamai-apis/blob/b106ff167cb582efdaf4aa185c962ab33d3d6dee/apis/siem/v1/schemas/siem-event-200.yaml#L199) part of the message, unless variations could be possible in other parts of the message too.

PR #12407, opened 2025-01-21
Add support of extra mongodb collstats metrics

### LINKED ISSUE #12406: Add new MongoDB collstats metricset properties
This task is `integrations` part of the job. `beats` part was implemented in [this issue](https://github.com/elastic/beats/issues/42003).

The new mongodb collstats properties that have been added in `beats` are:
- collstats.stats.size
- collstats.stats.count
- collstats.stats.avgObjSize
- collstats.stats.storageSize
- collstats.stats.totalIndexSize
- collstats.stats.totalSize
- collstats.stats.max
- collstats.stats.nindexes

Now that the properties are captured in `beats` we need to handle the properties in `integrations` as well.

stefans-elastic:
## Proposed commit message

Add support of extra mongodb collstats metrics (size, count, avgObjSize, storageSize, totalIndexSize, totalSize, max, nindexes)

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #12406

## Screenshots

Screenshot 2025-01-21 at 1 32 08 PM


packages/mongodb/data_stream/collstats/fields/fields.yml
```diff
@@ -110,3 +110,38 @@
...
+          description: >
+            The number of objects or documents in this collection.
+        - name: stats.avgObjSize
+          type: long
+          description: >
+            The average size of an object in the collection (in bytes).
+        - name: stats.storageSize
+          type: long
+          description: >
+            The total amount of storage allocated to this collection for document storage (in bytes).
```
gpop63, line 131: You could add `metric_type: gauge` and `unit: byte`. Also field namings could be improved check naming conventions: https://www.elastic.co/guide/en/beats/devguide/current/event-conventions.html

stefans-elastic:
> @stefans-elastic not sure in which release this feature ended up in metricbeat but you will have to bump the kibana version to that one.

I don't think `beats` part got released already (maybe I didn't need to be so quick with this PR. Do you think I should move this PR to `Draft` until `beats` gets a new release?

shmsr:
> > @stefans-elastic not sure in which release this feature ended up in metricbeat but you will have to bump the kibana version to that one.
> 
> I don't think `beats` part got released already (maybe I didn't need to be so quick with this PR. Do you think I should move this PR to `Draft` until `beats` gets a new release?

@stefans-elastic Your changes will be able in next minor release of beats i.e., 8.18. So, if you set kibana.version in top-level manifest.yml correctly, then you can also merge it now. This change will only be available when some user installs the 8.18 or above stack; and should choose to install this specific version of package.

packages/mongodb/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: "1.16.1"
```
ishleenk17, line 1: 1.17.0 since its an enhancement

packages/mongodb/data_stream/collstats/fields/fields.yml
```diff
@@ -110,3 +110,38 @@
       metric_type: counter
       description: |-
         Number of database commands executed.
+    - name: stats
+      type: group
+      fields:
+        - name: stats.size
```
ishleenk17, line 116: If the group has stats, the field name can be size, count. We need not append stats.* to the field name for a stats group.

packages/mongodb/manifest.yml
```diff
@@ -1,6 +1,6 @@
 name: mongodb
 title: MongoDB
-version: "1.16.0"
+version: "1.16.1"
```
ishleenk17, line 3: We should update the Kibana version as well to the version from which the change is available in beats.
Is the change backported to 8.17.x and 8.18 ? We should change Kibana version accordingly

ishleenk17:
> > @stefans-elastic not sure in which release this feature ended up in metricbeat but you will have to bump the kibana version to that one.
> 
> I don't think `beats` part got released already (maybe I didn't need to be so quick with this PR. Do you think I should move this PR to `Draft` until `beats` gets a new release?

8.18 release is quite late. 
We can backport it to 8.17.x as well and then we will be able to merge this PR sometime next month.

stefans-elastic:
> > > @stefans-elastic not sure in which release this feature ended up in metricbeat but you will have to bump the kibana version to that one.
> > 
> > 
> > I don't think `beats` part got released already (maybe I didn't need to be so quick with this PR. Do you think I should move this PR to `Draft` until `beats` gets a new release?
> 
> 8.18 release is quite late. We can backport it to 8.17.x as well and then we will be able to merge this PR sometime next month.

trying to backport to 8.17 (https://github.com/elastic/beats/pull/42358). I think we should be able to include this in 8.17.2 release

packages/mongodb/data_stream/collstats/fields/fields.yml
stefans-elastic: >> You could add metric_type: gauge and unit: byte.
is this only for `stats.storageSize` property?

>> Also field namings could be improved
don't names have to be the same as they are in `beats`? Also the names match naming of the fields returned by mongo's collstats command - https://www.mongodb.com/docs/manual/reference/command/collStats/#output

packages/mongodb/manifest.yml
```diff
@@ -14,7 +14,7 @@ icons:
 format_version: "3.0.2"
 conditions:
   kibana:
-    version: "^8.13.0"
+    version: "^8.17.2 || ^9.0.0"
```
stefans-elastic, line 17: I've noticed that quite a few packages have `|| ^9.0.0` in the condition. Correct me if we don't need it in this package

packages/mongodb/changelog.yml
stefans-elastic: done

packages/mongodb/data_stream/collstats/fields/fields.yml
stefans-elastic: done

packages/mongodb/manifest.yml
stefans-elastic: updated

stefans-elastic:
> Can we add a commit message in the PR description to maintain PR hygiene?

added

packages/mongodb/manifest.yml
```diff
@@ -1,6 +1,6 @@
 name: mongodb
 title: MongoDB
-version: "1.16.0"
+version: "1.17.0-next"
```
mrodm, line 3: It's not allowed to publish "next" versions. 

I cannot find the documentation, now but it was here the steps previously
https://github.com/elastic/integrations/pull/9796/files#diff-7b9748c4f848852d8b40b331e43f9fd58560e923d065183d0eb3e5916d831c9bL146

Should it be set again to 1.16.0 here (keeping the changelog "-next" entry)? @ishleenk17 @jsoriano

mrodm: The same happened in this PR https://github.com/elastic/integrations/pull/12407

That new version of the "iis" package was not published, since it is not allowed to publish "-next" versions.

stefans-elastic: to  clarify: in manifest it should be `1.16.0` and in changelog it should be `1.17.0-next`, right?

packages/mongodb/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: "1.17.0-next"
```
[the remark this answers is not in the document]
jsoriano: Appending the `-next` version is an approach that can be used to introduce a change without publishing a version. This can be used to delay the decision on publishing a package, for example if a set of related changes are introduced on different PRs.

This is not linked to the stack version, a package can be published for an stack version that doesn't exist yet.

packages/mongodb/manifest.yml
ishleenk17: > The same happened in this PR #12407
> 
> That new version of the "iis" package was not published, since it is not allowed to publish "-next" versions.
> 
> Maybe you would need to update the manifest also for that other package ("iis" package).
> 
> WDYT ? @ishleenk17 @jsoriano

You wanted to point out to another PR here for IIS  @mrodm ?

packages/mongodb/manifest.yml
ishleenk17: > I've noticed that quite a few packages have `|| ^9.0.0` in the condition. Correct me if we don't need it in this package

We are doing a bulk change of adding support of 9.0.0 . So you can remove that from this PR.

ishleenk17: > Even if the new fields are only available from 8.17, they probably don't do any harm on 8.13, and this way you can more easily keep maintaining support for versions from 8.13 to 8.17.

@jsoriano : In case we don't upgrade to 8.17.2, we will have new fields which actually don't exist for stack version lower than this but will still be shown in the README of the Integration. The user sitting on 8.13 might then also expect those fields which can't be supported. WDYT ?

packages/mongodb/manifest.yml
ishleenk17: If we use next in this PR and merge it. I don't think we will be able to publish any other change to the mongodb package until this change is published.
There are other mongodb changes ongoing in parallel which would be merged soon. 
So, IMO lets wait to push this PR until release of 8.17.2.

stefans-elastic: I've removed `-next`

packages/mongodb/manifest.yml
stefans-elastic: > We are doing a bulk change of adding support of 9.0.0 . So you can remove that from this PR.

removed 9.0.0 from the PR

packages/mongodb/manifest.yml
jsoriano: Why does this change need 8.17.2? It is only adding support for more fields, it should be safe to use in older versions.

stefans-elastic:
since the [PR](https://github.com/elastic/beats/pull/42358) in beats to backport this to 8.17 has been finally merged I think it is safe to assume the feature is going to be available in 8.17.2. So I think we can go ahead and approve/merge this PR (I'm pretty sure I've addressed all PR comments. If you notice something I haven't addressed - please let me know). WDYT?

packages/mongodb/manifest.yml
ishleenk17: @stefans-elastic : Lets keep the version as 8.13.
We should figure out a way to depict in the fields on how to showcase that a particular set of fields are available only from 8.17.2 onwards. 
But the kibana version can remain unchanged.

stefans-elastic: @ishleenk17 done

jsoriano (approved):

ishleenk17 (approved):
Looks good!

PR #12417, opened 2025-01-21
[MS Exchange Server] Added fields to ignore if empty.

matthewscherer:
## Proposed commit message
A customer came across a couple log lines with missing networkmessageid and senderaddress fields. This in turn rejected the logs messages. Allowing these two fields to be empty or otherwise missing will allow them to pass the ingest function. 

## Checklist

## Author's Checklist
- [ ] 

## Related issues

- Relates https://github.com/elastic/sdh-beats/issues/5552


andrewkroh:
> changes lgtm but it requires manifest.yml version bump and the changelog.yml

My recommended way to do this is to use

```
elastic-package changelog add \
--link https://github.com/elastic/integrations/pull/12417 \
--type bugfix \
--next patch \
--description 'Handle events where `networkmessageid` or `senderaddress` are not present.'
```

packages/microsoft_exchange_server/data_stream/messagetracking/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -51,6 +51,7 @@ processors:
 - set:
     field: email.local_id
     copy_from: microsoft.exchange.networkmessageid
+    ignore_empty_value: true
 - append:
     field: email.sender.address
     value: "{{{microsoft.exchange.senderaddress}}}"
```
andrewkroh, line 57: According to ECS, [email.sender.address](https://www.elastic.co/guide/en/ecs/current/ecs-email.html#field-email-sender-address) should be a scalar. So I think we should have

```yaml
- set:
    field: email.sender.address
    copy_from: microsoft.exchange.senderaddress
    ignore_empty_value: true
```

This also handles the case where senderaddress is empty.

packages/microsoft_exchange_server/data_stream/messagetracking/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -71,6 +72,7 @@ processors:
     ignore_failure: true
```
andrewkroh, line 72: The above append processor needs to handle the case where senderaddress is emtpy. The current `if` condition looks wrong.

```yaml
- append:
    field: email.from.address
    value: "{{{microsoft.exchange.senderaddress}}}"
    if: ctx.microsoft?.exchange?.senderaddress != null && ctx.microsoft.exchange.senderaddress != ""
```

andrewkroh (commented):
This change needs an accompanying test case where a new log sample{s} with empty senderaddress/networkmessageid is added to a packages/microsoft_exchange_server/data_stream/messagetracking/_dev/test/pipeline/*.log file.

matthewscherer:
This PR was replaced with this one. https://github.com/elastic/integrations/pull/12846

PR #12421, opened 2025-01-22
[GCP VertexAI] Remove zone filter and fix regions

muthu-mps:
This PR removes the zone filter and fixes the regions filter default value.

## Checklist

## Related issues

- 

## Screenshots

Screenshot 2025-01-22 at 12 40 58 PM

Screenshot 2025-01-22 at 12 40 43 PM

### DIFF
--- a/packages/gcp_vertexai/changelog.yml
+++ b/packages/gcp_vertexai/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Remove zone and fix default value in regions filter.
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/12420
+      link: https://github.com/elastic/integrations/pull/12421
 - version: "0.3.0"
   changes:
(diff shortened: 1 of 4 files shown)


ishleenk17 (approved):
Looks good!

shmsr:
@muthu-mps But why did we remove the zone filter?

muthu-mps:
> @muthu-mps But why did we remove the zone filter?

The zone filter was removed to make it similar to what we have in other cloud providers and the zone comes under the region where we decide to keep only region filter.

cc: @ishleenk17

shmsr:
Oh ok

ishleenk17:
Moreso, we are now using `regions` (array) for taking the region as input from user. 
Having a singular zone and multiple regions doesn't make sense together.

We can have `region and zone` together as input configs but not `regions and zone`.

Issue #12420, opened 2025-01-22
[Amazon Bedrock] Guardrails dashboard improvements


agithomas:
This issue is created to address the below mentioned inputs

1. [Input set 1](https://github.com/elastic/integrations/pull/12332#issuecomment-2601842898)
2. [Input set 2](https://github.com/elastic/integrations/pull/12332#issuecomment-2602322175)
3. Other dashboard panel rearrangements.

agithomas:
PR link : https://github.com/elastic/integrations/pull/12422

PR #12422, opened 2025-01-22
Add Amazon Bedrock Guardrails Improvements

agithomas:
- Enhancement

## Proposed commit message

Add improvements to the Bedrock overview dashboard and Guardrails dashboard

## Checklist

## Author's Checklist

- [x] Integration testing
- [x] Screenshot rendering testing  

- elastic-package build
- elastic-package stack up -v -d --services package-registry
- Perform package upgrade

## Related issues

- Closes https://github.com/elastic/integrations/issues/12420 

## Screenshots

**Amazon Bedrock Overview dashboard**

aws_bedrock_invocation
aws_bedrock_chat_image
aws_bedrock_logs

---

**Guardrails Overview Dashboard**
amazon_bedrock_guardrails_overview

amazon_bedrock_guardrails_invocation_details


muthu-mps (approved):
LGTM!

ishleenk17:
Are we using LLM or models in the charts ?

agithomas:
> Eg: "Overall latency by LLM" or "Overall Latency by model"

@daniela-elastic , the alternate option is to use "by foundation models". [Reference](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html). What do you suggest?

shmsr:
Yes, dashboard looks LGTM. Yes, let's make the titles FM or models instead of LLM. Rest, looks good.

ishleenk17:
@agithomas  We were reviewing the GCP vertex AI dashboard. Some feedback from AWS as part of those discussions.

1. For the Number of invocations that are intervened panel, the Y axis should be Invocations, instead of Interventions.
2. The model usage can be a donut instead of a pie chart, like in Vertex AI.

cc: @daniela-elastic

daniela-elastic:
For no1 yes it makes sense as long as we don't change the name of the widget - "umber of invocations where the guardrails intervened"
And yes, we agreed let's use donuts instead of pie charts

daniela-elastic:
I'm fine to call it "- by foundation model" (singular, not plural). Or even just "by model". Is there any reason why we need to specify that it's "foundation" or it can just be assumed?

ishleenk17:
> I'm fine to call it "- by foundation model" (singular, not plural). Or even just "by model". Is there any reason why we need to specify that it's "foundation" or it can just be assumed?

I think model should be a generic term that we can use across LLM Integrations.
@muthu-mps : In case Azure OpenAI needs such changes.

shmsr:
> I'm fine to call it "- by foundation model" (singular, not plural). Or even just "by model". Is there any reason why we need to specify that it's "foundation" or it can just be assumed?

The term "Foundation Models" originated in a 2021 paper (https://arxiv.org/pdf/2108.07258). While almost all models we are plotting here qualify as Foundation Models (FM), except most embedding models, because:

- Limited scope: They focus on data representations in vectors and act as components of FM (like a building block).
- No emergent behavior: While embedding models can capture complex relationships, they lack the emergent properties (performing arithmetic, reasoning, answering questions, summarizing passages seen in language model, and more).
- Limited adaptability: They serve as building blocks for FMs but aren't typically fine-tuned for various downstream tasks in the same way that a language model is.

But yes in case an embedding model starts showing these properties, then they come under FM.

Since "models" is a superset of "foundation models", using "model" as also suggested by Daniela seems a better choice for the dashboard - it looks cleaner while future-proofing our terminology for any new model types that may not exhibit foundation model properties.

muthu-mps:
@agithomas - The `i` icon to add description in the metrics chart is updated in the recent version. This doesn't include the icon by default. You have to hover over the chart to see the icon.  This is with the `8.17.1` release version. 
## Without hover
Screenshot 2025-01-28 at 11 09 57 AM

## Hover over the total invocations chart.
Screenshot 2025-01-28 at 11 10 03 AM

agithomas:
Presently, the amazon bedrock integration has 8.16.2 dependency. I think it may not be good to increment the version to `8.17.1` for a small visual improvement. I assume the users who upgrade to `8.17.1` stack will be able to take the benefit of the above feature

PR #12425, opened 2025-01-22
[integrations][trend_micro_vision_one] - Fixed inconsistent time interval issue leading to data loss in the detections data stream

ShourieG:
## Type of change
- Bug

## Proposed commit message
Fixed inconsistent time interval issue leading to data loss in the detections data stream. The cursor time has been updated to the "**eventTimeDT**" field of the 1st event for every response cycle leading to a smoother time interval without gaps. Events returned are always sorted in decreasing order of time, thus the 1st event of every response will always contain the latest event for that response.

## Checklist

## Author's Checklist

- [ ] 

## Related issues
- Supersedes https://github.com/elastic/integrations/pull/12382


ShourieG:
@chrisberkhout, could you take a look at this updated logic. I think with the recent info provided, this small change should be enough for the interval gap issue.

chrisberkhout (approved):
Good. Very nice commit message!

Maybe as a separate change, but it would be good the same thing for the `alert` and `audit` as well. They're less affected because they only a `last_update_at` cursor value for the beginning of a new pagination sequence, but the fix should be as easy there (although with different field names).

PR #12440, opened 2025-01-23
[cisco_asa] Add advanced option for time zone mapping and support parsing extra timestamp in header

### LINKED ISSUE #12442: [cisco_asa] Support mapping short tz names to IANA names/offsets
The date processor has limited to no support for short TZ names (EDT, CST, etc) due to a variety of issues with these identifiers (usage in Java is more or less deprecated, short names are often ambiguous, etc). Other integrations, such as Cisco IOS and Sophos, support mapping these short names to identifiers which are much more reliable, such as IANA names (America/Chicago) and UTC offsets (-06:00). Support for this mapping should be added to Cisco ASA as well, as logs may have short name identifiers.

taylor-swanson:
## Proposed commit message

- Added an advanced option for mapping short tz names to long tz names
- Support parsing the extra timestamp that is sometimes included in messages, often containing the time zone.

## Checklist

```
cd packages/cisco_asa
elastic-package test
```

## Related issues

- Closes #12442

## Screenshots

Screenshot 2025-01-23 at 8 42 25 AM


packages/cisco_asa/data_stream/log/_dev/test/pipeline/test-additional-messages.log-expected.json
```diff
@@ -11577,7 +11577,6 @@
                 "original": "Sep 25 15:47:07 host1.example.com : Sep 25 15:47:07 EDT: %ASA-auth-4-113004: AAA user authentication Successful : server = myservername : user = myusername",
                 "outcome": "success",
                 "severity": 4,
-                "timezone": "UTC",
```
taylor-swanson, line 11580: This is caused by two things:

- This log has a second timestamp with a time zone and is now being parsed.
- Since there is no mapping for `EDT`, it hits the fallback date parser and `event.timezone` gets removed. See: https://github.com/elastic/integrations/blob/main/packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml#L142

The correct solution in this case is to add the mapping for EDT, but leaving this in since it's an interesting scenario to test.

mjwolf (approved):

Issue #11911, opened 2024-11-28
microsoft_exchange_online_message_trace: indexing errors due to duplicate documents


efd6:
It seems that in some cases documents are not uniquely identified by the set of fields that are used to [fingerprint during ingest](https://github.com/elastic/integrations/blob/fc1c51338d9966a24b581f38f52e87a5d692db5f/packages/microsoft_exchange_online_message_trace/data_stream/log/elasticsearch/ingest_pipeline/default.yml#L290-L301)

- email.from.address
- email.to.address
- email.message_id
- email.local_id
- email.subject
- event.outcome
- email.attachments.file.size

We need to investigate the what set makes a reasonable identification for unique documents (`event.original` has been suggested, but we should see if we can get away with a smaller signal if possible).

jvalente-salemstate:
Are you able to tell if the duplicate documents have differing delivery statuses? That one bit is not included in the fingerprint, while it's the one that is likely changing and causing duplicate documents. 

These 3 may be enough to identify each uniquely:

- email.local_id
- email.to.address
- microsoft.online_message_trace.Status

The first two are generally "good enough" for identifying a unique message but the message trace may also have Pending/Getting Status as an initial status before being updated to Delivered/Quarantined/etc.

efd6:
Thanks for that suggestion @jvalente-salemstate.

PR #8717, opened 2023-12-13
[Exchange Online Message Trace] Add Additional Look-back Time & Fix Cursor Value

BenB196:
Enhancement & Bug

## Proposed commit message

This PR makes two changes to the Exchange Online Message Trace integration:

~1. Performs a refactor of `httpjson.yml.hbs` file to move the `request.*` & `response.*` parts together, or inversely moving all the fields/tags stuff towards the end of the file. (making it easier to follow from my perspective)~
1. Fixes an issue with how the value for `last_execution_datetime` cursor was being set.
    - Previously, the cursor value was using `now`, the issue with this is that the cursor sets it value once ["all events ... are published"](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-httpjson.html#cursor). This meant that if a cycle took five minutes to complete, there could be a five minute gap in time/data between the last iteration's `EndDate` and the next iteration's `StartDate`. This appears to have always been an "issue" but became more noticeable with #8550 where it changed the default interval from `1m` to `1h`, making it take far longer to process an iteration and making the gap in data more noticable.
    - ~To "fix" this issue, rather than using `now`, we take the `.last_event.EndDate` as the value. This should ensure that the next iteration always starts where the last one ended.~
    - ~Side Note: This also fixes another issue, for if Microsoft ever has an issue publishing Message Traces. Before this change, the integration would continue on as if there was no problem and leave a potentially large gap in data. Now the integration will hold at the last published event and pick up once Microsoft fixes the event publishing.~
2. Adds the ability to configure an additional look-back time to ensure data is properly collected.
    - See: https://github.com/elastic/integrations/pull/8717#issuecomment-1859322741

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).

## Author's Checklist

[...]

PR #5159, opened 2023-02-01
microsoft_exchange_online_message_trace: map fields to ECS

efd6:
Also retain original fields in data stream object.

## What does this PR do?

This improves the ECS compliance of the package.

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).

## Author's Checklist

- [ ]

## Related issues

- Closes #5083

PR #12446, opened 2025-01-23
[microsoft_exchange_online_message_trace] Remove event.start/end

chrisberkhout:
## Proposed commit message

```
[microsoft_exchange_online_message_trace] Remove event.start/end

Don't set `event.start` and `event.end` to `StartDate` and `EndDate`,
because those are query parameters, not properties of the event itself.

Also, extend the description of the `additional_look_back` setting to
note that up to 24h may be necessary (according to the API
documentation[1]).

[1]: https://learn.microsoft.com/en-us/previous-versions/office/developer/o365-enterprise-developers/jj984335%28v%3doffice.15%29
```

## Discussion

Elsewhere I suggested standard solutions for pagination when there is:
1. a delay between data creation and data availability, and/or
2. an unstable a period after initial creation of the data.

<details>

image

</details>

I suggested switching from the look-back strategy (introduced in https://github.com/elastic/integrations/pull/8717) to the standard solutions, in order to reduce fetching of the same data multiple times (and reduce indexing failures due to duplicate `_id` values).

However, the standard solution for delayed data availability isn't possible, because there is no creation or update time associated with the log entries we receive. `StartDate` and `EndDate` are query parameters. In the [API documentation](https://learn.microsoft.com/en-us/previous-versions/office/developer/o365-enterprise-developers/jj984335%28v%3doffice.15%29) these fields are marked "[In]" rather than "[In/Out]" (input parameters and report output columns). We just see them echoed back in the response.

This PR cleans up incorrect population of ECS fields with `StartDate` and `EndDate`.

Shall we drop `microsoft.online_message_trace.StartDate` and `microsoft.online_message_trace.EndDate` as well?

The standard solution for an unstable period is complicated by the fact that this period can be quite long. In https://github.com/elastic/integrations/pull/8717 it was reported to usually be less than an hour, but some users on the Internet report delays of multiple hours and the API documentation says up to 24 hours.

So, regarding pagination and duplicate `_id` values, the choice is between:

* a long (up to 24h) delay for all data (by only fetching data once it's known to be there), or
* indexing rejections due to duplicate `_id` values (using the look-back strategy).

In this case I think the look-back strategy is not a bad one.

## Checklist

## Related issues

- Relates #11911

### DIFF
--- a/packages/microsoft_exchange_online_message_trace/changelog.yml
+++ b/packages/microsoft_exchange_online_message_trace/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.25.3"
+  changes:
+    - description: Remove incorrect use of `event.start`, `event.end`.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/12446
 - version: "1.25.2"
   changes:
     - description: Fix handling of SenderAddress where the value is '<>'.
--- a/packages/microsoft_exchange_online_message_trace/data_stream/log/_dev/test/pipeline/test-log.log-expected.json
+++ b/packages/microsoft_exchange_online_message_trace/data_stream/log/_dev/test/pipeline/test-log.log-expected.json
@@ -58,10 +58,8 @@
                 "category": [
                     "email"
                 ],
-                "end": "2022-09-05T21:46:46.420Z",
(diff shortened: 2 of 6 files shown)


ShourieG (approved):
LGTM from my end but I think we should wait for @efd6 to review since he has more context with regards to this [PR](https://github.com/elastic/integrations/pull/5159/files) that introduced these fields.

PR #12456, opened 2025-01-24
[O365_metrics]Split and refactor other data streams.

ritalwar:
## Proposed commit message

This PR includes the following updates:  
1. Splits the `mailbox_usage` datastream into two separate datastreams: `mailbox_usage_detail` and `mailbox_usage_quota_status`.  
2. Renames the `outlook_app_usage` datastream to `outlook_app_usage_version_counts`.  
3. Renames the `yammer_device_usage` datastream to `viva_engage_device_usage_user_counts`.  
4. Refactors datastreams to use `report_date` as the timestamp when available; otherwise, defaults timestamp to `report_refresh_date`.  
5. Adds fingerprinting over dimension fields to handle duplicates. 
6. Refactors datastreams to explicitly convert fields to long in the pipeline and makes related adjustments.  
7. Updates the Codeowner file.

## Checklist


packages/o365_metrics/data_stream/outlook_activity/manifest.yml
```diff
@@ -13,7 +13,15 @@ streams:
         description: The interval at which the API is polled, supported in seconds, minutes, and hours.
         show_user: true
         required: true
-        default: 3m
+        default: 12h
```
ishleenk17, line 16: IIRC, didn't we decide this interval as 24 hrs ?

packages/o365_metrics/data_stream/outlook_app_usage_version_counts/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,192 @@
...
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - rename:
+      field: o365.metrics.outlook.app.usage.version.counts.outlook_2007
+      target_field: o365.metrics.outlook.app.usage.version.counts.outlook_2007.count
+      ignore_missing: true
+  - convert:
+      field: o365.metrics.outlook.app.usage.version.counts.outlook_2010
+      tag: convert_o365.metrics.outlook.app.usage.version.counts.outlook_2010
+      type: long
```
ishleenk17, line 60: By default are these metrics coming as integer through O365 ?
Why the need to convert them to long ?

ritalwar: Field values are coming as strings by default; therefore, we need to explicitly convert them to long.

packages/o365_metrics/data_stream/outlook_activity/manifest.yml
ritalwar: Yes, but we noticed that reports occasionally get refreshed daily, so to prevent any data loss, we decided to call the API twice a day.

muthu-mps:
@ritalwar - When changing the name of the data stream, Is there a reason to keep outlook_app_usage_version_counts instead of outlook_app_usage_version**s_user**_counts as per the documentation [here](https://learn.microsoft.com/en-us/graph/api/reportroot-getemailappusageversionsusercounts?view=graph-rest-1.0&tabs=http#:~:text=Get%20versions%20user%20counts)? As this is `versions user count` why we are not including the user in the data stream name?

packages/o365_metrics/data_stream/mailbox_usage_detail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,262 @@
...
+        - remove:
+            field: o365.metrics.mailbox.usage.detail.deleted_item_count
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - rename:
+      field: o365.metrics.mailbox.usage.detail.deleted_item_count
+      target_field: o365.metrics.mailbox.usage.detail.deleted_item.count
+      ignore_missing: true
+  - convert:
```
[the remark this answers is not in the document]
ritalwar: I understand, but since not all fields need conversion and the selection is conditional, applying the `convert` processor selectively was the most feasible approach for now. Using Painless would require more complex logic, but we can definitely explore it for future optimizations.

ritalwar:
> @ritalwar - When changing the name of the data stream, Is there a reason to keep outlook_app_usage_version_counts instead of outlook_app_usage_version**s_user**_counts as per the documentation [here](https://learn.microsoft.com/en-us/graph/api/reportroot-getemailappusageversionsusercounts?view=graph-rest-1.0&tabs=http#:~:text=Get%20versions%20user%20counts)? As this is `versions user count` why we are not including the user in the data stream name?

I understand your concern. While we split the datastreams and provided detailed names based on the reports, we also aimed to keep them concise to avoid overly long names. Where possible, we shortened them without losing essential information. The [description](https://github.com/elastic/integrations/blob/6a73fb8abe8126603f86fd213c0f29601dc812c2/packages/o365_metrics/data_stream/outlook_app_usage_version_counts/manifest.yml#L7) clearly specifies what each datastream fetches. Similarly, we named  data stream `mailbox_usage_quota_status` instead of `mailbox_usage_quota_status_mailbox_counts` to keep it clear yet manageable.

muthu-mps (approved):
- For the users who has already installed 0.3.0 might have a different data stream(Viva engage) for yammer device usage data stream while upgrading to newer version. I hope this is okay as the integration is still in Beta.
- Change looks good to me otherwise!

ritalwar:
> * For the users who has already installed 0.3.0 might have a different data stream(Viva engage) for yammer device usage data stream while upgrading to newer version. I hope this is okay as the integration is still in Beta.

Yes, since it's in beta, it's always recommended to use the latest version. Given that version 0.4.0 includes some fixes also as mentioned in the changelog, they should update to it.

ishleenk17 (approved):
Looks good!

lucian-ioan (approved):

PR #12461, opened 2025-01-24
fixed build README to generate

flexitrev:
Documentation

## Proposed commit message
Fixed  build README to generate appropriate documentation. Was broken via https://github.com/elastic/integrations/pull/11761

PR #12464, opened 2025-01-24
Adding health report to logstash integration

flexitrev:
## Proposed commit message

Added data and dashboards including Logstash Health report

## Checklist

## Author's Checklist

- [ ] 

## Related issues

Closes https://github.com/elastic/integrations/issues/12465

### Node Health
Node Health

### Pipeline Health
Pipeline Health


packages/logstash/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.5.0"
+  changes:
+    - description: Adding data and dashboards from Logstash Health Report
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/12394
```
andrewkroh, line 6: I think you want 12464.

```suggestion
      link: https://github.com/elastic/integrations/pull/12464
```

strawgate (changes requested):

yaauie:
- On the "General Pipeline Information" section:
   - I would not expect `TERMINATED` pipelines to have load averages; the presence of the 0's and green color in those columns _visually_ tells me that everything is A-okay.
   - Is "Load" in this section short-hand for Worker Utilization? Although it's shorter, I'm wary of overloading the word "Load" to mean new things (from a dev-ops perspective, it already has a very specific meaning related to CPU instructions queued per CPU tick)
   - Can we color-code the whole row based on the "Status" column (and if so, do we have prior-art for color-sets that are less in-your-face than #00ff00 #999999 #00ffff #ff0000 )?
   - It _feels_ like this is a smash-up of our two initial probes, but what happens when we add more?
     - in our diagnostic tool, we can compare two windows, IIRC it is always "current" against a selectable baseline, which lets us reduce the number of columns.

strawgate:
> * I would not expect `TERMINATED` pipelines to have load averages; the presence of the 0's and green color in those columns _visually_ tells me that everything is A-okay.

@flexitrev With the above sections covering terminated pipelines, we could change the bottom section to be running pipelines and just filter out terminated pipelines.

> Can we color-code the whole row based on the "Status" column (and if so, do we have prior-art for color-sets that are less in-your-face than #00ff00 #999999 #ffff00 #ff0000 )?

I do not believe you can color-code the whole row based on a cell -- you can color code the cell based on its value, though. As for the colors themselves, the color sets are the EUI palettes that ship with Kibana and will follow any changes the kibana team makes to the values for the colors. In general I think we should stick with the built-in palettes for integrations we ship to customers.

flexitrev:
Added a filter to the Worker utilization by pipeline to exclude terminated pipelines. Should I exclude finished too?

Colors are expected to change with project borealis - If they don't automatically get picked up by us picking builtin colors, we'll update later. https://github.com/elastic/search-team/issues/8943

flexitrev:
pipeline_health
 Updated dashboard, simplified worker utilization by pipeline table, modified visualization names

packages/logstash/data_stream/health_report/agent/stream/stream.yml.hbs
```diff
@@ -0,0 +1,64 @@
...
+
+redact: 
+  fields: ~
+
+program: |
+  get(state.url)
+  .as(resp, bytes(resp.Body)
+  .decode_json().as(body,
+    [   
+        [{"event.dataset": "logstash.health_report",
```
strawgate, line 27: If we're not customizing the dataset to be specific to the document type then we don't need to populate it at all

Was having event.dataset set to node and pipeline causing an issue?

flexitrev: Yes, that was a reason for failing CI

robbavey:
Some great and helpful stuff here!

Some comments:

### Add Logstash Integration

`Metrics (Elastic Agent)` _and_ `Metrics (Stack Monitoring)` are both enabled by default. We should only enable one by default, and warn that only one of these should be enabled.

Screenshot 2025-01-29 at 4 00 37 PM

--

`Metrics (Elastic Agent)` only has Health Report enabled by default - is this intentional? Plus some "technial preview" language still present.
Screenshot 2025-01-29 at 4 03 04 PM

## Node Health

Is it intentional for `Problem Nodes` to include details of all nodes, regardless of whether they are problematic? 
Screenshot 2025-01-29 at 4 04 13 PM

## Pipeline Health

Should we update the y-axis of `Pipeline Status over time`, to show something like "Percentage of Pipelines", rather than count, as the units are percentages?
Also, should we consider Dropping Partial Intervals on the Date Histogram to avoid showing potentially misleading partial results? 
Screenshot 2025-01-29 at 4 08 02 PM

### Concerning Pipelines

Is it possible to include a "last reported state" and/or something like "last reported in this state" field:
  
eg: Last 15 minute view:
Screenshot 2025-01-29 at 4 11 01 PM

Last 30 minute view: 
Screenshot 2025-01-29 at 4 12 36 PM

Over a long enough time period, there is a decent chance of seeing a pipeline in a `loading` state, which may be misleading or alarming?

---

I'd also like to see drilldowns/links to the existing dashboards to help users track down where their bottlenecks could be

packages/logstash/_dev/build/docs/README.md
```diff
@@ -541,16 +541,113 @@ An example event for `node` looks as following:
...
+Users that prefer the Stack Monitoring UI should uncheck `Metrics (Elastic Agent)` and continue to use `Metrics (Stack Monitoring)`. 
 
 ### Fields and Sample Event
 
+#### Health Report
+This is the health report data set which drives the Health dashboard pages
+
+#### Example
+An example event for 'health_report' looks as following:
+```json
```
robbavey, line 559: This should use the field/event macros as was worked on in https://github.com/elastic/integrations/pull/12461/files

robbavey (changes requested):
Main comments left in body of PR

packages/logstash/_dev/build/docs/README.md
flexitrev: I think I've got it now

robbavey:
README.md makes sense now (should https://github.com/elastic/integrations/pull/12461 be closed or committed...) - but I'd like to see your thoughts on my [comments](https://github.com/elastic/integrations/pull/12464#issuecomment-2622878839) outside of code review

flexitrev:
@robbavey I updated the defaults, and removed erroneous experimental release on Plugin and Pipeline dataset 
image

I don't know how to disable datasets by default and would like to address that in a subsequent dot version to make this report available. The previous version without the health report had all three datasets enabled by default. 

The filter for problem nodes is any node whose status is not "green", so a node not reporting status would show up.

robbavey:
To disable datasets by default, add `enabled: false` to the manifest.yml in all the `data_streams` for that dataset - eg, to disable the `stack monitoring` setting:

```patch
--- a/packages/logstash/data_stream/node/manifest.yml
+++ b/packages/logstash/data_stream/node/manifest.yml
@@ -8,6 +8,7 @@ elasticsearch:
 streams:
   - input: logstash/metrics
     title: Logstash node metrics
+    enabled: false
     description: Collect Logstash node metrics
     vars:
       - name: period
diff --git a/packages/logstash/data_stream/node_stats/manifest.yml b/packages/logstash/data_stream/node_stats/manifest.yml
index 450f8f13bd..a7504cb578 100644
--- a/packages/logstash/data_stream/node_stats/manifest.yml
+++ b/packages/logstash/data_stream/node_stats/manifest.yml
@@ -9,6 +9,7 @@ streams:
   - input: logstash/metrics
     title: Logstash node stats metrics
     description: Collect Logstash node stats metrics
+    enabled: false
     vars:
       - name: period
         type: text
```

> The filter for problem nodes is any node whose status is not "green", so a node not reporting status would show up.
Add this to the filter for `Problem Nodes` and it should go away
Screenshot 2025-02-03 at 3 33 18 PM

strawgate:
It looks like we removed event.dataset prematurely -- it looks like CI was failing because sample-event.json had two values for event.dataset: 
https://buildkite.com/elastic/integrations/builds/21145#0194aee4-6d12-43ad-8f03-233889b509f1

[build/test-coverage/coverage-logstash-pipeline-1738101594676982809-report.xml](https://buildkite.com/organizations/elastic/pipelines/integrations/builds/21145/jobs/0194aee4-6d12-43ad-8f03-233889b509f1/artifacts/0194aeed-d8dc-4be5-aa78-32b43bbdb5e0)

```
<testcase name="static test: Verify sample_event.json" classname="logstash.health_report" time="0.115375683">
<failure>test case failed: one or more errors found in document: [0] field "event.dataset" should have value in ["logstash.health_report"], it has "logstash.pipeline"</failure>
</testcase>
```

In the sample_event.json when CI was failing there were two `event.dataset` values in the document: https://github.com/elastic/integrations/commit/13ed93d495d7cfdf6cc2ec0f46e6cda999f50d6b

flexitrev:
Thanks! I tried it at the top level manifest but not on the individual streams - 
image

flexitrev:
> It looks like we removed event.dataset prematurely -- it looks like CI was failing because sample-event.json had two values for event.dataset: https://buildkite.com/elastic/integrations/builds/21145#0194aee4-6d12-43ad-8f03-233889b509f1
> 
> [build/test-coverage/coverage-logstash-pipeline-1738101594676982809-report.xml](https://buildkite.com/organizations/elastic/pipelines/integrations/builds/21145/jobs/0194aee4-6d12-43ad-8f03-233889b509f1/artifacts/0194aeed-d8dc-4be5-aa78-32b43bbdb5e0)
> 
> ```
> <testcase name="static test: Verify sample_event.json" classname="logstash.health_report" time="0.115375683">
> <failure>test case failed: one or more errors found in document: [0] field "event.dataset" should have value in ["logstash.health_report"], it has "logstash.pipeline"</failure>
> </testcase>
> ```
> 
> In the sample_event.json when CI was failing there were two `event.dataset` values in the document: [13ed93d](https://github.com/elastic/integrations/commit/13ed93d495d7cfdf6cc2ec0f46e6cda999f50d6b)
> 
> ```
> "event": {
>   "agent_id_status": "verified",
>   "ingested": "2025-01-28T18:41:28Z",
>   "dataset": "logstash.health_report"
> },
> "event.dataset": "logstash.pipeline"
> }
> ```

@strawgate Yes, I thought it best to just keep logstash.health_report. Should I keep both somehow? Rename event.dataset:logstash.pipeline?

strawgate:
Ideally, there would be two sample events, the node sample event and the pipeline sample event?

flexitrev:
Cutting a new branch and creating a new PR - https://github.com/elastic/integrations/pull/12601

PR #12494, opened 2025-01-28
packages/openai: New OpenAI integration

shmsr:
## Proposed commit message

This is a new integration for OpenAI to collect usage metrics using OpenAI's newly launched usage API. It uses the admin key for programmatic access to usage API.

## Checklist

# How to test this PR locally

Run tests using elastic-package and manual verification with OpenAI usage dashboard (https://platform.openai.com/usage)

## Related issues

- Relates https://github.com/elastic/beats/pull/41516

## Screenshots

usage


shmsr:
Update: 28th Jan, 2025

I am still working on the dashboard but rest seems to be more or less done. Except dashboards, other things can be reviewed.

packages/openai/data_stream/audio_speeches/fields/fields.yml
```diff
@@ -28,6 +28,9 @@
         - name: end_time
           type: date
           description: End timestamp of the usage bucket
+        - name: object
```
agithomas, line 31: `object` being one of the types, could you consider a different name. eg: `object_type` or `usage_record_type`?

packages/openai/data_stream/audio_speeches/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -46,6 +46,10 @@ processors:
         field: openai.audio_speeches.num_model_requests
         target_field: openai.base.num_model_requests
         ignore_missing: true
+    - rename:
+        field: openai.audio_speeches.object
+        target_field: openai.base.object
```
agithomas, line 51: I am not sure if `base` is a great option here. Can `summary` be a better option? I assume the base represents the overall summary of metric within a timeslice.

shmsr: So, `base` holds all the common fields. Not summary.

For example, `project_id` is common across all data_streams.

agithomas:
I refer to the screenshot attached to the issue. Adding a few points for your consideration.  

1. Kindly align the axis label text similar to other gen-ai integration dashboards. It would be best, i think, to have an alignment with the color palette used as well. 

2. When we have a time-series dashboard, please see if there is a need to have "over time" in the panel title.

packages/openai/data_stream/audio_speeches/fields/fields.yml
shmsr: sure. usage_object_type?

shmsr:
@agithomas Yes. Still working on refinements of dashboard. From the issue that Daniela created and also incorporating the comments given by all in AWS Bedrock PR. Will be done soon!

shmsr:
Done!

packages/openai/data_stream/code_interpreter_sessions/sample_event.json
```diff
@@ -0,0 +1,33 @@
+{
+    "openai": {
+      "code_interpreter_sessions": {
+        "sessions": 16
+      },
+      "base": {
+        "start_time": "2024-09-04T00:00:00.000Z",
+        "project_id": "",
+        "end_time": "2024-09-05T00:00:00.000Z",
+        "usage_object_type": "organization.usage.code_interpreter_sessions.<dummy>"
```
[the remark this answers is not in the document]
shmsr: So, for code_interpreter_session, OpenAI's APIs are down as of now. Not working since 24h or more. So, that's why kept it like that for now to pass the tests.

packages/openai/data_stream/audio_speeches/sample_event.json
```diff
@@ -0,0 +1,37 @@
+{
+    "openai": {
+        "audio_speeches": {
```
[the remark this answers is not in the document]
shmsr: Nope. Here: https://platform.openai.com/docs/api-reference/usage/audio_speeches_object

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,138 @@
+# OpenAI
+
+The OpenAI integration allows you to monitor OpenAI API usage and performance. OpenAI is a leading AI platform providing various AI models and APIs for natural language processing, image generation, and other AI capabilities.
+
+Use the OpenAI integration to track API usage, costs, and performance metrics across your OpenAI implementations. Then visualize that data in Kibana, create alerts to notify you if usage limits are approaching, and view metrics when troubleshooting API issues.
+
+For example, if you wanted to monitor costs across different OpenAI models, you could track token usage and API calls per model. Then you can visualize various trends in Kibana dashboards, set up alerts for unusual usage spikes, or troubleshoot by analyzing the number of requests.
```
daniela-elastic, line 7: Change "For example, if you wanted to monitor costs across different OpenAI models, you could track token usage and API calls per model. Then you can visualize various trends in Kibana dashboards, set up alerts for unusual usage spikes, or troubleshoot by analyzing the number of requests."
TO:
"The integrations lets you see your metrics split across the different OpenAI models. For example, you can track token usage and API calls per model."

muthu-mps:
- The panel name `Model invocation frequency - by model` can be changed to `Invocation frequency - by model` WDYT?

packages/openai/data_stream/audio_speeches/manifest.yml
```diff
@@ -0,0 +1,73 @@
...
+        description: >
+          The request tracer logs requests and responses to the agent's local file-system for debugging configurations. Enabling this request tracing compromises security and should only be used for debugging. See [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-cel.html#_resource_tracer_filename) for details.
+      - name: processors
+        type: yaml
+        title: Processors
+        multi: false
+        required: false
+        show_user: false
+        description: >-
+          Processors are used to reduce the number of fields in the exported event or to enhance the event with metadata.  This executes in the agent before the logs are parsed. See [Processors](https://www.elastic.co/guide/en/beats/filebeat/current/filtering-and-enhancing-data.html) for details.
```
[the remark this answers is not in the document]
muthu-mps: @shmsr - This change should be done for rest of the data streams as well.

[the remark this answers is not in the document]
shmsr: Sure.

shmsr:
> * The panel name `Model invocation frequency - by model` can be changed to `Invocation frequency - by model` WDYT?

Yes, makes sense. Thanks!

packages/openai/data_stream/audio_speeches/manifest.yml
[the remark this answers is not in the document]
shmsr: Done

muthu-mps (approved):
LGTM!

ishleenk17:
@shmsr : I'll be reviewing the code today. 
Couple starter comments looking at the dashboard:

1. Do customers benefit from knowing about API Keys, Project ID's, object types in the dashboard ? Is it worth putting this info there? This info being available in the documents should suffice ?
2. First 4 panels (invocation frequesncy, token usage....) are generic across models ?
3. The timestamp is per day, hope that's a discussed value.
4. We can remove the word "Dashboard" in the title of the dashboard. Just have "OpenAI".

shmsr:
> @shmsr : I'll be reviewing the code today. Couple starter comments looking at the dashboard:
> 
> 1. Do customers benefit from knowing about API Keys, Project ID's, object types in the dashboard ? Is it worth putting this info there? This info being available in the documents should suffice ?
> 2. First 4 panels (invocation frequesncy, token usage....) are generic across models ?
> 3. The timestamp is per day, hope that's a discussed value.
> 4. We can remove the word "Dashboard" in the title of the dashboard. Just have "OpenAI".

1. See the screenshot. So, OpenAI usage dashboard does contain top user, projects, api keys (they have mapping of id -> names, that's why they show names; but we have the ids that they can map or add a processor with add those mapping in the document and make a visualization out of it) that are using and also it seems a good metric to know who are top n users.
Screenshot 2025-01-31 at 12 00 29 PM
2. Tokens is not a common concept for all models. Like for image models, no concept of tokens. But it is placed there to maintain consistency across Elastic's LLM integrations. But later, you see the proper seggragation — token usage, image model, audio speeches, etc.
3. Yes. timestamp is per day wise, because, OpenAI themselves keep the bucket_width of daily. That is data is aggregated over a day; it is configurable though like minute wise, and hour wise. But also the older API used to give daily data. In their dashboard as well, it's daily data. So, kept it like that only.
4. Sure, on it!

ishleenk17:
To maintain consistency with other LLM Integrtaions, the agrregated values can be shown for 
Total Tokens, Total Invocations. We need not have input output divisioning here in the top panel?
Also we already have them in the Token usage where we are showing input and output.


Do you think its better to have the name as "Overall Invocation Frequency" for the topmost generic panel ?
Since thats the cumulated one. And this is a stacked graph, right ?

packages/openai/data_stream/audio_speeches/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,94 @@
+config_version: 2
+interval: {{interval}}
+{{#if enable_request_tracer}}
+resource.tracer.filename: "../../logs/cel/openai-audio-speeches-http-request-trace-*.ndjson"
+resource.tracer.maxbackups: 5
```
ishleenk17, line 5: For my knowledge, what is this used for ?

packages/openai/data_stream/audio_speeches/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,94 @@
...
+          "initial_start_time": state.?cursor.last_bucket_start.orValue(int(timestamp(now - duration(state.initial_interval))))
+        })
+    ).as(state,
+      request(
+        "GET",
+        state.url + "?" + {
+          "start_time": [string(int(state.initial_start_time))],
+          "page": state.page != null ? [state.page] : [],
+          "bucket_width": ["{{ bucket_width }}"],
+          "group_by": ["project_id,user_id,api_key_id,model"]
```
ishleenk17, line 41: Is it like saying if you would have had these fields in the document, these would have become your dimension fields to avoid duplication ?

packages/openai/data_stream/audio_speeches/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,94 @@
...
+          "page": null,
+          "initial_start_time": state.?cursor.last_bucket_start.orValue(int(timestamp(now - duration(state.initial_interval))))
+        })
+    ).as(state,
+      request(
+        "GET",
+        state.url + "?" + {
+          "start_time": [string(int(state.initial_start_time))],
+          "page": state.page != null ? [state.page] : [],
+          "bucket_width": ["{{ bucket_width }}"],
```
ishleenk17, line 40: Say customer changes this aggregation bucket width to say 1m, our dashboards are 1 day timestamp. 
Not sure but will that hinder experience for user.
Maybe not, but thoughts ?

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,136 @@
+# OpenAI
+
+The OpenAI integration allows you to monitor OpenAI API usage and performance. OpenAI is a leading AI platform providing various AI models and APIs for natural language processing, image generation, and other AI capabilities.
+
+Use the OpenAI integration to track API usage, costs, and performance metrics across your OpenAI implementations. Then visualize that data in Kibana, create alerts to notify you if usage limits are approaching, and view metrics when troubleshooting issues.
```
alaudazzi, line 5:
```suggestion
With the OpenAI integration, you can track API usage, costs, and performance metrics across your OpenAI implementations. You will use Kibana to visualize your data, create alerts if usage limits are approaching, and view metrics when you troubleshoot issues.
```

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+The integrations lets you see your metrics split across the different OpenAI models. For example, you can track token usage and API calls per model.
+
```
alaudazzi, line 8:
```suggestion
The OpenAI integration lets you monitor your metrics across the different OpenAI models. For example, you can track token usage and API calls per model.

```

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,136 @@
...
+## Data collection
+
+The OpenAI integration collects comprehensive usage data through the [OpenAI Usage API](https://platform.openai.com/docs/api-reference/usage). The Usage API provides detailed insights into your activity across the OpenAI API and helps you understand how your organization is using OpenAI APIs.
+
+## Data streams
+
+The OpenAI integration collects the following logs data streams:
+
+Data streams:
+
```
alaudazzi, line 18: Remove this line

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,136 @@
...
+### Audio speeches
+
+The `audio_speeches` data stream captures audio speeches usage events for the organization.
+
+{{event "audio_speeches"}}
+
+**ECS Field Reference**
+
+Please refer to the following [document](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) for detailed information on ECS fields.
```
alaudazzi, line 50:
```suggestion
Refer to this [document](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) for detailed information on ECS fields.
```

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Audio transcriptions
+
+The `audio_transcriptions` data stream captures audio transcriptions usage events for the organization.
+
+{{event "audio_transcriptions"}}
+
+**ECS Field Reference**
+
+Please refer to the following [document](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) for detailed information on ECS fields.
```
alaudazzi, line 62: This paragraph is repeated 7 times. It is probably enough to keep it only once, at the top of the section?

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,136 @@
...
+- `vector_stores`: Collects vector stores usage events for the organization.
+
+See more details for data streams in the [Logs](#logs-reference).
+
+## Requirements
+
+You need Elasticsearch for storing and searching your data and Kibana for visualizing and managing it.
+
+You need an OpenAI account to access the [Usage API](https://platform.openai.com/docs/api-reference/usage) with a valid [Admin key](https://platform.openai.com/settings/organization/admin-keys). Admin keys are being used for programmatic access to endpoints detailed in the [Usage API](https://platform.openai.com/docs/api-reference/usage).
+
```
alaudazzi, line 35: Do we need a Setup section here?

alaudazzi (commented):
I left a few comments, feel free to address them where appropriate.

packages/openai/data_stream/audio_speeches/agent/stream/cel.yml.hbs
shmsr: For debugging. Suppose there's some issue. So, if you enable this feature, you can see the raw request made by CEL execution and also the response.

```
<request 1>
<response 1>
<request 2>
<response 2>
```

Helps in debugging.

packages/openai/data_stream/audio_speeches/agent/stream/cel.yml.hbs
shmsr: Let's suppose we consider this example: https://platform.openai.com/docs/api-reference/usage/completions_object

See the response?

Unless you group with `project_id,user_id,api_key_id,model`, the values for those fields is going to be empty as they are not grouped.

For example this is a real response when dont use group by:
```
{
  "object": "page",
  "data": [
    {
      "object": "bucket",
      "start_time": 1738195200,
      "end_time": 1738281600,
      "results": [
        {
          "object": "organization.usage.completions.result",
          "input_tokens": 22,
          "output_tokens": 124,
          "num_model_requests": 1,
          "project_id": null,
          "user_id": null,
          "api_key_id": null,
          "model": null,
          "batch": null,
          "input_cached_tokens": 0,
          "input_audio_tokens": 0,
          "output_audio_tokens": 0
        }
      ]
    },
    {
      "object": "bucket",
      "start_time": 1738281600,
      "end_time": 1738314727,
      "results": []
    }
  ],
  "has_more": false,
  "next_page": null
}
```

See `results`? They are nulls.

Yes, they can be dimension fields. But here, we did not enabled TSDB as earlier discussed. Historical data support is there, so we are not enabling the same. Also, we chose logs-* here as discussed earlier.

packages/openai/_dev/build/docs/README.md
shmsr: Added.

packages/openai/data_stream/audio_speeches/agent/stream/cel.yml.hbs
shmsr: Screenshot 2025-01-31 at 4 00 23 PM

So, yes it will be fine. See this screenshot for when bucket width is set to hour `1h`. As it x-axis timestamp per day, the viz still properly aggregates.

But yes, if someone set to `minute` i.e., `1m` and suppose collecting `6 months` of data, imagine the number of requests.

Calculate per data-stream:

- 6 months to minutes = 262800. So this how many buckets we have to process.
- Now when querying with bucket_width of `1m`, OpenAI by default, returns 60 buckets in 1 request. So, `262800/60` = 4380.

So for 8 data_streams, 35040 requests in such short time which will definitely trigger some rate limit. In fact I have tested this as well, it takes a lot of time as OpenAI blocks so much requests in such short time i.e., 35k requests.

I will put the rate limits in another next enhancement as it is not urgent and immediate requirement when going with default or even `1h`. But with `1m`, there's a huge amount of data to process, so rate limit would make sense there. But then we also expect users to understand that pulling so much historical data with such granularity is definitely a lot.

shmsr: But yes, giving the users the flexibility to pull data `1d`/ `1h` / `1m` makes sense. There might be users who want to make viz and want `1h` / `1m` of granularity.

shmsr:
@ishleenk17 Can you again take a look at the screenshot? I've addressed the comments.

ishleenk17:
1. We can have token usage by model. That's what we have done for other LLM's. Would help user to see which model consumes most tokens.
Screenshot 2025-01-31 at 4 39 41 PM

2. Also, the "i" (info) icons for Total tokens, Totla invocations.

packages/openai/data_stream/completions/fields/fields.yml
```diff
@@ -0,0 +1,56 @@
...
+        - name: batch
+          type: boolean
+          description: Whether the request was processed as a batch
+        - name: input_audio_tokens
+          type: long
+          description: Number of audio input tokens used, including cached tokens
+        - name: input_cached_tokens
+          type: long
+          description: Number of text input tokens that has been cached from previous requests. For customers subscribe to scale tier, this includes scale tier tokens
+        - name: input_tokens
```
ishleenk17, line 51: This includes only the text tokens, correct ?
So maybe input_text_tokens.
Currently to me the name suggested total input completion tokens.

packages/openai/data_stream/embeddings/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,91 @@
...
+        field: openai.embeddings.results
+        ignore_missing: true
+    # Add base OpenAI fields
+    - rename:
+        field: openai.embeddings.model
+        target_field: openai.base.model
+        ignore_missing: true
+    - rename:
+        field: openai.embeddings.project_id
+        target_field: openai.base.project_id
```
ishleenk17, line 35: Will the values of these base fields be same across datastreams ?

packages/openai/data_stream/embeddings/fields/fields.yml
```diff
@@ -0,0 +1,41 @@
...
+          description: Start timestamp of the usage bucket
+        - name: end_time
+          type: date
+          description: End timestamp of the usage bucket
+        - name: usage_object_type
+          type: keyword
+          description: Type of the usage record
+    # Embeddings-specific fields
+    - name: embeddings
+      type: group
```
ishleenk17, line 36: For couple of DS, there is only a single field per DS. 
The reason for having them as separate DS ?
Is it because we didn't want to have multiple API calls in a single CEL program ?

shmsr: > The reason for having them as separate DS ?

If you remember the Slack discussion. Gives user ability to turn off/on.

> Is it because we didn't want to have multiple API calls in a single CEL program ?

Yes, also this.

And also combining with other data_stream won't make sense.

For example, for completions data_steam, just last month only it used to give input and output tokens. Now they give input audio tokens as well as the text-generating models now accept audio as input as well. So, there's chance for each category to grow. So, keeping them separate makes sense.

packages/openai/data_stream/embeddings/elasticsearch/ingest_pipeline/default.yml
shmsr: Yes. Like project_id is there in all DS. But yes, vector_stores and code_interpreter_sessions are two special data_streams; they are not models. So, only a few base fields are there.

packages/openai/data_stream/completions/fields/fields.yml
shmsr: > Currently to me the name suggested total input completion tokens.

True.

> This includes only the text tokens, correct ?

Also, true.

Actual definition:

The aggregated number of text input tokens used, including cached tokens. For customers subscribe to scale tier, this includes scale tier tokens.

---

But, I think the name `input_tokens` is fine. We have not changed names for any and descriptions for are on point. Also almost everywhere it is referred to as "Input tokens".

shmsr:
> 1. We can have token usage by model. That's what we have done for other LLM's. Would help user to see which model consumes most tokens.
> 
> Screenshot 2025-01-31 at 4 39 41 PM
> 2. Also, the "i" (info) icons for Total tokens, Totla invocations.

@ishleenk17 

1. I did have the first one and then removed. But the thing is we have to rank then. With ranking, Kibana allows one field. So, for models like moderation and embedding, there's no output tokens (only input tokens). You can also see that in Vertex as well (it is 0). 

See: https://platform.openai.com/docs/api-reference/usage/embeddings_object
See: https://platform.openai.com/docs/api-reference/usage/moderations_object

So total tokens becomes a wrong metric. Although, it is possible to plot. But if it's there in other, let me plot it.

2. Sure

shmsr:
Dashboard screenshot updated in PR comment.

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,114 @@
+# OpenAI
+
+The OpenAI integration allows you to monitor OpenAI API usage and performance. OpenAI is a leading AI platform providing various AI models and APIs for natural language processing, image generation, and other AI capabilities.
```
daniela-elastic, line 3: Question: should we say "foundation models" instead of "AI models"?

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+With the OpenAI integration, you can track API usage, costs, and performance metrics across your OpenAI implementations. You will use Kibana to visualize your data, create alerts if usage limits are approaching, and view metrics when you troubleshoot issues.
+
+The OpenAI integration lets you monitor your metrics across the different OpenAI models. For example, you can track token usage and API calls per model.
+
+## Data collection
+
+The OpenAI integration collects comprehensive usage data through the [OpenAI Usage API](https://platform.openai.com/docs/api-reference/usage). The Usage API provides detailed insights into your activity across the OpenAI API and helps you understand how your organization is using OpenAI APIs.
+
```
daniela-elastic, line 12: are we only tracking how your org is using the APIs (i.e. usage) or are we also giving you an understanding of how those OpenAI models are performing in terms of latency, errors,  etc?

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,114 @@
...
+- `embeddings`: Collects embeddings usage events for the organization.
+- `images`: Collects images usage events for the organization.
+- `moderations`: Collects moderations usage events for the organization.
+- `vector_stores`: Collects vector stores usage events for the organization.
+
+See more details for data streams in the [Logs](#logs-reference).
+
+## Requirements
+
+You need Elasticsearch for storing and searching your data and Kibana for visualizing and managing it.
```
daniela-elastic, line 30: Please work with @alaudazzi  to make sure the requirements section presence (or absence) is aligned with all other  integration pages

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+You need an OpenAI account to access the [Usage API](https://platform.openai.com/docs/api-reference/usage) with a valid [Admin key](https://platform.openai.com/settings/organization/admin-keys) to programmatic access to Usage APIs.
+
```
daniela-elastic, line 33: There seems to be some unnecessary repetition in this sentence, can we simplify:
"You need an OpenAI account to access the Usage API with a valid Admin key to programmatic access to Usage APIs."
@alaudazzi  can you help please?

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,114 @@
...
+## Setup
+
+For step-by-step instructions on how to set up an integration, see the [Getting started](https://www.elastic.co/guide/en/starting-with-the-elasticsearch-platform-and-its-solutions/current/getting-started-observability.html) guide.
+
+### Generate an Admin key
+
+To generate an Admin key, please generate a key or use an existing one from the [Admin keys](https://platform.openai.com/settings/organization/admin-keys) page. Use the Admin key to configure the OpenAI integration.
```
daniela-elastic, line 40: Shouldn't this section be right next to the section that talks about needing an OpenAI account and Admin keys? We should decide of it will be before or after that section given might be a dependency. @alaudazzi  any thoughts on this?

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Collection behavior
+
+By default, the OpenAI integration fetches events with a bucket width of 1 day (`1d`), which means events are aggregated by day. Events are collected from the set initial start time until the current time except the bucket that is currently being used by OpenAI, that is the last bucket is skipped as it is not complete. So, based on configured bucket width, the integration collects events from the initial start time until the current time minus the bucket width.
```
daniela-elastic, line 44: is "bucket width" a standard term we use? Will customers understand what we mean? What do we say in other integrations and can we use the same language if it's different?

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Logs reference
+
+**ECS Field Reference**
+
+Refer to this [document](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) for detailed information on ECS fields.
+
+### Audio speeches
+
+The `audio_speeches` data stream captures audio speeches usage events for the organization.
```
daniela-elastic, line 54: Do we need to repeat "for the organization" everywhere? Isn't it assumed?  @alaudazzi  would it make sense to get rid of this?

daniela-elastic (commented):
I've added a bunch of questions and suggestions. Would welcome recommendations from @alaudazzi

packages/openai/_dev/build/docs/README.md
shmsr: I did take a look at many "Requirements" sections; looks similar.

packages/openai/_dev/build/docs/README.md
shmsr: Based on what I've seen in other READMEs:

- The Requirements section should only list the requirements, nothing more.
- The Setup section elaborates on how to perform the actual setup. Also, made "### Generate an Admin key" a subsection under "## Setup"

So, in Requirements I listed what's required, and in the Setup section, I explained how to generate one Admin Key.

Should I club them?

packages/openai/_dev/build/docs/README.md
shmsr: https://github.com/elastic/integrations/pull/12494#discussion_r1938502879

packages/openai/_dev/build/docs/README.md
shmsr: Removed this. Makes sense to not repeat it.

packages/openai/_dev/build/docs/README.md
shmsr: So, we do give flexibility to the users to change the "bucket_width". It is not our terminology but OpenAI's. Check the 3rd query parameter here: https://platform.openai.com/docs/api-reference/usage/completions

So, by default, we keep this as `1d` (one day). Infact, OpenAI's own dashboard works on daily aggregated data as well; no knobs there. But as their new API gives the flexibility to change it, that's why we have made it configurable.

So, concept of bucket I explained here because it is important to make the user understand how the collection works:

So, let's think about last 2 days (Feb 2, Feb 3). The bucket_width is set to `1d`, that is OpenAI will return aggregated data for a day. So, each bucket has a start and end timestamp i.e., for Feb 2, you have one bucket, and for Feb 3, another bucket.

Now, as today's Feb 3, the ended hasn't ended yet. So, if you collect data, as the bucket is incomplete as still filling, we won't be able to get the correct data. Like we can collect, but the aggregation on our side, and the visualizations will not match up with OpenAI's. So, incomplete buckets are left out. As soon as the bucket is closed and it is not being filled anymore, we collect it.

packages/openai/_dev/build/docs/README.md
shmsr: Yes, only usage. Latency, errors, etc. are not present.

See: https://github.com/elastic/observability-dev/issues/3904#issuecomment-2469711313 for more detailed explanation.

packages/openai/_dev/build/docs/README.md
shmsr: Yes, I updated it. Now it's very close to OpenAI's own website and very crisp.

Ref: 
- https://openai.com/about/
- https://openai.com/api
- etc.

agithomas:
What does this Top 10 User IDs stand for?

Does the count of records mean, number of sessions or invocations? It will add more clarity if a more appropriate panel title and column title are chosen.  Same for top 10 project IDs, top 10 API IDs. If there is no means to edit the title / column label, please ignore this feedback.


image

Same is the case with the below-mentioned visualisation. Does the count signify the invocations or token usage?

image



Can you see why the project id is empty in this screenshot?


image


I understand that size is the term when doing image generation. Should we keep the technical term here or rather a more apt `resolution` here?


image


Some of the graphs are having the term "frequency" and "distribution" in the y-axis. If it is a frequency, the unit of y-axis must be request / sec,requests /hr.  In here, this is just the count. So, removing the word "frequency" will be more apt. Similarly the word "distribution" shows how a range of values are distributed ("Token usage distribution - by model"). I doubt this is the case here. So, recommend to remove the word "distribution" from the y-axis. 

image


The x-axis label is missing from the below mentioned panel

image

agithomas:
@shmsr ,

When i refer to the manifest.yml, i note the title as
- Collect audio transcriptions usage details for the organization.
- Collect audio speeches usage details for the organization
- Collect code interpreter sessions usage details for the organization

and so on.

I understand that to create the admin key, the user has to visit the `https://platform.openai.com/settings/organization/admin-keys`

A couple of questions here?

1. Can the integration work with [Project API Key](https://platform.openai.com/settings/organization/api-keys) ?
2. Even though we are saying "for the organisation", and a given account can have [multiple organisations](https://platform.openai.com/docs/guides/production-best-practices) , the name of the organisation (or id) does not appear as a field. I think, the organization specific metrics would be a key customer requirement (a [sample requirement](https://community.openai.com/t/multiple-organizations-for-billing-purposes/81872)) . Does the API return the name / id of the organisation? If yes, kindly include it, add it as a part of input control drop down

shmsr:
> What does this Top 10 User IDs stand for?
> 
> Does the count of records mean, number of sessions or invocations? It will add more clarity if a more appropriate panel title and column title are chosen. Same for top 10 project IDs, top 10 API IDs. If there is no means to edit the title / column label, please ignore this feedback.
> 
> image
> Same is the case with the below-mentioned visualisation. Does the count signify the invocations or token usage?
> 
> image
> Can you see why the project id is empty in this screenshot?
> 
> image
> I understand that size is the term when doing image generation. Should we keep the technical term here or rather a more apt `resolution` here?
> 
> image
> Some of the graphs are having the term "frequency" and "distribution" in the y-axis. If it is a frequency, the unit of y-axis must be request / sec,requests /hr. In here, this is just the count. So, removing the word "frequency" will be more apt. Similarly the word "distribution" shows how a range of values are distributed ("Token usage distribution - by model"). I doubt this is the case here. So, recommend to remove the word "distribution" from the y-axis.
> 
> image
> The x-axis label is missing from the below mentioned panel
> 
> image

@agithomas 

1. For each model, over a day, data is aggregated. So 1 record equates to 1 aggregated data per model. So, it's something hard to write there in column. Any suggestion which is short? I can make them show sum of requests but the thing is aggregates request field only comes all datastreams, except vector store and code interpreter session. What do you think? That's the reason I kept it count of records. But yes, number of requests makes more sense but 2 data_streams (although they are not very important datasets for most) don't have those fields.
2. Yeah, I am aware. Unfortunately, the data is like that only:

```
    {
      "object": "bucket",
      "start_time": 1725926400,
      "end_time": 1726012800,
      "results": [
        {
          "object": "organization.usage.completions.result",
          "input_tokens": 28,
          "output_tokens": 119,
          "num_model_requests": 1,
          "project_id": "",
          "user_id": "user-iAityJpTPqeX89hlfRWYAZ8W",
          "api_key_id": "key_10xSzP3zPsz8zB5O",
          "model": "gpt-3.5-turbo-0125",
          "batch": null,
          "input_cached_tokens": 0,
          "input_audio_tokens": 0,
          "output_audio_tokens": 0
        }
      ]
    }
```

I think a bug on their side. For example, their code interpreter session API does not work as of now. Also one example is wrong in their website itself. Also this occasional occurrence of empty project could be another bug.

3. Yes, kept it size as OpenAI was using size. Yeah, I can change. I am making it "dimension" because technically dimension == size, resolution is usually PPI/ DPI precisely. ([ref](https://dsp.stackexchange.com/a/89410))
4. Good point. I'll remove them. Use more apt titles.

shmsr:
@agithomas Updated the screenshot.

shmsr:
> @shmsr ,
> 
> When i refer to the manifest.yml, i note the title as
> 
> * Collect audio transcriptions usage details for the organization.
> * Collect audio speeches usage details for the organization
> * Collect code interpreter sessions usage details for the organization
> 
> and so on.
> 
> I understand that to create the admin key, the user has to visit the `https://platform.openai.com/settings/organization/admin-keys`
> 
> A couple of questions here?
> 
> 1. Can the integration work with [Project API Key](https://platform.openai.com/settings/organization/api-keys) ?
> 2. Even though we are saying "for the organisation", and a given account can have [multiple organisations](https://platform.openai.com/docs/guides/production-best-practices) , the name of the organisation (or id) does not appear as a field. I think, the organization specific metrics would be a key customer requirement (a [sample requirement](https://community.openai.com/t/multiple-organizations-for-billing-purposes/81872)) . Does the API return the name / id of the organisation? If yes, kindly include it, add it as a part of input control drop down

1. No. So there are 3 API of OpenAI. 2 undocumented. 1 recently launched (documented). Undocumented ones work with Project keys, but the new one uses the Admin Key. Also, see [this](https://platform.openai.com/docs/api-reference/admin-api-keys).
2. About this -- So, a user would have manually change the org in the settings and then get the [Admin Key](https://platform.openai.com/docs/api-reference/administration) for that org specifically. No changing of org by just adding a request header like we do for normal OpenAI API. So, ideal solution is setup a new integration per org. Also, 1 org can have multiple projects. But the thing, we do not get the name of project (or any other field) but just the id. So that's why did not add a filter, as it will be very hard without a name as id are just very hard to read hashes. Also see 1st comment [here](https://github.com/elastic/integrations/pull/12494#issuecomment-2626418393).

shmsr:
@agithomas Address all your review comments. PTAL. Thanks for all the feedback!

shmsr:
@alaudazzi / @daniela-elastic Just waiting for your go-ahead on this PR. Dashboard changes seem to be done. Can I get an approval for the documentation?

shmsr:
Thank you for all the feedback so far. Today I am targeting the OpenAI release. Please let me know if there's any feedback I haven't addressed yet.

packages/openai/manifest.yml
```diff
@@ -0,0 +1,45 @@
+format_version: 3.3.0
+name: openai
+title: OpenAI
+version: 0.1.0
+description: |
+  Collect OpenAI usage events directly from OpenAI.
```
agithomas, line 6:
```suggestion
  Collect OpenAI usage metrics with Elastic Agent.
```

packages/openai/_dev/build/docs/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+With the OpenAI integration, you can track API usage metrics across their models, as well as for vector store and code interpreter. You will use Kibana to visualize your data, create alerts if usage limits are approaching, and view metrics when you troubleshoot issues. For example, you can track token usage and API calls per model.
+
+## Data collection
+
+The OpenAI integration leverages the [OpenAI Usage API](https://platform.openai.com/docs/api-reference/usage) to collect detailed usage metrics. The Usage API delivers comprehensive insights into your API activity, helping you understand and optimize your organization's OpenAI API usage.
+
+## Data streams
+
+The OpenAI integration collects the following logs data streams:
```
agithomas, line 13: It is mentioned - "logs data stream". Shouldn't this be metrics?

agithomas (approved):
Approved with minor suggestions for improvements.

AndersonQ:
@shmsr are you sure this PR is correct? It's showing 5000+ files changed. Perhaps it's missing a rebase?

ishleenk17:
> @shmsr are you sure this PR is correct? It's showing 5000+ files changed. Perhaps it's missing a rebase?

This is due to a recent activity done to address an incident
https://groups.google.com/a/elastic.co/g/security-deployment-and-devices/c/yPJgVO-ovUI/m/W2wrD7dDBQAJ

packages/openai/_dev/build/docs/README.md
alaudazzi: LGTM

packages/openai/_dev/build/docs/README.md
alaudazzi: @shmsr these two sections -- Requirements and Setup -- should remain separate

alaudazzi (approved):
LGTM

PR #12070, opened 2024-12-11
Add support of memory and LocksAndThreads .NET CLR metrics

stefans-elastic:
## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [ ] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## Related issues

- 

## Screenshots
Screenshot 2024-12-20 at 4 28 06 PM

PR #12503, opened 2025-01-28
Jlind23/deprecate loginput

jlind23:
With the introduction of https://github.com/elastic/beats/pull/42295 the log input will soon not be able to run anymore as it is deprecated. To continue running we need to add this allow deprecated parameter to all log input template.


jlind23:
@rdner @cmacknz I'd love to get a sanity check on this 😅

jlind23:
@elastic/sec-linux-platform could you help me understand why the auditd package tests are failing?

packages/auditd/manifest.yml
```diff
@@ -14,7 +14,7 @@ categories:
   - auditd
 conditions:
   kibana:
-    version: ^8.11.0
+    version: "^8.13.0"
```
rdner, line 17: Why did this change?

packages/cassandra/manifest.yml
```diff
@@ -1,15 +1,15 @@
...
+version: "1.17.0"
 description: This Elastic integration collects logs and metrics from cassandra.
 type: integration
 categories:
   - datastore
   - observability
 conditions:
   kibana:
-    version: "^8.13.0"
+    version: "^8.13.0 || ^9.0.0"
```
rdner, line 12: Why do we need it here but not in the rest of the packages?

packages/iis/manifest.yml
```diff
@@ -1,6 +1,6 @@
 name: iis
 title: IIS
-version: "1.21.1-next"
+version: "1.22.0"
```
rdner, line 3: Do we know why it had the `-next` prefix before? Is it okay to drop it?

jlind23: That means this was a pre-release not shipped to end users. I would need @lalit-satapathy's inputs here.

packages/cassandra/manifest.yml
jlind23: That was a mistake, thanks for spotting this.

packages/auditd/manifest.yml
jlind23: Rolling it back, missed it.

packages/iis/manifest.yml
rdner: If it was a pre-release I assume we would have to keep it.

jsoriano: `-next`, apart from being a prerelease suffix, it avoids publishing the package. So if you add a version without this suffix it will also publish previously unreleased changes.

Though I am not sure why the previous change was included under a `-next` suffix, it only adds definitions for new fields, it looks safe and releaseable.

@stefans-elastic @muthu-mps do you see any reason not to release https://github.com/elastic/integrations/pull/12070?

packages/iis/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.22.0"
+  changes:
+    - description: Allow the usage of deprecated log input and support for stack 9.0
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/12503
 - version: "1.21.1-next"
   changes:
```
jsoriano, line 8: The change under the version with the `-next` suffix was never published, so if we go on with this version I would suggest to include both changes in the new version, removing the one that was not released.
```suggestion
```

packages/activemq/data_stream/audit/agent/stream/log.yml.hbs
```diff
@@ -13,6 +13,7 @@ tags:
 {{#contains tags "forwarded"}}
 publisher_pipeline.disable_host: true
 {{/contains}}
+allow_deprecated_use: true
```
jsoriano, line 16: Is it ok to use this setting in old versions of the stack? (I guess so because tests are passing).

jlind23: It is, Elastic Agent will just ignore any unknown settings.

cmacknz (approved):
LGTM. This is a no-op until we make the change in Beats to refuse to start the log input without the `allow_depreacted_use: true` flag when it is run from Elastic Agent. Right now there is an exception for this. We would want this change made first regardless so that integrations aren't temporarily broken we do that.

jlind23:
@elastic/obs-ds-hosted-services @elastic/obs-infraobs-integrations  @elastic/sec-deployment-and-devices @elastic/sec-linux-platform @elastic/sec-windows-platform @elastic/security-service-integrations @elastic/stack-monitoring 
Can I get your code owner approval on this please?

consulthys (approved):
LGT Stack Monitoring

taylor-swanson (approved):

packages/forcepoint_web/data_stream/logs/agent/stream/log.yml.hbs
efd6, line 1: Changelog and manifest changes for this?

packages/infoblox_nios/data_stream/log/agent/stream/log.yml.hbs
efd6, line 1: Changelog and manifest changes for this?

jlind23:
Thanks @efd6 for the review, I have updated the two packages you pointed out.

efd6 (approved):
For security-service-integrations.

packages/iis/changelog.yml
```diff
@@ -1,6 +1,9 @@
 # newer versions go on top
-- version: "1.21.1-next"
+- version: "1.22.0"
```
ishleenk17, line 2: @muthu-mps : Is it ok if we are upgrading the version here like this for IIS.
I am not sure why we had next. If yes, would we want to have 1.21.1 first and then this changelog ?

jsoriano: Another thread about this in https://github.com/elastic/integrations/pull/12503#discussion_r1933471357.

muthu-mps: - We are waiting for `8.17.2 release` to make this change available for the user. As this change was planned to merge before beats release we added `-next` tag to make it not available for the user.

packages/iis/manifest.yml
muthu-mps: @jsoriano,
The `-next` tag was added to have the newer fields available from 8.17.x version and user has to upgrade the stack to get monitor those fields. I have noticed there was a similar [discussion](https://github.com/elastic/integrations/pull/12407#discussion_r1925741195) around handing such a scenario without upgrading the Kibana version. We can drop the pre-release tag.

muthu-mps (commented):
Add changelog entry for this change,
`packages/mongodb/data_stream/log/agent/stream/log.yml.hbs`

muthu-mps (approved):
Added comment on changelog entry otherwise,
LGTM!

jlind23:
@muthu-mps just updated mongodb changelog and manifest.

jlind23:
@norrietaylor @andrewkroh @bturquet can we please get some eyes on this?

jlind23:
@robbavey @jsvd If one of you can review this it would be great!

jlind23:
@mrodm @jsoriano can one of you help me with the CI failure? The reason makes no sense to me 🤐

jsoriano:
@jlind23 this issue is fixed by https://github.com/elastic/integrations/pull/12630, try to cherry pick this change.

jlind23:
@nfritts @robbavey can I please get some eyes on this? Both you teams are missing reviews here.

robbavey (approved):
LGTM

nfritts (approved):

cmacknz (approved):

PR #12508, opened 2025-01-28
crowdstrike: fix handling of network direction

### LINKED ISSUE #12476: [Crowdstrike FDR]: Pipeline sets fields wrongly based on bad assumption
### Integration Name

CrowdStrike [crowdstrike]

### Dataset Name

crowdstrike.fdr

### Integration Version

1.49.0

### Agent Version

8.17.3

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.17.3

Not relevant

### Software/API Version

Not relevant

### Error Message

The pipeline renames "remote" fields into destination fields. Which is wrong because it is done based on the assumption that all connections are outbound. For inbound connections this renaming is wrong. There needs to be if statements that bases the renaming on the network direction.

It looks to be sourced from this commit https://github.com/elastic/integrations/commit/512db6f56f69d58b904c06d2f398133c1765c274#diff-9b8e65a89b684b4faab3770dbf228c2d81ba7dcb4836169d221c73ef3e70d8d2 which is 4(!!!) years old. Anyone who has been using this pipeline of-the-shelf has had their data wrongly renamed for the past four years.

### Event Original

_No response_

### What did you do?

Ingest data via the official integration

### What did you see?

Fields are being renamed wrongly. 

I would expect the destination address to correlate to the direction of the network session.

### Anything else?

_No response_

efd6:
## Proposed commit message

Assume that network direction that is no inbound can be validly semantically represented as outbound using the ECS fields available. This is probably not true; documented values of the ConnectionDirection are 0 - outbound, 1 - inbound, 2 - neither, and 3 - both[1].

Adhering strictly to inbound/outbound makes it impossible to map the data to ECS since neither and both would only be expressible as unknown.

[1]https://docs.panther.com/data-onboarding/supported-logs/crowdstrike/falcon-data-replicator#crowdstrike.networkconnect

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #12476


packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/outbound_network.yml
```diff
@@ -0,0 +1,63 @@
...
+      type: ip
+      ignore_missing: true
+  - rename:
+      field: crowdstrike.RemoteAddressIP6
+      target_field: destination.ip
+      ignore_missing: true
+  - set:
+      field: destination.address
+      copy_from: destination.ip
+      ignore_empty_value: true
```
kcreddy, line 59: This is identical to L#43 to L#47. One of them could be deleted.

packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/outbound_network.yml
```diff
@@ -0,0 +1,63 @@
...
+  - convert:
+      tag: convert_RemoteAddressIP6_ip
+      field: crowdstrike.RemoteAddressIP6
+      type: ip
+      ignore_missing: true
+  - rename:
+      field: crowdstrike.RemoteAddressIP6
+      target_field: destination.ip
+      ignore_missing: true
```
kcreddy, line 55: It seems we are setting `crowdstrike.RemoteAddressIP6` as `destination.ip` irrespective of `convert` succeeds or not. Is this okay?

Also, `RemoteAddressIP6` and `LocalAddressIP6` are both defined as `string` in documentation, but one of them is being handled as string, while other as List. 
Not sure if we need to handle cases of both fields being string or List types.

packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1970,56 +1963,28 @@ processors:
...
-      ignore_empty_value: true
-  - rename:
-      field: crowdstrike.RemotePort
-      target_field: destination.port
-      ignore_missing: true
+
+  - pipeline:
+      tag: pipeline_outbound_network
+      if: ctx.network?.direction != 'inbound'
+      name: '{{ IngestPipeline "outbound_network" }}'
```
kcreddy, line 1980:
```suggestion
      name: '{{ IngestPipeline "not_inbound_network" }}'
```
WDYT? because it corresponds to both outbound and unknown?

kcreddy (commented):
It would be nice to hear from user on https://github.com/elastic/integrations/issues/12476#issuecomment-2620207963

packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/outbound_network.yml
efd6: Done. Also in inbound.

packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/outbound_network.yml
efd6: The logic is the same here as previously present, but I'll add changes for this.

packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
efd6: I would prefer not; I'm embedding some cognitive dissonance here to ensure that we don't get comfortable with what we are doing. I'll add a comment.

kcreddy (approved):

efd6:
Waiting for @monkeyonbranch. Setting as draft to avoid accidental merge.

monkeyonbranch:
Sorry for the long wait. Sometimes life happens.
Yes, the fixes looks good to me. It should provide and improvement in making sure the right data is being set when normalizing it.

efd6:
Thanks @monkeyonbranch. If there are any issues with the heuristic we are using, please file a new issue.

PR #12468, opened 2025-01-25
Enablement & Sanity test run for OBS Non-Cloud integration packages for 9.0.0

kush-elastic:
- Enhancement

# Description
OBS non-cloud packages for 9.0.0 enablement activities of O11y integrations.

# Steps performed for the testing
1. update manifest.yml, package minor version upgrade and kibana dependency addition ` || ^9.0.0`.
2. build package
3. run package tests
4. install integration from kibana
5. check for any issues

## List of non-cloud packages and results of sanity

Below mentioned results are from running tech stack `9.0.0`.
Pass: Successfully Passed
Pending: Either permission issue or agent will stuck on 0 hits
Failed: Failed because of beats and integration changes are not in sync
Not Available: Tests are not available

No. | Packages | build & check | pipeline | system | asset | static
-- | -- | -- | -- | -- | -- | --
1 | activemq | Pass | Pass | Pass | Pass | Pass
2 | apache_spark | Pass | Not Available | Pass | Pass | Pass
3 | apache_tomcat | Pass | Pass | Pending | Pass | Pass
4 | cassandra | Pass | Pass | Pass | Pass | Pass
5 | ceph | Pass | Pass | Pass | Pass | Pass
6 | citrix_adc | Pass | Pass | Pass | Pass | Pass
7 | cockroachdb | Pass | Not Available | Pass | Pass | Pass
8 | coredns | Pass | Pass | Pending | Pass | Pass
9 | couchbase | Pass | Pass | Pass | Pass | Pass
10 | couchdb | Pass | Pass | Pass | Pass | Pass
11 | etcd | Pass | Not Available | Pass | Pass | Pass
12 | golang | Pass | Pass | Pass | Pass | Pass
13 | hadoop | Pass | Pass | Pass | Pass | Pass
14 | haproxy | Pass | Pass | Pass | Pass | Pass
15 | ibmmq | Pass | Pass | Pass | Pass | Pass
16 | influxdb | Pass | Not Available | Pass | Pass | Pass
17 | jolokia_input | Pass | Not Available | Pass | Not Available | Pass
18 | kafka | Pass | Pass | Not Available | Pass | Pass
19 | kafka_log | Pass | Not Available | Pass | Pass | Pass
20 | memcached | Pass | Not Available | Pass | Pass | Pass
21 | microsoft_sqlserver | Pass | Pass | Pending | Pass | Pass
22 | mongodb | Pass | Pass | Fail | Pass | Pass
[...]

PR #12516, opened 2025-01-29
[CI] Allow to test Pull Requests with specific stack version

mrodm:
## Proposed commit message

Enable to add a comment in the Pull Requests, to test the packages updated in the given PR with a different Elastic stack version.

This process would create a new build in a different Buildkite pipeline (to be created): https://buildkite.com/elastic/integrations-test-stack.

Posting GitHub comments into a PR would just be available (and requires to be tested) once this PR is merged.

## Author's Checklist

- [x] Test overriding stack version 
    - https://buildkite.com/elastic/integrations/builds/21224

## Related issues

- Relates https://github.com/elastic/integrations/pull/12468


mrodm:
@jsoriano WDYT about the approach followed here? 

These builds will always fail with an specific test even if the corresponding package steps finish successfully.

That would ensure that the author of the PR would run a CI build without any override as a final execution.

I was thinking to add a comment. However, that would not block the PR, and the PR could be merged without being tested with the minimum Kibana version set in the package.

Example of buildkite build: https://buildkite.com/elastic/integrations/builds/21215

jsoriano:
> These builds will always fail with an specific test even if the corresponding package steps finish successfully.
> 
> That would ensure that the author of the PR would run a CI build without any override as a final execution.

Could we run an additional pipeline when the comment is found? So the required pipeline is not altered and is executed and expected in any case.

I think it would be similar to what we do with `test serverless` in elastic-package.

mrodm:
> Could we run an additional pipeline when the comment is found? So the required pipeline is not altered and is executed and expected in any case.

Yes, this could be done. That new pipeline could also use the same the pipeline definition as the main one (`.buildkite/pipeline.yml`), including the logic here.

mrodm:
This PR will create a new Buildkite pipeline https://github.com/elastic/integrations-test-stack 

A new build of this pipeline will be triggered when a GitHub comment like `/test stack 9.0.0-SNAPSHOT` is posted in the PR.

jsoriano (approved):
Looks great, thanks!

PR #12517, opened 2025-01-29
[PANW] Remove leftover message field after parse

### LINKED ISSUE #12511: [PANW]: Ingest pipeline does not remove the partial message after parsing it
### Integration Name

Palo Alto Next-Gen Firewall [panw]

### Dataset Name

panw.panos

### Integration Version

4.2.0

### Agent Version

8.17.1

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.17.1

Elastic Cloud

### Software/API Version

_No response_

### Error Message

The PANW Ingest pipeline has a default pipeline that will parse the csv message from palo alto and creates a partial of the message in the field `message`, this field has the message that will be parsed by one of the auxiliary pipelines according with the log type, the value of the field `panw.panos.type`.

For example, if the value is `TRAFFIC` then it will be parsed by the traffic ingest pipeline, but this `message` field is not removed after the parse in the auxiliary pipelines.

This partial message is created in this [grok](https://github.com/elastic/integrations/blob/edad203551498c222573f9c1ecc60d228baa41d8/packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/default.yml#L29-L37) filter in the default ingest pipeline.

This can have a huge impact in the space storage used by the PANW data. 

### Event Original

_No response_

### What did you do?

Used the native PANW integration to parse Palo Alto device logs.

### What did you see?

The log is correctly parsed, but the partial message used by each auxiliary pipeline is not removed and ends up in the final document, requiring more space than needed.

The partial message should be removed after it is parsed.

### Anything else?

The `on_failure` configuration in the default ingest pipeline has a `remove` filter that remove the temporary fields `_temp_`, `_conf` and also the `message` field. [[processor here](https://github.com/elastic/integrations/blob/edad203551498c222573f9c1ecc60d228baa41d8/packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/default.yml#L1721-L1733)]

The `remove` processor that is used to remove temporary fields only removes `_temp_` and `_conf`, it needs to be changed to also remove the `message` field. [[processor here](https://github.com/elastic/integrations/blob/edad203551498c222573f9c1ecc60d228baa41d8/packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/default.yml#L1609-L1614)]

leandrojmp:
- Breaking change

## Proposed commit message

During the parse of the PANW log messages, the original `message` field is parsed by a `grok` filter and generates a new partial `message` field.

This partial `message` field will then be parsed by the auxiliary ingest pipelines, but after the parse is completed, the field is not removed and is part of the final document.

Since ths `message` fields is the majority of the original source message, this can represent an increase of almost 50% in storage requirements for the PANW data stream.

This PR aims to remove this field.

## Related issues

- Closes #12511

### DIFF
--- a/packages/panw/changelog.yml
+++ b/packages/panw/changelog.yml
@@ -1,5 +1,5 @@
 # newer versions go on top
-- version: "4.3.0"
+- version: "5.0.0"
   changes:
     - description: Remove leftover message field after parse
       type: breaking-change
--- a/packages/panw/changelog.yml
+++ b/packages/panw/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Remove leftover message field after parse
       type: breaking-change
-      link: https://github.com/elastic/integrations/pull/XXXXX #TODO
+      link: https://github.com/elastic/integrations/pull/12517
 - version: "4.2.0"
   changes:
     - description: Parse URL in threat logs with 'domain-edl' threat category
--- a/packages/panw/changelog.yml
+++ b/packages/panw/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "4.3.0"
+  changes:
+    - description: Remove leftover message field after parse
+      type: breaking-change
+      link: https://github.com/elastic/integrations/pull/XXXXX #TODO
 - version: "4.2.0"
   changes:
     - description: Parse URL in threat logs with 'domain-edl' threat category
--- a/packages/panw/manifest.yml
+++ b/packages/panw/manifest.yml
@@ -1,6 +1,6 @@
 name: panw
 title: Palo Alto Next-Gen Firewall
-version: "4.2.0"
+version: "4.3.0"
 description: Collect logs from Palo Alto next-gen firewalls with Elastic Agent.
 type: integration
 format_version: "3.0.3"
--- a/packages/panw/manifest.yml
+++ b/packages/panw/manifest.yml
@@ -1,6 +1,6 @@
 name: panw
 title: Palo Alto Next-Gen Firewall
-version: "4.3.0"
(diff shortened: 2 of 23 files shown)


packages/panw/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "4.3.0"
```
taylor-swanson, line 2: Since this is a breaking change, let's make this a major version bump.

```suggestion
- version: "5.0.0"
```

packages/panw/manifest.yml
```diff
@@ -1,6 +1,6 @@
 name: panw
 title: Palo Alto Next-Gen Firewall
-version: "4.2.0"
+version: "4.3.0"
```
taylor-swanson, line 3: Ditto

```suggestion
version: "5.0.0"
```

taylor-swanson (commented):
Overall this change looks good to me.

I suppose if someone wanted the original behavior, they could enable "preserve original event", which would give them `event.original`, which contains what `message` had before, plus a header.

It doesn't seem like it was intended to save `message` like this, rather it was an oversight.

packages/panw/changelog.yml
leandrojmp: done!

packages/panw/manifest.yml
leandrojmp: done!

leandrojmp:
The test failed, it seems to be complaining about the message received being different from the expected.

I remove the `message` field from all expected files, what else do I need to change to fix ?

taylor-swanson (approved):
LGTM

PR #12518, opened 2025-01-29
[mimecast] Fix cursor time formatting and selection

chrisberkhout:
## Proposed commit message

```
[mimecast] Fix cursor time formatting and selection

In the `audit_events` data stream, the `state.cursor.last` value was
handled as a time type, but when cursor data is reloaded from the file
the value will be a string. The change matches the behavior in other
data streams: explicitly format the time as a string when storing it,
and handle it as a string when reading it.

The logic for selecting a time for the cursor was to take the max of
returned times and `now`. This will usually use the time `now` as the
new starting point, instead of the time of last data. The fix is to only
use `now` if there is no returned data, and otherwise use the latest
returned time.
```

## Checklist

### DIFF
--- a/packages/mimecast/data_stream/archive_search_logs/agent/stream/cel.yml.hbs
+++ b/packages/mimecast/data_stream/archive_search_logs/agent/stream/cel.yml.hbs
@@ -93,11 +93,18 @@ program: |
             {
               "events": body.data.map(e, e[state.data_path].map(l, {"message": l.encode_json()})).flatten(),
               "cursor": {
-                "last": ([now] + body.data.map(e,
-                  e[state.data_path].map(l,
-                    l[state.time_field].parse_time(["2006-01-02T15:04:05-0700", time_layout.RFC3339])
-                  )
-                ).flatten()).max().format(time_layout.RFC3339),
+                "last": (
+                  body.data.map(e,
+                    e[state.data_path].map(l,
+                      l[state.time_field].parse_time(["2006-01-02T15:04:05-0700", time_layout.RFC3339])
+                    )
+                  ).flatten().as(time_fields,
+                    (size(time_fields) > 0) ?
+                      time_fields.max()
+                    :
+                      now
+                  ).format(time_layout.RFC3339)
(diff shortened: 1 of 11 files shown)


packages/mimecast/data_stream/archive_search_logs/agent/stream/cel.yml.hbs
```diff
@@ -93,11 +93,18 @@ program: |
...
+                    e[state.data_path].map(l,
+                      l[state.time_field].parse_time(["2006-01-02T15:04:05-0700", time_layout.RFC3339])
+                    )
+                  ).flatten().as(time_fields,
+                    (size(time_fields) > 0) ?
+                      time_fields.max()
+                    :
+                      now
+                  ).format(time_layout.RFC3339)
+                ),
```
efd6, line 107: This set of changes is described in the change description, but I'm wondering how this impacts on the behaviour of the integration as a whole; there should only be a difference in the case that `now` is past events that have been seen by the API backend but have not been made available to the API client. Is this the behaviour that you have seen and are fixing, or is this hypothetical?

chrisberkhout: The 1st commit is to fix a reported problem. Maybe it's better to separate the 2nd commit?

The 2nd commit could stop some records being missed. If you make a request at midday, for a time range ending at midday, you'll get events from the morning. The existing code was asking for the maximum of the morning times and the midday time, which is always going to be midday. I assumed the intent was to get the latest of the morning times, so we can continue from where we left off, and have midday as a fallback. Continuing from the last request time rather than the last data time can leave small gaps.

efd6: Thanks. I think it's fine to have in a single PR.

This, "Continuing from the last request time rather than the last data time can leave small gaps" is what I was wondering about. What you have is good, but I was just curious.

efd6 (approved):
Thanks

PR #12534, opened 2025-01-30
[Crowdstrike Alerts] change hash compution

SimonKoetting:
Currently, the integration uses multiple fields, including the updated_timestamp to compute a unique _id for the alerts.
Due to the nature of the updated_timestamp field, which can be changed due to internal background processes of CrowdStrike without having any change to the alert, this can lead to duplicates.
This PR changes this to use the created and context timestamp, which are static per alert, instead of the static one to avoid these duplicates


chrisberkhout (commented):
The original list of fingerprint fields was set in https://github.com/elastic/integrations/pull/10567.

> crowdstrike has several background processes that can change this value without any change to the alert itself

This change seems good if changes to the `updated_timestamp` value _never_ come with changes to the alert itself.

To catch all alert changes we could fingerprint everything except the `updated_timestamp`.

PR #12538, opened 2025-01-30
Update Kubernetes container logs documentation

belimawr:
## Proposed commit message

This commit updates the Kubernetes Container Logs documentation to better explain that an input is always generated for every container.

It also fixes a broken link.

## Checklist

- [ ] ~~I have verified that all data streams collect metrics or logs.~~

~~## Author's Checklist~~

```
cd packages/kubernetes
elastic-package build
elastic-package stack up -v -d
```
Navigate to https://localhost:5601/app/integrations/detail/kubernetes-1.70.2/overview
Check the new entry under "container-logs"

Add the integration, go to Collect Kubernetes container logs -> Advanced options -> Kubernetes container log path and check the new documentation for this field.

~~## Related issues~~
~~## Screenshots~~

### DIFF
--- a/packages/kubernetes/data_stream/container_logs/manifest.yml
+++ b/packages/kubernetes/data_stream/container_logs/manifest.yml
@@ -13,6 +13,12 @@ streams:
         multi: true
         default:
           - /var/log/containers/*${kubernetes.container.id}.log
+        description: >-
+          For every container in the cluster an instance of the input
+          will be created harvesting all paths defined here, even if
+          the paths contain no variable! Refer to the [integration
+          documentation](https://www.elastic.co/guide/en/integrations/current/kubernetes.html)
+          for more details.
       - name: symlinks
         type: bool
         title: Use Symlinks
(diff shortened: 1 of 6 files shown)


packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -13,6 +13,12 @@ streams:
         multi: true
         default:
           - /var/log/containers/*${kubernetes.container.id}.log
+        description: >-
+          For every container in the cluster an instance of the input
+          will be created harvesting all paths defined here, even if
+          the paths contain no variable! Refer to the [integration
```
gizas, line 19: Here you mean that the default value will be used ?
Maybe below is more clear?
```suggestion
          For every container in the cluster an instance of the input
          will be created harvesting all paths defined here, based on the default value! 
```

gizas (approved):

MichaelKatsoulis (approved):

packages/kubernetes/data_stream/container_logs/manifest.yml
belimawr: It's actually every container on the node/scope the Elastic-Agent can see, I'll re-word it.

Another thing I'm trying to get across is the fact that one input per container will be created regardless of what is set as `paths`.

What do you think of this version
```
          For every container the Elastic-Agent can see (usually every
          container on the node) an instance of the input will be
          created harvesting all paths defined here, even if 
          the paths contain no variable! Refer to the [integration
          documentation](https://www.elastic.co/guide/en/integrations/current/kubernetes.html)
          for more details.
```

gizas: Sorry maybe I have wrong impression of what is happening, os to clarify: The inputs will be created for every container despite the paths defined? Or first we search based on the paths and we create the inputs?

MichaelKatsoulis: The inputs are created based on the id of the stream that contains variables. So one input per pod and container 
`kubernetes-container-logs-${kubernetes.pod.name}-${kubernetes.container.id}`
Then the paths are checked. If the paths are containing variables then the the input will only harvest logs for that container log file. If for any reason, the user has not set variables in the paths (e.g. `/var/log/containers/*.log`), then this will lead to multiple instances(as many as the containers in the node) harvesting all the logs of the containers and setting the wrong metadata.

gizas: Yeap now makes sense. Thanks Micahlis. The combination of id and the path is what users need to understand here

PR #12543, opened 2025-01-30
apm: change tail_sampling_storage_limit config default to 0GB

carsonip:
## Proposed commit message

`sampling.tail.storage_limit` is 0 by default in 9.0. See https://github.com/elastic/apm-server/pull/15467 .
As UI validation requires unit (e.g. GB), set apm integration default storage limit to `0GB` which carries the same meaning.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/apm/changelog.yml
+++ b/packages/apm/changelog.yml
@@ -1,6 +1,6 @@
-- version: 9.0.0
+- version: 9.0.0-preview-1738343125
   changes:
-    - description: Remove data streams from integration package
+    - description: Change default sampling.tail.storage_limit to 0
       type: breaking-change
       link: https://github.com/elastic/integrations/pull/12543
 - version: 8.15.0-preview-1716438434
--- a/packages/apm/changelog.yml
+++ b/packages/apm/changelog.yml
@@ -1,3 +1,8 @@
+- version: 9.0.0
+  changes:
+    - description: Remove data streams from integration package
+      type: breaking-change
+      link: https://github.com/elastic/integrations/pull/12543
 - version: 8.15.0-preview-1716438434
   changes:
     - description: Remove data streams from integration package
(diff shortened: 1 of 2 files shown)


packages/apm/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: 9.0.0
```
carsonip, line 1: [to reviewer] should it be a preview version?

packages/apm/manifest.yml
```diff
@@ -10,7 +10,7 @@ conditions:
     capabilities:
       - apm
   kibana:
-    version: ^8.15.0
+    version: ^9.0.0
```
carsonip, line 13: [to reviewer] is it correct?

packages/apm/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 3.1.0
 name: apm
 title: Elastic APM
-version: 8.15.0-preview-1716438434
+version: 9.0.0
```
carsonip, line 4: [to reviewer] should it be a preview version?

carsonip (commented):
Q: is it even possible to publish a 8.18 integration package version in the future?

carsonip:
Seems that the APM integration UI doesn't like `0`, and I'm afraid that setting the default to 0 will result in a validation error by default. On the other hand, setting it to empty or `0GB` (or `0B`) is fine.

1. I'm inclined to set integration package default to `0GB` in to make it explicit.
2. Updating the description to mention the new 0 behavior will be helpful, and we might as well change the validation to allow for `0`. But we still need this PR to set integration package default to 0.

Let me see how much effort is needed to do (2), If it is too much, I'll go with (1).

image

simitt (approved):

Issue #12249, opened 2025-01-07
Due diligence ticket for clean up Aruba feature-branch integration before merging into `main`


qcorporation:
## Parent Ticket
https://github.com/elastic/integrations/issues/5255

## Description

- consolidate common fields into a general `aruba.*`
- event categorization painless script that labels required fields
- Apply `rules_export.ndjson` to see if there are rule violations within the integration (github does now allow ndjson attachments - use offline material)
- consolidate throttle event messages from the different event ids
- upgrade all event types of OS 10.15, some old implementations are still at 10.7

PR #12553, opened 2025-01-31
#12249-9b Updated events.* required fields

qcorporation:
## Parent Ticket
https://github.com/elastic/integrations/issues/12249

## Description
- generated the expected value from the logs
- generated but did not properly populate the values for most event types besides LLDP (this will come later)
- tested that if( xxx != null ) logic works for each event.* field being populated


dwhyrock (approved):
i'm going to trust that you got the event code parameters correct since i'm not going to go through a over a thousand logs and check the mappings.

gogochan (approved):

PR #12555, opened 2025-01-31
packages/cel - add 'condition' variable

### LINKED ISSUE #12521: [Custom API using CEL]: Add kubernetes autodiscover condition field to integration configuration page
### Integration Name

Custom API using Common Expression Language [cel]

### Dataset Name

cel.cel

### Integration Version

1.13.0

### Agent Version

8.16.0

Elastic Agent 8.16.0 image deployed with Kuberneetes

### User Goal

My customer has defined a single policy for all his kubernetes nodes and is using the the autodiscovery conditions to assign inputs based on the kubernetes labels.  This works well for Prometheus integration. Here is an example where we assign the Prometheus integration to collect from pods that match kubernetes.labels.component==bookie (this is for apache pulsar metrics).  We have another custom integration (NSQ) that uses CEL that we want to assign, but the condition field is not in the Kibana CEL integration config page.  I guess the alternative is to create different policies and assign the CEL policy only to those nodes. But the customer is administrating hundreds of nodes so this would add to his workload.

Image

### Existing Features

No autodiscovery condition option in the configuration

### What did you see?

No ingested docs since we can't configure.

### Anything else?

_No response_

andrewkroh:
## Proposed commit message

```
Add a user configurable 'condition' variable to allow users to control when 
the input should run. An example use case would be to apply Kubernetes 
autodiscovery conditions.

See https://www.elastic.co/guide/en/fleet/current/dynamic-input-configuration.html

Closes #12521
```

### DIFF
--- a/packages/cel/agent/input/input.yml.hbs
+++ b/packages/cel/agent/input/input.yml.hbs
@@ -1,3 +1,7 @@
+{{#if condition}}
+condition: {{condition}}
+{{/if}}
+
 data_stream:
   dataset: {{data_stream.dataset}}
 interval: {{resource_interval}}
--- a/packages/cel/changelog.yml
+++ b/packages/cel/changelog.yml
@@ -1,3 +1,8 @@
+- version: "1.14.0"
+  changes:
+    - description: Add a 'condition' configuration option for specifying to Elastic Agent when it should run the input.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/12555
 - version: "1.13.0"
   changes:
     - description: Add xsd configuration option inside input.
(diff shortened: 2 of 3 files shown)


packages/cel/agent/input/input.yml.hbs
```diff
@@ -1,3 +1,7 @@
+{{#if condition}}
+condition: {{condition}}
```
efd6, line 2: Looking for all the ways that this could break, what happens when someone writes `false` in the UI?

efd6 (approved):

packages/cel/agent/input/input.yml.hbs
efd6: Given that the there are other ways to achieve the same outcome (to enable toggle), I don't think it's a concern, and should just prompt a support answer of "Please don't do that."

efd6:
This needs to have the logstash ownership fix picked up.

andrewkroh:
Reincarnated at #12647.

PR #12572, opened 2025-02-03
Update Deployment and Devices integrations to ECS 8.17.0 (part 3)

taylor-swanson:
## Proposed commit message

Updates the following integrations to ECS 8.17.0:

- juniper_srx
- modsecurity
- netflow
- osquery
- panw
- pfsense
- proxysg

## Checklist

## Related issues

- Relates #11952


packages/pfsense/data_stream/log/_dev/test/pipeline/test-pfsense-haproxy.log-expected.json
```diff
@@ -406,14 +406,14 @@
             ]
         },
         {
-            "@timestamp": "2024-06-13T20:53:49.000-04:00",
+            "@timestamp": "2025-06-13T20:53:49.000-04:00",
```
dwhyrock, line 409: Curious why the date year is being updated here, but not in the file below (`packages/pfsense/data_stream/log/_dev/test/pipeline/test-pfsense-ipsec.log-expected.json`), even though they are both part of the `pfsense` package?

packages/proxysg/data_stream/log/fields/fields.yml
```diff
@@ -10,32 +10,32 @@
...
-        type: keyword
-      - name: transaction_uuid
-        type: keyword
-      - name: access_type
-        type: keyword
-      - name: access_security_policy_action
-        type: keyword
-      - name: access_security_policy_reason
-        type: keyword
+        - name: appliance_name
```
dwhyrock, line 13: Did these have negative effects being at the same level as `fields:` ?

packages/proxysg/data_stream/log/fields/fields.yml
```diff
@@ -204,7 +204,7 @@
           type: keyword
         - name: icap_status
           type: keyword
-        - name: certificate_hostname
```
dwhyrock, line 207: 🥇

dwhyrock (approved):
Couple minor questions, otherwise LGTM.

packages/pfsense/data_stream/log/_dev/test/pipeline/test-pfsense-haproxy.log-expected.json
taylor-swanson: `@timestamp` is a weird field when it comes to pipeline tests.

The main issue is that (most) of these logs don't contain a year, so when the expected file generates, it gets whatever year the current year is. This works fine until the new year comes, and then the test will fail, since the pipeline will generate a document with the new year, but this no longer matches what is in the expected file.

The "solution" is to mark the `@timestamp` as a dynamic field (see `test-common-config.yml`). That config takes a regex pattern. You could do `.*`, but to make it slightly more robust, some integrations do some form of regex against the `RFC3339` format, although it tends to miss things like the time zone offset at the end. What this does it it will validate the value of `@timestamp` against the pattern, but not the value in the expected file. Why it updates some of the entries and not all of them, I'm not sure. Technically, it's not supposed to update any of the values if the field is marked as dynamic.

packages/proxysg/data_stream/log/fields/fields.yml
taylor-swanson: Functionally, they are the same thing, just a formatting change.

PR #12615, opened 2025-02-05
[rubrik] Move templating in CEL state

ghost:
## Proposed commit message
This PR moves hbs templating in CEL state from CEL program.
Minor fixes.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/rubrik/data_stream/drives/agent/stream/cel.yml.hbs
```diff
@@ -75,20 +76,8 @@ program: |-
         "Body": {
           "query": state.queries[state.index],
           "variables": {
-              {{#if pageSize}}
-              "first": {{pageSize}},
-              {{/if}}
+              "first": has(state.pageSize) ? state.pageSize : null,
               "after": has(state.cursor) ? state.cursor.after : null,
-              "filter": [
```
muthu-mps, line 82: Why the `filter` is getting removed from the datastream?

gpop63: Because hbs templating doesnt work the same way in cel state and for now in the beta version we can proceed with just removing the filters.

efd6: You can just move it up into the state definition, say at `filter`, and then refer to it via `state.filter`.

Something like
```
state:
  index: 0
  pageSize: {{pageSize}}
{{#if filters}}
  filters:
  {{#each filters}}
    - field: "{{field}}",
      texts: [{{#each values}}"{{this}}"{{#unless @last}}, {{/unless}}{{/each}}]
  {{/each}}
{{/if}}
  queries:
```
and then here
```
            "first": has(state.pageSize) ? state.pageSize : null,
            "filter": state.?filter.orValue([]),
```

Issue #10277, opened 2024-06-26
[qualys_vmdr.asset_host_detection] Align `cloud.*` fields with Qualys data and not elastic agent metadata


clement-fouque:
When ingesting vulnerability data, `cloud.*` fields are a mix of the Elastic agent metadata and Qualys vulnerability data. I didn't find a way to disable metadata coming from Elastic agent (`publisher_pipeline.disable_host` seems to apply to field `host` only). 

A decision must be made to either have no cloud fields set, either have only the data coming from Qualys, either from the elastic agent metadata.

I have a strong preference to remove any properties set by the elastic agent and only have data from Qualys. If we want to build dashboard with data coming from different scanning solutions (Qualys, Tenable, Snyk, ...), we'll heavily rely on ECS fields. 

We might have a rule that if the field `tags` contains `forwarded`, we use only the data from Qualys. 

Here is an example of what I'm getting
```json
"cloud": {
  "availability_zone": "us-central1-a", # Elastic agent metadata
  "instance": {
    "name": "gke-xxx", # Elastic agent metadata
    "id": "12346789" # Qualys vulnerability data
  },
  "provider": "GCP", # Qualys vulnerability data
  "service": {
    "name": "Compute Engine" # Qualys vulnerability data
  },
  "project": {
    "id": "my-project" # Elastic agent metadata
  },
  "region": "us-central1", # Elastic agent metadata
  "account": {
    "id": "my-project" # Elastic agent metadata
  }
}
```

efd6:
@clement-fouque Would you be satisfied with setting the env var `BEATS_ADD_CLOUD_METADATA_PROVIDERS=""`? This has the effect of turning off the processor as can be seen here https://github.com/elastic/beats/blob/1ce84438ece3bfdd94fd8868365fb4fbf750256e/libbeat/processors/add_cloud_metadata/providers.go#L86-L91

andrewkroh:
If the qualy integration can set `cloud.*` fields it sounds like we would never want to cloud fields from the Agent's environment. I think we could unconditionally remove the `cloud` field before continuing to parse the qualys JSON.

clement-fouque:
I prefer the solution proposed by @efd6 as it gives the control to user on what he wants to do. I don't see a use case where users would want to keep elastic agent metadata but it's good to have the option though.

The ideal solution would be to have an option in the UI to select what we want to do: either set agent data, either set Qualys data.

clement-fouque:
[...]

PR #12624, opened 2025-02-05
Update security service integrations packages mappings

mrodm:
## Proposed commit message

Updated failed mappings found in system tests while validating the documents based on mappings in the data-streams ~and the corresponding transforms~.

Example of Buildkite build where the failures are raised: https://buildkite.com/elastic/integrations/builds/20794

Packages updated here:
- box_events
```
[0] field "threat.enrichments.indicator.first_seen" is undefined: actual mapping type (keyword) does not match with ECS definition type: date
[1] field "threat.enrichments.indicator.last_seen" is undefined: actual mapping type (keyword) does not match with ECS definition type: date
```
- claroty_ctd
```
[0] field "threat.indicator.modified_at" is undefined: actual mapping type (keyword) does not match with ECS definition type: date
```
- crowdstrike
```
[0] field "crowdstrike.assessments.analytics_and_improvements_mac" is undefined: field definition not found
[1] field "crowdstrike.assessments.application_firewall_mac" is undefined: field definition not found
[2] field "crowdstrike.assessments.crendential_dumping_hash_mac" is undefined: field definition not found
...
```
- mimecast
```
[0] field "threat.indicator.first_seen" is undefined: actual mapping type (keyword) does not match with ECS definition type: date
[1] field "threat.indicator.modified_at" is undefined: actual mapping type (keyword) does not match with ECS definition type: date
```
- sublime_security
```
[0] field "sublime_security.email_message.headers.hops.fields.position" is undefined: field definition not found
```
- ti_anomali
```
[0] field "threat.indicator.modified_at" is undefined: actual mapping type (keyword) does not match with ECS definition type: date
```
<details>
<summary>Packages with changes related to transforms moved to #12841</summary>

- github (transforms)
```
field "message" is undefined: actual mapping type (keyword) does not match with ECS definition type: match_only_text
```
- ti_custom (transforms)
```
[0] field "threat.indicator.url.original" is undefined: actual mapping type (keyword) does not match with ECS definition type: wildcard
```
- tychon (transforms)
```
[0] field "related.ip" is undefined: actual mapping type (keyword) does not match with ECS definition type: ip
[1] field "server.ip" is undefined: actual mapping type (keyword) does not match with ECS definition type: ip
```
- wiz (transforms)
```
[0] field "ecs.version" is undefined: actual mapping type (text) does not match with ECS definition type: keyword
[1] field "message" is undefined: actual mapping type (text) does not match with ECS definition type: match_only_text

[0] field "device.id" is undefined: actual mapping type (text) does not match with ECS definition type: keyword
[1] field "ecs.version" is undefined: actual mapping type (text) does not match with ECS definition type: keyword
[2] field "message" is undefined: actual mapping type (text) does not match with ECS definition type: match_only_text
[3] field "related.ip" is undefined: actual mapping type (text) does not match with ECS definition type: ip
[4] field "tags" is undefined: actual mapping type (text) does not match with ECS definition type: keyword
[5] field "vulnerability.reference" is undefined: actual mapping type (text) does not match with ECS definition type: keyword
```

</details>

<details>

<summary>Teleport changes moved to #12851</summary>

- teleport (ingest pipeline `event-groups`)
```
[0] field "teleport.audit.account_id" is undefined: field definition not found
[1] field "teleport.audit.aws_host" is undefined: field definition not found
[2] field "teleport.audit.aws_region" is undefined: field definition not found
[3] field "teleport.audit.aws_service" is undefined: field definition not found
[4] field "teleport.audit.exit_code" is undefined: field definition not found
[5] field "teleport.audit.instance_id" is undefined: field definition not found
[6] field "teleport.audit.region" is undefined: field definition not found
[7] field "teleport.audit.status" is undefined: field definition not found
[8] field "teleport.audit.target" is undefined: field definition not found
```

</details>

## Checklist

## Author's Checklist

- [x] Add changelog entries for all the updated packages
- [x] Test with `elastic-package` enabled mappings https://buildkite.com/elastic/integrations/builds/21924
- [x] Is it required to update transform settings (destination index)?
- [x] Remove debug changes `.buildkite` folder and `go.mod`/`go.sum` files.
- [x] Remove unnecessary comments from package files.

Run elastic-package with mappings validation enabled and run system tests:
```shell
cd packages/<package>
elastic-package stack up -v -d --version <kibana_constraint>
export ELASTIC_PACKAGE_FIELD_VALIDATION_TEST_METHOD=mappings
# Add defer-cleanup to check the data ingested in the corresponding indices/data streams
elastic-package test system -v --defer-cleanup 900s

elastic-package stack down -v
```

## Related issues

- Relates https://github.com/elastic/elastic-package/issues/2207
- Relates https://github.com/elastic/elastic-package/issues/2341


packages/box_events/data_stream/events/fields/ecs.yml
```diff
@@ -0,0 +1,7 @@
+# This definition could be removed once Kibana constraint is updated
+# to 8.15.2 or higher. "ecs@mappings" component template would define
+# the correct dynamic template for it.
+- external: ecs
+  name: threat.enrichments.indicator.first_seen
+- external: ecs
+  name: threat.enrichments.indicator.last_seen
```
mrodm, line 7: These fields were mapped as `keyword`, but they should be `date` according to ECS:
> field "threat.enrichments.indicator.first_seen" is undefined: actual mapping type (keyword) does not match with ECS definition type: date

Added field definitions for them using external.

packages/claroty_ctd/data_stream/event/fields/ecs.yml
```diff
@@ -0,0 +1,6 @@
+# This definition could be removed once Kibana constraint is updated
+# to 8.15.2 or higher. "ecs@mappings" component template would define
+# the correct dynamic template for it.
+- external: ecs
+  name: threat.indicator.modified_at
```
mrodm, line 5: Same as in `box_events`.
Error:
> field "threat.indicator.modified_at" is undefined: actual mapping type (keyword) does not match with ECS definition type: date

packages/crowdstrike/data_stream/fdr/fields/fields.yml
```diff
@@ -27,7 +27,7 @@
       type: long
     - name: AsepWrittenCount
       type: long
-    - name: assessments.*
+    - name: assessments
```
mrodm, line 30: All the fields under assessments were not defined (no mappings for it).
This field definition should be without `*` to avoid creating a dynamic template.

packages/github/elasticsearch/transform/latest_code_scanning/transform.yml
```diff
@@ -10,7 +10,7 @@ source:
 # that ability in order to prevent having duplicate data and prevent query
 # time field type conflicts.
 dest:
-  index: "logs-github_latest.dest_code_scanning-1"
+  index: "logs-github_latest.dest_code_scanning-2"
```
mrodm, line 13: If a new field definition is added to a transform, should this index be updated (suffix)?

packages/github/elasticsearch/transform/latest_code_scanning/fields/ecs.yml
```diff
@@ -38,3 +38,5 @@
   name: rule.name
 - external: ecs
   name: tags
+- external: ecs
+  name: message
```
mrodm, line 42: `message` field was mapped as keyword, but according to ECS should be `match_only_text`

Error:
> field "message" is undefined: actual mapping type (keyword) does not match with ECS definition type: match_only_text


In the data stream does not fail, because the mapping for that field is set by the `ecs@mappings` component template.

packages/sublime_security/data_stream/email_message/fields/fields.yml
```diff
@@ -641,9 +641,22 @@
                     - name: type
                       type: keyword
                       description: The type of authentication result, derived from the field name.
+                # https://github.com/elastic/kibana/pull/204104
```
mrodm, line 644: What option should we use here for `sublime_security.email_message.headers.hops.fields.position` field definition?

The first one would map `position` as keyword.
The second option keeps the same mapping type (`long`) that it is created for that field in the last version of this package.

Should it be added descriptions to those fields? If so, could you provide them?

packages/sublime_security/data_stream/email_message/fields/fields.yml
```diff
@@ -641,9 +641,22 @@
...
-                  type: object
-                  object_type: keyword
+                  type: group
+                  fields:
+                    - name: "*"
+                      type: object
+                      object_type: keyword
+                    - name: position
+                      # description: ?
+                      type: long
```
mrodm, line 659: What option should we use here for `sublime_security.email_message.headers.hops.fields.position` field definition?

The first one would map `position` as keyword.
The second option keeps the same mapping type (`long`) that it is created for that field in the last version of this package.

Should it be added descriptions to those fields? If so, could you provide them?

packages/wiz/elasticsearch/transform/latest_cdr_misconfigurations/fields/ecs.yml
```diff
@@ -28,3 +28,7 @@
   external: ecs
 - name: observer.vendor
   external: ecs
+- name: message
+  external: ecs
+- name: ecs.version
+  external: ecs
```
mrodm, line 34: Error: `field "ecs.version" is undefined: actual mapping type (text) does not match with ECS definition type: keyword`

This change ensures that `ecs.version` is set as `keyword`.

packages/wiz/elasticsearch/transform/latest_cdr_misconfigurations/fields/ecs.yml
```diff
@@ -28,3 +28,7 @@
   external: ecs
 - name: observer.vendor
   external: ecs
+- name: message
+  external: ecs
```
mrodm, line 32: Error: `field "message" is undefined: actual mapping type (text) does not match with ECS definition type: match_only_text`

This change ensures that `message` is set as `match_only_text`.

packages/wiz/elasticsearch/transform/latest_cdr_vulnerabilities/fields/fields.yml
```diff
@@ -6,6 +6,8 @@
   external: ecs
 - name: cloud.region
   external: ecs
+- name: device.id
+  external: ecs
```
mrodm, line 10: Error: `field "device.id" is undefined: actual mapping type (text) does not match with ECS definition type: keyword`

This change ensures that `device.id` is mapped as `keyword`.

packages/wiz/elasticsearch/transform/latest_cdr_vulnerabilities/fields/fields.yml
```diff
@@ -14,6 +16,8 @@
   external: ecs
 - name: vulnerability.id
   external: ecs
+- name: vulnerability.reference
+  external: ecs
```
mrodm, line 20: Error: `field "vulnerability.reference" is undefined: actual mapping type (text) does not match with ECS definition type: keyword`

This change ensures that `vulnerability.reference` is mapped as `keyword`.

packages/wiz/elasticsearch/transform/latest_cdr_vulnerabilities/fields/fields.yml
```diff
@@ -34,6 +38,14 @@
   external: ecs
 - name: event.type
   external: ecs
+- name: ecs.version
+  external: ecs
+- name: tags
+  external: ecs
```
mrodm, line 44: Error: `field "tags" is undefined: actual mapping type (text) does not match with ECS definition type: keyword`

This change ensures that `tags` is mapped as `keyword`.

packages/wiz/elasticsearch/transform/latest_cdr_vulnerabilities/fields/fields.yml
```diff
@@ -34,6 +38,14 @@
...
+- name: related.ip
+  external: ecs # should it be keyword instead of IP ? Would this be breaking change?
```
mrodm, line 46: Error: `field "related.ip" is undefined: actual mapping type (text) does not match with ECS definition type: ip`

This change is ensured that `related.ip` is mapped as `ip`.

Would this be a breaking change? Should it be kept as `text`?

packages/tychon/elasticsearch/transform/harddrive/fields/ecs.yml
```diff
@@ -60,3 +60,5 @@
   name: log.file.path
 - external: ecs
   name: tags
+- external: ecs
+  name: related.ip  # previously it was set as keyword but now it would be type IP, would that be a breaking change?
```
mrodm, line 64: Error: `field "related.ip" is undefined: actual mapping type (text) does not match with ECS definition type: ip`

This change is ensured that `related.ip` is mapped as `ip`.

Would this be a breaking change? Should it be kept as `text`?

packages/tychon/elasticsearch/transform/ciphers/fields/ecs.yml
```diff
@@ -96,6 +96,8 @@
   name: process.user.name
 - external: ecs
   name: server.address
+- external: ecs
+  name: server.ip # previously it was set as keyword but now it would be type IP, would that be a breaking change?
```
mrodm, line 100: Error: `field "server.ip" is undefined: actual mapping type (text) does not match with ECS definition type: ip`

This change is ensured that `server.ip` is mapped as `ip`.

Would this be a breaking change? Should it be kept as `text`?

packages/mimecast/data_stream/threat_intel_malware_grid/fields/ecs.yml
```diff
@@ -0,0 +1,8 @@
+# This definition could be removed once Kibana constraint is updated
+# to 8.15.2 or higher. "ecs@mappings" component template would define
+# the correct dynamic template for it.
+- external: ecs
+  name: threat.indicator.modified_at
+- external: ecs
+  name: threat.indicator.first_seen
```
mrodm, line 7: These two fields are mapped as keywords but they should `date`.

```
[0] field "threat.indicator.first_seen" is undefined: actual mapping type (keyword) does not match with ECS definition type: date
[1] field "threat.indicator.modified_at" is undefined: actual mapping type (keyword) does not match with ECS definition type: date
```

Using external `ecs` definition to update their mappings.

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
mrodm, line 1: :warning: Not sure what would be best here. 

The problem is that this Ingest pipeline fails in some steps, but this was not detected since the trigger in the default pipeline was ignoring the failures:

```yaml
   # Process most of the field groups.
   - pipeline:
       name: '{{ IngestPipeline "event-groups" }}'
       ignore_failure: true
```

Those failures cause that some fields are not removed or renamed, and therefore the mappings created in the data stream are not the expected ones. These are the errors found while testing with the sample logs in the package:
```
[0] field "teleport.audit.account_id" is undefined: field definition not found
[1] field "teleport.audit.aws_host" is undefined: field definition not found
[2] field "teleport.audit.aws_region" is undefined: field definition not found
[3] field "teleport.audit.aws_service" is undefined: field definition not found
[4] field "teleport.audit.exit_code" is undefined: field definition not found
[5] field "teleport.audit.instance_id" is undefined: field definition not found
[6] field "teleport.audit.region" is undefined: field definition not found
[7] field "teleport.audit.status" is undefined: field definition not found
[8] field "teleport.audit.target" is undefined: field definition not found
```

Could you check if there are other better options in the pipeline to solve these errors? Here, I just set override as true, but I miss the context to know what it is the best approach.

This could happen in other steps for other fields too (depending on the documents ingested). For instance:
- `teleport.audit.mfa_device.uuid` is tried to be updated with `rename` processor several times in that pipeline.
    - If it is set in the first processor (line 10), the other steps would fail since the field already exists, and the pipeline would fail and all the following step would not be run.

Maybe there are other cases.

.buildkite/scripts/common.sh
```diff
@@ -757,7 +757,7 @@ teardown_test_package() {
 }
 
 list_all_directories() {
-    find . -maxdepth 1 -mindepth 1 -type d | xargs -I {} basename {} | sort
+    find . -maxdepth 1 -mindepth 1 -type d | xargs -I {} basename {} | sort |grep -E '^(box_events|claroty_ctd|crowdstrike|github|mimecast|sublime_security|teleport|ti_anomali|ti_custom|tychon|wiz)$'
```
mrodm, line 760: To be removed changes in `.buildkite/` folder as well as `go.mod` and `go.sum` files.

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
mrodm, line 1: :warning: Not sure what would be best here. 

The problem is that this Ingest pipeline fails in some steps, but this was not detected since the trigger in the default pipeline was ignoring the failures:

```yaml
   # Process most of the field groups.
   - pipeline:
       name: '{{ IngestPipeline "event-groups" }}'
       ignore_failure: true
```

Those failures cause that some fields are not removed or renamed, and therefore the mappings created in the data stream are not the expected ones. These are the errors found while testing with the sample logs in the package:
```
[0] field "teleport.audit.account_id" is undefined: field definition not found
[1] field "teleport.audit.aws_host" is undefined: field definition not found
[2] field "teleport.audit.aws_region" is undefined: field definition not found
[3] field "teleport.audit.aws_service" is undefined: field definition not found
[4] field "teleport.audit.exit_code" is undefined: field definition not found
[5] field "teleport.audit.instance_id" is undefined: field definition not found
[6] field "teleport.audit.region" is undefined: field definition not found
[7] field "teleport.audit.status" is undefined: field definition not found
[8] field "teleport.audit.target" is undefined: field definition not found
```

Could you check if there are other better options in the pipeline to solve these errors? In the case of the [rename processor](https://www.elastic.co/guide/en/elasticsearch/reference/8.16/rename-processor.html), if the name already exists, it throws an error. Here, I just set override as true to avoid that error, but I miss the context to know what it is the best approach.


This could happen in other steps for other fields too (depending on the documents ingested). Just some examples (there could be more):
- `teleport.audit.mfa_device.uuid` is tried to be updated with `rename` processor several times in that pipeline.
    - If it is set in the first processor (line 10), the other steps would fail since the field already exists, and the pipeline would fail and all the following step would not be run.
- `teleport.audit.kubernetes.labels`
- `teleport.audit.database.query_parameters`


Maybe there are other cases.

packages/box_events/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.11.1"
+  changes:
+    - description: Add missing ECS mappings
```
efd6, line 4:
```suggestion
    - description: Add missing ECS mappings.
```

packages/box_events/data_stream/events/fields/ecs.yml
efd6: This needs a check for timestamp format to ensure no mapping failure prevents ingest.
```diff
diff --git a/packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml b/packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
index ed9382709c..b0c0baca2d 100644
--- a/packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
@@ -889,6 +889,32 @@ processors:
         }
         ctx.threat.indicator.type = "software";
         ctx.related.indicator_type.add(ctx.threat.indicator.type);
+  - date:
+      field: threat.indicator.first_seen
+      tag: date_threat_indicator_first_seen
+      target_field: threat.indicator.first_seen
+      formats:
+        - ISO8601
+      if: ctx.threat?.indicator?.first_seen != null && ctx.threat.indicator.first_seen != ''
+      on_failure:
+        - remove:
+            field: threat.indicator.first_seen
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - date:
+      field: threat.indicator.last_seen
+      tag: date_threat_indicator_last_seen
+      target_field: threat.indicator.last_seen
+      formats:
+        - ISO8601
+      if: ctx.threat?.indicator?.last_seen != null && ctx.threat.indicator.last_seen != ''
+      on_failure:
+        - remove:
+            field: threat.indicator.last_seen
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```

packages/claroty_ctd/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.4.2"
+  changes:
+    - description: Add missing ECS mappings
```
efd6, line 4:
```suggestion
    - description: Add missing ECS mappings.
```

packages/claroty_ctd/data_stream/event/fields/ecs.yml
efd6: This one is good; the field is obtained from a `date` processor.

packages/crowdstrike/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.50.1"
+  changes:
+    - description: Avoid using dynamic template for flattened fields
```
efd6, line 4:
```suggestion
    - description: Avoid using dynamic template for flattened fields.
```

packages/crowdstrike/data_stream/fdr/fields/fields.yml
efd6: Based on what we see in the data, this could be a `{"name": "assessments.*", "type": "keyword"}`, but that would be a breaking change, so this is probably the way to go. Do we need to do this?

packages/github/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.3.2"
+  changes:
+    - description: Add missing ECS field in latest_code_scanning transform
```
efd6, line 4:
```suggestion
    - description: Add missing ECS field in latest_code_scanning transform.
```

packages/github/elasticsearch/transform/latest_code_scanning/transform.yml
efd6: /cc @kcreddy

packages/github/elasticsearch/transform/latest_code_scanning/fields/ecs.yml
efd6: Why is this not more prevalent? I only see two cases in the whole repo where we define this.

packages/mimecast/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.4.4"
+  changes:
+    - description: Add missing ECS field mappings
```
efd6, line 4:
```suggestion
    - description: Add missing ECS field mappings.
```

packages/mimecast/data_stream/threat_intel_malware_customer/fields/ecs.yml
```diff
@@ -0,0 +1,8 @@
+# This definition could be removed once Kibana constraint is updated
+# to 8.15.2 or higher. "ecs@mappings" component template would define
+# the correct dynamic template for it.
+- external: ecs
+  name: threat.indicator.modified_at
+- external: ecs
+  name: threat.indicator.first_seen
+
```
efd6, line 8: Remove unnecessary final blank line.

packages/mimecast/data_stream/threat_intel_malware_grid/fields/ecs.yml
```diff
@@ -0,0 +1,8 @@
...
+
```
efd6, line 8: Remove unnecessary final blank line

packages/mimecast/data_stream/threat_intel_malware_grid/fields/ecs.yml
efd6: Fields are safe; obtained from `date` processor.

packages/sublime_security/data_stream/email_message/fields/fields.yml
efd6: I think your choice here is the correct one.

packages/sublime_security/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.5.1"
+  changes:
+    - description: Fix sublime_security.email_message.headers.hops.fields group mappings
```
efd6, line 4:
```suggestion
    - description: Fix `sublime_security.email_message.headers.hops.fields` group mappings.
```

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -872,14 +872,20 @@ processors:
       field: teleport.audit.aws_region
       target_field: cloud.region
       ignore_missing: true
+      # This was failing due to `cloud.region` already existed
+      override: true # Should it be added an if condition? Should it be added a remove processor?
```
efd6, line 876: I suspect this is the correct approach. It looks to me like we want the teleport data rather than the agent data.

/cc @andrewkroh

packages/ti_anomali/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.25.2"
+  changes:
+    - description: Add missing ECS field in intelligence datastream
```
efd6, line 4:
```suggestion
    - description: Add missing ECS field in intelligence datastream.
```

packages/ti_anomali/data_stream/intelligence/fields/ecs.yml
```diff
@@ -0,0 +1,6 @@
+# This definition could be removed once Kibana constraint is updated
+# to 8.15.2 or higher. "ecs@mappings" component template would define
+# the correct dynamic template for it.
+- external: ecs
+  name: threat.indicator.modified_at
+
```
efd6, line 6: Remove unnecessary blank line.

packages/ti_anomali/data_stream/intelligence/fields/ecs.yml
```diff
@@ -0,0 +1,6 @@
+# This definition could be removed once Kibana constraint is updated
+# to 8.15.2 or higher. "ecs@mappings" component template would define
+# the correct dynamic template for it.
+- external: ecs
+  name: threat.indicator.modified_at
```
efd6, line 5: This needs a format check:
```diff
diff --git a/packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml b/packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
index 60397ffa26..e1ca531a56 100644
--- a/packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/ti_anomali/data_stream/intelligence/elasticsearch/ingest_pipeline/default.yml
@@ -31,10 +31,13 @@ processors:
       target_field: event.created
       ignore_missing: true
 
-  - rename:
-      tag: rename_json_modified_ts
+  - date:
+      tag: date_json_modified_ts
       field: json.modified_ts
       target_field: threat.indicator.modified_at
+      formats:
+        - ISO8601
+      if: ctx.json?.modified_ts != null && ctx.json.modified_ts != ''
 
   - script:
       tag: script_calculate_deletion_scheduled_at
@@ -558,6 +561,7 @@ processors:
         - json.locations
         - json.meta.detail
         - json.meta.detail2
+        - json.modified_ts
         - json.resource_uri
         - json.sort
         - json.source_locations
```

packages/ti_custom/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.7.2"
+  changes:
+    - description: Add mapping for threat.indicator.url.original in transform
```
efd6, line 4:
```suggestion
    - description: Add mapping for threat.indicator.url.original in transform.
```

packages/tychon/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: "0.2.2"
+  changes:
+    - description: Add missing field mappings in transforms
```
efd6, line 3:
```suggestion
    - description: Add missing field mappings in transforms.
```

packages/tychon/elasticsearch/transform/arp/fields/ecs.yml
```diff
@@ -68,3 +68,5 @@
   name: network.type
 - external: ecs
   name: tags
+- external: ecs
```
efd6, line 71: I think it would be good to get @chrisberkhout to think about this when he is back.

packages/wiz/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.7.1"
+  changes:
+    - description: Add missing field mappings in transforms
```
efd6, line 4:
```suggestion
    - description: Add missing field mappings in transforms.
```

packages/wiz/elasticsearch/transform/latest_cdr_misconfigurations/fields/ecs.yml
efd6: Same query as above.

efd6 (commented):
A couple of sets of date processors need to be added and then have the tests be regenerated, patch for the processors attached (can be applied with git-am after gunzip).

[0001-add-date-processors.patch.gz](https://github.com/user-attachments/files/18761588/0001-add-date-processors.patch.gz)

packages/github/elasticsearch/transform/latest_code_scanning/transform.yml
kcreddy: We increment the index to avoid any conflicts due to mapping changes. 
I also simulated an integration upgrade from `keyword` to `match_only_text` on same index. After the upgrade, the type remained to be `keyword` and didn't change to `match_only_text`.

packages/github/elasticsearch/transform/latest_code_scanning/fields/ecs.yml
mrodm: In transforms, the `ecs@mappings` component template is not added to the index template. So, that field gets the default mapping for strings (`keyword`).

These are the component templates for the index template used in the transform:
- logs-github.latest_code_scanning-template@package
- logs-github.latest_code_scanning-template@custom

mrodm: Details of the index template:
index template transform

packages/box_events/data_stream/events/fields/ecs.yml
mrodm: Change applied via the patch you provided, thanks!

packages/ti_anomali/data_stream/intelligence/fields/ecs.yml
mrodm: Change applied via the patch you provided, thanks!

packages/sublime_security/data_stream/email_message/fields/fields.yml
mrodm: Updated field definition to use the second approach in https://github.com/elastic/integrations/pull/12624/commits/879aa336020748dc4070bab6732624c2bf292090

Should I leave the description undefined for now for this `position` field? Not familiar with the meaning of these values.

packages/crowdstrike/data_stream/fdr/fields/fields.yml
mrodm: The change I propose here is to keep using `flattened`, so everything under `assessments` will be stored as a whole JSON object.

Example of errors detected:
```
[0] field "crowdstrike.assessments.analytics_and_improvements_mac" is undefined: field definition not found
[1] field "crowdstrike.assessments.application_firewall_mac" is undefined: field definition not found
[2] field "crowdstrike.assessments.crendential_dumping_hash_mac" is undefined: field definition not found
...
```

Currently, that `flattened` mapping is not taking effect (since it creates a dynamic template that does not match), so Elasticsearch would be creating default mappings for those fields under `assessments` (as it uses Terraform this package I cannot tell exactly what mappings are).

Should we keep `flattened`?

efd6: Thanks for explaining. I'm happy with what you have.

packages/sublime_security/data_stream/email_message/fields/fields.yml
efd6: @janvi-elastic Can you help here?

janvi-elastic: Hi @mrodm ,
Below is the link for MDM Schema and also i have attached the screenshot for reference:
https://docs.sublime.security/docs/mdm

image

mrodm: Updated the description for the `position` field in https://github.com/elastic/integrations/pull/12624/commits/3ed832dc06cf70ba31ebd57a1807b0926e43d5d7, thanks!

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
mrodm: As tested in this PR https://github.com/elastic/integrations/pull/12801
Failures in this package are independent of the validation used.

In that PR, it is ensured that the system tests wait for all the test documents (270 in total) before running the validation, and it fails too.

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
mrodm: Tested to wait for all 270 documents defined in the test folder, and it also fails with the current validation used in elastic-package (no mapping checks involved).

https://github.com/elastic/integrations/pull/12801

Locally, this does not fail for me, I guess it is related to VM agents created/used in CI.

mrodm: > I suspect this is the correct approach. It looks to me like we want the teleport data rather than the agent data.

The concern with the changes that I set in this PR is that for instance `cloud.instance.id` could be overwritten 3 times (depending on the values).

As this is also can be reproduced with the current validation performed by elastic-package, should it be created a separated issue for it so you can take a look at it? In this PR, it is shown how to reproduce it in the CI #12801 

The problem is that the errors in this package could be blocking the next release elastic-package version (when merging it into the integrations repo).

cc @efd6 @jsoriano

mrodm:
I was wondering if it would be worth to move all the changes related to transforms to another Pull Request so all the other packages can be updated (exception for teleport), WDYT ? @jsoriano @efd6 
I'm not sure if there are any concerns related to the changes applied in the transforms here in this PR (e.g. this one https://github.com/elastic/integrations/pull/12624#discussion_r1951904339)

With the latest changes applied in `elastic-package`, validations in transforms are not going to be checked for now. The changes in those transforms were applied in this PR to take advantage of fixing them since they were detected here in this Buildkite build (https://buildkite.com/elastic/integrations/builds/20794).

efd6:
> I was wondering if it would be worth to move all the changes related to transforms to another Pull Request so all the other packages can be updated (exception for teleport), WDYT ?

That wouldn't make me unhappy.

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
efd6: There were related semantic concerns in #10277 where the issue is that the user wanted cloud provider identities being stored in these fields rather than the elastic agent identity.

Summary of behaviour:
- `cloud.region`: explicitly from `teleport.audit.aws_region`, `teleport.audit.db_aws_region` or `teleport.audit.region`
- `cloud.service.name`: explicitly from `teleport.audit.aws_service`
- `cloud.instance.id`: explicitly from `teleport.audit.aws_host`, `teleport.audit.db_gcp_instance_id` or `teleport.audit.instance_id`
- `cloud.account.id`: explicitly from `teleport.audit.account_id`

The filebeat `add_cloud_metadata` is always run, and it implicitly adds `cloud.{account.id,availability_zone,instance.id,machine.type,image.id,provider,region,service.name}` explaining why `cloud.service.name` and `cloud.account.id` which only have single explicit sets still have this error path behaviour.

If we enabled an approach like the solution that was implemented for the qualys case (user can specify what they are interested in), or a hard "use the data from the data source" approach, then those two fields cease to be an issue. The remaining fields, `cloud.region` and `cloud.instance.id` look like they would then just fall out (depending on how teleport handles the `teleport.audit.{region,instance_id}` cases (are they disjoint with the explicitly named fields, `teleport.audit.{aws,db_aws}` and `teleport.audit.{aws_host,db_gcp_instance_id}` respectively. Unfortunately the test cases don't give any indication about this and I cannot find any reference for it in the teleport documentation; my feeling is that the more generic case should be overridden by the more specific case, but if an override occurs, then an error should be written in and the pipeline be continued. In the case that the we see apparently inconsistent fields being set (e.g. `teleport.audit.aws_host`, `teleport.audit.db_gcp_instance_id`) this would just be an error. I'm not sure what the behaviour of `teleport.audit.aws_region` v. `teleport.audit.db_aws_region` should be; are these mutually exclusive?

mrodm:
> > I was wondering if it would be worth to move all the changes related to transforms to another Pull Request so all the other packages can be updated (exception for teleport), WDYT ?
> 
> That wouldn't make me unhappy.

The changes of the packages that were updating transform field definitions have been moved to #12841 (github, ti_custom, tychon and wiz). I've tried to keep links to the previous discussions.

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
mrodm: One option here to move forward could be creating an issue with the information we have now, and skip this `teleport` system test linking to that issue.

If the test is skipped, I'll remove the changes done to the `event-groups` ingest pipeline in this PR.

WDYT ?

mrodm: I've added the `assert` configuration to ensure that `elastic-package` would wait for that number of docs ingested into the data stream (https://github.com/elastic/integrations/pull/12624/commits/a21024c30473527a0ccc7380214dd9d9466fb18e)
```yaml
assert:
  hit_count: 270
```

.buildkite/scripts/common.sh
```diff
@@ -757,7 +757,7 @@ teardown_test_package() {
 }
 
 list_all_directories() {
-    find . -maxdepth 1 -mindepth 1 -type d | xargs -I {} basename {} | sort
+    find . -maxdepth 1 -mindepth 1 -type d | xargs -I {} basename {} | sort |grep -E '^(box_events|claroty_ctd|crowdstrike|mimecast|sublime_security|teleport|ti_anomali)$'
```
efd6, line 760: This looks like its equivalent to `ls -1 -d {box_events,claroty_ctd,crowdstrike,mimecast,sublime_security,teleport,ti_anomali}` (maybe not on mac), but is it supposed to still be in place?

go.mod
```diff
@@ -231,3 +231,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.2 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/elastic/elastic-package => github.com/mrodm/elastic-package v0.53.1-0.20250205174526-5b3f7cdba246
```
efd6, line 235: Is this still supposed to be in place?

efd6 (commented):
LGTM after clarification of nits.

.buildkite/scripts/common.sh
mrodm: These changes are going to be removed before merging. They were set to be able to test these packages with a specific elastic-package version.

go.mod
mrodm: These changes are going to be removed before merging. They were set to be able to test these packages with a specific elastic-package version.

All changes in `go.mod`, `go.sum` and files under `.buildkite` folder are going to be reverted.

efd6 (approved):

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
mrodm: @efd6 As I'm not totally confident about the changes for this ingest pipeline, I'll create a new PR just for the `teleport` changes, so it can be checked independently.

To not block merging the changes of the other packages.

mrodm:
Moved `teleport` changes to https://github.com/elastic/integrations/pull/12801

PR #12625, opened 2025-02-05
Changed ownership of logstash package to logstash team

consulthys:
## Proposed commit message

This PR changes the code ownership of the `logstash` integration package from the @elastic/stack-monitoring team to the @elastic/logstash team.

The main reason behind this proposed change is that the Logstash team completely owns this package and since they are actively developing and maintaining it, it makes sense to cede code ownership on that folder and allow them to move faster.

PR #12630, opened 2025-02-05
Update logstash owner in manifest

mrodm:
## Proposed commit message

Update `logstash` owner in the package manifest to be the same defined in `.github/CODEOWNERS` file.

This difference between `.github/CODEOWNERS` and the package manifest is causing errors in builds triggered from PRs. Example: https://buildkite.com/elastic/integrations/builds/21798

## Related issues

- https://github.com/elastic/integrations/pull/12625

### DIFF
--- a/packages/logstash/manifest.yml
+++ b/packages/logstash/manifest.yml
@@ -18,7 +18,7 @@ conditions:
   elastic:
     subscription: basic
 owner:
-  github: elastic/stack-monitoring
+  github: elastic/logstash
   type: elastic
 screenshots:
   - src: /img/kibana-logstash-log.png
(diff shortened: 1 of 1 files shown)


packages/logstash/manifest.yml
```diff
@@ -18,7 +18,7 @@ conditions:
   elastic:
     subscription: basic
 owner:
-  github: elastic/stack-monitoring
+  github: elastic/logstash
```
mrodm, line 21: Would it be needed to release a new version of this package with this change ?

cc @elastic/ecosystem

andrewkroh: I don't think a release is required given that this has no user-observable effect.

andrewkroh (approved):

jsoriano (approved):

PR #12633, opened 2025-02-06
[MS365] [Teams Call Quality] Add Teams Call Quality datastream

lucian-ioan:
## Proposed commit message
Add Teams Call Quality datastream

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/o365_metrics/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "0.4.1"
```
ritalwar, line 2: Since we are adding two new data streams, let's include them in version 0.5.0.

packages/o365_metrics/data_stream/teams_call_quality/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,143 @@
...
+      ignore_missing: true
+
+  - fingerprint:
+      fields:
+        - o365.metrics.teams.call.quality.id
+      target_field: _id
+
+  - set:
+      field: '@timestamp'
+      tag: set_timestamp_from_report_date
```
ritalwar, line 119:
```suggestion
      tag: set_timestamp_from_teams.call.quality.start_date_time
```

packages/o365_metrics/data_stream/teams_call_quality/manifest.yml
```diff
@@ -0,0 +1,48 @@
...
+    description: Collect Microsoft 365 Teams Call Quality metrics.
+    template_path: cel.yml.hbs
+    vars:
+      - name: interval
+        type: text
+        title: Interval
+        description: How often the API is polled, supports seconds, minutes and hours.
+        show_user: true
+        required: true
+        default: 12h
```
ritalwar, line 16: Should we use a shorter interval here?

lucian-ioan: We could use 1h for example. I thought this is ok to be 12h since the data can be very large and we wouldn't want to keep pulling it over and over due to API costs.

ritalwar: Okay, maybe we can revisit this discussion later.

ritalwar (approved):
I'm approving this for now for the initial release of the datastream, but we’ll need to update the CEL code to handle pagination. This can be addressed in the relevant backlog issues that have been created.

lucian-ioan:
Added next page support in the cel code. Trying to see If I can convert some list fields into JSONs, if not, I think this should be v1 for Teams Call Quality.

Issue #11919, opened 2024-11-28
[AWS] Introduce ignore_older & start_timestamp


Kavindu-Dodan:
**Background**

This issue is to track and upgrade AWS log integrations that utilize filebeat with `ignore_older` & `start_timestamp` introduced through https://github.com/elastic/beats/pull/41817. 

With the mentioned properties, users can control S3 (or S3 compatible) polling based log ingestion and improve filebeat registry memory usage.

**AC**

 Once https://github.com/elastic/beats/pull/41817 is merged, expose the above-mentioned properties through relevant integrations. 

Note - Configurations will be released with beats `8.18.0`

lucabelluccini:
Hello @Kavindu-Dodan 

Thank you for the work on this enhancement already on Beats.

Please try to reach out integration developers who use the `aws-s3` input (with polling mode) in their integrations to notify them about the new settings available in 8.18.0

One key integration which would possibly benefit from this is https://www.elastic.co/guide/en/integrations/current/cloudflare_logpush.html#cloudflare_logpush-to-collect-data-from-s3-compatible-cloudflare-r2-buckets-follow-the-below-steps

Kavindu-Dodan:
@lucabelluccini Good point. New configurations introduced at Beats are related to any integration that use S3 input for data ingestion. Most of the related integrations belong to my team. And I have added the team label that maintains the Cloudflare logs integrations.

PR #12645, opened 2025-02-06
[AWS S3] Introduce start timestamp and ignore older timespan to AWS S3 based integrations

Kavindu-Dodan:
## Proposed commit message

Introduce `Ignore Older Timespan` and `Start Timestamp` properties to integrations backed by AWS S3 input,

- **Ignore Older Timespan**: Accepts a timespan in which entries are accepted for processing
- **Start Timestamp**: Accepts a timestamp from which objects are accepted for processing

Configuring these properties allows S3 input to efficiently manage its internal registry. For example, setting `Ignore Older Timespan` to `2h` makes the S3 input registry only track entries within the last 2 hours. Once entries are beyond the timespan, input can remove them from the registry, thus reducing memory consumption. 

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/11919
- Relates https://github.com/elastic/beats/pull/41817

### Screenshots

Configuration rendered (Title matching existing format)

image

### DIFF
--- a/packages/amazon_security_lake/changelog.yml
+++ b/packages/amazon_security_lake/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.5.0"
+  changes:
+    - description: Add support to configure start_timestamp & ignore_older configurations for AWS S3 backed inputs
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/12645
 - version: "2.4.0"
   changes:
     - description: Update Kibana constraint to support 9.0.0.
--- a/packages/amazon_security_lake/changelog.yml
+++ b/packages/amazon_security_lake/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "2.5.0"
   changes:
-    - description: Add support to configure start_timestamp & ignore_older configurations for AWS S3 backed inputs
+    - description: Add support to configure start_timestamp and ignore_older configurations for AWS S3 backed inputs.
       type: enhancement
       link: https://github.com/elastic/integrations/pull/12645
 - version: "2.4.0"
--- a/packages/amazon_security_lake/data_stream/event/manifest.yml
+++ b/packages/amazon_security_lake/data_stream/event/manifest.yml
@@ -98,14 +98,14 @@ streams:
         multi: false
         required: false
         show_user: false
-        description: If set, only accept bucket entries with last modified timestamp newer than the given timestamp. Accepts a timestamp in `YYYY-MM-DDTHH:MM:SSZ` format. For example, "2020-10-10T10:30:00Z" (UTC) or "2020-10-10T10:30:00Z+02:30" (with zone offset).
+        description: If set, only read S3 objects with last modified timestamp newer than the given timestamp. Accepts a timestamp in `YYYY-MM-DDTHH:MM:SSZ` format. For example, "2020-10-10T10:30:00Z" (UTC) or "2020-10-10T10:30:00Z+02:30" (with zone offset).
       - name: ignore_older
         type: text
         title: "[S3] Ignore Older Timespan"
         multi: false
         required: false
         show_user: false
-        description: If set, ignore bucket entries not within the provided timespan. Timespan is checked from the current time to processing entry's last modified timestamp. Accepts a timestamp like `48h`, `2h30m`.
+        description: If set, ignore S3 objects whose Last-Modified time is before the ignore older timespan. Timespan is checked from the current time to S3 object's Last-Modified time. Accepts a duration like `48h`, `2h30m`.
       - name: queue_url
         type: text
         title: "[SQS] Queue URL"
--- a/packages/amazon_security_lake/data_stream/event/manifest.yml
+++ b/packages/amazon_security_lake/data_stream/event/manifest.yml
@@ -92,6 +92,20 @@ streams:
         show_user: true
         default: 5
         description: Number of workers that will process the S3 objects listed. It is a required parameter for collecting logs via the AWS S3 Bucket.
+      - name: start_timestamp
+        type: text
+        title: "[S3] Start Timestamp"
+        multi: false
+        required: false
+        show_user: false
+        description: If set, only accept bucket entries with last modified timestamp newer than the given timestamp. Accepts a timestamp in `YYYY-MM-DDTHH:MM:SSZ` format. For example, "2020-10-10T10:30:00Z" (UTC) or "2020-10-10T10:30:00Z+02:30" (with zone offset).
+      - name: ignore_older
+        type: text
+        title: "[S3] Ignore Older Timespan"
+        multi: false
+        required: false
+        show_user: false
+        description: If set, ignore bucket entries not within the provided timespan. Timespan is checked from the current time to processing entry's last modified timestamp. Accepts a timestamp like `48h`, `2h30m`.
       - name: queue_url
         type: text
         title: "[SQS] Queue URL"
--- a/packages/aws/manifest.yml
+++ b/packages/aws/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.3.1
 name: aws
 title: AWS
-version: 2.41.1
+version: 2.42.0
 description: Collect logs and metrics from Amazon Web Services (AWS) with Elastic Agent.
 type: integration
 categories:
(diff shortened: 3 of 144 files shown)


packages/cloudflare_logpush/data_stream/audit/agent/stream/aws-s3.yml.hbs
```diff
@@ -21,6 +21,14 @@ bucket_list_prefix: {{ bucket_list_prefix }}
 bucket_list_interval: {{ bucket_list_interval }}
 {{/if}}
 
+{{#if start_timestamp}}
+start_timestamp: {{start_timestamp}}
+{{/if}}
+
```
gizas, line 27: nit: not needed the newline

gizas, line 27:
```suggestion
```

packages/cloudflare_logpush/data_stream/casb/agent/stream/aws-s3.yml.hbs
```diff
@@ -21,6 +21,15 @@ bucket_list_prefix: {{ bucket_list_prefix }}
 bucket_list_interval: {{ bucket_list_interval }}
 {{/if}}
 
+{{#if start_timestamp}}
+start_timestamp: {{start_timestamp}}
+{{/if}}
+
```
gizas, line 27: nit: remove newline

gizas (approved):

agithomas:
@Kavindu-Dodan , could you please consider backporting the beats changes so that we can bump up the package versions only to `8.16.5` instead of `8.18.0`?

packages/aws/manifest.yml
```diff
@@ -11,7 +11,7 @@ conditions:
   elastic:
     subscription: basic
   kibana:
-    version: "^8.16.2"
+    version: "^8.18.0"
```
zmoog, line 14: Increasing the min version from 8.16.2 to 8.18 is a big step up. Enterprise customers may not be able to leverage this change in months.

We should consider backporting the Beats change to 8.16.

zmoog (commented):
We should consider a backporting the Beats change to keep the min stack version in the 8.16.x series.

packages/aws/manifest.yml
Kavindu-Dodan: But let me know if you both think this is worth a backport and if we could release this earlier

zmoog: The main downsides I see in bumping the min stack from 8.16 to 8.18 are:

- enterprise users will not benefit from this change before the 8.18.1 [^1]release (which means 2-3 months from now?)
- the need to backport bug fixes on any integration affected by this PR for 8.16 and 8.17 users

[^1]: some users have the practice—sometimes a policy—not to upgrade before a .1 release.

Kavindu-Dodan: @zmoog yeah that's a valid point. But do we backport features to minor releases ? If we backpot to 8.16.x & 8.17.x, release of them are on March 4th. And if we do not backport features to minor releases and wait for 8.18.0, then the release is on March 25

Kavindu-Dodan: After some discussions, we decided to go ahead with backports to 8.16.x and 8.17.x tracks.

Kavindu-Dodan:
Pending backports to 8.16.x [^1] & 8.17.x tracks. I will update this PR accordingly 

[^1] : https://github.com/elastic/beats/pull/42716
[^2]: https://github.com/elastic/beats/pull/42717

agithomas:
@Kavindu-Dodan , we have a great illustration of this feature's behaviour as part of the description of this [issue](https://github.com/elastic/beats/pull/41817). 

Do you think that it will be good to include this illustration in [AWS documentation](https://www.elastic.co/guide/en/integrations/current/aws.html)?  [Reference: Azure Integration](https://www.elastic.co/guide/en/integrations/current/azure.html) I think, if we include this illustration, it would avoid misconfiguration of this setting.

@andrewkroh, could you please share your opinion - if it is a good idea to include the illustration as part of the AWS documentation?

The changes look good to me. As discussed, kindly merge the changes after the `8.16.5` is available. Also, let us get the approval from the security-integrations team.

Kavindu-Dodan:
@agithomas good point on the documentation. I thought of adding some diagrams, but when checking the current AWS integration documentation, I couldn't find a suitable place to add the details. So I went with detailed descriptions for the integration input fields instead. 

Anyway, let me know if there's better suggestion.

packages/amazon_security_lake/data_stream/event/manifest.yml
```diff
@@ -92,6 +92,20 @@ streams:
...
+        required: false
+        show_user: false
+        description: If set, only accept bucket entries with last modified timestamp newer than the given timestamp. Accepts a timestamp in `YYYY-MM-DDTHH:MM:SSZ` format. For example, "2020-10-10T10:30:00Z" (UTC) or "2020-10-10T10:30:00Z+02:30" (with zone offset).
+      - name: ignore_older
+        type: text
+        title: "[S3] Ignore Older Timespan"
+        multi: false
+        required: false
+        show_user: false
+        description: If set, ignore bucket entries not within the provided timespan. Timespan is checked from the current time to processing entry's last modified timestamp. Accepts a timestamp like `48h`, `2h30m`.
```
andrewkroh, line 108: Similar to my previous comment,

- "bucket entries" -> S3 objects
- "processing entry's" -> the S3 object's

-----

> a timestamp like `48h`

I would call that a "duration" instead of "timestamp"

-----

> not within

It probably doesn't matter to most readers, but I am left wondering if this means for the boundary condition. I think it would be more clear to use the term "before" like, "Ignore S3 objects whose Last-Modified time is _before_ the ignore older time-span."

andrewkroh (commented):
> @andrewkroh, could you please share your opinion - if it is a good idea to include the illustration as part of the AWS documentation?

Personally, I don't think they are necessary to convey the meaning of the configuration options. If we need to provide additional context, then I would suggest linking to the reference docs like, "See the [input reference documentation] for more details." And add images in the reference docs.

agithomas:
> If we need to provide additional context, then I would suggest linking to the reference docs like, "See the [input reference documentation] for more details." And add images in the reference docs.

Thanks @andrewkroh for your inputs. Yes, i agree that it would be the best approach. 

@Kavindu-Dodan , it need not be part of this PR, it could be a separate PR by including the documentation team as reviewers.

packages/amazon_security_lake/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.4.0"
+  changes:
+    - description: Add support to configure start_timestamp & ignore_older configurations for AWS S3 backed inputs
```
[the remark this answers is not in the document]
Kavindu-Dodan: Thank you, change is done. Now it reads,

```
Add support to configure start_timestamp and ignore_older configurations for AWS S3 backed inputs.
```

packages/amazon_security_lake/data_stream/event/manifest.yml
```diff
@@ -92,6 +92,20 @@ streams:
         show_user: true
         default: 5
         description: Number of workers that will process the S3 objects listed. It is a required parameter for collecting logs via the AWS S3 Bucket.
+      - name: start_timestamp
+        type: text
+        title: "[S3] Start Timestamp"
+        multi: false
+        required: false
+        show_user: false
+        description: If set, only accept bucket entries with last modified timestamp newer than the given timestamp. Accepts a timestamp in `YYYY-MM-DDTHH:MM:SSZ` format. For example, "2020-10-10T10:30:00Z" (UTC) or "2020-10-10T10:30:00Z+02:30" (with zone offset).
```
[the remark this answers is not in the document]
Kavindu-Dodan: Updated description,

```
If set, only read S3 objects with last modified timestamp newer than the given timestamp. Accepts a timestamp in `YYYY-MM-DDTHH:MM:SSZ` format. For example, "2020-10-10T10:30:00Z" (UTC) or "2020-10-10T10:30:00Z+02:30" (with zone offset).
```

packages/amazon_security_lake/data_stream/event/manifest.yml
Kavindu-Dodan: Yup agree with the points. 

Updated description,

```
If set, ignore S3 objects whose Last-Modified time is before the ignore older timespan. Timespan is checked from the current time to S3 object's Last-Modified time. Accepts a duration like `48h`, `2h30m`.
```

andrewkroh (approved):

agithomas (approved):
LGTM!

Kavindu-Dodan:
Marking as draft and avoiding merging till the First week of March 2025

Kavindu-Dodan (commented):
Merging today as 8.16.5 was released and available - https://www.elastic.co/blog/elastic-stack-8-16-5-released

PR #12666, opened 2025-02-07
google_workspace: Omit pageToken from interval requests in chrome events

kcreddy:
## Proposed commit message
```
In chrome events, next_page_token is stored at the root-level 
of state. As it is enclosed in state.with(), this leads to 
next_page_token getting carried into subsequent 
interval requests. 

This change moves next_page_token inside cursor, thus 
removing from root-level. Also simplifies the pageToken query 
parameter calculation.
```

## Checklist

**Tested using mito:**
1. Token at root-level (this carries token to next request.):
    ```
    mito -data <(echo '{"token": "TOKEN1"}') <(echo '
    state.with(
      {
        ?"token": (1 > 2) ? optional.of("TOKEN2") : optional.none()
      }
    )
    ')
    ```
    Returns:
    ```
    {
	    "token": "TOKEN1"
    }
    ```
    

2. Token at next-level (this omits token from next request):
    ```
    mito -data <(echo '{"next": {"token": "TOKEN1"}}') <(echo '
    state.with({
      "next": {
        ?"token": (1 > 2) ? optional.of("TOKEN2") : optional.none(),
      }
    })
    ')
    ```
    Returns:
    ```
    {
	    "next": {}
    }
    ```

### DIFF
--- a/packages/google_workspace/changelog.yml
+++ b/packages/google_workspace/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.30.2"
+  changes:
+    - description: Fix pageToken from reappearing in interval requests.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/1
 - version: "2.30.1"
   changes:
     - description: Fix flaky test by renaming incorrect as.asn and as.organization_name fields.
--- a/packages/google_workspace/changelog.yml
+++ b/packages/google_workspace/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "2.30.2"
   changes:
-    - description: Fix pageToken from reappearing in interval requests.
+    - description: Prevent pageToken from incorrectly reappearing in interval requests.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/12666
 - version: "2.30.1"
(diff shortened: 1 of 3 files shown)


packages/google_workspace/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.30.2"
+  changes:
+    - description: Fix pageToken from reappearing in interval requests.
```
efd6, line 4:
```suggestion
    - description: Prevent pageToken from incorrectly reappearing in interval requests.
```

packages/google_workspace/data_stream/chrome/agent/stream/cel.yml.hbs
```diff
@@ -56,9 +56,9 @@ program: |
           ),
         "cursor": {
            "last_timestamp": state.end_time,
+           ?"next_page_token": body.?nextPageToken,
          },
```
efd6, line 60: I was wondering whether this should be
```
"next_page": {
    ?"token": body.?nextPageToken,
}
```
and leaving the cursor alone, but I think this is ok, and it has the behaviour that if execution is interrupted needing to go back to cursor state, it will continue from the correct point.

efd6 (approved):

packages/google_workspace/changelog.yml
kcreddy: Updated in [efae084](https://github.com/elastic/integrations/pull/12666/commits/efae084c725011a6b0d954880c7a8566b2dba4f3)

efd6 (approved):

PR #12668, opened 2025-02-07
Try to trigger workflow to block paths on PRs created by github actions

jsoriano:
Github actions are not executed on events caused by creating pull requests from Github actions, to avoid infinite recursion. In our case this blocks PRs automated with actions.

Try to trigger the block paths workflow after checking the suites.


mrodm (approved):
:+1:

Issue #12045, opened 2024-12-09
Retain event.original value upon pipeline errors


andrewkroh:
When an ingest pipeline error occurs, our log pipelines should retain the `event.original` value  to ensure that no data loss occurs and to facilitate correcting the failure. If processing is interrupted due to an error, some data may not have been extracted (i.e. incomplete processing) so it's important to retain the `event.original`. And secondly, in order for the package maintainers to be able to take action on reports of pipeline failures they nearly always need the `event.original` value to reproduce and understand the issue.

To implement this we should complete the work related to #10072. This not strictly required, but it helps ensure the `event.original` is consistently handled. We want the Fleet final_pipeline to be responsible for deleting event.original when `tags` does not contain `preserve_original_event`.

Next, in the primary pipeline of each log data stream we update the global `on_failure` handler to inject `preserve_original_event` into `tags`. This will accompany `event.kind: pipeline_error`. With this mechanism, users can still override this behavior through the various levels of `@custom` pipelines by deleting the tag value. I expect the work to be accomplished "mechanically", and this mechanism can be applied separately on the integrations owned by each SIT team.

TODO

- [ ] Tag events with `preserve_original_event` in primary on_failure handlers

andrewkroh:
https://github.com/elastic/integrations/pull/12046 contains the script that was used to apply changes to the security service team integrations.

PR #12685, opened 2025-02-10
Cloudtrail add actor and target

romulets:
**This is a re-open of https://github.com/elastic/integrations/pull/11245, that had to be closed because a full commit history rewrite in the integration main.**

### Background
Elastic Cloud Security Team has been focusing, this past year, on Cloud Detection and Response (CDR). One of the first steps towards the CDR vision is to enhance investigation workflows for the Cloud Security use-case in SIEM.

As part of enhancing investigation workflows it's necessary to be able to correlate events and entities. Meaning, if an alert is triggered on the ec2 instance i-000000000, it is of great value to easily be able to search all the events related to that entity, across multiple indices, with one query. Therefore we are working on extracting entities and enabling them to be correlated.

### What is an entity?

The term "entity" is broader than the current set of available fields under related. Although ip, user and hosts can be identities, there is a lack of space to represent messaging queues, load balancers, storage systems, databases and others. Therefore the proposal to add a new field.

### The proposed structure

There are two fields being added on this PR:

- `actor.entity.id` captures entities that started the event, the actors
- `target.entity.id` captures entities that were affected by the event. Being that created, updated, listed. We try to do as much as possible with the data present in the event.

#### Structure

The painless script turned very large. There are essentially three parts to it:

- Definition of helper functions. They are meant to facilitate the handling of the collections (`related`, `actor` and `target`). 
- Definition of enriching functions per AWS service. Even though there is no defined structure to `requestParameters` and `repsonseElements`, there is, usually a somewhat coherent structure per AWS service. I believe such separation brings better reading, creates a better headspace once working in a specific service and also breaks down the huge `if else` chain present in the previous state of the code
- Calling functions and setting fields.

#### Why [TreeSet](https://docs.oracle.com/javase/8/docs/api/java/util/TreeSet.html) as datastructure to hold `related`, `actor`, `target`.

There are two properties that this script must have:

- Values must be unique
- Values must be sorted (for testability and consistency on production)

Previously I had ensured both properties on "post processing", at the end of the script. Now it's ensured by the data structure itself.

I have not performance tested myself, but the usage of TreeSet should improve the time complexity of the algorithm, since we sort data on `add`, and previously we had to `sort` afterwards. I couldn't find a reliable source for time complexity of `TreeSet.add` vs `Collections.sort` - and honestly, the size of the list is so small that might not even matter.

### Amount of tests

The testing was essential to me to validate what I was doing, to verify each output. And I would like to keep the tests for future reference and ensuring we are not changing anything by mistake. But the tests are starting to get slow. Specially if you compare with other integrations, such as `okta`.


efd6 (approved):
Approving change in packages/aws/docs/cloudtrail.md as codeowner.

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -51,13 +51,14 @@ processors:
   - script:
       description: Appends any relevant entity to `related.entity` for all events
       lang: painless
+      if: "ctx?.json?.eventName != null && ctx?.tags != null && ctx.tags.contains('actor_target_mapping')"
       on_failure: 
         - set:
             description: Add error reason
             field: error.message
             value: "{{{ _ingest.on_failure_message }}}"
```
andrewkroh, line 59: Please add more details to the error.

```suggestion
            value: "{{{ _ingest.on_failure_message }}}"
            value: >-
              Processor '{{{ _ingest.on_failure_processor_type }}}'
              {{#_ingest.on_failure_processor_tag}}with tag '{{{ _ingest.on_failure_processor_tag }}}'
              {{/_ingest.on_failure_processor_tag}}failed with message '{{{ _ingest.on_failure_message }}}'
```

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -51,13 +51,14 @@ processors:
   - script:
       description: Appends any relevant entity to `related.entity` for all events
       lang: painless
```
andrewkroh, line 53: Add a `tag` field to the processor to help with debugging and observing metrics. e.g.

`tag: script_append_related_entity`

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -51,13 +51,14 @@ processors:
...
+      if: "ctx?.json?.eventName != null && ctx?.tags != null && ctx.tags.contains('actor_target_mapping')"
       on_failure:
```
andrewkroh, line 55:
```suggestion
      on_failure: 
        - append:
            field: tags
            value: preserve_original_event
            allow_duplicates: false
```

Relates to https://github.com/elastic/integrations/issues/12045. This would allow for easier resolutions when things fail.

romulets: Why do we need the append if we would have added the tag `script_append_related_entity` already? (suggested in a previous comment)

andrewkroh: Those are two different tag concepts. This comment is about the ECS `tags` field in the event. `script_append_related_entity` is about tagging the processor in the ingest pipeline (which shows up in [_nodes/stats](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-stats.html#cluster-nodes-stats-ingest-ex) and exceptions).

romulets: Check, I'm adding

agithomas:
For the [pipeline segment](https://github.com/elastic/integrations/pull/12685/files#diff-7b31d92d1c238c069c67b64b87d3f6fe33205ce68880888fb3edc0e4b4512ba8R774-R790), do you still want to keep it? This may be irrelevant following the change mentioned [here](https://aws.amazon.com/blogs/security/modifications-to-aws-cloudtrail-event-data-of-iam-identity-center/). Please re-check

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -146,113 +173,599 @@ processors:
...
+        enrichElasticLoadBalancing(enrichCtx, eventSource, eventName);
+        enrichDynamoDB(enrichCtx, eventSource, eventName);
         
-        field("related.entity").set(entities);
+        if (!enrichCtx.target.isEmpty()) {
+          field("target.entity.id").set(enrichCtx.target);
+          enrichCtx.related.addAll(enrichCtx.target);
+        }
+
+        field("actor.entity.id").set([ enrichCtx.actor ]);
```
agithomas, line 766: Could there be any anonymous access or cases where the value of `enrichCtx.actor`/`enrichCtx.related`  is null?

I hope,  i am referring to the correct [AWS documentation ](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-user-identity.html)

The `userIdentity element` for a request made on behalf of an IAM Identity Center user, 

```
"userIdentity": {
    "type": "IdentityCenterUser",
    "accountId": "123456789012",
    "onBehalfOf": {
        "userId": "544894e8-80c1-707f-60e3-3ba6510dfac1",
        "identityStoreArn": "arn:aws:identitystore::123456789012:identitystore/d-9067642ac7" 
    },
    "credentialId": "EXAMPLEVHULjJdTUdPJfofVa1sufHDoj7aYcOYcxFVllWR_Whr1fEXAMPLE"
}
```

Can you verify if this scenario is handled?

romulets: You are correct, this has not been handled, I'll add it

agithomas (approved):
LGTM!

PR #12231, opened 2025-01-06
[jamf_pro] Inventory pagination fix

chrisberkhout:
## Proposed commit message

```
[jamf_pro] Inventory pagination fix

* Always include the General section of inventory data.
* Update the cursor whenever there is new data.
* Assume there isn't another page if the current page isn't full.
* Correct the name of the `enable_section_content_caching` variable.
* Adjust the `page_size` variable description to correct the maximum.
* Simplify code to set the inventory filter parameter.

This has been manually tested against the live API.
```

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Discussion

I ran this against the live endpoint and checked a couple of pagination sequences. Here's some partial data from the logs (not the full results objects), that shows the cursor updates and sequence continuation/end are done correctly:

<details>
<summary>Expand for the partial request log data</summary>

```json
// first sequence

{
  "@timestamp": "2025-01-06T11:14:20.308Z",
  "url.query": "page-size=4&section=GENERAL&section=HARDWARE&section=OPERATING_SYSTEM&sort=general.reportDate%3Aasc"
}
{
  "http.response.body.content": {
    "totalCount": 8,
    "results": [
      { "reportDate": "2024-06-19T15:54:33.186Z" },
      { "reportDate": "2024-06-19T15:54:34.353Z" },
      { "reportDate": "2024-06-19T15:54:37.692Z" },
      { "reportDate": "2024-06-19T15:54:39.687Z" }
    ]
  }
}
{
  "@timestamp": "2025-01-06T11:14:21.195Z",
  "url.query": "filter=general.reportDate%3E%3D%222024-06-19T15%3A54%3A39.687Z%22&page-size=4&section=GENERAL&section=HARDWARE&section=OPERATING_SYSTEM&sort=general.reportDate%3Aasc"
}
{
  "http.response.body.content": {
    "totalCount": 5,
    "results": [
[...]

Issue #12021, opened 2024-12-06
[jamf_pro]: Integration is only sporadically importing device inventory


dkdavlar:
### Integration Name

Jamf Pro [jamf_pro]

### Dataset Name

inventory

### Integration Version

0.1.3

### Agent Version

8.15.2

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.15.2

### OS Version and Architecture

Windows Server 2022 Datacenter 21H2 - Build 20348.2849

### Software/API Version

Jamf Pro 11.10.2-t1729874551

### Error Message

Logs from the agent https://pastebin.com/yvTsuWeL

### Event Original

Data does not get imported into Elastic so there is no event.

### What did you do?

```
{
  "package": {
    "name": "jamf_pro",
    "version": "0.1.3",
    "experimental_data_stream_features": []
  },
  "name": "jamf_pro-1",
  "namespace": "",
  "description": "",
  "policy_ids": [
    "fleet-first-agent-policy"
  ],
  "vars": {},
  "inputs": {
    "general_settings-cel": {
      "enabled": true,
      "vars": {
        "api_host": "https://SERVERNAME.com",
        "client_id": "CLIENTID",
        "client_secret": {
          "id": "CLIETSECRET",
          "isSecretRef": true
        }
      },
      "streams": {
        "jamf_pro.inventory": {
          "enabled": true,
          "vars": {
            "interval": "24h",
            "page_size": "50",
            "preserve_original_event": false,
            "enable_request_tracer": true,
            "tags": [
              "forwarded"
            ],
            "enable_section_general": false,
            "enable_section_hardware": true,
            "enable_section_operating_system": false,
            "enable_section_user_and_location": false,
            "enable_section_disk_encryption": false,
            "enable_section_purchasing": false,
            "enable_section_applications": false,
            "enable_section_storage": false,
            "enable_section_configuration_profiles": false,
            "enable_section_printers": false,
            "enable_section_security": false,
            "enable_section_services": false,
            "enable_section_local_user_accounts": false,
            "enable_section_certificates": false,
            "enable_section_attachments": false,
            "enable_section_plugins": false,
            "enable_section_package_receipts": false,
            "enable_section_fonts": false,
            "enable_section_licensed_software": false,
            "enable_section_ibeacons": false,
            "enable_section_software_updates": false,
            "enable_section_extension_attributes": false,
            "CONTENT_CACHING": false,
            "enable_section_group_memberships": false
          }
        }
      }
    },
[...]

PR #12689, opened 2025-02-10
Jamf Pro: fix flaky system tests

chemamartinez:
## Proposed commit message

After [this pagination fix](https://github.com/elastic/integrations/pull/12231), the condition to ask for more data is `size(body.results) >= state.page_size` (more context of this change [here](https://github.com/elastic/integrations/issues/12021#issuecomment-2569658632)). This condition was always met in the system tests with `page_size == 1` so the test keeps constantly requesting more data.

- Tests have been adapted to don't request more data when receiving fewer events than the page size.
- It has been added the limitation of the `page_size` minimum value.
- Removed the unused variable `max_executions`.
- Store `prev_last_report_date` to compare with current `last_report_date` and avoid repeated API requests when last report filter is the same as the previous request.

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/12314
- Closes https://github.com/elastic/integrations/issues/12278
- Closes https://github.com/elastic/integrations/issues/12280


efd6 (commented):
Does this mean that if a user sets the page size to unity, they will have a problem?

efd6:
I think in that case we need to document the limitation or fix it.

chemamartinez:
Looking at the PR that changed the logic, I think the fix here would be going back to the previous condition
```
"want_more": body.totalCount > size(body.results),
```
as the bug that Chris mentioned was fixed in the logic of populating `cursor.last_report_date`. I don't see any issues in getting back to the old logic of the `want_more` check.

What do you think about it?

Otherwise, I can just add the limitation to the `page_size` description.

efd6:
I'd be OK with that provided it was checked against a real endpoint.

chemamartinez:
@efd6, I have tested both ways of defining `want_more` and both have the same limitation when `page_size == 1`.

In that case, due to the API design, it gets stuck in an infinite loop asking for the first event on the list. That's because the filter `filter=general.reportDate` is always populated from the only event received and when making the request, the same event is returned as we can only get one.

I'll leave things as they are and document that limitation.

efd6:
Can we add a check for the last-used `last_report_date` matching the current value? If they match, this would indicate non-progression. Suggest something like
```diff
diff --git a/packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs b/packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs
index b6f2131ed4..e0da25d578 100644
--- a/packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs
+++ b/packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs
@@ -62,9 +62,9 @@ program: |-
                                                ?"event.original": state.?preserve_original_event.orValue(false) ? optional.of(e.encode_json()) : optional.none(),
                                        }
                                ),
-                               "want_more": size(body.results) >= state.page_size,
                                "cursor": state.?cursor.orValue({}).with(
                                        {
+                                               ?"prev_last_report_date": state.?cursor.last_report_date,
                                                ?"last_report_date": (size(body.results) > 0) ?
                                                        optional.of(
                                                                body.results[size(body.results) - 1].general.reportDate.as(d,
@@ -76,7 +76,9 @@ program: |-
                                        }
                                ),
                        }
-               )
+               ).as(state, state.with({
+                       "want_more": size(body.results) >= state.page_size && (state.?cursor.prev_last_report_date != state.?cursor.last_report_date)
+               }))
        )
   )
 state:
```
(untested)

chemamartinez:
Good point, added at https://github.com/elastic/integrations/pull/12689/commits/bb83c619ae100cd51d3d80c97a9f86d22fc4a58c.

It has been tested against the real API, even though we still have the limitation of `page_size == 1`, it helps with unnecessary requests every interval. Now, we just see one request per interval when the `last_report_date` filter cannot advance.

packages/jamf_pro/data_stream/inventory/manifest.yml
```diff
@@ -22,7 +22,7 @@ streams:
         required: false
         title: Inventory request page size
         show_user: false
-        description: Set page size for computers-inventory API Call. Must not exceed 1000.
+        description: Set page size for computers-inventory API Call. Its value must be within the range of 2 to 1000.
```
efd6, line 25: Is this change still required?

packages/jamf_pro/_dev/deploy/docker/files/config.yml
efd6, line 1: Do we still need to change the tests?

packages/jamf_pro/data_stream/inventory/manifest.yml
chemamartinez: I think so, the difference between before and after the last commit when `page_size = 1` is that before, `want_more` was `true` when `size(body.results) >= state.page_size` (one result was returned, and one is expected), so it was constantly asking for more data in the same interval. Now, even though `want_more` is false because of the new condition `state.?cursor.prev_last_report_date != state.?cursor.last_report_date`, in the next interval the filter `last_report_date` is the same and will return the same event every time.

I checked that behaviour with `page_size = 1` and the latest changes with the API.

packages/jamf_pro/_dev/deploy/docker/files/config.yml
chemamartinez: Same reason as for the page size limitation.

packages/jamf_pro/data_stream/inventory/manifest.yml
efd6: OK, so the fix we have is just for the cpu overuse. It would be nice to fix the whole issue, but this will do for now.

efd6 (approved):

PR #12716, opened 2025-02-11
prisma_cloud: map alert.policy.{description,name,recommendation} as text or multi_field text

### LINKED ISSUE #12609: [Prisma Cloud] Change the mappings of fields with description and recommendation for policies
Hello,

There are some fields in the Prisma Cloud integration that could have the data type changed from `keyword` to `text` and one field should be a multi-field with a `.text` field.

The fields `prisma_cloud.alert.policy.description` and `prisma_cloud.alert.policy.recommendation` should be changed from `keyword` to `text` because they provide some the description of the policy and the recommendation to fix it, which can both be pretty long.

The field `prisma_cloud.alert.policy.name` needs to be a multi-field as it has the name of the rule and it is pretty common to need to search for part of the name of the rule, since this is a `keyword` field we need to use leading and trailing wildcards.

This is an example of some values for those fields, for the last two there is no need for them to be `keyword`, for the first one, it may be used un visualizations, so it also needs to be a multi-field.

Image

efd6:
## Proposed commit message

```
prisma_cloud: map alert.policy.{description,name,recommendation} as text or multi_field text

These fields are textually flavoured, so map them as such.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #12609


chemamartinez (commented):
Could this change generate any kind of breaking with data that is already indexed on upgrades?

chemamartinez (approved):

Issue #12217, opened 2025-01-02
[Azure] [Platform Logs] Issues with processing certain fields


lucian-ioan:
There are a few issues with the current pipeline/field processing for Azure Platform Logs:

1. Missing event.kind: "pipeline_error" on Pipeline Failure
Issue: Logs are missing the event.kind: "pipeline_error" field when pipelines fail.

2. Missing azure.subscription_id in Some Logs
Issue: azure.subscription_id is missing in certain logs.
Cause: Microsoft uses the non-standard field azure.platformlogs.ResourceId instead of azure.subscription_id.

3. Inconsistent Casing for azure.subscription_id
Issue: azure.subscription_id appears in mixed cases across logs.

4. Incorrect Data Type for azure.platformlogs.durationMs
Issue: The field azure.platformlogs.durationMs contains float values, which causes processors expecting integers to fail.

lucian-ioan:
I've started testing this and will open a PR soon.

PR #12735, opened 2025-02-12
[Azure] [PlatformLogs] Fix pipeline for edge cases

lucian-ioan:
## Proposed commit message
 Fix pipeline for edge cases.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/integrations/issues/12217


ishleenk17:
@muthu-mps : Can you please review this.

packages/azure/data_stream/platformlogs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -148,7 +148,7 @@ processors:
   - convert:
       field: azure.platformlogs.durationMs
       target_field: event.duration
-      type: integer
+      type: float
```
muthu-mps, line 151: - Not sure why the `event.duration` is of type integer. As we reference the ECS field for event.duration changing the type to float may not align with the ECS field as the event.duration is of **type long**. You have to convert it to long instead of float.
- `event.duration ECS reference`- https://www.elastic.co/guide/en/ecs/current/ecs-event.html#field-event-duration
- Will this change create a mapping conflict? If Yes, then lets file a separate issue for this fix.

lucian-ioan: event.duration gets converted from azure.platformlogs.durationMs which can come as float. Since it's in ms I think I will just convert the original value to long.

muthu-mps: Yes, This value is again converted to nano seconds here https://github.com/elastic/integrations/blob/8bb411cfe291a2a51bd53cc85391a35baea43767/packages/azure/data_stream/platformlogs/elasticsearch/ingest_pipeline/default.yml#L160

So we can keep the type as long.

ishleenk17: Why do we use event.duration ?
I expect conflicts to happen in case this field is already present and datatype is long. 
Can't we use a time duration field specific for azure platform logs ?

muthu-mps: Generally, event.duration is used in all Azure integrations to capture the request duration. I don't think we can add a separate field for platform logs alone. 
To fix this we need to change `integer` field type to `long`.

lucian-ioan: Thank you, I think it will work. Will test this live before merging.

ishleenk17: @muthu-mps : We have never confronted conflicts for this field ?
Also, I assume all the event.duration are long in existing azure logs as well to adhere to the ECS fields.

muthu-mps: > We have never confronted conflicts for this field ?

No, We have not seen any conflict error.

> Also, I assume all the event.duration are long in existing azure logs as well to adhere to the ECS fields.

yes, all other Azure integrations adhere to ECS field.

PR #12757, opened 2025-02-12
Fixed ingest pipeline used in Elasticsearch transform

3kt:
## Proposed commit message

An update to the `elasticsearch` integration requires an update of the ingest pipeline reference in `packages/elasticsearch/elasticsearch/transform/index_pivot/transform.yml` since version interpolation isn't possible until https://github.com/elastic/package-spec/issues/833 is fixed.


3kt:
Will need to wait for https://github.com/elastic/integrations/pull/12813 before merging

consulthys (approved):
LGT Stack Monitoring

PR #12769, opened 2025-02-13
[cisco_meraki_metrics] wrap organization IDs in quotes

ghost:
## Overview

This PR fixes the issue where large numbers are interpreted as integers and converted to scientific notation. Wrapping org IDs in double quotes to ensure they are handled as string.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/cisco_meraki_metrics/changelog.yml
+++ b/packages/cisco_meraki_metrics/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.3.1"
+  changes:
+    - description: Wrap organization ids in quotes to avoid parsing errors.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/999
 - version: "0.3.0"
   changes:
     - description: Add support for Kibana `9.0.0`.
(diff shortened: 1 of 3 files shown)


ishleenk17:
When I check the [sample event](https://github.com/elastic/integrations/blob/015ff7f808de07219d3b65ebdae109874ed85a0e/packages/cisco_meraki_metrics/data_stream/device_health/sample_event.json#L63), looks like org id is treated as a string/keyword only. 
Is that not the case?

I assume the sample json would not need any change here.

gpop63:
> I assume the sample json would not need any change here.

@ishleenk17 yes, sample event doesn't need any change the issue is only in the agent config.

ishleenk17:
> @gpop63 When I check the [sample event](https://github.com/elastic/integrations/blob/015ff7f808de07219d3b65ebdae109874ed85a0e/packages/cisco_meraki_metrics/data_stream/device_health/sample_event.json#L63), looks like org id is treated as a string/keyword only.
> Is that not the case?

Wasn't it treated as a string before itself as I can see in the sample event.

gpop63:
> Wasn't it treated as a string before itself as I can see in the sample event.

This issue is inside the agent config that we send to beats - for some reason kibana/fleet would interpret the organization ids values as numbers even if I used the type `text`. In documents it is and was a string so that's fine.

ishleenk17 (approved):
Looks good!

PR #12773, opened 2025-02-13
[cisco_asa] Add parsing for 4 new message types

dwhyrock:
## Proposed commit message
This adds 3 previously-unhandled event codes and another example for a fourth.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -2874,8 +2894,12 @@ processors:
           outcome: success
         "722034":
           type: [ connection, info ]
+        "722035":
+          type: [ connection, info ]
         "722037":
           type: [ connection, end ]
+        "722041":
+          type: [ connection, info ]
```
dwhyrock, line 2902: @taylor-swanson do you think I should set the `event.outcome` or any other fields for these messages?

taylor-swanson: Sorry, missed the notification for this...

If it's clear from the log description that we can set it, we should. It gets a lot trickier on logs which have a lot of different outcomes. I tend to leave those alone for right now.

dwhyrock: The message for `722035` is categorized under the following list:

```
Error Messages, Severity 3

The following messages appear at severity 3, errors:
```

So I'm going to set that one as `outcome: failure`, while the `722041` is under the `warning` category in the docs, so I'll not set an `outcome` there.  

Let me know if you disagree.

Source: https://www.cisco.com/c/en/us/td/docs/security/asa/syslog/b_syslog/syslogs-sev-level.html

taylor-swanson: Agree with `722035`, `722041` could be `outcome: failure`, or at least it seems that way for me. I guess where things get more complicated is if this message isn't about the actual event/action failing, but rather an additional log explaining the failure. I took a look at the [ECS docs for event.outcome](https://www.elastic.co/guide/en/ecs/current/ecs-event.html#field-event-outcome), and this is what it had to say:

>  Note that when a single transaction is described in multiple events, each event may populate different values of event.outcome, according to their perspective.
>
> Also note that in the case of a compound event (a single event that contains multiple logical events), this field should be populated with the value that best captures the overall success or failure from the perspective of the event producer.
>
> Further note that not all events will have an associated outcome. For example, this field is generally not populated for metric events, events with event.type:info, or any events for which an outcome does not make logical sense.

So in that case, it makes sense to me to enrich it with `outcome: failure`

taylor-swanson (approved):

PR #12796, opened 2025-02-15
Billing integration dedup credits & stalling fixes

3kt:
## Proposed commit message

This fixes multiple bugs, while slightly changing the collection behavior for credits data.

1. Wrong credit computation in the credit dashboard.
2. Stalling collection when an error is encountered.

The wrong credit computation was caused by the lack of deduplication. This is fixed by ensuring we only look at the last 24h, and controlling the IDs within 24h (edge case with multiple collecting agents, or agent full reinstall).

The stalling collection was caused by a missing `api_key` carry-over when an error is faced. As a nice benefit, i've also ensured that an error while back-feeding doesn't completely stops the collection, ie the agent will attempt to move forward until "live" in spite of the error(s).

## TODOs

- [x] ID control for credits, there should be a single document for the tuple `organization_id`/`daily timestamp`/`ecu_quantity`/`start` (assuming we never have 2 orders on the same date with the exact amount of ECUs)
- [x] Fix the credits dashboard to exclude old expired line items
- [x] Ensure at least an event is returned when `want_more` is true ([reference](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-cel.html#_execution))

## Stalling collection fix

The integration will now return "empty" documents if no cost is returned for the point of collection. This ensures the `want_more` behavior is enforced.

image


packages/ess_billing/data_stream/credits/agent/stream/cel.yml.hbs
```diff
@@ -47,6 +47,8 @@ program: |
           }
         ),
         "want_more": false,
+        "organization_id": state.organization_id,
+        "api_key": state.api_key,
```
efd6, line 51: Rather than doing this, I'd recommend wrapping the whole program with `state.with()`; this will ensure that any future changes will also have state variables carried over.

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
```diff
@@ -23,86 +22,103 @@ redact:
...
+          "events": [{
+            "ess": {
+              "billing": {
+                "organization_id": state.organization_id,
+                "from": req.from.format(time_layout.RFC3339),
+                "to": req.to.format(time_layout.RFC3339),
+                "total_ecu": 0,
+              }
+            }
+          }],
```
efd6, line 91: You can return a sentinel event in the array that is dropped by the ingest pipeline. One that did not have the ess field would do, then you just check in the preample of the ingest pipeline whether that exists and drop if not.

3kt: This is a great idea! Prevents pollution from reaching the actual DS - thanks for suggesting :)

ryanburnhamelastic (approved):

efd6 (commented):
Post merge LGTM

PR #12814, opened 2025-02-17
[o365_metrics]Update ownership for O365 Metrics integration

narph:
Update ownership for O365 Metrics integration
Currently some of the data streams are owned by the @elastic/security-service-integrations because they were created by one of the team members.
All the data streams that the security team owns:
- /packages/o365_metrics/data_stream/groups_activity_group_detail
- /packages/o365_metrics/data_stream/onedrive_usage_account_detail 
- /packages/o365_metrics/data_stream/teams_user_activity_user_detail
- /packages/o365_metrics/data_stream/viva_engage_groups_activity_group_detail

Based on the integration name and scope and looking at the sample logs:
https://github.com/elastic/integrations/blob/main/packages/o365_metrics/data_stream/groups_activity_group_detail/sample_event.json
https://github.com/elastic/integrations/blob/main/packages/o365_metrics/data_stream/onedrive_usage_account_detail/sample_event.json
https://github.com/elastic/integrations/blob/main/packages/o365_metrics/data_stream/teams_user_activity_user_detail/sample_event.json
https://github.com/elastic/integrations/blob/main/packages/o365_metrics/data_stream/viva_engage_groups_activity_group_detail/sample_event.json

the data streams return exclusively metrics (specifically count and duration) and not logs or security information. 

@kcreddy feel free to chime in here.

@lalit-satapathy do you see any reason the service team should continue owning these data streams?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -462,26 +462,4 @@
 /packages/zscaler_zpa @elastic/security-service-integrations
 /packages/cisco_meraki_metrics @elastic/obs-infraobs-integrations
 /packages/panw_metrics @elastic/obs-infraobs-integrations
-/packages/o365_metrics @elastic/obs-infraobs-integrations @elastic/security-service-integrations
-/packages/o365_metrics/data_stream/active_users_services_user_counts @elastic/obs-infraobs-integrations
-/packages/o365_metrics/data_stream/groups_activity_group_detail @elastic/security-service-integrations
-/packages/o365_metrics/data_stream/onedrive_usage_account_counts @elastic/obs-infraobs-integrations
-/packages/o365_metrics/data_stream/onedrive_usage_file_counts @elastic/obs-infraobs-integrations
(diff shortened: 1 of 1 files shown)


kcreddy:
It makes sense to divide the ownership based on the data domain instead of development team. All of the data in the mentioned data-streams pertains to [Microsoft 365 Reports](https://learn.microsoft.com/en-us/microsoft-365/admin/activity-reports/activity-reports?view=o365-worldwide) and are thus related to `metrics`. These shouldn't have been split up into data-stream based ownership to begin with. 

In these 4 data-streams, the ingest pipelines and field naming convention are closely aligned with existing observability-owned data-streams. The CEL input program is slightly different and built based around user requirements. But that shouldn't be ideal for security to own these data-streams IMO as they are primarily `metrics` based. 

Also when it comes to future enhancements for these data-streams such as enabling [TSDS](https://www.elastic.co/guide/en/integrations-developer/current/developer-tsds-guidelines.html), observability has better guidance in executing these tasks rather than security.

lalit-satapathy (approved):
LGTM.

PR #12821, opened 2025-02-18
Osquery Manager - add support for Kibana 9.0.0

szwarckonrad:
Upgrade Osquery Manager for Kibana 9.0.0

### DIFF
--- a/packages/osquery_manager/manifest.yml
+++ b/packages/osquery_manager/manifest.yml
@@ -1,14 +1,14 @@
 format_version: 3.0.0
 name: osquery_manager
 title: Osquery Manager
-version: 1.15.0
+version: 1.16.0
 description: Deploy Osquery with Elastic Agent, then run and schedule queries in Kibana
 type: integration
 categories:
   - security
 conditions:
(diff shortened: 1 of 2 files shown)


packages/osquery_manager/manifest.yml
```diff
@@ -1,14 +1,14 @@
...
-version: 1.15.0
+version: 1.16.0
 description: Deploy Osquery with Elastic Agent, then run and schedule queries in Kibana
 type: integration
 categories:
   - security
 conditions:
   kibana:
-    version: ^8.16.0
+    version: "^8.16.0 || ^9.0.0"
```
tomsonpl, line 11: why not ^8.18.0 ?

tomsonpl (approved):

packages/osquery_manager/manifest.yml
marc-gr: unless there is anything specific requiring a bump to 8.18 generally we try so support as many versions as possible to avoid to force current users to upgrade

PR #12830, opened 2025-02-18
[Windows] Add custom conditions support for Perfmon

### LINKED ISSUE #12829: [Windows]: Enable custom conditionals for Perfmon
### Integration Name

Windows [windows]

### Dataset Name

perfmon

### Integration Version

2.4.2

### Agent Version

8.17.1

Windows

### User Goal

Enable using conditionals to apply custom perfmon queries to specific systems

### Existing Features

We are rolling out custom perfmon queries for MSMQ and need to be able to apply those integrations only to supported servers without needing to put each server in a unique agent policy.

### What did you see?

There is a hard coded conditional right now, want to be able to append additional conditionals.

### Anything else?

_No response_

mholttech:
## Proposed commit message
Enable using conditionals to apply custom perfmon queries to specific systems.

## Checklist

## Related issues

- Closes #12829

### DIFF
--- a/packages/windows/data_stream/perfmon/agent/stream/stream.yml.hbs
+++ b/packages/windows/data_stream/perfmon/agent/stream/stream.yml.hbs
@@ -1,9 +1,5 @@
 metricsets: ["perfmon"]
-{{#if condition }}
-condition: ${host.platform} == 'windows' and {{ condition }}
-{{ else }}
-condition: ${host.platform} == 'windows'
-{{/if}}
+condition: {{ condition }}
 perfmon.group_measurements_by_instance: {{perfmon.group_measurements_by_instance}}
 perfmon.ignore_non_existent_counters: {{perfmon.ignore_non_existent_counters}}
 perfmon.refresh_wildcard_counters: {{perfmon.refresh_wildcard_counters}}
--- a/packages/windows/data_stream/perfmon/agent/stream/stream.yml.hbs
+++ b/packages/windows/data_stream/perfmon/agent/stream/stream.yml.hbs
@@ -1,5 +1,9 @@
 metricsets: ["perfmon"]
+{{#if condition }}
+condition: ${host.platform} == 'windows' and {{ condition }}
+{{ else }}
 condition: ${host.platform} == 'windows'
+{{/if}}
 perfmon.group_measurements_by_instance: {{perfmon.group_measurements_by_instance}}
 perfmon.ignore_non_existent_counters: {{perfmon.ignore_non_existent_counters}}
 perfmon.refresh_wildcard_counters: {{perfmon.refresh_wildcard_counters}}
--- a/packages/windows/data_stream/perfmon/manifest.yml
+++ b/packages/windows/data_stream/perfmon/manifest.yml
@@ -51,8 +51,15 @@ streams:
         default: 10s
       - name: condition
         title: Condition
-        description: Condition to filter when to collect this input. See [Dynamic Input Configuration](https://www.elastic.co/guide/en/fleet/current/dynamic-input-configuration.html) for details.
+        description: | 
+               Condition to filter when to apply this input. Refer to 
+               [Host provider](https://www.elastic.co/guide/en/fleet/current/host-provider.html) 
+               to find the available keys and to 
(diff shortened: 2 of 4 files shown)


packages/windows/data_stream/perfmon/agent/stream/stream.yml.hbs
```diff
@@ -1,5 +1,9 @@
 metricsets: ["perfmon"]
+{{#if condition }}
+condition: ${host.platform} == 'windows' and {{ condition }}
+{{ else }}
```
belimawr, line 4: The only problem with this approach is that `condtition` is a space character (` `), `{{#if condition }}` will evaluate to true, rendering
```
condition: ${host.platform} == 'windows' and
```

Which will make the Elastic-Agent fail to start the input with the following error:
```
PS C:\elastic-agent-8.17.2-windows-x86_64> & 'C:\Program Files\Elastic\Agent\elastic-agent.exe' status
+- fleet
¦  +- status: (HEALTHY) Connected
+- elastic-agent
   +- status: (FAILED) Invalid component model: rendering inputs failed: invalid condition "${host.platform} == 'windows' and": condition line 1 column 33: mismatched input '<EOF>' expecting {TRUE, FALSE, FLOAT, NUMBER, NOT, NAME, STEXT, DTEXT, '(', '[', '{', '$${', '${'}
PS C:\elastic-agent-8.17.2-windows-x86_64>
```

The way it shows in Fleet UI is not the most clear either:


If I recall correctly I tried something similar in the past, the solution was to have the whole condition in the field, so `${host.platform} == 'windows'` would be the default value allowing users to freely modify it.

belimawr (commented):
@mholttech you missed updating the version on `manifest.yml`

packages/windows/data_stream/perfmon/manifest.yml
```diff
@@ -49,6 +49,13 @@ streams:
         required: true
         show_user: true
         default: 10s
+      - name: condition
+        title: Condition
+        description: Condition to filter when to collect this input. See [Dynamic Input Configuration](https://www.elastic.co/guide/en/fleet/current/dynamic-input-configuration.html) for details.
+        type: text
+        multi: false
+        required: false
+        show_user: false
```
strawgate, line 58: Won't this hide it from the integration configuration?

strawgate (changes requested):

packages/windows/data_stream/perfmon/manifest.yml
mholttech: `show_user: false` puts this under "Advanced Options" instead of making it always appear. I figured that would be a good default for this one because it only needs to be modified in specific use cases, especially since it looks like we need to change the implementation slightly per @belimawr's comments

https://github.com/elastic/package-spec/blob/964c4a69e024cc464c4808720ba0db9f001a82a7/spec/integration/data_stream/manifest.spec.yml#L119-L123

packages/windows/data_stream/perfmon/agent/stream/stream.yml.hbs
mholttech: That's odd. I'll go ahead and make the change and update the manifest version but it seems like there might be some inconsitencies on how the agent handles this in various concepts.

I grabbed this particular pattern from a k8s datastream that i'm familiar with https://github.com/elastic/integrations/blob/01ffd843ae73826ecf98d223580ce645e4d6d6a3/packages/kubernetes/data_stream/scheduler/agent/stream/stream.yml.hbs#L13-L17

belimawr: > (...) it seems like there might be some inconsitencies on how the agent handles this in various situations.

Integrations are owned by different teams and we have a huge number of integrations, sometimes different teams have different approaches to solve the same problem.

belimawr (approved):

packages/windows/data_stream/perfmon/manifest.yml
belimawr: I like keeping it under the "Advanced options"

VihasMakwana (approved):

andrewkroh (approved):

belimawr:
@mholttech I see there are some conflicts, could you take a look at them?

PR #12834, opened 2025-02-19
jamf_pro: normalize jamf_pro.inventory.operating_system.version and os.version to three-part versions

### LINKED ISSUE #12799: [jamf_pro]: Normalize macOS version number
### Integration Name

Jamf Pro [jamf_pro]

### Dataset Name

inventory

### Integration Version

0.2.3

### Agent Version

8.15.5

Windows Server 2022 Datacenter 21H2 - Build 20348.2849

### User Goal

I want to use the reporting for macOS versions.

### Existing Features

The macOS version is currently imported, but because the version is not normalized, then each version can be written in two different ways making dashboards using this data useless.

### What did you see?

Jamf has two ways of collecting macOS versions:

- Using the Jamf binary which returns the version like: 15.1.0
- Using Declarative Device Management which will only return 15.1

This means that the dashboard looks like this and is pretty much useless:

Image

### Anything else?

Should this be fixed on Jamfs end? Probably, but it's been an open issue for several years.

efd6:
## Proposed commit message

```
jamf_pro: normalize jamf_pro.inventory.operating_system.version and os.version to three-part versions

Depending on the source of the Jamf data, we may receive two-part (e.g.
15.1) or three-part (e.g. 15.1.0) versions for versions where the last
part is zero. This leads to multiplication of the numbers of
semantically identical versions in search results and dashboards. So
normalise all OS versions to the three-part form, unless there are other
syntax components (this should never happen).
```

> [!NOTE]
> The dashboard uses `jamf_pro.inventory.operating_system.version`, so this why we need to change both.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #12799

### DIFF
--- a/packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json
+++ b/packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json
@@ -388,6 +388,70 @@
                 "purchasing": null,
                 "udid": "21ED95A7-FF9D-52BD-A55B-36D54585083A"
             }
+        },
+        {
+            "message": {
+                "applications": [
+                    {
+                        "bundleId": "com.apple.freeform",
+                        "externalVersionId": "0",
+                        "macAppStore": false,
+                        "name": "Freeform.app",
+                        "path": "/System/Applications/Freeform.app",
+                        "sizeMegabytes": 57,
+                        "updateAvailable": false,
+                        "version": "2.4"
+                    }
+                ],
+                "operatingSystem": {
+                    "activeDirectoryStatus": "Not Bound",
+                    "build": "23H311",
+                    "extensionAttributes": [],
+                    "fileVault2Status": "NOT_ENCRYPTED",
+                    "name": "macOS",
+                    "rapidSecurityResponse": null,
+                    "softwareUpdateDeviceId": "J314cAP",
+                    "supplementalBuildVersion": "23H311",
+                    "version": "14.7"
+                },
+                "packageReceipts": null,
+                "plugins": null,
+                "printers": null,
+                "purchasing": null,
+                "udid": "21ED95A7-FF9D-52BD-A55B-36D54585083A"
+            }
+        },
+        {
+            "message": {
+                "applications": [
+                    {
+                        "bundleId": "com.apple.freeform",
+                        "externalVersionId": "0",
+                        "macAppStore": false,
+                        "name": "Freeform.app",
+                        "path": "/System/Applications/Freeform.app",
+                        "sizeMegabytes": 57,
+                        "updateAvailable": false,
+                        "version": "2.4"
+                    }
+                ],
+                "operatingSystem": {
+                    "activeDirectoryStatus": "Not Bound",
+                    "build": "23H311",
+                    "extensionAttributes": [],
+                    "fileVault2Status": "NOT_ENCRYPTED",
+                    "name": "macOS",
+                    "rapidSecurityResponse": null,
+                    "softwareUpdateDeviceId": "J314cAP",
+                    "supplementalBuildVersion": "23H311",
+                    "version": "14"
+                },
+                "packageReceipts": null,
+                "plugins": null,
+                "printers": null,
+                "purchasing": null,
+                "udid": "21ED95A7-FF9D-52BD-A55B-36D54585083A"
+            }
         }
     ]
 }
--- a/packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json-expected.json
+++ b/packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json-expected.json
@@ -421,6 +421,7 @@
                 }
             },
             "os": {
+                "full": "sonoma",
                 "name": "macOS",
                 "version": "14.7.2"
             },
@@ -464,6 +465,7 @@
                 }
             },
             "os": {
+                "full": "sonoma",
                 "name": "macOS",
                 "version": "14.7.0"
             },
@@ -507,6 +509,7 @@
                 }
             },
             "os": {
+                "full": "sonoma",
                 "name": "macOS",
                 "version": "14.0.0"
             },
--- a/packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json-expected.json
+++ b/packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json-expected.json
@@ -429,6 +429,92 @@
                     ""
                 ]
             }
+        },
+        {
+            "ecs": {
+                "version": "8.11.0"
+            },
+            "event": {
+                "kind": "asset"
+            },
+            "jamf_pro": {
+                "inventory": {
+                    "applications": [
+                        {
+                            "bundle_id": "com.apple.freeform",
+                            "external_version_id": "0",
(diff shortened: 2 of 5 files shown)


packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json
```diff
@@ -388,6 +388,70 @@
...
+                "operatingSystem": {
+                    "activeDirectoryStatus": "Not Bound",
+                    "build": "23H311",
+                    "extensionAttributes": [],
+                    "fileVault2Status": "NOT_ENCRYPTED",
+                    "name": "macOS",
+                    "rapidSecurityResponse": null,
+                    "softwareUpdateDeviceId": "J314cAP",
+                    "supplementalBuildVersion": "23H311",
+                    "version": "14.7"
```
efd6, line 415: Note this.

packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json-expected.json
```diff
@@ -429,6 +429,92 @@
...
+                        }
+                    ],
+                    "operating_system": {
+                        "active_directory_status": "Not Bound",
+                        "build": "23H311",
+                        "file_vault2status": "NOT_ENCRYPTED",
+                        "name": "macOS",
+                        "software_update_device_id": "J314cAP",
+                        "supplemental_build_version": "23H311",
+                        "version": "14.7.0"
```
efd6, line 461: Normalised.

packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json-expected.json
```diff
@@ -429,6 +429,92 @@
...
+                    },
+                    "udid": "21ED95A7-FF9D-52BD-A55B-36D54585083A"
+                }
+            },
+            "os": {
+                "name": "macOS",
+                "version": "14.7.0"
```
efd6, line 468: Normalised.

packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json
```diff
@@ -388,6 +388,70 @@
...
+                "operatingSystem": {
+                    "activeDirectoryStatus": "Not Bound",
+                    "build": "23H311",
+                    "extensionAttributes": [],
+                    "fileVault2Status": "NOT_ENCRYPTED",
+                    "name": "macOS",
+                    "rapidSecurityResponse": null,
+                    "softwareUpdateDeviceId": "J314cAP",
+                    "supplementalBuildVersion": "23H311",
+                    "version": "14"
```
efd6, line 447: Note this.

packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json-expected.json
```diff
@@ -429,6 +429,92 @@
...
+                        }
+                    ],
+                    "operating_system": {
+                        "active_directory_status": "Not Bound",
+                        "build": "23H311",
+                        "file_vault2status": "NOT_ENCRYPTED",
+                        "name": "macOS",
+                        "software_update_device_id": "J314cAP",
+                        "supplemental_build_version": "23H311",
+                        "version": "14.0.0"
```
efd6, line 504: Normalised.

packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json-expected.json
```diff
@@ -429,6 +429,92 @@
...
+                    },
+                    "udid": "21ED95A7-FF9D-52BD-A55B-36D54585083A"
+                }
+            },
+            "os": {
+                "name": "macOS",
+                "version": "14.0.0"
```
efd6, line 511: Normalised.

packages/jamf_pro/data_stream/inventory/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -123,6 +123,79 @@ processors:
...
+        }
+        return s + ".0.0";
+      }
+      String full_name(String s) {
+        if (s.startsWith('15.')) {
+          return 'sequoia';
+        }
+        if (s.startsWith('14.')) {
+          return 'sonoma';
+        }
```
kcreddy, line 158: Could it be simplified using `params` in script processor?

kcreddy (approved):
LGTM. 
Just some clarification.

packages/jamf_pro/data_stream/inventory/elasticsearch/ingest_pipeline/default.yml
efd6: It can't be a look-up since we are doing a check on the prefix. We could iterate over a map, doing the prefix check on the key and then returning the value, I'm not convinced that that is better; this is almost certainly faster.

efd6: It would look like this
```
- script:
    tag: script_normalize_operating_system_version
    lang: painless
    if: ctx.jamf_pro?.inventory?.operating_system?.version != null && ctx.jamf_pro.inventory.operating_system.version != ''
    params:
      versions:
        '15.': sequioa
        '14.': sonoma
        '13.': ventura
        '12.': monterey
        '11.': big sur
        '10.15.': catalina
        '10.14.': mojave
        '10.13.': high sierra
        '10.12.': sierra
        '10.11.': el capitan
        '10.10.': yosemite
        '10.9.': mavericks
    source: |-
      String normalize(String s) {
        int n = 0;
        for (int i = 0; i < s.length(); i++){
            char c = s.charAt(i);
            if (c == (char)'.') {
              n++;
              continue;
            }
            if (c < (char)'0' || (char)'9' < c) {
              // If we have non-numeric parts, bail.
              return s;
            }
        }
        if (n >= 2) {
            return s;
        }
        if (n == 1) {
            return s + ".0";
        }
        return s + ".0.0";
      }
      String full_name(String s, Map versions) {
        for (def e: versions.entrySet()) {
          if (s.startsWith(e.getKey())) {
            return e.getValue();
          }
        }
        return '';
      }
      ctx.jamf_pro.inventory.operating_system.version = normalize(ctx.jamf_pro.inventory.operating_system.version);
      String name = full_name(ctx.jamf_pro.inventory.operating_system.version, params.get('versions'));
      if (name != '') {
        if (ctx.os == null) {
          ctx.os = [:];
        }
        ctx.os.full = name;
      }
```

kcreddy: I was thinking since it's already normalised,  we can split based on . and get first value and get that key's value

efd6: Version 10.x. has different behaviour to all the others. So it would be a two step in that case; find anything in `11+.`, otherwise find anything in `10.x.`. The normalisation doesn't really help here.

kcreddy: Got it 👍🏼

PR #12845, opened 2025-02-19
Make the Filestream integration GA

belimawr:
## Proposed commit message

See title

## Checklist

~~## Author's Checklist~~

- Build the integration package
- Start a stack with `elastic-package` from the Filestream integration folder
- Open Kibana
- Click on "Add integrations"
- Search for "Filestream"
- Ensure it can be found *without* selecting "Display beta integrations"
- Ensure it does not have the "technical preview" badge.

~~## Related issues~~

### DIFF
--- a/packages/filestream/manifest.yml
+++ b/packages/filestream/manifest.yml
@@ -3,7 +3,7 @@ name: filestream
 title: Custom Filestream Logs
 description: Collect log data using filestream with Elastic Agent.
 type: integration
-version: 0.1.0
+version: 1.0.0
 conditions:
   kibana:
     version: "^8.15.0 || ^9.0.0"
--- a/packages/filestream/changelog.yml
+++ b/packages/filestream/changelog.yml
@@ -1,3 +1,8 @@
+- version: "1.0.0"
+  changes:
+    - description: Make it GA
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/12845
 - version: "0.1.0"
   changes:
     - description: Make it available in 9.0


jlind23 (approved):

packages/filestream/manifest.yml
```diff
@@ -3,7 +3,7 @@ name: filestream
 title: Custom Filestream Logs
 description: Collect log data using filestream with Elastic Agent.
 type: integration
```
jsoriano, line 5: @belimawr I think this package should be of input type. Is there any reason to use integration type here?

belimawr: Honestly, I don't fully understand the difference between them :sweat_smile: 

IIRC that was the type of the previous one I "reverted".

We can update it to type input if you prefer.

What is the difference between them? Also, is there an easy place to discover all the supported configurations and their possible values/meaning?

jsoriano: > What is the difference between them?

Integration packages are intended for complete solutions, they can contain multiple data streams, dashboards and so on, for some specific service. Examples: apache, nginx, aws...

Input packages are intended for generic inputs, they define a single stream, but they don't do much on their own. Examples: log, httpjson, sql...

In general input packages have better support for use cases where users might have multiple different customizations for policies using the same package.

I think the `filestream` package should be an input package. Some packages, like the `log` one were migrated at some point from integration to input package (see the PR for the log package: https://github.com/elastic/integrations/commit/4e3ca294d88b15e182fbba055507074f03a641e0).

Converting this package to an input package would consist on removing the `data_stream` level, and moving this information to the root level.

> Also, is there an easy place to discover all the supported configurations and their possible values/meaning?

Sadly Package Spec documentation is currently limited to some descriptions in the repository, that are not very easy to consume. Probably the best option at the moment is to look to other input packages in the integrations repository.

jsoriano: Migration to input package would be something like this: https://github.com/elastic/integrations/pull/12878

PR #12851, opened 2025-02-20
[teleport] Update event-groups ingest pipeline to manage cloud fields if already present

mrodm:
## Proposed commit message

Update event-groups ingest pipeline to manage cloud fields (like cloud.region, cloud.istance.id...) if they were already set by filebeat or other processors.

These errors were detected while testing the validation based on mappings: https://buildkite.com/elastic/integrations/builds/22313

But sometimes it was not detected, and it was due to `elastic-package` was not waiting to get all the docs defined in the test folder to be ingested. After being ingested a few documents, `elastic-package` started the validation of those docs found. 

This PR has two goals:
- adding the required configuration to change the behavior of the system tests to wait for the 270 documents defined in [`_dev/deploy/docker/sample_logs/`](https://github.com/elastic/integrations/tree/b6e4aaf381b6294733eede706902e74852999173/packages/teleport/_dev/deploy/docker/sample_logs).
- ~adding the `override` setting into the rename processors to avoid throwing exceptions.~
    - ~**IMPORTANT**: The corresponding cloud fields will be set with the value of the last rename processor.~
- adding the `if` setting into the rename processors to avoid throwing exceptions if those fields were already present in the document.
    - **IMPORTANT**: The corresponding cloud fields will keep the original values (if they exist), or the value from the first rename processor applicable.
    - Related discussion: https://github.com/elastic/integrations/pull/12851#pullrequestreview-2646676378

Errors found when `elastic-package` waits for all 270 documents ([buildkite link](https://buildkite.com/elastic/integrations/builds/22414) - [pr link](https://github.com/elastic/integrations/pull/12801)):
```
[0] found error.message in event: [Processor rename with tag  in pipeline logs-teleport.audit-1.2.1-event-groups failed with message: field [cloud.instance.id] already exists]
[1] found error.message in event: [Processor rename with tag  in pipeline logs-teleport.audit-1.2.1-event-groups failed with message: field [cloud.region] already exists]
```

In another test running just validation based on mappings, these were the fields undefined ([buildkite link](https://buildkite.com/elastic/integrations/builds/22313)):
```
[0] field "teleport.audit.account_id" is undefined: field definition not found
[1] field "teleport.audit.aws_host" is undefined: field definition not found
[2] field "teleport.audit.aws_region" is undefined: field definition not found
[3] field "teleport.audit.aws_service" is undefined: field definition not found
[4] field "teleport.audit.exit_code" is undefined: field definition not found
[5] field "teleport.audit.instance_id" is undefined: field definition not found
[6] field "teleport.audit.region" is undefined: field definition not found
[7] field "teleport.audit.status" is undefined: field definition not found
[8] field "teleport.audit.target" is undefined: field definition not found
```

Related discussions from previous PR:
- https://github.com/elastic/integrations/pull/12624/files#r1950970242
- https://github.com/elastic/integrations/pull/12624/files#r1951898376

## Checklist

## Author's Checklist

- [x] Test with `elastic-package` enabled mappings https://buildkite.com/elastic/integrations/builds/22587
- [x] Remove debug changes `.buildkite` folder and `go.mod`/`go.sum` files.

## Related issues

- Follows #12624


packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -1426,11 +1436,17 @@ processors:
...
       target_field: cloud.account.id
       ignore_missing: true
+      # This was failing due to `cloud.account.id` already existed
+      override: true # Should it be added an if condition? Should it be added a remove processor?
   - rename:
       field: teleport.audit.region
       target_field: cloud.region
       ignore_missing: true
-      ignore_failure: true
+      ignore_failure: true # it could already exist this field
```
mrodm, line 1445: `cloud.region` probably already set by `rename` processor in line 871
https://github.com/elastic/integrations/pull/12851/files#diff-d28d80fdbcd472b933ce90de9cf75954604df56b9cd6ddc158c644e87da59141R872

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -1407,6 +1415,8 @@ processors:
       field: teleport.audit.instance_id
       target_field: cloud.instance.id
       ignore_missing: true
+      # This was failing due to `cloud.instance.id` already existed
+      override: true # Should it be added an if condition? Should it be added a remove processor?
```
mrodm, line 1419: `cloud.instance.id` could be set in three different processors in this pipeline renaming:
- teleport.audit.aws_host
- teleport.audit.db_gcp_instance_id
- teleport.audit.instance_id

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -872,14 +872,20 @@ processors:
       field: teleport.audit.aws_region
       target_field: cloud.region
       ignore_missing: true
+      # This was failing due to `cloud.region` already existed
+      override: true # Should it be added an if condition? Should it be added a remove processor?
```
mrodm, line 876: `cloud.region` could be set in three different processors using different source fields:
- teleport.audit.aws_region
- teleport.audit.db_aws_region
- teleport.audit.region

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -1426,11 +1436,17 @@ processors:
       field: teleport.audit.account_id
       target_field: cloud.account.id
       ignore_missing: true
+      # This was failing due to `cloud.account.id` already existed
+      override: true # Should it be added an if condition? Should it be added a remove processor?
```
mrodm, line 1440: There is no other processor updating `cloud.account.id`

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -1426,11 +1436,17 @@ processors:
...
   - rename:
       field: teleport.audit.region
       target_field: cloud.region
       ignore_missing: true
-      ignore_failure: true
+      ignore_failure: true # it could already exist this field
+  # in case it fails previous rename processor, remove the field (not defined in the package)
+  - remove:
+      field: teleport.audit.region
+      ignore_missing: true
```
mrodm, line 1449: As `cloud.region` could be set in 3 different processors, should it be set here also `override: true` as in the others ?

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -872,14 +872,20 @@ processors:
...
       target_field: cloud.region
       ignore_missing: true
+      # This was failing due to `cloud.region` already existed
+      override: true # Should it be added an if condition? Should it be added a remove processor?
   - rename:
       field: teleport.audit.aws_service
       target_field: cloud.service.name
       ignore_missing: true
+      # This was failing due to `cloud.service.name` already existed
+      override: true # Should it be added an if condition? Should it be added a remove processor?
```
mrodm, line 882: No other processor updating `cloud.service.name`

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -1407,6 +1417,8 @@ processors:
       field: teleport.audit.instance_id
       target_field: cloud.instance.id
       ignore_missing: true
+      # This could fail if `cloud.account.id` field already exists in the doc. Forced to override the value to avoid throw an exception.
+      override: true
```
mrodm, line 1421: `cloud.instance.id` could be set in three different processors using different source fields:
- teleport.audit.aws_host
- teleport.audit.db_gcp_instance_id
- teleport.audit.instance_id

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -1426,11 +1438,14 @@ processors:
...
       ignore_missing: true
+      # This could fail if `cloud.account.id` field already exists in the doc.  Forced to override the value to avoid throw an exception.
+      override: true
   - rename:
       field: teleport.audit.region
       target_field: cloud.region
       ignore_missing: true
-      ignore_failure: true
+      # This could fail if `cloud.region` field already exists in the doc. Forced to override the value to avoid throw an exception.
+      override: true
```
mrodm, line 1448: `cloud.region` could be set in three different processors using different source fields:
- teleport.audit.aws_region
- teleport.audit.db_aws_region
- teleport.audit.region

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -1426,11 +1438,14 @@ processors:
       field: teleport.audit.account_id
       target_field: cloud.account.id
       ignore_missing: true
+      # This could fail if `cloud.account.id` field already exists in the doc.  Forced to override the value to avoid throw an exception.
+      override: true
```
mrodm, line 1442: There is no other processor updating `cloud.account.id`.

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -872,14 +872,20 @@ processors:
...
       target_field: cloud.region
       ignore_missing: true
+      # This could fail if `cloud.region` already exists in the doc. Forced to override the value to avoid throw an exception.
+      override: true
   - rename:
       field: teleport.audit.aws_service
       target_field: cloud.service.name
       ignore_missing: true
+      # This could fail if `cloud.service.name` field already exists in the doc. Forced to override the value to avoid throw an exception.
+      override: true
```
mrodm, line 882: No other processor updating `cloud.service.name`.

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -872,14 +872,20 @@ processors:
...
   - rename:
       field: teleport.audit.aws_host
       target_field: cloud.instance.id
       ignore_missing: true
+      # This could fail if `cloud.instance.id` field already exists in the doc. Forced to override the value to avoid throw an exception.
+      override: true
```
mrodm, line 888: Adding `override: true` into these rename processors makes that the corresponding cloud field values will be set with the value of the latest rename processor executed.

Is that ok/expected?

go.mod
```diff
@@ -213,21 +213,23 @@ require (
...
+	k8s.io/kubectl v0.32.1 // indirect
 	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738 // indirect
 	sigs.k8s.io/json v0.0.0-20241010143419-9aa6b5e7a4b3 // indirect
 	sigs.k8s.io/kustomize/api v0.18.0 // indirect
 	sigs.k8s.io/kustomize/kyaml v0.18.1 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.2 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/elastic/elastic-package => github.com/elastic/elastic-package v0.109.2-0.20250220090015-bbd8bcadb505
```
mrodm, line 235: To be removed before merging.
The same for the changes in files under `.buildkite` folder.

These are required to test with the validation based on mappings.

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
efd6: Direct link to the discussion in the original PR: https://github.com/elastic/integrations/pull/12624#discussion_r1951898376.

mrodm: > If we enabled an approach like the solution that was implemented for the qualys case (user can specify what they are interested in), or a hard "use the data from the data source" approach, then those two fields cease to be an issue. The remaining fields, `cloud.region` and `cloud.instance.id` look like they would then just fall out (depending on how teleport handles the `teleport.audit.{region,instance_id}` cases (are they disjoint with the explicitly named fields, `teleport.audit.{aws,db_aws}` and `teleport.audit.{aws_host,db_gcp_instance_id}` respectively. Unfortunately the test cases don't give any indication about this and I cannot find any reference for it in the teleport documentation; my feeling is that the more generic case should be overridden by the more specific case, but if an override occurs, then an error should be written in and the pipeline be continued. In the case that the we see apparently inconsistent fields being set (e.g. `teleport.audit.aws_host`, `teleport.audit.db_gcp_instance_id`) this would just be an error. I'm not sure what the behaviour of `teleport.audit.aws_region` v. `teleport.audit.db_aws_region` should be; are these mutually exclusive?


So, I don't have neither the context to answer that question nor to provide a better solution for the errors found in this ingest pipeline :(.

According to what you're suggesting (or I understand about it), do you mean to add an `on_failure` action in each rename processor modified here?
However, how could it be distinguished that the value to be override is the one set by `filebeat` or the one set by a previous processor?

Could I get some help here to move forward? :pray:

andrewkroh (commented):
Today, we know that no overrides are happening because that would result in an error.

Instead of adding `override: true`, which could change the behavior of `cloud` fields, how about adding a guard that checks if the target field exists to prevent the error (e.g., `if: ctx.cloud?.x == null`)?

I think the ideal solution is the one that was discussed, where the user gets to make the choice (i.e., what was added to Qualys). If we can't get this solution in time to unblock @mrodm's work on enhancing elastic-package, then I propose we replace override: true with an if, and then open an issue for the ideal solution.

-----

_side-note_: We need to audit every system test that uses canned logs to find the ones that are missing `assert.hit_count`.

mrodm:
> Today, we know that no overrides are happening because that would result in an error.
> 
> Instead of adding `override: true`, which could change the behavior of `cloud` fields, how about adding a guard that checks if the target field exists to prevent the error (e.g., `if: ctx.cloud?.x == null`)?

Ok, I'll replace those `overrides` by `if` conditions, thanks!

As a note, there would be some kind of change of behavior due to the fact that until now that pipeline was interrupted with an error if there ware `cloud.*` fields defined.

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
```diff
@@ -1426,11 +1450,19 @@ processors:
...
+      ignore_missing: true
   - rename:
       field: teleport.audit.region
       target_field: cloud.region
       ignore_missing: true
       ignore_failure: true
+      if: ctx.cloud?.region == null
+  - remove:
+      field: teleport.audit.region
+      ignore_missing: true
```
mrodm, line 1465: When using `if` conditions, as the source fields is not renamed, it must be added a new `remove` processor for each one.

packages/teleport/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.2.2"
+  changes:
+    - description: Update event-groups pipeline to take into account if cloud fields were already set.
```
efd6, line 4:
```suggestion
    - description: Update event-groups pipeline to take into account of whether cloud fields were already set.
```

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
efd6: The answer to the question is unfortunately a question. The issue is one of whether we know what the user wants from the data set. In the case of Qualys it was _fairly_ clear that at least some users wanted the data source's values for these fields rather than the values that are added by the agent processor. ISTM that this is likely to be the case here too, but I don't know this for a fact.

mrodm:
@efd6 @andrewkroh could I get another review for this PR? :blush: 

Thanks in advance!!

packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/event-groups.yml
efd6: Could be in a single remove processor block, but I think this is less prone to future damage.

efd6 (approved):
Thanks

PR #12862, opened 2025-02-21
[Amazon MQ] Add ActiveMQ overview dashboard

agithomas:
- Enhancement

## Proposed commit message

Add ActiveMQ overview dashboard

## Checklist

## Author's Checklist

- [x] Dashboard guidelines
- [x] Integration testing

- `elastic-package build`
- `elastic-package stack up -v -d --services package-registry`

## Screenshots

Screenshot


daniela-elastic (approved):
Looks good. Couple of small comments:
1. For CPU we show round values but followed by two zeroes after the decimal point (2.00%, 3.00%). Could we just show 2%, 3% or there's a reason why we want to show the extra zeroes?
2. In the "Producer vs Consumer count" graph we only show a line for the consumers. Any chance we can also show producers? Also, since the graph shows Consumers on the left and Producers on the right we should probably rename it to "Consumer vs Producer count" (reverse the order so the name and the two y-axis align

shmsr (approved):

muthu-mps:
Correct me if I am wrong here, Single document will not hold all the broker, topic and queue metrics together? If yes, then the first set of values will be blank when the topic and queue filter is applied. Do you think we can remove broker, queue and topic filters and top level?

Screenshot 2025-02-24 at 11 23 35 AM

agithomas:
> Do you think we can remove broker, queue and topic filters and top level?

As you mentioned, the dashboard will be partially loaded. However, I think, getting insights into specific queue/topic metrics would become very important to configure the producer/consumer count optimising or to determine the queue latency. The drop down filter adds this convenience - having large number of topics and queues are not un-common.

PR #12865, opened 2025-02-21
Add custom configuration to nginx/metrics

### LINKED ISSUE #12746: [NGINX]: DNS Lookup of multiple ingress pods via headless service
### Integration Name

Nginx [nginx]

### Dataset Name

stubstatus

### Integration Version

1.24.1

### Agent Version

8.14.3

Debian GNU/Linux 11 (bullseye) - 6.1.0-0.deb11.21-cloud-amd64

### User Goal

I would like to use a headless service to scrape the stubstatus of all my ingress nginx instances individually.

### Existing Features

Currently I would need to add each host individually, or use a service, which load balances in a round-robin fashion.
Adding each host is no option, because the pod IPs will change from time to time.

### What did you see?

Data ingest is working fine for a single instance.

### Anything else?

If there is an easy option to instruct the agent to use a headless service and scrape each IP resolved via DNS, I am happy to create a PR for it.
If not, I think this is a general request for all integrations.

toms-place:
## Proposed commit message

This PR adds the possibility to add custom configuration to the nginx/metrics input.
This allows to add custom conditions to nginx metrics, in order to use `http://${kubernetes.pod.ip}` in the hosts field on fleet managed agents.

## Checklist

## Author's Checklist

- [ ] Discuss if this is also needed for logs

Add ingress nginx to a k8s cluster, deploy a fleet agent on an ingress nginx node and scrape the metrics directly from the pod.

## Related issues

With this PR the headless service is not needed anymore.

- Closes #12746


packages/nginx/data_stream/stubstatus/manifest.yml
```diff
@@ -33,6 +33,24 @@ streams:
...
         description: >
           Processors are used to reduce the number of fields in the exported event or to enhance the event with metadata. This executes in the agent before the events are shipped. See [Processors](https://www.elastic.co/guide/en/fleet/current/elastic-agent-processor-configuration.html) for details.
+      - name: custom
+        title: Custom configurations
+        description: >
+          Additional settings to be added to the configuration.
+
+          Example:
+
+              condition: ${kubernetes.labels.app} == 'ingress-nginx'
```
muthu-mps, line 43: - Why do we apply the condition only for stubstatus datastream and not for the access and error logs?
- Is this applicable for Nginx ingress controller and for Nginx on Kubernetes? Because this condition is already applied for the Nginx_Ingress_Controller integration.
https://github.com/elastic/integrations/blob/fc127eb4482d15531f2848711ca21538155e7956/packages/nginx_ingress_controller/data_stream/access/manifest.yml#L47

toms-place: Hi @muthu-mps :)
Thanks for reviewing!

The nginx_ingress_contoller package only supports logs. It currently does not support metrics/stubstatus.
That's why I adapted the nginx package.

Adding the condition only is also an option, but I thought having more control by just using a custom field is appreciated.

I also have an open bullet point to discuss, if we want to have this for acces/error logs as well in the nginx package as well. What do you think?

muthu-mps: @toms-place - As the logs can be fetched using the `nginx_ingress_controller` integration, we can adopt the condition for the metrics.

toms-place: So do you want me to change it to condition only with an empty string as default , instead of using a custom field?

muthu-mps: As we try to fetch data by applying condition, we could probably change this to condition instead of custom input without applying default variable.

@toms-place - Can we try applying the config and process the data using `stubstatus` datastream to make sure that it process the expected metrics.

toms-place: I will change it to condition then. :+1: 

I already tested it on my end:
In order for it to work I needed to adapt the ingress-nginx-controller config to allow `nginx-status-ipv4-whitelist: <HOST_CIDR>|<POD_CIDR>`. The CIDR depends on the fleet agents network configuration (`hostNetwork: true|false`).
See: [https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#nginx-status-ipv4-whitelist](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#nginx-status-ipv4-whitelist) for reference.

toms-place: I now changed it to condition ;)

toms-place: @muthu-mps can this be merged now? :)

toms-place:
Please reopen it.

packages/nginx/data_stream/stubstatus/manifest.yml
muthu-mps: @toms-place - Sorry for the delay here, I would like to understand if this is going to be a redundant feature when we enable this data stream to `nginx-ingress-controller` integration. Will it be beneficial either way when used in both the integrations? Can we add the metrics data stream to the ingress-controller and adopt this change?  WDYT?

toms-place: @muthu-mps - Ah I think the confusion here is that I am talking about metrics. the `nginx-ingress-controller` integration only handles logs atm.

toms-place: But it could be beneficial there as well I guess.

ishleenk17:
@toms-place : Would you like to resume it ? I would take it to completion.

toms-place:
> @toms-place : Would you like to resume it ? I would take it to completion. 

@ishleenk17 - Yes, what is still left here?

ishleenk17:
> > @toms-place : Would you like to resume it ? I would take it to completion.
> 
> @ishleenk17 - Yes, what is still left here?

Some merge conflict resolutions. Then we should be GTG.

toms-place:
> > > @toms-place : Would you like to resume it ? I would take it to completion.
> > 
> > 
> > @ishleenk17 - Yes, what is still left here?
> 
> Some merge conflict resolutions. Then we should be GTG.

All resolved, please review again :)

PR #12878, opened 2025-02-24
Migrate filestream package to input type

jsoriano:
## Proposed commit message

Migrate filestream package to input package.

The kind of use cases that the filestream package support are more suited to generic use cases supported by the input package type.

## Checklist

- [x] ~~I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices)~~

## Author's Checklist

- [ ] 

## Related issues

-


jsoriano:
There should not be any issue if we keep the same variables. With the migration from the log integration to the log input we had some issues, but they were related to invalid datasets, that should not be allowed in recent versions of Kibana.

It'd be great if you can check if the new package works for you in any case. Thanks!

jsoriano:
Updated and opening for review. Happy to close it if there were no plans to migrate this package to input package, though I think it would be good to do so.

belimawr:
> Updated and opening for review. Happy to close it if there were no plans to migrate this package to input package, though I think it would be good to do so.

@jsoriano I see no issue moving forward with this. The PR looks good, I just want to do a quick test before approving it.

belimawr (changes requested):
@jsoriano, I tested upgrading the policy from the current `1.1.0` to the `2.0.0` from this PR and, as it stands, the upgrade generates data duplication :/

In the current `1.1.0` the ID for the Filestream input is:
```
filestream-filestream.generic-0e551301-3916-4d99-81c6-406aa390859b
```

However, after the upgrade the ID changes to:
```
filestream-filestream.filestream-0e551301-3916-4d99-81c6-406aa390859b
```

The ID of the Filestream input is part of the key used to save the state from the files being ingested, if this ID changes, all files get re-ingested.

I'm not sure what caused this change. Do you know if this is caused by the change in the type?

jsoriano:
@belimawr good catch, thanks for testing this.

This seems quite inconvenient for migration of integration packages to input type.

It looks like this id is generated by Fleet based on the names of the package and the data stream. Input packages don't have data streams, so the data stream part is replaced by the package name. I haven't managed to control this id.
But I couldn't find where these ids are generated, so not sure.

@kpollich do you know where Fleet generates these input ids?

jsoriano:
> i don't think it's the right move to make this an input type package

Why? Generic data streams are the typical example for input packages.

Is it because of the problems upgrading? If built from scratch, would you also make this an integration package?

kpollich:
> @kpollich do you know where Fleet generates these input ids?

Sorry for the delay in response. I just took a look and tracked down where these input ID's are generated. There's a `storedPackagePolicyToAgentInputs` method here that contains this logic: 

https://github.com/elastic/kibana/blob/6a8deddadce4c3209d0fa1ca1252b69716fc50d0/x-pack/platform/plugins/shared/fleet/server/services/agent_policies/package_policies_to_agent_inputs.ts#L27-L95

belimawr:
> > @kpollich do you know where Fleet generates these input ids?
> 
> Sorry for the delay in response. I just took a look and tracked down where these input ID's are generated. There's a `storedPackagePolicyToAgentInputs` method here that contains this logic:
> 
> https://github.com/elastic/kibana/blob/6a8deddadce4c3209d0fa1ca1252b69716fc50d0/x-pack/platform/plugins/shared/fleet/server/services/agent_policies/package_policies_to_agent_inputs.ts#L27-L95

No worries, I was on PTO and came back today ;)

I'm far from being an expert in TypeScript, but if I'm not mistaken, there is no way to change Filestream integration from package type to input type without affecting the ID. Is that it @kpollich ?

kpollich:
> I'm far from being an expert in TypeScript, but if I'm not mistaken, there is no way to change Filestream integration from package type to input type without affecting the ID. Is that it @kpollich ?

There should be a way to do this, but I think the block of code I linked would need to be updated to consider the move from input -> integration type as a special case.

belimawr:
> > I'm far from being an expert in TypeScript, but if I'm not mistaken, there is no way to change Filestream integration from package type to input type without affecting the ID. Is that it @kpollich ?
> 
> There should be a way to do this, but I think the block of code I linked would need to be updated to consider the move from input -> integration type as a special case.

Thanks @kpollich! 

Do you think this is something we should invest time on? Integrations are not my area of expertise, so I can't judge that. We just need to be careful to make sure those IDs are stable across upgrades/downgrades.

belimawr:
@kpollich, I've been thinking about the ID change problem. The Filestream input supports taking over the state of another instance of Filestream (coming up in the next release, see the [`config`](https://github.com/elastic/beats/blob/7bfa78e27177ef1ef8d3e52254c941eb932d9542/filebeat/filebeat.reference.yml#L888-L893) from `main`).

We'd need access to the previous ID, the configuration looks like this:
```yaml
  take_over:
   enabled: true
   from_ids: ["old-id-1", "old-id-2"]
```

Are the parts of the input ID available in any form that can be used in the input template?

nimarezainia:
@jsoriano @belimawr what is the path forward with this one?

belimawr:
> @jsoriano @belimawr what is the path forward with this one?

From the data-plane side, we can deal with changing IDs since `v9.1.0` by using [`take_over`](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-filestream#filebeat-input-filestream-take-over) and providing the  previous Filestream IDs. Keep in mind that this will tie the change in input type with the Elastic-Agent version `>= v9.1.0`.

However, this requires access to the "old IDs" and that requires changes in Fleet. Another option is to change the package type and have Fleet still generate the same IDs. Either way this requires changes in Fleet.

So from my perspective, this is a prioritisation issue. I'm happy to discuss/dive deeper do decide on a technical solution, but that's need to be aligned with the data-plane and fleet priorities.

jsoriano:
As there are changes required in Fleet in any case, I would say that it would be better to guarantee that input ids don't change on upgrades of existing policies (https://github.com/elastic/kibana/issues/223256). This would avoid more unexpected situations caused by these id changes in the future, on any input.

Supporting `take_over` sounds good too, but would only work for this specific situation, of changing ids in the `filestream` input. Though maybe this is fine enough as this is probably the input most impacted by this issue.

kpollich:
With discussions ongoing about leaning into input packages for OTel integrations, I think we should prioritize https://github.com/elastic/kibana/issues/223256 as a compliment to that work.

belimawr:
> I would say that it would be better to guarantee that input ids don't change on upgrades of existing policies ([elastic/kibana#223256](https://github.com/elastic/kibana/issues/223256))

I agree, that's the best solution and covers all cases.

There are also other inputs that might experience the same problem (input ID tied to the state) if their ID changes.

packages/filestream/manifest.yml
```diff
@@ -2,25 +2,307 @@ format_version: 3.1.5
...
 title: Custom Logs (Filestream)
 description: Collect log data using filestream with Elastic Agent.
-type: integration
-version: 1.2.0
+type: input
+version: 2.0.0
 conditions:
   kibana:
-    version: "^8.15.0 || ^9.0.0"
+    version: "^9.2.0"
```
jsoriano, line 9: Changed kibana constraint to the version that contains the fix to keep state (https://github.com/elastic/kibana/issues/223256).

jsoriano:
@belimawr now that https://github.com/elastic/kibana/issues/223256 is merged, what do you think about going on with this PR?
I have updated the kibana constraint to `^9.2.0` so the package is only used on versions with the fix.

belimawr:
> @belimawr now that [elastic/kibana#223256](https://github.com/elastic/kibana/issues/223256) is merged, what do you think about going on with this PR? I have updated the kibana constraint to `^9.2.0` so the package is only used on versions with the fix.

It sound good. I'll test it today.

belimawr:
@jsoriano, I've just tested it using `elastic-package` and pulling a new `Kibana:9.2.0-SNAPTHOT`, and data was duplicated again, the stream ID changed
- from `filestream-filestream.generic-227b4766-7234-4536-aba8-acea8b7dcdd7`
- to `filestream-filestream.filestream-227b4766-7234-4536-aba8-acea8b7dcdd7`

I'm wondering if the Kibana version I used contains the change you mentioned.

The build date is `2025-09-14T11:26:59.337Z` and the version `6be1de873486a707485492bfc8b13555707e15fa`.

<details><summary>docker inspect output</summary>
<p>

```json
            "Labels": {
                "architecture": "x86_64",
                "build-date": "2025-08-20T13:12:41",
                "com.redhat.component": "ubi9-minimal-container",
                "com.redhat.license_terms": "https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI",
                "description": "Your window into the Elastic Stack.",
                "distribution-scope": "public",
                "io.buildah.version": "1.40.1",
                "io.k8s.description": "The Universal Base Image Minimal is a stripped down image that uses microdnf as a package manager. This base image is freely redistributable, but Red Hat only supports Red Hat technologies through subscriptions for Red Hat products. This image is maintained by Red Hat and updated regularly.",
                "io.k8s.display-name": "Red Hat Universal Base Image 9 Minimal",
                "io.openshift.expose-services": "",
                "io.openshift.tags": "minimal rhel9",
                "maintainer": "infra@elastic.co",
                "name": "Kibana",
                "org.label-schema.build-date": "2025-09-14T11:26:59.337Z",
                "org.label-schema.license": "Elastic License",
                "org.label-schema.name": "Kibana",
                "org.label-schema.schema-version": "1.0",
                "org.label-schema.url": "https://www.elastic.co/products/kibana",
                "org.label-schema.usage": "https://www.elastic.co/guide/en/kibana/reference/index.html",
                "org.label-schema.vcs-ref": "6be1de873486a707485492bfc8b13555707e15fa",
                "org.label-schema.vcs-url": "https://github.com/elastic/kibana",
                "org.label-schema.vendor": "Elastic",
                "org.label-schema.version": "9.2.0-SNAPSHOT",
                "org.opencontainers.image.created": "2025-09-14T11:26:59.337Z",
                "org.opencontainers.image.documentation": "https://www.elastic.co/guide/en/kibana/reference/index.html",
                "org.opencontainers.image.licenses": "Elastic License",
                "org.opencontainers.image.revision": "6be1de873486a707485492bfc8b13555707e15fa",
                "org.opencontainers.image.source": "https://github.com/elastic/kibana",
                "org.opencontainers.image.title": "Kibana",
                "org.opencontainers.image.url": "https://www.elastic.co/products/kibana",
                "org.opencontainers.image.vendor": "Elastic",
                "org.opencontainers.image.version": "9.2.0-SNAPSHOT",
                "release": "1",
                "summary": "Kibana",
                "url": "https://catalog.redhat.com/en/search?searchType=containers",
                "vcs-ref": "f4b088292653bbf5ca8188a5e59ffd06a8671d4b",
                "vcs-type": "git",
                "vendor": "Elastic",
                "version": "9.2.0-SNAPSHOT"
            },
```

</p>
</details> 


Do you know how I can ensure I'm running the correct Kibana version?

jsoriano:
> just tested it using `elastic-package` and pulling a new `Kibana:9.2.0-SNAPTHOT`, and data was duplicated again, the stream ID changed
> * from `filestream-filestream.generic-227b4766-7234-4536-aba8-acea8b7dcdd7`
> * to `filestream-filestream.filestream-227b4766-7234-4536-aba8-acea8b7dcdd7`

You are right, I can also reproduce this :thinking: 

> I'm wondering if the Kibana version I used contains the change you mentioned.
> 
> The build date is `2025-09-14T11:26:59.337Z` and the version `6be1de873486a707485492bfc8b13555707e15fa`.

In principle this commit includes the fix.

@nchaulet do you know what could be happening?

nchaulet:
> @nchaulet do you know what could be happening?

I was able to repro this locally, input id seems to stay during migration, but not stream ids, I am taking a look and will come with a PR.

jsoriano:
Issue with ids solved by https://github.com/elastic/kibana/pull/235170.

belimawr:
> Issue with ids solved by [elastic/kibana#235170](https://github.com/elastic/kibana/pull/235170).

Awesome!

Do you know if this commit is already in the latest snapshot? I'm not quite sure how to track down which commits are included in a Docker image from Kibana...

jsoriano:
> Do you know if this commit is already in the latest snapshot? I'm not quite sure how to track down which commits are included in a Docker image from Kibana...

It is in the latest snapshot built yesterday.

belimawr:
> > Do you know if this commit is already in the latest snapshot? I'm not quite sure how to track down which commits are included in a Docker image from Kibana...
> 
> It is in the latest snapshot built yesterday.

Thanks! I'll test soon.

belimawr:
I did one test and it worked :tada:, no data duplication.

Here are the input configuration before and after the upgrade:
**Before the upgrade:**
```yaml
    - data_stream:
        namespace: default
      id: filestream-filestream-b178a25f-661b-4365-b02a-7b4111a66455
      meta:
        package:
            name: filestream
            version: 1.3.0
      name: filestream-1
      package_policy_id: b178a25f-661b-4365-b02a-7b4111a66455
      revision: 1
      streams:
        - clean_inactive: -1
          data_stream:
            dataset: filestream.generic
            type: logs
          file_identity:
            fingerprint: null
          id: filestream-filestream.generic-b178a25f-661b-4365-b02a-7b4111a66455
          parsers: null
          paths:
            - /tmp/flog.log
          prospector:
            scanner:
                exclude_files:
                    - \.gz$
                fingerprint:
                    enabled: true
                    length: 1024
                    offset: 0
                recursive_glob: true
      type: filestream
      use_output: default
```
The stream ID is `filestream-filestream.generic-b178a25f-661b-4365-b02a-7b4111a66455`

**After the upgrade:**
```yaml
    - data_stream:
        namespace: default
      id: filestream-filestream-b178a25f-661b-4365-b02a-7b4111a66455
      meta:
        package:
            name: filestream
            version: 2.0.0
      name: filestream-1
      package_policy_id: b178a25f-661b-4365-b02a-7b4111a66455
      revision: 2
      streams:
        - clean_inactive: -1
          data_stream:
            dataset: filestream.generic
            elasticsearch:
                dynamic_dataset: true
                dynamic_namespace: true
            type: logs
          file_identity:
            fingerprint: null
          id: filestream-filestream.generic-b178a25f-661b-4365-b02a-7b4111a66455
          parsers: null
          paths:
            - /tmp/flog.log
          prospector:
            scanner:
                exclude_files:
                    - \.gz$
                fingerprint:
                    enabled: true
                    length: 1024
                    offset: 0
                recursive_glob: true
      type: filestream
      use_output: default
```
The stream ID is `filestream-filestream.generic-b178a25f-661b-4365-b02a-7b4111a66455`.

I'll add my review soon.

packages/filestream/manifest.yml
```diff
@@ -2,25 +2,333 @@ format_version: 3.1.5
...
+        title: Encoding
+        description: |
+          The file encoding to use for reading data that contains international characters. For a full list of valid encodings, see the [Documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream.html#_encoding_2)
+        required: false
+        show_user: false
+      - name: recursive_glob
+        type: bool
+        title: Recursive Glob
+        description: |
+          Enable expanding ** into recursive glob patterns. With this feature enabled, the rightmost ** in each path is expanded into a fixed number of glob patterns. For example: /foo/** expands to /foo, /foo/*, /foo/*/*, and so on. If enabled it expands a single ** into a 8-level deep * pattern.
```
belimawr: The correct format for bold text is `**text**` (no spaces after or before the `**`). There are quite a few places here that make the mistake of adding spaces, like here.

```suggestion
          Enable expanding **into recursive glob patterns. With this feature enabled, the rightmost** in each path is expanded into a fixed number of glob patterns. For example: /foo/** expands to /foo, /foo/*, /foo/*/*, and so on. If enabled it expands a single ** into a 8-level deep * pattern.
```

Could you fix this and the other ones as well?

belimawr (commented):
LGTM, but there are those formatting issues that would be nice to fix on this PR as we're bumping the version to 2.0.0.

packages/filestream/manifest.yml
jsoriano: Oh, it looks like I missed some fomat fixes between merges, good catch! I have copied these vars again from main, in https://github.com/elastic/integrations/pull/12878/commits/5b5fb26756bae6974c4a71e7418e60e4b286bccb.

jsoriano:
Thanks for testing!

jsoriano:
Waiting for serverless to receive the fix as severless doesn't check `kibana.version`.

jsoriano:
Tested with serverless, ids are preserved and data is not duplicated, merging.

PR #12896, opened 2025-02-25
[Falco] Split datastream based on CNCF or agent-based ingest type

mjwolf:
## Proposed commit message

In order to assist with collection of telemetry on the ingest method used with the Falco integration, this will set the datastream based on whether data was collected with the agent-based or CNCF native ingest method.

Since the CNCF-native ingest method writes directly to the datastream, there's less control over what clients set in documents. It's used as the default case, since there may be no fields that are guaranteed unique to the CNCF documents to route on. The agent-based ingest method will reliably set some fields that can be used to determine the agent was used to write the documents. Routing to the agent datastream is done when an agent-only field is set.
 
This is a breaking change, since it will change the datastream used for agent-based ingestion of Falco data.

## Checklist

To test, set up the falco integration using the elastic agent on the host. Verify that data being ingested into elasticsearch with `data_stream.dataset=falco.alerts_agent`, `event.dataset=falco.alerts_agent`, and in a `.ds-logs-falco.alerts-agent*` index. 

Also set up CNCF-native ingest, and check that the data is ingested into logs-falco.alerts-all

## Screenshots

**Agent-based ingest** documents redirected to `falco.alerts_agent` dataset and `.ds-logs-falco.alerts-agent*` index:

Screenshot 2025-02-18 at 2 25 45 PM
Screenshot 2025-02-18 at 2 26 10 PM

**CNCF-native ingest** documents use `falco.alerts` dataset and `.ds-logs-falco.alerts-all` index
Screenshot 2025-02-25 at 3 18 11 PM
Screenshot 2025-02-25 at 3 17 52 PM


mjwolf:
\cc @nick-alayil

efd6:
I think you need to sort out the routing some more:
```
test case failed: one or more errors found in document: [0] field "event.dataset" should have value in ["falco.alerts"], it has "falco.alerts_agent"
[1] field "data_stream.dataset" should have value in ["falco.alerts"], it has "falco.alerts_agent"
[2] parsing field value failed: field "event.dataset"'s value "falco.alerts_agent" does not match the declared constant_keyword value "falco.alerts"
[3] parsing field value failed: field "data_stream.dataset"'s value "falco.alerts_agent" does not match the declared constant_keyword value "falco.alerts"
```

packages/falco/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "2.0.0"
+  changes:
+    - description: |
+        Route documents to datastream based on ingest source.
```
andrewkroh, line 5:
```suggestion
        Route documents to data stream based on ingest source.
```

packages/falco/changelog.yml
```diff
@@ -1,4 +1,12 @@
...
+
+        Ingest with CNCF datasources will use 'falco-*' dataset, and agent-based ingestion will use 'falco-agent-*'.
```
andrewkroh, line 7:
```suggestion
        Ingest with CNCF data sources will use 'falco-*' dataset, and agent-based ingestion will use 'falco-agent-*'.
```

packages/falco/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "2.0.0"
+  changes:
+    - description: |
+        Route documents to datastream based on ingest source.
+
```
andrewkroh, line 6: I wonder how this will render on our documentation web pages. I'm not sure how much value the newline adds so perhaps just remove it.

packages/falco/data_stream/alerts/manifest.yml
```diff
@@ -1,5 +1,12 @@
 title: "Falco Alerts"
 type: logs
+dataset: falco.alerts
+elasticsearch:
+  dynamic_dataset: true
+  dynamic_namespace: true
+  index_template:
+    mappings:
+      dynamic: true
```
andrewkroh, line 9: I thought this was the default. Is specifying `dynamic: true` necessary?

packages/falco/data_stream/alerts/routing_rules.yml
```diff
@@ -0,0 +1,7 @@
+- source_dataset: falco.alerts
+  rules:
+    - target_dataset: falco.alerts_agent
+      if: 'ctx?.elastic_agent?.id != null'
```
andrewkroh, line 4:
```suggestion
      if: ctx.elastic_agent?.id != null
```

packages/falco/data_stream/alerts_agent/manifest.yml
```diff
@@ -0,0 +1,9 @@
+title: Falco Alerts Agent
+dataset: falco.alerts_agent
+type: logs
+elasticsearch:
+  dynamic_dataset: true
+  dynamic_namespace: true
+  index_template:
+    mappings:
+      dynamic: true
```
andrewkroh, line 9: Same question, is this necessary?

taylor-swanson:
Regarding the test failures, I believe this is caused by the system test running in the `alerts` data stream, but the log(s) get rerouted to the `alerts_agent` data stream. The system test doesn't know what to do with re-routed documents (I ran into a simliar issue with the syslog router integration). I thought there was some way to configure the system test to better handle this (as in, giving it the data stream to validate against), but I can't seem to find that in package spec.

It may not be possible yet to write a system test for this. There's an issue somewhere for adding support for this scenario to elastic-package, but I'd have to dig up that issue.

mjwolf:
I've disabled system tests, and fixed the static tests. I don't think there's any way to have passing system tests with rerouted datasets, until elastic-package adds support for this, as is being tracked by https://github.com/elastic/elastic-package/issues/1917

packages/falco/data_stream/alerts/_dev/test/system/test-default-config.yml
```diff
@@ -1,13 +0,0 @@
-service: falco-tcp
```
kcreddy, line 1: Looking at other examples that define `routing_rules.yml`, the system tests are present in that datastream.
Example: [`sentinel_one_cloud_funnel`](https://github.com/elastic/integrations/blob/main/packages/sentinel_one_cloud_funnel/data_stream/event/routing_rules.yml), [`entityanalytics_okta`](https://github.com/elastic/integrations/blob/main/packages/entityanalytics_okta/data_stream/entity/routing_rules.yml). 
Is there any limitation for removing system tests?

taylor-swanson (approved):

packages/falco/data_stream/alerts/_dev/test/system/test-default-config.yml
mjwolf: I'm not too familar with those integrations, but I think the reason system tests work with those is because there's at least one event in the system tests that remains in the original dataset, so the system tests will see at least one event, and pass.

With this, we want all events that come from the agent to be redirected, so there isn't a way we can have any events remain in the original dataset, since all events from system tests come from the agent.

I've added back the system tests though, so at least others can see that the tests should exist, and set them to skip, with a link to the elastic-package issue tracking this problem

kcreddy: Thanks for the context.

kcreddy (approved):
LGTM once the CI error is fixed.

`link` is required param for [skip config](https://github.com/elastic/package-spec/blob/main/spec/integration/data_stream/_dev/test/skip.spec.yml#L19)

packages/falco/data_stream/alerts/_dev/test/system/test-default-config.yml
kcreddy: System test skip configuration requires `link` as well and CI error is due to that.

PR #12906, opened 2025-02-26
Agentless deployment of `ess_billing` integration

3kt:
Enables agentless deployment mode for the `ess_billing` integration.

### DIFF
--- a/packages/ess_billing/_dev/build/docs/README.md
+++ b/packages/ess_billing/_dev/build/docs/README.md
@@ -29,7 +29,7 @@ For private cloud, or admin users, the cloud endpoint can be altered to match yo
 For step-by-step instructions on how to set up an integration, see the
 [Getting started](https://www.elastic.co/guide/en/starting-with-the-elasticsearch-platform-and-its-solutions/current/getting-started-observability.html) guide.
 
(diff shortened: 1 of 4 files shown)


pathoge (approved):
LGTM

qcorporation:
@3kt just an FYI: agentless backend infrastructure isn't available until 8.17 but truthfully we should be targeting 8.18 stack.
I'm not sure what will happen if you keep the kibana version stack `version: "^8.15.0"`

3kt:
> @3kt just an FYI: agentless backend infrastructure isn't available until 8.17 but truthfully we should be targeting 8.18 / 9.0 stack.
> 
> I'm not sure what will happen if you keep the kibana version stack `version: "^8.15.0"`

Interesting - are we considering "split" version condition ?
Said differently, I'd like the default deployment to have the native 8.15 considtion, while agentless only becomes available if we're 8.17+

kpollich:
> Said differently, I'd like the default deployment to have the native 8.15 considtion, while agentless only becomes available if we're 8.17+

I _think_ this will work as expected as none of the agentless functionality is actually available in the UI pre 8.17.0, but I haven't actually tested this exact version constraint behavior.

3kt:
@kpollich well let's guinea-pig this theory ;)

Issue #12873, opened 2025-02-24
Event Loss with Trend Micro Vision One Integration "detections" data stream


ShourieG:
The trend micro vision one integration is based on the httpjson input. The **list detections** api documented [here](https://automation.trendmicro.com/xdr/api-v3/#tag/Search/paths/~1v3.0~1search~1detections/get), seems to return events with a hardcoded sorting order of **"descending"** type based on the **eventTimeDT** value of the events. This cannot be manipulated using query params. (Confirmed via a customer who directly spoke with trendmicro )

Explanation of the scenario- 

Suppose there are 100 events in total, and we query the api with a value of  **top=25**, this means for 100 events there will be a total of 4 pages.

During 1st pagination httpjson does the following:
- split body.items (contains 25 events in descending order in the items[] array) items[0] -> latest event for entire pagination cycle of 100 events
- publishes the 25 events
- updates the cursor value to .first_event.eventTimeDT,  i.e items[0].eventTimeDT

Now lets suppose the agent is restarted due to some reason, our cursor contains **eventTimeDT** of the latest event, i.e the 100th event.

When next request cycle starts we resume the request with:

```yml
- set:
      target: url.params.startDateTime
      value: '[[formatDate ((parseDate .cursor.last_update_at).Add (parseDuration "-{{additional_look_back}}"))]]'
``` 

but this logic is flawed since this timestamp leads to us dropping/not ingesting the events [75-0], since their **eventTimeDT** is **earlier** than our current cursor value, and this leads to data loss in the long run. We do not have control of when the **.first_event.eventTimeDT** is saved in httpjson as it's configured to do so during each pagination cycle. If the events were ordered in ascending order, this would not have been an issue since **.last_event.eventTimeDT** would always contain the latest event and our cursor would be consistent with our request cycles.

**The only way I see mitigating this is creating a CEL equivalent where we say, only save the cursor after the entire pagination cycle has been completed.** 

Would like more thoughts on this @andrewkroh, @efd6 , @chrisberkhout

andrewkroh:
The Google Workspace API has similar behavior and resulted in a similar issue.See https://github.com/elastic/integrations/issues/4796. The integration was modified to accommodate the API behavior.

ShourieG:
@andrewkroh, thanks for bringing this up, I saw the process used here and was wondering if it would be simpler to just convert it to CEL instead of using the more complex templating approach used in httpjson in this scenario.

andrewkroh:
[...]

PR #12916, opened 2025-02-27
[Trendmicro Vision One] - Fixed cursor & pagination logic for "detections" data stream

ShourieG:
## Type of change
- Bug
- Breaking Change

## Proposed commit message
 Fixed cursor & pagination logic for "detections" data stream to mirror a similar logic in implemented for a  [google_workspace fix](https://github.com/elastic/beats/pull/34274).
 
Please refer to [this](https://github.com/elastic/integrations/issues/12873) issue to see the details.

## NOTE
This will be a breaking change for the detections data stream as cursor values are completely changed, so existing state will get affected.

## Checklist

## Author's Checklist

- [ ] 

## Related issues
- Closes https://github.com/elastic/integrations/issues/12873
- Relates https://github.com/elastic/beats/pull/34274


ShourieG:
#TODO - test this with a live account

packages/trend_micro_vision_one/data_stream/detection/agent/stream/httpjson.yml.hbs
```diff
@@ -27,29 +27,49 @@ request.transforms:
       value: '{{page_size}}'
   - set:
       target: url.params.startDateTime
-      value: '[[formatDate ((parseDate .cursor.last_update_at).Add (parseDuration "-{{additional_look_back}}"))]]'
+      # If last pagination cycle was finished successfully
+      # we move the startDate pointer forward
+      # else we reprocess the last interval from the beginning.
+      # If none of the values are in the cursor it means is a fresh start
```
efd6, line 33:
```suggestion
      # If none of the values are in the cursor it means it is a fresh start
```

packages/trend_micro_vision_one/data_stream/detection/agent/stream/httpjson.yml.hbs
```diff
@@ -27,29 +27,49 @@ request.transforms:
       value: '{{page_size}}'
   - set:
       target: url.params.startDateTime
-      value: '[[formatDate ((parseDate .cursor.last_update_at).Add (parseDuration "-{{additional_look_back}}"))]]'
+      # If last pagination cycle was finished successfully
```
efd6, line 30:
```suggestion
      # If the last pagination cycle was completed successfully
```

packages/trend_micro_vision_one/data_stream/detection/agent/stream/httpjson.yml.hbs
```diff
@@ -27,29 +27,49 @@ request.transforms:
...
+      # we move the startDate pointer forward
+      # else we reprocess the last interval from the beginning.
```
efd6, line 32:
```suggestion
      # otherwise we reprocess the last interval from the beginning.
```

packages/trend_micro_vision_one/data_stream/detection/agent/stream/httpjson.yml.hbs
```diff
@@ -27,29 +27,49 @@ request.transforms:
...
+      # If none of the values are in the cursor it means is a fresh start
+      # and we pick the default value.
+      value: >-
+        [[- if eq .cursor.pagination_finished "true" -]]
+        # look-back time should only be applied for fresh pagination intervals
```
efd6, line 37: If you want to include comments in the Go template, you will need to use the template comment syntax.
```suggestion
        [[- /* look-back time should only be applied for fresh pagination intervals */ -]]
```
What is the precise definition of "fresh" here? I can see two possible interpretations; one matches the behaviour here (not started but after one that has completed), the other doesn't (never started). The immediately obvious interpretation is the second one, so I think this needs to be clearer.

efd6 (commented):
This seems reasonable, but needs some commentary clean ups.

packages/trend_micro_vision_one/data_stream/detection/agent/stream/httpjson.yml.hbs
ShourieG: By fresh pagination cycle I meant a completely new pagination cycle that occurs on every new interval not between pages during a particular interval, I'll make it a bit clearer

ShourieG:
@efd6, addressed all the comments

packages/trend_micro_vision_one/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.25.2"
+  changes:
+    - description: Added pagination and cursor fix to mirror the logic for a similar google_workspace scenario in pull - https://github.com/elastic/beats/pull/34274.
```
efd6, line 4:
```suggestion
    - description: Fixed pagination and cursor handling.
```
(users who want to know about pull requests will know how to find them, presenting them to other people is not really helpful, so instead of putting it here, put it in the commit message or just rely on the discussion in this PR — sorry, I missed this on the first round)

efd6 (approved):
Pending whether [this TODO](https://github.com/elastic/integrations/pull/12916#issuecomment-2687294752) was TODONE.

ShourieG:
cc: @narph @jamiehynds, planning on merging this soon but just as a fyi, this will be a breaking change for the detection data stream as the existing state will be lost due to a complete overhaul of the cursor logic. Should we give any other warnings for our customers before merging ?

narph:
> cc: @narph @jamiehynds, planning on merging this soon but just as a fyi, this will be a breaking change for the detection data stream as the existing state will be lost due to a complete overhaul of the cursor logic. Should we give any other warnings for our customers before merging ?

@ShourieG lets not merge this till we discuss the impact on our users/customers

ShourieG:
@jamiehynds, So if we implement this, the detections data stream will lose all existing state, the reingestion process will depend on the customers configuration as to how far they want to look back. If the customer does not change any configs and let things stay the way they are, the ingestion would start with the default startDateTime value which is calculated on the basis of (now - initial_interval) where initial_interval is user configured.

If we don't implement this, we would end up in a similar state with what happened with google_workspace a couple of years back when the integration was dropping quite a bit of data due to how the api returned the responses. Basically we would get users complaining of data inconsistency which can vary from a small amount to quite large amounts on a case by case basis.

jamiehynds:
@ShourieG while not ideal, I think implementing this change now is the less of two evils. Most users will likely prefer to deal with some reingestion than run the risk of dropping data due to current cursor logic. Plus, our default is 24h initial interval, and it's 'just' detections, so don't expect it do be a vast amount of data. We just need to make it clear in the changelog description that there's a breaking change included in this update and be clear on the impact (reingestion)

PR #12924, opened 2025-02-28
[AmazonMQ] Add RabbitMQ metrics dataset

agithomas:
- Enhancement

## Proposed commit message

- Add support for RabbitMQ metrics.
- Add minor documentation updates.

## Checklist

## Author's Checklist

- [x] Integration testing with producer clients and consumer clients
- [x] Comparison of values with RabbitMQ console metrics
- [x] Metrics dependencies on dimensions and adjustments
- [x] TSDS enablement testing

- `elastic-package build`
- `elastic-package stack up -v -d --services package-registry`

## Screenshots

image

### DIFF
--- a/packages/aws_mq/_dev/build/docs/README.md
+++ b/packages/aws_mq/_dev/build/docs/README.md
@@ -11,15 +11,16 @@ The Amazon MQ integration allows you to efficiently collect and monitor broker p
 
 ## Compatibility
 
-This integration presently supports Amazon MQ for [Apache ActiveMQ](http://activemq.apache.org/) metrics.
+This integration presently supports Amazon MQ for [Apache ActiveMQ](http://activemq.apache.org/) and [RabbitMQ](https://www.rabbitmq.com/) metrics.
 
 ## Data streams
 
-The Amazon MQ integration collects Apache ActiveMQ metrics. 
+The Amazon MQ integration collects Apache ActiveMQ and RabbitMQ metrics. 
 
 
 Data streams:
  - `activemq_metrics`: Collects broker metrics and destination (queue and topic) metrics.
+ - `rabbitmq_metrics`: Collects broker, queue and node metrics.
 
 
 ## Requirements
@@ -48,9 +49,9 @@ documentation](https://docs.elastic.co/integrations/aws#requirements).
 
 ## Metrics
 
-### ActiveMQ Metrics
+### ActiveMQ metrics
 
-Amazon MQ provides a range of broker and queue metrics that help monitor system performance, resource utilization, and message flow. These metrics can be used for various use cases, including:
+AmazonMQ for ActiveMQ provides a range of broker and queue metrics that help monitor system performance, resource utilization, and message flow. These metrics can be used for various use cases, including:
 
 - Tracking broker resource utilization, such as compute, memory, and storage.
 - Monitoring message throughput and queue performance.
@@ -66,3 +67,15 @@ The following metrics are related to [Amazon MQ quotas](https://docs.aws.amazon.
 
 {{event "activemq_metrics"}}
 {{fields "activemq_metrics"}}
+
+
+### RabbitMQ metrics
+
+Amazon MQ for RabbitMQ offers a variety of broker and queue metrics to monitor system performance, resource utilization, and message flow. These metrics are essential for:
+
+- Assessing broker resource usage, including CPU, memory, and storage.
+- Tracking message rates and queue depths to ensure efficient message processing.
+- Analyzing connection counts and consumer activity to optimize messaging workloads.
+
+{{event "rabbitmq_metrics"}}
+{{fields "rabbitmq_metrics"}}
--- a/packages/aws_mq/_dev/build/docs/README.md
+++ b/packages/aws_mq/_dev/build/docs/README.md
@@ -51,7 +51,7 @@ documentation](https://docs.elastic.co/integrations/aws#requirements).
 
 ### ActiveMQ metrics
 
-AmazonMQ for ActiveMQ provides a range of broker and queue metrics that help monitor system performance, resource utilization, and message flow. These metrics can be used for various use cases, including:
+Amazon MQ for ActiveMQ provides a range of broker and queue metrics that help monitor system performance, resource utilization, and message flow. These metrics can be used for various use cases, including:
 
 - Tracking broker resource utilization, such as compute, memory, and storage.
 - Monitoring message throughput and queue performance.
--- a/packages/aws_mq/data_stream/rabbitmq_metrics/agent/stream/stream.yml.hbs
+++ b/packages/aws_mq/data_stream/rabbitmq_metrics/agent/stream/stream.yml.hbs
@@ -1,0 +1,74 @@
+metricsets: ["cloudwatch"]
+period: {{period}}
+{{#if data_granularity}}
+data_granularity: {{data_granularity}}
+{{/if}}
+{{#if include_linked_accounts}}
+include_linked_accounts: {{include_linked_accounts}}
+{{#if owning_account}}
+owning_account: "{{owning_account}}"
+{{/if}}
+{{/if}}
+{{#if access_key_id}}
+access_key_id: {{access_key_id}}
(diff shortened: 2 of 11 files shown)


packages/aws_mq/data_stream/rabbitmq_metrics/agent/stream/stream.yml.hbs
```diff
@@ -0,0 +1,74 @@
...
+proxy_url: {{proxy_url}}
+{{/if}}
+{{#if processors}}
+processors:
+{{processors}}
+{{/if}}
+metrics:
+- namespace: AWS/AmazonMQ
+  statistic: ["Maximum"]
+  name:
```
gpop63, line 55: I wonder if `Average` statistic would provide some value for some of these metrics, `Maximum` would show the peak values which usually are momentary but the normal trend could be different.

Would `Average` be better for these metrics: `RabbitMQIOReadAverageTime`, `RabbitMQIOWriteAverageTime`, `RabbitMQMemLimit`, `RabbitMQDiskFreeLimit` WDYT?

agithomas: Here is what i observed as part of testing `RabbitMQIOReadAverageTime` and `RabbitMQIOWriteAverageTime` itself an average value. 

I also compared against the value appearing in the rabbitmq's web console

image and `max` aggregation seem to match the value in the web-console. 

However, there is a change i just made, `RabbitMQDiskFree.max` is updated as `RabbitMQDiskFree.min`

packages/aws_mq/_dev/build/docs/README.md
```diff
@@ -48,9 +49,9 @@ documentation](https://docs.elastic.co/integrations/aws#requirements).
 
 ## Metrics
 
-### ActiveMQ Metrics
+### ActiveMQ metrics
 
-Amazon MQ provides a range of broker and queue metrics that help monitor system performance, resource utilization, and message flow. These metrics can be used for various use cases, including:
+AmazonMQ for ActiveMQ provides a range of broker and queue metrics that help monitor system performance, resource utilization, and message flow. These metrics can be used for various use cases, including:
```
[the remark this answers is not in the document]
agithomas: Good catch. Thanks. I've updated the README

packages/aws_mq/data_stream/rabbitmq_metrics/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,45 @@
...
+processors:
+  - dot_expander:
+      field: "*"
+      ignore_failure: true
+  - drop:
+      description: "To drop the documents having RabbitMQ metrics"
+      if: >
+        ctx.aws?.amazonmq?.metrics?.size() == 1 &&
+        ctx.aws.amazonmq.metrics.ConsumerCount?.max != null
+      ignore_failure: true
```
gpop63, line 12: Is this because AWS sometimes sends this metric in isolation and its not very useful?

agithomas: Yes, it is to isolate from the ActiveMQ metrics

gpop63 (approved):

packages/aws_mq/data_stream/rabbitmq_metrics/fields/fields.yml
```diff
@@ -0,0 +1,176 @@
...
+              metric_type: gauge
+              type: long
+              unit: byte
+              description: The disk limit for a RabbitMQ node.
+            - name: RabbitMQDiskFree.min
+              metric_type: gauge
+              type: long
+              unit: byte
+              description: The total volume of free disk space available in a RabbitMQ node.
+            - name: RabbitMQFdUsed.max
```
devamanv, line 143: From the [docs](https://docs.aws.amazon.com/amazon-mq/latest/developer-guide/rabbitmq-logging-monitoring.html), this metrics looks to be a count. Should the name be changed to `RabbitMQFdUsed.count` instead to reflect the same? 
Consider removing the `unit: byte` if we happen to make the change.

packages/aws_mq/data_stream/rabbitmq_metrics/manifest.yml
```diff
@@ -0,0 +1,67 @@
...
+          # - key: "created-by"
+            # value: "foo"
+      - name: include_linked_accounts
+        type: bool
+        title: Include Linked Accounts
+        multi: false
+        required: false
+        show_user: false
+        default: true
+        description: When include_linked_accounts is set to true, CloudWatch metrics will be collected from both linked accounts and the monitoring account. Default is true.
```
devamanv, line 50:
```suggestion
        description: If `include_linked_accounts` is set to true, CloudWatch metrics are collected from both the linked accounts and the monitoring account. Default: true.
```

packages/aws_mq/data_stream/rabbitmq_metrics/manifest.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: owning_account
+        type: integer
+        title: Owning Account
+        multi: false
+        required: false
+        show_user: false
+        description: Accepts an AWS account ID linked to the monitoring account. Works only if include_linked_accounts is set to true. If set, monitoring data will only include data from the given account.
```
devamanv, line 57:
```suggestion
        description: Accepts an AWS account ID linked to the monitoring account. This option is applicable only if 'include_linked_accounts' is set to true. If provided, monitoring data will be limited to the specified account.```

File: packages/aws_mq/data_stream/rabbitmq_metrics/manifest.yml [hunk 7]
```diff
@@ -0,0 +1,67 @@
...
+        multi: true
+        required: false
+        show_user: true
+      - name: latency
+        type: text
+        title: Latency
+        multi: false
+        required: false
+        show_user: false
+        description: The 'latency' parameter adjusts the Metricbeat collection start and end times. AWS CloudWatch might experience delay in processing metrics for some services causing data points to be missed during the integration collection period. To mitigate this potential issue, specify a latency parameter such as `15m`.
```
devamanv, line 33:
```suggestion
        description: The 'latency' parameter adjusts the Metricbeat collection start and end times. AWS CloudWatch may experience delays in processing metrics for some services, which can cause data points to be missed during the integration collection period. To mitigate this issue, specify a latency parameter (e.g., 15m).
```

packages/aws_mq/data_stream/rabbitmq_metrics/fields/fields.yml
```diff
@@ -0,0 +1,176 @@
...
+          description: The name of the queue.
+          type: keyword
+          dimension: true
+        - name: VirtualHost
+          description: The name of the virtual host.
+          type: keyword
+          dimension: true
+    - name: cloudwatch.namespace
+      type: keyword
+      description: The namespace specified when query cloudwatch api.
```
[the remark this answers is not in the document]
agithomas: I agree. It is written to align with the field description with other aws datasets. A general update can be taken up across all aws packages in future.

packages/aws_mq/data_stream/rabbitmq_metrics/fields/fields.yml
agithomas: This approach is taken for aws metrics dataset to support multiple statistical aggregation type

packages/aws_mq/data_stream/rabbitmq_metrics/manifest.yml
agithomas: I agree. It is written to align with the field description with other aws datasets. A general update can be taken up across all aws packages in future.

packages/aws_mq/data_stream/rabbitmq_metrics/manifest.yml
agithomas: I agree. It is written to align with the field description with other aws datasets. A general update can be taken up across all aws packages in future.

PR #12770, opened 2025-02-13
mimecast.audit_events: Prevent pageToken from incorrectly reappearing in interval requests

kcreddy:
## Proposed commit message
See title

> [!NOTE]
> To reviewers:
> - Issue is described in detail in a similar fix: https://github.com/elastic/integrations/pull/12666
> - Also see test 1 below to know how pagetoken (in current version) is wrongly getting carried into next request and how PR fixes it.

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally

1. Final `response state` from system tests in current vs updated (PR):
  - Current:

    ```json
        {
        ...
        "cursor": {
              "last": "2025-02-13T15:32:15Z"
          },
        "events": [],
        "last_page": {
              "data": [
                  {
                      "endDateTime": "2025-02-13T15:32:13Z",
                      "startDateTime": "2025-02-12T15:32:13Z"
                  }
              ],
              "next": "lasttoken"
          },
        ...
        }
    ``` 

  - Updated (PR):

    ```json
        {
        ...
        "cursor": {
              "last": "2025-02-13T15:17:25Z"
          },
        "events": [],
        "last_page": {},
        ...
        }
    ``` 

2. System tests should pass.

```
2025/02/13 21:02:15 DEBUG found 0 hits in logs-mimecast.audit_events-87315 data stream
2025/02/13 21:02:17 DEBUG found 0 hits in logs-mimecast.audit_events-87315 data stream
2025/02/13 21:02:18 DEBUG found 4 hits in logs-mimecast.audit_events-87315 data stream
--- Test results for package: mimecast - START ---
╭──────────┬──────────────┬───────────┬───────────┬────────┬───────────────╮
│ PACKAGE  │ DATA STREAM  │ TEST TYPE │ TEST NAME │ RESULT │  TIME ELAPSED │
[...]

PR #12936, opened 2025-03-03
mimecast: Prevent pageToken from incorrectly reappearing in interval requests in multiple datastreams

kcreddy:
## Proposed commit message
```
last_page is stored at the root-level of state. As it is enclosed 
in state.with(), this leads to last_page.next incorrectly getting 
carried into subsequent interval requests whenever last_page 
evaluates to optional.none().

This change removes last_page assignment to optional.none(), 
thus removing last_page.next it from interval requests. 
```

> [!NOTE]
> This fix is similar to audit events in #12770.

## Checklist

### DIFF
--- a/packages/mimecast/data_stream/archive_search_logs/agent/stream/cel.yml.hbs
+++ b/packages/mimecast/data_stream/archive_search_logs/agent/stream/cel.yml.hbs
@@ -90,7 +90,7 @@ program: |
           }
         }).do_request().as(resp, resp.StatusCode == 200 ?
           bytes(resp.Body).decode_json().as(body, body.?fail.orValue([]).size() == 0 ?
-            {
+            (has(body.?meta.pagination.next) && size(body.data) != 0).as(want_more, {
               "events": body.data.map(e, e[state.data_path].map(l, {"message": l.encode_json()})).flatten(),
               "cursor": {
                 "last": (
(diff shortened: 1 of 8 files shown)


packages/mimecast/data_stream/archive_search_logs/agent/stream/cel.yml.hbs
```diff
@@ -106,19 +106,16 @@ program: |
...
-                optional.of({
-                  ?"next": body.?meta.pagination.next,
-                  "data": req.data,
-                })
-              :
-                optional.none(),
+              "last_page": {
+                ?"next": want_more ? body.?meta.pagination.next : optional.none(),
+                ?"data": want_more ? req.?data : optional.none(),
+              },
```
chrisberkhout, line 112: I don't really like the pattern of wrapping everything in `state.with()`. I think it's better to be able to look at a certain branch of the expression and see what is returned there, without having to remember that things may be added far away.

This seems like a good example of that. I had to read the PR description a couple of time to understand that this wasn't a no-op change.

In any case, if we stick with our current convention of wrapping it, I do think it's good to have a comment so the override doesn't get removed accidentally.

Maybe something like this?

```suggestion
              "last_page": want_more ?
                {
                  ?"next": body.?meta.pagination.next,
                  ?"data": req.?data,
                }
              :
                null, // required to clear the incoming state
```

kcreddy: I'm okay with the suggestion. Thanks Chris

kcreddy: Actually, it doesn't seem to work:
```
failed to check program: failed compilation: ERROR: <input>:84:36: found no matching overload for '_?_:_' applied to '(bool, map(string, dyn), null)'\n |             \"last_page\": want_more ?\n | ...................................^ accessing config
```
Doesn't allow returning a `map` and `null` in opposite conditions.

efd6: You can wrap the `null` and object in `dyn`, `<cond> ? dyn({…}) : dyn(null)` and it will work.

efd6 (approved):
LGTM, but please wait for @chrisberkhout

packages/mimecast/data_stream/archive_search_logs/agent/stream/cel.yml.hbs
chrisberkhout: Thanks!

PR #12949, opened 2025-03-04
[O365 Metrics] Parse nested fields for Teams Call Quality and Subscriptions data streams

lucian-ioan:
## Proposed commit message
Parse nested fields for Teams Call Quality and Subscriptions data streams.

This is already being done for O365 and is a compromise between generating too many fields and not parsing the data at all.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- 

## Screenshots

Before data structure:
image

After data structure.
image


packages/o365_metrics/data_stream/subscriptions/_dev/test/pipeline/test-subscriptions.json-expected.json
```diff
@@ -38,6 +38,18 @@
                                 "provisioning_status": "Success",
                                 "service_plan_id": "8c098270-9dd4-4350-9b30-ba4703f3b36b",
                                 "service_plan_name": "ADALLOM_S_O365"
+                            },
+                            {
+                                "applies_to": "User",
+                                "provisioning_status": "Pending",
```
ishleenk17, line 44: In Discover will we be able to search for o365.metrics.applies_to field and also its corelation with o365.metrics.provisioning_status ?
Or we can see the corelation only in document as an array ?

ritalwar (approved):
LGTM!

packages/o365_metrics/data_stream/subscriptions/_dev/test/pipeline/test-subscriptions.json-expected.json
ishleenk17: Thanks @ritalwar . So as I see there would still be no way in which we will be able to tell which  service_plans.appliesto, corresponds to which service_plans.provisioning_status. Am I right ?
Since both these fields are coming as separate arrays.

ishleenk17: Thanks Lucian. I would not want this to be a blocker to the PR. So we can go ahead with it.
But can you please open another issue stating the above discussion.
Lets see if we can come up with another solution for it. 
I'll also look more into it. 
For now lets go ahead with "group".

ishleenk17 (approved):

PR #12961, opened 2025-03-05
Add new Wiz Cloud Configuration Finding Full Posture data stream and update misconfig transform to use it

maxcold:
## Proposed commit message

Adding a new `Wiz Cloud Configuration Finding Full Posture` data stream, which in contrast to the existing `Wiz Cloud Configuration Finding` data stream, ingests all Cloud Configuration Finding data every 24h to match the logic of the native Cloud Security Posture integration used by `cloudbeat`. Updated latest misconfigurations transform to use this new data stream as a source. Due to the change in the transform, this might be a breaking change for customers, they need to enable the new data stream when updating to still receive findings in their Findings > Misconfigurations view and other flows relying on cloud security posture data

## Checklist

## Author's Checklist

- [ ] 

1. build package with `elastic-package build`
2. start the stack corresponding to the Kibana version from the manifest with `elastic-package stack up`
3. Install the integration and ingest Wiz data, findings from Wiz should appear in Findings > Misconfiguration

## Related issues

Related to:
- https://github.com/elastic/kibana/issues/208935


packages/wiz/data_stream/cloud_configuration_finding_full_posture/manifest.yml
```diff
@@ -0,0 +1,66 @@
...
+        show_user: false
+        default: 30s
+      - name: enable_request_tracer
+        type: bool
+        title: Enable request tracing
+        multi: false
+        required: false
+        show_user: false
+        description: >-
+          The request tracer logs requests and responses to the agent's local file-system for debugging configurations. Enabling this request tracing compromises security and should only be used for debugging. Disabling the request tracer will delete any stored traces. See [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-cel.html#_resource_tracer_enable) for details.
```
efd6, line 33: Please wrap this text as it was in the other cases where the option is used.

packages/wiz/elasticsearch/transform/latest_cdr_misconfigurations/fields/base-fields.yml
```diff
@@ -12,6 +12,6 @@
 - name: event.dataset
   external: ecs
   type: constant_keyword
-  value: wiz.cloud_configuration_finding
```
efd6, line 15: Why is this changed?

packages/wiz/manifest.yml
```diff
@@ -1,15 +1,15 @@
...
+version: "3.0.0"
 description: Collect logs from Wiz with Elastic Agent.
 type: integration
 categories:
   - security
   - cloudsecurity_cdr
 conditions:
   kibana:
-    version: "^8.16.0 || ^9.0.0"
+    version: "~8.16.6 || ~8.17.4 || ^8.18.0 || ^9.0.0"
```
efd6, line 12: Can you explain why this complexity is required?

efd6 (commented):
Test expectations need to be updated.

packages/wiz/data_stream/cloud_configuration_finding_full_posture/fields/resource.yml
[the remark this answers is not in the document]
maxcold: I took this approach from our native Cloud Security Posture integration https://github.com/elastic/integrations/tree/main/packages/cloud_security_posture/data_stream/findings/fields 
Find it easier to navigate when split by ECS part. Is there a convention around that?

packages/wiz/elasticsearch/transform/latest_cdr_misconfigurations/fields/base-fields.yml
maxcold: the source of the transform now the new data stream `wiz.cloud_configuration_finding_full_posture`, that's why the change

packages/wiz/manifest.yml
maxcold: When adding a new data stream as a source for the transform we need a change in Elasticsearch, see related PR https://github.com/elastic/elasticsearch/pull/124074 This change was backported to 8.16.x, 8.17.x, that's why it will only work starting from 8.16.6 and 8.17.4

maxcold:
@efd6 
> Test expectations need to be updated.
did you mean adding system tests for the new data stream or smth else?

packages/wiz/data_stream/cloud_configuration_finding_full_posture/fields/resource.yml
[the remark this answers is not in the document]
efd6: I can't think of any other integrations of ours that do this.

packages/wiz/elasticsearch/transform/latest_cdr_misconfigurations/fields/base-fields.yml
efd6: Thanks

packages/wiz/data_stream/cloud_configuration_finding_full_posture/_dev/test/pipeline/test-cloud-configuration-finding-full-posture.log-expected.json
```diff
@@ -1,7 +1,6 @@
 {
     "expected": [
         {
-            "@timestamp": "2025-03-21T13:16:37.507860047Z",
```
efd6: This will get undone next time someone regenerates tests.

packages/wiz/_dev/deploy/docker/docker-compose.yml
```diff
@@ -26,6 +26,19 @@ services:
...
+      ports:
+        - 8090
+      volumes:
+        - ./files:/files:ro
+      environment:
+        PORT: '8090'
+      command:
+        - http-server
+        - --addr=:8090
+        - --config=/files/config-cloud_configuration_finding_full_posture.yml
```
efd6, line 41:
```suggestion
  wiz-cloud_configuration_finding_full_posture:
    image: docker.elastic.co/observability/stream:v0.15.0
    hostname: wiz-cloud_configuration_finding_full_posture
    ports:
      - 8090
    volumes:
      - ./files:/files:ro
    environment:
      PORT: '8090'
    command:
      - http-server
      - --addr=:8090
      - --config=/files/config-cloud_configuration_finding_full_posture.yml
```

packages/wiz/_dev/deploy/docker/files/config-cloud_configuration_finding_full_psoture.yml
efd6, line 1: config-cloud_configuration_finding_full_p**os**ture.yml

packages/wiz/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.0.0"
+  changes:
+    - description: Add new Cloud Configuration Finding Full Posture data stream. If you rely on Findings > Misconfigurations view, enable this new data stream.
```
efd6, line 4: I don't think the changelog is the right place to put instructions.

packages/wiz/data_stream/cloud_configuration_finding_full_posture/fields/resource.yml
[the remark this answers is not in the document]
maxcold: any specific issue with it?

packages/wiz/data_stream/cloud_configuration_finding_full_posture/_dev/test/pipeline/test-cloud-configuration-finding-full-posture.log-expected.json
maxcold: yep, fixed with the `dynamic_fields`

packages/wiz/data_stream/cloud_configuration_finding_full_posture/fields/resource.yml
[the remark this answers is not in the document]
maxcold: Just to clarify, I can move everything to one or a smaller number of files if you think it is a hard requirement. I just don't understand the guidance here. I see different approaches in different data streams and am not sure what is the preferred way and why. For example, just took a random data stream: https://github.com/elastic/integrations/tree/main/packages/windows/data_stream/sysmon_operational/fields. What is the logic of the split by files there? I'm not sure. I haven't had issues with the split we have in `cloud_security_posture` integration; that's why the question

packages/wiz/_dev/build/docs/README.md
```diff
@@ -105,6 +106,16 @@ This is the `Cloud Configuration Finding` dataset.
...
 
+### Cloud Configuration Finding Full Posture
+
+This is the `Cloud Configuration Finding Full Posture` dataset.
+
+#### Example
+
+{{event "cloud_configuration_finding_full_posture"}}
+
+{{fields "cloud_configuration_finding__full_posture"}}
```
kcreddy, line 117: Looks like there is a typo which is making fields unrendered in docs/README.

```suggestion
{{fields "cloud_configuration_finding_full_posture"}}
```

packages/wiz/elasticsearch/transform/latest_cdr_misconfigurations/transform.yml
```diff
@@ -1,6 +1,6 @@
 source:
   index:
-    - "logs-wiz.cloud_configuration_finding-*"
+    - "logs-wiz.cloud_configuration_finding_full_posture-*"
```
kcreddy, line 3: I think any updates to transform requires bumping `fleet_transform_version`.
In this case, it could be `fleet_transform_version: 0.3.0`

packages/wiz/manifest.yml
kcreddy: But wouldn't this change make make users believe agentless is compatible on `~8.16.6 || ~8.17.4`?
Because we are merging the PR into `main` which had agentless change: https://github.com/elastic/integrations/pull/13105

maxcold: @kcreddy this is a good question, thanks for bringing it up. I will ask the agentless team starting from which version we support the `deployment_modes` . But even if we don't support them in 8.16 and 8.17 which we probably don't, I wonder if smth breaks because of this new addition to the manifest, eg. validation. If agentless options just won't be available, I think it should be fine, Agentless feature by itself is going GA in 8.18, so it should be clear for the users. 
But ofc we can go with only 8.18/9.0 for this change, it will just reduce the adoption

packages/wiz/elasticsearch/transform/latest_cdr_misconfigurations/transform.yml
maxcold: good catch, I updated the version to 0.2.0 initially but with merging `main` another update came in

seanrathier: 💡  Good information to know.  I was perplexed why the transform versions did not match the Kibana version.

packages/wiz/manifest.yml
seanrathier: The `deployment_mode` was added in package-spec and Fleet support in 8.16, in 8.17 it was add for the Fleet extensions like CSPM.

seanrathier (commented):
LGTM.  Let me know when you finish testing and I will review again.

packages/wiz/manifest.yml
maxcold: I tested and 8.16.6-SNAPSHOT works fine with the changes I have here. Plus the mode itself was added in 8.16 https://github.com/elastic/integrations/pull/12961#discussion_r2014325623

kcreddy: Sounds good. Thanks for testing @maxcold 👍🏼

packages/wiz/data_stream/cloud_configuration_finding_full_posture/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,149 @@
+config_version: 2
+interval: 24h
```
kcreddy, line 2: If I look at this CEL program, it is almost identical to its corresponding incremental data stream [`cloud_configuration_finding`](https://github.com/elastic/integrations/blob/main/packages/wiz/data_stream/cloud_configuration_finding/agent/stream/cel.yml.hbs). There are few differences such as hardcoded `interval`, removing incremental data fetching using [`analyzedAt`](https://github.com/elastic/integrations/blob/main/packages/wiz/data_stream/cloud_configuration_finding/agent/stream/cel.yml.hbs#L89-L95), and removing handling of [`cursor`](https://github.com/elastic/integrations/blob/main/packages/wiz/data_stream/cloud_configuration_finding/agent/stream/cel.yml.hbs#L104-L126).

What if we present user with an option inside the existing data stream `cloud_configuration_finding` itself, such as 
a [boolean](https://github.com/elastic/integrations/blob/main/packages/wiz/data_stream/cloud_configuration_finding/manifest.yml#L42-L53) variable `full_pull` which defaults to `true`, and let the user choose full or incremental payloads?
Was this already thought about? Are there any issues you foresaw? It seems the only reason we are adding this new data stream is to match it with `Cloud Security Posture` style `24h` pull, and the data is exactly same as existing [`cloud_configuration_finding`](https://github.com/elastic/integrations/blob/main/packages/wiz/data_stream/cloud_configuration_finding).

kcreddy (commented):
Just wanted a clarification on the approach. otherwise it looks good.

packages/wiz/data_stream/cloud_configuration_finding_full_posture/agent/stream/cel.yml.hbs
maxcold: @kcreddy thanks for bringing it up! yes, this indeed is the challenge we discussed with product and opted out for a separate data stream due to the following challenges:
- integration default UX can become very confusing when different options are presented to users and it's not clear how they work in combination. For example, if we present the full posture option on the existing data stream, it's not possible to hide other params (initial lookback interval and period interval) . So it will be confusing for the users how do they relate to each other and if they work together. Ofc we can solve it in the docs and in the field descriptions but it's not a good UX
- another problem is that the data from incremental and full posture are two separate usecases solving different problems. Users might want to have them both potentially. Incremental to base their alerts on and full posture to enable our Cloud Security flows (Findings page and context in different places of Security solution). Adding the option to existing data stream will make users to chose one flow, which might be limiting. Ofc it is probably possible to solve all with a more complex CEL logic, but in my opinion it defeats this purpose `The problem with multiple data streams handling same API is in maintenance during enhancement/bugfixes which needs to be propagated to multiple places now` - overly complex data stream with two modes that can coexist or work separately is very hard to maintain without incidental bugs in my opinion.

kcreddy: I think to add onto good UX, maybe we should also change the `title` and `description` of existing [cloud_configuration_finding](https://github.com/elastic/integrations/blob/main/packages/wiz/data_stream/cloud_configuration_finding) data stream? Do you think `Full posture` vs `Incremental posture` are self-intuitive or do you think documenting the difference in the integration docs makes more sense?

kcreddy: Also, is this going to be applied to all existing and future implementations, such as [AWS Security Hub](https://github.com/elastic/integrations/blob/main/packages/aws/data_stream/securityhub_findings/agent/stream/httpjson.yml.hbs#L30-L34), which pulls incrementally?

How about vulnerability flows?

maxcold: @smriti0321 
> I think to add onto good UX, maybe we should also change the title and description of existing [cloud_configuration_finding](https://github.com/elastic/integrations/blob/main/packages/wiz/data_stream/cloud_configuration_finding) data stream? Do you think Full posture vs Incremental posture are self-intuitive or do you think documenting the difference in the integration docs makes more sense?

Do we want to change the existing data stream to state clearly "Incremental"? Also what do we do about the documentation, do we want to document that users need to enable this new data stream if they want to leverage CDR flows?

maxcold: @kcreddy , yes, the plan is to implement similar for AWS Security Hub. For vulnerability flows we decided not to deal with the problem of incremental data flow for now. The posture much more relevant for CSPM data (Misconfiguration Fingings). Even though it's nice to have the full data of all the vulnerabilities in the system (and as we know our InfoSec team pulls the whole data set from Qualys) , it's less critical

packages/wiz/_dev/deploy/docker/files/config-cloud_configuration_finding_full_posture.yml
```diff
@@ -0,0 +1,36 @@
...
...as full permissions across the entire account, creating access keys for it increases the chance that they will be compromised. Instead, it is recommended to create IAM users with predefined roles.\n>**Note** \nSee Cloud Configuration Rule `IAM-207` to see if the Root account's access keys are active.","id":"563ed717-4fb6-47fd-929e-9c794e201d0a","name":"Root account access keys should not exist","remediationInstructions":"Perform the following steps, while being signed in as the Root user, in order to delete the root user's access keys via AWS CLI: \n1. Use the following command to list the Root user's access keys. \nCopy the `AccessKeyId` from the output and paste it into the `access-key-id` value in the next step. \n```\naws iam list-access-keys\n```\n2. Use the following command to delete the access key(s). \n```\naws iam delete-access-key /\n --access-key-id <value>\n```\n>**Note** \nOnce an access key is removed, any application using it will not work until a new one is configured for it.","shortId":"IAM-006"},"severity":"MEDIUM"}],"pageInfo": {"hasNextPage": true,"endCursor": "eyJmaWVsZHMiOlt7IkZpZWxkIjoiVGltZXN0YW1wIiwiVmFsdWUiOiIyMDIzLTA5LTA0VDExOjE5OjM3LjgwMTU0MVoifV19"}}}}
```
efd6, line 23: For the same of future maintenance, please pretty print the object and use the `minify_json` helper ([example](https://github.com/elastic/integrations/blob/ca62d9645d1baed7e65fc9d721a5a7424d5afa77/packages/abnormal_security/_dev/deploy/docker/files/config.yml#L60-L76)). Also below.

packages/wiz/data_stream/cloud_configuration_finding_full_posture/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,149 @@
...
+              "message": "POST " + state.url.trim_right("/") + "/graphql:" + (
+                size(resp.Body) != 0 ?
+                  string(resp.Body)
+                :
+                  string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
+              ),
+            },
+          },
+          "want_more": false,
+        }
```
efd6, line 130:
```suggestion
      resp.Body.decode_json().as(body, body.?data.configurationFindings.nodes.orValue(null) != null ?
        {
          "events": body.data.configurationFindings.nodes.map(e, {
            "message": e.encode_json(),
          }),
          "end_cursor": {
            ?"value": body.?data.configurationFindings.pageInfo.hasNextPage.orValue(false) ?
                body.?data.configurationFindings.pageInfo.endCursor
              :
                optional.none()
          },
          "want_more": body.?data.configurationFindings.pageInfo.hasNextPage.orValue(false),
        }
      :
        {
          "events": [],
          "want_more": false,
        }
      )
    :
      {
        "events": {
          "error": {
            "code": string(resp.StatusCode),
            "id": string(resp.Status),
            "message": "POST " + state.url.trim_right("/") + "/graphql:" + (
              size(resp.Body) != 0 ?
                string(resp.Body)
              :
                string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
            ),
          },
        },
        "want_more": false,
      }
```

kcreddy (approved):
LGTM. 
The reasons behind adding a new data-stream is here: https://github.com/elastic/integrations/pull/12961#discussion_r2016032512

Can wait for @efd6 approval before merging.

packages/wiz/_dev/deploy/docker/files/config-cloud_configuration_finding_full_posture.yml
```diff
@@ -0,0 +1,127 @@
...
+      Authorization:
+        - 'Bearer xxxx'
+    request_body: /.*"after":"eyJmaWVsZHMiOlt7IkZpZWxkIjoiVGltZXN0YW1wIiwiVmFsdWUiOiIyMDIzLTA5LTA0VDExOjE5OjM3LjgwMTU0MVoifV19".*/
+    responses:
+      - status_code: 200
+        headers:
+          Content-Type:
+            - application/json
+        body: |-
+              {{ minify_json `
```
efd6, line 79: nit: consistent indentation

packages/wiz/data_stream/cloud_configuration_finding_full_posture/manifest.yml
```diff
@@ -0,0 +1,70 @@
+title: Collet full Cloud Configuration Finding posture from Wiz.
```
efd6, line 1:
```suggestion
title: Collect full Cloud Configuration Finding posture from Wiz.
```

packages/wiz/data_stream/cloud_configuration_finding_full_posture/manifest.yml
```diff
@@ -0,0 +1,70 @@
...
+        multi: false
+        default: false
+      - name: processors
+        type: yaml
+        title: Processors
+        multi: false
+        required: false
+        show_user: false
+        description: >-
+          Processors are used to reduce the number of fields in the exported event or to enhance the event with metadata. This executes in the agent before the logs are parsed. See [Processors](https://www.elastic.co/guide/en/beats/filebeat/current/filtering-and-enhancing-data.html) for details.
```
efd6, line 70: nit: break at reasonable line lengths

efd6 (approved):
One fix required, two nits that you can take or leave.

PR #13009, opened 2025-03-07
checkpoint_harmony_endpoint: improve handling of 404 and 503 errors

chemamartinez:
## Proposed commit message

```
Changes in the error handling:

- When receiving a 503 Service Unavailable response, the sequence is restarted gratefully, cleaning the task ID and page token and waiting for the next interval.
- When receiving a 404 Not Found, the task ID is requested again for the same timeframe.
```

> [!TIP]
> Review commit by commit for a better understanding. The first one just propagates improvements from https://github.com/elastic/integrations/pull/12795 and https://github.com/elastic/integrations/pull/12158 to the rest of the data streams, while the second contains the new changes.

> [!NOTE]

## Checklist

### DIFF
--- a/packages/checkpoint_harmony_endpoint/data_stream/antibot/agent/stream/cel.yml.hbs
+++ b/packages/checkpoint_harmony_endpoint/data_stream/antibot/agent/stream/cel.yml.hbs
@@ -69,7 +69,9 @@ program: |
   			"startTime": state.?cursor.next_startTime.orValue(
   				timestamp(now() - duration(state.initial_interval)).format(time_layout.RFC3339)
   			),
-  			"endTime": timestamp(now() - duration("1m")).format(time_layout.RFC3339),
+  			"endTime": state.?cursor.next_endTime.orValue(
+  				timestamp(now() - duration("1m")).format(time_layout.RFC3339)
+  			),
   		}.as(timeframe,
   			request("POST", state.url.trim_right("/") + "/app/laas-logs-api/api/logs_query").with(
   				{
@@ -116,7 +118,10 @@ program: |
   									"auth_data": auth_data,
   									"task_id": body.data.taskId,
   									"page_token": null,
+  									"current_startTime": timeframe.startTime,
+  									"current_endTime": timeframe.endTime,
   									"next_startTime": timeframe.endTime,
+  									"next_endTime": null,
   								},
   							}
   						)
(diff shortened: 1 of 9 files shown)


packages/checkpoint_harmony_endpoint/data_stream/antibot/agent/stream/cel.yml.hbs
```diff
@@ -144,7 +149,24 @@ program: |
   					}
   				)
   			:
-  			resp.StatusCode == 200 ?
+  			(resp.StatusCode == 404) ?
+  				// 404 Not Found - Resubmit the task ID query for the same timeframe.
+  				state.with(
+  					{
+  						"events": [{"message": {"event": {"reason": "polling"}}.encode_json()}],
```
chrisberkhout, line 156: When `want_more` is true, a dummy event is required to make it immediately re-run. If there are no events it will wait for an interval.

Here, `want_more` is false, so that dummy event is unnecessary.

I think it should either be an empty array:
```suggestion
  						"events": [],
```
or a single event with an error message (single events get logged):
```suggestion
  						"events": { "error": {"message": "404: task ID not found" }},
```
or the same with the polling message, so the event will be logged, but the pipeline will drop it:
```suggestion
  						"events": {"message": {"event": {"reason": "polling"}}.encode_json(), "error": {"message": "404: task ID not found"}},
```

The last one would be easier if the pipeline expected parsed JSON instead of `{ "message": "JSON string" }`, because the CEL input logging code only logs a single message if it's parsed JSON.

I don't have a strong opinion about which one. The later ones are more messy, but get some useful information in the log/ES. The first one is clean and 404s shouldn't be happening a lot, and if they are they can always be seen with request tracing.

Here's a relevant part of the [CEL documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-cel.html):

>The field should be an array, but in the case of an error condition in the CEL program it is acceptable to return a single object instead of an array; this will will be wrapped as an array for publication and an error will be logged. If the single object contains a key, "error", the error value will be used to update the status of the input to report to Elastic Agent. This can be used to more rapidly respond to API failures.

chrisberkhout (approved):
Looks good. One point of clean-up with some options.

packages/checkpoint_harmony_endpoint/data_stream/antibot/agent/stream/cel.yml.hbs
chemamartinez: I think you are talking about the 503 case instead of the 404 (that's the one that sets `want_more` to false).

Thanks for the detailed explanation, I would choose to log the error, as receiving a 503 is not very common and so we can discover it more easily. My question is if we are going to log the error, why not use the same format as for the rest of the errors? I have submitted a commit with the changes, would you mind taking a look at it again?

chrisberkhout: Yes, you're right it was 503. The change looks good. I think that kind of logged error event could be used for all error cases. Doesn't necessarily have to be now.

For these error cases there are two choices:
- return single error event to be logged?
- keep that error event in Elasticsearch or drop it from the ingest pipeline?

So for now it's logged and not dropped, which is fine. I think in the future we could have some conventions for those choices.

chrisberkhout (approved):

PR #13016, opened 2025-03-07
[CI] Update teams permissions in pull request pipeline

mrodm:
## Proposed commit message

Update team permissions related to Pull Request pipeline in `catalog-info.yaml`.

Teams added in this PR are the ones found in `.github/CODEOWNERS` file. These are the teams:
```
@elastic/cloud-security-posture
@elastic/customer-architects
@elastic/ecosystem
@elastic/elastic-agent
@elastic/elastic-agent-data-plane
@elastic/fleet
@elastic/logstash
@elastic/ml-ui
@elastic/obs-ds-hosted-services
@elastic/obs-ds-intake-services
@elastic/obs-infraobs-integrations
@elastic/obs-ux-management-team
@elastic/profiling
@elastic/protections
@elastic/search-extract-and-transform
@elastic/sec-applied-ml
@elastic/sec-deployment-and-devices
@elastic/sec-linux-platform
@elastic/security-asset-management
@elastic/security-service-integrations
@elastic/sec-windows-platform
@elastic/stack-monitoring
```

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -61,6 +61,48 @@ spec:
       teams:
         ingest-fp:
           access_level: MANAGE_BUILD_AND_READ
+        ecosystem:
+          access_level: MANAGE_BUILD_AND_READ
+        cloud-security-posture:
+          access_level: BUILD_AND_READ
+        customer-architects:
+          access_level: BUILD_AND_READ
+        elastic-agent:
+          access_level: BUILD_AND_READ
+        elastic-agent-data-plane:
+          access_level: BUILD_AND_READ
+        fleet:
+          access_level: BUILD_AND_READ
+        logstash:
+          access_level: BUILD_AND_READ
+        ml-ui:
+          access_level: BUILD_AND_READ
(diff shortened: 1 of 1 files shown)


jsoriano (approved):
:+1:

Would it be possible to expand the codeowners scripts so this file is updated automatically too? No need to do it now, we can do it later.

mrodm:
> Would it be possible to expand the codeowners scripts so this file is updated automatically too? No need to do it now, we can do it later.

I guess, it could be updated the logic in `dev/codeowners/codeowners.go` to warn if the configuration of that pipeline does not contain some team (codeowner).

Another option could be to run some script via updatecli as it is done to update the templates for issues here: https://github.com/elastic/integrations/blob/16b897d40c5dfa6aea80f662584932c5be07e249/.github/workflows/updatecli/updatecli.d/sync-packages-to-bug-issue-template.yml#L31

The concern is that `catalog-info.yml` contain multiple YAML documents and I don't know how easy or difficult it would be to update it automatically.

PR #13020, opened 2025-03-07
Aligned data pulls to calendar days

3kt:
In the previous implementation of the integration, the billing data was pulled with an offset starting when the integration is first run. This had the side effect of aligning days to whatever current time it was at installation.
IE instead of pulling from 00:00 to 23:59, it could be from 11:00 to 10:59 for instance.

This fixes this behavior, while also changing the lookbehind behavior: it used to be a count of hours, whereas it is now a count of days.


VimCommando (approved):
LGTM

PR #13056, opened 2025-03-11
crowdstrike.fdr: Increase field limits in FDR data-streams to avoid unindexed ECS fields

kcreddy:
## Proposed commit message
```
Crowdstrike FDR data stream can collect more than 1000 fields
and the data stream explicitly defines more than 500 fields.
In such cases, the total_fields.limit is automatically updated to 
10000[1] from the default 1000. 
Having such dynamic resizing could be detrimental when some of 
these explicitly mapped fields are later removed and the limit goes 
back to 1000. If it goes back to 1000, then it could end up not 
indexing all fields, even the ECS fields.

This PR sets the field limit to 2000 to facilitate 
indexing of all fields, especially ECS.

Ref: 
- [1] https://www.elastic.co/guide/en/integrations-developer/current/general-guidelines.html#_field_limits
```

## Checklist

Current (1.54.0) version:
`GET .ds-logs-crowdstrike.fdr-default-2025.03.11-000001/_settings`
```json
{
  ".ds-logs-crowdstrike.fdr-default-2025.03.11-000001": {
    "settings": {
      "index": {
        "mapping": {
          "total_fields": {
            "limit": "10000",
            "ignore_dynamic_beyond_limit": "true"
          },
          "ignore_malformed": "true"
        },
```

With new PR change
`GET .ds-logs-crowdstrike.fdr-default-2025.03.11-000001/_settings`
```json
{
  ".ds-logs-crowdstrike.fdr-default-2025.03.11-000001": {
    "settings": {
      "index": {
        "mapping": {
          "total_fields": {
            "limit": "2000",
            "ignore_dynamic_beyond_limit": "true"
          },
          "ignore_malformed": "true"
        },
```

## Related issues

-

### DIFF
--- a/packages/crowdstrike/changelog.yml
+++ b/packages/crowdstrike/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Increase field limits in FDR data-streams to avoid unindexed ECS fields.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/
+      link: https://github.com/elastic/integrations/pull/13056
 - version: "1.54.0"
   changes:
     - description: Enable request trace log removal.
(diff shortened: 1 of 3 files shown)


efd6 (approved):
Behaviour change confirmed.

PR #13059, opened 2025-03-11
[cisco_asa] Time with fraction-of-seconds support

jrmolin:
## Proposed commit message

[cisco_asa] Timestamp with fraction-of-seconds now parsed

* Some Cisco timestamps had fraction-of-second showing up and causing pipeline errors
* Two new log examples added
* No existing pipeline tests failed
* New logs parsed successfully

## Checklist

- ~[ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices)~

## Author's Checklist

- [ ] 

```
cd packages/cisco_asa
elastic-package test pipeline
```

## Related issues

-


jrmolin:
previous PR: https://github.com/elastic/integrations/pull/12202

mjwolf (commented):
This looks good to me, but could you also add a test case with fractions-of-seconds and a non-UTC timezone? It would be good to verify they work together

packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -172,24 +172,53 @@ processors:
...
+                    - "MMM  d yyyy HH:mm:ss[.SSSSSSSSS]"
+                    - "MMM dd yyyy HH:mm:ss[.SSSSSSSSS]"
+                    - "EEE MMM  d yyyy HH:mm:ss[.SSSSSSSSS]"
+                    - "EEE MMM dd yyyy HH:mm:ss[.SSSSSSSSS]"
+                    - "MMM  d yyyy HH:mm:ss[.SSSSSSSSS] z"
+                    - "MMM dd yyyy HH:mm:ss[.SSSSSSSSS] z"
+                    - "MMM  d yyyy HH:mm:ss[.SSSSSSSSS] z"
+                    - "MMM dd yyyy HH:mm:ss[.SSSSSSSSS] z"
+                    - "EEE MMM  d yyyy HH:mm:ss[.SSSSSSSSS] z"
+                    - "EEE MMM dd yyyy HH:mm:ss[.SSSSSSSSS] z"
```
taylor-swanson, line 221: This is getting a bit ridiculous... I think we should look at re-doing this part.

I was playing around with the pipeline simulator and came up with this:

```
"[EEE ]MMM [ ]d[ yyyy] HH:mm:ss[.nnn][ z]"
```

I haven't tested all cases yet, but this does seem to catch most of them. One caveat, I truncated the fractional seconds down to three digits (millisecond precision). The date processor does this already, so I don't think this is a problem.

taylor-swanson (changes requested):
Okay, can confirm, I did test the following and all tests pass, except one, where `event.timezone: "UTC"` shows up in the results 🤷 

```yaml
  #
  # Parse the date included in FTD logs
  #
  - date:
      if: ctx._temp_?.raw_date != null
      tag: parse_raw_date
      timezone: "{{{ event.timezone }}}"
      field: "_temp_.raw_date"
      formats:
        - "ISO8601"
        - "[EEE ]MMM [ ]d[ yyyy] HH:mm:ss[.nnnnnnnnn][.SSS][ z]"
      on_failure:
        # Try to re-parse as UTC to catch when TZ is invalid or unknown.
        - remove:
            field: event.timezone
            ignore_missing: true
        - date:
            if: ctx._temp_?.raw_date != null
            tag: "parse_raw_date_fallback"
            field: "_temp_.raw_date"
            formats:
              - "ISO8601"
              - "[EEE ]MMM [ ]d[ yyyy] HH:mm:ss[.nnnnnnnnn][.SSS][ z]"
```
By the way, I made a slight adjustment to the fractional seconds... nanoseconds is tried first, but will only match if there's 9 digits (as expected), or it will match against the milliseconds one.

As Michael mentioned, a couple more test cases are needed (especially with fractional seconds).

packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
jrmolin: I agree! I was getting some weird errors when updating, and I have to use the `.SSS` instead of `.n`, because the `.n` doesn't actually get output.

taylor-swanson (commented):
Updated changes seem fine to me. My original suggestion only supported milliseconds and nanoseconds, not microseconds. I don't think there's a clever way to do all of that in one pattern.

My only last request is that we have a nanosecond test case (can literally just copy one the microsecond case and add 3 extra digits at the end to make it nanoseconds).

taylor-swanson (approved):
LGTM

PR #13067, opened 2025-03-11
update `audit_manager` Kibana constraint integration to support to 9.0

fearful-symmetry:
## Proposed commit message

Updates the constraint to support the 9.0 stack. 

## Checklist

### DIFF
--- a/packages/auditd_manager/manifest.yml
+++ b/packages/auditd_manager/manifest.yml
@@ -1,7 +1,7 @@
 format_version: "3.0.0"
 name: auditd_manager
 title: "Auditd Manager"
-version: "1.18.3"
+version: "1.19.0"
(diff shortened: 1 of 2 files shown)


packages/auditd_manager/manifest.yml
```diff
@@ -1,15 +1,15 @@
...
+version: "1.19.0"
 description: "The Auditd Manager Integration receives audit events from the Linux Audit Framework that is a part of the Linux kernel."
 type: integration
 categories:
   - security
   - auditd
 conditions:
   kibana:
-    version: "^8.16.0"
+    version: ^8.16.0 || ^9.0.0
```
nicholasberlin, line 12: I notice this had `"` around it before, is the change necessary?

fearful-symmetry: hmm, yeah

nicholasberlin (approved):

PR #13080, opened 2025-03-12
[CI] Remove deps from with_mage

mrodm:
## Proposed commit message

Remove dependencies that are already defined in `go.mod` when calling `with_mage` in buildkite scripts.

Applied same change as #11143 

Relates #13077


jsoriano (approved):

mrodm:
Merging since CI failures are unrelated to the change in this PR

PR #13088, opened 2025-03-12
Cisco ise time format update

jrmolin:
- Bug fix

## Proposed commit message

[cisco_ise] Add fraction-of-second support for parsing timestamp
- Updated time formats similar to the cisco_asa formats (optional fraction-of-second fields)
- Added new tests
- No regressions

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/cisco_ise/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -69,9 +69,8 @@ processors:
   - date:
       field: _tmp.timestamp
       formats:
-        - MMM  d HH:mm:ss
-        - MMM dd HH:mm:ss
-        - MMM d HH:mm:ss
+        - MMM [ ]d HH:mm:ss[.SSSSSS]
+        - MMM [ ]d HH:mm:ss[.SSS]
```
bhapas, line 73: Can we not combine this way?

```suggestion
        - MMM [ ]d HH:mm:ss[.SSS][.SSSSSS]
```

jrmolin: i'm trying it now

ilyannn: But this would allow `.123.456789` as well which would be weird

taylor-swanson: Seems to work if you reverse it (tested with pipeline simulator):

```
[.SSSSSS][.SSS]
```

packages/cisco_ise/data_stream/log/elasticsearch/ingest_pipeline/pipeline_ad_connector.yml
```diff
@@ -57,12 +57,11 @@ processors:
       ignore_failure: true
   - date:
       field: _tmp.timestamp
-      formats:
+      formats: &formats
```
taylor-swanson, line 60: What is this supposed to be? I vaguely remember seeing this before, but I'm not sure of its purpose.

packages/cisco_ise/data_stream/log/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: @ilyannn 
> But this would allow `.123.456789` as well which would be weird

That does not appear to be the case:

```json
{
  "error": {
    "root_cause": [
      {
        "reason": "unable to parse date [Jan 9 15:23:20.123456.987]"
      }
    ],
    "reason": "unable to parse date [Jan 9 15:23:20.123456.987]",
    "caused_by": {
      "reason": "failed to parse date field [Jan 9 15:23:20.123456.987] with format [MMM [ ]d HH:mm:ss[.SSSSSS][.SSS]]",
      "caused_by": {
        "type": "date_time_parse_exception",
        "reason": "Text 'Jan 9 15:23:20.123456.987' could not be parsed, unparsed text found at index 21"
      }
    }
  }
}
```

My guess is the datetime formatter doesn't like using the same pattern again. Even if it did, I wouldn't be concerned. It would grab either one, but it would be a bogus timestamp anyway, since I'm not aware of any timestamps that do this sort of thing. If for some reason the two numbers had different meanings, I would be more concerned, but that's not the case here.

So to summarize, it will work with either format (3 or 6 digits), but not both at the same time.

packages/cisco_ise/data_stream/log/elasticsearch/ingest_pipeline/pipeline_administrative_and_operational_audit.yml
```diff
@@ -133,7 +133,7 @@ processors:
       ignore_missing: true
   - date:
       field: _tmp.timestamp
-      formats:
+      formats: &formats
         - yyyy-MM-dd HH:mm:ss.SSS
         - yyyy-MM-dd HH:mm:ss.SSSSSS
         - MMM [ ]d HH:mm:ss[.SSS]
```
bhapas, line 139: So the main difference between this `date` processor and the one [below](https://github.com/elastic/integrations/pull/13088/commits/b9a9caf2e30c7dbc88e40c02b1fe942ad2cde2b4#diff-f68a8a291ac91ab8bc1fb8984578589ca5745ebef40f495d59d0f53c222e1861R148) in line 148 is the timezone handling. So we are converting to UTC timezone first and then to the event timezone , if exists? Isn't there a way to combine them? I haven't tried though.

packages/cisco_ise/data_stream/log/elasticsearch/ingest_pipeline/pipeline_ad_connector.yml
```diff
@@ -57,12 +57,11 @@ processors:
...
         - yyyy-MM-dd HH:mm:ss.SSS
         - yyyy-MM-dd HH:mm:ss.SSSSSS
```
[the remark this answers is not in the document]
jrmolin: i didn't want to mess with anything that wasn't already needing an update. i can, though

packages/cisco_ise/data_stream/log/elasticsearch/ingest_pipeline/pipeline_ad_connector.yml
taylor-swanson: Gotcha, thanks!

packages/cisco_ise/data_stream/log/elasticsearch/ingest_pipeline/pipeline_administrative_and_operational_audit.yml
jrmolin: there is probably a way to combine them, but i'm trying to change as little as possible

bhapas: Not for this PR though. Just as an overall improvement of the pipeline

packages/cisco_ise/data_stream/log/elasticsearch/ingest_pipeline/pipeline_ad_connector.yml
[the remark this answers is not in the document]
taylor-swanson: That's fair, something that can be revisited in the future if we improve/de-dup time parsing.

taylor-swanson (approved):

PR #13098, opened 2025-03-13
Update: Add support for proxy in CEL for rubrik integration

redcinelli:
I added support of for proxy settings in rubriks integration. A client needed it.

## Checklist

### DIFF
--- a/packages/rubrik/changelog.yml
+++ b/packages/rubrik/changelog.yml
@@ -1,4 +1,10 @@
 # newer versions go on top
+- version: "0.9.1"
+  changes:
+    - description: |
+        Add support for proxy in the CEL output
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/tdb
 - version: "0.9.0"
   changes:
     - description: |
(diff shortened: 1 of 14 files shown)


gpop63:
Thanks for the contribution @redcinelli i'm wondering if we should move `proxy_url` to the other integration vars for consistency purposes wdyt @devamanv?

devamanv:
> we should move proxy_url to the other integration vars for consistency purposes 

@redcinelli what @gpop63 means here is that we have kept the config options at the top level vars in the package manifest file. This was done to avoid duplication as well as keeping all the common config required by all datastreams at the top level, and only have datastream-specific options in the respective datastream manifest. 
So, I'd also suggest to move the `proxy_url` option at the top level vars. Rest looks good to me.

redcinelli:
@gpop63 and @devamanv I believe this is fixed. Or did I miss understood what you were talking about.

@gpop63, I do think this is a setting that would gain from being set as default in all relevant (http based input) integrations. So many users as requesting it.

i'll mirror the changes on this [draft pr](https://github.com/elastic/integrations/pull/13115) too :D

gpop63 (approved):

devamanv:
@redcinelli thanks for making this change. Looks good to me.

redcinelli:
Thanks you both very much for your time. 💚

PR #13115, opened 2025-03-14
Update: Add proxy setting input for anomali's package

redcinelli:
## Proposed commit message

Add proxy setting input for anomali's package

## Checklist


paulrossmeier:
Hello @redcinelli 

This is a very popular enhancement that would drive the adoption of the anomali integration.

Is there any blockers to moving this draft to a pull request and merge that I can help with?

Thanks!

redcinelli:
@paulrossmeier the buildkite check does not pass. But so does the main branch ... what's up with that ?

packages/ti_anomali/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.26.3"
```
efd6, line 2:
```suggestion
- version: "1.27.0"
```
enhancement bump minor

packages/ti_anomali/manifest.yml
```diff
@@ -42,6 +50,10 @@ policy_templates:
       - type: cel
         title: "Anomali ThreatStream API"
         description: Retrieves indicators from the intelligence endpoint of the Anomali ThreatStream API
+        inputs:
+          - type: cel
+            title: "Collect indicators from the intelligence API using CEL Input"
+            description: "Retrieves indicators from the intelligence endpoint of the Anomali ThreatStream API using CEL Input"
```
efd6, line 56: Why is this added?

packages/ti_anomali/manifest.yml
```diff
@@ -34,6 +34,14 @@ icons:
...
     size: 216x216
     type: image/svg+xml
+vars:
+  - name: proxy_url
+    type: text
+    title: Proxy URL
+    multi: false
+    required: false
+    show_user: false
+    description: URL to proxy connections in the form of http[s]://<user>:<password>@<server name/ip>:<port>. Please ensure your username and password are in URL encoded format.
```
efd6, line 44: This should be in the intelligence manifest since the common manifest suggest that it applies to both data streams, and the threatstream data stream is not a REST API-based stream.

efd6 (commented):
This needs a change to the version field in the manifest file to match the version in the changelog and there will need to be an addition to the agent configs [here](https://github.com/elastic/integrations/blob/main/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs) adding the template text to make use of the var that you added.
```
{{#if proxy_url}}
resource.proxy_url: {{proxy_url}}
{{/if}}
```
similar to [here](https://github.com/elastic/integrations/blob/283ae7240e9333ed2ccdddf316d3538e5bcab13d/packages/abnormal_security/data_stream/audit/agent/stream/cel.yml.hbs#L7-L9).

packages/ti_anomali/manifest.yml
redcinelli: ... I have no clue this must be from the merge / rebase I have done earlier. good catch

redcinelli:
> his needs a change to the version field in the manifest file to match the version in the changelog and there will need to be an addition to the agent configs [here](https://github.com/elastic/integrations/blob/main/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs) adding the template t

I don't think it is needed. As in [packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs](https://github.com/elastic/integrations/blob/main/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs)

We already have it. Right ?

efd6:
> I don't think it is needed. As in [packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs](https://github.com/elastic/integrations/blob/main/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs)
>
> We already have it. Right ?

Yup, I missed that.

We do need the version bump in the manifest though.

redcinelli:
@efd6 1.26.2 to 1.27.0 right ? it should be good

redcinelli:
@efd6 it turn out this https://github.com/elastic/integrations/pull/13826 addressed already what this PR does. I beleive we can drop it.

Issue #11916, opened 2024-11-28
[ThreatConnect] Future Enhancements


jamiehynds:
ThreatConnect's integration with Elastic provides the ability to ingest indicators which can be leveraged by analytics including indicator match rules to alert analysts to observables in their environment matching IOC's from ThreatConnect. A user has provided some ideas for future enhancements to the integration to arm analysts with even more context while investigating threats in Elastic. 

**Indicator Fields**
Import all indicator fields directly associated with identified threat actors
Ensure full fidelity of original ThreatConnect intelligence metadata
Create mappings that preserve original context and classification of each indicator
Threat Actor Enrichment Data
Geolocation Information
Geographic origin of threat actors
Associated IP ranges and regions
Potential operational zones
-- Target Sector Intelligence

**Industries/sectors historically targeted**
Vertical-specific attack patterns
Threat actor specialization details
-- Threat Actor Associations

**Linked threat groups**
Known affiliations
Potential collaboration networks
Relationship mapping between actors

andrewkroh:
I've had talks with ThreatConnect and the user that requested to add optional fetching of associated group attributes to the ti_threatconnect integration. Here's what I believe needs to be done.

## Background
Currently, the ThreatConnect integration fetches indicators with their associated groups, but doesn't retrieve the attributes of these groups. This enhancement will add functionality to optionally fetch these group attributes through additional API calls, enriching the indicator data.

## Technical Context
- When the ThreatConnect `/api/v3/indicators` API returns data, it includes associatedGroups, but not their attributes
- To get the group attributes, we need to make additional requests to `/api/v3/groups/{associatedGroup.id}?fields=attributes`
- This requires one API call per unique group ID, which could significantly increase the number of API calls

## Implementation Requirements
1. Add an advanced configuration option (boolean) to enable/disable associated group attribute collection
2. The option's description should contain a warning that enabling this feature will:
   - Significantly increase the number of API calls to ThreatConnect
   - Increase storage requirements for the integration
3. When enabled, implement the following data flow:
   - Extract unique associatedGroup IDs from the indicator response
   - Make requests to `/api/v3/groups/{associatedGroup.id}?fields=attributes` for each unique group ID
   - Transform the attributes response and add to the corresponding groups in the indicator documents
   - Format each attribute as an object with `type` and `value` fields
4. Optimize API calls by:
   - Only fetching attributes once per unique group ID (even if the same group appears in multiple indicators)
[...]

PR #13121, opened 2025-03-14
[ti_threatconnect] Allow fetch of associated group attributes, map tag tactics

chrisberkhout:
## Proposed commit message

```
[ti_threatconnect] Allow fetch of associated group attributes, map tag tactics

This has been tested against the live API.

Fetching an indicator's associated groups and their attributes in a single
request, and the required permissions, are discussed in the ThreatConnect
documentation under "Include Additional Association Levels for a Field"[1].

[1]: https://docs.threatconnect.com/en/latest/rest_api/v3/additional_fields.html#include-additional-association-levels-for-a-field
```

## Discussion

For mapping the attributes I followed what was done for tags, etc. and made it a group of keywords.

For attributes and tactics I removed intermediate wrapper objects.

I only added tactics because that was being returned by the server for the the existing integration's query.

The new option is under "Advanced options". It has an impact on response size described in [this comment](https://github.com/elastic/integrations/issues/11916#issuecomment-2722214668) and provides users with the same guidance as other options that increase response size.

Activating "Include Group Associations in the response" and "Include Group Associations with their Attributes in the response" simultaneously seems to behave the same as activating the latter only.

## Checklist

## Related issues

- Relates #11916

## Screenshots

Permission changes are required to use the feature as per [documentation](https://docs.threatconnect.com/en/latest/rest_api/v3/additional_fields.html#include-additional-association-levels-for-a-field).

Screenshot 2025-03-19 at 12 41 53

New setting in integration:

Screenshot 2025-03-19 at 12 45 48

### DIFF
--- a/packages/ti_threatconnect/data_stream/indicator/fields/fields.yml
+++ b/packages/ti_threatconnect/data_stream/indicator/fields/fields.yml
@@ -568,13 +568,13 @@
               fields:
                 - name: last_used
                   type: date
-                  description: Date and time when tag was last used.
+                  description: Date and time when the tag was last used.
                 - name: name
                   type: keyword
-                  description: Name of tag.
+                  description: Name of the tag.
                 - name: owner
                   type: keyword
-                  description: The Organization, Community, or Source to which the Tag belongs.
+                  description: The Organization, Community, or Source to which the tag belongs.
                 - name: platforms
                   type: group
                   fields:
@@ -586,7 +586,7 @@
                       description: Platform on which tag is added.
                 - name: tactics
                   type: keyword
-                  description: Attack tactics associated with the tag technique.
+                  description: Attack tactics associated with the tag.
                 - name: technique
                   type: group
                   fields:
--- a/packages/ti_threatconnect/data_stream/indicator/manifest.yml
+++ b/packages/ti_threatconnect/data_stream/indicator/manifest.yml
@@ -21,7 +21,7 @@ streams:
         multi: false
         required: false
         show_user: false # Drives whether the input is shown at top and not under advanced settings.
-        description: Elastic has a default async search response size of 10MB. Enabling this option may result in larger response size, which may cause the Kibana discover view to have issues.
+        description: Elastic has a default async search response size of 10MB. Enabling this option may result in larger response size, which may cause problems when using the Kibana discover view.
         default: false # Default this bool to false on setup
       - name: include_group_assoc
         type: bool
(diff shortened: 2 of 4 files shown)


packages/ti_threatconnect/data_stream/indicator/fields/fields.yml
```diff
@@ -577,6 +584,9 @@
                     - name: data
                       type: keyword
                       description: Platform on which tag is added.
+                - name: tactics
+                  type: keyword
+                  description: Attack tactics associated with the tag technique.
```
efd6, line 589: Is this word order in the original documentation? ISTM that this should either be "technique tag" or "tagged technique". If this is what they have (suggested to be true from the docs below), then I guess it's fine to leave it.

packages/ti_threatconnect/data_stream/indicator/manifest.yml
```diff
@@ -31,6 +31,14 @@ streams:
         show_user: false
         description: Elastic has a default async search response size of 10MB. Enabling this option may result in larger response size, which may cause the Kibana discover view to have issues.
         default: false
+      - name: include_group_assoc_attribs
+        type: bool
+        title: Include Group Associations with their Attributes in the response
+        multi: false
+        required: false
+        show_user: false
+        description: This requires the API user account to have the option "Allow User to Exceed API Link Limit" enabled. Elastic has a default async search response size of 10MB. Enabling this option may result in larger response size, which may cause the Kibana discover view to have issues.
```
efd6, line 40:
```suggestion
        description: This requires the API user account to have the option "Allow User to Exceed API Link Limit" enabled. Elastic has a default async search response size of 10MB. Enabling this option may result in larger response size, which may cause the problems when using the Kibana discover view.
```

packages/ti_threatconnect/data_stream/indicator/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -800,6 +800,39 @@ processors:
             - _ingest._value.from
           tag: remove_duplicate_fields
           ignore_missing: true
+  - foreach:
```
efd6, line 803: I like that the unnecessary level is removed, but this doesn't happen for the other fields; they all have a "data" path component. Is that just unfortunate?

packages/ti_threatconnect/data_stream/indicator/manifest.yml
chrisberkhout: Changed to
> ... which may cause the problems when using the Kibana discover view.

in 4 places.

andrewkroh (approved):
LGTM. I'm happy you discovered the `associatedGroups.attributes` option.

packages/ti_threatconnect/data_stream/indicator/elasticsearch/ingest_pipeline/default.yml
efd6: Yeah, what's here is fine. This conversation documents that, so I think we are good.

efd6 (approved):

PR #13122, opened 2025-03-14
[tanium] Use `host.name` instead of `host.hostname`

chrisberkhout:
## Proposed commit message

```
[tanium] Use `host.name` instead of `host.hostname`

ECS defines both of these fields. The `host.hostname` field is
specifically for what is output by the `hostname` command. The
`host.name` field is broader and is the better field to use in this
case.
```

## Checklist


packages/tanium/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.16.0"
```
kcreddy, line 2:
```suggestion
- version: "1.16.1"
```
Looks like CI is failing because of this.

kcreddy (approved):
Just version change. LGTM otherwise 👍🏼

PR #13126, opened 2025-03-14
Fix org ID scientific notation

3kt:
## Proposed commit message

The org ID can sometimes be interpreted in scientific notation, which causes issues in the ESS billing dashboard. This fixes it by changing the type of org_id back to a string, directly from the Elastic Agent collection.


pathoge (approved):
LG

PR #13132, opened 2025-03-16
[O365 Metrics] Fix Teams Call Quality cel code in case of multiple page responses

lucian-ioan:
## Proposed commit message
Fix teams call quality cel code in case of multiple page responses.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/o365_metrics/changelog.yml
+++ b/packages/o365_metrics/changelog.yml
@@ -1,7 +1,12 @@
 # newer versions go on top
+- version: "0.6.6"
+  changes:
+    - description: Fix cel code for `teams_call_quality` in case of multiple page responses.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/12345
 - version: "0.6.5"
   changes:
-    - description: Convert nested fields to group for `teams_call quality` and `subscriptions` data streams.
+    - description: Convert nested fields to group for `teams_call_quality` and `subscriptions` data streams.
       type: enhancement
       link: https://github.com/elastic/integrations/pull/12949
 - version: "0.6.4"
--- a/packages/o365_metrics/manifest.yml
+++ b/packages/o365_metrics/manifest.yml
@@ -1,6 +1,6 @@
 name: o365_metrics
 title: Microsoft Office 365 Metrics
-version: "0.6.5"
+version: "0.6.6"
 description: Collect metrics from Microsoft Office 365 with Elastic Agent.(This integration is currently in development and not yet ready for general use)
 type: integration
 format_version: "3.0.2"


tommyers-elastic (approved):

packages/o365_metrics/data_stream/teams_call_quality/_dev/test/system/test-default-config.yml
```diff
@@ -11,7 +11,6 @@ data_stream:
     interval: 5m
     preserve_original_event: true
 assert:
-  hit_count: 1
```
ritalwar, line 14: Why are we removing this check?

packages/o365_metrics/data_stream/teams_call_quality/fields/fields.yml
```diff
@@ -5,6 +5,9 @@
     - name: id
       type: keyword
       description: Unique identifier for the call quality record
+    - name: call_record_id
```
ritalwar, line 8: Could you update the sample event JSON file and generate it with your system test? 
Also, since a new field has been added, it would be helpful to update the sample event file.

packages/o365_metrics/data_stream/teams_call_quality/_dev/test/system/test-default-config.yml
lucian-ioan: Because there can also be 2 hit counts if there is one document with 2 JSONs, each with it's own id.

ritalwar: Shouldn't we update our tests and this check accordingly, rather than removing it, since we are aware of the test data here?

lucian-ioan: If we want to keep the original test case as well, then yes.

tommyers-elastic: the config should always provide a predictable number of results, so i agree with richa, we should keep the assertions explicit.

packages/o365_metrics/data_stream/teams_call_quality/fields/fields.yml
```diff
@@ -2,233 +2,73 @@
...
+                      description: The display name of the caller
+                    - name: id
+                      type: keyword
+                      description: The unique user ID for the caller
+                    - name: tenant_id
+                      type: keyword
+                      description: The tenant ID of the caller's organization
+                    - name: user_principal_name
+                      type: keyword
+                      description: The tenant ID of the caller's organization
```
muthu-mps, line 66: Fix the description for user_principal_name.

packages/o365_metrics/data_stream/teams_call_quality/fields/fields.yml
```diff
@@ -241,264 +81,211 @@
...
-          type: group
-          description: Information about the caller (initiator)
-          fields:
-            - name: name
-              type: keyword
-              description: The name of the caller
-            - name: user_agent
+                    - name: user_principal_name
+                      type: keyword
+                      description: The tenant ID of the callee's organization
```
muthu-mps, line 86: Fix the description here.

packages/o365_metrics/data_stream/teams_call_quality/fields/fields.yml
```diff
@@ -2,233 +2,73 @@
...
-                  description: Whether other no sound issues were reported by the callee
-                - name: stopped
-                  type: boolean
-                  description: Whether the call was stopped prematurely by the callee
-        - name: identity
-          type: group
-          description: Identity of the callee
-          fields:
-            - name: user
+              description: Unique identifier for the segment
```
muthu-mps, line 30: nit: Can you add period for the descriptions to maintain consistency?  
```suggestion
              description: Unique identifier for the segment.              
```

packages/o365_metrics/data_stream/teams_call_quality/fields/fields.yml
```diff
@@ -241,264 +81,211 @@
...
+                    - name: port
+                      type: long
+                      description: Network port number used by media endpoint.
+                    - name: received_quality_event_ratio
+                      type: long
+                      description: Fraction of the call that the media endpoint detected the network was causing poor quality of the audio received.
+                    - name: reflexive_ip_address
+                      type: keyword
+                      description: IP address of the media endpoint as seen by the media relay server. This is typically the public internet IP address associated to the endpoint.
+                    - name: relay_ip_address
```
muthu-mps, line 134: Can we make the `IP fields type as ip`? Do you see any challenges?

packages/o365_metrics/data_stream/teams_call_quality/fields/fields.yml
```diff
@@ -241,264 +81,211 @@
...
+                      type: keyword
+                      description: IP address of the media relay server allocated by the media endpoint.
+                    - name: relay_port
+                      type: long
+                      description: Network port number allocated on the media relay server by the media endpoint.
+                    - name: sent_quality_event_ratio
+                      type: long
```
muthu-mps, line 141: The fields which are in `float earlier are now converted to long`. Why do we change the field type for ratio here? Do you think the ratio fields should be of type float?

muthu-mps:
The Json structure for this data stream has been changed. The metrics gets ingested in a different field format now. Are we making this change as the integration is in **beta** and not considering it as a breaking change?

packages/o365_metrics/data_stream/teams_call_quality/fields/fields.yml
ritalwar: Thanks for pointing that out. I’ll double-check the field types with Microsoft’s docs and make updates if needed.

ritalwar:
> The Json structure for this data stream has been changed. The metrics gets ingested in a different field format now. Are we making this change as the integration is in **beta** and not considering it as a breaking change?

Yes, we discussed this internally with @tommyers-elastic and @lalit-satapathy . In fact, we’ve even cleaned up several fields and retained/added only those that are necessary. Since the integration is still in beta, we felt it was better to improve the data stream now rather than later.

muthu-mps (approved):
LGTM!

muthu-mps:
@ritalwar - Performed the local testing to validate the data processing.

PR #13148, opened 2025-03-17
imperva_cloud_waf: clarify imperva file compression configuration

efd6:
## Proposed commit message

See title.

Apparently it is possible for people to interpret the original text as a conditional imperative rather than an imperative with an explanatory clause as is intended, resulting in incorrect configuration.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


chrisberkhout (approved):
Both look okay to me. I guess we'll see if the new one works better.

I wonder if the "Compress logs." at the start of the line is being misinterpreted. It could be "Log compression." which might be clearer for the documentation but might not match the UI anymore.

PR #13152, opened 2025-03-18
imperva_cloud_waf: step over non-UTF-8 data work items

efd6:
## Proposed commit message

```
imperva_cloud_waf: step over non-UTF-8 data work items

It is possible for users to incorrectly configure their Imperva Cloud
WAF logging to send compressed data (this is the default). This results
in a CEL evaluation failure since we depend on the data being string
CEF. When the non-UTF-8 data is converted to a string, CEL refuses to
peform the conversion by design. When a user has done this, it appears
that logs in the compressed form persist, so the agent is unable to move
past the integration-invalid data. This change tries to perform the
conversion, falling back to an error message being sent to the index
when it is not possible. This helps identify cases where the
configuration is incorrect, and allows the collection to step over the
bodies that are not consumable.

Unfortunately we cannot make use of fleet health notifications since
sending object errors prevents cursor updates, and so would result in
continuing to be stuck.
```

> [!WARNING]
> The CEL input in this integration is not tested in system tests. Please review with extra care.

PoC:
```
mito src.cel
! stderr .
cmp stdout want.txt

-- src.cel --
[
	b"\xc3\x28",
	b"good message",
].map(body,
	{
		"events": try(string(body), "error").as(body, type(body) == type("") ?
			[{"message": body}]
		:
			[{"error": {"message": body.error}}]
		),
	}
)
-- want.txt --
[
	{
		"events": [
			{
				"error": {
					"message": "invalid UTF-8 in bytes, cannot convert to string"
				}
			}
		]
	},
	{
		"events": [
			{
				"message": "good message"
			}
		]
	}
]
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/imperva_cloud_waf/data_stream/event/agent/stream/cel.yml.hbs
+++ b/packages/imperva_cloud_waf/data_stream/event/agent/stream/cel.yml.hbs
@@ -72,7 +72,22 @@ program: |
         }
       }).do_request().as(resp, resp.StatusCode == 200 ?
         bytes(resp.Body).as(body, {
-          "events": (string(body)+"|==|").split("|==|")[1].split("\n").filter(x,x!="").map(x,{"message":x}),
+          "events": try(string(body), "error").as(body, type(body) == type("") ?
+          	(body+"|==|").split("|==|")[1].split("\n").filter(x,x!="").map(x,{"message":x})
+          :
+            // This is an unrecoverable error for the worklist element; it indicates
+            // most likely that the data was compressed, or was encrypted. We cannot
+            // handle this, so we step past it and log the failure. We are in an
+            // unfortunate position where ideally we would like this to cause a health
+            // status note in fleet to result, but we cannot do this since this would
+            // delete the cursor update and we would still be stuck here, so we just
+            // log into the index in the hope that someone will see it.
+            [{"error":
+              {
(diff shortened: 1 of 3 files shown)


packages/imperva_cloud_waf/data_stream/event/agent/stream/cel.yml.hbs
```diff
@@ -72,7 +72,22 @@ program: |
...
+            // handle this, so we step past it and log the failure. We are in an
+            // unfortunate position where ideally we would like this to cause a health
+            // status note in fleet to result, but we cannot do this since this would
+            // delete the cursor update and we would still be stuck here, so we just
+            // log into the index in the hope that someone will see it.
+            [{"error":
+              {
+                "message": v.worklist[v.next].filename + ": " + body.error + ": check imperva waf logging is configured correctly",
+              },
+            }]
```
chrisberkhout, line 89: what about making it a single event so it's logged?

efd6: That's explained in the commit message and in the comment above; if we do that the cursor that gets written here will be dropped.

chrisberkhout (approved):
Very nice.

The commit description is good and clear.

PR #13158, opened 2025-03-18
Fixed transform job definition that was inflating ingest volume

3kt:
## Proposed commit message

Negative values were cut from the calculation of the delta between the start and end values of the metrics. This was causing the ingest volume to be inflated.

Values can be negative when a merge, or document deletion occurs.

This deletes the `Math.max(0, XX)` in the transform job bucket scripts where relevant (ie where the metrics aren't ever-increasing gauges that are reset upon node restart or shard migration)

## Checklist

### DIFF
--- a/packages/elasticsearch/changelog.yml
+++ b/packages/elasticsearch/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Fixed transform job definition that was inflating ingest volume
       type: bugfix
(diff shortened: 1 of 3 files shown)


pickypg (approved):
LGTM

consulthys (commented):
LGT Stack Monitoring

Issue #12911, opened 2025-02-26
[opencanary]: Sets all log message event.kind == alert leading to unwanted alert noise in Elastic Security


colin-stubbs:
### Integration Name

OpenCanary [opencanary]

### Dataset Name

opencanary.events

### Integration Version

0.3.0

### Agent Version

8.17.2

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.0.0

### OS Version and Architecture

Not Applicable

### Software/API Version

All

### Error Message

`opencanary.events` ingest pipeline contains this code,

```
  - set:
      description: Set event.kind
      tag: set_event_kind
      field: event.kind
      value:
        - alert
      ignore_failure: true
```

This sets ALL logs generated by OpenCanary with `event.kind == alert`.

This leads to the default External Alerts rule in Elastic Security generating unwanted noise alerts when OpenCanary honeypots restart.

The ingest pipeline should selectively set event.kind to "alert" *ONLY* when it's clear the log message is an actual canary interaction from a network client.

From the look of it we can do this using the `log.logger` field, as the boot time message are ID 1000-10006, e.g. translated as "LOG_BASE_MSG"

While Elastic Security exceptions *can* be made, it would be better if the integration just worked and didn't generate noise immediately.

Example noise,

<img width="1186" alt="Image" src="https://github.com/user-attachments/assets/105b3c25-617b-432b-a9e5-fe3e6852cf44" />

### Event Original

```
{
    "dst_host": "",
    "dst_port": -1,
    "local_time": "2025-02-24 16:45:09.388562",
    "local_time_adjusted": "2025-02-24 16:45:09.388591",
    "logdata": {
        "msg": {
            "logdata": "Added service from class CanaryFTP in opencanary.modules.ftp to fake"
        }
    },
    "logtype": 1001,
    "node_id": "opencanary-1",
    "src_host": "",
    "src_port": -1,
    "utc_time": "2025-02-24 16:45:09.388586"
}
```

and

```
{
    "dst_host": "",
    "dst_port": -1,
    "local_time": "2025-02-24 16:45:09.388754",
    "local_time_adjusted": "2025-02-24 16:45:09.388766",
    "logdata": {
        "msg": {
            "logdata": "Canary running!!!"
        }
    },
    "logtype": 1001,
    "node_id": "opencanary-1",
    "src_host": "",
    "src_port": -1,
    "utc_time": "2025-02-24 16:45:09.388762"
}
```

### What did you do?

[...]

PR #13162, opened 2025-03-19
[opencanary] Update Event ECS Fields to Remove Noise

mohitjha-elastic:
## Proposed Commit Message
**Update Event ECS Fields to Remove Unwanted Alert Noise.**
Log ID 1000-1006 is related to base message hence event.* fields should not be set in that case. Also, as requested in the issue, message ECS field is already present in the pipeline.

## Checklist

Clone integrations repo.
Install the elastic package locally.
Start the elastic stack using the elastic package.
Move to integrations/packages/opencanary directory.
`elastic-package test -v`

## Related issues

- Closes https://github.com/elastic/integrations/issues/12911


packages/opencanary/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.4.1"
+  changes:
+    - description: Update event ecs fields to remove unwanted alert noise.
```
efd6, line 4:
```suggestion
    - description: Update event ECS fields to remove unwanted alert noise.
```

packages/opencanary/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -48,6 +48,28 @@ processors:
...
       ignore_missing: true
       ignore_failure: true
+  - set:
+      description: Set event.category according to event details
+      tag: set_event_category
+      field: event.category
+      value:
+        - network
+        - intrusion_detection
+      if: ctx.opencanary?.logtype != null && ![1000, 1001, 1002, 1003, 1004, 1005, 1006].contains(ctx.opencanary.logtype)
```
efd6, line 58: Since opencanary.logtype is a long, can we
```suggestion
      if: ctx.opencanary?.logtype != null && (ctx.opencanary.logtype < 1000 || 1006 < ctx.opencanary.logtype)
```
instead? This saves an allocation, and reduces the number of comparisons. Similar below.

efd6 (approved):
Thanks

efd6:
Ah, OK. The test shows that that is already set. We don't need to mention that in the commit, and the "Anything else?" section in the issue is not a "Would you like fries with that?", it's a "Is there anything else we should know?"

PR #13163, opened 2025-03-19
Universal Profiling: Add Kibana constraint to support 9.0.0

florianl:
## Proposed commit message

Universal Profiling: Add Kibana constraint to support 9.0.0

## Checklist

### DIFF
--- a/packages/universal_profiling_agent/changelog.yml
+++ b/packages/universal_profiling_agent/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: 8.17.3
+  changes:
+    - description: Update Kibana constraint to support 9.0.0.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/13163
 - version: 8.17.2
   changes:
     - description: Update ownership
--- a/packages/universal_profiling_agent/manifest.yml
+++ b/packages/universal_profiling_agent/manifest.yml
@@ -1,11 +1,11 @@
 name: profiler_agent
 title: Universal Profiling Agent
-version: 8.17.2
+version: 8.17.3
 categories: ["elastic_stack", "monitoring"]
 description: Fleet-wide, whole-system, continuous profiling with zero instrumentation.
 conditions:
   kibana:
-    version: ^8.17.1
+    version: "^8.17.3 || ^9.0.0"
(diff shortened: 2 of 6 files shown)


rockdaboot (commented):
Just for my understanding:
Wouldn't it be more intuitive to implement a 9.0.0 version and saying that it is compatible with >= 8.17.1? Currently (if I understand correctly), this PR implements a version 8.17.3 that says it is compatible with everything >= 9.0.0, which might not be true any more in the future.

rockdaboot:
> > [..]  this PR implements a version 8.17.3 that says it is compatible with everything >= 9.0.0, which might not be true any more in the future.
> 
> This is not correct. This PR bumps the integration packages to version `8.17.3` and sets the minimal supported kibana version to `8.17.3 OR 9.0.0`.

Right, excuse my brevity, I left away the 8.17.3 compatibility part as that seemed to be obvious. The question was more if it possibly makes more sense to introduce a 9.0.0 package version that is compatible a minimal 8.x version of 8.17.3 *and* a minimal 9.x version of 9.0.0. Because it is expected that future versions of 8.x stay compatible with 9.0.0. But it is not clear if 8.17.3 stays compatible with future 9.x versions.
Just curious, not sure if my understanding is correct here.

florianl:
> The question was more if it possibly makes more sense to introduce a 9.0.0 package version that is compatible a minimal 8.x version of 8.17.3 and a minimal 9.x version of 9.0.0. 

The integration package version can be pushed any time. If there will be a feature, that is only compatible with 9.x, then it makes sense to me, to push also the integration package version to 9.x.
Personally, I think it is easier to use the minimal supported kibana version as integration package version, as long as the forwards compatibility is given. Having a 9.x integration package and needing to figure out which 8.x kibana version is supported feels like unneeded overhead.
Also, to follow the general Elastic version policy, everything that is marked as 9.x is not supported with 8.x components, while 8.x components might work with 9.x unless there is a breaking change.

> Because it is expected that future versions of 8.x stay compatible with 9.0.0. But it is not clear if 8.17.3 stays compatible with future 9.x versions.

As far as I know, there is no integration package concept that supports diverging integrations for major versions.

christos68k (approved):

PR #13189, opened 2025-03-20
[Google Threat Intelligence] Add Integration Package

niraj-crest:
- Enhancement

## What does this PR do?

- Added 4 data streams (cryptominer, first_stage_delivery_vectors, infostealer, iot).
- Added data collection logic for the data streams.
- Added the ingest pipeline for the data streams.
- Mapped fields according to the ECS schema and added Fields metadata in the appropriate yaml files.
- Added dashboards and visualizations.
- Added system test cases for the data stream.

## Checklist

- Clone integrations repo.
- Install elastic-package locally.
- Move to integrations/packages/ti_google_threat_intelligence) directory.
- Run the following command to run tests. elastic-package test


niraj-crest:
Hey @jamiehynds & @andrewkroh
Can anyone please add reviewers to this PR?

kcreddy (commented):
@niraj-crest, could you fix the CI error?
```
Error: there is no owner for "packages/ti_google_threat_intelligence" in ".github/CODEOWNERS"
```
by adding a line into [`.github/CODEOWNERS`](https://github.com/elastic/integrations/blob/main/.github/CODEOWNERS#L415)

niraj-crest:
Hello @kcreddy,
Added codeowners, can please run the tests again?

kcreddy:
@niraj-crest, looks like the PR is failing [`Check for blocked path changes`](https://github.com/elastic/integrations/actions/runs/14187368809/job/39759267735#logs). 
The git diff `git diff --name-only f9794072abccfc51e4549ecd7510d0d2248f2c72 1b3d6ddc5affe5f7215b2668ecdd1db7682202e9` is identifying files from [`blocked_paths`](https://github.com/elastic/integrations/blob/main/.github/workflows/block-paths.yml#L19-L22).

Can you re-sync your branch with elastic:main?

niraj-crest:
> @niraj-crest, looks like the PR is failing [`Check for blocked path changes`](https://github.com/elastic/integrations/actions/runs/14187368809/job/39759267735#logs). The git diff `git diff --name-only f9794072abccfc51e4549ecd7510d0d2248f2c72 1b3d6ddc5affe5f7215b2668ecdd1db7682202e9` is identifying files from [`blocked_paths`](https://github.com/elastic/integrations/blob/main/.github/workflows/block-paths.yml#L19-L22).
> 
> Can you re-sync your branch with elastic:main?

Done @kcreddy.

kcreddy:
@niraj-crest, I don't think its done. 

Screenshot 2025-04-02 at 11 25 27 AM

niraj-crest:
Hello @kcreddy,
CI is passing now, i guess you can start reviewing the PR.

kcreddy:
> @kcreddy,
> CI is passing now, i guess you can start reviewing the PR.

I've added few comments. I haven't looked into the entire PR yet. Will do.

packages/ti_google_threat_intelligence/elasticsearch/transform/domain_ioc/transform.yml
```diff
@@ -0,0 +1,48 @@
...
+        - exists:
+            field: error.message
+        - match_phrase:
+            event.dataset: ti_google_threat_intelligence.enriched_ioc
+dest:
+  index: "logs-ti_google_threat_intelligence_latest.dest_domain_ioc-1"
+  aliases:
+    - alias: "logs-ti_google_threat_intelligence_latest.domain_ioc"
+      move_on_creation: true
+  pipeline: 0.1.0-ti_google_threat_intelligence-latest_domain_ioc-transform-pipeline
```
kcreddy, line 20: We tried to avoid this hardcoded version dependency, especially on integration assets. This is because of the frequent changed to the integrations and this value is often forgotten, leading to errors during upgrade. Hence all the ti_* integrations don't have a destination pipeline.
There were already few instances where not bumping `dest.index` suffix or `fleet_transform_version` caused problems, but those 2 options only gets changed when the transform or destination index configuration itself needs a change (which is rare). But this version dependency is major one as it now needs to be bumped inside every PR. 
Unless we have some check in `elastic-package` which verifies the compares changelog version to the prefix of this `pipeline:` parameter (which we don't have right now), I am not quite confident going forward with this approach.
Is it possible to avoid the destination pipeline?

packages/ti_google_threat_intelligence/_dev/build/docs/README.md
```diff
@@ -0,0 +1,207 @@
+# Google Threat Intelligence
+
+## Overview
+
+[Google Threat Intelligence](https://gtidocs.virustotal.com/) is a security solution that helps organizations detect, analyze, and mitigate threats. It leverages Google's global telemetry, advanced analytics, and vast infrastructure to provide actionable insights. Key features include threat detection, malware and phishing analysis, and real-time threat alerts.
```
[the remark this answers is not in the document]
niraj-crest: Yes it is.

packages/ti_google_threat_intelligence/_dev/build/docs/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Enabling the integration in Elastic:
+
+1. In Kibana, go to **Management > Integrations**.
+2. In the "Search for integrations" search bar, type **Google Threat Intelligence**.
+3. Click on the **Google Threat Intelligence** integration from the search results.
+4. Click on the **Add Google Threat Intelligence** button to add the integration.
+5. While adding the integration, to collect logs via REST API, provide the following details:
+   - Enable the type of threat feed you have access to.
+   - Access Token
```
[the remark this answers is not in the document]
niraj-crest: No. user does not need to have any additional permissions.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,100 @@
...
+      state.with({
+        "start_time": state.?cursor.last_timestamp.orValue((now - duration(state.initial_interval)).format("2006010215")),
+        "end_time": now.format("2006010215"),
+      })
+  ).as(state, state.with(
+    request(
+      "GET",
+      state.url.trim_right("/") + "/api/v3/threat_lists/cryptominer/"+state.start_time+"?"+{
+        ?"query": has(state.query) && state.query != null ? optional.of([state.query]) : optional.none(),
+        "limit": ["2000"]
```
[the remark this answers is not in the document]
niraj-crest: Limit is increased to 4000 as per latest [doc](https://gtidocs.virustotal.com/reference/get-hourly-threat-list#:~:text=QUERY%20PARAMS-,limit,Maximum%20number%20of%20IoCs%20to%20retrieve%20(max%204000),-type). i will update it to 4000.

niraj-crest:
I will address all the comments and get back to you soon.

packages/ti_google_threat_intelligence/data_stream/cryptominer/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,766 @@
...
+  - set:
+      field: threat.indicator.timeout
+      tag: set_threat_indicator_timeout
+      copy_from: gti.cryptominer.attributes.last_analysis_stats.timeout
+      ignore_empty_value: true
+  - set:
+      field: threat.indicator.undetected
+      tag: set_threat_indicator_from_last_analysis_stats_field
+      copy_from: gti.cryptominer.attributes.last_analysis_stats.undetected
+      ignore_empty_value: true
```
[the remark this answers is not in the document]
niraj-crest: Sure @kcreddy,
We are thinking of changing all the custom threat.indiactor fields to gti.threat.indicator filed.
What we are trying to achieve is the following 

Once data collection is done and we have collected all the IOCs, We run transforms to filter the latest IOC documents based on type(File,url,domain,ip). At this stage we want a common root field across all the data streams, so that when users want to create visualization on latest IOCs, they don't have to take all the fields from all the data streams into consideration. For example we rename fields like gti.linux.malisious, gti.osx.malisious etc to gti.threat.indicator.malisious, so that users can easily analyze all the data. Additionally We pass on entire gyi.threat.indicator object to alerts generated by detection rules(We can only pass one object). We also perform one more transform on alerts so that we get latest alers and create enrichment visualization on the same.

So to complete this entire flow, it is necessary to have a single root filed after first transform for all the data streams.

We are thinking of gti.threat.indicator. 

Let us know what are your thoughts about it?

unnamed

packages/ti_google_threat_intelligence/data_stream/cryptominer/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,766 @@
...
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - set:
+      field: threat.indicator.id
+      tag: set_threat_indicator_id
+      copy_from: json.data.id
+      ignore_empty_value: true
+  - rename:
+      field: json.data.id
+      tag: rename_data_id_with_domain
+      target_field: gti.cryptominer.id.domain
```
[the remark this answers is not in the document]
niraj-crest: Sure, I will use `gti.cryptominer.id` field, but I suggest also keeping fields like  `domain.id` , `file.ip` etc as well, as it will make creating visualization easy for users as they won't have to apply `type` filter every time to byforget. 
let me know your thoughts!

packages/ti_google_threat_intelligence/elasticsearch/transform/domain_ioc/transform.yml
niraj-crest: As i mentioned above, we are palnning to create a uniformed document for all the data streams after first transform and also we are doing some imoprant processing in `Rule transform` as well to enrich IOCs. So we won't be able to avoid pipelines.

packages/ti_google_threat_intelligence/data_stream/cryptominer/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
kcreddy: @niraj-crest, sorry for the delay. I was OOO. I will get back to you on this tomorrow.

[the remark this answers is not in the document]
kcreddy: Thanks for the analysis @niraj-crest. 
The difference between this integration PR and earlier threat intel integrations are following:
1. New transform on detected indicator matches. This transform queries for alerts generated into `.internal.alerts-security.alerts-default-*` by the indicator match rules and stores them inside a destination index `logs-ti_google_threat_intelligence.enriched_ioc`.
2. A latest transform per [`threat.indicator.type`](https://www.elastic.co/guide/en/ecs/current/ecs-threat.html#field-threat-indicator-type) instead of per data stream (example: [Abusech](https://github.com/elastic/integrations/tree/main/packages/ti_abusech/elasticsearch/transform))

I see some potential benefits and also downsides.

**Benefits:**
1. By creating another latest transform on detected indicators and with the dashboard on its destination index, we derive more value from the data for the users.
2. Lesser number of transforms (currently 4 + 1) using per `threat.indicator.type` approach.

**Downsides:**
1. Because you now need a common field, for example: `gti.threat.indicator`, you had to use a destination pipeline on all the transform configurations. 
https://github.com/elastic/integrations/blob/d976d105ba87880763bb44009178aaf869ae300c/packages/ti_google_threat_intelligence/elasticsearch/transform/rule_transform/transform.yml#L17
The `pipeline` option needs to contain a prefix for the integration version. Since this value is updated in every PR, the `pipeline` option inside transform config also needs to be updated per PR (including bumping `fleet_transform_version`).

2. As per https://www.elastic.co/guide/en/security/current/detections-permissions-section.html, the users need a `Read` on `Security` feature to use the internal alerts indices. But to create/manage an integration, they need `Kibana` privileges https://www.elastic.co/guide/en/fleet/current/fleet-roles-and-privileges.html. The users who only want to create integration and ingest indicators and who don't have privileges to use the internal alerts indices will not be able to use the integration.

3. Currently GTI only has 4 indicator types with lot of products. While it may seem creating latest transform per `threat.indicator.type` instead of per data stream is beneficial, if GTI products are updated to send/consume other indicator types, this might also grow into an exhaustive list: https://www.elastic.co/guide/en/ecs/current/ecs-threat.html#field-threat-indicator-type

@andrewkroh, any thoughts on this approach?

niraj-crest:
@kcreddy Waiting for reply on this [comment](https://github.com/elastic/integrations/pull/13189#discussion_r2041353605). Once we have a finalized approach, I will push all the changes accordingly.

packages/ti_google_threat_intelligence/data_stream/cryptominer/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
andrewkroh: Lots of good points @kcreddy .

Regarding the rule_transform, I question whether we need another transform copying alert data. Have we explored using ES|QL queries to directly query the alerts index? This would avoid data duplication unless there's a specific performance or access pattern need that direct queries can't satisfy.

If the goal is to provide aggregated insights (e.g., "indicator X triggered N alerts"), a pivot transform would be more appropriate than a latest transform, as it would summarize rather than duplicate the alert data.

----

The pipeline versioning issue is concerning:

> The `pipeline` option needs to contain a prefix for the integration version... needs to be updated per PR

This manual update process seems error-prone and unsustainable. We should explore:
1. Adding automation in elastic-package or Fleet to update these references
2. Creating validation checks that catch missing updates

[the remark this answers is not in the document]
kcreddy: Thanks @andrewkroh.

@niraj-crest, based on Andrew's suggestion, it is unviable for us to use `pipeline` option inside the transform.
Can you check if ES|QL queries can accomplish a similar result instead of rule_transform?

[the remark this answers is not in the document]
niraj-crest: Hello @kcreddy The document generated in alert index stores the important information in threat.enrichments field, which is array of objects.
`{"threat": { "enrichments": [{ "indicator": {......}}]}}`
So by using rule transform pipeline we convert this array to single object which can be used to create visualizations. Now accessing this nested fields does not seem to be feasible in ES|QL. Now to achieve the above conversion we could try script parameter of pivot transform but it seems that it is [not available](https://www.elastic.co/docs/reference/aggregations/search-aggregations-metrics-scripted-metric-aggregation#:~:text=scripted_metric%20is%20not%20available%20in%20Elastic%20Cloud%20Serverless.) in Elastic cloud Serverless. So we need to have rule transform pipeline.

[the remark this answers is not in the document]
kcreddy: Hey @niraj-crest, thanks for the analysis.
I have created https://github.com/elastic/elastic-package/issues/2552 for adding the automation to the `pipeline` property. I believe this may be closely aligned with Fleet like Andrew said and probably wouldn't be available anytime soon. 
If we are looking forward to moving ahead with this integration, we should ship the 4 latest transforms (without rule transform) without `pipeline` property. This will make the integration lack insights into indicator matches, but it can be added as a future enhancement. WDYT?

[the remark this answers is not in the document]
niraj-crest: Let me explore little more. I will get back to you soon.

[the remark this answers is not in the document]
niraj-crest: Hey @kcreddy,
Just to clarify, the only concern with this approach is that updating the pipeline version is easily overlooked when bumping the package version. Since the pipelines are version-dependent, this could lead to issues for users.
Could you please confirm?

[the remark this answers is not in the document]
kcreddy: @niraj-crest, I understand the concern that its going to reduce the value for the users. Please note that none of the Threat Intel integrations currently derive this information. ES|QL would also have been my bet, but looks like it didn't work. 
If we want to get past this blocker for now, we can release the integration without the rule transform and `pipeline` parameter. Once https://github.com/elastic/elastic-package/issues/2552 is handled, we can revisit and make enhancement to the integration.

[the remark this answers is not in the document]
niraj-crest: Hello @kcreddy,
When can we expect [elastic/elastic-package#2552](https://github.com/elastic/elastic-package/issues/2552) to get added in elastic-package?
This would help us plan things out.

packages/ti_google_threat_intelligence/elasticsearch/ingest_pipeline/ti_google_threat_intelligence-latest_domain_ioc-transform-pipeline.yml
niraj-crest, line 1: Hey @kcreddy,

As discussed, we are proceeding with the initial integration without including the transform pipelines. However, the aggregation performed by the transform pipeline here is crucial, as without it, important data will remain scattered across all 15 data streams.

While the integration will still be functional, this lack of aggregation will complicate tasks like creating visualizations, saved searches, and detection rules. Additionally, once the transform pipelines are eventually added, the resulting schema changes will likely break any custom assets users may have already created.

To address this, we're considering providing users with steps to create a custom pipeline that replicates the behavior of our transform pipeline. This would allow us to build dashboards and visualizations on top of the processed data from the start.

Let us know your thoughts on this approach.

kcreddy: @niraj-crest, I am okay with going ahead with documenting steps for custom pipeline for now to unblock you. However, once the fix for https://github.com/elastic/elastic-package/issues/2552 is made, and we add the pipelines into our code, can you please ensure that users will have seemless experience with the upgrade?

niraj-crest:
@kcreddy I have addressed review comments, please run CI tests.

niraj-crest:
@kcreddy can you please run tests on the latest commit, also let me know if there are any other suggestions from your side.

niraj-crest:
@kcreddy can you please share SonarQube results?

niraj-crest:
@kcreddy Let me know if there are any other suggestion from your side. We are looking forward to merge this PR if everything seems fine.

packages/ti_google_threat_intelligence/elasticsearch/transform/ip_ioc/fields/iot_fields.yml
```diff
@@ -0,0 +1,264 @@
+- name: gti
```
kcreddy, line 1: Do we need these `<product>_fields.yml` files?
I see that `gti.<product>` is being renamed to `gti.threat.indicator` at the beginning of destination ingest pipelines of these transforms. So `gti.<product>` fields shouldn't be available in the destination index and we shouldn't need mapping for them.

packages/ti_google_threat_intelligence/elasticsearch/transform/url_ioc/fields/ecs.yml
```diff
@@ -0,0 +1,32 @@
...
+- external: ecs
+  name: related.hash
+- external: ecs
+  name: related.hosts
+- external: ecs
+  name: threat.feed.name
+- external: ecs
+  name: related.ip
+- external: ecs
+  name: threat.indicator.last_seen
```
kcreddy, line 24: I see some of the fields missing from this file, such as `threat.indicator.name`, `url.*`.
Could you please verify all fields are mapped?

packages/ti_google_threat_intelligence/elasticsearch/transform/ip_ioc/transform.yml
```diff
@@ -0,0 +1,45 @@
...
+retention_policy:
+  time:
+    field: '@timestamp'
+    max_age: 30d
+_meta:
+  managed: false
+  # Bump this version to delete, reinstall, and restart the transform during
+  # package installation.
+  fleet_transform_version: 1.0.0
+  run_as_kibana_system: false
```
kcreddy, line 45:
```suggestion
  run_as_kibana_system: false
```
Remove this, it shouldn't be needed for TI* transforms, they have permissions to run as kibana_system role.

The new `rule` transform may still need `run_as_kibana_system: false`, because its reading for `.internal*` indices

packages/ti_google_threat_intelligence/elasticsearch/ingest_pipeline/ti_google_threat_intelligence-correlation_detection_rule-pipeline.yml
```diff
@@ -0,0 +1,307 @@
...
+        for (def obj : ctx.threat.enrichments) {
+          if (obj.containsKey('matched') && obj.containsKey('indicator')) {
+            ctx.temp_fields = obj;
+            break;
+          }
+        }
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```
kcreddy, line 18: Add a `tag:` to the `script` processor to catch the error properly.
Same for other places as well.

packages/ti_google_threat_intelligence/elasticsearch/ingest_pipeline/ti_google_threat_intelligence-correlation_detection_rule-pipeline.yml
```diff
@@ -0,0 +1,307 @@
...
+      ignore_missing: true
+  - script:
+      lang: painless
+      source: |-
+        // Define the list of fields to retain
+        def fieldsToKeep = ["@timestamp"];
+
+        // Define regex patterns
+        def gtiPattern = /^temp_fields.*/;  // Matches keys starting with 'temp_fields'
+        def underscorePattern = /^_.*/;  // Matches keys starting with '_'
```
kcreddy, line 40:
```suggestion
        def gtiPattern = /^temp_fields/;  // Matches keys starting with 'temp_fields'
        def underscorePattern = /^_/;  // Matches keys starting with '_'
```

packages/ti_google_threat_intelligence/elasticsearch/ingest_pipeline/ti_google_threat_intelligence-correlation_detection_rule-pipeline.yml
```diff
@@ -0,0 +1,307 @@
...
+        for (def key : keys) {
+          // If the key is not in fieldsToKeep, does not match any Pattern, remove it
+          if (!fieldsToKeep.contains(key) && !(key =~ gtiPattern) && !(key =~ underscorePattern)) {
+            ctx.remove(key);
+          }
+        }
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```
kcreddy, line 55: Add `tag`

packages/ti_google_threat_intelligence/elasticsearch/transform/rule/transform.yml
```diff
@@ -0,0 +1,46 @@
...
+            kibana.alert.rule.tags: Google Threat Intelligence
+        - match_phrase:
+            kibana.alert.rule.category: Indicator Match Rule
+      must_not:
+        - match_phrase:
+            kibana.alert.rule.tags: IOC Stream
+dest:
+  index: "logs-ti_google_threat_intelligence.enriched_ioc-1"
+  aliases:
+    - alias: "logs-ti_google_threat_intelligence.enriched_ioc"
```
kcreddy, line 18: Can you share a sample document from `.internal.alerts-security.alerts-*` and its transformed (destination) version from `logs-ti_google_threat_intelligence.enriched_ioc`?

kcreddy: Also, please ensure docs inside `logs-ti_google_threat_intelligence.enriched_ioc` doesn't match [detection rules](https://github.com/elastic/detection-rules/blob/main/rules/threat_intel/threat_intel_indicator_match_url.toml#L124-L125) again, as it may trigger a cyclic loop.

packages/ti_google_threat_intelligence/manifest.yml
```diff
@@ -0,0 +1,98 @@
...
+              #    874A4YI7YUwOVsVAdbWtgp1d0zKcPRR+r2OdSbTAV5/gcS3jgBJ3i1BN34JuDVFw
+              #    3DeJSYT3nxy2Y56lLnxDeF8CUTUtVQx3CuGkRg1ouGAHpO/6OqOhwLLorEmxi7tA
+              #    H2O8mtT0poX5AnOAhzVy7QW0D/k4WaoLyckM5hUa6RtvgvLxOwA0U+VGurCDoctu
+              #    8F4QOgTAWyh8EZIwaKCliFRSynDpv3JTUwtfZkxo6K6nce1RhCWFAsMvDZL8Dgc0
+              #    yvgJ38BRsFOtkRuAGSf6ZUwTO8JJRRIFnpUzXflAnGivK9M13D5GEQMmIl6U9Pvk
+              #    sxSmbIUfc2SGJGCJD4I=
+              #    -----END CERTIFICATE-----
+owner:
+  github: elastic/security-service-integrations
+  type: community
```
kcreddy, line 98:
```suggestion
  type: elastic
```

I believe this should be `elastic` for all integrations created by Crest.

packages/ti_google_threat_intelligence/elasticsearch/transform/rule/transform.yml
```diff
@@ -0,0 +1,46 @@
...
+  query:
+    bool:
+      filter:
+        - match_phrase:
+            kibana.alert.rule.tags: Google Threat Intelligence
+        - match_phrase:
+            kibana.alert.rule.category: Indicator Match Rule
+      must_not:
+        - match_phrase:
+            kibana.alert.rule.tags: IOC Stream
```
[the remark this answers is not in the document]
kcreddy: Can you share an example doc that matches above `filters` and also contains `kibana.alert.rule.tags: IOC Stream`?

packages/ti_google_threat_intelligence/kibana/dashboard/ti_google_threat_intelligence-0b0fb6b4-d250-4e31-a56a-bb872e4c7c4a.json
```diff
@@ -0,0 +1,953 @@
...
+                                    "language": "kuery",
+                                    "query": ""
+                                }
+                            }
+                        },
+                        "description": "",
+                        "id": "",
+                        "params": {
+                            "fontSize": 12,
+                            "markdown": "[Threat Intelligence](#/dashboard/ti_google_threat_intelligence-9e8de699-a623-4a1b-9f63-7d641116f531), **Adversary Intelligence**, [Threat Feed Overview](#/dashboard/ti_google_threat_intelligence-95187e5c-b4a2-45ad-b6a4-d6ce68e1f43e)\n\nThis dashboard provides adversary intelligence insights into threat intelligence events detected in the user's environment by Google Threat Intelligence.\n\nIt categorizes detected threats into malware families, collections, campaigns, reports, software toolkits, threat actors, and vulnerabilities, using saved searches and uses metrics for count of unique adversaries.\n\n[**Integrations Page**](/app/integrations/detail/ti_google_threat_intelligence/overview)",
```
efd6, line 102: I'm not convince that this need commas between the targets.
```suggestion
                            "markdown": "[Threat Intelligence](#/dashboard/ti_google_threat_intelligence-9e8de699-a623-4a1b-9f63-7d641116f531) **Adversary Intelligence** [Threat Feed Overview](#/dashboard/ti_google_threat_intelligence-95187e5c-b4a2-45ad-b6a4-d6ce68e1f43e)\n\nThis dashboard provides adversary intelligence insights into threat intelligence events detected in the user's environment by Google Threat Intelligence.\n\nIt categorizes detected threats into malware families, collections, campaigns, reports, software toolkits, threat actors, and vulnerabilities, using saved searches and uses metrics for count of unique adversaries.\n\n[**Integrations Page**](/app/integrations/detail/ti_google_threat_intelligence/overview)",
```

packages/ti_google_threat_intelligence/kibana/dashboard/ti_google_threat_intelligence-9e8de699-a623-4a1b-9f63-7d641116f531.json
```diff
@@ -0,0 +1,3460 @@
...
.../ti_google_threat_intelligence-0b0fb6b4-d250-4e31-a56a-bb872e4c7c4a), [Threat Feed Overview](#/dashboard/ti_google_threat_intelligence-95187e5c-b4a2-45ad-b6a4-d6ce68e1f43e)\n\nThis dashboard provides a comprehensive view of threat intelligence events detected using Google Threat Intelligence data. It delivers insights into event distribution, threat analysis, and categorized breakdowns by threat feed type, severity, verdict, and threat score, including detected events over time.\n\nKey metrics include total detected events and unique files, URLs, domains, and IPs. Advanced filtering allows refinement by indicator type, threat score, verdict, severity, and threat feed type. Events are further analyzed by file types, country, tags, and top-level domains to identify patterns.\n\nNetwork and organizational insights highlight the top 10 autonomous system owners, autonomous system numbers, categories, and IPv4 network ranges. Additionally, it identifies the top 10 source event datasets and modules. Saved searches streamline investigations by summarizing detected file, IP, URL, and domain events.\n\n[**Integrations Page**](/app/integrations/detail/ti_google_threat_intelligence/overview)",
```
efd6, line 225:
```suggestion
                            "markdown": "**Threat Intelligence** [Adversary Intelligence](#/dashboard/ti_google_threat_intelligence-0b0fb6b4-d250-4e31-a56a-bb872e4c7c4a) [Threat Feed Overview](#/dashboard/ti_google_threat_intelligence-95187e5c-b4a2-45ad-b6a4-d6ce68e1f43e)\n\nThis dashboard provides a comprehensive view of threat intelligence events detected using Google Threat Intelligence data. It delivers insights into event distribution, threat analysis, and categorized breakdowns by threat feed type, severity, verdict, and threat score, including detected events over time.\n\nKey metrics include total detected events and unique files, URLs, domains, and IPs. Advanced filtering allows refinement by indicator type, threat score, verdict, severity, and threat feed type. Events are further analyzed by file types, country, tags, and top-level domains to identify patterns.\n\nNetwork and organizational insights highlight the top 10 autonomous system owners, autonomous system numbers, categories, and IPv4 network ranges. Additionally, it identifies the top 10 source event datasets and modules. Saved searches streamline investigations by summarizing detected file, IP, URL, and domain events.\n\n[**Integrations Page**](/app/integrations/detail/ti_google_threat_intelligence/overview)",
```

packages/ti_google_threat_intelligence/elasticsearch/transform/ip_ioc/transform.yml
```diff
@@ -0,0 +1,45 @@
+# Use of "*" to use all namespaces defined.
+source:
+  index:
+    - "logs-ti_google_threat_intelligence.*"
+  query:
+    bool:
+      filter:
+        - match_phrase:
+            threat.indicator.type: ipv4-addr
```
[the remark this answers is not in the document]
niraj-crest: Yes, we’ve checked both the [documentation ](https://gtidocs.virustotal.com/reference/ip-object#:~:text=timestamp%3E%0A%20%20%20%20%7D%2C-,%22id%22%3A%20%22%3Cipv4%3E%22%2C,-%22links%22%3A)and the live data, currently, all indicators are coming as `ipv4-addr`. There are no `ipv6-addr` indicators in any of the feed types.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,88 @@
...
+        "events": (
+          has(body.iocs) && size(body.iocs) > 0 ?
+            body.iocs.map(e, {
+              "message": e.encode_json(),
+            })
+          :
+            [{"message": "retry"}]
+        ),
+        "cursor": {
+          "last_timestamp": (state.start_time.parse_time("2006010215") + duration("1h")).format("2006010215"),
```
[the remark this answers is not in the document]
kcreddy: yeah, but it looks like if there are more than 4000 indicators inside the 1 hour duration, there is no way to ingest them. Maybe its just anticipated maximum or a limitation in the API that it doesn't support pagination.

packages/ti_google_threat_intelligence/elasticsearch/transform/rule/transform.yml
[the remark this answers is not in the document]
niraj-crest: We have two separate rule transforms one for the IOC stream in this [PR](https://github.com/elastic/integrations/pull/13449) and another for the Threat List. The Threat List includes 14 data streams (e.g., `cryptominer`, `iot`, `infostealer`, `first_stage_delivery_vector`), covered in this PR and others like [PR1](https://github.com/elastic/integrations/pull/13236) and [PR2](https://github.com/elastic/integrations/pull/13190).
These transforms serve different purposes and follow different log schemas. The condition
```yaml
must_not:
  - match_phrase:
      kibana.alert.rule.tags: IOC Stream
```

is added to ensure this transform processes only Threat List data and excludes alerts generated by the IOC stream transform.
Here’s a sample [document](https://jsonblob.com/1377898078895529984) that matches this filter:
As shown, this document doesn't contain `IOC Stream` in `kibana.alert.rule.tags`. In contrast, alerts generated by IOC stream rules [link](https://github.com/elastic/integrations/pull/13449/files#diff-3f10a70718a8281c478853a45e62586b9162817793fd6c670d89d8a677275b4b) will include this tag, which is why they are excluded by this condition.

packages/ti_google_threat_intelligence/elasticsearch/transform/rule/transform.yml
niraj-crest: Please find the sample documents:
**[.internal.alerts-security.alerts-*](https://jsonblob.com/1377898078895529984)**
**[logs-ti_google_threat_intelligence.enriched_ioc](https://jsonblob.com/1377898664655249408)**
Regarding the detection rules you mentioned, they include the condition `not labels.is_ioc_transform_source:"true"`. Since our enriched IoC index doesn’t include the `labels.is_ioc_transform_source` field, these documents won’t match the rules, so there’s no risk of a cyclic loop.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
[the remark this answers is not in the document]
niraj-crest: Thanks for the follow-up. We’ve rechecked both the live API and the documentation, and can confirm that there’s no way to paginate within a specific hourly window. This appears to be a behavior of the API itself it doesn’t provide a mechanism to fetch additional pages for a given hour.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,88 @@
...
+        },
+        "want_more": (state.start_time.parse_time("2006010215") + duration("1h")).format("2006010215") < now.format("2006010215")
```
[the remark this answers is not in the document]
niraj-crest: Thanks for pointing this out. I agree that comparing times directly is generally more efficient and reliable than comparing formatted strings.

In this case, the API only accepts time as a path parameter formatted to the hour. For example, if the current time is `2025-06-03 08:01:00` and `state.start_time.parse_time("2006010215")` gives `2025-06-03 08:00:00`, we want `want_more` to be `false`. This is because the API returns hourly data with a one-hour delay, and calling it with the current hour minus one would result in an error. We've confirmed this behavior in the live API.

To handle this properly, I’ve used the following logic:

```
(state.start_time.parse_time("2006010215") + duration("1h")) < (now.format("2006010215")).parse_time("2006010215")
```
This ensures the comparison respects the API's delay. 

Please, Let me know if this version works for you.

packages/ti_google_threat_intelligence/manifest.yml
niraj-crest: This integration was specifically requested by `Google` and is not part of the standard contract with Crest. As such, we are classifying it under the `community` category.

kcreddy: Thats good to know. Yeah it should remain as `community` in that case.

niraj-crest:
Hello @kcreddy / @efd6  /@andrewkroh 
Please let me know if we have any other review comments.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,88 @@
...
+{{#if query}}
+  query: {{query}}
+{{/if}}
+redact:
+  fields:
+    - access_token
+program: |
+  state.with({
+    "start_time": state.?cursor.last_timestamp.orValue((now - duration(state.initial_interval)).format("2006010215"))
+  }).as(state, state.with(
```
efd6:
```suggestion
  state.?cursor.last_timestamp.orValue((now - duration(state.initial_interval)).format("2006010215")).as(start_time,
```
since this is overwritten for every eval it doesn't need to be put in `state`. This will need bracket fixups.

Though, I would also reconsider doing the format here. Instead, that way you don't need to reparse each time you make an adjustment.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,88 @@
...
+    request(
+      "GET",
+      state.url.trim_right("/") + "/api/v3/threat_lists/cryptominer/" + state.start_time + "?" + {
```
efd6:
```suggestion
      state.url.trim_right("/") + "/api/v3/threat_lists/cryptominer/" + start_time + "?" + {
```

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,88 @@
...
+        "events": (
+          has(body.iocs) && size(body.iocs) > 0 ?
+            body.iocs.map(e, {
+              "message": e.encode_json(),
+            })
+          :
+            [{"message": "retry"}]
+        ),
+        "cursor": {
+          "last_timestamp": (state.start_time.parse_time("2006010215") + duration("1h")).format("2006010215"),
```
efd6:
```suggestion
          "last_timestamp": (start_time.parse_time("2006010215") + duration("1h")).format("2006010215"),
```

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,88 @@
...
+        },
+        "want_more": (state.start_time.parse_time("2006010215") + duration("1h")) < (now.format("2006010215")).parse_time("2006010215")
```
efd6:
```suggestion
        "want_more": (start_time.parse_time("2006010215") + duration("1h")) < (now.format("2006010215")).parse_time("2006010215")
```

packages/ti_google_threat_intelligence/data_stream/first_stage_delivery_vectors/agent/stream/cel.yml.hbs
efd6, line 1: Same comments as above.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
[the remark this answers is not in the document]
efd6: With the comments in my most recent review, this could become `(start_time + duration("1h")) < now`.

packages/ti_google_threat_intelligence/data_stream/iot/fields/fields.yml
```diff
@@ -0,0 +1,264 @@
...
+                  description: The name of the security engine that performed the analysis.
+                - name: result
+                  type: keyword
+                  description: The outcome of the analysis performed by the security engine.
+            - name: last_analysis_stats
+              type: group
+              fields:
+                - name: harmless
+                  type: long
+                  description: Number of reports saying that is harmless.
```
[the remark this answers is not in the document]
efd6: That is unfortunate.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,88 @@
...
+  state.with({
+    "start_time": state.?cursor.last_timestamp.orValue((now - duration(state.initial_interval)).format("2006010215"))
+  }).as(state, state.with(
+    request(
+      "GET",
+      state.url.trim_right("/") + "/api/v3/threat_lists/cryptominer/" + state.start_time + "?" + {
+        ?"query": has(state.query) ? optional.of([state.query]) : optional.none(),
+        "limit": ["4000"],
+        "x-tool": ["Elastic"],
+        "User-Agent": ["v0.1.0"]
```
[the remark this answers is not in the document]
efd6: Rather than in the changelog, I would suggest a comment above the `version` field in the manifest and a comment on this field saying that it should match the package version.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
niraj-crest: Thank you for the suggestion!
We’ve updated the code to use `start_time` directly as recommended, which definitely helped simplify things.

Regarding your additional point about reconsidering the formatting, we explored that approach but ran into a timing issue due to the API's behavior. Specifically, the API enforces a **1-hour delay**, and if we don’t round precisely to the hour, we risk calling a future window too early.

For example, if the current time is `10:21` and we start with an initial interval of `2h`, we’ll fetch data for `08` and `09`. When checking whether to proceed to the next hour, `(09:21 + 1h)` becomes `10:21`, which might still be less than now (say `10:22`) by the time the check happens. This leads us to attempt a `10:00` call prematurely and receive an error.

To avoid this, we’ve kept the strict hour-to-hour comparison using formatted values.
Please let us know if you have any further thoughts.

niraj-crest:
Hello @kcreddy / @efd6 ,
We’ve addressed all the review comments.
Kindly let us know if you have any additional suggestions. If everything looks good to you, we’d appreciate it if we could proceed with merging this PR.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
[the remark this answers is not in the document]
kcreddy: Thanks for checking again.

packages/ti_google_threat_intelligence/elasticsearch/transform/rule/transform.yml
```diff
@@ -0,0 +1,46 @@
+# Use of "*" to use all namespaces defined.
+source:
+  index:
+    - ".internal.alerts-security.alerts-*"
```
kcreddy: Can you check if you can query on alert index alias instead? `.alerts-security.alerts-*`

It seems to be the suggested way over alert indices. https://www.elastic.co/docs/solutions/security/detect-and-alert/query-alert-indices

packages/ti_google_threat_intelligence/elasticsearch/transform/rule/transform.yml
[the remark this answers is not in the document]
kcreddy: This is indeed confusing as this PR change is dependent on future PRs. We should avoid this if possible in the future.

What is the difference between IOCs ingested from this PR's API i..e., [Get Hourly Threat List](https://gtidocs.virustotal.com/reference/get-hourly-threat-list) vs the [`/ioc_stream` API](https://gtidocs.virustotal.com/reference/get-objects-from-the-ioc-stream) thats added in https://github.com/elastic/integrations/pull/13449?
Is one API better than the other, if so why don't we add only the better one?

packages/ti_google_threat_intelligence/elasticsearch/transform/ip_ioc/fields/iot_fields.yml
kcreddy: Please kindly make a note of all these after items and create a follow up PR. Thanks!

packages/ti_google_threat_intelligence/elasticsearch/transform/domain_ioc/fields/ecs.yml
```diff
@@ -0,0 +1,32 @@
+- external: ecs
+  name: ecs.version
+- external: ecs
+  name: event.category
+- external: ecs
+  name: event.created
+- external: ecs
+  name: event.kind
+- external: ecs
```
kcreddy, line 11: Missing `event.ingested`. Please add it.
Can cause issue like this: https://github.com/elastic/integrations/issues/14080

packages/ti_google_threat_intelligence/elasticsearch/transform/url_ioc/fields/ecs.yml
kcreddy: In the pipeline tests, you also have top-level `url` fields. Please add them as well.

packages/ti_google_threat_intelligence/elasticsearch/transform/ip_ioc/transform.yml
kcreddy: > After adding the pipeline, the transform turns Healthy as expected.

This behaviour is very strange. `run_as_kibana_system` shouldn't impact the transform health from a pipeline not being present, especially when we don't even have a pipeline by default https://github.com/elastic/integrations/pull/13189#discussion_r2071530420.

If its not too much, could you please kindly add a video capturing this behaviour?

packages/ti_google_threat_intelligence/elasticsearch/transform/rule/transform.yml
kcreddy: Thanks.
If you also have [`event.kind: enrichment`](https://www.elastic.co/docs/reference/ecs/ecs-allowed-values-event-kind#ecs-event-kind-enrichment) inside `logs-ti_google_threat_intelligence.enriched_ioc`, you should change it to something else. Because the detection rules [matches it](https://github.com/elastic/detection-rules/blob/main/rules/threat_intel/threat_intel_indicator_match_url.toml#L162) and also the document here itself isn't supposed to be used to enrich other events.

packages/ti_google_threat_intelligence/elasticsearch/transform/rule/transform.yml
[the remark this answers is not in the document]
niraj-crest: Hello @kcreddy ,

The key difference between the Threat List and IOC Stream APIs lies in their purpose and the nature of the data they provide:

**Threat List** delivers hourly threat logs across various categories defined by GTI (e.g., cryptominer, infostealer, etc.).

**IOC Stream**, on the other hand, provides IOC information specifically related to threat groups the user has subscribed to.

Each follows a different schema and serves a distinct use case, which is why both are required for this integration it's not a matter of one being better than the other.

**Regarding the PR structure:** all this PRs( https://github.com/elastic/integrations/pull/13189, https://github.com/elastic/integrations/pull/13190, https://github.com/elastic/integrations/pull/13236, https://github.com/elastic/integrations/pull/13449 ) are part of the same GTI integration. We've broken them down to make review and collaboration easier, considering the overall package is quite large. Once this PR is merged, we'll ensure the feedback here is also reflected and addressed in the subsequent PRs.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,88 @@
...
+  (
+    state.?cursor.last_timestamp.orValue((now - duration(state.initial_interval)).format("2006010215"))
+  ).as(start_time, state.with(
+    request(
+      "GET",
+      state.url.trim_right("/") + "/api/v3/threat_lists/cryptominer/" + start_time + "?" + {
+        ?"query": has(state.query) ? optional.of([state.query]) : optional.none(),
+        "limit": ["4000"],
+        "x-tool": ["Elastic"],
+        "User-Agent": ["v0.1.0"],  // Include integration version here; must match the 'version' in package level manifest.yml
```
efd6:
```suggestion
        "User-Agent": ["v0.1.0"], // Keep this in sync with 'version' in package level manifest.yml.
```
Same change elsewhere that this comment exists.

packages/ti_google_threat_intelligence/manifest.yml
```diff
@@ -0,0 +1,100 @@
+format_version: 3.3.1
+name: ti_google_threat_intelligence
+title: Google Threat Intelligence
+# This version must match the User-Agent version used in CEL code.
+# Remember to update the User-Agent in CEL code when changing this version.
```
efd6, line 5: :white_check_mark:

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
efd6: OK. That is also unfortunate. What we'd really like is a [timestamp round function](https://pkg.go.dev/time#Time.Round). https://github.com/elastic/mito/issues/92

packages/ti_google_threat_intelligence/elasticsearch/transform/ip_ioc/transform.yml
niraj-crest: Hello @kcreddy,
Please Find the Drive Link: https://drive.google.com/file/d/1_Z8kbECp890fuT2ULQtxyI_0Do4qEAyH/view?usp=drive_link

kcreddy: Thanks @niraj-crest, really appreciate!

Can you please check if it resolves by changing the source indices to match this generic pattern: `logs-ti_*.*-*`?
Possibly:
```yaml
source:
  index:
    - "logs-ti_google_threat_intelligence.*-*"
```

niraj-crest: Thanks for the suggestion.
Unfortunately, we're still seeing the same behavior with the rule transform. All other 4 transforms have worked fine.

image

kcreddy: Yes the `rule` transform still requires `run_as_kibana_system: false`
https://github.com/elastic/integrations/pull/13189#discussion_r2112389395

niraj-crest: By following the suggested pattern and removing `run_as_kibana_system` from non-rule transforms, the reauthorization warning issue has been resolved.

However, there's an important consideration here. As discussed in [this comment](https://github.com/elastic/integrations/pull/13189/files/595a25b8cf2f290ab278e11d05c4d38e13fd5986#r2142637521), this PR is part of the Google Threat Intelligence integration and this integration will includes both the threat list and IOC stream along with their assets. Therefore, to make the transform logic more robust and avoid processing enriched logs or errors, the following query has been added:
```yml
source:
  index:
    - "logs-ti_google_threat_intelligence.*-*"
  query:
    bool:
      filter:
        - match_phrase:
            threat.indicator.type: domain-name
        - match_phrase:
            event.dataset: ti_google_threat_intelligence.ioc_stream            
      must_not:
        - exists:
            field: error.message
        - match_phrase:
            event.dataset: ti_google_threat_intelligence.enriched_ioc
        - match_phrase:
            event.dataset: ti_google_threat_intelligence.enriched_ioc_stream

```

This works well in the live environment.
However, system tests are failing, likely because query conditions are not applied or evaluated during the system test.

Please find example of system test result for `iot` data-stream after configuring suggested changes.

```
--- Test results for package: ti_google_threat_intelligence - START ---
╭───────────────────────────────┬─────────────┬───────────┬───────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬─────────────────╮
│ PACKAGE                       │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT                                                                                                              │    TIME ELAPSED │
├───────────────────────────────┼─────────────┼───────────┼───────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────┤
│ ti_google_threat_intelligence │ iot         │ system    │ default   │ ERROR: no documents found in preview for transform "logs-ti_google_threat_intelligence.domain_ioc_st-default-0.1.0" │ 2m10.257972784s │
╰───────────────────────────────┴─────────────┴───────────┴───────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┴─────────────────╯
--- Test results for package: ti_google_threat_intelligence - END   ---
Done
Error: failed to process results: one or more test cases failed
```

kcreddy: Do you mean the system tests were working fine when `run_as_kibana_system: false` but not otherwise?

> However, system tests are failing, likely because query conditions are not applied or evaluated during the system test.

This needs to be debugged further. Please open an issue in https://github.com/elastic/elastic-package with screenshots or recording.
To unblock you, add a link to that issue in this transform file as a comment indicating the reason for `run_as_kibana_system: false`.

kcreddy (commented):
Only waiting on 1 outstanding comment: https://github.com/elastic/integrations/pull/13189#discussion_r2151827711

packages/ti_google_threat_intelligence/elasticsearch/transform/ip_ioc/transform.yml
niraj-crest: Hello @kcreddy ,

As requested, we’ve investigated this further, and here’s our detailed analysis:

Initially, after removing run_as_kibana_system: false, the system tests passed. However, we encountered a reauthorization warning (as discussed in [this comment](https://github.com/elastic/integrations/pull/13189#discussion_r2144306578)), which could negatively impact the user experience. To avoid this, we applied a broader index pattern:

```yml
source:
  index:
    - "logs-ti_google_threat_intelligence.*-*"
```
While this resolved the warning, the current transform is intended to work only with specific threat list data. To correctly scope the data source and maintain the intended behavior, we introduced a query filter:

```yml
source:
  index:
    - "logs-ti_google_threat_intelligence.*-*"
  query:
    bool:
      filter:
        - match_phrase:
            threat.indicator.type: domain-name
      must_not:
        - exists:
            field: error.message
        - match_phrase:
            event.dataset: ti_google_threat_intelligence.ioc_stream
        - match_phrase:
            event.dataset: ti_google_threat_intelligence.enriched_ioc
        - match_phrase:
            event.dataset: ti_google_threat_intelligence.enriched_ioc_stream
```

However, it appears that these query conditions are not being applied during the system test. As result system test is failing.
```
--- Test results for package: ti_google_threat_intelligence - START ---
╭───────────────────────────────┬─────────────┬───────────┬───────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬─────────────────╮
│ PACKAGE                       │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT                                                                                                              │    TIME ELAPSED │
├───────────────────────────────┼─────────────┼───────────┼───────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────┤
│ ti_google_threat_intelligence │ iot         │ system    │ default   │ ERROR: no documents found in preview for transform "logs-ti_google_threat_intelligence.domain_ioc_st-default-0.1.0" │ 2m10.257972784s │
╰───────────────────────────────┴─────────────┴───────────┴───────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┴─────────────────╯
--- Test results for package: ti_google_threat_intelligence - END   ---
Done
Error: failed to process results: one or more test cases failed
```
Based on this, it seems the root cause of the system test failure is that the query block is not  evaluated when the system test runs.

Would you recommend we open an issue in the [elastic-package](https://github.com/elastic/elastic-package) regarding this?

kcreddy: > Would you recommend we open an issue in the [elastic-package](https://github.com/elastic/elastic-package) regarding this?

@niraj-crest, yes please create the issue there with all details. Especially confirming that you are receiving data in the destination indices correctly, but the system tests are failing with `ERROR: no documents found`

kcreddy: Once the issue is created add a comment here in these transforms linking the issue.
I can approve this PR right after. Sorry for the very long wait.

niraj-crest: I have updated the PR. Could you please review the changes? If everything looks good, we can proceed with merging the PR.

niraj-crest: We have raised an issue in elastic package. https://github.com/elastic/elastic-package/issues/2676

kcreddy: Can you add comment to all transforms (except rule) - mentioning this issue and reason for choosing `run_as_kibana_system:false`

niraj-crest: Hello @kcreddy,
We have added suggested comment can you please review it.

kcreddy: @niraj-crest, thanks for adding comment. I will approve to unblock you. 
This fix should be done before making the integration GA.

When you have an outstanding list of pending items (that we discussed), please create a public issue and link to this PR. It will be the items required to make the integration GA. For example: https://github.com/elastic/integrations/pull/13189#discussion_r2135449116

niraj-crest: hello @kcreddy,
As suggested we have raised 2 issues one for https://github.com/elastic/integrations/pull/13189#discussion_r2135449116 and one for making package GA.

- https://github.com/elastic/integrations/issues/14286
- https://github.com/elastic/integrations/issues/14287

We will appreciate if you approve and merge this PR.

kcreddy (approved):
https://github.com/elastic/integrations/issues/14287 tracks list of pending issues for making the integration GA.

PR #13200, opened 2025-03-20
checkpoint_harmony_endpoint: fix assignment of endTime query parameter

chemamartinez:
## Proposed commit message

Fixing a CEL condition that made the `endTime` query parameter being null when `state.cursor.next_endTime` is also null, leading to the following error:
```
 {"success":false,"error":{"status":400,"name":"Bad Request","details":["endTime should not be empty"]}}
```

> [!NOTE]

## Checklist

### DIFF
--- a/packages/checkpoint_harmony_endpoint/data_stream/antibot/agent/stream/cel.yml.hbs
+++ b/packages/checkpoint_harmony_endpoint/data_stream/antibot/agent/stream/cel.yml.hbs
@@ -69,9 +69,10 @@ program: |
   			"startTime": state.?cursor.next_startTime.orValue(
   				timestamp(now() - duration(state.initial_interval)).format(time_layout.RFC3339)
   			),
-  			"endTime": state.?cursor.next_endTime.orValue(
-  				timestamp(now() - duration("1m")).format(time_layout.RFC3339)
-  			),
+  			"endTime": state.?cursor.next_endTime.orValue(null) == null ?
+  				optional.of(timestamp(now() - duration("1m")).format(time_layout.RFC3339))
+  			:
+  				optional.of(state.cursor.next_endTime),
   		}.as(timeframe,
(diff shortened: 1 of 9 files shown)


ShourieG (approved):
LGTM

packages/checkpoint_harmony_endpoint/data_stream/antibot/agent/stream/cel.yml.hbs
```diff
@@ -69,9 +69,10 @@ program: |
   			"startTime": state.?cursor.next_startTime.orValue(
   				timestamp(now() - duration(state.initial_interval)).format(time_layout.RFC3339)
   			),
-  			"endTime": state.?cursor.next_endTime.orValue(
-  				timestamp(now() - duration("1m")).format(time_layout.RFC3339)
-  			),
+  			"endTime": state.?cursor.next_endTime.orValue(null) == null ?
+  				optional.of(timestamp(now() - duration("1m")).format(time_layout.RFC3339))
+  			:
+  				optional.of(state.cursor.next_endTime),
```
kcreddy, line 75:
```suggestion
  				timestamp(now() - duration("1m")).format(time_layout.RFC3339)
  			:
  				state.cursor.next_endTime,
```
Do you need `optional.of(`?
It doesn't seem to be defined as optional var with `?"endTime"`

ShourieG: good catch @kcreddy, yes **optional.of** makes sense when the existence of **endTime** it self is also optional

chemamartinez: True, not needed.

kcreddy (approved):

PR #13238, opened 2025-03-21
ti_anomali: Fix lower bound query parameter in Threatstream API.

kcreddy:
## Proposed commit message
```
API doc doesn't have any reference to the 
query parameter search_after, which is supposed to handle
the lower bound on searching indicators. Without this, the 
integration always fetches all indicators every call because 
modified_ts__lt : now-1m. There are cases Anomali revoked 
API keys because of this requesting for proper lower bounding 
the requests.

Based on API doc, this PR updates the query parameter to be
update_id__gt which lower bounds based on last received 
update_id.
```

> [!NOTE]
> Not tested against live API, but only based on API doc. 
> DM me for the API doc.

## Checklist

Updated system tests pass. Not tested against live API.
```
--- Test results for package: ti_anomali - START ---
╭────────────┬──────────────┬───────────┬───────────┬────────┬───────────────╮
│ PACKAGE    │ DATA STREAM  │ TEST TYPE │ TEST NAME │ RESULT │  TIME ELAPSED │
├────────────┼──────────────┼───────────┼───────────┼────────┼───────────────┤
│ ti_anomali │ intelligence │ system    │ default   │ PASS   │ 44.176397125s │
╰────────────┴──────────────┴───────────┴───────────┴────────┴───────────────╯
--- Test results for package: ti_anomali - END   ---
Done
```

### DIFF
--- a/packages/ti_anomali/_dev/deploy/docker/files/intelligence-api-config.yml
+++ b/packages/ti_anomali/_dev/deploy/docker/files/intelligence-api-config.yml
@@ -7,7 +7,7 @@ rules:
       order_by: update_id
       modified_ts__lt: "{modified_ts__lt:.*}"
       limit: "3"
-      search_after: "100000006"
+      update_id__gt: "100000006"
     responses:
       - status_code: 200
         body: |-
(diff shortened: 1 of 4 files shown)


kcreddy:
> In the commit message, I suggest removing this part:
> 
> > Without this, the integration always fetches all indicators every call based on
> > "modified_ts__lt: now-1m".
> 
> Since that doesn't seem to be the case.
> 
> The `search_after` condition works, but using `update_id__gt` is recommended and probably performs better.

Done. Thanks @chrisberkhout !

PR #13301, opened 2025-03-25
[Enhancement] Tenable sc vulnerability mitigated added as new datastream

frederikb96:
## Proposed commit message

Added a new datastream to also query mitigated vulnerabilities. Tenable contains two independent databases, one for open and one for closed/mitigated vulnerabilities. This enhancement adds a datastream to query the database for mitigated vulnerabilities as well. The new datastream is a simple copy of the already available datastream for open vulnerabilities. I adjusted the query and renamed everything accordingly. It was necessary to create a new datastream since two independent databases are used to store the vulnerability data in tenable as explained in the [official tenable documentation](https://developer.tenable.com/docs/tenablesc-integrations).

## Checklist

The new datastream contains an example log message which can be used to verify that the integration works as expected. It is completely analog to the already existing datastream for open vulnerabilities.


packages/tenable_sc/data_stream/vulnerability_patched/_dev/test/pipeline/test-common-config.yml
```diff
@@ -0,0 +1,5 @@
+dynamic_fields:
+  "event.ingested": ".*"
```
efd6, line 2:
```suggestion
```
This is not necessary.

packages/tenable_sc/_dev/deploy/docker/files/config.yml
```diff
@@ -31,3 +31,12 @@ rules:
...
..."threat_recency\",\"name\":\"Threat Recency\",\"type\":\"string\",\"value\":\"No recorded events\"},{\"id\":\"threat_sources_last_28\",\"name\":\"Threat Sources\",\"type\":\"string\",\"value\":\"No recorded events\"}]","baseScore":"0.0","temporalScore":"","cvssVector":"AV:L/AC:L/Au:N/C:N/I:N/A:N","cvssV3BaseScore":"0.0","cvssV3TemporalScore":"","cvssV3Vector":"AV:L/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:N","cpe":"","vulnPubDate":"788961600","patchPubDate":"-1","pluginPubDate":"933508800","pluginModDate":"1570190400","checkType":"remote","version":"1.48","cve":"CVE-1999-0524","bid":"","xref":"CWE #200","pluginText":"<plugin_output>The remote clock is synchronized with the local clock.\n</plugin_output>","dnsName":"_gateway.lxd","macAddress":"00:16:3e:a1:12:f7","netbiosName":"","operatingSystem":"Linux Kernel 2.6","ips":"10.238.64.1","hostUniqueness":"repositoryID,ip,dnsName","uniqueness":"repositoryID,ip,dnsName","family":{"id":"30","name":"General","type":"active"},"repository":{"id":"1","name":"Live","description":"","sciID":"1","dataFormat":"IPv4"},"pluginInfo":"10114 (0/1) ICMP Timestamp Request Remote Date Disclosure"}]},"error_code":0,"error_msg":"","warnings":[],"timestamp":1634709360}
```
efd6, line 41: Please pretty print this and use `minify_json` helper (example [here](https://github.com/elastic/integrations/blob/bfd04a448cf4f12101e9152655047567030a852a/packages/abnormal_security/_dev/deploy/docker/files/config.yml#L60-L76))

packages/tenable_sc/data_stream/vulnerability_patched/agent/stream/httpjson.yml.hbs
efd6, line 1: Since this is a new data stream, I think it would be worth starting out with CEL instead of HTTP JSON.

packages/tenable_sc/data_stream/vulnerability_patched/fields/fields.yml
```diff
@@ -0,0 +1,262 @@
...
+      type: group
+      fields:
+        - name: data_format
+          type: keyword
+          description: |
+            The data format of the repository.
+        - name: description
+          type: keyword
+          description: |
+            The description of the repository.
```
efd6, line 188: Do we have cases where this is not empty? How large can it be and should it be `match_only_text` instead of `keyword`.

The `.severity.description` below looks like it is a keywordy type.

packages/tenable_sc/data_stream/vulnerability_patched/fields/fields.yml
```diff
@@ -0,0 +1,262 @@
+- name: tenable_sc.vulnerability
+  type: group
+  fields:
+    - name: accept_risk
+      type: keyword
+      description: |
+        N/A.
```
efd6, line 7:
```suggestion
```
or fill the description.

efd6 (commented):
Thanks for the contribution

frederikb96:
@efd6 Thanks for the swift reply and the suggestions. Since the new data stream is analog to the existing one, I am wondering if I should adjust the existing one as well according to your suggestions. The suggestions are all also applicable to the existing data streams of this integration :)

Should I put it all into this pull request?

About CEL: I decided to keep in analog to the existing data stream of the open vulnerabilities, since it's exactly the same code, apart from the fact that it is querying the other tenable database for mitigated vulnerabilities. I would suggest that it makes more sense to create another pull request where the code can be rewritten to CEL for both data streams at once after this one is merged. Otherwise, we have two different implementations of basically the same query.

packages/tenable_sc/data_stream/vulnerability_patched/fields/fields.yml
frederikb96: I checked and the repository description has never been non-empty in any tenable documents we have ingested so far.

However, I would vote to just keep this as is since this might break the data field mapping when changing it to match_only_text now for some users. And just adjusting this for the data stream with the mitigated vulnerabilities would be bad, then we had a different mapping here in the two vulnerability data streams and a data view on both would complain.

packages/tenable_sc/data_stream/vulnerability_patched/fields/fields.yml
frederikb96: removed the description

packages/tenable_sc/data_stream/vulnerability_patched/agent/stream/httpjson.yml.hbs
frederikb96: As said below, I would keep it in sync with the other existing data stream, since this new one is a almost 1:1 copy of it, just addresing the other database in tenable. Would rather open another PR maybe later and adjusting both to CEL at once.

packages/tenable_sc/_dev/deploy/docker/files/config.yml
frederikb96: Thanks for the example, nice! I fixed the code for all datastreams to keep it in sync with the existing once.

frederikb96:
@elastic/security-service-integrations

Anything else that needs to be done, or can it be merged? :)

packages/tenable_sc/data_stream/vulnerability_patched/fields/fields.yml
efd6: No worries. Thanks for looking into that for me.

packages/tenable_sc/data_stream/vulnerability_patched/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,594 @@
...
+  - set:
+      field: event.kind
+      value: pipeline_error
+  - append:
+      field: tags
+      value: preserve_original_event
+      allow_duplicates: false
+  - append:
+      field: error.message
+      value: '{{{ _ingest.on_failure_message }}}'
```
efd6, line 594:
```suggestion
      value: >-
        Processor '{{{ _ingest.on_failure_processor_type }}}'
        {{{#_ingest.on_failure_processor_tag}}}with tag '{{{ _ingest.on_failure_processor_tag }}}'
        {{{/_ingest.on_failure_processor_tag}}}failed with message '{{{ _ingest.on_failure_message }}}'
```

packages/tenable_sc/data_stream/vulnerability_patched/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,594 @@
...
+      formats:
+        - UNIX
+      on_failure:
+        - append:
+            field: error.message
+            value: '{{{_ingest.on_failure_message}}}'
+  - script:
+      lang: painless
+      tag: set_vulnerability_age
+      if: ctx.tenable_sc?.vulnerability?.first_seen != null && ctx.tenable_sc?.vulnerability?.last_seen != null
```
efd6, line 286:
```suggestion
      if: ctx.tenable_sc?.vulnerability?.first_seen != null && ctx.tenable_sc.vulnerability.last_seen != null
```

efd6 (commented):
I'm torn about the CEL/HTTP JSON question. I think you are correct that having them all the same, given the similarity of the data streams makes sense, but a future rewrite is more work than a new data stream write in CEL and (currently) has more user disruption baked in. We are working on mechanisms to reduce or eliminate this friction, but it doesn't exist yet.

I'll think some more. In the mean time, there were error handling changes that should be made, so I've also asked for a nit to be fixed as well. If you want to apply the suggested changes here in the other data streams where applicable, I would approve that.

PR #13312, opened 2025-03-26
[Cloud Security][Wiz] add vulnerability event.id to use ecs mapping

alexreal1314:
## add vulnerability event.id to use ecs mapping

Update event.id field in latest cdr vulnerabilities to use ECS mapping

### DIFF
--- a/packages/wiz/manifest.yml
+++ b/packages/wiz/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.2.3
 name: wiz
 title: Wiz
-version: "2.9.3"
+version: "2.10.0"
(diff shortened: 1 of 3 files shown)


packages/wiz/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 3.2.3
 name: wiz
 title: Wiz
-version: "2.9.3"
+version: "2.9.4"
```
kcreddy, line 4:
```suggestion
version: "2.10.0"
```

kcreddy (approved):

PR #13325, opened 2025-03-27
crowdstrike: improve ingest pipeline performance

efd6:
## Proposed commit message

```
crowdstrike: improve ingest pipeline performance

This improves the performance of the null cleaning script in the pipeline's
postamble, and the command line arg splitter in the body of the pipeline.

The empty field change reduces the number of allocations for iterators by
half and improves the readability of the code.

             │  old.bench  │           new.bench           │
             │   sec/op    │    sec/op     vs base         │
remove-nulls   4.000µ ± 0%   2.000µ ± 50%  -50.00% (n=100)

The command line splitting script was consuming a significant proportion
of the pipeline's total CPU use. The original code was based on Go code
used in libbeat, on the assumption that the allocation and GC behaviour
would be similar between the two languages; this is not the case. While
slicing strings in Go is extremely cheap (likely not involving an
allocation at all in most cases since the string header can be retained
on the stack), this is not the case in Java where the substring method
does allocate. So to address this, instead of consuming the string and
biting off chunks, only take fragments when they are known to be a
complete argument.

Before the change, using the elastic-package benchmarking tool with
munging to allow benchstat to consume the data, the performance was

                   │ cmd_old.bench │
                   │    sec/op     │
split-command-line     10.31µ ± 0%

After the change, the time taken is not great enough to be emitted by
the benchmark tool. This would place it in the approximately
sub-microsecond range.
```

> [!NOTE]
> Great appreciation to @joegallo for looking through this with me.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
@@ -2840,24 +2840,24 @@ processors:
       description: This script processor iterates over the whole document to remove fields with null values.
       source: |
         void handleMap(Map map) {
-          for (def x : map.values()) {
-            if (x instanceof Map) {
-                handleMap(x);
-            } else if (x instanceof List) {
-                handleList(x);
+          map.values().removeIf(v -> {
+            if (v instanceof Map) {
+                handleMap(v);
+            } else if (v instanceof List) {
+                handleList(v);
             }
-          }
-          map.values().removeIf(v -> v == null || v == '' || v == '-' || v == 'none' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0));
+            return v == null || v == '' || v == '-' || v == 'none' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0)
+          });
         }
         void handleList(List list) {
-          for (def x : list) {
-              if (x instanceof Map) {
-                  handleMap(x);
-              } else if (x instanceof List) {
-                  handleList(x);
-              }
-          }
-          list.removeIf(v -> v == null || v == '' || v == '-' || v == 'none' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0));
+          list.removeIf(v -> {
+            if (v instanceof Map) {
+                handleMap(v);
+            } else if (v instanceof List) {
+                handleList(v);
+            }
+            return v == null || v == '' || v == '-' || v == 'none' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0)
+          });
         }
         handleMap(ctx);
 on_failure:
--- a/packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
@@ -1735,19 +1735,19 @@ processors:
             return b;
         }
 
-        // readNextArg splits command line string cmd into next
-        // argument and command line remainder.
-        def readNextArg(String cmd) {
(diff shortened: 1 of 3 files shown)


packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1735,19 +1735,23 @@ processors:
...
-            for (; cmd.length() > 0; cmd = cmd.substring(1)) {
-                def c = cmd.charAt(0);
+            for (; offset < line.length(); offset++) {
+                def c = line.charAt(offset);
                 if (c == (char)' ' || c == (char)0x09) {
                     if (!inquote) {
+                        if (b.length() == 0) {
+                            // These would be removed by the null cleaner anyway.
+                            continue;
+                        }
```
efd6, line 1751: Currently this is required to maintain the behaviour that we have in the original code (when the original code also includes the empty field cleaner). This should not need to exist, but I have not been able to figure what the difference in the code is; this troubles me.

packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1776,23 +1780,23 @@ processors:
...
         // at http://daviddeley.com/autohotkey/parameters/parameters.htm#WINARGV
         // Original implementation found at: https://github.com/golang/go/commit/39c8d2b7faed06b0e91a1ad7906231f53aab45d1
-        def commandLineToArgv(String cmd) {
+        def commandLineToArgv(String line) {
             def args = new ArrayList();
-            while (cmd.length() > 0) {
-                if (cmd.charAt(0) == (char)' ' || cmd.charAt(0) == (char)0x09) {
-                    cmd = cmd.substring(1);
+            for (int i = 0; i < line.length();) {
+                if (line.charAt(0) == (char)' ' || line.charAt(0) == (char)0x09) {
```
kcreddy, line 1794:
```suggestion
                if (line.charAt(i) == (char)' ' || line.charAt(i) == (char)0x09) {
```

efd6: Thanks.

joegallo:
Related to https://github.com/elastic/elasticsearch/issues/125745, which would put the responsibility for this parsing in Elasticsearch itself (I'm just leaving breadcrumbs for future Github archaeologists).

joegallo:
> The empty field change reduces the number of allocations for iterators by half and improves the readability of the code.

The precise issue is that _iterating_ on the `values()` of the `ctx` happens to currently require that we allocate and populate a new HashSet with all the properties of the metadata of the `IngestDocument`. AFAICT there's no way around that in painless code -- you can't access the `getSource()` method that would allow you to iterate less expensively over the more correct object, so you're always going to be paying that cost. If we promote this up into Elasticeearch (see https://github.com/elastic/elasticsearch/issues/86157) then we can be faster (in addition to preventing all the current copy/paste duplication of this code).

efd6:
@joegallo Are you happy with the text of the commit message in terms of reflecting the reality of the situation closely enough for a future spelunker?

joegallo:
Oh, certainly! It's fine.

efd6:
@elastic/sec-linux-platform and @elastic/sec-windows-platform you may be interested in the second commit here for these ingest pipelines:
```
sysmon_linux/log/default.yml
windows/powershell/default.yml
windows/forwarded/powershell.yml
windows/powershell_operational/default.yml
windows/forwarded/sysmon_operational.yml
windows/forwarded/powershell_operational.yml
windows/sysmon_operational/default.yml
```

kcreddy (approved):

andrewkroh (approved):

PR #13329, opened 2025-03-27
[Jamf_Pro] - Fix empty string issue for date query param in filter for Jamf Pro inventory data stream

ShourieG:
## Type of change
- Bug

## Proposed commit message
**WHAT:** Fixed empty string issue for date query param in filter for Jamf Pro inventory data stream.

**WHY:** Recently there was a commit changing how the filter works detailed [here](https://github.com/harnish-elastic/integrations/commit/c730ea0f0ddbaa955e2a7189043617827fa880b5#diff-0ffc4d4d6aab9bfef1f294ac1e04c5560811b7c9ad54baba832b9d369b989476R40). The issue is that this change does not take into account the fact that due to the [previously existing cursor logic](https://github.com/harnish-elastic/integrations/commit/c730ea0f0ddbaa955e2a7189043617827fa880b5#diff-0ffc4d4d6aab9bfef1f294ac1e04c5560811b7c9ad54baba832b9d369b989476L73), there could be scenarios where the `cursor.last_report_date` is an **empty string**. 

Sending the date in the filter as an empty string causes the API to respond with a status_code: 400 leading to failures. The fix in this PR tries to address this specific issue along with some minor cleanup.

**Commit Message:**  `The current cursor handling assumes that state.cursor.last_report_date is
either absent or a valid cursor. However, previously stored cursor values may
include an empty string for this value to indicate an absence. The outcome of
this is that an invalid filter parameter was being sent in those cases. This
changes the logic to correctly handle cursors that have used an empty string
as an absence marker.`

## NOTE
Some minor cleanup and additions are included since they are all + changes. Some are just elastic-package format behaviour.

## Checklist

## Author's Checklist

- [ ] 

## Related issues
- Relates https://github.com/elastic/integrations/pull/12231

### DIFF
--- a/packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json
+++ b/packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json
@@ -229,7 +229,9 @@
                     "supervised": false,
                     "mdmCapable": {
                         "capable": false,
-                        "capableUsers": ["user.mcuserface@company.com"]
+                        "capableUsers": [
+                            "user.mcuserface@company.com"
+                        ]
                     },
                     "reportDate": "2024-06-19T15:54:37.692Z",
                     "lastContactTime": "2024-04-18T14:26:51.514Z",
--- a/packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs
+++ b/packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs
@@ -29,6 +29,8 @@ publisher_pipeline.disable_host: true
 {{/contains}}
 
 keep_null: true
+redact:
+    fields: ~
 program: |-
   request(
   	"GET",
@@ -36,7 +38,10 @@ program: |-
   		"section": state.sections,
   		"page-size": [string(state.page_size)],
   		"sort": ["general.reportDate:asc"],
-  		?"filter": state.?cursor.last_report_date.optMap(d, ["general.reportDate>=\"" + d + "\""]),
+  		?"filter": (has(state.?cursor.last_report_date) && state.cursor.last_report_date.orValue("") != "") ?
+   			optional.of(["general.reportDate>=\"" + state.cursor.last_report_date + "\""])
+   		:
+   			optional.none(),
   	}.format_query()
   ).with(
   	{
--- a/packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs
+++ b/packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs
@@ -38,7 +38,7 @@ program: |-
   		"section": state.sections,
   		"page-size": [string(state.page_size)],
   		"sort": ["general.reportDate:asc"],
-  		?"filter": (has(state.?cursor.last_report_date) && state.cursor.last_report_date.orValue("") != "") ?
+  		?"filter": state.?cursor.last_report_date.orValue("") != "" ?
    			optional.of(["general.reportDate>=\"" + state.cursor.last_report_date + "\""])
    		:
    			optional.none(),
--- a/packages/jamf_pro/data_stream/inventory/manifest.yml
+++ b/packages/jamf_pro/data_stream/inventory/manifest.yml
@@ -4,11 +4,7 @@ streams:
   - input: cel
     template_path: cel.yml.hbs
     title: Inventory data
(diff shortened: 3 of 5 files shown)


packages/jamf_pro/data_stream/inventory/_dev/test/pipeline/test-inventory.json
```diff
@@ -229,7 +229,9 @@
                     "supervised": false,
                     "mdmCapable": {
                         "capable": false,
-                        "capableUsers": ["user.mcuserface@company.com"]
+                        "capableUsers": [
+                            "user.mcuserface@company.com"
+                        ]
```
ShourieG, line 234: cause by elastic-package, but its a + change so I let it stay.

packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs
```diff
@@ -29,14 +29,19 @@ publisher_pipeline.disable_host: true
 {{/contains}}
 
 keep_null: true
+redact:
+    fields: ~
```
ShourieG, line 33: added recommended `redact` field since it's a + change.

packages/jamf_pro/data_stream/inventory/manifest.yml
```diff
@@ -4,11 +4,7 @@ streams:
   - input: cel
     template_path: cel.yml.hbs
     title: Inventory data
-    description: |
-      Settings for inventory requests.  
-      *NOTE* Enabling many sections may impact query processing.
-      It is recommended to only enable sections of interest.
-      The General section is always included.
+    description: "Settings for inventory requests.  \n*NOTE* Enabling many sections may impact query processing.\nIt is recommended to only enable sections of interest.\nThe General section is always included.\n"
```
ShourieG, line 7: caused by elastic-package lint

packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs
```diff
@@ -29,14 +29,19 @@ publisher_pipeline.disable_host: true
...
   	"GET",
   	state.url.trim_right("/") + "?" + {
   		"section": state.sections,
   		"page-size": [string(state.page_size)],
   		"sort": ["general.reportDate:asc"],
-  		?"filter": state.?cursor.last_report_date.optMap(d, ["general.reportDate>=\"" + d + "\""]),
+  		?"filter": (has(state.?cursor.last_report_date) && state.cursor.last_report_date.orValue("") != "") ?
+   			optional.of(["general.reportDate>=\"" + state.cursor.last_report_date + "\""])
+   		:
+   			optional.none(),
```
efd6, line 44:
```suggestion
  		?"filter": state.?cursor.last_report_date.orValue("") != "" ?
   			optional.of(["general.reportDate>=\"" + state.cursor.last_report_date + "\""])
   		:
   			optional.none(),
```
`state.cursor.last_report_date.orValue("")` is not valid since `state.cursor.last_report_date` is not an optional type.

packages/jamf_pro/data_stream/inventory/manifest.yml
efd6: I'm not sure this improves things.

efd6 (commented):
Suggest the following for the commit message:
```
The current cursor handling assumes that state.cursor.last_report_date is
either absent or a valid cursor. However, previously stored cursor values may
include an empty string for this value to indicate an absence. The outcome of
this is that an invalid filter parameter was being sent in those cases. This
changes the logic to correctly handle cursors that have used an empty string
as an absence marker.
```

Note that commit messages do not render markdown.

packages/jamf_pro/data_stream/inventory/manifest.yml
ShourieG: Yes agree but since it aligns with the format expected by elastic-package, I think it's ok to keep it in.

ShourieG:
@efd6, addressed your suggested changes.

packages/jamf_pro/data_stream/inventory/manifest.yml
efd6: I think it reflects a bug in the yaml implementation and it makes the code harder to read, but I won't push.

efd6 (approved):
Thanks.

Please make sure the line wraps are sane in the final commit message.

PR #13333, opened 2025-03-27
checkpoint_harmony_endpoint: improve handling of API errors

chemamartinez:
## Proposed commit message

```
The two changes included are:
- The handling of the 404 error when hitting the retrieve endpoint (that happens when the integration already received a valid task ID and page token, in this case, both parameters are cleaned and a new task ID is requested again for the same timeframe.
- In the generic error handling, also clean task ID and page token to avoid entering in a loop making requests with stale parameters.
```

> [!NOTE]

## Checklist

### DIFF
--- a/packages/checkpoint_harmony_endpoint/changelog.yml
+++ b/packages/checkpoint_harmony_endpoint/changelog.yml
@@ -3,10 +3,10 @@
   changes:
     - description: Improve cleaning of query parameters on errors.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/99999
+      link: https://github.com/elastic/integrations/pull/13333
     - description: Handle 404 error for the retrieve endpoint.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/99999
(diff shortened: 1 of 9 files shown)


chrisberkhout (commented):
The 404 changes look good.

For the generic error handling changes, is it intentional that the time advances? It seems like it should also repeat.

chemamartinez:
> The 404 changes look good.
> 
> For the generic error handling changes, is it intentional that the time advances? It seems like it should also repeat.

@chrisberkhout, yes, that makes sense I think. I've trying to get some troubleshooting advices from the API docs but unfortunately there's nothing there.

Do you think it would be worthy to merge both 404 and generic cases by setting `want_more: false` in 404 case? that would simplify the error logic.

chrisberkhout:
> Do you think it would be worthy to merge both 404 and generic cases by setting `want_more: false` in 404 case? that would simplify the error logic.

Yes, good idea!

chrisberkhout (approved):

PR #12781, opened 2025-02-13
Updated description on ssl node in package level manifest.yml to be uniform and include links to online documentation for integrations owned by security-service-integrations

StacieClark-Elastic:
## Proposed commit message

Updates description field on ssl nodes in package level manifest.yml file to include links to online documentation and to be consistent with other integrations. Changes are for integrations owned by security-service-integrations

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [ x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

# How to test this PR locally

Originally, we updated the descriptions on all the files in integration. This created some issues with so many teams needing to validate the files. This is currently a partial update with files that are owned by security-service-integrations.

git diff main | grep description: | grep + | sort -u
results in the update fields for the ssl node description and the changelog.yml

## Related issues

- Closes #12705
- Relates #11792
- Requires #123
- Supersedes #123

PR #13362, opened 2025-03-31
[Entity Analytics AD] Add config option to set TLS options

### LINKED ISSUE #10335: [entityanalytics_ad] Add config option to set TLS options
When communicate to an AD server over LDAPS users might need to configure additional TLS settings in order to establish a connection or to meet security requirements (e.g. only use TLS 1.3). The integration should expose a way to set TLS options for the underlying input.

References

- https://github.com/elastic/beats/blob/8bf258131a76c534f7272baccb7bca7dc607d030/x-pack/filebeat/input/entityanalytics/provider/activedirectory/conf.go#L43-L44

moxarth-rathod:
## Proposed commit message

Adding a new configuration option to set TLS for the input in the integration.

## Checklist

## Related issues

- Closes #10335


packages/entityanalytics_ad/data_stream/user/manifest.yml
```diff
@@ -51,6 +51,35 @@ streams:
         multi: false
         required: true
         show_user: true
+      - name: ssl
+        type: yaml
+        title: SSL Configuration
+        description: i.e. certificate_authorities, supported_protocols, verification_mode etc.
```
kcreddy, line 57: Can you use the description from here?
https://github.com/elastic/integrations/pull/12781/files

kcreddy, line 57:
```yaml
description: i.e. certificate_authorities, supported_protocols, verification_mode etc.
```

@moxarth-rathod, I am now curious where did you copy this from? which package did you reference and why?

moxarth-rathod: @kcreddy I think there was some mistake happened while adding this. I've updated the description as it should be 👍.

kcreddy: It would've been good to know where this was copied from, in case there's a script or template, which can be changed, to avoid any future changes that could slip through the PR review process.

kcreddy (approved):

packages/entityanalytics_ad/data_stream/user/manifest.yml
moxarth-rathod: Actually i copied it from my local and that seems to be outdated/incorrect there, so i checked it in this repo and there is no issue with it. There was a problem with my local files, nothing else.

kcreddy: thank you!

PR #13366, opened 2025-03-31
[azure_logs] Add the missing storage_account_container option

zmoog:
This wasn't included by mistake

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/azure_logs/manifest.yml
+++ b/packages/azure_logs/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.3.0
 name: azure_logs
 title: "Custom Azure Logs"
-version: 0.3.0
+version: 0.3.1
 source:
   license: Elastic-2.0
 description: "Collect log events from Azure Event Hubs with Elastic Agent"
@@ -93,7 +93,15 @@ policy_templates:
           The ingest pipeline ID to use for processing the data. If provided,
           replaces the default pipeline for this integration.
         required: false
-        show_user: true        
+        show_user: true
+      - name: storage_account_container
+        type: text
+        title: Storage Account Container
+        multi: false
+        required: false
+        show_user: false
+        description: >
+          The storage account container where the integration stores the checkpoint data for the consumer group. It is an advanced option to use with extreme care. You MUST use a dedicated storage account container for each Azure log type (activity, sign-in, audit logs, and others). DO NOT REUSE the same container name for more than one Azure log type. See [Container Names](https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names) for details on naming rules from Microsoft. The integration generates a default container name if not specified.
       - name: resource_manager_endpoint
         type: text
         title: Resource Manager Endpoint


bvader:
@zmoog I tested the following combinations

- V1 processor with multiple consumers with distinct `storage_account_container`  values
- V2 processor with multiple consumers without setting  `storage_account_container`
- V2 processor with multiple consumers with setting a common `storage_account_container`

All these tests were successful, excellent!!

packages/azure_logs/manifest.yml
```diff
@@ -93,7 +93,15 @@ policy_templates:
...
-        show_user: true        
+        show_user: true
+      - name: storage_account_container
+        type: text
+        title: Storage Account Container
+        multi: false
+        required: false
+        show_user: false
+        description: >
+          The storage account container where the integration stores the checkpoint data for the consumer group. It is an advanced option to use with extreme care. If you are using the [processor_version](https://www.elastic.co/guide/en/integrations/current/azure_logs.html#azure_logs-event-hub-processor-v2-only) v1, you MUST use a dedicated storage account container for each event hub. See [Container Names](https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names) for details on naming rules from Microsoft. The integration generates a default container name if not specified.
```
muthu-mps, line 104: Shouldn't this be v2 instead of v1?
```suggestion
          The storage account container where the integration stores the checkpoint data for the consumer group. It is an advanced option to use with extreme care. If you are using the [processor_version](https://www.elastic.co/guide/en/integrations/current/azure_logs.html#azure_logs-event-hub-processor-v2-only) v2, you MUST use a dedicated storage account container for each event hub. See [Container Names](https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names) for details on naming rules from Microsoft. The integration generates a default container name if not specified.
```

zmoog: No, this limit applies to processor v1 only. 

The processor v1 stores the blobs in the container root only. So, if you want to ingest logs from the same event hub with two consumer groups (for different use cases), you must use two containers.

The processor v2 stores the blobs in folder with the `/namespace/eventhub/consumer-group/` structure, so it can use the same container.

I'll specify that processor v1 needs a dedicated container for each unique eventhub/consumer group combination, instead of event hub only. WDYT?

zmoog: I posted an updated version of storage_account_container description.

zmoog:
> One more question, why do we add link to the processor V2(https://www.elastic.co/guide/en/integrations/current/azure_logs.html#azure_logs-event-hub-processor-v2-only) for V1?

Uhm, the `processor_version` option currently is located in the 'v2 only section', but I should probably restructure this section into 'processor options' and annotate the v2 options accordingly.

Great point. I'm making the change, thanks!

zmoog:
@muthu-mps here's the final revision:

CleanShot 2025-04-01 at 13 44 45@2x

CleanShot 2025-04-01 at 13 46 21@2x

Issue #12563, opened 2025-02-03
Standardize Ingested Data for Response Actions


raqueltabuyo:
### **Description**
To improve bidirectional response actions, we need consistent data ingestion from CrowdStrike, SentinelOne, and Microsoft Defender for Endpoint. This includes host listings, machine action tracking, and essential identifiers.

### **Requirements**
Ensure all three integrations provide:
- A **list of hosts** (following SentinelOne’s current approach).
- **List of action statuses** (e.g., "Machine Actions" in Microsoft Defender).
- `agent.id` - Some systems (e.g,. SentinelOne) have multiple “agent”/“Host” ID, the one needed to be captured is the ID that in turn is used with the external system’s API.
- `host.os.type`

### **Impact**
- Inconsistent data limits the effectiveness of automated response actions.
- Without standard fields, response workflows may not work uniformly across vendors.

### **Next Steps**
1. Identify discrepancies in host listings, action status tracking, and key identifiers across vendors.
2. Ensure proper ECS mapping and normalization for these fields.
3. Implement required transformations or enrichments to fill data gaps.
4. Prioritize Microsof Defender.

raqueltabuyo:
Hey @paul-tavares , could you please provide additional technical information to @kcreddy in this ticket? 
Can you provide a deep-dive of the current Sentinel approach for list of hosts as well as more info about what you would like to have for `host.os.type`, `agent.id` and list of action statuses?
Thanks in advance.

kcreddy:
Hey @paul-tavares,
 
In addition to the [above info](https://github.com/elastic/integrations/issues/12563#issuecomment-2736050728), I would like to get clarity on the following:
 
Crowdstrike has [`host`](https://github.com/elastic/integrations/tree/main/packages/crowdstrike/data_stream/host) data-stream for fetching host information, similarly SentineOne has [`agent`](https://github.com/elastic/integrations/tree/main/packages/sentinel_one/data_stream/agent). 
Currently [`Microsoft Defender for Endpoint`](https://github.com/elastic/integrations/tree/main/packages/microsoft_defender_endpoint/_dev/build/docs) integration has `log` data-stream, which seem to fetching `alerts` as per the docs:
> Note: When giving the application the API permissions described in the documentation (Windows Defender ATP Alert.Read.All), it will only grant access to read alerts from ATP and nothing else in the Azure Domain

[...]

PR #13373, opened 2025-04-01
[crowdstrike, microsoft_defender_endpoint, sentinel_one] Update host.* ECS mappings

brijesh-elastic:
## Proposed commit message

To improve bidirectional response actions, maintain consistent `host.*` mappings for CrowdStrike, SentinelOne, and Microsoft Defender for Endpoint packages.

All three integrations will provide `host.id` (which will be used for sending action requests to the EDR platform) and `host.os.type` values.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Relates #12563


packages/crowdstrike/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1401,6 +1402,32 @@ processors:
...
+          - windows
+          - ios
+          - android
+      source: |
+        if (ctx.crowdstrike?.alert?.device?.platform_name != null) {
+          for (String os: params.os_type) {
+            if (ctx.crowdstrike.alert.device.platform_name.toLowerCase().contains(os)) {
+              ctx.host.os.put('type', os);
+            }
+          }
```
efd6, line 1423: Is there a reason that we cannot do a single map look-up? If there is, the linear scan I guess is OK, but we should not be making a new string for each iteration of the os_type list.

```suggestion
          String platform_name = ctx.crowdstrike.alert.device.platform_name.toLowerCase();
          for (String os: params.os_type) {
            if (platform_name.contains(os)) {
              ctx.host.os.put('type', os);
            }
          }
```
(same below)

Can you explain why we cannot use a map look-up?

packages/crowdstrike/data_stream/host/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -643,6 +648,21 @@ processors:
       tag: set_host_os_platform_from_host_platform_name
       copy_from: crowdstrike.host.platform.name
       ignore_empty_value: true
+  - set:
+      field: host.os.type
+      tag: set_host_os_type_to_linux
+      value: linux
+      if: ctx.crowdstrike?.host?.platform?.name != null && ctx.crowdstrike.host.platform.name.toLowerCase().contains('lin')
```
efd6, line 655: Do we need to use `contains`? Can we not do an equality on a case-folded basis? (same below)

packages/crowdstrike/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -571,6 +582,25 @@ processors:
...
+        os_type:
+          - linux
+          - macos
+          - unix
+          - windows
+          - ios
+          - android
+      source: |
+        for (String os: params.os_type) {
+          if (ctx.crowdstrike.vulnerability.host_info.platform.toLowerCase().contains(os)) {
```
efd6, line 600: Map look-up or case-folded equality? If not, hoist the string construction outside the loop.

packages/sentinel_one/data_stream/activity/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -421,6 +425,25 @@ processors:
...
+          - macos
+          - unix
+          - windows
+          - ios
+          - android
+      source: |
+        for (String os: params.os_type) {
+          if (ctx.host.os.family.toLowerCase().contains(os)) {
+            ctx.host.os.put('type', os);
+          }
```
efd6, line 445: Same here.

packages/sentinel_one/data_stream/agent/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -618,10 +618,31 @@ processors:
...
+          - windows
+          - ios
+          - android
+      source: |
+        if (ctx.host == null) {
+          ctx.put('host', new HashMap());
+        }
+        if (ctx.host.os == null) {
+          ctx.host.put('os', new HashMap());
+        }
```
efd6, line 640:
```suggestion
        ctx.host = ctx.host ?: [:];
        ctx.host.os = ctx.host.os ?: [:];
```

packages/sentinel_one/data_stream/agent/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -618,10 +618,31 @@ processors:
...
+        for (String os: params.os_type) {
+          if (ctx.json.osType.toLowerCase().contains(os)) {
+            ctx.host.os.put('type', os);
+          }
```
efd6, line 644: Map look-up or case-folded equality? Otherwise hoist string construction.

packages/sentinel_one/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -132,10 +132,31 @@ processors:
...
+          ctx.put('host', new HashMap());
+        }
+        if (ctx.host.os == null) {
+          ctx.host.put('os', new HashMap());
+        }
+        for (String os: params.os_type) {
+          if (ctx.sentinel_one.alert.agent.os.type.toLowerCase().contains(os)) {
+            ctx.host.os.put('type', os);
+          }
+        }
```
efd6, line 159: Same here.

packages/sentinel_one/data_stream/threat/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -354,13 +354,31 @@ processors:
...
+          ctx.put('host', new HashMap());
+        }
+        if (ctx.host.os == null) {
+          ctx.host.put('os', new HashMap());
+        }
+        for (String os: params.os_type) {
+          if (ctx.json.agentRealtimeInfo.agentOsType.toLowerCase().contains(os)) {
+            ctx.host.os.put('type', os);
+          }
+        }
```
efd6, line 381: And here.

packages/crowdstrike/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
brijesh-elastic: I'm not sure what values we'll get in the `crowdstrike.alert.device.platform_name` field, so to cover all possible cases, I'm converting them to lowercase and checking if they contain any expected values from [host.os.type](https://www.elastic.co/guide/en/ecs/current/ecs-os.html#field-os-type).

applying same at other places...

packages/crowdstrike/data_stream/host/elasticsearch/ingest_pipeline/default.yml
brijesh-elastic: Yes, we can.
I was just trying to populate `host.os.type` to account for a wider range of values that may appear in `crowdstrike.host.platform.name`

packages/crowdstrike/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
efd6: OK. This is unfortunate, but if we have to do it, we have to do it.

packages/crowdstrike/data_stream/host/elasticsearch/ingest_pipeline/default.yml
efd6: Do you know that that will be necessary? Do you have a set of strings that you've seen?

packages/crowdstrike/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
efd6: Please hoist the string construction out of the loop.

packages/crowdstrike/data_stream/host/elasticsearch/ingest_pipeline/default.yml
brijesh-elastic: I haven’t seen anything for `lin`, but we do have `mac` and `windows` values in the pipeline logs. I did this because the [FDR ingest pipeline](https://github.com/elastic/integrations/blob/main/packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml#L1727-L1730) is doing the same.

efd6: OK

efd6 (approved):
Thanks

PR #13393, opened 2025-04-02
[ZPA User Activity] do not set error.message for expected hostname behaviour

SimonKoetting:
For the User activity Log stream as the Zscaler response format, we
document that users should use '"Host": %j{Host}'.

According to the Zscaler documentation[1], Host is defined as "The host
domain or IP address", so it's expected to get hostnames or IP
addresses. So do not set error.message, since it's an expected behavior
and the value is already copied to the correct field in the error
handler processor.

[1]https://help.zscaler.com/zpa/about-user-activity-log-fields


packages/zscaler_zpa/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.21.1"
+  changes:
+    - description: Remove error.message for expected behavior related to Host
```
efd6, line 4:
```suggestion
    - description: Do not set `error.message` for expected behavior related to Zscaler `Host` field.
```

efd6 (commented):
Suggest the following for the commit message body:
```
For the User activity Log stream as the Zscaler response format, we
document that users should use '"Host": %j{Host}'.

According to the Zscaler documentation[1], Host is defined as "The host
domain or IP address", so it's expected to get hostnames or IP
addresses. So do not set error.message, since it's an expected behavior
and the value is already copied to the correct field in the error
handler processor.

[1]https://help.zscaler.com/zpa/about-user-activity-log-fields
```

efd6 (approved):

efd6 (approved):

PR #13421, opened 2025-04-04
proofpoint_on_demand: set subobjects false to msg_parts.metadata field

chemamartinez:
## Proposed commit message

Applying `subobjects: false` to the `message.msg_parts.metadata` field in the messages data stream.

This object in source documents can look like:
```
"metadata": {
  "mxtype": "Documentation",
  "mxtype.localized": "Documentation",
  ...
}
```

Leading to the error when ingesting it:
```
(status=400): {"type":"illegal_argument_exception","reason":"can't merge a non object mapping [proofpoint_on_demand.message.msg_parts.metadata.mxtype] with an object mapping"}, dropping event!"
```

Setting `subobjects: false` so fields can be stored avoiding the nesting and the conflict.

## Checklist

### DIFF
--- a/packages/proofpoint_on_demand/changelog.yml
+++ b/packages/proofpoint_on_demand/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.5.1"
+  changes:
+    - description: Set subobjects false to `message.msg_parts.metadata` to avoid mapping conflicts.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/99999
 - version: "1.5.0"
   changes:
     - description: Update Kibana constraint to support 9.0.0.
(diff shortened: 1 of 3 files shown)


kcreddy (approved):
LGTM.
Is it possible to add a sample test?

chemamartinez:
> LGTM. Is it possible to add a sample test?

I wonder if is possible to add case to the system tests right now. I have tried without success so far.

From the real case, the error before applying the change points to the field after being processed by the ingest pipeline (`can't merge a non object mapping [proofpoint_on_demand.message.msg_parts.metadata.mxtype] with an object mapping`). However, when adding an example of subfields in the system test, [here](https://github.com/elastic/integrations/blob/235390f2fb9724469081a4d184d99b37e9e58fc6/packages/proofpoint_on_demand/_dev/deploy/docker/websocket-mock-service/main.go#L52), it fails pointing to the field before being processed (`can't merge a non object mapping [json.msgParts.metadata.mock] with an object mapping`).

Still investigating the reason of this behaviour.

kcreddy (approved):

PR #13426, opened 2025-04-04
ti_anomali: Add `initial_interval` parameter to API ingestion.

kcreddy:
## Proposed commit message
Without initial polling interval, users fetch indicators from years ago
as seen in their Anomali Threatstream.

Adding configurable `initial_interval` option (with default 90 days) 
will help users adjust how far back to fetch indicators from.

## Checklist

Added `initial_interval` option to system tests config. System tests should pass.

```
--- Test results for package: ti_anomali - START ---
╭────────────┬──────────────┬───────────┬───────────┬────────┬───────────────╮
│ PACKAGE    │ DATA STREAM  │ TEST TYPE │ TEST NAME │ RESULT │  TIME ELAPSED │
├────────────┼──────────────┼───────────┼───────────┼────────┼───────────────┤
│ ti_anomali │ intelligence │ system    │ default   │ PASS   │ 37.072411042s │
╰────────────┴──────────────┴───────────┴───────────┴────────┴───────────────╯
--- Test results for package: ti_anomali - END   ---
Done
```

### DIFF
--- a/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
+++ b/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
@@ -39,7 +39,7 @@ program: |-
   		state.url.trim_right("/") + "/api/v2/intelligence/?" + {
   			"order_by": ["update_id"],
   			"modified_ts__lt": [(now - duration(state.delay)).format(time_layout.RFC3339)],
-  			?"modified_ts__gt": !has(state.cursor) ? optional.of([(now - duration(state.initial_interval)).format(time_layout.RFC3339)]) : optional.none(),
+  			?"modified_ts__gt": state.?cursor.last_update_id.orValue(null) == null ? optional.of([(now - duration(state.initial_interval)).format(time_layout.RFC3339)]) : optional.none(),
   			"limit": [string(state.page_size)],
   			?"update_id__gt": state.?cursor.last_update_id.optMap(id, [string(int(id))]),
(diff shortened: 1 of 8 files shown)


packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
```diff
@@ -38,6 +39,7 @@ program: |-
   		state.url.trim_right("/") + "/api/v2/intelligence/?" + {
   			"order_by": ["update_id"],
   			"modified_ts__lt": [(now - duration(state.delay)).format(time_layout.RFC3339)],
+  			?"modified_ts__gt": !has(state.cursor) ? optional.of([(now - duration(state.initial_interval)).format(time_layout.RFC3339)]) : optional.none(),
```
chrisberkhout, line 42: It probably should check for `state.?cursor.last_update_id.orValue(null) != null` rather than `!has(state.cursor)`, because currently, if the first request returns no data, `state.cursor` will exist as an empty object.

chrisberkhout (changes requested):
Looks good except for one edge case that should probably be handled.

packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
kcreddy: Thanks @chrisberkhout.
I updated to `state.?cursor.last_update_id.orValue(null) == null` because I was checking other way around.

chrisberkhout (approved):
Looks good. It works against the live API.

PR #13427, opened 2025-04-04
[system][syslog] add a pattern in filebeat.system module to capture greedy multiline logs with ISO timestamps

stefans-elastic:
## Proposed commit message

add a pattern in filebeat.system module to capture greedy multiline logs with ISO timestamps

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/obs-integration-team/issues/198


packages/system/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.67.4"
+  changes:
+    - description: Add new grok pattern to system (syslog) module to capture multiline logs with ISO 8601 timestamps.
+      type: bugfix
```
ishleenk17, line 5: enhancement

packages/system/data_stream/syslog/_dev/test/pipeline/test-iso8601-timestamp-multiline-syslog.log-expected.json
```diff
@@ -0,0 +1,27 @@
...
+        "kind": "event",
+        "timezone": "GMT-0200"
+      },
+      "host": {
+        "hostname": "a-mac-with-esc-key"
+      },
+      "input": {
+        "type": "log"
+      },
+      "message": "2016-12-13 11:35:28.421\n\t\tGoogleSoftwareUpdateAgent[21412/0x700007399000] [lvl=2] -[KSUpdateEngine updateAllExceptProduct:] KSUpdateEngine updating all\n\t\tinstalled products, except:'com.google.Keystone'.",
```
ishleenk17, line 17: This test log has actually matched the 3rd pattern. Please add a log that doesn't mathc 3rd, but only the 4th newly added grok pattern.

ishleenk17:
Also, can you state details about the second ask as per this [issue](https://github.com/elastic/obs-integration-team/issues/198).

packages/system/changelog.yml
stefans-elastic: fixed

stefans-elastic:
> Also, can you state details about the second ask as per this [issue](https://github.com/elastic/obs-integration-team/issues/198).

Sure.


The gh issue contains two points:

Include the greedy grok pattern not only for SYSLOGTIMESTAMP, but for ISO8601 as well.
A process might send logs to /var/log/messages with white space characters and not the white spaces themselves, something like this
So this PR covets point 1. As for point 2: I'm not sure it is possible to have log entries with escaped whitespace characters. The example given in the issue seems to be coming from result event and for those the whitespace characters are escaped (as can be seen here: https://github.com/elastic/beats/blob/main/filebeat/module/system/syslog/test/darwin-syslog.log-expected.json#L31)

packages/system/data_stream/syslog/_dev/test/pipeline/test-iso8601-timestamp-multiline-syslog.log-expected.json
```diff
@@ -0,0 +1,27 @@
...
+        "kind": "event",
+        "timezone": "GMT-0200"
+      },
+      "host": {
+        "hostname": "a-mac-with-esc-key"
+      },
+      "input": {
+        "type": "log"
+      },
+      "message": "2022-04-21T14:30:00Z\n\t\tGoogleSoftwareUpdateAgent[21412/0x700007399000] [lvl=2] -[KSUpdateEngine updateAllExceptProduct:] KSUpdateEngine updating all\n\t\tinstalled products, except:'com.google.Keystone'.",
```
ishleenk17, line 17: This is still matching the 3rd pattern.
If you see we are getting fields as host.hostname, process.name which matches the 3rd pattern .
To match the 4th pattern , one of these fields should be sipped. So that other than timestamp the complete message becomes part of the greedymessage.

ishleenk17:
> > Also, can you state details about the second ask as per this [issue](https://github.com/elastic/obs-integration-team/issues/198).
> 
> Sure.
> 
> The gh issue contains two points:
> 
> Include the greedy grok pattern not only for SYSLOGTIMESTAMP, but for ISO8601 as well. A process might send logs to /var/log/messages with white space characters and not the white spaces themselves, something like this So this PR covets point 1. As for point 2: I'm not sure it is possible to have log entries with escaped whitespace characters. The example given in the issue seems to be coming from result event and for those the whitespace characters are escaped (as can be seen here: https://github.com/elastic/beats/blob/main/filebeat/module/system/syslog/test/darwin-syslog.log-expected.json#L31)

Since, we don't observe these logs and this could be a corner case.lets focus on just adding the ISO timestamp for this PR.

packages/system/data_stream/syslog/_dev/test/pipeline/test-iso8601-timestamp-multiline-syslog.log-expected.json
stefans-elastic: hmm, what is the "tell" of it (I think I'm missing it)?

packages/system/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 3.0.2
 name: system
 title: System
-version: "1.67.3"
+version: "1.67.4"
```
ishleenk17, line 4:
```suggestion
version: "1.68.0"
```

packages/system/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.67.4"
```
ishleenk17, line 2:
```suggestion
- version: "1.68.0"
```

stefans-elastic:
> Please make the version change. Otherwise, LGTM.

updated the version

PR #13443, opened 2025-04-07
[Asset Discovery] Change entity.name from text to keyword

romulets:
`entity.name` was mistakenly mapped as `text` and not `keyword` impacting entity store aggregations


kubasobon (approved):
Sure 🚀

PR #13450, opened 2025-04-07
[cisco_ios] Fix parsing of FQDN hostnames

mjwolf:
## Proposed commit message

In some cases messages with FQDN hostnames were not being parsed properly. The message body and header were not being properly parsed, so some header data (e.g. `event.code`) was not being parsed. This corrects parsing of these cases.

The problem was the most greedy case of consuming the most of the body into `message` was hit before the case of properly parsing more header fields, so rearranging the patterns fixes things. This also reduces the number of patterns by combining very similar patterns.

## Checklist


packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -38,19 +38,17 @@ processors:
...
-        - '^%{SYSLOGTIMESTAMP} (?:%{IP}|%{HOSTNAME:log.syslog.hostname}) %{CISCO_PRIORITY_MSGCOUNT}?(?:%{NUMBER:cisco.ios.sequence}: )%{GREEDYDATA:_temp_.message}$'
       pattern_definitions:
         ISO8601_TIMEZONE: "(?:Z|[+-]%{HOUR}(?::?%{MINUTE}))"
         TIMESTAMP_ISO8601: "%{YEAR}-%{MONTHNUM}-%{MONTHDAY}[T ]%{HOUR}:?%{MINUTE}(?::?%{SECOND})?%{ISO8601_TIMEZONE:_temp_.tz}?"
         CISCO_PRIORITY_MSGCOUNT: '<%{NONNEGINT:log.syslog.priority:long}>(?:%{NONNEGINT:cisco.ios.message_count})?(?:: )?'
         CISCO_TIMESTAMP: '[*]?%{CISCOTIMESTAMP_EX:_temp_.cisco_timestamp}(?: %{CISCO_TZ:_temp_.tz})?'
         CISCOTIMESTAMP_EX: '(%{CISCOTIMESTAMP})|(%{YEAR} %{MONTH} %{MONTHDAY} %{TIME})'
         CISCO_UPTIME: '[0-9a-zA-Z]+'
-        CISCO_HOSTNAME: '[a-zA-Z][.0-9a-zA-Z_-]{0,253}[0-9a-zA-Z]?'
+        CISCO_HOSTNAME: '(?!\d+)[0-9a-zA-Z][.0-9a-zA-Z_-]{0,253}[0-9a-zA-Z]?'
```
ilyannn, line 51: I'm confused about the negative lookahead. Taking the string 

```
Jul 11 09:34:00 1my-router-hostname 1663469: Jul 11 09:34:00.334: %FMANFP-6-IPACCESSLOGP: R0/0: fman_fp_image: list internet_in_gig0 permitted udp 172.16.0.26(1985) -> 10.100.8.34(1985), 327 packets
```

the expression

```
/(?!\d+)[0-9a-zA-Z][.0-9a-zA-Z_-]{0,253}[0-9a-zA-Z]?/g
```

does not match `1my-router-hostname`, though it matches `my-router-hostname`. Is this how it works in the `grok` processor?

packages/cisco_ios/data_stream/log/_dev/test/pipeline/test-fqdn.log-expected.json
```diff
@@ -40,6 +40,47 @@
...
+            },
+            "ecs": {
+                "version": "8.17.0"
+            },
+            "event": {
+                "category": [
+                    "network"
+                ],
+                "code": "MACREPLACE",
+                "original": "Apr 2 08:03:15 xx-xx-xx-999.xxxx.xxxxxxxxxx.example <189>942562724: Apr 2 2025 06:03:15.915 UTC: %AUTHMGR-6-MACREPLACE: MAC address (0000.a636.6867) on Interface GigabitEthernet1/0/4 is replaced by MAC (66aa.6636.6867) AuditSessionID 0A66666A1A3AC8899A66AA6A]\"",
```
qcorporation, line 60: nit: I know a bit out of scope, but shouldn't we be parsing the MAC address into format AA-BB-CC-DD-EE-FF? both existing mac and replaced MAC

packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
qcorporation: I also don't completely understand this logic, couldn't you have gotten away with `[0-9a-zA-Z][...........]`
Running the (?!\d+)[0-9a-zA-Z][.0-9a-zA-Z_-]{0,253}[0-9a-zA-Z]? strips the number
Screenshot 2025-04-08 at 5 50 23 PM

mjwolf: Yes, you're both correct that this wasn't doing exactly what I thought it was doing, although it did somehow make all the existing tests pass.

But it turns out this is not required. The Cisco IOS documentation states that a hostname with leading digits is not supported, [in the hostname section](https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/fundamentals/command/cf_command_ref/F_through_K.html): `The name must also follow the rules for ARPANET hostnames. They must start with a letter, end with a letter or digit, and have as interior characters only letters, digits, and hyphens.`

So I've reverted this to the original pattern. This PR will now just fix the FQDN problem

packages/cisco_ios/data_stream/log/_dev/test/pipeline/test-fqdn.log-expected.json
mjwolf: The message body is not parsed at all for this message type, so we don't do anything with MAC in this case.

There are thousands of possible message types in IOS, and we only parse a handful. If we want to add support for all of them, it'll be a big project. I don't want to add support for this message, when I don't know if it's any more or less important than the other currently unsupported messages right now

packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -35,21 +35,19 @@ processors:
...
-        - '^%{SYSLOGTIMESTAMP} (?:%{IP}|%{HOSTNAME:log.syslog.hostname}) %{CISCO_PRIORITY_MSGCOUNT}?(?:%{NUMBER:cisco.ios.sequence}: )(\\*)?(?:%{CISCO_UPTIME:cisco.ios.uptime}|%{CISCO_TIMESTAMP}): %{GREEDYDATA:_temp_.message}$'
-        - '^%{SYSLOGTIMESTAMP} (?:%{IP}|%{HOSTNAME:log.syslog.hostname}) %{CISCO_PRIORITY_MSGCOUNT}?(?:%{NUMBER:cisco.ios.sequence}: )%{GREEDYDATA:_temp_.message}$'
       pattern_definitions:
         ISO8601_TIMEZONE: "(?:Z|[+-]%{HOUR}(?::?%{MINUTE}))"
         TIMESTAMP_ISO8601: "%{YEAR}-%{MONTHNUM}-%{MONTHDAY}[T ]%{HOUR}:?%{MINUTE}(?::?%{SECOND})?%{ISO8601_TIMEZONE:_temp_.tz}?"
         CISCO_PRIORITY_MSGCOUNT: '<%{NONNEGINT:log.syslog.priority:long}>(?:%{NONNEGINT:cisco.ios.message_count})?(?:: )?'
         CISCO_TIMESTAMP: '[*]?%{CISCOTIMESTAMP_EX:_temp_.cisco_timestamp}(?: %{CISCO_TZ:_temp_.tz})?'
         CISCOTIMESTAMP_EX: '(%{CISCOTIMESTAMP})|(%{YEAR} %{MONTH} %{MONTHDAY} %{TIME})'
-        CISCO_UPTIME: '[0-9a-zA-Z]+'
+        CISCO_UPTIME: '(?:\b\d{1,4}:\d{2}:\d{2}:|(?:(\d+)y)?(?:(\d+)w)?(?:(\d+)d)?(?:(\d+)h)?(?:(\d+)m)?(?:(\d+)s)?)'
```
qcorporation, line 50: The uptime changes look OK. Is there an example where we are parsing `1111:11:11:` as an example?

mjwolf: This comes from the [cisco documentation](https://www.cisco.com/E-Learning/bulk/public/tac/cim/cib/using_cisco_ios_software/cmdrefs/service_timestamps.htm), uptime can be `The uptime form of the command adds time stamps in the format HHHH:MM:SS`.

The other format is from existing test cases and is documented here: https://www.cisco.com/c/en/us/td/docs/ios/fundamentals/command/reference/cf_book.pdf

mjwolf: your right that the trailing `:` isn't correct, I'll remove it

qcorporation (approved):
Thanks for the updates

Issue #12718, opened 2025-02-11
Qualys VMDR: Implement transform for enhancing cloud security workflow


kcreddy:
To leverage cloud security workflow, mapping changes are implemented in https://github.com/elastic/integrations/pull/11999 as part of https://github.com/elastic/integrations/issues/11673. 

In addition to this, a latest transform must be implemented similar to [AWS SecurityHub](https://github.com/elastic/integrations/blob/main/packages/aws/elasticsearch/transform/latest_cdr_misconfigurations/transform.yml) as a final step.

PR #13455, opened 2025-04-08
qualys_vmdr: Add latest transform for Asset Host Detections

kcreddy:
## Proposed commit message
```
Adds a latest transform for vulnerabilities which allows data
from Qualys VMDR to be displayed in Elastic Security CNVM workflow.

Ref: https://github.com/elastic/integrations/issues/11673
```

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/12718


packages/qualys_vmdr/elasticsearch/transform/latest_cdr_vulnerabilities/fields/package.yml
efd6, line 1: I know that having these in separate files reflects the situation in the data stream's defs, but it makes me sad. I the rationale for this diff collision avoidance?

efd6 (approved):

packages/qualys_vmdr/elasticsearch/transform/latest_cdr_vulnerabilities/fields/package.yml
kcreddy: The fields in these files `package.yml`, `resource.yml`, `vulnerability.yml` are meant for extended ECS fields that are not in ECS definitions, but are used by the Elastic workflows.
Here both `package.name, package.version` are present in ECS: https://www.elastic.co/guide/en/ecs/current/ecs-package.html. But field `package.fixed_version` isn't. Hence it was separated. 
It could also help checking diff between source fields and destination fields: https://github.com/elastic/integrations/blob/main/packages/qualys_vmdr/data_stream/asset_host_detection/fields/package.yml as they are now identical.

kcreddy:
The PR will be moved to `ready for review` once live testing is finished.
@maxcold @alexreal1314

alexreal1314:
> The PR will be moved to `ready for review` once live testing is finished. cc: @maxcold @alexreal1314

still waiting for credentials to test it.

packages/qualys_vmdr/elasticsearch/transform/latest_cdr_vulnerabilities/transform.yml
```diff
@@ -0,0 +1,32 @@
...
+  index:
+    - "logs-qualys_vmdr.asset_host_detection-*"
+dest:
+  index: "security_solution-qualys_vmdr.vulnerability_latest-v1"
+  aliases:
+    - alias: "security_solution-qualys_vmdr.vulnerability_latest"
+      move_on_creation: true
+latest:
+  unique_key:
+    - vulnerability.id
```
maxcold, line 11: this might be causing duplication of the documents (we see duplication in the test env). I remember we had this conversation before, but I need to find out the outcome. If having fields with multiple values as unique_key causing duplication we might need to find another way, eg. use QID+resource.id+namespace as a key. We need to understand better our options here.

alexreal1314: @kcreddy @maxcold IMO we should use event.id+resource.id+namespace since we set event.id based on [unique_vuln_id](https://github.com/elastic/integrations/blob/195342148b508a875724de1c14aec30d267182e8/packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml#L1069) in host detection, or use qualys_vmdr.knowledge_base.qid field in knowledge_base ingest pipeline.

kcreddy: Thanks for the suggestion @alexreal1314.
I have implemented it here: https://github.com/elastic/integrations/pull/13455/commits/591a0f09e3dfbe18e97e449842043b67d06e417f.

packages/qualys_vmdr/elasticsearch/transform/latest_cdr_vulnerabilities/transform.yml
```diff
@@ -0,0 +1,32 @@
...
+settings:
+  unattended: true
+frequency: 5m
+sync:
+  time:
+    field: event.ingested
+retention_policy:
+  time:
+    field: "@timestamp"
+    max_age: 4h
```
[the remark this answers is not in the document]
kcreddy: Thanks for the suggestion @maxcold and @clement-fouque 
I updated the retention to `24h` here: https://github.com/elastic/integrations/pull/13455/commits/591a0f09e3dfbe18e97e449842043b67d06e417f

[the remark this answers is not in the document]
maxcold: @clement-fouque thanks for sharing more context! We will work toward connecting the data retention param on the transform with the params of the integration, but for now, I think we should proceed with the longer retention period

maxcold:
@kcreddy we probably need to update `kibana.version` condition https://github.com/elastic/integrations/blob/main/packages/qualys_vmdr/manifest.yml#L12 to the `version: "^8.19.0 || ^9.1.0"` . @alexreal1314 as far as I can see you didn't backport the changes of supporting multiple CVEs and packages to other branches. We might need to backport at least to `8.x` so the changes go out in 8.19 as well.

kcreddy:
> @kcreddy we probably need to update `kibana.version` condition https://github.com/elastic/integrations/blob/main/packages/qualys_vmdr/manifest.yml#L12 to the `version: "^8.19.0 || ^9.1.0"` . @alexreal1314 as far as I can see you didn't backport the changes of supporting multiple CVEs and packages to other branches. We might need to backport at least to `8.x` so the changes go out in 8.19 as well.

Okay. Do you think merging this PR closer to `8.19` or after `8.19` makes sense? 
Prior to this, you can test the integration on `8.19` using the fleet API.

maxcold:
@nick-alayil can you provide your input on 
> Do you think merging this PR closer to 8.19 or after 8.19 makes sense?

This would mean that the Qualys data won't be available in Serverless before 8.19/9.1 release in June. Are we ok with that? do we have customers in Serverless we would like to enable with this feature?

nick-alayil:
> Do you think merging this PR closer to 8.19 or after 8.19 makes sense?

If the changes look solid and testing is good, I think we should merge this right away. We're following the serverless first approach as an org now anyway, so this aligns with that direction.

For ECH, do we actually need more testing and/or require any kibana changes to get in? If that's the case, then fine - we'll wait. But if not, we can leverage our ability to do independent integration releases and get the Qualys integration changes into the 8.18 stack rather than waiting for 8.19/9.1 in June.

maxcold:
@nick-alayilThanks for your input! To provide some context and clarify a potential issue with merging this change: For ECH, we control the compatible stack versions for this integration using `kibana.version: "^8.19.0 || ^9.1.0"` in the manifest. Therefore, these changes won't be immediately available in ECH, regardless of whether we merge now or later. This is because they depend on Alex's changes, which will only be included in Kibana versions 9.1 and 8.19.

The main concern with merging now is primarily engineering-related for the Security Service Integrations team. They would be unable to add new changes to the Qualys integration that might be compatible with older stack versions. This is because any changes made after merging this pull request will only be included in the versions compatible with the 8.19/9.1 stack. Consequently, the integrations team previously requested that we hold off on merging changes that limit the stack version to an unreleased version until closer to its release date.

However, I agree with the statement: 'We're following a serverless-first approach as an organization now anyway, so this aligns with that direction.' In my opinion, this technical limitation within the integrations ecosystem shouldn't block our serverless-first approach.

kcreddy:
> If the changes look solid and testing is good, I think we should merge this right away. We're following the serverless first approach as an org now anyway, so this aligns with that direction.

> However, I agree with the statement: 'We're following a serverless-first approach as an organization now anyway, so this aligns with that direction.' In my opinion, this technical limitation within the integrations ecosystem shouldn't block our serverless-first approach.

Thanks for the input @nick-alayil and @maxcold. 
I've updated the min stack version as per your suggestion -> [89b8323](https://github.com/elastic/integrations/pull/13455/commits/89b832352764b5bd59c7eece4bda8a91bcaef26a)

kcreddy:
The CI error should be fixed by https://github.com/elastic/elasticsearch/pull/126417 which was also introduced in `8.19.0` by https://github.com/elastic/elasticsearch/pull/125232.

efd6:
Running this locally on v8.19.0-SNAPSHOT, I have minimised this to
```
---
description: Pipeline for processing Asset Host Detection data.
processors:
  - set:
      field: message
      value: '{"ATTRIBUTE":[{"OTHER_FIELD":""}]}'
  - json:
      field: message
      target_field: json
  - remove:
      tag: remove_json
      field:
        - json.ATTRIBUTE.LAST_ERROR_DATE
      ignore_missing: true
on_failure:
  - append:
      field: error.message
      value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```

This ceases to fail when `ATTRIBUTE` is not an array. 

Using the dev tools gives an indication why this might be:
```
POST /_ingest/pipeline/_simulate
{
  "pipeline": {
    "processors": [
      {
        "json": {
          "field": "message",
          "target_field": "json"
        }
      },
     {
        "remove": {
          "field": [
            "json.ATTRIBUTE.LAST_ERROR_DATE"
          ],
          "ignore_missing": true
        }
      }
    ]
  },
  "docs": [
    {
      "_index": "idx",
      "_id": "id",
      "_source": {
        "message": """{"ATTRIBUTE":[{"OTHER_FIELD":""}]}"""
      }
    }
  ]
}
```
gives
```
{
  "docs": [
    {
      "error": {
        "root_cause": [
          {
            "reason": "[LAST_ERROR_DATE] is not an integer, cannot be used as an index as part of path [json.ATTRIBUTE.LAST_ERROR_DATE]"
          }
        ],
        "reason": "[LAST_ERROR_DATE] is not an integer, cannot be used as an index as part of path [json.ATTRIBUTE.LAST_ERROR_DATE]",
        "caused_by": {
          "type": "number_format_exception",
          "reason": "For input string: \"LAST_ERROR_DATE\""
        }
      }
    }
  ]
}
```
while when the document is `{"ATTRIBUTE":{"OTHER_FIELD":""}}` there is no error.

If I run the same failing simulate request on v8.17.3, I get no failure.

The image for the v8.19.0-SNAPSHOT is 3c3225efef7e8285bb2c5de5cf0b15fe4747677fc84473a2b3db53fe2391212e, which reports `{"@timestamp":"2025-04-17T02:29:50.155Z", "log.level": "INFO", "message":"version[8.19.0-SNAPSHOT], pid[233], build[docker/c070c976010455cf7d9b45b6270cb6709e84c1ae/2025-04-15T17:38:52.824533143Z], OS[Linux/6.10.14-linuxkit/amd64], JVM[Oracle Corporation/OpenJDK 64-Bit Server VM/24/24+36-3646]", "ecs.version": "1.2.0","service.name":"ES_ECS","event.dataset":"elasticsearch.server","process.thread.name":"main","log.logger":"org.elasticsearch.node.Node","elasticsearch.node.name":"3c3225efef7e","elasticsearch.cluster.name":"elasticsearch"}` which is [here](https://github.com/elastic/elasticsearch/commit/c070c976010455cf7d9b45b6270cb6709e84c1ae) and **should** contain the fix in elastic/elasticsearch#126417. So I have to conclude that the issue here is not the same as the one that was fixed in that PR — this is not completely surprising since the error messages are subtly different.

kcreddy:
Thanks @efd6 for the analysis. It is indeed a variation of https://github.com/elastic/elasticsearch/pull/126417 but for lists.

@joegallo made the fix in https://github.com/elastic/elasticsearch/pull/127006 and should be available in `v8.19.0` stack version.

alexreal1314:
Hi @kcreddy @maxcold I've tested latest changes on my env https://alex-dev91.kb.us-west2.gcp.elastic-cloud.com/ and seems that after we changed the unique key of the transform there are no duplicates.

this is the query i ran:
POST security_solution-qualys_vmdr.vulnerability_latest-v1/_search
```
{
  "size": 0,
  "aggs": {
    "duplicate_check": {
      "composite": {
        "size": 10000,
        "sources": [
          { "event_id": { "terms": { "field": "event.id.keyword" } } },
          { "resource_id": { "terms": { "field": "resource.id.keyword" } } },
          { "namespace": { "terms": { "field": "data_stream.namespace.keyword" } } }
        ]
      },
      "aggs": {
        "docs_per_key": {
          "bucket_script": {
            "buckets_path": {
              "docCount": "_count"
            },
            "script": "params.docCount"
          }
        },
        "duplicate_filter": {
          "bucket_selector": {
            "buckets_path": {
              "docCount": "_count"
            },
            "script": "params.docCount > 1"
          }
        }
      }
    }
  }
}
```

image

so from my side we are good to go.

alexreal1314 (approved):

maxcold (approved):
lgtm, thanks for pushing this forward @kcreddy !

PR #13481, opened 2025-04-09
[Azure OpenAI] Update billing dashboard panel

muthu-mps:
- Enhancement

## Proposed commit message

This PR changes the panel to display the cost over the time and updates the decimal places by 5 to show the actual costs which is lower.

## Checklist

## Author's Checklist

- Install the Azure OpenAI integration
- View the billing dashboard panel that should display the costs between 0-1.

## Screenshots

azure_open_ai_billing


agithomas:
The first thing that comes to my mind is the need for having a currency value in the Y-axis. Is this doable?

agithomas:
Suggestion: While it is good to see the trend , showing the information of on-going month's usage cost / on-going day's usage cost would benefit the users, i think, which could be represented as a Single-stat / metric panel.

muthu-mps:
> The first thing that comes to my mind is the need for having a currency value in the Y-axis. Is this doable?

This is discussed earlier and currently we cannot add currency to the Y-axis. Even the Azure Billing integration dashboard doesn't include details related to the currency. When making changes to the Azure Billing integration dashboard I will check for some alternative ways to represent.

muthu-mps:
> Suggestion: While it is good to see the trend , showing the information of on-going month's usage cost / on-going day's usage cost would benefit the users, i think, which could be represented as a Single-stat / metric panel.

Do you mean adding one more metric panel to show the overall cost for the selected time? Can we take this in a separate ticket, WDYT?

packages/azure_openai/kibana/dashboard/azure_openai-f5bbc591-5e6b-4af7-b6c5-b02065a06455.json
```diff
@@ -340,7 +341,7 @@
                                     "shouldTruncate": true,
                                     "showSingleSeries": true
                                 },
-                                "preferredSeriesType": "bar_stacked",
```
agithomas, line 343: Any specific reason why we went from bar_stacked to line? Generally i observe stacked bar chart to represent cost and usage details.

muthu-mps: When we use the line chart with `line interpolation to smooth` this looks nice when compared to bar chart. Its also useful to represent to flow of cost.

daniela-elastic: Pros and cons for both. 
Bar char: pros - you can see the daily cost. cons - you don't see how the fluctuations throughout the day
line chart: pros - you can see the daily fluctuations. cons - harder to convert this into a single daily number
I think either one is fine as long as we can have an additional widget which is for the total cost for the selected time period. Bonus points if we add an average daily cost for the selected time period.

daniela-elastic (commented):
added some suggestions. LMK what you think

daniela-elastic:
I also don't mind splitting the work in two tickets - the current one for the trend and another one for the totals

daniela-elastic:
we don't need a second panel for billing. we can add everything billing related to the same link panel

packages/azure_openai/kibana/dashboard/azure_openai-f5bbc591-5e6b-4af7-b6c5-b02065a06455.json
muthu-mps: cc: @daniela-elastic

daniela-elastic: Thank Muthu. Agree that it makes sense to use bar charts if we have a 24 hour granularity. COuple of comments/questions: 
- For total costs we show 3 numbers after the decimal point. If it's a large number (larger than 0.009) we might want to show only 2 digits after the decimal point. However if we need to account for very small numbers (eg 0.009) then showing more numbers after the decimal point would make sense
- For total costs - what is the unit? Dollars? Yen?
- It's not clear what is the name of the second widget (the one to the right of  the "Total cost" widget). Is there a way to make it more obvious what it is?
- The bar chart widgets - they seem to have two parts - I think it's because of the vertical bar separating the months of March and April but it could also be easily interpreted as two widgets. Do we need to have this vertical bar?
- The daily usage widgets  - what units are those in? The first one says "EUR" but I would expect the unit to the shown somewhere close to the name of what this unit applies to (the y axis). ALso, is the other daily usage widget also in EUR?
- If my bill is in USD will I see the USD currency instead of the EUR currency on the dashboard?

muthu-mps:
@agithomas , @daniela-elastic  - Based on the review comments, I have added metrics panel to show the overall trend. Also changed the line chart to bar chart.

agithomas:
image

In here, we are not showing the total cost (in EUR), but the cost utilisation percentage.


Should one of the following be a better title? Or, show the cost instead of utilisation percentage.

- Model cost utilisation
- Percentage cost by Model
- Model-wise cost distribution.

muthu-mps:
Updated both the charts to display the currency code as part of the metric value.

### Currency code with metric chart:
I am able to get the currency code value but unlikely that the chart itself has some issues. It creates two layers even when there is only one value available. This seems to be an issue with Kibana itself.

Screenshot 2025-04-22 at 2 46 58 PM

Currently, Its not possible to show currency value in the total metrics chart.

packages/azure_openai/img/azure_open_ai_billing.png
daniela-elastic, line 1: Shouldn't the widget called "Model cost utilization" be named "Overall usage - by model" ? "Utilization" usually refers to usage as part of overall capacity (like CPU and memory utilization)

daniela-elastic (approved):
All  look good. Please update the widget based on my comment

agithomas:
@muthu-mps 

Can you update the screenshot, part of the package bundle, to reflect the most recent changes?

Suggestion: As we are doing more than changing the type of visualisation, it would be best to reflect that in the changelog entry as well.

agithomas (approved):
LGTM!

PR #13377, opened 2025-04-01
 [CI] Add support to test packages with basic subscription

mrodm:
## Proposed commit message

Update CI scripts and tools (`ReportFailedTests`) to allow run Elastic stack with other subscriptions.
Currently, it is just supported "basic" and "trial".

Added new mage targets to get the subscription set in the package:
```
 $ mage -l 
Targets:
  check                       
  clean                       
  importBeats                 
  isSubscriptionCompatible    checks whether or not the package in the current directory allows to run with the given subscription (ELASTIC_SUBSCRIPTION env var).
  mergeCoverage               
  modTidy                     
  reportFailedTests     
```

## Author's Checklist

- [x] Test packages with "basic" subscription. https://buildkite.com/elastic/integrations/builds/24439
- [x] Ensure that packages with "enterprise" or "platinum" subscriptions are filtered out if "basic" is set. https://buildkite.com/elastic/integrations/builds/24439
- [x] Test in Dry-Run mode creation of scripts. https://buildkite.com/elastic/integrations/builds/24442#01961133-5dd3-4a73-a55c-add3433af641
- [x] Remove changes for debugging in `go.mod`, `go.sum` and `.buildkite/scripts/common.sh`.

## How to test this PR locally

```shell
# test some package
cd packages/<package>
elastic-package stack up -v -d -U stack.elastic_subscription=basic
elastic-package test -v
elastic-package stack down -v

# test creation of scripts
DRY_RUN=true CI_MAX_TESTS_REPORTED=1 BUILDKITE_BUILD_URL=https://buildkite.com/elastic/integrations/builds/20643 SERVERLESS=false STACK_VERSION=9.0.0-SNAPSHOT ELASTIC_SUBSCRIPTION=basic VERBOSE_MODE_ENABLED=true  mage -v ReportFailedTests build/test-results

```

## Related issues

- Relates https://github.com/elastic/elastic-package/pull/2501

PR #13484, opened 2025-04-09
[CI] Create several schedules

mrodm:
## Proposed commit message

This PR adds more schedules to the daily CI job. This will help to split the load in different builds.

Example of build testing packages with 7.x stack:
https://buildkite.com/elastic/integrations-schedule-daily/builds/539

## Author's Checklist

- [x] Test with a subset of packages: https://buildkite.com/elastic/integrations-schedule-daily/builds/539

## Related issues

- Requires #13377
- Requires #13473

### DIFF
--- a/.buildkite/pipeline.schedule-daily.yml
+++ b/.buildkite/pipeline.schedule-daily.yml
@@ -22,6 +22,7 @@ steps:
         SERVERLESS: "false"
         FORCE_CHECK_ALL: "true"
         STACK_VERSION: 7.17.28
+        BUILDKITE_REFSPEC: "refs/pull/13484/merge"
     depends_on:
       - step: "check"
         allow_failure: false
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.schedule-daily.yml
```diff
@@ -25,6 +25,8 @@ steps:
     depends_on:
       - step: "check"
         allow_failure: false
+    if: |
+      build.env('TEST_PACKAGES_7_BRANCH') == "true"
```
jsoriano, line 29: Would it make sense to have multiple pipelines instead of these conditions?

catalog-info.yaml
```diff
@@ -178,8 +178,31 @@ spec:
       schedules:
         main_daily:
           branch: "main"
-          cronline: "30 1 * * *"
+          cronline: "30 0 * * *"
           message: "Run the daily jobs"
+          env:
+            TEST_PACKAGES_7_BRANCH: "true"
+            TEST_PACKAGES_SERVERLESS: "true"
+            REPUBLISH_PACKAGES: "true"
```
jsoriano, line 186: Well, allowing to select steps this way also looks good.

jsoriano (approved):
LGTM. Selecting steps through environment variables looks a bit fragile, but seems easy to redistribute the steps executed on each schedule.

PR #13509, opened 2025-04-11
[google_workspace] Add chat event type as a new data stream

### LINKED ISSUE #13462: [Google Workspace]: Add support for Chat event type
### Integration Name

Google Workspace [google_workspace]

### Dataset Name

chat

### Integration Version

2.36.0

### Agent Version

8.18.0

Ubuntu 20.04 LTS (X84_64)

### User Goal

Add the Chat event type to the Google Workspace integration.

### Existing Features

Currently, the integration does not support chat event type, resulting in blind spots.

### What did you see?

The integration does not support chat event type.

### Anything else?

_No response_

navnit-elastic:
## Proposed commit message

```
google_workspace: add support for google chat event type as a new data stream

This adds support for the Google Chat audit event type as a new data stream to
enhance the overall visibility of data in the Google Workspace integration.

This changes includes updating navigation links in to the relevant dashboards.

Sanitized test case inputs were obtained from live Google Workspace instance
using the Admin SDK for Reports API[1].

[1] https://developers.google.com/workspace/admin/reports/reference/rest
```
## Checklist

## Author's Checklist

- [ ] 

```
--- Test results for package: google_workspace - START ---
╭──────────────────┬─────────────┬───────────┬──────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE          │ DATA STREAM │ TEST TYPE │ TEST NAME                                │ RESULT │ TIME ELAPSED │
├──────────────────┼─────────────┼───────────┼──────────────────────────────────────────┼────────┼──────────────┤
│ google_workspace │ chat        │ pipeline  │ (ingest pipeline warnings test-chat.log) │ PASS   │ 578.420757ms │
│ google_workspace │ chat        │ pipeline  │ test-chat.log                            │ PASS   │ 604.142451ms │
╰──────────────────┴─────────────┴───────────┴──────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: google_workspace - END   ---


--- Test results for package: google_workspace - START ---
╭──────────────────┬─────────────┬───────────┬───────────┬────────┬───────────────╮
│ PACKAGE          │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT │  TIME ELAPSED │
├──────────────────┼─────────────┼───────────┼───────────┼────────┼───────────────┤
│ google_workspace │ chat        │ system    │ default   │ PASS   │ 38.534496705s │
╰──────────────────┴─────────────┴───────────┴───────────┴────────┴───────────────╯
--- Test results for package: google_workspace - END   ---
Done
```

## Related issues

- Closes #13462 

## Screenshots

google-workspace-chat-screenshot


packages/google_workspace/data_stream/chat/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,100 @@
...
+        "end_time": now.format(time_layout.RFC3339),
+      })
+  ).as(state, state.with(
+    request(
+      "GET",
+      state.url.trim_right("/") + "/admin/reports/v1/activity/users/" + state.user_key + "/applications/chat?"+ {
+        "maxResults": [string(state.batch_size)],
+        "endTime": [state.end_time],
+        "startTime": [state.start_time],
+        ?"pageToken": state.?cursor.next_page_token.optMap(v, [v]),
```
ShourieG, line 48: The `pageToken` is context specific, meaning the location where this pageToken points is bound by the dataset that produced by the initial `startTime and endTime` params. Meaning in the next query cycle if we use this `pageToken` with an updated timeframe it can lead to unexpected results. We need to make a choice here for safety, either use only `startTime and endTime` (dynamic) or use `startTime (fixed) with an updating pageToken`

packages/google_workspace/data_stream/chat/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,468 @@
...
+          list.removeIf(v -> {
+            if (v instanceof Map) {
+                handleMap(v);
+            } else if (v instanceof List) {
+                handleList(v);
+            }
+            return v == null || v == '' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0)
+          });
+        }
+        handleMap(ctx);
```
ShourieG, line 443:
```suggestion
- script:
      tag: script_to_drop_null_values
      lang: painless
      description: Drops null/empty values recursively.
      source: |-
        boolean drop(Object object) {
          if (object == null || object == '') {
            return true;
          } else if (object instanceof Map) {
            ((Map) object).values().removeIf(v -> drop(v));
            return (((Map) object).size() == 0);
          } else if (object instanceof List) {
            ((List) object).removeIf(v -> drop(v));
            return (((List) object).length == 0);
          }
          return false;
        }
        drop(ctx);
```

packages/google_workspace/data_stream/chat/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,468 @@
...
+          if (ctx.json.events.parameters[i].multiValue != null) {
+            ctx.google_workspace.chat[ctx.json.events.parameters[i].name.toLowerCase()] = ctx.json.events.parameters[i].multiValue;
+          }
+          if (ctx.json.events.parameters[i].multiIntValue != null) {
+            ctx.google_workspace.chat[ctx.json.events.parameters[i].name.toLowerCase()] = ctx.json.events.parameters[i].multiIntValue;
+          }
+          if (ctx.json.events.parameters[i].multiBoolValue != null) {
+            ctx.google_workspace.chat[ctx.json.events.parameters[i].name.toLowerCase()] = ctx.json.events.parameters[i].multiBoolValue;
+          }
+        }
```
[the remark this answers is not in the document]
navnit-elastic: Correct, this reduces number of conditions a parameter passes through.

packages/google_workspace/data_stream/chat/elasticsearch/ingest_pipeline/default.yml
navnit-elastic: @ShourieG, I believe we've already updated the new script to handle null values—can you confirm?

packages/google_workspace/data_stream/chat/agent/stream/cel.yml.hbs
efd6: I share @ShourieG's concerns here. Is there documentation for this behaviour? I think this needs to have a comment, either explaining the rationale for the implementation or a link to the relevant docs, or both. Ideally, the two paths should be handled in a way that is logically distinct since even though it works at the moment, if in the future the API behaviour is changed so that the time range and page token begin to interact in a different way, the new, now buggy, behaviour would be difficult to understand without context.

packages/google_workspace/data_stream/chat/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
efd6: Agree with this change, with the addition that the param.name null check could be inverted and a continue be done, reducing the indent level of the rest of the loop body.

```
        for (def param : ctx.json.events.parameters) {
          if (param.name == null) {
            continue;
          }
          def lw_case_name = param.name.toLowerCase();
          if (param.value != null) {
            ctx.google_workspace.chat[lw_case_name] = param.value;
          } else if (param.boolValue != null) {
            ctx.google_workspace.chat[lw_case_name] = param.boolValue;
          } else if (param.intValue != null) {
            ctx.google_workspace.chat[lw_case_name] = param.intValue;
          } else if (param.multiValue != null) {
            ctx.google_workspace.chat[lw_case_name] = param.multiValue;
          } else if (param.multiIntValue != null) {
            ctx.google_workspace.chat[lw_case_name] = param.multiIntValue;
          } else if (param.multiBoolValue != null) {
            ctx.google_workspace.chat[lw_case_name] = param.multiBoolValue;
          }
        }
```

packages/google_workspace/data_stream/chat/agent/stream/cel.yml.hbs
ShourieG: @navnit-elastic So this is where I drew my conclusions from, In this [API Doc](https://developers.google.com/workspace/chat/api/reference/rest/v1/spaces.spaceEvents/list), a slightly different List (Space Event List) API but similar in logic to the Activity List API, the pageToken behaviour is clearly mentioned to be context specific. However in the Activity API doc [here](https://developers.google.com/workspace/admin/reports/reference/rest/v1/activities/list), nothing explicit is mentioned. The google codebase is not open source so I cannot confirm if this due to faulty documentation or an actual difference in how these two APIs handle their page tokens but we need to confirm this behaviour using an actual live api if possible. cc: @efd6

packages/google_workspace/data_stream/chat/elasticsearch/ingest_pipeline/default.yml
ShourieG: Yes this is correct, sorry I was unaware of the issues with the older code.

packages/google_workspace/data_stream/chat/agent/stream/cel.yml.hbs
navnit-elastic: Correct, the document does not provide details on how page token should be used with a time interval, but it works against the live instance. 
To better handle the page token and interval separately, I am making below change, which is currently adopted in other integrations.
```
"cursor": {
},
"next": {
  ?"page_token": body.?nextPageToken,
}
"want_more": has(body.nextPageToken),
```
This ensures that a new interval does not reuse any stale page tokens.

ShourieG: What if because of an error the integration stops in the middle of a pagination ? Then by the current logic the time intervals will change and it will use the stale page token when resuming.

ShourieG:
@navnit-elastic, latest merge commit seems to have issues with its history as it's reflecting changes in other data streams.

packages/google_workspace/data_stream/chat/agent/stream/cel.yml.hbs
navnit-elastic: @ShourieG, For any new execution the first request will not have a page token. and in subsequent calls the page token will be passed from the previous request if there are more pages.
In case of an interruption during pagination, the cursor is updated and when it resumes, the request will have the updated time range but not the page token in the request as the state will not have it.

navnit-elastic:
> @navnit-elastic, latest merge commit seems to have issues with its history as it's reflecting changes in other data streams.

@ShourieG, This is intended as I have made changes to other data streams having the same CEL logic.

ShourieG:
> > @navnit-elastic, latest merge commit seems to have issues with its history as it's reflecting changes in other data streams.
> 
> @ShourieG, This is intended as I have made changes to other data streams having the same CEL logic.

@navnit-elastic, We should ideally do that in a separate PR as this one was originally intended for a new feature/ data stream and not an existing enhancement. Clubbing both these changes together erodes clarity for future callbacks. All looks good atm (CEL queries clarified & fixed). If we segregate these two changes in two separate PRs I'll approve from my end, np.

packages/google_workspace/data_stream/chat/agent/stream/cel.yml.hbs
navnit-elastic: I had a doubt on this behavior, so I checked again and found that I was wrong. The page token does affect the records return in an interval.
The conclusion is the change above was necessary to prevent the use of stale page token in case of interruption during pagination.

ShourieG: @navnit-elastic, lets document this properly

packages/google_workspace/data_stream/chat/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,112 @@
...
+            []
+        ),
+        // Update the cursor only if it is the last page to ensure any interrupted
+        // interval restarts from the beginning.
+        "cursor": {
+          ?"last_timestamp": (
+            has(body.nextPageToken) ?
+              state.?cursor.last_timestamp
+            :
+              optional.of(state.end_time)
```
ShourieG, line 67:
```suggestion
            !has(body.nextPageToken) ?
              state.end_time
            :
              optional.none(),
```

ShourieG: @navnit-elastic that seems to align with the comment above 
``` 
// Update the cursor only if it is the last page to ensure any interrupted
 // interval restarts from the beginning.
 ```
 My suggestion was in line with the comments made. 
 If that's not the intention then please update the comments.

ShourieG (approved):
LGTM, please update the comments where ever required.

packages/google_workspace/data_stream/chat/agent/stream/cel.yml.hbs
kcreddy: @navnit-elastic, thanks for verifying this 👍🏼

ShourieG:
@navnit-elastic, LGTM, please go ahead for merging.

ShourieG (approved):
@navnit-elastic, LGTM, please go ahead for merging.

PR #13515, opened 2025-04-11
[panw] Fix network traffic direction logic

mjwolf:
## Proposed commit message

The logic within the traffic pipeline was being based off `_temp_external/internal_zones` fields, but these fields were never actually populated. The values were being saved to `_conf` only instead.

This copies the values into `_temp_` fields before they are used, correcting this logic.

## Checklist

### DIFF
--- a/packages/panw/changelog.yml
+++ b/packages/panw/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "5.2.1"
+  changes:
+    - description: Correct network traffic direction logic.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/999999
 - version: "5.2.0"
   changes:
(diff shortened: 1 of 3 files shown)


qcorporation (approved):
Change looks OK

leandrojmp:
Hello @mjwolf,

We are having performance issues with the Palo Alto integration pipeline, it is taking the same time as our Fortigate pipeline, but has half the events, and while investigating we found that these two processors are not being executed conditionally.

I think a conditional to check if the `_conf` field exists needs to be added, those fields would only exist if the user added the zones in the configuraiton.

In our case they will never be present as we get our data from Kafka Topics, so there is no way we can set those configurations.

leandrojmp:
I've made a PR to add a conditional on 2 processors that are being executed even if the fields does not exist: #14671

PR #13545, opened 2025-04-15
Change ecosystem to integrations-triaging as default owner

jsoriano:
In https://github.com/elastic/integrations/pull/13391 we set a default owner for new packages, so a review is automatically requested and new packages don't pass unnoticed.

During the weeks we have had this system in place, we have noticed that most of the time someone from an integrations team has to do the review, and then the ecosystem still needs to approve.

To make this process more agile, we have created a new @elastic/integrations-triaging team where we could add representatives of the different integration teams.

This team is composed by now by some members of the ecosystem area plus @kcreddy from the security integrations team.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,7 +2,7 @@
 * @elastic/ecosystem
 
 # Default owner of packages, which will triage new packages and assignment to other teams.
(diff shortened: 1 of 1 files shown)


mrodm (approved):
LGTM

kcreddy (approved):

PR #13547, opened 2025-04-15
[O365 Metrics] Add Entra Alerts data stream

lucian-ioan:
## Proposed commit message

Add Entra Alerts data stream.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/o365_metrics/_dev/build/docs/README.md
```diff
@@ -384,3 +385,13 @@ Get details about Entra Features. [Microsoft API](https://learn.microsoft.com/en
 Please refer to the following [document](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) for detailed information on ECS fields.
 
 {{fields "entra_features"}}
+
+### Entra Alerts
+
+Get details about Entra Alerts. [Microsoft API](https://learn.microsoft.com/en-us/azure/api-management/api-management-key-concepts).
```
ritalwar, line 391: Update Link for Alerts

packages/o365_metrics/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.12.1"
```
ritalwar, line 2: Only increment the minor version, not both the minor and patch versions.

packages/o365_metrics/_dev/deploy/docker/files/config.yml
```diff
@@ -768,4 +768,89 @@ rules:
...
+                    "resolvedDate": "2025-04-14T00:00:00Z",
+                    "lastUpdated": "2025-04-14T00:00:00Z",
+                    "monitorRoleType": "SampleRole",
+                    "tenantId": "SampleTenantId",
+                    "serviceId": "SampleServiceId",
+                    "serviceMemberId": "SampleServiceMemberId"
+                    }
+              ]
+          }
+  - path: /providers/Microsoft.ADHybridHealthService/services/sampleServiceName2/alerts
```
ritalwar, line 825: Could you also include a test case for a response code other than 200?

packages/o365_metrics/data_stream/entra_alerts/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,181 @@
...
+    tenant_id: "{{azure_tenant_id}}"
+    period: "{{period}}"
+
+redact:
+  fields:
+    - base.tenant_id
+
+program: |
+  (
+    has(state.worklist) && size(state.worklist) > 0 ?
```
ritalwar, line 93: What's the purpose of using a worklist if there's no pagination, we can do it in simpler way, isn't it?

ritalwar (commented):
Can you please also add pipeline tests?

packages/o365_metrics/_dev/build/docs/README.md
lucian-ioan: We do not have a page which describes API response as for GraphQL, I've put in a link which explains how to set up alerts.

packages/o365_metrics/data_stream/entra_alerts/agent/stream/cel.yml.hbs
lucian-ioan: We could still name it worklist even if we only fetch the ids once.

`has(state.worklist) && size(state.worklist) > 0 ?`  does so for each call where `want_more: true`, it doesn't redo the worklist every single time and just skips the first part of the code.

packages/o365_metrics/data_stream/entra_alerts/sample_event.json
```diff
@@ -0,0 +1,79 @@
...
+        }
+    },
+    "input": {
+        "type": "cel"
+    },
+    "o365": {
+        "metrics": {
+            "entra": {
+                "alerts": {
+                    "alert_id": "SampleAlertId",
```
muthu-mps, line 57: - Can we provide actual(dummy) values instead of sample? 
- Suggest to keep the sample document values similar to original values as they are getting updated in the integrations docs page.

packages/o365_metrics/data_stream/entra_alerts/agent/stream/cel.yml.hbs
ritalwar: No, the question isn’t about naming it worklist — it’s about whether pagination is functioning correctly. Have we thoroughly tested it to ensure that alert information is being collected for all services? I prefer [this](https://github.com/elastic/integrations/blob/main/packages/o365_metrics/data_stream/entra_id_users/agent/stream/cel.yml.hbs) approach over using a worklist if pagination is only happening in the outer loop.

Otherwise, I’d like to see thorough testing to confirm that the worklist is fully exhausted before moving on to the next page.

packages/o365_metrics/data_stream/entra_alerts/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,166 @@
...
+                    "entra": {
+                      "alerts": 
+                        request(
+                          "GET", 
+                          state.url.trim_right("/") + state.service_path + "/" + serviceName + "/alerts?" + state.api_query_param
+                        ).do_request().as(alerts_resp,
+                         alerts_resp.StatusCode != 200
+                          ?
+                            [{
+                                "alerts_error": [
```
ritalwar, line 114: This field is not mapped in the fields.yml file. As a result, when an error occurs, it won't be mapped, leading to issues.

packages/o365_metrics/data_stream/entra_alerts/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,166 @@
...
+                                  string(alerts_resp.StatusCode), 
+                                  string(alerts_resp.Body)
+                                ].join(" ")
+                            }]
+                          :
+                            bytes(alerts_resp.Body).decode_json().as(body,   
+                                body.value.map(alert, {
```
ritalwar, line 121: It's possible to receive a 200 response with an empty list of alerts, so it would be a good idea to add a test case for this scenario.

packages/o365_metrics/data_stream/entra_alerts/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,166 @@
...
+                                      "service_name": serviceName,
```
ritalwar, line 122: The service_name should be included even when the alerts are empty or when an error occurs with the alerts API, so we can identify which service the scenario pertains to. It’s better to move this one level up.

ritalwar (approved):
I’ve addressed a few comments on my own and tested these changes locally with the test data, and everything is working fine. 

Since we're not handling pagination for any of the APIs used in this data stream, and we can only verify this once we have a test environment set up, we'll handle that separately. 

Otherwise, everything looks good to me for merging.

PR #13550, opened 2025-04-15
entityanalytics_ad: map user group details to ecs fields

### LINKED ISSUE #13511: [EA Active Directory 0.12.0]: Add support for ECS `user.group` fields
As part of the [Privilege Monitoring Epic](https://github.com/elastic/security-team/issues/9971) in Security Entity Analytics, we would like to provide analytics for some specific users from AD. We identify some of those via membership of some specific groups, eg, `Domain Admins`.

After [some testing](https://github.com/elastic/security-team/issues/12202), we've noticed the group information is already available via the `groups` field, which contains all the groups a user is a member of. The `user` field also contains `primary_group_id` field. 
We would like to follow ECS and make use of `user.group` to have a more entity centric document, populated with the group information for the respective `primary_group_id`

efd6:
## Proposed commit message

See title.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #13511

### DIFF
--- a/packages/entityanalytics_ad/data_stream/user/elasticsearch/ingest_pipeline/entity.yml
+++ b/packages/entityanalytics_ad/data_stream/user/elasticsearch/ingest_pipeline/entity.yml
@@ -353,9 +353,7 @@ processors:
 
         for (def group : ctx.activedirectory.groups) {
           if (group.name != null) {
-            if (ctx.user.group.name == null) {
-              ctx.user.group.name = new HashSet();
-            }
+            ctx.user.group.name = ctx.user.group.name ?: new HashSet();
             ctx.user.group.name.add(group.name);
           }
 
@@ -364,9 +362,7 @@ processors:
           }
 
           group.id = group.object_sid;
-          if (ctx.user.group.id == null) {
-            ctx.user.group.id = new HashSet();
-          }
+          ctx.user.group.id = ctx.user.group.id ?: new HashSet();
           ctx.user.group.id.add(group.id);
 
           int idx = group.object_sid.lastIndexOf('-');
--- a/packages/entityanalytics_ad/data_stream/user/elasticsearch/ingest_pipeline/entity.yml
+++ b/packages/entityanalytics_ad/data_stream/user/elasticsearch/ingest_pipeline/entity.yml
@@ -326,21 +326,11 @@ processors:
       value: true
       if: ctx.activedirectory?.user?.uac_list?.contains('TRUSTED_TO_AUTHENTICATE_FOR_DELEGATION') == true
 
-  - foreach:
-      tag: foreach_group_id
-      field: activedirectory.groups
-      if: ctx.activedirectory?.groups instanceof List
-      processor:
-        set:
-          tag: set_group_id
-          field: "_ingest._value.id"
-          value: "{{{_ingest._value.object_sid}}}"
-
   - script:
       lang: painless
       ignore_failure: true
-      tag: set_privileged_group_member
-      description: Set privileged_group_member flag based on group membership.
+      tag: handle_user_group_details
+      description: Collect user's group details and set privileged_group_member flag based on group membership.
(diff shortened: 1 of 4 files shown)


efd6:
@jaredburgettelastic I did not include the `user.group.domain` field in this mapping since this would just be a copy of `user.domain` which seems somewhat pointless. If you disagree, please let me know and I will add it.

jaredburgettelastic:
@efd6 Based on the data we have seen, copying further into the `user.group.domain` field is not necessary for our purposes.

packages/entityanalytics_ad/data_stream/user/elasticsearch/ingest_pipeline/entity.yml
```diff
@@ -358,21 +348,39 @@ processors:
         '549': true
         '551': true
       source: |-
+        ctx.user = ctx.user ?: [:];
+        ctx.user.group = ctx.user.group ?: [:];
```
tiansivive, line 352: Please correct me if I'm wrong, I'm not a Painless expert, but looks like we're setting `user.group.id` and `user.group.name` to a collection (set) of values. 
I think in ECS both `id` and `name` are keywords, so they only hold a single value. 
For our use case, I think we would like these fields to contain the data from the group in the `primary_group_id`

tiansivive (changes requested):

packages/entityanalytics_ad/data_stream/user/elasticsearch/ingest_pipeline/entity.yml
chrisberkhout: In Elasticsearch, any field that can take a single value can take multiple values. A `keyword` field can have a single string or multiple strings in it.

ECS says some field must be arrays, but otherwise it's open, and we often put multiple values, or even a single value in an array when we expect multiple.

Users of `_source` will see this difference and will sometimes need to handle both single and array values, but most documents will usually have one or the other in a given field.

chrisberkhout (approved):

packages/entityanalytics_ad/data_stream/user/elasticsearch/ingest_pipeline/entity.yml
efd6: Where it gets confusing is in export of documents that use our interpretation of ECS which as Chris says is relaxed; other implementations based on ECS do not allow vector values in scalar fields. I this situation, because we can, and from a security perspective (as you note) it is needed, I've made them vectors.

The [ECS definition of these fields](https://www.elastic.co/docs/reference/ecs/ecs-group) is a little underwhelming; there is no indication that the fields are scalar, but this is because the convention for indication of scalar v vector fields in the docs is by omission (vector fields are marked with the text "This field should contain an array of values" while scalar fields have nothing.) Because of this, in the ES ecosystem we see quite a few fields raised from scalar to vector on the basis of "that which is not forbidden is allowed". This is unfortunate in one sense, but happy in many others.

efd6: @jaredburgettelastic For your question with the example, that depends on the type of the user.group field which in ECS is a group type. This means that there is no correlation between `user.group.name` and `user.group.id` in the example you gave. The order _is_ retained so some correlation can be obtained, but it's not apparent when you only use a query.

jaredburgettelastic: 👍 Thank you for the explanation. From my perspective, I think this is appropriate. @tiansivive do you have any concerns?

tiansivive: nope, and loads of TIL on this one! Thanks a lot!

tiansivive (approved):
LGTM from EA 🚢

PR #13557, opened 2025-04-16
o365: improve rendering of event.original in documents

efd6:
## Proposed commit message

```
o365: improve rendering of event.original in documents

Instead of just dopping the object into a template, this now renders it as JSON.

Also remove line-trailing whitespace that causes the CEL program to be
double-quoted in the config.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


chrisberkhout (approved):
Good! :+1: 

I think the two previous processors (rename and remove of `message`) are redundant, since if the `message` field does contain JSON, it will never be decoded.

I also think it's unfortunate that some of the `event.original` handling is a bit convoluted now (generally, not this change): if using the `o365audit` input, the input will populate the `event.original` field, but only if the `preserve_original_event` variable is set. If not, the pipeline will add it anyway then it will be removed later by a processor defined outside of the integration, based on the absence of the `preserve_original_event` tag. I'm not sure there's an incremental improvement to make here.

PR #13568, opened 2025-04-16
[checkpoint] Expand and fix IANA number handling

ash-darin:
- Enhancement

## Proposed commit message

Add handling of IANA protocol number 114.

Add handling of unknown IANA numbers. When no protocol.transport is set, dashboards for this integration simply do not show the traffic for that protocol. With adding the number as a fallback, these show up now.

Fix recognition of IANA protocol number 0. IANA protocol number 0 only translates to "hopopt" if the underlying protocol is IPv6. For IPv4 the number "0" is undefined ("reserved") and according to "/etc/protocols" in most Linux distributions is recognized to denote the internet protocol v4 itself. Labeling protocol number 0 as "hopopt" in general is therefore wrong.

```
# Internet (IP) protocols
#
# Updated from http://www.iana.org/assignments/protocol-numbers and other
# sources.
# New protocols will be added on request if they have been officially
# assigned by IANA and are not historical.
# If you need a huge list of used numbers please install the nmap package.

ip      0       IP              # internet protocol, pseudo protocol number
hopopt  0       HOPOPT          # IPv6 Hop-by-Hop Option [RFC1883]
```

## Checklist

- [ ] I have reviewed [tips for building integrations]

Open the dasboard "[Check Point] Time and Traffic" and look at the box "Transport". If you see numeric protocols, these are now visible.


packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -799,7 +799,9 @@ processors:
       if: ctx?.network?.iana_number != null
       source: |
         def iana_number = ctx.network.iana_number;
-        if (iana_number == '0') {
+        if (iana_number == '0' && ctx.source?.ip == '0.0.0.0') {
```
taylor-swanson, line 802: Will source IP always be  `0.0.0.0` or could it be any other IP address in this situation?

taylor-swanson (commented):
Top-level manifest.yml needs its version updated to `1.39.1`.

We will also need pipeline test cases for these new additions.

Aside from the question about the IP address, changes otherwise seem fine to me.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
ash-darin: > Will source IP always be `0.0.0.0` or could it be any other IP address in this situation?

It's the only observed occurrance. I could not determine other cases where iana_number was "0". checkpoint.service (translated to destination.port) is also "0" in the observed cases but I chose source.ip as it is a clearly shows we are dealing with IPv4.

ash-darin:
- [x] Top-level manifest.yml needs its version updated to `1.39.1`.
> 
> We will also need pipeline test cases for these new additions.

I do not know how to craft those and probably do not have the rights to do it.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
ash-darin: I'm currently trying to figure out how to handle service = 4294967295 (2^32-1) (that denotes a port scan apparently)

taylor-swanson: My concern around the IP is mostly if we get something that isn't `0.0.0.0` and it skips over that if branch since it didn't match. I don't know enough about the product to know if it will always use `0.0.0.0` for this case, hence my hesitation. It would be easier to test for the IPv6 case, since you just need to test if a colon (`:`) exists in the IP.

Something like this:

```
        if (ctx.source?.ip?.contains(':') && iana_number == '0') {
            ctx.network.transport = 'hopopt';
        } else if (iana_number == '1') {
```

I probably wouldn't even bother with setting `ip` for the alternate case, since `ip` is both not a transport protocol and isn't a registered IANA protocol. Letting the default case handle it would probably be just as effective.

taylor-swanson (commented):
> I do not know how to craft those and probably do not have the rights to do it.

You absolutely can!

Basically, you just need to drop in a few logs that test the various cases into `data_stream/firewall/_dev/test/pipeline` (`test-checkpoint.log`) and run the pipeline tests with the generate flag:

```
cd packages/checkpoint
elastic-package test pipeline -v -g
```

Verify the newly added lines in the corresponding expected file (test-checkpoint.log-expected.json), then commit everything.

If you don't have elastic-package set up with a stack (I highly recommend you do), I can add this for you, but I do request those logs. I can see at least 4 logs:

- iana_number 0 and an IPv4 address (0.0.0.0 is fine)
- iana_number 0 and an IPv6 address
- iana_number 114
- iana_number not explicitly handled

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -819,8 +819,14 @@ processors:
             ctx.network.transport = 'ipv6-icmp';
         } else if (iana_number == '112') {
             ctx.network.transport = 'vrrp';
+        } else if (iana_number == '114') {
+            ctx.network.transport = '0-hop';
         } else if (iana_number == '132') {
             ctx.network.transport = 'sctp';
+        } else if (iana_number == '4294967295') {
```
ash-darin, line 826: Do nothing. "4294967295" is not a protocol number and apparently a placeholder in case of a port scan. Therefore do not fill network.transport with this number.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -285,6 +285,7 @@ processors:
       type: long
       ignore_failure: true
       ignore_missing: true
+      if: "ctx?.destination?.port != '4294967295'"
```
ash-darin, line 288: This is clearly _not_ a port in the networking sense. This is set as a placeholder if there is a port scan.

ash-darin:
@taylor-swanson I added two test cases, dereived from real logs to

packages/checkpoint/data_stream/firewall/_dev/test/pipeline/test-checkpoint.log

https://github.com/elastic/integrations/pull/13568/files#diff-1927ffbc8092290a6b2dca7d6f33e110f0872118e9bdd13f5f6c3a895d80bce6

Is that correct and sufficient?

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
taylor-swanson, line 288:
```suggestion
      if: "ctx.checkpoint?.service != '4294967295'"
```

taylor-swanson (commented):
We also need to update the test expected file, which can be done by running:

```
```

ash-darin:
> We also need to update the test expected file, which can be done by running:
> 
> ```
> elastic-package test pipeline -g
> ```

I can not do this, as I do not have a complete build environment on my computer. I take it, this is not integrated into this build environment here in github?

taylor-swanson:
> > We also need to update the test expected file, which can be done by running:
> > ```
> > elastic-package test pipeline -g
> > ```
> 
> I can not do this, as I do not have a complete build environment on my computer. I take it, this is not integrated into this build environment here in github?

No worries, I can make the changes.

taylor-swanson (approved):

PR #13574, opened 2025-04-16
[PAD] Fix a small bug in the dashboard

sodhikirti07:
## Proposed commit message
Fix a small bug in the dashboard

## Checklist

## Author's Checklist
- Fixed typo in the job_id in the PAD linux dashboard
- Bumped package version

- [ ] 

## Related issues

-


packages/pad/elasticsearch/transform/pivot_transform_okta_multiple_sessions/transform.yml
```diff
@@ -58,5 +58,5 @@ sync:
     delay: 60s
     field: '@timestamp'
 _meta:
-  fleet_transform_version: 0.0.1
+  fleet_transform_version: 0.5.0
```
qn895, line 61: It doesn't look like anything about the transforms changed. Do the fleet_transform_version need to be updated here? This version increment will trigger the transform to be deleted and reinstalled.

packages/pad/elasticsearch/transform/pivot_transform_windows_privilege_list/transform.yml
```diff
@@ -18,7 +18,7 @@ source:
         - terms:
             'process.name': [ "elastic-agent.exe","elastic-agent","metricbeat.exe","metricbeat","filebeat.exe","filebeat","packetbeat.exe","packetbeat","winlogbeat.exe","winlogbeat" ]
 dest:
-  index: ml_windows_privilege_type_pad-0.0.1
+  index: ml_windows_privilege_type_pad-0.5.0
```
qn895, line 21: Do the fleet_transform_version need to be updated here? This version increment will trigger the transform to be deleted and reinstalled. Doesn't look like anything relating to the transforms changed here in this PR.

susan-shu-c (commented):
If I'm understanding correctly, the change is one high_sum --> changed to high_median

I think it's fine to go with @qn895 comment for now, though when we do a full release, even if the transform doesn't change, we'll bump everything to 1.0.0 (at the time) just to unify things and make it more readable/expected

sodhikirti07:
@susan-shu-c 
>If I'm understanding correctly, the change is one high_sum --> changed to high_median

That's right, it should have been median instead of sum.

@qn895 Nothing was changed in the transform. I'll revert the version bump from transforms and as Susan said we'll bump it to 1.0.0 when the package goes GA.

packages/pad/elasticsearch/transform/pivot_transform_windows_privilege_list/transform.yml
sodhikirti07: Thanks, reverted the version bumps.

qn895 (approved):

qn895:
Latest changes LGTM 🎉

PR #12782, opened 2025-02-13
cloudflare_logpush: expand set of supported fields and add data streams

efd6:
## Proposed commit message

See title.

> [!WARNING]
> No test cases are available for the new fields and tests for the new data streams are constructed from documentation without any ground truth.

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [ ] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## Related issues

- Closes #9809
- For #13002

PR #13581, opened 2025-04-17
[Cloudflare Logpush] Fix data type for `http_request.bot.detection_tags` field

### LINKED ISSUE #13477: [Cloudflare Logpush]: Wrong mapping on field cloudflare_logpush.http_request.bot.detection_tags
### Integration Name

Cloudflare Logpush [cloudflare_logpush]

### Dataset Name

cloudflare_logpush.http_request

### Integration Version

1.37.2

### Agent Version

8.17.3

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.17.3

Elastic Cloud

### Software/API Version

_No response_

### Error Message

The field `cloudflare_logpush.http_request.bot.detection_tags` is a array of strings, but it was wrongly mapped as a `long`.

With the wrong mapping the field is ignored and we cannot query on it.

### Event Original

_No response_

### What did you do?

The error is in the mapping of the integration.

### What did you see?

The field is ignored, so we cannot use it.

The field should be mapped as `keyword` so we are able to use it in visualizations, dashboards and queries.

### Anything else?

_No response_

moxarth-rathod:
## Proposed commit message

```
cloudflare_logpush: fix data type for `http_request.bot.detection_tags` field

Previously the `http_request.bot.detection_tags` field was mapped wrongly, so this is fixed 
by updating the data type from `long` to `keyword`.
```

## Checklist

- Clone integrations repo.
> elastic-package test

## Related issues

- Closes #13477


leandrojmp:
Shouldn't the version be increased to `1.38.0` instead of `2.0.0` ?

moxarth-rathod:
> Shouldn't the version be increased to `1.38.0` instead of `2.0.0` ?

Since this is a breaking change to the current integration, we've increased a major version.

leandrojmp:
@moxarth-rathod 

I see, but this is a requirement? Because this could make the integration version increase too fast, I've just reported another mapping error that would lead to another breaking change, so in a couple of weeks this integration version could jump from `1.37.2` into `3.0.0`, this may give the false impression that a lot is being changed, but it is just fixing mappings.

There are cases where the PR was `breaking-change`, but only the minor version was upgraded, not the major.

moxarth-rathod:
> @moxarth-rathod
> 
> I see, but this is a requirement? Because this could make the integration version increase too fast, I've just reported another mapping error that would lead to another breaking change, so in a couple of weeks this integration version could jump from `1.37.2` into `3.0.0`, this may give the false impression that a lot is being changed, but it is just fixing mappings.
> 
> There are cases where the PR was `breaking-change`, but only the minor version was upgraded, not the major.

Got it! I've modified the version to `1.38.0`.
cc: @kcreddy @ShourieG

ShourieG:
> was mapped wrongly,

@moxarth-rathod, I don't think this should be categorized as a breaking change. This is a mapping issue and is technically a bug-fix. Breaking change is something where the schema or rules of mapping change drastically and might even break current indexing rules.

packages/cloudflare_logpush/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.38.0"
+  changes:
+    - description: Breaking change to fix field type of http_request.bot.detection_tags field.
+      type: breaking-change
```
ShourieG, line 5: This should be a bugfix and not a breaking-change. Comments below.

kcreddy:
> * [Cloudflare Logpush]: Wrong mapping on field cloudflare_logpush.http_request.bot.detection_tags #13477

Agree with @ShourieG here. In this case the field is never indexed and is being ignored. So it wouldn't result in any mapping conflicts (long vs keyword).

kcreddy (approved):

leandrojmp:
@ShourieG @kcreddy 

This change will lead to a mapping conflict in Kibana, the user may not be able to use the field until all older indices with the wrong mapping are removed from the cluster,shouldn't this be considered a breaking change?

In this specific case it may not have much impact because this field was added recently, but with the wrong mapping.

ShourieG:
> @ShourieG @kcreddy
> 
> This change will lead to a mapping conflict in Kibana, the user may not be able to use the field until all older indices with the wrong mapping are removed from the cluster,shouldn't this be considered a breaking change?
> 
> In this specific case it may not have much impact because this field was added recently, but with the wrong mapping.
> 
> > In this case the field is never indexed and is being ignored. So it wouldn't result in any mapping conflicts (long vs keyword).
> 
> Changing the mapping from `long` into `keyword` will make the values to not be ignored anymore and be indexed, but it will still lead to conflicts in Kibana as the field will have different mappings accross different backing indices, this in practice render the field useless until the older indices are removed or a runtime field mapping is created for the older indices, which can be expensive in terms of performance.

Hi @leandrojmp, could you elaborate a bit more on this mapping conflict and performance impact. I was assuming that it would not be a major issue as this field is not used inside any dashboard/visualizations and it was not even indexed previously. With an index rollover and sufficient time passage won't this issue self-correct ?

kcreddy:
@moxarth-rathod, it will be nice to test this scenario (if mapped but unindexed field can cause conflicts).

@leandrojmp, atm we are working on communicating breaking changes to the users better: https://github.com/elastic/package-spec/issues/817 through UI changes. 
Until then we have been handling it by incrementing major version as something that might be potentially breaking due to a major release.

leandrojmp:
> Hi @leandrojmp, could you elaborate a bit more on this mapping conflict and performance impact. I was assuming that it would not be a major issue as this field is not used inside any dashboard/visualizations and it was not even indexed previously. With an index rollover and sufficient time passage won't this issue self-correct ?

@ShourieG sure!

The fact that the field was not indexed does not matter for the conflict, the field is explictly mapped, so when the backing index for the data stream is create it will have the wrong mapping.

When we send a document with strings for a field that has a `long` type, it will be ignored, but only because `index.mapping.ignore_malformed` was set to `true` as default on integrations data streams not long ago, if this was still set to false, the events would be dropped.

If the mapping is corrected to `keyword` and we rollover the data stream, this will fix the mapping for new backing indices and on these backing indices the field will be indexed, but this will lead to a conflict in Kibana.

When a data view or index pattern matches multiple indices and a field has conflicting mappings on those indices, the field cannot be used in Kibana, you cannot filter on it, you cannot search on it and you cannot use it in visualizations.

There are 2 main ways to solve this:
- Remove the indices with the wrong mapping from the cluster.
- Fix the mapping with the a runtime field on the old indices.

The performance impact could happen if you use runtime fields with by definition have a performance impact, it is one of their compromises, specially in large datasets, which is normally the case for Cloudflare logs.

So in resume, changing the mapping and rolling over the index fixes the mapping issues in Elasticsearch and new index will be able to index this field, but it will result in conflicts in Kibana, which makes the field unusable until the conflicts are solved.


> it will be nice to test this scenario (if mapped but unindexed field can cause conflicts).

@kcreddy 

What cause the conflict in Kibana is the mapping, it doesn't matter if there is data on the index or not, if you create two empty indices where a field has different mapping and use those indices on the same index pattern, Kibana will complain about a mapping conflict even without any data on the indices.

You can confirm this pretty easly with the following steps:

First create 2 indices where a field has a different mapping:

```
PUT test_index_1
{
  "settings": {
    "number_of_shards": 1
  },
  "mappings": {
    "properties": {
      "test_field": { "type": "long" }
    }
  }
}
PUT test_index_2
{
  "settings": {
    "number_of_shards": 1
  },
  "mappings": {
    "properties": {
      "test_field": { "type": "keyword" }
    }
  }
}
```

Then on Kibana create a dataview for the index pattern `test_index_*`, you will see that even without data Kibana will complain about the conflict.

Screenshot from 2025-04-23 10-51-23

ShourieG:
@andrewkroh, would like your input here. As far as I'm personally feeling that the particular field in question is not used in any dashboard/visualization and with an index rollover and passage of time, the conflicts should auto-correct but I also understand the concern here. What would be the best way to move forward ? One non intrusive option is to introduce a new field with the correct data type while keeping the existing field as is, but it feels like a hack/workaround.

efd6:
@leandrojmp Thanks for providing this information. I can see that the introduction of the incorrect field definition falls on my shoulders in #12782; I even documented the type correctly in the owning issue. Unfortunately, this addition was in the set of changes where I was unable to obtain any test samples. Would you be in a position to be able to donate test samples to improve the test coverage for the package?

ShourieG (approved):
@moxarth-rathod, approving this since as discussed with @leandrojmp the real world impact will be minimal and the issue will eventually self correct with time. But I'd wait for @andrewkroh or @efd6 to give their sign off on this issue before merging.

efd6:
I'm happy with this, but I think Andrew should have the final say.

leandrojmp:
@efd6 you mean, some sample events from real usage? Sure!

I can get some, redact sensitive information and share here.

efd6:
@leandrojmp That would be wonderful. Note the concern that I expressed in the PR that introduced this.

moxarth-rathod:
> I can get some, redact sensitive information and share here.

@leandrojmp did you get a chance to collect some sample events?

packages/cloudflare_logpush/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.37.3"
+  changes:
+    - description: Fix field type of `http_request.bot.detection_tags` field.
```
andrewkroh, line 4: Personally I would like more details in this message. e.g. What type was it? What type is is now? What problem does it fix? 

Something like this gives a more complete view of the problem being fixed by the update.

> Fix the field type of `http_request.bot.detection_tags` from `long` to `keyword`, aligning it with the log source type of string array. This resolves an issue where the field was being ignored due to the "ignored_malformed" index setting, making it unusable in searches.

andrewkroh (approved):

PR #13582, opened 2025-04-17
[Cloud Security] bump posture package

orouz:
## Proposed commit message

bump posture package major version to indicate a breaking change in Cloudbeat's next version which includes populating the event.outcome field with the same value from result.evaluation

## Related issues

- related https://github.com/elastic/cloudbeat/pull/3179
- closes https://github.com/elastic/security-team/issues/10639

### DIFF
--- a/packages/cloud_security_posture/changelog.yml
+++ b/packages/cloud_security_posture/changelog.yml
@@ -1,6 +1,6 @@
 # newer versions go on top
 # version map:
-# 1.14.x - 8.19.x, 9.1.x
+# 2.0.x - 8.19.x, 9.1.x
 # 1.13.x - 8.18.x, 9.0.x
 # 1.12.x - 8.17.x
 # 1.11.x - 8.16.x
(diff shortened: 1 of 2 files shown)


packages/cloud_security_posture/changelog.yml
```diff
@@ -1,6 +1,6 @@
 # newer versions go on top
 # version map:
-# 1.14.x - 8.19.x, 9.1.x
+# 1.14.0 | 2.0.x - 8.19.x, 9.1.x
```
maxcold, line 3: I don't think we can have two versions like this in parallel in the main branch. If we want to make a major change, we need to bump to 2.0.0, meaning 1.14.0 will never be released. There will only be `1.14.0-previewX` versions relesed. If we want to maintain two versions in parallel, we might need to think about how to achieve it with backport branches https://www.elastic.co/docs/extend/integrations/developer-workflow-support-old-package

orouz: so you're saying this should read: `2.0.x - 8.19.x, 9.1.x` without the `1.14.0` even though there are change log entries referencing `1.14.0-previewX` ? i'm good either way, it's just a comment, didn't want to make the change log ambiguous

maxcold: yeah, the comment is not that important, the important think that we all clear on the fact that there will be no `1.14.0` released after `1.13.0`, only `2.0.0`

oren-zohar (approved):

Issue #13580, opened 2025-04-17
[VMware vSphere]: Conflict with System package


lduvnjak:
### Integration Name

VMware Carbon Black Cloud [carbon_black_cloud]

### Dataset Name

vsphere.host

### Integration Version

1.18.2

### Agent Version

8.17.3

### Agent Output Type

logstash

### Elasticsearch Version

8.17.3

### OS Version and Architecture

RHEL 9.5

### Error Message

![Image](https://github.com/user-attachments/assets/adc42882-87d4-4e8e-965d-44b5f89ef189)

### What did you do?

Installed both integrations

### What did you see?

There is a field conflict between the two integrations. It is because the vSphere integration uses a `.bytes` field suffix for it's metrics.

### What did you expect to see?

They both use the same field to store the data and not cause conflicts.

### Anything else?

As far as I can tell, it's supposed to be stored at the "root" field, aka without the `.bytes` suffix.
There is also no ECS field reference for the `system.*` metricset that I could find, the only thing I found was a reference set for [metricbeat](https://www.elastic.co/docs/reference/beats/metricbeat/metricbeat-metricset-system-memory).
Also the chosen integration is Carbon Black because there is no vSphere in the dropdown menu.

lduvnjak:
Hi @andrewkroh , as is mentioned in the issue, the integration is not actually Carbon Black.
I've selected that since there is no VMware vSphere to select from the dropdown.

Can you please update the labels accordingly?

andrewkroh:
It's in the dropdown, but the sort order is confusing since it is based on the package name (in the brackets) instead of the title.

<img width="252" alt="Image" src="https://github.com/user-attachments/assets/98bc6ea9-8234-4901-a301-3b158e01921b" />

lduvnjak:
You're right, my bad. I didn't see it next to the other VMware integrations so I assumed it's missing. Thanks again!

andrewkroh:
PR to fix the sort order: https://github.com/elastic/integrations/pull/13583

andrewkroh:
@lduvnjak Could you share the contents (as JSON) of one of the conflicting documents from `ds-metrics-vsphere.host-default-2025.04.03-000001` or `.ds-metrics-vsphere.host-prod-2025.03.13-000003` that contain the `system.memory.free` as an object.

PR #13583, opened 2025-04-17
fix(issue template): sort packages by title

andrewkroh:
## Proposed commit message

Sort the list of integrations in the issue template by title (ignoring case). Previously they were ordered by directory name which was confusing.

## Related

- https://github.com/elastic/integrations/issues/13580#issuecomment-2812766382


mrodm (approved):
:+1:

PR #13588, opened 2025-04-17
[trend_micro_vision_one] Collect telemetry via the Datalake Pipeline API

### LINKED ISSUE #10192: [Trend Micro Vision One] Adding support for Datalake Pipeline
Our Trend Micro Vision One integration is currently limited to ingesting alerts/detections/audit events from their platform, which is useful as a starting point to support SIEM use cases. However, similar to other integrations we have with CrowdStrike & SentinelOne, Trend Micro provides an API to ingest raw telemetry, which would unlock significant value for our mutual users. This telemetry could be used for detections, dashboards, Analyzer view and more in Elastic Security.

The aim of this task is to add support for Trend's [Datalake Pipeline API ](https://automation.trendmicro.com/xdr/api-v3#tag/Datalake-Pipeline) to ingest and map telemetry to ECS, to power our analytics. 

Worth noting that Datalake Pipeline will incur additional costs for Trend Micro users (similar to how other vendors approach raw telemetry), which we'll need to make clear to our users.

chrisberkhout:
## Proposed commit message

```
[trend_micro_vision_one] Collect telemetry via the Datalake Pipeline API

Adds a data stream to collect telemetry data (of all sub-types) from the
Datalake Pipeline API[1]. Includes ECS mappings based on available data,
and a basic dashboard.

Also enables the request trace deletion feature in the existing HTTPJSON
data streams.

This has been manually tested against a test server with a lot of data.

[1]: https://automation.trendmicro.com/xdr/api-v3/#tag/Datalake-Pipeline
```

## Discussion

Reviewers: Ask me for the field mapping spreadsheet.

It registers a pipeline for (all sub-types of) telemetry data. It doesn't register a pipeline for detections data, since an existing data stream collects that from another API.

## Checklist

## Author's Checklist

- [x] System test with a mock server based on API documentation and sample data.
- [x] Manual testing of pipeline registration and checking against a live API.
- [ ] Additional manual testing against a live API with data.

## Related issues

- Closes #10192

## Screenshots

A new dashboard screenshot has been added to the integration.


packages/trend_micro_vision_one/_dev/deploy/docker/files/config.yml
```diff
@@ -24,3 +24,131 @@ rules:
...
+
+  - path: /v3.0/datalake/dataPipelines
+    methods: ['GET']
+    request_headers:
+      Authorization:
+        - "Bearer xxxx"
+    responses:
+      - status_code: 200
+        body: |
+          {
```
efd6, line 36: `minify_json` unless the API sends the data in this format.

packages/trend_micro_vision_one/_dev/deploy/docker/files/config.yml
```diff
@@ -24,3 +24,131 @@ rules:
...
...escription":"Windows 10 Pro (64 bit) build 19045","osName":"Windows","osType":"0x00000030","osVer":"10.0.19045","plang":1,"pname":"Endpoint Sensor","pplat":5889,"processCmd":"C:\\Windows\\system32\\svchost.exe -k LocalSystemNetworkRestricted -p -s SysMain","processFileCreation":"1733468133525","processFileHashId":"-821808160829839906","processFileHashMd5":"7469cc568ad6821fd9d925542730a7d8","processFileHashSha1":"e4e3f6bbad17b41a42687b3d75ade4a10b0870ec","processFileHashSha256":"6fc3bf1fdfd76860be782554f8d25bd32f108db934d70f4253f1e5f23522e503","processFileModifiedTime":"1733468133525","processFilePath":"C:\\Windows\\System32\\svchost.exe","processFileSize":"57528","processHashId":"-7291293652695135564","processLaunchTime":"1739418294389","processName":"C:\\Windows\\System32\\svchost.exe","processPid":1708,"processSigner":["Microsoft Windows Publisher"],"processSignerFlagsAdhoc":[false],"processSignerFlagsLibValid":[false],"processSignerFlagsRuntime":[false],"processSignerValid":[true],"processTrueType":7,"processUser":"SYSTEM","processUserDomain":"NT AUTHORITY","productCode":"xes","pver":"1.2.0.5960","sessionId":0,"timezone":"UTC+00:00","uuid":"b3d95071-ccad-44c3-9c99-f08be03bc786"}
```
efd6, line 142: Expand the lines and use `minify_json`.

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,297 @@
...
+  (state.?cursor.packages_to_get.orValue([]).size() > 0) ?
+
+    //--------------------------------------------------------------------------
+    // have a package to get -> get the package
+    //--------------------------------------------------------------------------
+
+    request("GET", [
+      state.url.trim_suffix("/"),
+      "/v3.0/datalake/dataPipelines/", state.cursor.packages_to_get[0].pipeline_id,
+      "/packages/", state.cursor.packages_to_get[0].package_id,
```
efd6, line 92: I think this logic could be simplified by using `head`(`path.?to.possible.array[0]`) and `tail`, though this would require that the kibana.version be updated to 8.15.0 to get the `tail` function.

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,297 @@
...
+{{#each tags as |tag|}}
+  - {{tag}}
+{{/each}}
+{{#contains "forwarded" tags}}
+publisher_pipeline.disable_host: true
+{{/contains}}
+{{#if enable_request_tracer}}
+resource.tracer.filename: "../../logs/cel/http-request-trace-*.ndjson"
+resource.tracer.maxbackups: 5
+{{/if}}
```
efd6, line 27: Is it worth bumping the stack version so that we can have trace deletion?

packages/trend_micro_vision_one/data_stream/telemetry/_dev/test/pipeline/test-data.log-expected.json
```diff
@@ -0,0 +1,4201 @@
...
+                ],
+                "name": "tvo-5901-1"
+            },
+            "os": {
+                "full": "Windows 10 Pro (64 bit) build 19045",
+                "name": "Windows",
+                "version": "10.0.19045"
+            },
+            "process": {
+                "command_line": "\"C:\\ProgramData\\Microsoft\\Windows Defender\\platform\\4.18.24090.11-0\\MpDefenderCoreService.exe\"",
```
efd6, line 48: Do we want to unquote the `command_line` field?

packages/trend_micro_vision_one/data_stream/telemetry/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,988 @@
...
+  ##############################################################################
+  # Add type, subtype fields
+  ##############################################################################
+
+  - script:
+      tag: script_add_event_type_based_on_event_id
+      lang: painless
+      description: Add event_type based on event_id
+      source: |
+        def key = (ctx.trend_micro_vision_one?.telemetry?.event_id ?: "").toString();
```
efd6, line 115: Why do this? ISTM that if `ctx.trend_micro_vision_one?.telemetry?.event_id == null` then we can immediately fall to assigning 'Other'.

Something like
```
def key = ctx.trend_micro_vision_one?.telemetry?.event_id;
if (key == null || params[key] == null) {
  ctx.trend_micro_vision_one.telemetry.event_type = 'Other';
  return;
}
ctx.trend_micro_vision_one.telemetry.event_type = params[key];
```

packages/trend_micro_vision_one/_dev/deploy/docker/files/config.yml
chrisberkhout: I changed it to have the individual items compacted.

Compacting the outer fields seems like less readability for no benefit.

packages/trend_micro_vision_one/_dev/deploy/docker/files/config.yml
chrisberkhout: I'm not sure what you mean here. I want a minified JSON object on each line and that's what it's doing.

packages/trend_micro_vision_one/data_stream/telemetry/_dev/test/pipeline/test-data.log-expected.json
chrisberkhout: I think they're correct, because there's a space in the name.
This is what you'd type at the command prompt to run it:
```
"C:\ProgramData\Microsoft\Windows Defender\platform\4.18.24090.11-0\MpDefenderCoreService.exe"
```
Additional arguments would come afterwards, outside the quotes.

packages/trend_micro_vision_one/data_stream/telemetry/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: The event IDs are sometimes numbers and sometimes strings. I made the params keys strings, so we need a string to check the table.

Could still short-cut the null but it doesn't seem worth it. Some sample data I have is 6% nulls for event ID.

It could have mixed-type params keys and depend on particular values always having a given type, but that seems brittle and more confusing.

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
chrisberkhout: Yes. Actually I used head and tail originally then realized it wouldn't work on 8.13.0 so went to filter. Thinking about it again, updating to 8.15.0 seems better.

Done. The `head()` macro is for HTTP but `state.?cursor.packages_to_get.hasValue()` is good.

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
chrisberkhout: Yes, I think so. Done.

chrisberkhout: Also done for the existing data streams.

packages/trend_micro_vision_one/_dev/deploy/docker/files/config.yml
efd6: Yeah. We have competing needs; readability and minimised diff noise for future changes, and minified objects to replicate the real API. So to allow both of those, we can pretty print the object in the source, and render it through the `minify_json` helper at runtime so that a minified object is received by the agent.

packages/trend_micro_vision_one/_dev/deploy/docker/files/config.yml
efd6: What I'm suggesting is
```
    responses:
      - status_code: 200
        body: |
          {{ minify_json `
          {
              "items": [
                  {
                      "createdDateTime": "2022-08-20T12:34:56Z",
                      "description": "to siemhost1",
                      "id": "8746fc45-6b9d-4923-b476-931aec6e06eb",
                      "subType": [
                          "endpointActivity",
                          "cloudActivity"
                      ],
                      "type": "telemetry"
                  },
                  {
                      "createdDateTime": "2022-08-22T01:23:45Z",
                      "description": "to siemhost1",
                      "id": "3db80190-2172-11ed-861d-0242ac120002",
                      "type": "detection"
                  }
              ],
              "count": 2
          }
          `}}

File: packages/trend_micro_vision_one/manifest.yml [hunk 7]
```diff
@@ -1,25 +1,84 @@
...
     title: Trend Micro Vision One Logo
     size: 32x32
     type: image/svg+xml
+vars:
+  - name: hostname
+    type: text
+    title: Regional Domain URL
+    description: "Trend Micro Vision One URL to connect to the API. The URL domain used for this configuration is the domain for the region where your service endpoint is hosted. See the [Trend Vision One documentation](https://automation.trendmicro.com/xdr/Guides/Regional-domains) for the domain for your region. Enter the the HTTPS URL for your domain, `https://<your-regional-domain` without any trailing slash. "
+    required: true
+  - name: enable_request_tracer
```
efd6, line 33: This needs a `default: false` [now](https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#request-trace-logging-must-be-safely-configured-if-used).

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,297 @@
...
+    (state.?cursor.pipeline_id.orValue(null) != null) ?
+
+      //------------------------------------------------------------------------
+      // have pipeline id -> check for packages, update list of packages to get
+      //------------------------------------------------------------------------
+
+      {
+        "startDateTime": [
+          state.?cursor.newStartDateTime.optMap(t, t.parse_time(time_layout.RFC3339)).orValue(now - duration(state.initial_interval)),
+          now - duration(string(7*24)+"h"), // now-7d is the oldest allowed start
```
efd6, line 152: What is the API response if we are outside this range? Can we be outside the range between the setting of `now` by the input and getting to this point in the program (it's a tight race to lose, but I guess it's possible).

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,297 @@
...
+      "Header": { "Authorization": ["Bearer " + state.api_token] }
+    }).do_request().as(resp,
+      (resp.StatusCode != 200 && resp.StatusCode != 404) ? // Other error - log it, skip this package and continue after interval
+        state.with({
+          "events": {
+            "error": {
+              "code": string(resp.StatusCode),
+              "id": string(resp.Status),
+              "message": "GET fetch a package: "+(
+                size(resp.Body) != 0 ?  string(resp.Body) : string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
```
efd6, line 103: nit (also below)
```suggestion
                size(resp.Body) != 0 ? string(resp.Body) : string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
```

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,297 @@
...
+        )
+      )
+
+    :
+
+      //------------------------------------------------------------------------
+      // no pipeline id -> setup: try to get the pipeline id
+      //------------------------------------------------------------------------
+
+      (state.?hard_coded_pipeline_id.orValue('') != '') ?
```
efd6, line 207: Do they sometimes return `""`?

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,297 @@
...
+                      })
+                    :
+                      // no pipeline id is returned, but it will be found next time
+                      state.with({
+                        "events": [{"event": {"reason": "want_more"}}],
+                        "want_more": int(state.registration_count) == 0, // avoid a busy loop by waiting if the registration was previously attempted
+                        "registration_count": int(state.registration_count) + 1,
+                      })
+                  )
+
```
efd6, line 294:
```suggestion
```
style only, feel free to ignore

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
chrisberkhout: It doesn't fail. It seems to just return what it wants.

Still, I don't think we should request things that we've been told we can't have.

Easy fix:
```diff
-          now - duration(string(7*24)+"h"), // now-7d is the oldest allowed start
+          now - duration(string(7*24)+"h1m"), // now-7d is the oldest allowed start, so limit to now-7d-1m
```

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
chrisberkhout: I guess I'm going to keep it.

I like this
```
body.items.filter(i, i.description == state.managed_pipeline_description)[?0].id.orValue(null).as(id,
  (id != null) ?

    //...

  :

    //...

)
```
Better than 
```
body.items.filter(i, i.description == state.managed_pipeline_description)[?0].id.orValue(null).as(id,
  (id != null) ?

    //...

  :

    //...
)
```

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
chrisberkhout: Sorry, I thought I wrote something about this, but I don't see it here now.

When I changed the check to `state.hard_coded_pipeline_id != null`, I did manually create a policy without a value and saw that it included `hard_coded_pipeline_id: null`. And I thought it was fine since the system test kept passing locally.

But maybe there was a build problem or something that kept me stuck on an older version. I cleared old builds and today it does fail.

The system test agent policy also gets `state.hard_coded_pipeline_id: null`, but it's dropped later. I think [here](https://github.com/elastic/go-ucfg/blob/2a22403faaea7d9e9d492f210a0f2c576c5d3dfc/reify.go#L229-L231) in `go-ucfg`.

Changing the check to `has(state.hard_coded_pipeline_id)` makes it all work.

My first thought is that extra handlebars logic to remove the key from the policy is redundant, because keys with null values will be removed later anyway. On the other hand, this is a situation where mito will run the program differently than the CEL input, which I don't like. Mito keeps nulls in the input state:
```
mito -data <(echo '{ "k": null }') <(echo '{
  "k_exists": has(state.k),
  "k_is_null": state.k == null,
  "l_exists": has(state.l),
}')
```
```
{
  "k_exists": true,
  "k_is_null": true,
  "l_exists": false
}
```

So now I think either:
- add the handlbars condition, or
- change the CEL check to `has(state.hard_coded_pipeline_id) && state.hard_coded_pipeline_id != null`.

What do you think?

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,297 @@
...
+                  "message": "GET list available packages: "+(
+                    size(resp.Body) != 0 ? string(resp.Body) : string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
+                  ),
+                },
+              },
+              "cursor": (resp.StatusCode == 404) ? state.cursor.drop("pipeline_id") : state.cursor,
+              "want_more": false,
+            })
+          :
+            bytes(resp.Body).decode_json().as(body,
```
efd6, line 185:
```suggestion
            resp.Body.decode_json().as(body,
```

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,297 @@
...
+                  "id": string(resp.Status),
+                  "message": "GET list pipelines: "+(
+                    size(resp.Body) != 0 ? string(resp.Body) : string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
+                  ),
+                },
+              },
+              "want_more": false,
+            })
+          :
+            bytes(resp.Body).decode_json().as(body,
```
efd6, line 242:
```suggestion
            resp.Body.decode_json().as(body,
```

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,297 @@
...
+              body.items.filter(i, i.description == state.managed_pipeline_description)[?0].id.orValue(null).as(id,
+                (id != null) ?
+
+                  //------------------------------------------------------------
+                  // got the pipeline id -> return it in the cursor
+                  //------------------------------------------------------------
+
+                  state.with({
+                    "events": [{"event": {"reason": "want_more"}}],
+                    "cursor": state.?cursor.orValue({}).with({ "pipeline_id": id }),
```
efd6, line 252:
```suggestion
              body.items.filter(i, i.description == state.managed_pipeline_description)[?0].id.as(id,
                id.hasValue() ?

                  //------------------------------------------------------------
                  // got the pipeline id -> return it in the cursor
                  //------------------------------------------------------------

                  state.with({
                    "events": [{"event": {"reason": "want_more"}}],
                    "cursor": state.?cursor.orValue({}).with({ "pipeline_id": id.value() }),
```

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,297 @@
...
+        })
+    )
+
+  :
+
+    //--------------------------------------------------------------------------
+    // no package to get -> check or setup
+    //--------------------------------------------------------------------------
+
+    (state.?cursor.pipeline_id.hasValue()) ?
```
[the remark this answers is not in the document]
chrisberkhout: I do like the `has()` version better. I was checking some things and found this issue https://github.com/elastic/mito/issues/89.

It doesn't apply in this case though, so I switched to `has(state.?cursor.pipeline_id)`.

packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs
chrisberkhout: Done.

efd6 (approved):
Nicely tamed large CEL program. LGTM

PR #13589, opened 2025-04-17
fix(packages/falco): use ecs definition for process.group.{id,name}

andrewkroh:
## Proposed commit message

Falco was using data type 'text' for the process.group.id and
process.group.name fields, but those fields are defined as 'keyword' in 
ECS leading to conflicts with other packages that are following ECS.

I modified the field definitions to use 'external: ecs' where possible.
And removed unused attributes from fields.yml files.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/falco/data_stream/alerts/fields/base-fields.yml
```diff
@@ -1,22 +1,18 @@
...
-  description: Data stream dataset.
-  value: 'falco.alerts'
+  external: ecs
+  value: falco.alerts
 - name: data_stream.namespace
-  type: constant_keyword
-  description: Preserved Falco field
+  external: ecs
 - name: event.dataset
   type: constant_keyword
```
ShourieG, line 10:
```suggestion
```

packages/falco/data_stream/alerts/fields/base-fields.yml
```diff
@@ -1,22 +1,18 @@
...
-  description: Data stream / event dataset.
-  value: 'falco.alerts'
+  external: ecs
+  value: falco.alerts
 - name: event.module
   type: constant_keyword
```
ShourieG, line 14:
```suggestion
```

packages/falco/data_stream/alerts/fields/base-fields.yml
ShourieG: Any specific reason we are not removing the type here, since we are already referencing ecs ?

packages/falco/data_stream/alerts/fields/base-fields.yml
ShourieG: same query as above

packages/falco/data_stream/alerts/fields/base-fields.yml
andrewkroh: event.dataset is not a `constant_keyword` hence the override is still necessary to retain the same behavior.

packages/falco/data_stream/alerts/fields/base-fields.yml
andrewkroh: `event.module` is not a `constant_keyword` hence the override is still necessary to retain the same behavior.

ShourieG (approved):
LGTM

packages/falco/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.3.1"
+  changes:
+    - description: Align field definitions with ECS. The `process.group.id` and `process.group.name` fields were changed from text to keword.
```
andrewkroh, line 4:
```suggestion
    - description: Align field definitions with ECS. The `process.group.id` and `process.group.name` fields were changed from text to keyword.
```

Issue #13591, opened 2025-04-17
[ded] source.bytes field declared as type double conflicts with the ECS data type long


andrewkroh:
`source.bytes` field declared as type double conflicts with the ECS data type long. This field should use the `external: ecs` definition to prevent data view conflicts with other integrations.

https://github.com/elastic/integrations/blob/e5218567da48062dc4e067640fad7c2eabc738e3/packages/ded/elasticsearch/transform/pivot_transform/fields/fields.yml#L29-L30

jmcarlock:
@qn895 Also wanted to double check here, since we are changing the transform mappings, I believe we also need to bump the transform version.

PR #13601, opened 2025-04-18
Change DED field mappings to ECS

jmcarlock:
## Proposed commit message

Change to transform mappings to ECS so they do not conflict with dashboards.

## Checklist

- [ ] ~I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) ~

Tested locally with elastic-package.

## Related issues

- https://github.com/elastic/integrations/issues/13591


qn895 (dismissed):
LGTM 🎉

packages/ded/elasticsearch/transform/pivot_transform/transform.yml
```diff
@@ -88,5 +88,5 @@ sync:
     delay: 120s
     field: "@timestamp"
 _meta:
-  fleet_transform_version: 2.2.0
+  fleet_transform_version: 2.3.1
```
andrewkroh, line 91: I think you may also need to change the destination index name on line 3 for the new mappings to apply on upgrade.

andrewkroh (commented):
The average aggregation that sets `source.bytes` will emit a float, and I assume that ES will [coerce](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/coerce) that to a long.

I highly recommend manually testing an upgrade from the current version to this new version to verify there are no issues with the transform. This is a big gap in our automated testing at the moment.

packages/ded/elasticsearch/transform/pivot_transform/transform.yml
jmcarlock: Thanks @andrewkroh ! I'm retesting now, it might have picked up the last version of the transform.

jmcarlock:
Successfully tested on transform output.

Screenshot 2025-04-21 at 11 03 14 AM

sodhikirti07 (approved):
lgtm!

jmcarlock:
> The average aggregation that sets source.bytes will emit a float, and I assume that ES will [coerce](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/coerce) that to a long.
>
> I highly recommend manually testing an upgrade from the current version to this new version to verify there are no issues with the transform and dest index upgrade. This is a big gap in our automated testing at the moment.

Before merging, I double checked the mappings that are input to the transform:

```
{
  "logs-test_ded_high_sent_bytes_destination_ip": {
    "mappings": {
      "source.bytes": {
        "full_name": "source.bytes",
        "mapping": {
          "bytes": {
            "type": "long"
          }
        }
```

`source.bytes` is correctly mapped to `long`. However the transform output was not coercing as expected:
```
{
  "ml_network_ded-2.3.1": {
    "mappings": {
      "source.bytes": {
        "full_name": "source.bytes",
        "mapping": {
          "bytes": {
            "type": "double"
          }
...
```

I think the outputs are still being deduced by the transform. However, I suggesting [changing this field name to `avg_source_bytes`](https://github.com/elastic/integrations/pull/13601/commits/54ad73264dce94e00898c0bf9278cbd392813993) and defining it as a `double`, as it doesn't actually represent `source.bytes`, but the average.

One note though, the transform output is only writing to `ml_network_ded-<TRANSFORM_VERSION>` (and aliased to `ml_network_ded.all`), and isn't picked up by any dashboard I'm aware of.

@sodhikirti07 do you think this is okay? Should we leave a more verbose note in the readme about this field changing names?

sodhikirti07:
@jmcarlock I think we can try rounding the number by using [bucket script aggregations](https://www.elastic.co/docs/reference/aggregations/search-aggregations-pipeline-bucket-script-aggregation) or anything similar. That way we won't need to change anything, hopefully.

jmcarlock:
Thanks @sodhikirti07 for taking a look!

@andrewkroh I don't think bucket aggs will work here as even if we put everything into one large bucket, we still will need to call the average function.

Is there an alternative to renaming `source.bytes` as I have done? Is there a way to coerce this field properly to a long in the transform?

andrewkroh:
If there is no other means, then a pipeline can be applied to a transform to execute a [convert processor](https://www.elastic.co/docs/reference/enrich-processor/convert-processor).

jmcarlock:
Thanks @andrewkroh ! I will give this a try.

jmcarlock:
When trying to add the convert processor it looks like it is unable to convert from to Long:
Screenshot 2025-04-22 at 8 50 59 AM

I also tried to use a Painless script in the pipeline, but it seems that was ignored. For instance when testing adding an integer:

```java
PUT _ingest/pipeline/test_pipeline
{
  "processors": [
    {
      "script": {
        "lang": "painless",
        "source": """
          double sourceBytes = ctx['source']['bytes'];
          int intValue = (int) sourceBytes;
          ctx['source']['bytes'] = intValue;
        """
      }
    }
  ]
}
```

Then I add it to the destination index settings:
```
dest:
  index: ml_network_ded-2.3.1
  aliases:
    - alias: ml_network_ded.latest
      move_on_creation: true
    - alias: ml_network_ded.all
      move_on_creation: false
  pipeline: test_pipeline
```

This produces no different results even though when using `_simulate` it works correctly. I'm assuming the pipeline is being executed before the pivot?

I still think this is the best course of action:
> I think the outputs are still being deduced by the transform. However, I suggesting [changing this field name to avg_source_bytes](https://github.com/elastic/integrations/pull/13601/commits/54ad73264dce94e00898c0bf9278cbd392813993) and defining it as a double, as it doesn't actually represent source.bytes, but the average.

sodhikirti07 (changes requested):
@jmcarlock I agree with your solution. If a user upgrades this package, will they be able to see both versions of the transform? And I assume the same applies to the destination indices and ML jobs as well? They'll need to reconfigure the jobs as well. You might need to change the `source.bytes` field in [ML jobs](https://github.com/elastic/integrations/blob/54ad73264dce94e00898c0bf9278cbd392813993/packages/ded/kibana/ml_module/ded-ml.json#L65). Please make note of this change in the README.

jmcarlock:
I found this morning that we can't change the field name to `avg_source_bytes` because the field is used downstream in AD jobs.

@sodhikirti07 helped me find a way to cast properly to long! However, @andrewkroh when we try to specify the field from ECS it says the mapping for `source.bytes` is `keyword`. It only worked in this case when we explicitly defined it as `long`. Can you advise?

andrewkroh:
To use `external: ecs` you need to add _dev/build/build.yml file:

```yaml
dependencies:
  ecs:
    reference: "git@v8.17.0"
```

packages/ded/elasticsearch/ingest_pipeline/ml_ded_ingest_pipeline.yml
```diff
@@ -0,0 +1,17 @@
+---
+description: Pipeline for the Data Exfiltration Detection Integration transform.
+processors:
+  - script:
+      description: Casts the value of the field `avg_source_bytes` to a long and assigns it to `source.bytes`.
+      lang: painless
+      source: |
+        if (ctx.avg_source_bytes != null) {
+            ctx['source']['bytes'] = (long)(ctx.avg_source_bytes);
+        }
```
andrewkroh, line 10:
```suggestion
  - convert: 
      field: avg_source_bytes
      type: long
      ignore_missing: true
  - remove:
      field: avg_source_bytes
      ignore_missing: true
```

This can be accomplished with `convert`. Also, do we need to retain `avg_source_bytes` for anything?

packages/ded/elasticsearch/ingest_pipeline/ml_ded_ingest_pipeline.yml
```diff
@@ -0,0 +1,17 @@
...
+on_failure:
+  - set:
+      field: event.kind
+      value: pipeline_error
+  - append:
+      field: error.message
+      value: '{{{ _ingest.on_failure_message }}}'
```
andrewkroh, line 17:
```suggestion
      value: >-
        Processor '{{{ _ingest.on_failure_processor_type }}}'
        {{#_ingest.on_failure_processor_tag}}with tag '{{{ _ingest.on_failure_processor_tag }}}'
        {{/_ingest.on_failure_processor_tag}}failed with message '{{{ _ingest.on_failure_message }}}'
```

packages/ded/elasticsearch/ingest_pipeline/ml_ded_ingest_pipeline.yml
jmcarlock: No, we do not. I will verify that this works correctly with convert and try it in-place.

jmcarlock:
Reverted to a Painless script for the pipeline as the convert didn't work. @andrewkroh @sodhikirti07 now ready for final review!

packages/ded/elasticsearch/transform/pivot_transform/fields/fields.yml
```diff
@@ -1,30 +1,32 @@
...
+- external: ecs
+  name: destination.geo.country_name
+- external: ecs
+  name: destination.geo.country_iso_code
+- external: ecs
+  name: destination.geo.region_name
+- external: ecs
+  name: destination.geo.city_name
+- name: avg_source_bytes
+  type: double
```
andrewkroh, line 32: This field is dropped by pipeline so this mapping should be removed.

Everything else LGTM.

andrewkroh (approved):

sodhikirti07 (approved):
Changes lgtm!

PR #13621, opened 2025-04-21
[Servicenow] Add support to append `sysparm query` in CEL input

### LINKED ISSUE #13413: [servicenow]: User-supplied sysparm_query
### Integration Name

ServiceNow [servicenow]

### Dataset Name

event

### Integration Version

0.12.1

### Agent Version

Any

Any

### User Goal

Allow the user to supply a value for the `sysparm_query` parameter.  This can contain any number of arbitrary boolean expressions used to filter the result set returned by the query.   Without this, the query is functionally equivalent to a `SELECT *` in SQL.

[docs](https://www.servicenow.com/docs/bundle/yokohama-api-reference/page/integrate/inbound-rest/concept/c_TableAPI.html#title_table-GET)

| Name | Description |
| ------ | ----------- |
| sysparm_query | Encoded query used to filter the result set. You can use a UI filter to obtain a properly encoded query. Syntax: sysparm_query=<col_name><operator><value>.<col_name>: Name of the table column to filter against.<operator>: Supports the following values:=: Exactly matches <value>.!=: Does not match <value>.^: Logically AND multiple query statements.^OR: Logically OR multiple query statements.LIKE: <col_name> contains the specified string. Only works for <col_name> fields whose data type is string.STARTSWITH: <col_name> starts with the specified string. Only works for <col_name> fields whose data type is string.ENDSWITH: <col_name> ends with the specified string. Only works for <col_name> fields whose data type is string.<value>: Value to match against.All parameters are case-sensitive. Queries can contain more than one entry, such as sysparm_query=<col_name><operator><value>[<operator><col_name><operator><value>].For example:(sysparm_query=caller_id=javascript:gs.getUserID()^active=true)Encoded queries also support order by functionality. To sort responses based on certain fields, use the ORDERBY and ORDERBYDESC clauses in sysparm_query.Syntax:ORDERBY<col_name>ORDERBYDESC<col_name>For example: sysparm_query=active=true^ORDERBYnumber^ORDERBYDESCcategoryThis query filters all active records and orders the results in ascending order by number, and then in descending order by category.If part of the query is invalid, such as by specifying an invalid field name, the instance ignores the invalid part. It then returns rows using only the valid portion of the query. You can control this behavior using the property glide.invalid_query.returns_no_rows. Set this property to true to return no rows on an invalid query.Note: The glide.invalid_query.returns_no_rows property controls the behavior of all queries across the instance, such as in lists, scripts (GlideRecord.query()), and web service APIs.Data type: String |

### Existing Features

The integration uses the `sysparm_query` parameter to [specify the ordering of the resultset](https://github.com/elastic/integrations/blob/0c977305cb6ed830e6ca6625edd5ae4026f47a3a/packages/servicenow/data_stream/event/agent/stream/cel.yml.hbs#L37).

This enhancement would require dynamic logic to concatenate the user-supplied `sysparm_query` string (joined by the logical "AND" operator (`^`) only when present and non-empty, etc.

### What did you see?

N/A

### Anything else?

This may just be the tip of the iceberg, so to speak.  According to the [documentation](https://www.servicenow.com/docs/bundle/yokohama-api-reference/page/integrate/inbound-rest/concept/c_TableAPI.html#d248029e1461) there are several additional parameters that can be added to the GET TABLE route.

moxarth-rathod:
## Proposed commit message

```
Introduce an optional parameter to append additional query conditions
for advanced filtering needs. It is advised to avoid using timestamp
fields in the appended query to prevent potential issues with pagination and
cursor logic.
```

## Checklist

- Clone integrations repo.
> elastic-package test

## Related issues

- Closes #13413

### DIFF
--- a/packages/servicenow/changelog.yml
+++ b/packages/servicenow/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.13.0"
+  changes:
+    - description: Add option to supply a value for the sysparm_query parameter.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
 - version: "0.12.1"
   changes:
     - description: Fix handling of SQS worker count configuration.
--- a/packages/servicenow/changelog.yml
+++ b/packages/servicenow/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Add option to supply a value for the sysparm_query parameter.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/1
+      link: https://github.com/elastic/integrations/pull/13621
 - version: "0.12.1"
   changes:
     - description: Fix handling of SQS worker count configuration.
--- a/packages/servicenow/data_stream/event/agent/stream/cel.yml.hbs
+++ b/packages/servicenow/data_stream/event/agent/stream/cel.yml.hbs
@@ -21,6 +21,7 @@ state:
   batch_size: {{batch_size}}
   initial_interval: {{initial_interval}}
   timestamp_field: {{timestamp_field}}
+  query: {{query}}
 regexp:
   next_link: '<([^,]*)>;rel="next"'
   content: 'text/html'
@@ -34,7 +35,7 @@ program: |
           "sysparm_display_value": ["all"],
           "sysparm_exclude_reference_link": ["true"],
           "sysparm_limit": [string(state.batch_size)],
-          "sysparm_query": ["ORDERBY"+state.timestamp_field+"^"+state.timestamp_field+">"+start],
+          "sysparm_query": ["ORDERBY"+state.timestamp_field+"^"+state.timestamp_field+">"+start+"^"+state.?query.orValue("")],
         }.format_query()
       )
     )).as(resp, resp.StatusCode == 200 && (resp.Header["Content-Type"][0].re_find_submatch("content") == []) ?
--- a/packages/servicenow/data_stream/event/agent/stream/cel.yml.hbs
+++ b/packages/servicenow/data_stream/event/agent/stream/cel.yml.hbs
@@ -35,7 +35,7 @@ program: |
           "sysparm_display_value": ["all"],
           "sysparm_exclude_reference_link": ["true"],
           "sysparm_limit": [string(state.batch_size)],
-          "sysparm_query": ["ORDERBY"+state.timestamp_field+"^"+state.timestamp_field+">"+start+"^"+state.?query.orValue("")],
+          "sysparm_query": ["ORDERBY"+state.timestamp_field+"^"+state.timestamp_field+">"+start+(state.?query.hasValue() ? "^"+state.?query.orValue("") : "")],
         }.format_query()
       )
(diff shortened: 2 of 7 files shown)


packages/servicenow/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.13.0"
+  changes:
+    - description: Add option to supply a value for the sysparm_query parameter.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
```
ShourieG, line 6:
```suggestion
      link: https://github.com/elastic/integrations/pull/13621
```

packages/servicenow/data_stream/event/agent/stream/cel.yml.hbs
```diff
@@ -34,7 +35,7 @@ program: |
           "sysparm_display_value": ["all"],
           "sysparm_exclude_reference_link": ["true"],
           "sysparm_limit": [string(state.batch_size)],
-          "sysparm_query": ["ORDERBY"+state.timestamp_field+"^"+state.timestamp_field+">"+start],
+          "sysparm_query": ["ORDERBY"+state.timestamp_field+"^"+state.timestamp_field+">"+start+"^"+state.?query.orValue("")],
```
ShourieG, line 38: if `query` is missing, won't the final query params be something like `sysparm_query=ORDERBYstate.timestamp_field^state.timestamp_field>start^` this trailing "^" might cause issue.

ShourieG (approved):
Approving to unblock, but I still feel we should handle the trailing tokens in a more graceful manner. Having dead tokens as a suffix does not seem right, even if the API handles it.

packages/servicenow/data_stream/event/agent/stream/cel.yml.hbs
ShourieG, line 38: Let's solve this properly

ShourieG (changes requested):

packages/servicenow/data_stream/event/agent/stream/cel.yml.hbs
moxarth-rathod: Done, I've handled the trailing `^` in case the query is empty.

packages/servicenow/data_stream/event/agent/stream/cel.yml.hbs
```diff
@@ -34,7 +35,7 @@ program: |
           "sysparm_display_value": ["all"],
           "sysparm_exclude_reference_link": ["true"],
           "sysparm_limit": [string(state.batch_size)],
-          "sysparm_query": ["ORDERBY"+state.timestamp_field+"^"+state.timestamp_field+">"+start],
+          "sysparm_query": ["ORDERBY"+state.timestamp_field+"^"+state.timestamp_field+">"+start+(state.?query.hasValue() ? "^"+state.?query.orValue("") : "")],
```
ShourieG, line 38: @moxarth-rathod, can we add a system test that tests the "true" path in the condition `(state.?query.hasValue() ? "^"+state.?query.orValue("") : "")`  ?. Currently there is no system test with "query". This is just to make sure that there are no CEL eval error in any branch of evaluation here.

ShourieG (approved):
LGTM

PR #13634, opened 2025-04-22
[Teleport] Add support for configuring which cloud metadata to use

### LINKED ISSUE #12918: [teleport] Allow the user to decide which values should be set in `cloud.*` fields
Relates #12851 
Relates #12624 

It was observed that the `event-groups` ingest pipeline defined in the `teleport` package was failing when `cloud.*` fields were already present in the documents.

In #12851 those errors were solved by adding `if` conditions to the corresponding `rename` processors to avoid changing the behavior. Therefore, the values set in those `cloud.*` fields are not changed.

But it would be interesting to allow the user to make the choice of what values they are interested in, as it was done for Qualys in https://github.com/elastic/integrations/issues/10277

Related discussion threads:
- https://github.com/elastic/integrations/pull/12624#discussion_r1951898376
- https://github.com/elastic/integrations/pull/12851#pullrequestreview-2646676378

moxarth-rathod:
## Proposed commit message

```
teleport: allow the user to decide which values should be set in
cloud.* fields.

This PR adds a UI configurator that allows the user to select
which of the two sources (or both) to use to populate the cloud fields. It
also tags the document with an indicator of which sources were available. 
```

## Checklist

## Related issues

- Closes #12918

## Screenshots

image

### DIFF
--- a/packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
@@ -13,12 +13,15 @@ processors:
       if: ctx.tags?.contains('both') == true
   - script:
       lang: painless
-      if: ctx.tags?.contains('both') == true
-      source: ctx.tags.remove(ctx.tags.indexOf('both'));
-  - script:
-      lang: painless
-      if: ctx.cloud == null && ctx.tags?.contains('elastic_cloud_data') == true
-      source: ctx.tags.remove(ctx.tags.indexOf('elastic_cloud_data'));
+      source: |
+        if (ctx.tags != null) {
+          if (ctx.tags.contains('both')) {
+            ctx.tags.remove(ctx.tags.indexOf('both'));
+          }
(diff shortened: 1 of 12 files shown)


packages/teleport/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -6,6 +6,29 @@ processors:
...
+      value: ["elastic_cloud_data", "provider_cloud_data"]
+      if: ctx.tags?.contains('both') == true
+  - script:
+      lang: painless
+      if: ctx.tags?.contains('both') == true
+      source: ctx.tags.remove(ctx.tags.indexOf('both'));
+  - script:
+      lang: painless
+      if: ctx.cloud == null && ctx.tags?.contains('elastic_cloud_data') == true
+      source: ctx.tags.remove(ctx.tags.indexOf('elastic_cloud_data'));
```
ShourieG, line 21: we can combine them into one script processor and have 2 different if conditions instead of having 2 separate scripts here.

Issue #13593, opened 2025-04-17
[tenable_io] vulnerability.description should use ECS definition


andrewkroh:
The tenable_io integration needs to follow ECS field definitions.

`vulnerability.description` field declared as type text conflicts with the ECS data type keyword

https://github.com/elastic/integrations/blob/5743fabad0863f937273cc0cc56c3b32e30a06bf/packages/tenable_io/data_stream/vulnerability/fields/fields.yml#L456-L458

Issue #12827, opened 2025-02-18
TenableIO: Implement transform for Cloud Security Workflows


kcreddy:
meta: https://github.com/elastic/integrations/issues/12823

Issue #12826, opened 2025-02-18
TenableIO: Implement mappings for Cloud Security Workflows


kcreddy:
meta: https://github.com/elastic/integrations/issues/12823

Issue #12823, opened 2025-02-18
[meta] Update Tenable IO integration to Leverage Native Cloud Security Workflows


kcreddy:
As part of effort to leverage Cloud Security workflows such as CSPM and CNVM for 3rd party integrations, the vulnerabilities data from Tenable IO needs to be enriched just like previous enhancements for Wiz, AWS Security Hub, and Qualys VMDR.

Tasks:

- [ ] Get access to Tenable IO environment.
- [ ] Setup Tenable IO data in demo cluster.
- [ ] Analyse required mapping changes for Tenable IO and get feedback from Cloud team.
- [ ] Implement mapping changes 
- [ ] Implement transform

cc: @narph

kcreddy:
Hey @cpascale43,
I would need the credentials for Tenable IO to get started on this work. If we don't have the environment, we will have to reach out to Tenable and get its license.

cc: @narph @jamiehynds

kcreddy:
Closing this as all requirements have been implemented:
- Setup Tenable data in SEI cluster.
- Approval of proposed mapping changes.
- Add required mappings and transform: https://github.com/elastic/integrations/pull/13636

PR #13636, opened 2025-04-22
tenable_io: Add mappings and transform for Cloud Detection and Response (CDR) workflow

kcreddy:
## Proposed commit message
```
tenable_io: Add mappings and transform for Cloud Detection and Response (CDR) workflow.

Add mappings and latest transform to tenable_io.vulnerability datastream 
to help with  Cloud Native Vulnerability Management (CNVM)[1] workflow.

Remove explicitly defined vulnerability.description field causing ECS conflict. 
It will be auto-mapped using ecs@mappings component template. Inside 
the destination index, the mapping is defined using - external: ecs.

[1] https://www.elastic.co/guide/en/security/current/vuln-management-overview.html

```

> [!NOTE]
> To Reviewers:
> 
> - Please refer to this guide: https://docs.elastic.dev/security-solution/cloud-security/cdr/3p-dev-guide which was used for proposed changes.

## Checklist

Both pipeline and system tests should pass:
`cd packages/tenable_io && elastic-package build && elastic-package stack up -d -v  --version=8.19.0-SNAPSHOT && eval "$(elastic-package stack shellinit)" && elastic-package test pipeline --generate  -v`
```
--- Test results for package: tenable_io - START ---
╭────────────┬───────────────┬───────────┬───────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE    │ DATA STREAM   │ TEST TYPE │ TEST NAME                                         │ RESULT │ TIME ELAPSED │
├────────────┼───────────────┼───────────┼───────────────────────────────────────────────────┼────────┼──────────────┤
│ tenable_io │ asset         │ pipeline  │ (ingest pipeline warnings test-asset.log)         │ PASS   │ 554.821375ms │
│ tenable_io │ asset         │ pipeline  │ test-asset.log                                    │ PASS   │  46.067709ms │
│ tenable_io │ audit         │ pipeline  │ (ingest pipeline warnings test-audit.log)         │ PASS   │ 450.386542ms │
│ tenable_io │ audit         │ pipeline  │ test-audit.log                                    │ PASS   │ 112.436125ms │
│ tenable_io │ plugin        │ pipeline  │ (ingest pipeline warnings test-plugin.log)        │ PASS   │ 427.393167ms │
│ tenable_io │ plugin        │ pipeline  │ test-plugin.log                                   │ PASS   │  62.169584ms │
│ tenable_io │ scan          │ pipeline  │ (ingest pipeline warnings test-scan.log)          │ PASS   │ 463.494792ms │
│ tenable_io │ scan          │ pipeline  │ test-scan.log                                     │ PASS   │   45.26575ms │
│ tenable_io │ vulnerability │ pipeline  │ (ingest pipeline warnings test-vulnerability.log) │ PASS   │ 477.573042ms │
│ tenable_io │ vulnerability │ pipeline  │ test-vulnerability.log                            │ PASS   │ 106.045875ms │
╰────────────┴───────────────┴───────────┴───────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: tenable_io - END   ---
Done
```

`cd packages/tenable_io && elastic-package build && elastic-package stack up -d -v --version=8.19.0-SNAPSHOT && eval "$(elastic-package stack shellinit)" && elastic-package test system --generate  -v --data-streams=vulnerability`
```
--- Test results for package: tenable_io - START ---
╭────────────┬───────────────┬───────────┬───────────┬────────┬─────────────────╮
│ PACKAGE    │ DATA STREAM   │ TEST TYPE │ TEST NAME │ RESULT │    TIME ELAPSED │
├────────────┼───────────────┼───────────┼───────────┼────────┼─────────────────┤
│ tenable_io │ vulnerability │ system    │ default   │ PASS   │ 2m20.248691833s │
╰────────────┴───────────────┴───────────┴───────────┴────────┴─────────────────╯
--- Test results for package: tenable_io - END   ---
Done
```

## Related issues

- Closes https://github.com/elastic/integrations/issues/12826
- Closes https://github.com/elastic/integrations/issues/12827
- Relates https://github.com/elastic/integrations/issues/12823
- Closes https://github.com/elastic/integrations/issues/13593

## Screenshots

**Ingested source data:**
Screenshot 2025-04-25 at 7 55 10 PM

**Transform running:**
Screenshot 2025-04-25 at 8 09 05 PM

**Populating destination index:**
Screenshot 2025-04-25 at 8 09 32 PM


packages/tenable_io/_dev/deploy/docker/files/config.yml
```diff
@@ -45,7 +45,7 @@ rules:
...
...6-2051)","family":"Web Clients","family_id": 1000020,"has_patch":false,"id":9062,"name":"Test  &lt; 48.0.2564.82 Multiple Vulnerabilities","risk_factor":"HIGH","see_also":["http://testreleases.blogspot.com/2016/01/beta-channel-update_20.html"],"solution":"Update the  browser to 48.0.2564.82 or later.","synopsis":"The remote host is utilizing a web browser that is affected by multiple vulnerabilities.","vpr":{"score":5.9,"drivers":{"age_of_vuln":{"lower_bound":366,"upper_bound":730},"exploit_code_maturity":"UNPROVEN","cvss_impact_score_predicted":false,"cvss3_impact_score":5.9,"threat_intensity_last28":"VERY_LOW","threat_sources_last28":["No recorded events"],"product_coverage":"LOW"},"updated":"2019-12-31T10:08:58Z"}},"port":{"port":"0","protocol":"TCP"},"scan":{"completed_at":"2018-12-31T20:59:47Z","schedule_uuid":"6f7db010-9cb6-4870-b745-70a2aea2f81ce1b6640fe8a2217b","started_at":"2018-12-31T20:59:47Z","uuid":"0e55ec5d-c7c7-4673-a618-438a84e9d1b78af3a9957a077904"},"severity":"low","severity_id":3,"severity_default_id":3,"severity_modification_type":"NONE","first_found":"2018-12-31T20:59:47Z","last_found":"2018-12-31T20:59:47Z","indexed":"2022-11-30T14:09:12.061Z","state":"OPEN"}]
```
kcreddy, line 48: Updated "output" to test `package` field.

packages/tenable_io/data_stream/vulnerability/fields/package.yml
```diff
@@ -0,0 +1,10 @@
+- name: package
```
kcreddy, line 1: `package` field is defined nested.

packages/tenable_io/data_stream/vulnerability/fields/vulnerability.yml
```diff
@@ -0,0 +1,24 @@
...
+      type: keyword
+    - name: package
+      type: nested
+    - name: package.fixed_version
+      type: keyword
+    - name: package.name
+      type: keyword
+    - name: package.path
+      type: keyword
+    - name: package.published_date
```
kcreddy, line 14: Because `package` is defined nested and contains array of objects, `package.published_date` is also populated multiple times for each element in the array from the raw data field `plugin.vuln_publication_date` as discussed in our spreadsheet.

cc: @maxcold @nick-alayil

packages/tenable_io/data_stream/vulnerability/_dev/test/pipeline/test-vulnerability.log-expected.json
```diff
@@ -166,13 +177,486 @@
                 ],
                 "report_id": "0e55ec5d-c7c7-4673-a618-438a84e9d1b78af3a9957a077904",
                 "scanner": {
+                    "name": "Google Chrome &lt; 48.0.2564.82 Multiple Vulnerabilities",
+                    "vendor": "Tenable"
+                },
+                "score": {
+                    "version": "3.0"
+                },
+                "severity": "low",
```
kcreddy, line 186: Our ECS `severity` has values with first letter as uppercase (ex: `Low`). Tenable sends them as all lowercase (ex: `low`).

@maxcold Let me know if this can be handled by CNVM, if not I need to update inside the pipeline.

packages/tenable_io/elasticsearch/transform/latest_cdr_vulnerabilities/transform.yml
```diff
@@ -0,0 +1,30 @@
...
+dest:
+  index: "security_solution-tenable_io.vulnerability_latest-v1"
+  aliases:
+    - alias: "security_solution-tenable_io.vulnerability_latest"
+      move_on_creation: true
+latest:
+  unique_key:
+    - event.id
+    - resource.id
+    - data_stream.namespace
```
kcreddy, line 13: Transform's uniqueness is defined similar to Qualys, instead of following the guide.

cc: @maxcold @alexreal1314

packages/tenable_io/data_stream/vulnerability/fields/vulnerability.yml
kcreddy: Please see screenshots in the PR description.

packages/tenable_io/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -311,10 +372,106 @@ processors:
...
+      tag: gsub_tenable_io_vulnerability_output_split_packages
+      ignore_missing: true
+  - gsub:
+      field: tenable_io.vulnerability.output
+      pattern: '\n\s*'
+      description: To facilitate splitOnToken in the subsequent script processor, which otherwise cannot split on \n. Within each package, details are seperated by single \n.
+      replacement: '||'
+      tag: gsub_tenable_io_vulnerability_output_split_package_info
+      if: ctx._temp?.output_has_package == true
+      ignore_missing: true
```
chrisberkhout, line 416: I think you can can avoid these by building the token strings from chars.

In Painless Lab, put this:
```
String lf = (String)(char)10;
String s = "line1" + lf + "line2";

"s:"+lf
+"---"+lf
+s+lf
+"---"+lf
+lf
+"s.splitOnToken(lf)[1]:"+lf
+s.splitOnToken(lf)[1]
```
And you'll see the output:
```
s:
---
line1
line2
---

s.splitOnToken(lf)[1]:
line2
```

For the following Painless script, I think it would be nice to define a function that does most of the work without being dependent on a specific context, then have minimal code to apply the function where necessary. That makes it easier to test or modify in the future, because you can copy the function out and work on it in other contexts.

chrisberkhout (changes requested):
The proposed commit message could be updated to mention it also adds the transform.

packages/tenable_io/elasticsearch/transform/latest_cdr_vulnerabilities/transform.yml
```diff
@@ -0,0 +1,30 @@
+source:
+  index:
+    - "logs-tenable_io.vulnerability-*"
+dest:
+  index: "security_solution-tenable_io.vulnerability_latest-v1"
+  aliases:
+    - alias: "security_solution-tenable_io.vulnerability_latest"
+      move_on_creation: true
```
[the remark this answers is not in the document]
kcreddy: The first run of the transform `triggered_count: 1` sometimes doesn't process/index the documents. I usually wait for next run of the transform to trigger at which point it processes/indexes the documents into destination.

packages/tenable_io/elasticsearch/transform/latest_cdr_vulnerabilities/fields/base-fields.yml
```diff
@@ -0,0 +1,17 @@
+- name: data_stream.type
+  external: ecs
+- name: data_stream.dataset
+  external: ecs
+- name: data_stream.namespace
+  external: ecs
+  type: keyword
```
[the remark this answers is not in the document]
kcreddy: Sorry, may I know why that is?

packages/tenable_io/data_stream/vulnerability/fields/vulnerability.yml
```diff
@@ -0,0 +1,24 @@
...
+    - name: title
+      type: keyword
+    - name: package
+      type: nested
+    - name: package.fixed_version
+      type: keyword
+    - name: package.name
+      type: keyword
+    - name: package.path
+      type: keyword
```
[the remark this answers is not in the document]
kcreddy: It is present actually. Can you please check:

https://github.com/elastic/integrations/blob/e8ec175e5d23f50868e93d2d9d9848c1a0afce09/packages/tenable_io/elasticsearch/transform/latest_cdr_vulnerabilities/fields/vulnerability.yml#L12-L13

packages/tenable_io/elasticsearch/transform/latest_cdr_vulnerabilities/fields/ecs.yml
```diff
@@ -0,0 +1,92 @@
+# Define ECS constant fields as constant_keyword
+- name: observer.vendor
+  type: constant_keyword
+  external: ecs
+- name: vulnerability.scanner.vendor
+  type: constant_keyword
+  external: ecs
+# Other ECS fields
```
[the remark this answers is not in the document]
kcreddy: Sure let me separate them 👍🏼

packages/tenable_io/data_stream/vulnerability/fields/vulnerability.yml
kcreddy: After discussing with @nick-alayil and also inside the spreadsheet, the field should be `vulnerability.published_date`. Updated in https://github.com/elastic/integrations/pull/13636/commits/4410d9f04368c7d2586d95730a362c270d1383b4

packages/tenable_io/elasticsearch/transform/latest_cdr_vulnerabilities/fields/ecs.yml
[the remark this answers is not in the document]
kcreddy: Done in https://github.com/elastic/integrations/pull/13636/commits/e4841d36273f5733c24b26d865485372a1ba3d33

packages/tenable_io/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
kcreddy: Thanks for the suggestion @chrisberkhout. 
For this script, we are trying to replace a pattern and not a single character. I will use the regexp inside the script processor itself and add functions inside the script itself as you suggested so that it will easy to maintain.

kcreddy: Done in [f9f1dc2](https://github.com/elastic/integrations/pull/13636/commits/f9f1dc2dde23b816e9a87229d7f8079ecbf3247a)

packages/tenable_io/data_stream/vulnerability/fields/vulnerability.yml
nick-alayil: Thanks @kcreddy 🙏 . @maxcold, we need the same to be updated in [CDR guide](https://docs.elastic.dev/security-solution/cloud-security/cdr/3p-dev-guide) as well.

kcreddy: I also noticed that the field is defined as `keyword` in `cloud_security_posture` integration. I think this field should be a `date` type.

packages/tenable_io/data_stream/vulnerability/fields/vulnerability.yml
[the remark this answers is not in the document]
chrisberkhout: Sorry, that was the wrong file. What I noticed was this:

```
diff ./data_stream/vulnerability/fields/package.yml ./elasticsearch/transform/latest_cdr_vulnerabilities/fields/package.yml
```

```diff
7,8d6
< - name: package.path
<   type: keyword
```

I guess ideally it'd be in both the `package.yml` files but neither of the `vulnerability.yml` files.

packages/tenable_io/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Sorry, for some reason I thought it was newlines and tabs, and that the problem was making a string literal with those characters in it.

A pattern is different. Moving it into the script is nice.

Splitting on the pattern rather than doing the substitutions is an option: https://discuss.elastic.co/t/can-we-use-split-processor-with-painless-scripting/98536/2

My original idea was to wrap everything in a function. Something like this:
```
List packagesFromOutput(String output) {
  // ...
}

List results = packagesFromOutput(ctx.tenable_io.vulnerability.output);

ctx.package = ctx.package ?: [];
ctx.vulnerability = ctx.vulnerability ?: [:];
ctx.vulnerability.package = ctx.vulnerability.package ?: [];

ctx.package.addAll(results);
ctx.vulnerability.package.addAll(results);
```

Anyway, these are all optional suggestions.

chrisberkhout (approved):
Looks good :+1: 

Minor point about the `package.path` definitions.

Optional improvement ideas for the Painless script.

packages/tenable_io/data_stream/vulnerability/fields/vulnerability.yml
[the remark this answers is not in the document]
kcreddy: perfect, thanks @chrisberkhout. it is indeed missed. Updated in [333d371](https://github.com/elastic/integrations/pull/13636/commits/333d371a99253c38f23a70b2d53958f551d64ce4)

packages/tenable_io/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -71,10 +90,22 @@ processors:
...
       target_field: tenable_io.vulnerability.asset.netbios.name
       ignore_missing: true
+  - rename:
+      field: json.asset.fqdn
+      target_field: tenable_io.vulnerability.asset.fqdn
+      ignore_missing: true
   - set:
       field: host.name
-      copy_from: tenable_io.vulnerability.asset.netbios.name
+      copy_from: tenable_io.vulnerability.asset.fqdn
```
alexreal1314, line 99: @kcreddy Is it possible to share some docs from Tenable? I would like to understand which fields they have. Because I see for example that out of 1844 documents, 52 are missing host.name which is quite important because alerts calculations are based on it. In comparison in Qualys out of 45k documents only 39 are missing this field.

packages/tenable_io/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -311,10 +372,91 @@ processors:
...
+      if: ctx.json?.plugin?.vuln_publication_date != null && ctx.json.plugin.vuln_publication_date != ''
+      on_failure:
+        - append:
+            field: error.message
+            value: '{{{_ingest.on_failure_message}}}'
   - rename:
       field: json.output
       target_field: tenable_io.vulnerability.output
       ignore_missing: true
+  - script:
```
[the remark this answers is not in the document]
kcreddy: Thanks @alexreal1314 for the testing. Does it mean we need to have a separate `package_nested` field for storing `nested` package data?
We did this in Qualys where [`package`](https://github.com/elastic/integrations/blob/main/packages/qualys_vmdr/data_stream/asset_host_detection/_dev/test/pipeline/test-asset-host-detection.log-expected.json#L39-L51) is an `object` type and [`package_nested`](https://github.com/elastic/integrations/blob/main/packages/qualys_vmdr/data_stream/asset_host_detection/_dev/test/pipeline/test-asset-host-detection.log-expected.json#L180-L190) is `nested` type?

packages/tenable_io/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
kcreddy: Hey @alexreal1314, here's the relevant discussion which lead to `host.name` populating with `asset.fqdn`: https://docs.google.com/spreadsheets/d/1ovxuXOtSXyNGZ_db7TaD5D0pt5m_60yUPe1oxJ1AQ8I/edit?gid=0#gid=0&range=G21 
Here's the Tenable API doc: https://developer.tenable.com/reference/exports-vulns-download-chunk. 
In case `asset.fqdn` field is missing, we could populate with some other field `asset.hostname` or `asset.netbios_name`. Would that be accepted solution?

packages/tenable_io/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
alexreal1314: @kcreddy creating a separate package_nested field won't fix the issue because we rely in the UI on the package field which is of type object. Also package_nested is not part of ECS.
Need to think how to overcome this issue, meanwhile I would think that we should store ctx.tenable_io?.vulnerability?.output fields in the format we have today package.name: array of packages, package.version: array of versions and package.fixed_version array of fixed versions.
WDYT @maxcold ?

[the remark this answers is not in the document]
kcreddy: Related discussion previously: https://github.com/elastic/security-team/issues/9987#issuecomment-2548605862. We wanted to have `package.*` fields but also `package_nested.*` nested fields to preserve the relationships.

packages/tenable_io/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
kcreddy: As per Nick's suggestion, I updated in [0fab27e](https://github.com/elastic/integrations/pull/13636/commits/0fab27e418660405b9ff767edc039c846c185a5d)

packages/tenable_io/data_stream/vulnerability/_dev/test/pipeline/test-vulnerability.log-expected.json
kcreddy: Updated to as per CVSS standard in [fa4aea3](https://github.com/elastic/integrations/pull/13636/commits/fa4aea3ff4089a4084cb6915fd1973b01217ca47)

packages/tenable_io/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
kcreddy: @alexreal1314 @maxcold, this seems to be the only pending comment to merge this PR. 
Let me know if there's anything needed here from my side.

packages/tenable_io/manifest.yml
```diff
@@ -1,15 +1,15 @@
...
+version: "4.0.0"
 description: Collect logs from Tenable Vulnerability Management with Elastic Agent.
 type: integration
 categories:
   - security
   - vulnerability_management
 conditions:
   kibana:
-    version: "^8.18.0 || ^9.0.0"
+    version: "^8.19.0 || ^9.1.0"
```
maxcold, line 12: @kcreddy We've been given guidance by the leadership team that Security Solution treats 8.19 as a bugfix release and that we shouldn't add new features to this release, only 9.1.0. Does Security Integrations follow the same approach? If yes, we need to limit this version to only `^9.1.0` stack version.
For Qyalys we already have the integration published with `^8.19.0 || ^9.1.0` stack condition, we will need to figure out what we need to do on Kibana side to support that, most likely backport all the changes we made to support multiple CVEs, packages etc.

kcreddy: > We've been given guidance by the leadership team that Security Solution treats 8.19 as a bugfix release

@maxcold I wasn't aware of this. Also, Qualys VMDR is the only integration that seems to define this. So the 
Let me know which versions are you targeting for backport, I can update this accordingly. 
BTW `8.17.7` and `8.18.2` FF is today.

kcreddy: @jamiehynds, it seems that we shouldn't be releasing new features targeting `^8.19.0`. Does it also mean new integrations shouldn't be targeting this version? 
Its not GA yet, but would like to know in case any future PRs comes up targeting this version.

packages/tenable_io/data_stream/vulnerability/_dev/test/pipeline/test-vulnerability.log-expected.json
maxcold: Thanks! yes, it makes sense to match CVSS. ECS is very loose and gives CVSS as an example, but it's better to be more consistent at least where we can control it

packages/tenable_io/data_stream/vulnerability/fields/package.yml
maxcold: let's change the mapping to match what we have in Qualys. Even though nested allows to keep the relation between fields in case of multiple packages, we don't support it in UI currently. And to properly support it we will need to figure out the way to make it work for both Qualys and Tenable

packages/tenable_io/data_stream/vulnerability/fields/vulnerability.yml
maxcold: Updated  to top level field in the guide https://github.com/elastic/security-team/pull/12644

packages/tenable_io/manifest.yml
maxcold: I'm waiting for the confirmation from @nick-alayil on the approach to stack versions and availability of CDR related changes in integrations. If we follow the guidance for Security Solution and treat 8.19.0 as bugfix release (it applies to all 8.x versions going forward as far as I understand) , then for Tenable we should have `version: "^9.1.0"`
For Qualys as the integration has already been published, I think we need to backport Kibana changes to 8.19.0, but that's solely on us. Let me know if I can give more context on this topic

packages/tenable_io/data_stream/vulnerability/fields/vulnerability.yml
```diff
@@ -0,0 +1,24 @@
+- name: vulnerability
+  type: group
+  fields:
+    - name: title
+      type: keyword
+    - name: package
```
maxcold, line 6: do we need the package mapping here if we have it separately?

packages/tenable_io/elasticsearch/transform/latest_cdr_vulnerabilities/fields/vulnerability.yml
```diff
@@ -0,0 +1,24 @@
+- name: vulnerability
+  type: group
+  fields:
+    - name: title
+      type: keyword
+    - name: package
```
maxcold, line 6: here as well

kcreddy: Updated in [35a8012](https://github.com/elastic/integrations/pull/13636/commits/35a801270b75e1f05feab4b0096bd800db49b63c)

kcreddy: Removed this field in [899aa71](https://github.com/elastic/integrations/pull/13636/commits/899aa71500eae8ddb070dea7508d3a03b4b60270)

packages/tenable_io/data_stream/vulnerability/fields/vulnerability.yml
kcreddy: Removed this field in [899aa71](https://github.com/elastic/integrations/pull/13636/commits/899aa71500eae8ddb070dea7508d3a03b4b60270)

maxcold (approved):
tested everything except the latest change of removing `vulnerability.package`. Looks good! Thanks for the very fast feedback loop!

maxcold:
For visibility: though approved, we are waiting from product and leadership decision on whether we can make these changes available starting from both 8.19 and 9.1 or only 9.1 . I will update when we have a final decision. Pls don't merge before that

packages/tenable_io/manifest.yml
kcreddy: After discussing with @maxcold, we are going ahead with version constraint: `version: "^8.19.0 || ^9.1.0"`. Will go ahead and merge the PR.

cc: @nick-alayil

packages/tenable_io/changelog.yml
```diff
@@ -1,4 +1,15 @@
 # newer versions go on top
+- version: "4.0.0"
+  changes:
+    - description: Add mappings required for Cloud Detection and Response (CDR) vulnerability workflow.
```
andrewkroh, line 4: I think it would be helpful to specify what the breaking changes are and include any advice we might have for users who are upgrading (if applicable).

kcreddy: Just missed this comment and merged it 😞.

I will create a quick PR with list of changes and update documentation regarding transform and its requirement.
Essentially there are few important mappings that were updated and addition of a transform adds atleast 2X data, and also requires Transform nodes, all of which the users could be made aware of.

kcreddy: Created https://github.com/elastic/integrations/pull/14130

PR #13638, opened 2025-04-22
[cisco_meraki_metrics] scale values in device channel utilization so they display correctly as percentages

tommyers-elastic:
scale values in device channel utilization so they display correctly as percentages

## Checklist

### DIFF
--- a/packages/cisco_meraki_metrics/data_stream/device_health/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/cisco_meraki_metrics/data_stream/device_health/elasticsearch/ingest_pipeline/default.yml
@@ -4,8 +4,42 @@ processors:
   - script:
       lang: painless
       source: >
-        if (ctx.meraki != null && ctx.meraki.uplink != null && ctx.meraki.uplink.loss != null && ctx.meraki.uplink.loss.pct != null) {
-            ctx.meraki.uplink.loss.pct = ctx.meraki.uplink.loss.pct / 100;
+        // some values have unit 'percent' in the mappings; we need to scale them down from 0->100 to 0->1.
+        // we round to 4 decimal places to avoid floating point errors.
+
+        if (ctx.meraki != null) {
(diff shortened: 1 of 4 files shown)


gpop63 (approved):

packages/cisco_meraki_metrics/data_stream/device_health/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -4,8 +4,42 @@ processors:
   - script:
       lang: painless
       source: >
-        if (ctx.meraki != null && ctx.meraki.uplink != null && ctx.meraki.uplink.loss != null && ctx.meraki.uplink.loss.pct != null) {
-            ctx.meraki.uplink.loss.pct = ctx.meraki.uplink.loss.pct / 100;
+        // some values have unit 'percent' in the mappings; we need to scale them down from 0->100 to 0->1.
+        // we round to 4 decimal places to avoid floating point errors.
+
+        if (ctx.meraki != null) {
```
ishleenk17, line 10: def scalePercent(value) {
    return value != null ? Math.round(value / 100 * 10000) / 10000.0 : null
}

We can define this and use this function to make it slightly more readable.

ishleenk17 (approved):
1 nit. Else looks good!

packages/cisco_meraki_metrics/data_stream/device_health/elasticsearch/ingest_pipeline/default.yml
tommyers-elastic: i don't think you can define functions in ingest pipeline script processors :(

ishleenk17: We can, in a slighlty different syntax.Eg [here](https://github.com/elastic/integrations/blob/e5a63de43184291164ebb7a904498611ced53ea6/packages/o365_metrics/data_stream/teams_call_quality/elasticsearch/ingest_pipeline/default.yml#L30C16-L30C28)

PR #13640, opened 2025-04-22
Make kv more resillient and fingerprint more accurate

ash-darin:
- Enhancement

## Proposed commit message

The fingerprint creation is broken.
-  @timestamp is renamed beforehand to "event.created" and is therefore always empty [here](https://github.com/ash-darin/integrations/blob/a0218ac45b5056f331fd566be81561210e69c546/packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml#L108)  and therefore useless . 
- The current constellation creates duplicates as practical tests show. The new constellation has created no duplicates in a 14 day run.

A combination of Sophos/Ironport creates strings in Email-headers, that show up in the logging and look exactely like 

`"; ` . This breaks the kv parser while taking the event apart. 

Fortunately some more spaces follow, so adding "(?=\w)) to the regex ignores those occurences. If \w is considered to "narrow", \S should also be possible. 

## Checklist

- [x] I have added an entry to the package's `changelog.yml` file.

## Author's Checklist

- [ ] Add PR number to changelog

If it behaves as before, nothing is lost.


ash-darin:
> /test

?? I don't understand?

taylor-swanson (commented):
Change seems fine to me. I'm not that worried about `\w` vs. `\S`, unless we expect there to be special/international characters for keys.

We're missing a pipeline test, specifically for the `kv` processor change. We need a test both to verify functionality, but also for regression testing to make sure it isn't accidentally changed in the future.

Would you be able to provide an additional test case for the pipeline tests? Alternatively, if you provide a sample log that utilizes this new change, I'd be happy to add it as well.

`/test` triggers the CI tests.

taylor-swanson:
To be more specific, a log that includes the email_headers field (especially one with multiple spaces), since it doesn't look like we have any existing tests that cover that field, yet.

Make sure any sensitive information is removed/anonymized from the log.

ash-darin:
> To be more specific, a log that includes the email_headers field (especially one with multiple spaces), since it doesn't look like we have any existing tests that cover that field, yet.
> 
> Make sure any sensitive information is removed/anonymized from the log.

I've scaled down the PR in order to have more time to compile and review the test data. I hope that is acceptable. I will open a separate PR to add my changes to the KV parser.

taylor-swanson:
@ash-darin, sounds good to me.

Only remaining issue is the merge conflicts. Could you merge main and update the version to `1.40.1`?

ash-darin:
> @ash-darin, sounds good to me.
> 
> Only remaining issue is the merge conflicts. Could you merge main and update the version to `1.40.1`?

oh swell.. github mislead me to delete all commits.. just one click away from desaster..

I don't have the energy to recreate this...

Issue #12988, opened 2025-03-06
[Atlassian Jira]: Duplicate record pulls due to cursor not progressing


Hythloday-zero:
### Integration Name

Atlassian Jira [atlassian_jira]

### Dataset Name

atlassian_jira.audit

### Integration Version

1.27.0

### Agent Version

8.16.3

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.17.1

### OS Version and Architecture

Elastic Cloud (GovCloud) i3en

### Error Message

As yet, no error message discovered.  Simply receiving duplicate records each time the integration pulls from the API.

### Event Original

Entire ingested records (including event.original) for duplicated event:

{
    "_index": ".ds-logs-atlassian_jira.audit-default-2025.02.25-000048",
    "_id": "WRd8PpUB8gda2vdpE7Hc",
    "_version": 1,
    "_source": {
        "agent": {
        "name": "SERVERNAME",
        "id": "36f5560b-a286-4354-9bd3-8045937719d7",
        "type": "filebeat",
        "ephemeral_id": "05f00c93-1bb1-4923-8f11-c130d0b8faf1",
        "version": "8.16.1"
        },
        "elastic_agent": {
        "id": "36f5560b-a286-4354-9bd3-8045937719d7",
        "version": "8.16.1",
        "snapshot": false
        },
        "tags": [
        "preserve_original_event",
        "forwarded",
        "jira-audit"
        ],
        "input": {
        "type": "httpjson"
        },
        "@timestamp": "2025-02-25T18:31:15.406Z",
        "ecs": {
        "version": "8.11.0"
        },
        "data_stream": {
        "namespace": "default",
        "type": "logs",
        "dataset": "atlassian_jira.audit"
        },
        "event": {
        "agent_id_status": "verified",
        "ingested": "2025-02-25T19:01:13Z",
[...]

PR #13665, opened 2025-04-24
[Atlassian JIRA and Atlassian Conflunce] Update Cursor Logic to Remove Duplicate Events

mohitjha-elastic:
## Proposed Commit Message

```
atlassian_jira and atlassian_cloud: update cursor logic to remove duplicate events.

After reviewing the Atlassian Jira[1] and Atlassian Confluence[2] API
documentation, It has been noticed that the APIs return data on or
after the specified start date. Currently, the date from the response
body is getting saved into the cursor. Because of this, when the
request is being made in the next interval, it includes data we've
already fetched—leading to duplicate events being published to
Elasticsearch. Hence, updated the cursor logic, added 1ms to it would
fetch the data afterwards.

This change has been tested on the data available in the test folder as
well as using the mock server.

[1]https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-audit-records/#api-rest-api-3-auditing-record-get
[2]https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-audit/#api-wiki-rest-api-audit-get
```

## Checklist

### To test atlassian_jira integration
Clone integrations repo.
Install the elastic package locally.
Start the elastic stack using the elastic package.
Move to integrations/packages/atlassian_jira directory.
`elastic-package test -v`

### To test atlassian_confluence integration
Clone integrations repo.
Install the elastic package locally.
Start the elastic stack using the elastic package.
Move to integrations/packages/atlassian_confluence directory.
`elastic-package test -v`

## Related issues

- Closes https://github.com/elastic/integrations/issues/12988


kcreddy (commented):
While adding fingerprint also works for this issue, can you check if it can be solved at cursor level itself without having to propagate the duplicate event into the ingest pipeline and then ignoring it?

ShourieG:
> @mohitjha-elastic, While adding fingerprint also works for this issue, can you check if it can be solved at cursor level itself without having to propagate the duplicate event into the ingest pipeline and then ignoring it?

Agree with @kcreddy here, 1st option would be to try and optimise better cursor date handling. If that's not possible we should do fingerprints.

mohitjha-elastic:
@kcreddy @ShourieG 
We currently don’t have any additional fields that can be stored in the cursor. One possible workaround I can think of is adding `+1 second` to the timestamp when storing it in the cursor. However, this might risk missing events if any are published during that one-second window (although I assume that scenario is unlikely).

Also, storing current timestamp in cursor does not seem feasible that might miss the data in case of API failure.

Since the filter fetches data on or after the start time, the only potential duplication after each interval would be those events (It must be only few) with a timestamp exactly matching the cursor timestamp. In that case, this approach seems quite reasonable.

Let me know what you think.

kcreddy:
> @kcreddy @ShourieG We currently don’t have any additional fields that can be stored in the cursor. One possible workaround I can think of is adding `+1 second` to the timestamp when storing it in the cursor. However, this might risk missing events if any are published during that one-second window (although I assume that scenario is unlikely).
> 
> Also, storing current timestamp in cursor does not seem feasible that might miss the data in case of API failure.
> 
> Since the filter fetches data on or after the start time, the only potential duplication after each interval would be those events (It must be only few) with a timestamp exactly matching the cursor timestamp. In that case, this approach seems quite reasonable.
> 
> Let me know what you think.

+1 second could infact miss events. We should increment to lowest unit allowed by the API, which seems to be milliseconds.
https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-audit/#api-wiki-rest-api-audit-get. Can you check if this can be done to avoid missing data?

The problem with fingerprint is that the fields that we believe are unique may not always be the that way and often need to add more fields because users complain of missing data. It might not be bad to just fingerprint on `event.original` field to avoid this problem. There is also a cost to fingerprinting (~25% lower indexing times) because it has to first check of that document exists in the index.

mohitjha-elastic:
Thanks, @kcreddy for clarifying the use case of adding fingerprints. 
Yes, I checked and increment the lowest unit (1 millisecond) to avoid missing data.
Updated the PR.

packages/atlassian_confluence/data_stream/audit/agent/stream/httpjson.yml.hbs
```diff
@@ -128,4 +128,4 @@ publisher_pipeline.disable_host: true
 {{#if processors}}
 processors:
 {{processors}}
-{{/if}} 
\ No newline at end of file
+{{/if}}
```
efd6, line 131: May as well fix the final new line as well.

packages/atlassian_jira/_dev/deploy/docker/files/config.yml
```diff
@@ -76,4 +76,4 @@ rules:
     responses:
       - status_code: 200
         body: |-
-          {"offset":0,"limit":2,"total":4,"records":[]}
+          {"offset":0,"limit":2,"total":4,"records":[]}
```
efd6, line 79: Revert this change?

packages/atlassian_jira/data_stream/audit/agent/stream/httpjson.yml.hbs
```diff
@@ -116,4 +116,4 @@ fields:
 {{#if processors}}
 processors:
 {{processors}}
-{{/if}} 
\ No newline at end of file
+{{/if}}
```
efd6, line 119: Fix final new line as well.

efd6 (commented):
Suggest the following for the commit message; line breaks and no markdown in git commit messages, otherwise unaltered.
```
atlassian_jira and atlassian_cloud: update cursor logic to remove duplicate events.

After reviewing the Atlassian Jira[1] and Atlassian Confluence[2] API
documentation, It has been noticed that the APIs return data on or
after the specified start date. Currently, the date from the response
body is getting saved into the cursor. Because of this, when the
request is being made in the next interval, it includes data we've
already fetched—leading to duplicate events being published to
Elasticsearch. Hence, updated the cursor logic, added 1ms to it would
fetch the data afterwards.

This change has been tested on the data available in the test folder as
well as using the mock server.

[1]https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-audit-records/#api-rest-api-3-auditing-record-get
[2]https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-audit/#api-wiki-rest-api-audit-get
```

kcreddy (approved):
LGTM. Please resolve @efd6 comments before merging.

mohitjha-elastic:
@efd6 @kcreddy 
Updated the commit message and resolved the comments.
Thanks!

efd6 (approved):
Thanks

PR #13666, opened 2025-04-24
Rename `Custom Filestream Logs` -> `Custom Logs (Filestream)`

rdner:
So, it's consistent with the previous `Custom Logs` package it's to replace.

Screenshot 2025-04-24 at 14 18 00
Screenshot 2025-04-24 at 14 32 02

### DIFF
--- a/packages/filestream/_dev/build/docs/README.md
+++ b/packages/filestream/_dev/build/docs/README.md
@@ -1,4 +1,4 @@
-# Custom Filestream Log integration
+# Custom Logs (Filestream) Package
 
 The `filestream` custom input is used to read lines from active log files. It is the
 new, improved alternative to the `log` input. It comes with various improvements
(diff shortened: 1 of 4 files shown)


packages/filestream/_dev/build/docs/README.md
```diff
@@ -1,4 +1,4 @@
-# Custom Filestream Log integration
+# Custom Logs (Filestream) Package
```
belimawr, line 1: The PR looks good @rdner, however I'd still add a warning about migrating from log to filestream.

Maybe add something like this:
```suggestion
# Custom Logs (Filestream) Package

WARNING: Migrating from "Custom Logs (Legacy)" to "Custom Logs (Filestream)" will cause all existing files to be re-ingested.
```

I'm also wondering if we should add some more information:
1. State that in the future we will support the state migration
2. Suggest using `ignore_older` to minimise data duplication while migrating.

rdner: This PR is only about renaming the package. Further changes should be separate.

belimawr (approved):

PR #13703, opened 2025-04-28
[google_workspace] Handle lag time

### LINKED ISSUE #13081: [Google Workspace] Incorrect logic for paginating through out-of-order data
## The problem

We are missing around 1% of events, probably because we assume that new data will have timestamps larger than what we've seen before.

## Background

The Google Workspace reports API has two complications:

* **The most recent data is returned first**
  That is true of the items in a page, and of the pages in a pagination sequence. A new pagination sequence's time range depends on the full time range of the previous sequence, not just on the last page fetched.
  This was addressed previously:
  https://github.com/elastic/integrations/issues/4796 (issue)
  https://github.com/elastic/integrations/pull/4982 (PR)

* **There is a lag between when an event happens and when it is available from the API**
  The lag depends on the data type, as documented in [Data retention and lag times](https://support.google.com/a/answer/7061566).
  There is no guarantee that events will become available in chronological order.
  New logic is needed to correctly handle out-of-order availability.

The integration sets a 2 hour `interval` for each data stream. The description of the `interval` variable says:

> Duration between requests to the API. Google Workspace defaults to a 2 hour polling interval because Google reports can go from some minutes up to 3 days of delay. For more details on this, you can read more at https://support.google.com/a/answer/7061566.

However, this setting does not resolve the issues that arise from the lag.

Having an interval time that is not short relative to the lag time reduces the number of events that will be missed, but it does not avoid misses. For example, in the following scenario, the event at 01:52 is missed. The query at 04:00 can't see it yet, and the query at 06:00 assumes it's already been seen (since 02:01 was already seen). The same problem happens with the data at 03:58.

Image

We can introduce a new variable: lag time, which can be set differently for each data stream, depending on the expected lag. The lag time variable can be used to avoid querying time ranges for which we expect the data is not yet fully available (`endTime=now-lag`). Or, alternatively, to know which time range to repeatedly query.

Especially if we use the strategy of requesting data once, after we are confident that it is available, we can reduce the interval time to 15 or 5 mins.

This issue is relevant to all data streams.

chrisberkhout:
## Proposed commit message

```
[google_workspace] Handle lag time, add proxy and SSL options

Avoid requesting data for time ranges that are expected to have
incomplete data. The lag times are different for each data set, and
default values have been chosen based on guidance in the Google
Workspace Admin documentation[1].

Waiting until the data is all available means we don't miss things if
it becomes available out of order.

HTTP JSON input config files were made identical where possible, which
involved adding Proxy URL and SSL Configuration options to the following
data streams: `admin`, `drive`, `groups`, `login`, `saml`,
`user_accounts`.

[1]: https://support.google.com/a/answer/7061566
```

## Notes for reviewers

Can be read commit-by-commit.

The initial commits refactor to make the relevant `httpjson.yml.hbs` and `cel.yml.hbs` files are identical. See via `find -name '*.yml.hbs' | xargs md5sum | sort`

The `alert` data stream uses a different API and is not relevant.

I manually checked via the system tests that the initial `startTime` and `endTime` values are being chosen correctly.

## Checklist

## Related issues

- Closes #13081

### DIFF
--- a/packages/google_workspace/changelog.yml
+++ b/packages/google_workspace/changelog.yml
@@ -4,6 +4,9 @@
     - description: Handle lag time.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/13703
+    - description: Add Proxy URL and SSL Configuration options where absent.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/13703
 - version: "2.39.1"
   changes:
     - description: Fix default request trace enabled behavior.
--- a/packages/google_workspace/changelog.yml
+++ b/packages/google_workspace/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.39.2"
+  changes:
+    - description: Handle lag time.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/13703
 - version: "2.39.1"
   changes:
     - description: Fix default request trace enabled behavior.
--- a/packages/google_workspace/data_stream/access_transparency/agent/stream/httpjson.yml.hbs
+++ b/packages/google_workspace/data_stream/access_transparency/agent/stream/httpjson.yml.hbs
@@ -34,7 +34,10 @@ request.transforms:
         [[- else -]]
           [[- .cursor.last_response_date -]]
         [[- end -]]
-      default: '[[formatDate (now (parseDuration "-{{initial_interval}}"))]]'
+      default: '[[formatDate (now (parseDuration "-{{initial_interval}}{{lag_time}}"))]]'
+  - set:
+      target: url.params.endTime
+      value: '[[formatDate (now (parseDuration "-{{lag_time}}"))]]'
 response.split:
   target: body.items
   ignore_empty_value: true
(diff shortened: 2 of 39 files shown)


packages/google_workspace/data_stream/access_transparency/agent/stream/httpjson.yml.hbs
```diff
@@ -34,7 +34,10 @@ request.transforms:
         [[- else -]]
           [[- .cursor.last_response_date -]]
         [[- end -]]
-      default: '[[formatDate (now (parseDuration "-{{initial_interval}}"))]]'
+      default: '[[formatDate (now (parseDuration "-{{initial_interval}}{{lag_time}}"))]]'
```
efd6, line 37: I think I am missing something. How does this work when there is a lag time and an initial interval where the units are the same? Is this code based on the assumption that they are not?

Apparently not: https://go.dev/play/p/kBMF448HXj-

Carry on.

packages/google_workspace/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.38.1"
+  changes:
+    - description: Handle lag time.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/13703
```
efd6, line 6: Add note about addition of SSL support?

efd6 (commented):
I think there should be some text in the commit message/title about the harmonisation of the manifests.

packages/google_workspace/data_stream/access_transparency/agent/stream/httpjson.yml.hbs
chrisberkhout: Yeah, I was a bit surprised when I found it works that way.

efd6 (approved):

PR #13712, opened 2025-04-29
ssi: ensure request trace is off by default

### LINKED ISSUE #13710: [BUG] Enable Request Tracer Defaults to null causing request tracing to be collected
With the change to enable request tracer to allow the removal of old trace logs it appears that as no default is set for the variable enable_request_tracer that it is getting set to null. This is causing the request tracer to still be enabled as it has other fields resulting in request tracing to be inadvertently be left on. 

My google workspace integration logs were being flooded with ```write error: write length 1108655 exceeds maximum file size 1048576```. 

The UI shows that request tracing is set to false but an agent diagnostic shows http trace files still being generated. The computed config appears as below:

```
          request:
            tracer:
                enabled: null
                filename: ../../logs/httpjson/http-request-trace-*.ndjson
                maxbackups: 5
```

Upon flipping it to true and then back to false in the UI. A change was detected and I reapplied the agent policy. A subsequent diagnostic showed the tracer logs were removed and the error stopped being generated.

I believe the bug was introduced in this PR #13035

efd6:
## Proposed commit message

```
all: ensure request trace is off by default

The change from truthy to boolean in #13710 resulted in policies with a
null *.tracer.enabled value which is treated as on by the inputs if
there are other *.tracer.* fields set. This change ensures that the value
is false unless altered to be true, fixing behavior in integrations that
have adopted request trace deletion capabilities.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #13710

### DIFF
--- a/packages/1password/changelog.yml
+++ b/packages/1password/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.32.1"
+  changes:
+    - description: Fix default request trace enabled behavior.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/13712
 - version: "1.32.0"
   changes:
     - description: Update Kibana constraint to support 9.0.0.
--- a/packages/1password/changelog.yml
+++ b/packages/1password/changelog.yml
@@ -1,9 +1,4 @@
 # newer versions go on top
-- version: "1.32.1"
-  changes:
-    - description: Fix default request trace enabled behavior.
-      type: bugfix
-      link: https://github.com/elastic/integrations/pull/13712
 - version: "1.32.0"
   changes:
     - description: Update Kibana constraint to support 9.0.0.
--- a/packages/1password/manifest.yml
+++ b/packages/1password/manifest.yml
@@ -1,7 +1,7 @@
 format_version: "3.0.2"
 name: 1password
 title: "1Password"
-version: "1.32.1"
+version: "1.32.0"
 description: Collect logs from 1Password with Elastic Agent.
 type: integration
 categories:
@@ -44,7 +44,6 @@ policy_templates:
           - name: enable_request_tracer
             type: bool
             title: Enable request tracing
-            default: false
             multi: false
             required: false
             show_user: false
--- a/packages/1password/manifest.yml
+++ b/packages/1password/manifest.yml
@@ -1,7 +1,7 @@
 format_version: "3.0.2"
 name: 1password
 title: "1Password"
-version: "1.32.0"
+version: "1.32.1"
 description: Collect logs from 1Password with Elastic Agent.
 type: integration
 categories:
@@ -44,6 +44,7 @@ policy_templates:
           - name: enable_request_tracer
             type: bool
             title: Enable request tracing
+            default: false
             multi: false
             required: false
             show_user: false
--- a/packages/abnormal_security/changelog.yml
+++ b/packages/abnormal_security/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.6.2"
+  changes:
+    - description: Fix default request trace enabled behavior.
+      type: bugfix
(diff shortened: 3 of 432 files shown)


kcreddy (commented):
Is this intentionally ignored?https://github.com/elastic/integrations/blob/main/packages/sailpoint_identity_sc/manifest.yml#L63

taylor-swanson (approved):
LGTM for deployment-and-devices

efd6:
@kcreddy Thanks, no, that was missed. Adding.

kcreddy (approved):

ishleenk17:
@efd6 : For users with existing policies (where this was true by default), these logs will stop getting collected once they upgrade the integrations. That would be a breaking behaviour. 
Although, I see this flag is just used for debugging purpose so might not be bothersome for the user.
WDYT ?

andrewkroh (approved):
I tested the upgrade process with `enable_request_tracer` unset in the policy. After the upgrade, it was assigned a value of `false`.

I didn't test the case where `enable_request_tracer` had a user-defined value in the policy. I assume Fleet would not overwrite a user-provided value simply because the default changed. But still might be worth testing.

efd6:
That is a fly in the ointment. You are correct, good catch. At the moment, all the old behaviour (non-deleting) relies on the existence of the `tracer` object, which is triggered in those integrations by adding the log destination filename ([e.g.](https://github.com/elastic/integrations/blob/037cda7c46f84ca7c1d8055105f5d7fab59fda50/packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs#L3-L6)). This is the vast majority of cases (290).

There are two ways forward:
1. add in `request.tracer.enabled: true` or `resource.tracer.enabled: true` for all the old behaviour cases (many); or
2. back out all the old behaviour cases.

I'm inclined towards the second approach since it greatly simplifies this PR (likely reducing it to only being our team's ownership), but it does leave open the potential risk of the incorrect default on behaviour leaking back in.

efd6:
@andrewkroh It will depend on the package as to whether this will break people. The 1password case I linked to should show that this does indeed break the tracer functionality.

ishleenk17:
> I'm inclined towards the second approach since it greatly simplifies this PR

When you say back out old behavior would that mean this change would not be done to packages where something like [this](https://github.com/elastic/integrations/blob/037cda7c46f84ca7c1d8055105f5d7fab59fda50/packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs#L3) is being used ?

efd6:
@ishleenk17 Yes, that's correct. The back-out is in the second commit here; just an automated revert of all the packages that do not have the new behaviour as identified by `rg -c -g '*.hbs' 'enabled: \{\{enable_request_tracer}}' | cut -d/ -f2|sort|uniq`.

zmoog (approved):

ishleenk17:
> @ishleenk17 Yes, that's correct. The back-out is in the second commit here; just an automated revert of all the packages that do not have the new behaviour as identified by `rg -c -g '*.hbs' 'enabled: \{\{enable_request_tracer}}' | cut -d/ -f2|sort|uniq`.

https://github.com/elastic/integrations/blob/478a859041ed54b68902e5478d03b966fe5f797f/packages/o365_metrics/data_stream/entra_features/agent/stream/cel.yml.hbs#L57

o65_metrics doesn't have the behaviour you mentioned above. But I see the change for that exists. 
Am I looking at something incorrectly ?

efd6:
Another good catch. This was a result of incorrect use of REs. Will remove.

efd6:
@ishleenk17 PTAL

ishleenk17:
> Another good catch. This was a result of incorrect use of REs. Will remove.

Thanks, it now has no integrations left from observability :) 
So no approval required now!

ishleenk17 (approved):

PR #13721, opened 2025-04-30
entityanalytics_okta,okta: record okta domain into host.name in ingested documents

efd6:
## Proposed commit message

```
entityanalytics_okta,okta: record okta domain into host.name in ingested documents

This change affects two packages, entityanalytics_okta and okta in a
coordinated way.

entityanalytics_okta:

Rather than waiting for a stack release this adds the okta_domain value
via a beats processor, placing it in a location that is unlikely to
collide with data in the input's event data in the future. Move this to
host.name in ingest, falling back to reasonable location on failure.

okta:

This records the Okta Domain value from the Okta API URL provided by the
configuration. This is chosen over using the okta_domain value since
the URL is always present, so this simplifies the logic. The URL is
passed outside the event.original to avoid collision. This means that
it does not end up in the event.original. The actual Okta Domain is
obtain from the URL in the ingest pipeline and then placed gingerly in
host.name if possible, falling back to the same locations used in the
entityanalytics_okta integration in order to harmonise the two
integrations. The work to do this is done last in the pipeline to allow
an unlikely failure to not interfere with other parts of the pipeline,
but to provide an informative error in that case that that does
happen.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


chemamartinez (approved):
LGTM

PR #13755, opened 2025-05-01
zscaler_zpa: fix handling of multiple remote IPs, and event categorisation

### LINKED ISSUE #13754: [Zscaler Private Access]: audit data stream does not tolerate remoteIP lists
### Integration Name

Zscaler Private Access [zscaler_zpa]

### Dataset Name

_No response_

### Integration Version

1.22.1

### Agent Version

8.17.1

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.17.1

Elastic Cloud

### Software/API Version

_No response_

### Error Message

CLient_ip and relevant_ip fields are being set to ignored.  These fields contian more than a single IP address and causing the pipeline to fail

### Event Original

    "event.original": [
      "{\"ModifiedTime\":\"\",\"CreationTime\":\"2025-04-30T16:23:40.000Z\",\"ModifiedBy\":288263728720249833,\"RequestID\":\"12d6eccc-718c-4657-b267-83cc1c3f35f6\",\"SessionID\":\"1samau4fwi7xbsf3317mkd5vz\",\"AuditOldValue\":\"\",\"AuditNewValue\":\"{\\\"loginAttempt\\\":\\\"2025-04-30 16:23:40 UTC\\\",\\\"remoteIP\\\":\\\"x.x.x.x, x.x.x.x\\\"}\",\"AuditOperationType\":\"Sign In\",\"ObjectType\":\"Authentication\",\"ObjectName\":\"xxxx\",\"ObjectID\":xxxxx,\"CustomerID\":xxxxx,\"User\":\"xxxx\",\"ClientAuditUpdate\":1}"
    ]

### What did you do?

Installed the integrations and started sending data into elastic

### What did you see?

Zscaler data was being marked as poor quality and fields were missing.  Investigating showed that some fields were being ignored because they were of the incorrect type.  After some additional investigation, it was found that these fields had contained more than one IP address, which was causing the pipeline to process these IP addresses as the wrong type.

These fields contain more than one IP address due to the environment being double-nat'ed and ZScaler was adding both Client IP address and Relevant IP address from each layer of the NAT

It was expected that the Client IP and Relevant_IP fields in the data and that the data be considered healthy

### Anything else?

_No response_

efd6:
## Proposed commit message
```
zscaler_zpa: fix handling of multiple remote IPs, and event categorisation

ZScaler appears to sometimes send events with multiple remote IPs. So
split the list an raise client.ip to an array if this is the case. The
Zscaler ZPA documentation does not make clear what the semantics of this
case are, so we bend the ECS a little here.

Also fix handling of event.type and event.category to ensure that they
agree with ECS definitions, and add event.outcome mappings.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #13754


packages/zscaler_zpa/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -216,12 +260,25 @@ processors:
           ctx.server = new HashMap();
           ctx.server.address = valuesMap?.domainOrIpAddress;
         }
+  - split:
+      field: client.ip
+      separator: ', *'
+      if: ctx.client?.ip instanceof String && ctx.client.ip.contains(',')
```
kcreddy, line 266: I think having a list of `client.ip`/ `source.ip` will prevent us from using `geoip` processor on them. Maybe we should find the public IP among the 2 IPs and populate just with it?

efd6: I'm following this up to see if there is a known behaviour we can rely on to do this.

efd6: @StellaFreyju can you provide some insight here? I understand that the first element of the array is the public IP address, but that the second may be somewhat public. I do not think that it is a good idea to do geoip look-ups on both, but doing it on the first element seems fine; users interested in the second (and subsequent) addresses can do look-ups in an `@custom` pipeline. Does this sound reasonable to you?

efd6:
@kcreddy I haven't had a response, so I would like to merge this either as is. PTAL

kcreddy (approved):
Sounds good. Thanks!

Issue #13752, opened 2021-08-25
[SLES 15]: No "system.auth" logs for system integration under Data Streams tab for SLES 15 linux agent.


amolnater-qasource:
**Kibana version:** 7.15.0 BC-2 Kibana cloud environment

**Host OS and Browser version:** All, All

**Build details:**
```
Build: 43742
Commit: 113d5f2a745ab9be7d567d13ff906dfd0f5b7a9a
Artifact Link: https://staging.elastic.co/7.15.0-d7816905/downloads/beats/elastic-agent/elastic-agent-7.15.0-linux-x86_64.tar.gz
```

**Preconditions:**
1. 7.15.0 BC-2 Kibana cloud environment should be available.
2. An SLES 15 linux agent must be installed with Default policy having System integration.

**Steps to reproduce:**
1. Login to Kibana environment.
2. Navigate to Data Streams tab.
3. Select `System` from Integration dropdown and `logs` from type dropdown.

**Logs:**
[logs.zip](https://github.com/elastic/integrations/files/7044131/logs.zip)

**Expected Result:**
"system.auth" logs should be generated for system integration under Data Streams tab for SLES 15 agent.

**Screenshot:**
![13](https://user-images.githubusercontent.com/77374876/130736666-9814dffc-bf95-4e61-9ea8-1456e458924b.JPG)

**Note:**
This issue is **not observed for Ubuntu 20** machines with .tar artifact.

pierrehilbert:
I just opened the PR to fix it.

PR #13759, opened 2025-05-02
Use journald input by default when running system integration for SLES 15-SP6

pierrehilbert:
## Proposed commit message

Use journald input by default when running system integration for SLES 15-SP6.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/system/changelog.yml
+++ b/packages/system/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.68.3"
+  changes:
+    - description: Change default to use journald input for SLES 15 SP6.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/xxxx
 - version: "1.68.2"
   changes:
     - description: Handle events without event_data
--- a/packages/system/manifest.yml
+++ b/packages/system/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.0.2
 name: system
 title: System
-version: "1.68.2"
+version: "1.68.3"
 description: Collect system logs and metrics from your servers with Elastic Agent.
 type: integration
 categories:
@@ -40,12 +40,12 @@ policy_templates:
               to find the available keys and to
               [Conditions](https://www.elastic.co/guide/en/fleet/current/dynamic-input-configuration.html#conditions)
               on how to use the available keys in conditions. It defaults to
-              '${host.os_version} != "12 (bookworm)" and (${host.os_platform} != "amzn" or ${host.os_version} != "2023")'
+              '${host.os_version} != "12 (bookworm)" and (${host.os_platform} != "amzn" or ${host.os_version} != "2023") and (${host.os_platform} != "sles" or ${host.os_version} != "15-SP6")'
             type: text
             multi: false
             required: false
             show_user: true
-            default: ${host.os_version} != "12 (bookworm)" and (${host.os_platform} != "amzn" or ${host.os_version} != "2023")
+            default: ${host.os_version} != "12 (bookworm)" and (${host.os_platform} != "amzn" or ${host.os_version} != "2023") and (${host.os_platform} != "sles" or ${host.os_version} != "15-SP6")
       - type: journald
         title: Collect logs from System instances using Journald
         description: Collecting System auth and syslog logs using Journald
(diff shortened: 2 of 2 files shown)


rdner:
@pierrehilbert there are not testing steps in the description.

Was it tested or you would like me to run the tests?

pierrehilbert:
@amolnater-qasource tested the fix by manually changing the configuration here: https://github.com/elastic/integrations/issues/13752#issuecomment-2844912849 

But always great to have more folks confirming this is working the correct way.

rdner (approved):

packages/system/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.0.1"
```
shmsr, line 2: As this is not a bugfix but a behaviour change (default change based on new condition), I think we should do a minor upgrade (2.0.0 -> 2.1.0) instead of a patch upgrade (2.0.0 -> 2.0.1)

packages/system/changelog.yml
```diff
@@ -1,4 +1,9 @@
...
+  changes:
+    - description: Change default to use journald input for SLES 15 SP6.
+      type: enhancement
```
ishleenk17, line 5: To me this looks to be a bugfix, since we don't support logfile for `sles` since the very beginning. Just that we had not added that to our config.
```suggestion
      type: bugfix
```

ishleenk17:
@pierrehilbert 
Also, there might be more such platform/os_version combinations we might have missed/not known. 
We always have the condition config option for the user to change that. So I don't think this should be a concern. WDYT ?
Current behaviour if in case these log inputs are used is that just they won't see any logs, is that right?

packages/system/manifest.yml
```diff
@@ -58,12 +58,12 @@ policy_templates:
               to find the available keys and to
               [Conditions](https://www.elastic.co/guide/en/fleet/current/dynamic-input-configuration.html#conditions)
               on how to use the available keys in conditions. It defaults to
-              '${host.os_version} == "12 (bookworm)" or (${host.os_platform} == "amzn" and ${host.os_version} == "2023")'
+              '${host.os_version} == "12 (bookworm)" or (${host.os_platform} == "amzn" and ${host.os_version} == "2023") or (${host.os_platform} == "sles" or ${host.os_version} == "15-SP6")'
             type: text
             multi: false
             required: false
             show_user: true
-            default: ${host.os_version} == "12 (bookworm)" or (${host.os_platform} == "amzn" and ${host.os_version} == "2023")
```
shmsr, line 66: qq: Why just SLES 15 (SP6)? I assume journald is primary logging system for other service packs as well other versions of SLES?

pierrehilbert: Good point, could probably be replaced by startsWith(${host.os_version}, "15") == true

shmsr: https://documentation.suse.com/sles/12-SP5/html/SLES-all/cha-journalctl.html (see first line)

Also SLES 12. SLES 12 onwards, they are using journal. SLES 13, 14 doesn't exist. So I think we should handle SLES 12 and 15 with startsWith

https://www.theregister.com/2018/06/25/suse_linux_enterprise_15/

^ Reason why they skipped 13 and 14
> SUSE jumped from SLES 12 to 15 as both 13 and 14 are unlucky numbers. While many are aware of the negative connotations attached to the number 13, in China, 14 is also considered unlucky. This is because the number 4 in Chinese (四, pinyin: sì; Cantonese Yale: sei) is almost homophonous to the word for 'death' (死 pinyin: sǐ; Cantonese Yale: séi). Therefore, SUSE was asked not to use these numbers by partners and customers alike.

cmacknz: With the condition we want to catch the point where an existing OS release no longer included syslog by default, not just the point where journald was available.

For existing OS releases that included both syslog and journald, this integration is working without issue by reading from the syslog files. We could switch from syslog to journald there but it opens us to the chance that it breaks something unexpectedly.

pierrehilbert:
@ishleenk17 / @shmsr Can I get another review here please?

ishleenk17 (approved):

PR #13765, opened 2025-05-02
[System] Add pipeline for AD FS Auditing to security data stream.

### LINKED ISSUE #11539: Support processing of AD FS logs with the system integration
When using the [system.security integration](https://www.elastic.co/docs/current/integrations/system#system), the agent is parsing the ADFS server name and service user name (copied from **winlog.computer_name** and **winlog.user.name**) into the "host.name" and "user.name" fields for the event. On the other hand, Kerberos events (for instance event id 4768), overwrite "user.name" with the authenticating user information.

It would be useful to have specific processing for [AD FS events — with ids between 1200 and 1210](https://learn.microsoft.com/en-us/windows-server/identity/ad-fs/troubleshooting/ad-fs-tshoot-logging#types-of-events).

The list at https://github.com/elastic/integrations/issues/3769 that track what is parsed with the system.security integration doesn't include these events in particular.

jvalente-salemstate:
## Proposed commit message

This PR adds support for AD FS Auditing (`event.provider: "AD FS Auditing"`) to `system.security`.  These events have an `event.code` of `1200-1207`.   Most of the information in the event logs is found in an XML object stored in `winlog.event_data.param2`.

Rather than trying to parse the XML, I am using the `grok` processor to extract most of this information. 

Rather than using the script in `standard.yml`, I chose to place an additional painless script processor for ECS categorization within the ADFS pipeline because it seems that these codes may not be exclusive to ADFS. This avoids setting values when not appropriate. 

- All four ECS categorization fields have been set as completely as possible
- Additionally, `event.action` and `event.reason` are set wherever possible.
- I've tested with roughly 4900 exported events from our production environment with no pipeline failures. `event.code: 1207` has not been tested. 

## Checklist

## Related issues

Closes #11539


yannickwellens:
Any chance this can be pushed through?

yannickwellens:
Bump

marc-gr:
Thanks for the PR! This one might be worth waiting for https://github.com/elastic/elasticsearch/pull/130337 so we can use proper xml parsing on the ingest pipeline, which seems might be a bit clearer for future maintainers

PR #13781, opened 2025-05-05
Align datatypes with ECS

ash-darin:
# Align datatypes with ECS

Changed the datatype for the fields 

- event.module
- event.dataset

from constant_keyword to keyword. The usage of constant_keyword does not conform to ECS:

https://www.elastic.co/docs/reference/ecs/ecs-event#field-event-dataset
https://www.elastic.co/docs/reference/ecs/ecs-event#field-event-module

ECS compliance is recommended per Development guidelines:

https://www.elastic.co/docs/extend/integrations/general-guidelines#_ecs_compliance

> An integration package should be compliant with the most recent version of ECS.

## Checklist

- [x] I have added an entry to the package's `changelog.yml` file.


efd6 (commented):
The use here is not inconsistent with ECS (see [ref](https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#optimizing-indices-using-constant-keyword-fields)). Is there a reason beyond the comment in the PR description that you feel that this change is needed?

ash-darin:
If you ever want to have different event.module or event.dataset field content in the same index, then yes. constant_keyword will not allow that.

> Is having these fields be `constant_keyword` causing some problem?

andrewkroh:
> If you ever want to have different event.module or event.dataset field content in the same index, then yes. constant_keyword will not allow that.

Correct, but that's not how integrations are designed. Each integration has a dedicated data stream for each dataset, meaning only one unique `event.{module,dataset}` will be used within that specific data stream. There won't be any mixing within data streams.

This is different for the destination index of transforms (as described [here](https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#optimizing-indices-using-constant-keyword-fields) under _special considerations_). Sometimes, transforms may source data from multiple data streams, leading to different values for `event.dataset`. Therefore, the mappings for transforms might need to avoid using `constant_keyword` for those fields. However, since the Darktrace integration doesn't use transforms, I don't think this would be an issue here.

ash-darin:
I'll not argue with you. It's your design decision.

PR #13816, opened 2025-05-07
[cisco_ios] Fix parsing of hostnames that start with a digit

mjwolf:
## Proposed commit message

This PR improves the Cisco IOS ingest pipeline to properly handle hostnames that begin with digits. Even though such hostnames do not strictly conform to the Cisco IOS naming specification, hostnames starting with digits are commonly used in real-world configurations.

To do this we modified the grok pattern to allow such hostnames, e.g. 

```
<190>3132811: 3router 3132807: Jul 14 2023 08:23:43.398 UTC: %FOO-6-BAR: Numeric hostname with sequence
```

Due to the limitations of the grok parser's regex support, this also allowed all-digit hostnames, which caught cases like

```
<190>2361044: Jul 14 2023 08:23:43.398 UTC: %FOO-6-BAR: Test header format
```

where `2361044` is in the same place as the hostname but is instead a sequence number (compare
```
<190>sw01: Jul 14 2023 08:23:43.398 UTC: %FOO-6-BAR: Test header format
```
). We had to special-case this processing with additional ingest processors.

Another problem was created by timestamps like this

```
<46>: 2023 Aug 27 21:40:50 PDT: %...
<46>: 2023host Aug 27 21:40:50 PDT: %...
```

where `2023` is in fact part of the timestamp but `2023host` is a valid hostname (and the timestamp does not contain the year at all). We fixed this issue by making sure that there is an additional attempt to recognize the timestamp pattern before proceeding with recognizing it as a hostname.

These cases are covered by new tests in `test-numeric-hostname.log`. Note that `@timestamp` is a dynamic field (the timestamp without a year refers to a current year), so correctness of some tests is not checked automatically.

## Checklist


packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -36,26 +36,58 @@ processors:
...
-        CISCO_HOSTNAME: '[a-zA-Z][.0-9a-zA-Z_-]{0,253}[0-9a-zA-Z]?'
+        CISCO_HOSTNAME: '[0-9a-zA-Z][.0-9a-zA-Z_-]{0,253}[0-9a-zA-Z]?'
         CISCO_TZ: '[a-zA-Z]{1,4}([+-]\d{1,2}|[+-]\d{2}:\d{2})?'
   - grok:
       field: _temp_.message
       tag: grok_message
       patterns:
         - '^%%{GREEDYDATA:message}$'
         - '^%{GREEDYDATA:_temp_.generic_message}$'
+  # Do not allow all-digit hostnames, these are sequence numbers. Due limitations of grok, and how
```
ilyannn, line 59: Is the Painless script pattern matching for number really necessary? It looks like we can just grok `log.syslog.hostname` as `NUMBER` into `cisco.ios.sequence` – if that works, we have a sequence.

We might need to match it against the timestamp pattern, but can't that also be done with grok?

taylor-swanson (commented):
Do we have a sample event that we can use as a test? Would like to see this in action.

mjwolf:
> Do we have a sample event that we can use as a test? Would like to see this in action.

The test in `test-yearfirst-timestamp.log` is what required special timestamp handling in the script

taylor-swanson:
> > Do we have a sample event that we can use as a test? Would like to see this in action.
> 
> The test in `test-yearfirst-timestamp.log` is what required special timestamp handling in the script

So re-generating `test-yearfirst-timestamp.log` results in this timestamp

```
2024 Aug 27 21:40:50 PDT

"@timestamp": "2025-08-27T21:40:50.000-07:00"
```
(I deleted the expected file and re-generated it, since `@timestamp` is marked as a dynamic field, and therefore doesn't get updated or verified):

Note the year. Unfortunately, this change is handling the year incorrectly (or not at all). `main` does not have this issue. This is the only regression, I also re-created the other expected files and they are the same.

taylor-swanson (approved):

PR #13822, opened 2025-05-07
varonis: add pre-processor option to allow ingestion of non-conforman…

### LINKED ISSUE #13764: [Varonis]: Ingest Pipeline Error
### Integration Name

Varonis [varonis]

### Dataset Name

varonis.logs

### Integration Version

0.1.0

### Agent Version

8.16.3

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.18.0

Ubuntu 24.04

### Software/API Version

_No response_

### Error Message

The ingest pipeline appears to be attempting to parse backslash values `\` as being escaped backslashes like `\\`, but none of the messages contain escaped backslashes. This is causing ingest pipeline errors:

```
malformed value for duser at pos 224, malformed value for filePath at pos 290, malformed value for msg at pos 480, malformed value for dpriv at pos 760, error in field 'deviceCustomDate1': value is not a valid timestamp, error in field 'cnt': strconv.ParseInt: parsing "": invalid syntax]
```

The sanitized `event.original` that I have provided retains the same character length of each of the sanitized values, so the above error should still be generated the same. 

### Event Original

<14>Apr 28 15:59:01 VARONISSRVR1 CEF:0|Varonis Inc.|DatAdvantage|8.7.1|5015|DS object permission added|3|rt=Apr 28 2025 15:01:42 cat=Alert cs2=Permission changes on OU cs2Label=RuleName cn1=142 cn1Label=RuleID end=Apr 28 2025 14:53:02 duser=contosofoo.com\Bourne Jr., Jason dhost=AD-contosofoo.com filePath=contosofoo.com\Home Office Site\Computers\Containered fname=Containered act=DS object permission added dvchost=Compute outcome=Success msg=Permissions were added to "Containered" for group "contosofoo.com\Group Admins-Region_200" cs3= cs3Label=AttachmentName cs4= cs4Label=ClientAccessType deviceCustomDate1= fileType= cs1= cs1Label=MailRecipient suser= cs5= cs5Label=MailboxAccessType cnt= cs6= cs6Label=ChangedPermissions oldFilePermission= filePermission=Write dpriv=contosofoo.com\Group Admins-Region_200 start=Apr 28 2025 14:53:02

### What did you do?

No special configurations have been done. 

### What did you see?

Error:

```
malformed value for duser at pos 224, malformed value for filePath at pos 290, malformed value for msg at pos 480, malformed value for dpriv at pos 760, error in field 'deviceCustomDate1': value is not a valid timestamp, error in field 'cnt': strconv.ParseInt: parsing "": invalid syntax]
```

No error. 

### Anything else?

_No response_

efd6:
## Proposed commit message

```
varonis: add pre-processor option to allow ingestion of non-conformant CEF messages

Add a config option to add "pre-processors" which will run before
the "decode_cef" processor on the agent. CEF messages from some sources
may be malformed by not following the CEF spec exactly, which can then
cause problems in the "decode_cef" processor. Having a pre-processor
will allow running scripts that can correct the malformed CEF messages,
and allow decode_cef to process them correctly.

New test sample provided in issue from user.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #13764


efd6:
> It would be good for the regex to match single backslashes rather than any, so that running it on correctly escaped backslashes doesn't double up the escaping.

I don't think this is a solvable problem (it is intrinsic to the nature of the in-band escaping solution); it is entirely possible that there is intentially a case where there is a literal `\\` in the text but the text is unescaped. This would be the case for an absolute Windows path such as `\\host\share\foo`.

> Since we're only adding escaping to backslashes, at least for now, I think it's better for the documentation to mention that, or at least hint at it, to avoid users hoping it will magically fix any data issues they have.

We're not actually doing anything at this point. The JS that's added is only in tests, not in the manifest, so it is still up to the user to configure the pre-processors that they what to add. They would need to understand the code that they used

> The `test()` function could exercise the `process()` function rather than just the the replace.

The test that I added was purely to test that the escaping sequence was correct, not that JS calls work. I think that what is here is OK>

chrisberkhout (approved):
All good :+1: 

> > It would be good for the regex to match single backslashes rather than any, so that running it on correctly escaped backslashes doesn't double up the escaping.
> 
> I don't think this is a solvable problem (it is intrinsic to the nature of the in-band escaping solution); it is entirely possible that there is intentially a case where there is a literal `\\` in the text but the text is unescaped. This would be the case for an absolute Windows path such as `\\host\share\foo`.

Good point.

> > Since we're only adding escaping to backslashes, at least for now, I think it's better for the documentation to mention that, or at least hint at it, to avoid users hoping it will magically fix any data issues they have.
> 
> We're not actually doing anything at this point. The JS that's added is only in tests, not in the manifest, so it is still up to the user to configure the pre-processors that they what to add. They would need to understand the code that they used

My mistake. I thought it was adding a kind of "fix data" option, but it's just the ability to add whatever pre-processors the user writes, so the README description is good.

> > The `test()` function could exercise the `process()` function rather than just the the replace.
> 
> The test that I added was purely to test that the escaping sequence was correct, not that JS calls work. I think that what is here is OK.

Yeah, it's fine. It was just an idea.

Issue #10272, opened 2024-06-26
[New Integration] AWS Config


cpascale43:
### Description
AWS Config provides configurations and compliance status of AWS resources within an account. It records configuration changes and can generate detailed snapshots of the current configuration state of resources. This helps in analyzing AWS environments to ensure compliance and identify potential security issues. 

### Architecture
AWS Config captures and delivers detailed configuration snapshots into an S3 bucket, which can then be forwarded to Elastic via SQS. Configuration snapshots are point-in-time captures of the configurations of all the resources in the AWS environment. These snapshots are formatted as JSON files. These snapshots can be delivered to an S3 bucket either manually via the [`DeliverConfigSnapshot`](https://docs.aws.amazon.com/config/latest/APIReference/API_DeliverConfigSnapshot.html) API call or automatically based on a schedule set in the AWS Config recorder settings. 

Refs:
- [Understanding the differences between configuration history and configuration snapshot files in AWS Config](https://aws.amazon.com/blogs/mt/configuration-history-configuration-snapshot-files-aws-config/)
- [Delivering Configuration Snapshots to an Amazon S3 Bucket](https://docs.aws.amazon.com/config/latest/developerguide/deliver-snapshot-cli.html)
- [AWS Config API Reference](https://docs.aws.amazon.com/config/latest/APIReference/Welcome.html)

### Metrics & Dashboards 
Alongside configuration snapshots, the integration provides various metrics to users such as compliance scores and notification failures. These are:
- **aws.config.change_notifications_delivery_failed (count):** Measures the number of failed attempts to deliver change notifications to an Amazon SNS topic.
- **aws.config.compliance_score (gauge):** Indicates the percentage of compliant rule-resource combinations in a conformance pack.
- **aws.config.config_history_export_failed (count):** Tracks the number of unsuccessful attempts to export configuration history to an S3 bucket.
- **aws.config.config_snapshot_export_failed (count)**: Counts the failures in exporting configuration snapshot files to an S3 bucket.
- **aws.config.configuration_items_recorded (count):** Logs the number of configuration items recorded for each resource or all resource types.
- aws.config.configuration_recorder_insufficient_permissions_failure (count): Monitors the number of permission failures due to insufficient IAM role policies for the configuration recorder.

Dashboards provide insight into resource configurations, compliance status and changes over time. Some specific ideas are:
- **Resource Inventory:** Provide a comprehensive view of all AWS resources, highlighting configuration states and changes.
[...]

PR #13595, opened 2025-04-17
[M365 Defender][Microsoft Defender Endpoint] Add support of vulnerability data-stream

sharadcrest:
## Proposed commit message

This release introduces the vulnerability data stream, 
along with its associated dashboard and visualizations.

Vulnerability fields are mapped to their corresponding ECS fields where possible.

Test samples were derived from live data samples, which were subsequently
sanitized.

## Checklist

- [X] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [X] I have verified that all data streams collect metrics or logs.
- [X] I have added an entry to my package's `changelog.yml` file.
- [X] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [X] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

### To test the m365_defender package:

- Clone integrations repo.
- Install elastic package locally.
- Start elastic stack using elastic-package.
- Move to integrations/packages/m365_defender directory.
- Run the following command to run tests.

> elastic-package test
```
--- Test results for package: m365_defender - START ---
╭───────────────┬───────────────┬───────────┬────────────────────────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE       │ DATA STREAM   │ TEST TYPE │ TEST NAME                                                                  │ RESULT │ TIME ELAPSED │
├───────────────┼───────────────┼───────────┼────────────────────────────────────────────────────────────────────────────┼────────┼──────────────┤
│ m365_defender │               │ asset     │ dashboard m365_defender-2690a440-7235-11ed-8657-c59f6ece834c is loaded     │ PASS   │       3.33µs │
│ m365_defender │               │ asset     │ dashboard m365_defender-3caf3c00-7456-11ed-8657-c59f6ece834c is loaded     │ PASS   │        317ns │
│ m365_defender │               │ asset     │ dashboard m365_defender-ac54d310-44ab-11ed-8375-0168a9970c06 is loaded     │ PASS   │        213ns │
│ m365_defender │               │ asset     │ dashboard m365_defender-afb93ff7-9903-4d91-9028-9fe9c5a434f8 is loaded     │ PASS   │        272ns │
[...]

PR #13830, opened 2025-05-07
Add new AWS Config datastream.

muskan-agarwal26:
## Proposed commit message

This release includes the Config data stream for supporting config findings via REST API.

AWS Config fields are mapped to their corresponding ECS fields where possible. Also added associated dashboards and visualizations.

Test samples were derived from live logs.

## Checklist

- Clone integrations repo.
- Move to integrations/packages/aws directory.

> elastic-package test
```
2025/05/07 21:50:34  INFO New version is available - v0.111.0. Download from: https://github.com/elastic/elastic-package/releases/tag/v0.111.0
2025/05/07 21:50:35  INFO License text found in "/root/ELASTIC_PROJECT/github/integrations/LICENSE.txt" will be included in package
2025/05/07 21:51:45  INFO Write container logs to file: /root/ELASTIC_PROJECT/github/integrations/build/container-logs/config-1746634905423590657.log
2025/05/07 21:51:50  INFO Write container logs to file: /root/ELASTIC_PROJECT/github/integrations/build/container-logs/elastic-agent-1746634910377459951.log
--- Test results for package: aws - START ---
╭─────────┬─────────────┬───────────┬───────────┬────────┬───────────────╮
│ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT │  TIME ELAPSED │
├─────────┼─────────────┼───────────┼───────────┼────────┼───────────────┤
│ aws     │ config      │ system    │ default   │ PASS   │ 55.030876912s │
╰─────────┴─────────────┴───────────┴───────────┴────────┴───────────────╯
--- Test results for package: aws - END   ---
Done
``` 

## Related issues
- Closes https://github.com/elastic/integrations/issues/10272

## Screenshots
config-1
config-2


packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,268 @@
...
+        "batch_size": state.batch_size,
+        "has_next": state.has_next,
+        "next": state.next,
+        "next_page": state.next_page,
+
+        // performing a succession of keyed hash operations (HMAC) on the request date, Region, and service, 
+        // with your AWS secret access key as the key for the initial hashing operation.
+        // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#derive-signing-key
+
+        "signing_key": "aws4_request".hmac("sha256", "config".hmac("sha256", state.aws_region.hmac("sha256", now.format("20060102").hmac("sha256", bytes("AWS4" + state.secret_key))))),
```
efd6, line 38:
```suggestion
        "signing_key": "aws4_request".hmac("sha256",
          "config".hmac("sha256",
            state.aws_region.hmac("sha256",
              now.format("20060102").hmac("sha256",
                bytes("AWS4" + state.secret_key)
              )
            )
          )
        ),
```

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,268 @@
...
+
+        // Creating a string_to_sign that includes the algorithm, request timestamp, credential scope, and hashed canonical request
+        // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#create-string-to-sign
+
+        "string_to_sign": ['AWS4-HMAC-SHA256', now.format("20060102T150405Z"), now.format("20060102") + '/' + state.aws_region + '/config/aws4_request', (['POST', '/', '', ["content-type:application/x-amz-json-1.1", "host:config." + state.aws_region + "." + state.tld, "x-amz-date:" + now.format("20060102T150405Z"), "x-amz-target:StarlingDoveService.GetComplianceDetailsByConfigRule"].join("\n"), '', "content-type;host;x-amz-date;x-amz-target", has(state.next_page) && has(state.next_page.result_token) ? {"NextToken": state.next_page.result_token, "ConfigRuleName": state.worklist.ConfigRules[int(state.next)].ConfigRuleName, "Limit": int(state.batch_size)}.encode_json().sha256().hex() : {"ConfigRuleName": state.worklist
+        .ConfigRules[int(state.next)].ConfigRuleName, "Limit": int(state.batch_size)}.encode_json().sha256().hex()].join("\n")).sha256().hex()].join("\n"),
```
efd6, line 44:
```suggestion
        "string_to_sign": [
          'AWS4-HMAC-SHA256',
          now.format("20060102T150405Z"),
          now.format("20060102") + '/' + state.aws_region + '/config/aws4_request',
          (
            [
              'POST',
              '/',
              '',
              [
                "content-type:application/x-amz-json-1.1",
                "host:config." + state.aws_region + "." + state.tld,
                "x-amz-date:" + now.format("20060102T150405Z"),
                "x-amz-target:StarlingDoveService.GetComplianceDetailsByConfigRule"
              ].join("\n"),
              '',
              "content-type;host;x-amz-date;x-amz-target",
              {
                ?"NextToken": state.?next_page.result_token,
                "ConfigRuleName": state.worklist.ConfigRules[int(state.next)].ConfigRuleName,
                "Limit": int(state.batch_size)
              }.encode_json().sha256().hex()
            ].join("\n")
          ).sha256().hex()
        ].join("\n"),
```
but it also seems to me that the double join is not necessary — untested.
```suggestion
        "string_to_sign": [
          'AWS4-HMAC-SHA256',
          now.format("20060102T150405Z"),
          now.format("20060102") + '/' + state.aws_region + '/config/aws4_request',
          (
            [
              'POST',
              '/',
              '',
              "content-type:application/x-amz-json-1.1",
              "host:config." + state.aws_region + "." + state.tld,
              "x-amz-date:" + now.format("20060102T150405Z"),
              "x-amz-target:StarlingDoveService.GetComplianceDetailsByConfigRule",
              '',
              "content-type;host;x-amz-date;x-amz-target",
              {
                ?"NextToken": state.?next_page.result_token,
                "ConfigRuleName": state.worklist.ConfigRules[int(state.next)].ConfigRuleName,
                "Limit": int(state.batch_size)
              }.encode_json().sha256().hex()
            ].join("\n")
          ).sha256().hex()
        ].join("\n"),
```
(also equivalents for the blocks below: prefer vertical space over horizontal space and separation over density, particularly for code like this; the optimisations that I have shown here were not obvious in the dense form)

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,268 @@
...
+          }.encode_json()
+        ).with(
+          {
+            "Header": {
+              "Content-Type": ["application/x-amz-json-1.1"],
+              "X-Amz-Date": [now.format("20060102T150405Z")],
+
+              // Performing a keyed hash operation on the string to sign using the derived signing key as the hash key to calculate the signature
+              // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#calculate-signature
+              // and constructing the Authorization header by combining the algorithm, credential scope, signed headers, and calculated signature.
```
efd6, line 96: This sentence no verb.
```suggestion
              // Perform a keyed hash operation on the string to sign using the derived signing key as the hash key to calculate the signature
              // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#calculate-signature
              // and construct the Authorization header by combining the algorithm, credential scope, signed headers, and calculated signature.
```

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,268 @@
...
+        state.with({
+
+          // performing a succession of keyed hash operations (HMAC) on the request date, Region, and service, 
+          // with your AWS secret access key as the key for the initial hashing operation.
+          // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#derive-signing-key
+
+          "signing_key": "aws4_request".hmac("sha256", "config".hmac("sha256", state.aws_region.hmac("sha256", now.format("20060102").hmac("sha256", bytes("AWS4" + state.secret_key))))),
+
+          // Creating a string_to_sign that includes the algorithm, request timestamp, credential scope, and hashed canonical request
+          // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#create-string-to-sign
```
efd6, line 77:
```suggestion
          // Create a string_to_sign that includes the algorithm, request timestamp, credential scope, and hashed canonical request.
          // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#create-string-to-sign
```

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,268 @@
...
+
+          "string_to_sign": ['AWS4-HMAC-SHA256', now.format("20060102T150405Z"), now.format("20060102") + '/' + state.aws_region + '/config/aws4_request', (['POST', '/', '', ["content-type:application/x-amz-json-1.1", "host:config." + state.aws_region + "." + state.tld, "x-amz-date:" + now.format("20060102T150405Z"), "x-amz-target:StarlingDoveService.DescribeConfigRules"].join("\n"), '', "content-type;host;x-amz-date;x-amz-target", has(state.next_page) && has(state.next_page.rule_token) ? {"NextToken": state.next_page.rule_token}.encode_json().sha256().hex() : {}.encode_json().sha256().hex()].join("\n")).sha256().hex()].join("\n")
+        })
+      :
+        state.with({
+
+          // performing a succession of keyed hash operations (HMAC) on the request date, Region, and service, 
+          // with your AWS secret access key as the key for the initial hashing operation.
+          // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#derive-signing-key
```
efd6, line 72:
```suggestion
          // Performing a succession of keyed hash operations (HMAC) on the request date, Region, and service, 
          // with the AWS secret access key as the key for the initial hashing operation.
          // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#derive-signing-key
```
Though this is a repeat of the text above, to I would suggest just referring to that.

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,268 @@
...
+            {
+              "worklist": body,
+              "next": 0,
+              "has_next": has(body.NextToken),
+              "batch_size": state.batch_size,
+              "next_page": {
+                ?"rule_token": body.?NextToken,
+              },
+
+              // performing a succession of keyed hash operations (HMAC) on the request date, Region, and service,
```
efd6, line 115: Refer to the text above.

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,268 @@
...
+          "ConfigRuleName": config_rules.worklist.ConfigRules[int(config_rules.next)].ConfigRuleName,
+          "Limit": int(config_rules.batch_size),
+        }.encode_json()
+      ).with(
+        {
+          "Header": {
+            "Content-Type": ["application/x-amz-json-1.1"],
+            "X-Amz-Date": [now.format("20060102T150405Z")],
+
+              // Performing a keyed hash operation on the string to sign using the derived signing key as the hash key to calculate the signature
```
efd6, line 175: Refer to the text above.

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,268 @@
...
+              // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#calculate-signature
+              // and constructing the Authorization header by combining the algorithm, credential scope, signed headers, and calculated signature.
+              // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#add-signature-to-request
+
+              "Authorization": ["AWS4-HMAC-SHA256 Credential=" + config_rules.access_key + "/" + now.format("20060102") + "/" + config_rules.aws_region + "/config/aws4_request, SignedHeaders=content-type;host;x-amz-date;x-amz-target, Signature=" + (config_rules.string_to_sign).hmac("sha256", config_rules.signing_key).hex()],
```
efd6, line 180: Split this over lines.

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,268 @@
...
+            "Header": {
+              "Content-Type": ["application/x-amz-json-1.1"],
+              "X-Amz-Date": [now.format("20060102T150405Z")],
+
+              // Performing a keyed hash operation on the string to sign using the derived signing key as the hash key to calculate the signature
+              // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#calculate-signature
+              // and constructing the Authorization header by combining the algorithm, credential scope, signed headers, and calculated signature.
+              // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#add-signature-to-request
+
+              "Authorization": ["AWS4-HMAC-SHA256 Credential=" + state.access_key + "/" + now.format("20060102") + "/" + state.aws_region + "/config/aws4_request, SignedHeaders=content-type;host;x-amz-date;x-amz-target, Signature=" + (state.string_to_sign).hmac("sha256", state.signing_key).hex()],
```
efd6, line 99: Split this over lines.

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,268 @@
...
+    :
+      (
+      state.?want_more.orValue(false) ?
+        state.with({
+
+          // performing a succession of keyed hash operations (HMAC) on the request date, Region, and service, 
+          // with your AWS secret access key as the key for the initial hashing operation.
+          // Refer https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#derive-signing-key
+
+          "signing_key": "aws4_request".hmac("sha256", "config".hmac("sha256", state.aws_region.hmac("sha256", now.format("20060102").hmac("sha256", bytes("AWS4" + state.secret_key))))),
```
efd6, line 60: As above.

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,268 @@
...
+              )
+            :
+              [{}],
+            "next_page": {
+              ?"result_token": body.?NextToken,
+              ?"rule_token": config_rules.?next_page.rule_token,
+            },
+            "want_more": config_rules.has_next || has(body.NextToken) || (int(config_rules.next) + 1 < size(config_rules.worklist.ConfigRules)),
+            "next": has(body.NextToken) ? config_rules.next : (int(config_rules.next) + 1 < size(config_rules.worklist.ConfigRules)) ? (int(config_rules.next) + 1) : 0,
+            "worklist": has(body.NextToken) ? config_rules.worklist : (int(config_rules.next) + 1 < size(config_rules.worklist.ConfigRules)) ? config_rules.worklist : {},
```
efd6, line 206: Vertical space.

packages/aws/data_stream/config/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,422 @@
...
+      tag: set_event_outcome_from_aws_config_compliance_type
+      lang: painless
+      description: set event.outcome from aws.config.compliance_type
+      if : ctx.aws?.config?.compliance_type instanceof String
+      source: >-
+        if (ctx.aws.config.compliance_type == 'NON_COMPLIANT') {
+          ctx.event.outcome = 'failure';
+        } else if (ctx.aws.config.compliance_type == 'COMPLIANT') {
+          ctx.event.outcome = 'success';
+        }else {
```
efd6, line 351:
```suggestion
        } else {
```

packages/aws/data_stream/config/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,422 @@
...
+            }
+            return v == null || v == '' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0)
+          });
+        }
+        handleMap(ctx);
+  - set:
+      field: event.kind
+      tag: set_pipeline_error_into_event_kind
+      value: pipeline_error
+      if: ctx.error?.message != null
```
efd6, line 409: Also need the original event recovery in this case.
```suggestion
  - set:
      field: event.kind
      tag: set_pipeline_error_into_event_kind
      value: pipeline_error
      if: ctx.error?.message != null
  - append:
      field: tags
      value: preserve_original_event
      allow_duplicates: false
      if: ctx.error?.message != null
```

efd6:
Also, need to add owner.
```
Error: package "packages/aws" shares ownership across data streams but these ones [packages/aws/data_stream/config] lack owners
```

packages/aws/elasticsearch/transform/latest_config_misconfigurations/transform.yml
```diff
@@ -0,0 +1,38 @@
+# Use of "*" to use all namespaces defined.
+source:
+  index:
+    - "logs-aws.config-*"
+dest:
+  index: "security_solution-aws.misconfiguration_latest-v3"
```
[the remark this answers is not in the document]
kcreddy: It could be `v1`. Probably copied from AWS Security Hub

packages/aws/data_stream/config/_dev/test/pipeline/test-common-config.yml
```diff
@@ -0,0 +1,5 @@
+fields:
+  tags:
+    - preserve_duplicate_custom_fields
+dynamic_fields:
+  "@timestamp": ".*"
```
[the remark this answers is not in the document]
muskan-agarwal26: We are mapping the @timestamp field with the `_ingest.timestamp` field, hence we need the above line of code to pass the pipeline tests.

efd6:
Please address https://github.com/elastic/integrations/pull/13830#discussion_r2088124296.

packages/aws/data_stream/config/_dev/test/pipeline/test-common-config.yml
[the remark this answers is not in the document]
muskan-agarwal26: Doing this, the pipeline test fails:
pipeline-test-fail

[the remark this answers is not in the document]
efd6:
```diff
diff --git a/packages/aws/data_stream/config/_dev/test/pipeline/test-common-config.yml b/packages/aws/data_stream/config/_dev/test/pipeline/test-common-config.yml
index df05cade2a..37e8fa225f 100644
--- a/packages/aws/data_stream/config/_dev/test/pipeline/test-common-config.yml
+++ b/packages/aws/data_stream/config/_dev/test/pipeline/test-common-config.yml
@@ -1,5 +1,3 @@
 fields:
   tags:
     - preserve_duplicate_custom_fields
-dynamic_fields:
-  "@timestamp": ".*"
diff --git a/packages/aws/data_stream/config/_dev/test/pipeline/test-event.log-expected.json b/packages/aws/data_stream/config/_dev/test/pipeline/test-event.log-expected.json
index d64117c97d..acec0338fa 100644
--- a/packages/aws/data_stream/config/_dev/test/pipeline/test-event.log-expected.json
+++ b/packages/aws/data_stream/config/_dev/test/pipeline/test-event.log-expected.json
@@ -1,7 +1,6 @@
 {
     "expected": [
         {
-            "@timestamp": "2025-05-07T13:18:18.562568394Z",
             "aws": {
                 "config": {
                     "annotation": "string",
diff --git a/packages/aws/data_stream/config/elasticsearch/ingest_pipeline/default.yml b/packages/aws/data_stream/config/elasticsearch/ingest_pipeline/default.yml
index 25b30dd7cb..1b9086099f 100644
--- a/packages/aws/data_stream/config/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/aws/data_stream/config/elasticsearch/ingest_pipeline/default.yml
@@ -250,10 +250,6 @@ processors:
         - append:
             field: error.message
             value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
-  - set:
-      tag: set_timestamp
-      field: "@timestamp"
-      value: "{{{_ingest.timestamp}}}"
   - rename:
       field: json.EvaluationResultIdentifier.EvaluationResultQualifier.ConfigRuleName
       tag: rename_EvaluationResultIdentifier_EvaluationResultQualifier_ConfigRuleName
```

packages/aws/data_stream/config/_dev/test/pipeline/test-event.log-expected.json
```diff
@@ -0,0 +1,117 @@
...
+                "user": [
+                    "string"
+                ]
+            },
+            "resource": {
+                "id": "i-0a4468fbfafee6a8f",
+                "type": "AWS::EC2::Instance"
+            },
+            "result": {
+                "evaluation": "COMPLIANT"
```
kcreddy, line 103: This can only be `passed`, `failed`, or `unknown` and always populated.

packages/aws/data_stream/config/sample_event.json
```diff
@@ -0,0 +1,108 @@
...
+        "type": "AWS::EC2::Instance"
+    },
+    "result": {
+        "evaluation": "COMPLIANT"
+    },
+    "rule": {
+        "description": "Checks whether the active access keys are rotated within the number of days specified in maxAccessKeyAge. The rule is non-compliant if the access keys have not been rotated for more than maxAccessKeyAge number of days.",
+        "name": "access-keys-rotated",
+        "reference": "arn:aws:config:us-east-1:11223344556:config-rule/config-rule-id1",
+        "uuid": "config-rule-id1"
```
kcreddy, line 100: I would suggest use `rule.id` similar to AWS SecurityHub and update the transform definition to `rule.id`

packages/aws/elasticsearch/transform/latest_config_misconfigurations/manifest.yml
```diff
@@ -0,0 +1,11 @@
+start: true
```
kcreddy, line 1: Remove this file.
Please replicate what we have under SecurityHub. https://github.com/elastic/integrations/tree/main/packages/aws/elasticsearch/transform/latest_cdr_misconfigurations

packages/aws/elasticsearch/transform/latest_config_misconfigurations/transform.yml
```diff
@@ -0,0 +1,38 @@
...
+  unattended: true
+sync:
+  time:
+    field: "event.ingested"
+    # sometimes transform wouldn't process all documents.
+    delay: 60s
+retention_policy:
+  time:
+    field: "@timestamp"
+    max_age: 2160h
```
kcreddy, line 32:
```suggestion
    max_age: 24h
```
Similar to SecurityHub

efd6 (commented):
LGTM but waiting for @kcreddy

packages/aws/elasticsearch/transform/latest_config_misconfigurations/transform.yml
```diff
@@ -0,0 +1,38 @@
...
+_meta:
+  managed: true
+  # Bump this version to delete, reinstall, and restart the transform during
+  # package installation.
+  fleet_transform_version: 0.1.0
+  run_as_kibana_system: false
```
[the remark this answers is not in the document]
kcreddy: @maxcold, @nick-alayil 
Adding onto https://github.com/elastic/integrations/pull/13830#discussion_r2090375397, this PR adds the transform that takes data from source `logs-aws.config-*` and populates destination index `security_solution-aws.misconfiguration_latest`. The destination index is same as AWS SecurityHub.

Now, as per the CDR guide, the transform's `run_as_kibana_system` property is not set. The [default value](https://github.com/elastic/kibana/blob/264194ea4e46dc260e36c807c166777865ea5a79/x-pack/platform/plugins/shared/fleet/server/services/epm/elasticsearch/transform/install.ts#L256-L258) for this property is when it is not set is `true`. This means the transform is created using `kibana_system` role. But if a transform needs to be created with `kibana_system` role, it requires [additional ES permissions for source indices](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/store/KibanaOwnedReservedRoleDescriptors.java#L474-L486). Currently the ES permissions doesn't include `logs-aws.config-*` source indices hence the transform is failing with following error: 
```
Transform encountered an exception: [org.elasticsearch.ElasticsearchSecurityException: Cannot start transform [logs-aws.latest_config_misconfigurations-default-0.1.0] because user lacks required permissions, see privileges_check_failed issue for more details]; Will automatically retry [1/-1]
```

We may have 2 options:
1. Set `run_as_kibana_system: false` which creates/runs the transform using [secondary credentials](https://github.com/elastic/kibana/issues/151186#issuecomment-1507074519) (not `kibana_system` role). This is in deviation to existing 3rd party CDR integrations which all use `kibana_system` role.
2. Add necessary ES permissions to source indices `logs-aws.config-*` and wait for next release of ES to take its effect and then add the transform. Until then, we could just add the datastream without a transform.

WDYT is preferred here?
cc: @muskan-agarwal26

packages/aws/data_stream/config/fields/ecs.yml
```diff
@@ -0,0 +1,4 @@
+- name: cloud.provider
+  type: constant_keyword
+- name: observer.vendor
+  type: constant_keyword
```
kcreddy, line 4: If its defined this way, the fields inside README is rendered without description.
```
| cloud.provider |  | constant_keyword |
| observer.vendor |  | constant_keyword |
```
Please update to:
```suggestion
- name: cloud.provider
  type: constant_keyword
  external: ecs
- name: observer.vendor
  type: constant_keyword
  external: ecs
```

packages/aws/data_stream/config/sample_event.json
```diff
@@ -0,0 +1,108 @@
...
+    "input": {
+        "type": "cel"
+    },
+    "message": "Checks whether the active access keys are rotated within the number of days specified in maxAccessKeyAge. The rule is non-compliant if the access keys have not been rotated for more than maxAccessKeyAge number of days.",
+    "observer": {
+        "product": "AWS Config",
+        "vendor": "Amazon"
+    },
+    "resource": {
+        "id": "arn:aws:config:us-east-1:11223344556:config-rule/config-rule-id1",
```
kcreddy, line 90: This is incorrect. It should be populated by `resource_id`. 
What I meant in https://github.com/elastic/integrations/pull/13830#discussion_r2090395561 was that if you can derive the full ARN of the `resource_id`. If thats not possible (either account ID or cloud region is not known), then let it just be `resource_id` like before.

packages/aws/elasticsearch/transform/latest_config_misconfigurations/transform.yml
```diff
@@ -0,0 +1,36 @@
...
+  unique_key:
+    - rule.id
+    - resource.id
+    - data_stream.namespace
+  sort: "@timestamp"
+description: >-
+  Latest Configs from AWS. As configs get updated, this transform stores only the latest state of each config inside the destination index. Thus the transform's destination index contains only the latest state of the config.
+frequency: 5m
+settings:
+  # This is required to prevent the transform from clobbering the Fleet-managed mappings.
```
kcreddy, line 20: Remove this comment, its not associated with below code.

packages/aws/elasticsearch/transform/latest_config_misconfigurations/transform.yml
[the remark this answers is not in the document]
kcreddy: @maxcold, I also hope the transform naming here is fine. The transform naming convention is not defined inside CDR guide, but we have been following the name `latest_cdr_misconfigurations`. But it was already taken during [AWS SecurityHub](https://github.com/elastic/integrations/blob/main/packages/aws/elasticsearch/transform/latest_cdr_misconfigurations/transform.yml) implementation.

kcreddy (commented):
Waiting for Cloud Security team's inputs on 3 outstanding comments:
1. Main blocker (Elasticsearch permissions): https://github.com/elastic/integrations/pull/13830#discussion_r2090401706
2. `observer.*` mappings: https://github.com/elastic/integrations/pull/13830#discussion_r2090390113
3. ILM policy for source indices: https://github.com/elastic/integrations/pull/13830#discussion_r2090375397

packages/aws/data_stream/config/sample_event.json
```diff
@@ -0,0 +1,108 @@
...
+        ]
+    },
+    "input": {
+        "type": "cel"
+    },
+    "message": "Checks whether the active access keys are rotated within the number of days specified in maxAccessKeyAge. The rule is non-compliant if the access keys have not been rotated for more than maxAccessKeyAge number of days.",
+    "observer": {
+        "product": "AWS Config",
+        "vendor": "Amazon"
+    },
```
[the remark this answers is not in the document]
maxcold: It is a good question. When we started rely on `observer.vendor` this question came up I believe but we decided to put the detailed name into it because `AWS` or `Amazon` wasn't specific enough and in the data grid we didn't want to rely on the combination two fields. Though it's not 100% semanitcally correct, I'd advice going with `AWS Config` as a value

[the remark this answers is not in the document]
kcreddy:
```suggestion
    "observer": {
        "vendor": "AWS Config"
    },
```
Thanks @maxcold. 
@muskan-agarwal26, please update as per above suggestion.

packages/aws/data_stream/config/elasticsearch/ilm/default_policy.json
```diff
@@ -0,0 +1,20 @@
...
+					}
+				}
+			},
+			"delete": {
+				"min_age": "7d",
+				"actions": {
+					"delete": {}
+				}
+			}
+		}
```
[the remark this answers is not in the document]
maxcold: I think it definitely makes sense to test it before merging
On the ILM policy - if we ingest the full posture every 24h, then 7d retention for the source index is fine, we don't need old data in the source index.
For vulnerability data stream though it might be a different case as we have 90d retention on transform, I'm not sure how it will work with 7d on ILM, need to think about it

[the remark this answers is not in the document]
kcreddy: > On the ILM policy - if we ingest the full posture every 24h, then 7d retention for the source index is fine, we don't need old data in the source index.

Thanks for the confirmation.

Regarding `vulnerability` data_stream in https://github.com/elastic/integrations/pull/13595/files#r2079624821, what we have is a full ingestion (not incremental) every `interval: 4h`. So my understanding is that a similar 7d ILM on source data is going to be fine. 
The CDR guides mentions 90d retention only incase the ingestion was incremental.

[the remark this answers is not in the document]
kcreddy: > I think it definitely makes sense to test it before merging

@muskan-agarwal26 @piyush-elastic, our cloud security team would like to test this feature before merging the PR. Once their testing is done, we can merge them.

packages/aws/elasticsearch/transform/latest_config_misconfigurations/transform.yml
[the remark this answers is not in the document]
kcreddy: Thanks @maxcold for the detailed clarifications.
> 1. what about keeping latest_cdr_misconfigurations part but adding some suffix, eg. latest_cdr_misconfigurations_awsconfig or smth?

@muskan-agarwal26, could you please update the transform name to `latest_cdr_misconfigurations_awsconfig` as suggested?

> 2. Plus I'm not sure how the case of versioning in the destination index and aliasing would work out. What about having a different destination index but keeping the pattern security_solution-*.misconfiguration_latest we use on the data view. For example security_solution-awsconfig.misconfiguration_latest would fit the pattern and allow us to use a different destination index compare to AWS Security Hub

Thats a great point. I think currently defined destination index and alias will effect AWS SecurityHub working. We should definitely update as per your suggestion. During dual testing we may come to know of more issues if any.
For now, @muskan-agarwal26 can you please update transform's destination configuration to following:
```
dest:
  index: "security_solution-awsconfig.misconfiguration_latest-v1"
  aliases:
    - alias: "security_solution-awsconfig.misconfiguration_latest"
      move_on_creation: true
```

> 3. With this in mind I think we should be consistent with other data streams and use kibana_system. We can add the required source index permissions for 9.1 release (and backport to 8.x supported versions as well) and update the version check accordingly

@maxcold Alright, will create the required PR and ping you for checking on backport versions.

kcreddy (commented):
@muskan-agarwal26 please implement suggestions from https://github.com/elastic/integrations/pull/13830#discussion_r2103942615

packages/aws/elasticsearch/transform/latest_config_misconfigurations/transform.yml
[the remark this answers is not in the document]
kcreddy: @maxcold, here's the Elasticsearch PR: https://github.com/elastic/elasticsearch/pull/128350. Please kindly check the backport labels. Its beyond FF for `v8.17.7, v8.18.2, and v9.0.2`, but looks like https://github.com/elastic/elasticsearch/pull/124074 was given an exception so I added them.

packages/aws/data_stream/config/elasticsearch/ilm/default_policy.json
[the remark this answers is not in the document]
maxcold: @kcreddy will let you know regarding AWS Config testing

As for the retention - if transform have `retention_policy.time.max_age: 24h` then having retention period of 7d on source index should be ok. I just don't see transform on m365 PR at all, that's why wasn't sure what's the approach there.

maxcold (changes requested):
I built the integration from the PR, installed it on a cloud env 9.0.1. Here are the issues I found, we need to fix them before merging:
1. `resource.name`, `resource.type`, `rule.number` are missing in the data, that's only what I can see in the data grid. 
2. the flyout opens but has no data, most likely some of the data missing but I need to dig deeper to understand what exactly

@kcreddy @muskan-agarwal26 before implementing the changes for CDR flows we were doing spreadsheet mapping of the fields required by the guide https://docs.elastic.dev/security-solution/cloud-security/cdr/3p-dev-guide . This allows to understand what crucial fields are missing or the data format is different. Has this exercise been done for AWS Config?

Screenshot 2025-05-26 at 18 05 24

kcreddy (commented):
After an internal discussion, we decided not to use transform for now. This will remove the issue of https://github.com/elastic/integrations/pull/13830#pullrequestreview-2868842851.
@muskan-agarwal26 please remove the transform (and any related changes) from the PR.

packages/aws/data_stream/config/sample_event.json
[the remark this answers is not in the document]
kcreddy: Thanks!

packages/aws/manifest.yml
```diff
@@ -172,6 +172,35 @@ policy_templates:
...
+        enabled: true
+      agentless:
+        enabled: true
+        organization: security
+        division: engineering
+        team: security-service-integrations
+    data_streams:
+      - config
+    categories:
+      - security
```
kcreddy, line 189: Add `- cloudsecurity_cdr`

packages/aws/docs/config.md
```diff
@@ -0,0 +1,241 @@
...
+
+**IMPORTANT: Extra AWS charges on API requests will be generated by this integration. Check [API Requests](https://www.elastic.co/docs/current/integrations/aws#api-requests) for more details.**
+
+## Data streams
+
+The AWS Config integration collects one type of data: logs.
+
+**Logs** help you keep a record of the findings in AWS Config, allowing you to track and audit compliance status of your resources.
+
+Within the `config` data stream, we first retrieve all config rules using the [DescribeConfigRules](https://docs.aws.amazon.com/config/latest/APIReference/API_DescribeConfigRules.html) action. Then, for each specific config rule, we fetch its evaluation results using the [GetComplianceDetailsByConfigRule](https://docs.aws.amazon.com/config/latest/APIReference/API_GetComplianceDetailsByConfigRule.html) action. These evaluation results enrich their respective config rules, ultimately producing a finding log.
```
kcreddy, line 15:
```suggestion
The AWS Config integration works by first retrieving all config rules using the [DescribeConfigRules](https://docs.aws.amazon.com/config/latest/APIReference/API_DescribeConfigRules.html) API. Then, for each specific config rule, the integration fetches its evaluation results using the [GetComplianceDetailsByConfigRule](https://docs.aws.amazon.com/config/latest/APIReference/API_GetComplianceDetailsByConfigRule.html) API. These evaluation results enrich their respective config rules, ultimately producing a finding log.
```

kcreddy (commented):
Nits

kcreddy (approved):
LGTM. Thanks!

maxcold (approved):

PR #13832, opened 2025-05-07
[CI] Do not use -q parameter together with git commands in pipe

mrodm:
## Proposed commit message

Remove the usage of the `-q` parameter when used as part of pipe with `git` commands.

It could report failure due to SIGPIPE errors. As it was, this code
```
if git diff --name-only "${commit_merge}" "${to}" | grep -q -E "^packages/${package}/" ; then
    echo "[${package}] PR is affected: found package files"
    return 0
fi
```

was returning failure (141 error code) even there were matching files with the given regex.

Related https://stackoverflow.com/questions/19120263/why-exit-code-141-with-grep-q

Buildkite build with tests:
- https://buildkite.com/elastic/integrations/builds/25683#0196ac79-4deb-448d-8609-c34b9a8a44d6

```bash
# get the changeset from the affected PR
git fetch upstream pull/13810/head:pr-13810

# try to check the modified files
# this should return zero code
(set -o pipefail ; git diff --name-only c08c07e2cd3268914d9c504d0a51273f33f63f4a befe8ead69811789c40cba2e907abf6d0bcd6046  | grep -q -E "^packages/security_detection_engine/")

# Without -q parameter, it works
(set -o pipefail ; git diff --name-only c08c07e2cd3268914d9c504d0a51273f33f63f4a befe8ead69811789c40cba2e907abf6d0bcd6046  | grep -E "^packages/security_detection_engine/" > /dev/null)

```

## Related issues

- Relates https://github.com/elastic/integrations/pull/13810
- Relates https://github.com/elastic/integrations/pull/13831/

### DIFF
--- a/.buildkite/scripts/common.sh
+++ b/.buildkite/scripts/common.sh
@@ -768,12 +768,18 @@ is_pr_affected() {
 
     commit_merge=$(git merge-base "${from}" "${to}")
     echoerr "[${package}] git-diff: check non-package files (${commit_merge}..${to})"
-    if git diff --name-only "${commit_merge}" "${to}" | grep -q -E -v '^(packages/|\.github/(CODEOWNERS|ISSUE_TEMPLATE|PULL_REQUEST_TEMPLATE)|README\.md|docs/)' ; then
+    # Avoid using "-q" in grep in this pipe, it could cause that some files updated are not detected due to SIGPIPE errors
+    # https://buildkite.com/elastic/integrations/builds/25606
+    # https://github.com/elastic/integrations/pull/13810/files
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/common.sh
jsoriano, line 1: Should we handle the issue in the greps in lines 304, 410 and 441 just in case? Though piped commands there don't seem to be affected.

jsoriano (approved):

.buildkite/scripts/common.sh
mrodm: I'll add the same change in those lines to avoid similar errors in the future just in case.

mrodm: Done in https://github.com/elastic/integrations/pull/13832/commits/be936dc2fc4ebfbd45546280c120a284f3184df6

jsoriano (approved):
Thanks :+1:

mrodm:
@Mergify backport backport-cloud_security_posture-1.13 backport-security_detection_engine-8.18 backport-security_detection_engine-8.17 backport-security_detection_engine-8.16

PR #13837, opened 2025-05-08
[slack-audit]: fix ingest where source address is empty in event

fabiogermann:
Added a check to only process the IP address if it is present in the event.
In some cases Slack logs events that do not have an IP which causes the pipeline to fail.

## Checklist

## Author's Checklist

n/a

Use any Slack event and remove the source.address field and test using the the regular pipeline tester.

## Related issues

n/a

## Screenshots

n/a

### DIFF
--- a/packages/slack/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/slack/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
@@ -202,6 +202,7 @@ processors:
       target_field: source.ip
       type: ip
       ignore_missing: true
+      if: ctx?.source?.address != ''
   - geoip:
       field: source.ip
       target_field: source.geo


packages/slack/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -202,6 +202,7 @@ processors:
       target_field: source.ip
       type: ip
       ignore_missing: true
+      if: ctx?.source?.address != ''
```
kcreddy, line 205:
```suggestion
      if: ctx.source?.address != ''
```

packages/slack/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.25.1"
```
kcreddy, line 2: Please update [`manifest.yml`](https://github.com/elastic/integrations/blob/main/packages/slack/manifest.yml#L4) as well.

kcreddy (changes requested):

PR #13838, opened 2025-05-08
[atlassian-confluence]: fix ingest pipeline handling case where event does not have ip address

fabiogermann:
Added a check to only process the IP address if it is present in the event.
In some cases Confluence sends events that do not have an IP / or an empty sting as IP which causes the pipeline to fail.

## Checklist

## Author's Checklist

n/a

Use any Slack event and remove the source.address field and test using the the regular pipeline tester.

## Related issues

n/a

## Screenshots

n/a

### DIFF
--- a/packages/atlassian_confluence/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/atlassian_confluence/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
@@ -50,7 +50,7 @@ processors:
       target_field: source.ip
       type: ip
       ignore_missing: true
-      if: ctx?.source?.address != ''
+      if: ctx.source?.address != ''
   - geoip:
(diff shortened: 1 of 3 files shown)


packages/atlassian_confluence/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -50,6 +50,7 @@ processors:
       target_field: source.ip
       type: ip
       ignore_missing: true
+      if: ctx?.source?.address != ''
```
efd6, line 53:
```suggestion
      if: ctx.source?.address != ''
```

efd6 (commented):
Please add a pipeline test and update the manifest.yml's version field to match the version in the changelog.

PR #13862, opened 2025-05-09
[Crowdstrike Alert] adjust batch size to API limit

SimonKoetting:
The CrowdStrike alert integrations uses 2 API endpoints from CrowdStrike:
/alerts/queries/alerts/v2 → to get the composite IDs of all open alerts → limit of 10,000 (see https://www.falconpy.io/Service-Collections/Alerts.html#getqueriesalertsv2) as It's currently also defined as the default value for the batch size in the description.
The returned composite_ids of the first API call are then sent with a post again:
/alerts/entities/alerts/v2 → to get the alert details → limit of 1,000 (see https://www.falconpy.io/Service-Collections/Alerts.html#postentitiesalertsv2) 
which causes the integration to get an HTTP 413 request too large error if there are more than 1000 composite IDs returned from the first API which returns up to 10,000 IDs due to the limit.

As we already implemented a pagination using the want_more for the get, the batch size should be set to a maximum 10 1,000 instead of 10,000 to avoid the integration failing in environments with a lot of alerts.

### DIFF
--- a/packages/crowdstrike/changelog.yml
+++ b/packages/crowdstrike/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.65.1"
+  changes:
+    - description: Adjust alert batch size to 1000 to match the API limit.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/13857
 - version: "1.65.0"
   changes:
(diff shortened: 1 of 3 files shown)


efd6 (approved):
Thanks, that's a nice foot-gun.

PR #13885, opened 2025-05-12
[Cloudflare Logpush] map cloudflare_logpush.http_request.client.ssl.cipher to tls.cipher

### LINKED ISSUE #13596: [Cloudflare Logpush]: The field cloudflare_logpush.http_request.client.ssl.cipher has the wrong mapping
### Integration Name

Cloudflare Logpush [cloudflare_logpush]

### Dataset Name

_No response_

### Integration Version

1.37.2

### Agent Version

8.17.3

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.17.3

Elastic Cloud

### Software/API Version

_No response_

### Error Message

There is no error, the field just have the wrong mapping

### Event Original

Not applicable

### What did you do?

Not applicable

### What did you see?

The field `cloudflare_logpush.http_request.client.ssl.cipher` is mapped as a `text` field, making it not possible to use it in visualizations in dashboards.

The field `cloudflare_logpush.http_request.client.ssl.cipher` should be mapped as `keyword` field.

### Anything else?

The content of the field `cloudflare_logpush.http_request.client.ssl.cipher` is the tls cipher being used in the connection, this is a single string, the equivalent ecs field is `tls.cipher`, which is mapped as `keyword`.

The user normally wants to create visualizations using this field, but it is not possible because of the wrong mapping.

moxarth-rathod:
## Proposed commit message

```
cloudflare_logpush: map cloudflare_logpush.http_request.client.ssl.cipher to tls.cipher

Map cloudflare_logpush.http_request.client.ssl.cipher to the tls.cipher in order
to pick up its search behavior without risking potential breaking behavior.
```

## Checklist

Clone integrations repo.
Install elastic package locally.
Start elastic stack using elastic-package.
Move to integrations/packages/cloudflare_logpush directory.
> elastic-package test

## Related issues

- Closes #13596

### DIFF
--- a/packages/cloudflare_logpush/changelog.yml
+++ b/packages/cloudflare_logpush/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "1.38.0"
   changes:
-    - description: Add tls.cipher ECS mapping for http_request.client.ssl.cipher.
+    - description: Add `tls.cipher` ECS mapping for `cloudflare_logpush.http_request.client.ssl.cipher`.
       type: enhancement
       link: https://github.com/elastic/integrations/pull/13885
 - version: "1.37.4"
--- a/packages/cloudflare_logpush/changelog.yml
+++ b/packages/cloudflare_logpush/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Fix the field type of http_request.client.ssl.cipher from text to keyword.
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/1
+      link: https://github.com/elastic/integrations/pull/13885
 - version: "1.37.4"
   changes:
     - description: Fix the field type of http_request.bot.detection_tags from long to keyword, aligning it with the log source type of string array. This resolves an issue where the field was being ignored due to the "ignored_malformed" index setting, making it unusable in searches.
--- a/packages/cloudflare_logpush/changelog.yml
+++ b/packages/cloudflare_logpush/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.37.5"
+  changes:
+    - description: Fix the field type of http_request.client.ssl.cipher from text to keyword.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/1
 - version: "1.37.4"
   changes:
     - description: Fix the field type of http_request.bot.detection_tags from long to keyword, aligning it with the log source type of string array. This resolves an issue where the field was being ignored due to the "ignored_malformed" index setting, making it unusable in searches.
--- a/packages/cloudflare_logpush/changelog.yml
+++ b/packages/cloudflare_logpush/changelog.yml
@@ -1,8 +1,8 @@
 # newer versions go on top
-- version: "1.37.5"
+- version: "1.38.0"
   changes:
-    - description: Fix the field type of http_request.client.ssl.cipher from text to keyword.
-      type: bugfix
+    - description: Add tls.cipher ECS mapping for http_request.client.ssl.cipher.
+      type: enhancement
       link: https://github.com/elastic/integrations/pull/13885
 - version: "1.37.4"
   changes:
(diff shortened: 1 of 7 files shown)


packages/cloudflare_logpush/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.37.5"
+  changes:
+    - description: Fix the field type of http_request.client.ssl.cipher from text to keyword.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/1
```
efd6, line 6:
```suggestion
      link: https://github.com/elastic/integrations/pull/13885
```

packages/cloudflare_logpush/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.37.5"
+  changes:
+    - description: Fix the field type of http_request.client.ssl.cipher from text to keyword.
+      type: bugfix
```
efd6, line 5: I do not believe that this is a fix since I am not convinced that this is a bug. It is unfortunate, but it is not a strict conflict.

efd6 (commented):
I can see the justification for this, but I do not agree that this is a fix; at best it is an enhancement since it is enabling something that could not be done in the past, but it is not a conflict with the ECS since the types under custom field hierachies are not constrained by the ECS (arguably), and the field here is not mentioned in the ECS (ssl v tls, yes, tenuous, but valid).

I think this needs thought, and the commit message rationale needs to be clarified.

/cc @ShourieG @leandrojmp

leandrojmp:
> I can see the justification for this, but I do not agree that this is a fix; at best it is an enhancement since it is enabling something that could not be done in the past, but it is not a conflict with the ECS since the types under custom field hierachies are not constrained by the ECS (arguably), and the field here is not mentioned in the ECS (ssl v tls, yes, tenuous, but valid).
> 
> I think this needs thought, and the commit message rationale needs to be clarified.
> 
> /cc @ShourieG @leandrojmp

I think this can be considered an enhancement, I mentioned on the Issue that I've opened that the current mapping is wrong because the value for the field `cloudflare_logpush.http_request.client.ssl.cipher` is a keyword and the equivalent ECS field would be `tls.cipher`, but this is indeed not a conflict.

Some values for this field are:

- `AEAD-AES256-GCM-SHA384`
- `ECDHE-ECDSA-AES128-GCM-SHA256`
- `NONE`

Which is a string that indicates the cipher used during the connection, the same description for ecs `tls.cipher` field.

A future improvement could be just renaming this into `tls.cipher` to make it easier to correlate with other data sources, this is something that we are doing using the custom ingest pipeline.

efd6:
@leandrojmp 

> Which is a string that indicates the cipher used during the connection, the same description for ecs tls.cipher field.

Oh, I absolutely agree that it is intended to be semantically the same, but ECS legalism…

> A future improvement could be just renaming this into tls.cipher to make it easier to correlate with other data sources, this is something that we are doing using the custom ingest pipeline.

TBH, this would be my preferred solution — copying the value over to `tls.cipher` — but I was concerned that the semantics of **client** might be lost in the move to that field where **client** is not indicated in the field name. As a user, do you think that would an issue?

efd6:
@leandrojmp, thanks. Let's wait for @ShourieG, but @moxarth-rathod I think we may want to do that rather than what is here now.

moxarth-rathod:
> @leandrojmp, thanks. Let's wait for @ShourieG, but @moxarth-rathod I think we may want to do that rather than what is here now.

Got it @efd6. I'll update this as per the suggestions.

ShourieG:
@moxarth-rathod LGTM from my end, but wait for Dan's approval before merging.

packages/cloudflare_logpush/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.38.0"
+  changes:
+    - description: Add tls.cipher ECS mapping for http_request.client.ssl.cipher.
```
efd6, line 4:
```suggestion
    - description: Add `tls.cipher` ECS mapping for `cloudflare_logpush.http_request.client.ssl.cipher`.
```

efd6 (commented):
Please also update the proposed commit message to match the behaviour here.
```
cloudflare_logpush: map cloudflare_logpush.http_request.client.ssl.cipher to tls.cipher

Map cloudflare_logpush.http_request.client.ssl.cipher to the tls.cipher in order
to pick up its search behavior without risking potential breaking behavior.
```

efd6 (approved):

PR #13896, opened 2025-05-13
[Apache_Tomcat] Ingest pipeline pattern enhancement

Linu-Elias:
## Proposed commit message
Added support for %D attribute to get the time taken to process the request, in millis.
Existing Supported Pattern:
Common Log Format :- '%h %l %u %t "%r" %s %b'
Combined Log Format :- '%h %l %u %t "%r" %s %b "%{Referrer}i" "%{User-Agent}i"'
Combined Log Format + X-Forwarded-For header :- '%h %l %u %t "%r" %s %b %A %X %T "%{Referer}i" "%{User-Agent}i" X-Forwarded-For="%{X-Forwarded-For}i"'

New Pattern:
Common Log Format :- '%h %l %u %t "%r" %s %b **ms:%D**'
Combined Log Format :- '%h %l %u %t "%r" %s %b **ms:%D** "%{Referrer}i" "%{User-Agent}i"'
Combined Log Format + X-Forwarded-For header :- '%h %l %u %t "%r" %s %b **ms:%D** %A %X %T "%{Referer}i" "%{User-Agent}i" X-Forwarded-For="%{X-Forwarded-For}i"'

## Checklist

## Author's Checklist

- [ ]


ishleenk17:
@Linu-Elias : As per documentation %T also gives the same time just in seconds..
%D - Time taken to process the request, in millis
%T - Time taken to process the request, in seconds

So are both same  ? If yes, do we need both of them in the log format ?

Linu-Elias:
> > But, after closely looking at the sample logs in the ingest pipeline and the corresponding Grok pattern, it appears that we actually support **%F** (time taken to commit the response, in milliseconds), which is mapped to `apache_tomcat.access.response_time`
> 
> Can we confirm on this from the apache tomcat logs and if required do the correction in the current code. So from what I ubderstand you are mentiioning that we are groking correctly but the value corresponds to a field with different definition. Lets confirm that and make the change else it would be confusing. Also please see it should not be a breaking change.

I have made the necessary doc changes in this PR for- changed `%T` to` %F` since we are grokking for `apache_tomcat.access.response_time `

packages/apache_tomcat/data_stream/catalina/fields/beats.yml
```diff
@@ -22,3 +22,6 @@
     - name: vol
       type: keyword
       description: The serial number of the volume that contains a file. (Windows-only)
+- name: log.flags
```
ishleenk17, line 25: What is this added for ?

PR #10741, opened 2024-08-08
[zscaler_zia] Update response format version numbers

chrisberkhout:
## Proposed commit message

```
[zscaler_zia] Update response format version numbers (#)

Response formats were often updated without incrementing the
corresponding version number.

These have been updated to reflect the number of changes made
to the response formats, as summarized in the following table.

| Data stream      | Was | Now | Note                                |
|------------------|-----|-----|-------------------------------------|
| Alerts           |  v1 |  v1 | Correct.                            |
| Audit Log        |  v1 |  v1 | Correct. Pattern added in 78f3eae.  |
| DNS Log          |  v1 |  v2 | Fixed. Pattern updated in 78f3eae.  |
| Endpoint DLP Log |  v1 |  v1 | Correct. Pattern added in 78f3eae.  |
| Firewall Log     |  v1 |  v2 | Fixed. Pattern updated in 78f3eae.  |
| Tunnel Log       |  v1 |  v2 | Fixed. Patterns updated in 78f3eae. |
| Web Log          |  v2 |  v5 | Fixed. Pattern updated in...        |
|                  |     |     | 3418fe2, a9783c8, 78f3eae, dbf3f74. |
```

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).

PR #13904, opened 2025-05-14
zscaler_zia: add strict field template mode for tcp and http_endpoint input data streams

### LINKED ISSUE #10743: [zscale_zia] Validate response format versions
The `zscale_zia` integration requires the user to set up Zscaler Internet Access (ZIA) with response format definitions for each type of data requested by the integration.

If a user upgrades the integration without updating their response format configuration in Zscaler, there can be a mismatch between the data in responses and what's expected by the ingest pipelines.

We could avoid such problems by having each format definition include a hard-coded version number, and having the ingest pipelines validate that they have received data with the expected version number.

When introducing this, it would be good to allow responses with the correct version number, or with no version number but the expected a set of field names (which could be validated by sorting, hashing and comparing to an expected value). That would avoid having a breaking change to introduce the version validation.

efd6:
## Proposed commit message

```
zscaler_zia: add strict field template mode for tcp and http_endpoint input data streams

This checks that the template version matches the current expected
template version if the template specifies it, falling back to an
assessment of the fields that are present in the message. The check is
only performed when enabled, which is not the case by default.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #10743


packages/zscaler_zia/data_stream/web/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -37,6 +37,30 @@ processors:
...
+          version: v9
```
kcreddy, line 44: After this, is the idea to maintain all versions or only the latest version?

packages/zscaler_zia/data_stream/web/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -37,6 +37,30 @@ processors:
...
+          def fields = [];
+          for (e in ctx.json.entrySet()) {
+            fields.add(e.getKey());
+          }
+          Collections.sort(fields);
+          String signature = String.join("|", fields);
+          if (signature != params.expect.fields) {
+            ctx.error = ctx.error ?: [:];
+            ctx.error.message = ctx.error.message ?: [];
+            ctx.error.message.add("field set mismatch: "+signature+" is not expected set of templated fields");
```
kcreddy, line 57: Do you think adding a link to integration docs would better help users understand this error?

packages/zscaler_zia/data_stream/web/elasticsearch/ingest_pipeline/default.yml
efd6: Only the latest version, and after some reasonable time we will remove the field enumeration fallback (it _is_ a pain to get this correct in the script).

packages/zscaler_zia/data_stream/web/elasticsearch/ingest_pipeline/default.yml
efd6: I think this is an excellent idea.

packages/zscaler_zia/data_stream/web/elasticsearch/ingest_pipeline/default.yml
efd6: Initial change for that, but I need to be able to link directly to the version that is running, rather than the latest, and I don't see a way to do that right now.

packages/zscaler_zia/data_stream/web/elasticsearch/ingest_pipeline/default.yml
efd6: Initial change for that, but I need to be able to link directly to the version that is running, rather than the latest, and I don't see a way to do that right now.

packages/zscaler_zia/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -43,6 +43,32 @@ processors:
       tag: rename_resp_event
       target_field: json
       ignore_missing: true
+  - script:
+      params:
+        pkg_version: 3.12.0
```
kcreddy, line 48: Looks like we are not using this `pkg_version` param anywhere. Can w remove it?

efd6: Waiting for feedback from docs for how I can link directly to the correct version; it will be used then, that's why this commit is a wip. Without a direct link to the correct version, I'd rather not link at all.

kcreddy: Got it. My bad, I didn't expand the commit.

efd6: I have confirmation that this is not possible. The new code links to the markdown. It's not friendly, but it is at least correct. I would prefer correctness over friendliness, but happy to hear your position.

packages/zscaler_zia/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -37,6 +37,32 @@ processors:
       tag: rename_resp_event
       target_field: json
       ignore_missing: true
+  - script:
+      params:
+        pkg_version: 3.12.0
+        data_stream: audit-log
```
efd6, line 43: This is now no longer used. I'm OK with leaving it here if for if we ever get the ability to link to actual versioned documentation.

packages/zscaler_zia/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
kcreddy: Got it. Yeah correctness makes much more sense. I am okay with this.
Its probably going to be slightly inconvenient to maintain this `params.pkg_version` flag as it needs to be bumped every PR on `zscaler_zia`. 
Do you think we should document such exceptions into our wiki?

efd6: I think at least adding a comment above the manifest.yml `version` field. Would that do for you?

packages/zscaler_zia/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
kcreddy: Maybe we should add a comment so others don't delete it.

efd6: I've used it in the error string.

kcreddy (approved):

packages/zscaler_zia/_dev/build/docs/README.md
```diff
@@ -138,12 +138,12 @@ See: [Zscaler Vendor documentation](https://help.zscaler.com/zia/nss-feed-output
...
..."deviceostype":"%s{deviceostype}","deviceosversion":"%s{deviceosversion}","deviceowner":"%s{deviceowner}","devicetype":"%s{devicetype}","dnsapp":"%s{dnsapp}","dnsappcat":"%s{dnsappcat}","dns_gateway_status":"%s{dnsgw_flags}","dns_gateway_rule":"%s{dnsgw_slot}","dns_gateway_server_protocol":"%s{dnsgw_srv_proto}","category":"%s{domcat}","durationms":"%d{durationms}","ecs_prefix":"%s{ecs_prefix}","ecs_slot":"%s{ecs_slot}","epochtime":"%d{epochtime}","error":"%s{error}","hour":"%02d{hh}","http_code":"%s{http_code}","istcp":"%d{istcp}","loc":"%s{location}","login":"%s{login}","minutes":"%02d{mm}","month":"%s{mon}","month_of_year":"%02d{mth}","odevicehostname":"%s{odevicehostname}","odevicename":"%s{odevicename}","odeviceowner":"%s{odeviceowner}","odomcat":"%s{odomcat}","protocol":"%s{protocol}","recordid":"%d{recordid}","dns_req":"%s{req}","reqaction":"%s{reqaction}","reqrulelabel":"%s{reqrulelabel}","dns_reqtype":"%s{reqtype}","dns_resp":"%s{res}","resaction":"%s{resaction}","respipcategory":"%s{respipcat}","resrulelabel":"%s{resrulelabel}","restype":"%s{restype}","srv_dip":"%s{sip}","srv_dport":"%d{sport}","second":"%02d{ss}","datetime":"%s{time}","tz":"%s{tz}","year":"%04d{yyyy}"\}\}
```
chrisberkhout, line 141: This should be v2 to match the text above.

I previously went through the history and corrected the numbers to match the different versions that had existed: https://github.com/elastic/integrations/pull/10741

chrisberkhout: Also below in sample response.

packages/zscaler_zia/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Without templating in the pipeline, one thing we could do is use the agent config to insert the package version in the event (under `_conf`). Then the pipeline can read it from there.

chrisberkhout (changes requested):
I think this is good.

I noted an idea about getting the package version for the error message.

I would prefer strict fields or something like it to always be on. It doesn't fail all the processing, it just adds an error message. That may cause some level of disruption, but if it's an advanced option that's off by default, I think we'll almost never get this information when it matters.
My preference would be to set the error message by default, and have an option to skip it or downgrade it to a warning (although I'm not sure we have a great ECS field for warnings).

packages/zscaler_zia/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Oh, but maybe the agent config template also can't get the version without it being set in a variable. If it was set in a variable, at least it would all be in the manifest. I don't know how great this idea is. What do you think?

efd6: I thought about that and I didn't like the idea of exposing this in a way that is visible in the UI.

chrisberkhout (approved):
Thanks!

Issue #12999, opened 2025-03-06
[abnormal_security] Enrich Threat Events with Attachment and Link Details


cpascale43:
## Description
The current Abnormal Security integration does not include details for attachments and links associated with threats. Customers have requested that these details be included in the threat document to provide full context for each threat.

## Proposed Solution

- Modify the ingestion pipeline to make additional API calls to the `/v1/threats/{threat_id}/attachments` and `/v1/threats/{threat_id}/links` endpoints.
- Join the attachment and link details with the corresponding threat document during ingestion.
- Ensure that the enriched threat document includes all relevant details in a single document for easier analysis.

## API References
- [Attachments API](https://app.swaggerhub.com/apis/abnormal-security/abx/1.4.3#/Threats/v1_threats_attachments_retrieve)
- [Links API](https://app.swaggerhub.com/apis/abnormal-security/abx/1.4.3#/Threats/v1_threats_links_retrieve)

## Additional Notes
- This approach avoids the complexities of enrichment and ensures that all data is available in a single document.
- Sample data can be provided for testing if needed.

PR #13933, opened 2025-05-19
abnormal_security: enrich threat events with attachment and link details

chemamartinez:
## Proposed commit message

Added enrichment for threat events with attachments and links details.

API references:
- [Attachments API](https://app.swaggerhub.com/apis/abnormal-security/abx/1.4.3#/Threats/v1_threats_attachments_retrieve)
- [Links API](https://app.swaggerhub.com/apis/abnormal-security/abx/1.4.3#/Threats/v1_threats_links_retrieve)

It also fixes a bug by replacing the use of `abxMessageId` by `abxMessageIdStr`, as the first one was deprecated in the API. In addition, using directly the string value is safer, as converting the integer value into string caused sometimes a change in the value, probably because an overflow.

image

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/12999

## Screenshots

Added new option to enable enrichment for threat events:

image


packages/abnormal_security/data_stream/threat/_dev/test/pipeline/test-threat.log
```diff
@@ -1,5 +1,6 @@
...
...ed":true,"fromAddress":"john@example.com","fromName":"john","impersonatedParty":"None / Others","internetMessageId":"<AZz8NUMEST-qmuz77_koic@example>","isRead":false,"links":[{"display_text":"This is not a spoof!","domain":"lamronba.com","source":"body","type":"html href","url":"http://spoof.lamronba.com"},{"display_text":"This is not a spoof!","domain":"lamronba2.com","source":"body","type":"html href","url":"http://spoof.lamronba2.com"}],"postRemediated":false,"receivedTime":"2024-07-17T23:25:38Z","recipientAddress":"bob@example.com","remediationStatus":"Auto-Remediated","remediationTimestamp":"2024-07-17T23:25:45.73564Z","sentTime":"2024-07-17T23:25:29Z","subject":"YoU.have.\ud835\uddea\ud835\udfec0\ud835\udde1\ud835\udde1 a K0baIt 215-piece_ToooI_Set_Noo0wW..#GBOB","threatId":"bf255f2d-a2ad-3f50-5075-fdcc24308bbd","toAddresses":["bob@example.com"],"ccEmails":[],"replyToEmails":[],"returnPath":"bounce-bob_H181S7GUCF@example.com","senderDomain":"example.com","senderIpAddress":"81.2.69.142","summaryInsights":["Abnormal Email Body HTML","Invisible characters found in Email","Suspicious Link","Unusual Sender","Unusual Sender Domain"],"urlCount":1,"urls":["https://www.example.com/"]}
```
efd6, line 6: Add final new line.

packages/abnormal_security/data_stream/threat/agent/stream/cel.yml.hbs
```diff
@@ -86,7 +87,43 @@ program: |
         }).do_request().as(resp, resp.StatusCode == 200 ?
           bytes(resp.Body).decode_json().as(body,{
```
efd6, line 88:
```suggestion
          resp.Body.decode_json().as(body,{
```
since the kibana version supports this now (throughout).

packages/abnormal_security/data_stream/threat/agent/stream/cel.yml.hbs
```diff
@@ -86,7 +87,43 @@ program: |
...
+                    []
+                  ),
+                  "links": request("GET",
+                    state.url.trim_right("/") + "/v1/threats/" + string(state.worklist.threats[state.next].threatId) + "/links"
+                  ).with({
+                    "Header": {
+                      "Authorization": ["Bearer " + string(state.access_token)],
+                    }
+                  }).do_request().as(linkResp, linkResp.StatusCode == 200 ?
+                    bytes(linkResp.Body).decode_json().as(linkBody,
```
efd6, line 112:
```suggestion
                    linkResp.Body.decode_json().as(linkBody,
```

packages/abnormal_security/data_stream/threat/agent/stream/cel.yml.hbs
```diff
@@ -86,7 +87,43 @@ program: |
...
+                      has(linkBody.links) ? linkBody.links.filter(l, l.abxMessageIdStr == e.abxMessageIdStr).map(l, {
+                        "url": l.linkUrl,
+                        "display_text": l.displayText,
+                        "domain": l.domainLink,
+                        "type": l.linkType,
+                        "source": l.source
+                      }) : []
```
efd6, line 119:
```suggestion
                      has(linkBody.links) ?
                        linkBody.links.map(l, l.abxMessageIdStr == e.abxMessageIdStr, {
                          "url": l.linkUrl,
                          "display_text": l.displayText,
                          "domain": l.domainLink,
                          "type": l.linkType,
                          "source": l.source,
                        })
                      :
                        []
```

packages/abnormal_security/data_stream/threat/agent/stream/cel.yml.hbs
```diff
@@ -86,7 +87,43 @@ program: |
...
+                e.with({
+                  "attachments": request("GET",
+                    state.url.trim_right("/") + "/v1/threats/" + string(state.worklist.threats[state.next].threatId) + "/attachments"
+                  ).with({
+                    "Header": {
+                      "Authorization": ["Bearer " + string(state.access_token)],
+                    }
+                  }).do_request().as(attachResp, attachResp.StatusCode == 200 ?
+                    bytes(attachResp.Body).decode_json().as(attachBody,
+                      has(attachBody.attachments) ? attachBody.attachments.filter(a, a.abxMessageIdStr == e.abxMessageIdStr).map(a, a.attachmentName) : []
```
efd6, line 100:
```suggestion
                      has(attachBody.attachments) ?
                        attachBody.attachments.map(a, a.abxMessageIdStr == e.abxMessageIdStr, a.attachmentName)
                      :
                        []
```

packages/abnormal_security/_dev/deploy/docker/files/config.yml
```diff
@@ -512,6 +552,68 @@ rules:
...
+        headers:
+          Content-Type:
+            - 'application/json'
+        body: |-
+          {{ minify_json `
+          {
+              "attachments": [
+                  {
+                      "abxMessageIdStr": "2260288475997441028",
+                      "abxMessageId": 2260288475997441028,
```
kcreddy, line 571: The API doesn't seem to contain `abxMessageId`. Could you confirm if this field is still received from the API?

packages/abnormal_security/data_stream/threat/manifest.yml
```diff
@@ -38,6 +38,13 @@ streams:
         required: true
         show_user: false
         default: 30s
+      - name: enable_enrichment
+        type: bool
+        title: Enable Attachments and Links enrichment
+        required: false
+        show_user: false
+        default: true
```
kcreddy, line 46: Just a thought, I am wondering if this should be `default: false`. With `default: true`, it is not necessarily breaking anything, but it would lead to longer ingestion times because we now make 2 API calls per threatID. 
Maybe we should set `default: false` and document this option inside README for users needing the enrichment?

chemamartinez: That's a good point, I actually added the option to be able to disable the enrichment thinking on that.

About the default value, I think each approach has advantages and disadvantages, so I followed the criteria of the most similar case I found https://github.com/elastic/integrations/pull/8474 (it's true that in that case it was just adding extra decoding and storage but not API calls).

On one hand, enabling it by default would help users that upgrade integrations without taking care of the added changes, getting the enrichment out-of-the-box.

On the other hand, setting `default: false` is less intrusive, especially if it can affect the performance.

Maybe a good solution would be disabling it, document the option and also set `show_user: true` for more visibility. Do you agree?

packages/abnormal_security/_dev/deploy/docker/files/config.yml
chemamartinez: Sample data that we got included that field, so I added it to the tests, also to make sure its appearance doesn't take any effect.

efd6 (approved):
LGTM but please wait for @kcreddy

kcreddy (approved):
Thanks

Issue #13861, opened 2025-05-09
[Discussion] - Breaking change resulting from data type changes of mapped attributes


ShourieG:
In some scenarios we might have to make subsequent updates to existing integrations where the update would involve changing the data type of an existing mapped attribute. 

This could be because of reasons such as :- 

1. Existing attribute was initially assigned a wrong type.
2. Updates in the original data source/ api has changed the original data type due to internal reasons.
3. Performance reasons (this would be very rare), where indexing the data with it's original data type is becoming expensive for the end-user. etc ... 

In such cases, an update made to change the data type would be considered a breaking as it would create conflicts in kibana search queries / dashboards (if they were being used). Till now we have generally followed the philosophy of incrementing the major version of integrations when it comes to breaking changes. However in some cases such breaking changes have  been assessed to have minimal impact mainly due to - 

1. The attribute was very recently introduced in the integration and has not seen much use.
2. The attribute is not being used in Kibana dashboards at all.
3. The attribute might not have been even indexed in the first place due to data type mis-match.

Such a case was recently occurred in this [PR](https://github.com/elastic/integrations/pull/13581), and in the [discussion here](https://github.com/elastic/integrations/pull/13581#issuecomment-2824397344) it was clarified that even if a field is not explicitly indexed (because of a data type mismatch), it is still explicitly mapped and that would be enough for conflicts to occur when the backing index for the data stream gets created, more details in the discussion.

After discussions across the team it was how ever assessed that the impact of this change would be minimal, and the conflicts would self correct with an index rollover and passage of time, hence it was decided to mark this particular [PR](https://github.com/elastic/integrations/pull/13581) as a **bugfix** and not a **breaking-change**. It was also kind of decided that we would analyse such cases moving forward on a case by case basis and decide if they are significant to be a breaking-change or a bugfix. 

Recently another such [PR](https://github.com/elastic/integrations/pull/13825) has come into play, where there is a data type change leading to a breaking-change. However we have only done an  increment for the **"minor"** version here **(which is generally reserved for enhancements)** and marked it as a **breaking-change**.

The issue is that there seems to be a bit of confusion on how to handle the version increments and the type of change it should be marked as due to the case-case policy we are going by at the moment.

Issues resulting from a case-by-case approach:

1. Lack of clarity regarding version increments.
2. Incrementing major versions often for such small but impactful changes can cause disruption in the community and with end users in general.
[...]

PR #13977, opened 2025-05-23
[Netskope Alerts] Add text multi-field to netskope.alerts.breach.description field

### LINKED ISSUE #13857: [Netskope]: Fields netskope.alerts.breach.date and netskope.alerts.breach.description have the wrong mapping.
The fields `netskope.alerts.breach.date` and `netskope.alerts.breach.description` in the Alerts dataset of the Netskope integration have the wrong mapping.

The field `netskope.alerts.breach.date` is mapped as a `double`, but this is not a numeric field, it is a date, the value in the logs is an epoch time and the field `netskope.alerts.breach.description` is mapped as just `keyword`, but its content is a lenghty description, so it needs to follow other `*.description` fields and be mapped as a multi-field with a `.text` field as `match_only_text`.

For example, considering the sample message in the test files:

```
{"event":{"id":"47eccb9569fe50460ad1200f"},"netskope":{"alerts":{"insertion_epoch_timestamp":1642755526,"acked":"false","is_alert":"yes","alert":{"name":"cit0day - download1.mios.com","type":"breach"},"type":"Compromised Credential","breach":{"date":1601942400,"description":"In September 2020, a threat actor began sharing the millions of stolen credentials that were associated with a prominent Dark Web credentials service shut down by US federal authorites in August 2020. The stolen credentials represent hundreds of websites and hundreds of  millions of users and their associated passwords affected by the illegal antics of the threat actor who managed the now defunct Dark Web forum. Users and companies from all over the world were affected by these various breaches. This file contains the download1.mios.com accounts dump.","id":"bc6952df4c61b469cf4a47f17d0ea384","media_references":"null","score":40,"target_references":"null"},"category":{"name":"null"},"cci":"0","ccl":"unknown","count":1,"email":{"source":"Active"},"external":{"email":1},"matched":{"username":"user@example.com"},"organization":{"unit":"null"},"other":{"categories":"null"},"user":{"group":"null"},"app":{"category":"null"}}},"@timestamp":"2022-01-19T21:39:15.000Z","user":{"email":{"1":"user@example.com","2":"user@example.com","3":"user@example.com"}}}
```

The value for the field `breach.date` is `1601942400`

And this value is an epoch date:
```
$ date -d@1601942400
seg 05 out 2020 21:00:00 -03
```

And for the field `breach.description`, the value is a text:

> n September 2020, a threat actor began sharing the millions of stolen credentials that were associated with a prominent Dark Web credentials service shut down by US federal authorites in August 2020. The stolen credentials represent hundreds of websites and hundreds of  millions of users and their associated passwords affected by the illegal antics of the threat actor who managed the now defunct Dark Web forum. Users and companies from all over the world were affected by these various breaches. This file contains the download1.mios.com accounts dump.

In resume, the current mappings are:

- `netskope.alerts.breach.date`: `double`
- `netskope.alerts.breach.description`: `keyword`

But the mapping needs to be changed to:

- `netskope.alerts.breach.date`: `date`
- `netskope.alerts.breach.description`: `keyword` (`multi-field`)
- `netskope.alerts.breach.description.text`: `match_only_text` (`multi-field`)

leandrojmp:
## Proposed commit message

The value of the field `netskope.alerts.breach.date` is a date in unix epoch format, this field is incorrectly mapped as a `double`, but it should be mapped as a `date` as it represents the date when the breach was identified.

For this to work as a date a new `date` processor is required and added as well.

The value of the field `netskope.alerts.breach.description` can also be a lenghty description of the breach, but it is mapped only as a `keyword`, since sometimes the user may want to search for strings in this field it should be a `multi-field` with a `.text` field mapped as a `match_only_field`

One thing to add is that **this is a breaking change**, the mapping of the field `netskope.alerts.breach.date` is being changed from `double` to `date` and this will lead to a Kibana Conflict, which makes the field unusable until the conflict is solved.

I think that is very unlikely that this field is being used with the current mapping.

Also, the changelog specs does not allow to use multiple types, so since this is correcting the mappings and also adding a new processor, I'm tagging as an `enhancement` and bumped the minor version, I don't think this change justify bumping the major version.

## Checklist


leandrojmp:
The test cases were generated running `elastic-package test pipeline -g` and the `README.md` file was updated running `elastic-package build`.

efd6 (commented):
The manifest specifies v8.13.0 as the minimum version, so the test expectations should be rendered with that stack version.

leandrojmp:
> The manifest specifies v8.13.0 as the minimum version, so the test expectations should be rendered with that stack version.

Just did it, using `elastic-package stack up --version 8.13.0`, not sure exactly what this changed.

efd6 (commented):
As you note, this is a breaking change. That should be reflected in the changelog version bump. But also, do we want to split this into to parts: one part for the description mapping (which is not breaking and so is just a minor bump); and the other which is arguably a breaking bugfix. This would allow people to pick up the enhancement without any breakage. It would require this being broken into two PRs so that separate versions get published.

leandrojmp:
> As you note, this is a breaking change. That should be reflected in the changelog version bump. But also, do we want to split this into to parts: one part for the description mapping (which is not breaking and so is just a minor bump); and the other which is arguably a breaking bugfix. This would allow people to pick up the enhancement without any breakage. It would require this being broken into two PRs so that separate versions get published.

No problem, I can change this PR to just change the mapping for the description field and then create a new one for the mapping of the date field with the new processor.

The change in the mapping on the description field I would consider an enhancement, as the original mapping is not being changed, just a new `.text` mapping is being added.

But regarding to the versions, should both bump the minor version? I don't think that the breaking change on the date filter for the new PR would be enough to justify bumping this to version 2, there is some discussion on how to deal with these changes here: #13861

efd6:
> But regarding to the versions, should both bump the minor version? I don't think that the breaking change on the date filter for the new PR would be enough to justify bumping this to version 2

What would be enough?

efd6:
> Not exactly sure, something more complex, I guess.

Yeah, this is the issue. That is a vibe concept, but we need to have a solid construction. That solid construction is provided by semver. This change would break people, so it's a breaking change. Absent a policy that excludes this kind of breakage from the set that we consider to be breaking, I think we need to at least consider it. This is something that we are discussing internally.

> The main issue is that bumping major versions on breaking changes can lead to a faster increase of the integration version, which can give the false impression that a lot of things is changing.

I think this is an education issue; major means no more and no less than, we are breaking our continuation of API/behaviour contract. That people interpret it to mean progress is an issue with the interpretation.

efd6 (approved):
Suggested commit message body
```
The value of the field netskope.alerts.breach.description can be a
lengthy description of the breach, but it is mapped only as a keyword.
Since sometimes the user may want to search for strings in this field
it should be a multi-field with a .text field mapped as a
match_only_field.
```

leandrojmp:
Hello @efd6 

As asked, just created the other PR here #14008

PR #13982, opened 2025-05-23
[CI] Add new entries to skip triggering Buildkite builds

mrodm:
## Proposed commit message

Allow also to skip triggering Buildkite builds if Pull Requests just update:
- `.buildkite/pull-requests.json`
- `catalog-info.yaml`

### DIFF
--- a/.buildkite/pull-requests.json
+++ b/.buildkite/pull-requests.json
@@ -13,7 +13,7 @@
       "always_trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:build|test)\\W+(?:this|it))|^/test$|^/test benchmark fullreport$",
       "skip_ci_labels": [],
       "skip_target_branches": [],
-      "skip_ci_on_only_changed": ["^.github/workflows/", "^.github/dependabot.yml", "^.github/ISSUE_TEMPLATE/", "^docs/"],
+      "skip_ci_on_only_changed": ["^.github/workflows/", "^.github/dependabot.yml$", "^.github/ISSUE_TEMPLATE/", "^docs/", "^catalog-info.yaml$", "^.buildkite/pull-requests.json$"],
       "always_require_ci_on_changed": []
     },
     {


.buildkite/scripts/common.sh
```diff
@@ -749,7 +749,7 @@ is_pr_affected() {
     # Example:
     # https://buildkite.com/elastic/integrations/builds/25606
     # https://github.com/elastic/integrations/pull/13810
-    if git diff --name-only "${commit_merge}" "${to}" | grep -E -v '^(packages/|\.github/(CODEOWNERS|ISSUE_TEMPLATE|PULL_REQUEST_TEMPLATE)|README\.md|docs/)' > /dev/null; then
+    if git diff --name-only "${commit_merge}" "${to}" | grep -E -v '^(packages/|\.github/(CODEOWNERS|ISSUE_TEMPLATE|PULL_REQUEST_TEMPLATE)|README\.md|docs/|catalog-info\.yaml|\.buildkite/pull-requests\.json)' > /dev/null; then
```
mrodm, line 752: The files added in `skip_ci_on_only_changed` setting from `.buildkite/pull-requests.json` are also ignored when identifying non-package modified files in Pull Requests.

So, if `catalog-info.yaml` or `.buildkite/pull-requests.json` are modified, it will not cause that all packages are tested in that build. As an example, if `.buildkite/pipeline.yml` is modified in a Pull Request (that it is not part of that Regex), that CI build would test all packages.

.buildkite/pull-requests.json
```diff
@@ -13,7 +13,7 @@
       "always_trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:build|test)\\W+(?:this|it))|^/test$|^/test benchmark fullreport$",
       "skip_ci_labels": [],
       "skip_target_branches": [],
-      "skip_ci_on_only_changed": ["^.github/workflows/", "^.github/dependabot.yml", "^.github/ISSUE_TEMPLATE/", "^docs/"],
+      "skip_ci_on_only_changed": ["^.github/workflows/", "^.github/dependabot.yml$", "^.github/ISSUE_TEMPLATE/", "^docs/", "^catalog-info.yaml$", "^.buildkite/pull-requests.json$"],
```
mrodm, line 16: Changes in `catalog-info.yaml` should be validated (at least syntax) by the GitHub Workflow https://github.com/elastic/integrations/blob/bb6c198364643d7eedbbb7b2c6d9daab27982d7c/.github/workflows/catalog-info.yml

v1v (approved):

PR #14001, opened 2025-05-26
[Box Events] Update and expand ECS `user.*` field mappings

### LINKED ISSUE #12971: [box_events]: user.XXX fields wrongly set, not conforming to ECS
### Integration Name

Box Events [box_events]

### Dataset Name

box_events.events

### Integration Version

2.13.0

### Agent Version

8.16.1

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.16.1

Ubuntu 20.04

### Software/API Version

Box API

### Error Message

@efd6 thanks for addressing #12920 

That seems to have affected all user.XXX fields now set, not only for the COLLABORATION_INVITE type of event.

However, I think some values aren't properly assigned

Given the event.json attached to #12920 which I'll reattach here:

the created_by.name ends up as: user.name, and the created_by.login ends up as user.id, and the domain part as user.domain. Same rules seem to be applied to the accessible_by values in the event.json.

I think the following should apply: 
 * created_by.name -> user.full_name
 * created_by.login -> user.email
   * split into user.name and user.domain
 * created_by.id -> user.id

I believe that should conform a bit better to ECS.

And if there is no created_by.name in the source for some events as it seems, currently "Unknown User" is set. Maybe just have that field empty string or not set at all? Having "Unknown User" in many of the events, isn't overly helpful.

And same applies to the accessible_by values.

As it was just only introduced, I don't see this would cause much trouble to anyone, or if at all only few.

### Event Original

[event(1).json](https://github.com/user-attachments/files/19092260/event.1.json)

### What did you do?

updated integration to 2.13.0

### What did you see?

as explained above, what I think wrong assignment of user fields.

as explained above, a better ECS conformity of user.* and user.target.* fields.

### Anything else?

_No response_

moxarth-rathod:
## Proposed commit message

```
box_events: improve the ecs field mappings for `user.*` fields

This PR is based on the earlier PR [1] that addressed improvements to the user.* field mappings.
In this update, some of the existing mappings have been revised and several new ones have been added.

[1] https://github.com/elastic/integrations/pull/12944
```

## Checklist

Clone integrations repo.
Install elastic package locally.
Start elastic stack using elastic-package.
Move to integrations/packages/box_events directory.
> elastic-package test

## Related issues

- Closes #12971

### DIFF
--- a/packages/box_events/changelog.yml
+++ b/packages/box_events/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.14.0"
+  changes:
+    - description: Update and expand ECS `user.*` field mappings.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
 - version: "2.13.0"
   changes:
     - description: Handle collaboration invite events and improve user field handling.
--- a/packages/box_events/changelog.yml
+++ b/packages/box_events/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Update and expand ECS `user.*` field mappings.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/1
+      link: https://github.com/elastic/integrations/pull/14001
 - version: "2.13.0"
   changes:
     - description: Handle collaboration invite events and improve user field handling.
--- a/packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
@@ -619,7 +619,7 @@ processors:
       if: ctx._tmp?.created_login instanceof List && ctx._tmp.created_login.length == 2
   - set:
       field: user.id
-      copy_from: box.created_by.id
+      copy_from: box.created_by.login
       ignore_empty_value: true
       if: ctx.user?.id == null && ctx.box?.created_by?.type == 'user'
   - append:
@@ -656,10 +656,10 @@ processors:
       field: user.target.domain
       copy_from: _tmp.accessible_login.1
       if: ctx._tmp?.accessible_login instanceof List && ctx._tmp.accessible_login.length == 2
-  - set:
-      field: user.target.id
-      copy_from: box.accessible_by.id
-      ignore_empty_value: true
+  - rename:
+      field: box.accessible_by.login
+      target_field: user.target.id
+      ignore_missing: true
       if: ctx.user?.target?.id == null && ctx.box?.accessible_by?.type == 'user'
   - append:
       field: related.user
--- a/packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
@@ -594,64 +594,82 @@ processors:
       target_field: event.id
       ignore_missing: true
 # User details: set rather than rename is used in these to avoid breaking change.
+  - set:
+      field: user.full_name
+      copy_from: box.created_by.name
+      ignore_empty_value: true
+      if: ctx.user?.full_name == null && ctx.box?.created_by?.type == 'user'
+  - set:
+      field: user.email
+      copy_from: box.created_by.login
+      ignore_empty_value: true
+      if: ctx.user?.email == null && ctx.box?.created_by?.login != null
   - split:
       field: box.created_by.login
       target_field: _tmp.created_login
       separator: '@'
       if: ctx.box?.created_by?.login instanceof String && ctx.box.created_by.login.length() > 2
   - set:
+      field: user.name
+      copy_from: _tmp.created_login.0
+      if: ctx._tmp?.created_login instanceof List && ctx._tmp.created_login.length == 2
+  - set:
       field: user.domain
       copy_from: _tmp.created_login.1
       if: ctx._tmp?.created_login instanceof List && ctx._tmp.created_login.length == 2
   - set:
       field: user.id
-      copy_from: box.created_by.login
+      copy_from: box.created_by.id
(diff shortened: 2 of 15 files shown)


packages/box_events/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.14.0"
+  changes:
+    - description: Update and expand ECS `user.*` field mappings.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
```
ShourieG, line 6:
```suggestion
      link: https://github.com/elastic/integrations/pull/14001
```

packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -594,64 +594,82 @@ processors:
...
+      if: ctx.user?.email == null && ctx.box?.created_by?.login != null
   - split:
       field: box.created_by.login
       target_field: _tmp.created_login
       separator: '@'
       if: ctx.box?.created_by?.login instanceof String && ctx.box.created_by.login.length() > 2
+  - set:
+      field: user.name
+      copy_from: _tmp.created_login.0
+      if: ctx._tmp?.created_login instanceof List && ctx._tmp.created_login.length == 2
```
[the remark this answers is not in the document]
ShourieG: Also is this always the scenario if `ctx._tmp.created_login.length == 2` then the 1st element will contain the name ? What does the [1] position contain then ?

packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -594,64 +594,82 @@ processors:
...
-      copy_from: box.created_by.login
+      copy_from: box.created_by.id
       ignore_empty_value: true
       if: ctx.user?.id == null && ctx.box?.created_by?.type == 'user'
   - append:
       field: related.user
-      value: '{{{user.id}}}'
-      if: ctx.user?.id != null && ctx.user.id != ''
+      value: '{{{user.email}}}'
+      if: ctx.user?.email != null && ctx.user.email != ''
```
ShourieG, line 628: Why are we replacing `id` with `email` ? Can we not use the [ ECS email fields](https://www.elastic.co/guide/en/ecs/8.17/ecs-email.html) to add more info related to emails instead of replacing the one in `related.user`

efd6: This actually maintains the current behaviour due to other changes here.

packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -594,64 +594,82 @@ processors:
...
+      field: user.target.id
+      copy_from: box.accessible_by.id
+      ignore_empty_value: true
       if: ctx.user?.target?.id == null && ctx.box?.accessible_by?.type == 'user'
   - append:
       field: related.user
-      value: '{{{user.target.id}}}'
-      if: ctx.user?.target?.id != null && ctx.user.target.id != ''
+      value: '{{{user.target.email}}}'
+      if: ctx.user?.target?.email != null && ctx.user.target.email != ''
```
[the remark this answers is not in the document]
efd6: I don't think this is the case, here, the email is associated with the user that is the target of the box operation, not to any email transaction.

packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -594,64 +594,82 @@ processors:
...
+      copy_from: _tmp.created_login.0
+      if: ctx._tmp?.created_login instanceof List && ctx._tmp.created_login.length == 2
   - set:
       field: user.domain
       copy_from: _tmp.created_login.1
       if: ctx._tmp?.created_login instanceof List && ctx._tmp.created_login.length == 2
   - set:
       field: user.id
-      copy_from: box.created_by.login
+      copy_from: box.created_by.id
```
efd6, line 622: I'm not sure this is the right thing to do. The IDs here are internal box IDs, while previously we were using the login, which is more likely to be common across services (being an email address). This means that from an entitly analytics perspective, the previously used value is more likely to be useful.

packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -594,64 +594,82 @@ processors:
...
   - set:
       field: user.target.domain
       copy_from: _tmp.accessible_login.1
       if: ctx._tmp?.accessible_login instanceof List && ctx._tmp.accessible_login.length == 2
-  - rename:
-      field: box.accessible_by.login
-      target_field: user.target.id
-      ignore_missing: true
+  - set:
+      field: user.target.id
```
efd6, line 660: Same comment here.

packages/box_events/data_stream/events/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
moxarth-rathod: Yes, right. As Dan said [1] position will be the the domain.

> can we not use user.full_name here if present or is this specific user.name different in some way as in it's a custom username ?

yes, it can be a custom name in the email address so we've used user.name instead of user.full_name here.

ShourieG:
@moxarth-rathod LGTM from my end, please wait for Dan's approval

ShourieG (approved):
Approving from my end, but please wait for @efd6 to give final approval before merging.

efd6 (approved):

PR #14056, opened 2025-05-29
o365: validate field type in pipeline before accessing subfields

chemamartinez:
## Proposed commit message

As `organization` field may be a string instead of an object, the following error can appear when accessing its subfields being a string value.

```
Processor 'conditional' with tag '' failed with message 'dynamic getter [java.lang.String, id] not found'
```

Added a type check before accessing its subfields to avoid the above error.

## Checklist


efd6 (approved):

efd6:
Though it would be nice to have a test for this.

PR #14067, opened 2025-05-29
[symantec_endpoint_security] Fix processing errors for when some fields are missing

StacieClark-Elastic:
A customer was having issues with their ingest missing some expected fields. 

1. Nested foreach processor converting items in  attribute_ids array in a file object in the cybox.files array with fails when attribute_ids is missing. Tag is foreach_cybox_files.
2. Nested foreach processor converting items in  signature_value_ids array in a file object in the cybox.files array with fails when signature_value_id is missing. Tag is foreach_signature_value_ids.
3. A painless script fails when the type_id field  in a file object in  the cybox.files array is missing. Tag script_to_add_file_type.
4. foreach processor for the file.size array in the default processor fails when the size item is an empty string. Tag foreach_file_size.

## Proposed commit message
This fixes a variety of processing errors in pipeline_object_cybox.yml

* foreach over cybox.files array with tag foreach_signature_value_ids fails
  when the field signature_value_ids does not exist: replaced with a
  painless script with tag script_to_convert_signature_value_ids.
* foreach over cybox.files array with tag foreach_cybox_files that fails when the
  field attribute_ids does not exist: replaced with a painless script
  with tag tag: script_to_convert_attribute_ids, and added null check
  to an existing painless script that accesses attribute_ids.
* script processor with tag script_to_add_file_type that fails when the type_id
   does not exist: added a null check on type_id in painless script.

This fixes a processing error in default.yml
* covert processor with tag convert_file_size_to_long that fails when file.size is
   an empty string: replaced with a painless script

## Checklist

- [] I have verified that all data streams collect metrics or logs.
- [NA ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

Test files have been added to the pipeline tests. One file per error listed in the description. None of the older pipeline tests were modified and they still pass. 

One thing that I noticed is the system tests for the event datastream fail with a terraform error on main. They continue to do so with the PR.

Running on the old version to see the errors with the new pipeline tests then running on the new version verifies the fix.


packages/symantec_endpoint_security/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.11.1"
+  changes:
+    - description: Ignore failures on foreach processing of some missing subfields
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/99999
```
efd6, line 6:
```suggestion
      link: https://github.com/elastic/integrations/pull/14067
```

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/pipeline_object_cybox.yml
```diff
@@ -156,27 +156,33 @@ processors:
...
+              def new_ids = [];
+              for (int j = 0; j < file.attribute_ids.length; j++) {
+                def id = file.attribute_ids[j];
+                if (id != null) {
+                  new_ids.add(id.toString());
+                }
+              }
+              file.attribute_ids = new_ids;
+            }
+          }
```
efd6, line 179:
```suggestion
          for (int i = 0; i < ctx.ses.cybox.files.length; i++) {
            def file = ctx.ses.cybox.files[i];
            if (!file.containsKey('attribute_ids') || file.attribute_ids == null) {
              continue;
            }
            file.attribute_ids = [];
            for (int j = 0; j < file.attribute_ids.length; j++) {
              def id = file.attribute_ids[j];
              if (id != null) {
                file.attribute_ids.add(id.toString());
              }
            }
          }
```
- shift work left
- less intermediates (could keep the intermediate if it's only conditionally assigned to `file.attribute_ids`)

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/pipeline_object_cybox.yml
```diff
@@ -202,10 +208,12 @@ processors:
             }
           }
         for (def file : ctx.ses.cybox.files) {
-          for (def id : file.attribute_ids) {
-            if (params.containsKey(id.toString())) {
-              def type = params.get(id.toString());
-              var.add(type);
+        if (file.attribute_ids != null) {
```
efd6, line 211:
```suggestion
          if (file.attribute_ids != null) {
```

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/pipeline_object_cybox.yml
```diff
@@ -572,27 +580,32 @@ processors:
...
-              on_failure:
-                - remove:
-                    field: _ingest._value
-                    ignore_missing: true
-                - append:
-                    field: error.message
-                    value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+
+      lang: painless
+      source: >-
```
efd6, line 590: Similar changes here.

packages/symantec_endpoint_security/data_stream/event/_dev/test/pipeline/test-missing-attribute-ids.log
```diff
@@ -0,0 +1 @@
...ic_has_pii": true,"stic_hw_uid": "Test STIC_HW_UID","stic_ip_hash": "Test STIC_IP_Hash","stic_legacy_ent_uids": ["Test STIC_Legacy_Ent_UIDs 1","Test STIC_Legacy_Ent_UIDs 2"],"stic_legacy_hw_uids": ["Test STIC_Legacy_HW_UIDs 1","Test STIC_Legacy_HW_UIDs 2"],"stic_legacy_uids": ["Test STIC_Legacy_UIDs 1","Test STIC_Legacy_UIDs 2"],"stic_schema_id": "Test STIC_Schema_ID","stic_uid": "Test STIC_UID","stic_version": "Test STIC_Version","subfeature_name": "Test Subfeature_Name","time": "2024-02-29T02:00:00Z","timezone": 12345678901,"total": 12345678901,"type": "SCAN","type_id": 8020,"user": {"account_disabled": true,"cloud_resource_uid": "Test User-Cloud_Resource_UID","domain": "Test User-Domain","external_account_uid": "Test User-External_Account_UID","external_uid": "Test User-External_UID","full_name": "Test User-Full_Name","groups": ["Test User-Group 1","Test User-Groups 1"],"home": "Test User-Home","is_admin": true,"logon_name": "Test User-Logon_Name","name": "Test User-Name","password_expires": true,"shell": "Test User-Shell","sid": "Test User-SID","uid": "Test User-UID"},"user_name": "Test User_Name","user_uid": "Test User_UID","uuid": "Test UUID","verdict_id": 0,"version": "1.4"}
```
efd6, line 1: Add final new line, but can we make this more surgical so that the part that it is testing is more clearly obvious? Same for the tests below.

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1170,12 +1170,14 @@ processors:
       field: file.size
       if: ctx.file?.size instanceof List
       tag: foreach_file_size
+
```
efd6, line 1173:
```suggestion
```

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1170,12 +1170,14 @@ processors:
...
       processor:
         convert:
           field: _ingest._value
           tag: convert_file_size_to_long
           type: long
           ignore_missing: true
+          ignore_failure: true
```
efd6, line 1180: I don't think this is right. If this is to avoid a failure when the value is `""`, then I think this will need to be a script as well. As it is, the processor will now ignore all failures, and the `on_failure` below will never fire, even in cases where it should.

efd6 (commented):
Suggest the following for the commit message body
```
This fixes a variety of processing errors:

* foreach over cybox.files array with tag foreach_signature_value_ids fails
  when the field signature_value_ids does not exist: replaced with a
  painless script with tag script_to_convert_signature_value_ids.
* foreach over cybox.files array with tag foreach_cybox_files that fails when the
  field attribute_ids does not exist: replaced with a painless script
  with tag tag: script_to_convert_attribute_ids, and added null check
  to an existing painless script that accesses attribute_ids.
<snip>
```
`<snip>`ed because I could not find `script_to_add_file_type` and I think the `convert_file_size_to_long` change needs work.

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/pipeline_object_cybox.yml
StacieClark-Elastic: I see what you are trying to accomplish. I don't think the example will work.
file.attributes_ids is reassigned to an empty array, then you are trying to access items in the now empty array. 
`          file.attribute_ids = [];
            for (int j = 0; j < file.attribute_ids.length; j++)`
I would do a direct replacement in the array if I didn't need to check for null.

efd6: You are of course correct. I was getting ahead of myself. The more important part of what I was suggesting in the shift left.
```
          for (int i = 0; i < ctx.ses.cybox.files.length; i++) {
            def file = ctx.ses.cybox.files[i];
            if (file.attribute_ids == null || !file.containsKey('attribute_ids')) {
              continue;
            }
            def new_ids = [];
            for (int j = 0; j < file.attribute_ids.length; j++) {
              def id = file.attribute_ids[j];
              if (id != null) {
                new_ids.add(id.toString());
              }
            }
            file.attribute_ids = new_ids;
          }
```

packages/symantec_endpoint_security/data_stream/event/_dev/test/pipeline/test-missing-file-size.log
```diff
@@ -0,0 +1,3 @@
+{"category_id":3,"composite":1,"end_time":"2024-02-29T01:00:00.000Z","event_id":5000,"file":{"attribute_ids":[1],"attributes":12345678901,"signature_value":12345678901,"signature_value_ids":[1],"size":12345678901,"size_compressed":12345678901,"type_id":1,"version":"Test File-Version"},"severity_id":0,"time":"2024-02-29T02:00:00Z","timezone":12345678901,"type_id":5}
+{"actor":{"file":{"attribute_ids":[1],"attributes":12345678901,"signature_value_ids":[0],"size":"","type_id":1},"integrity_id":0,"start_time":1613021404000},"category_id":5,"end_time":"2024-02-29T01:00:00.000Z","is_user_present":false,"type_id":8000}
+{"actor":{"file":{"attribute_ids":[1],"attributes":12345678902,"signature_value_ids":[0],"type_id":1},"integrity_id":0,"start_time":1613021404000},"category_id":5,"end_time":"2024-02-29T01:00:01.000Z","is_user_present":false,"type_id":8000}
```
efd6: Final new line.

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1166,23 +1166,31 @@ processors:
...
+        def new_sizes = [];
+        for (int i = 0; i < ctx.file.size.length; i++) {
+          def sz = ctx.file.size[i];
+          if (ctx.file.size[i] != null) {
+            if (ctx.file.size[i] instanceof long || ctx.file.size[i] instanceof int) {
+              new_sizes.add((long)ctx.file.size[i]);
+            } else if (ctx.file.size[i] != "") {
+                new_sizes.add(Long.parseLong(ctx.file.size[i]));
+            }
+          }
```
efd6:
```suggestion
          def sz = ctx.file.size[i];
          if (sz != null) {
            continue;
          }
          if (sz instanceof long || sz instanceof int) {
            new_sizes.add((long)sz);
          } else if (sz instanceof String && sz != "") {
            new_sizes.add(Long.parseLong(sz));
          }
```
The `sz instanceof String` is to avoid the error. If you don't want to avoid the error, it can be omitted. I can see arguments both ways.

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/pipeline_object_cybox.yml
```diff
@@ -156,27 +156,32 @@ processors:
...
+              def new_ids = [];
+              for (int j = 0; j < file.attribute_ids.length; j++) {
+                if (file.attribute_ids[j] != null) {
+                  new_ids.add(file.attribute_ids[j].toString());
+                }
+              }
+              file.attribute_ids = new_ids;
+            }
+          }
+
```
efd6:
```suggestion
        for (int i = 0; i < ctx.ses.cybox.files.length; i++) {
          def file = ctx.ses.cybox.files[i];
          if (file.attribute_ids == null || !file.containsKey('attribute_ids')) {
            continue;
          }
          def new_ids = [];
          for (int j = 0; j < file.attribute_ids.length; j++) {
            def attr_id = file.attribute_ids[j];
            if (attr_id != null) {
              new_ids.add(attr_id.toString());
            }
          }
          if (new_ids.length != 0) {
            file.attribute_ids = new_ids;
          }
        }
```

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/pipeline_object_cybox.yml
```diff
@@ -202,10 +207,12 @@ processors:
...
-              def type = params.get(id.toString());
-              var.add(type);
+          if (file.attribute_ids != null) {
+            for (def id : file.attribute_ids) {
+              if (params.containsKey(id.toString())) {
+                def type = params.get(id.toString());
+                var.add(type);
+              }
             }
           }
```
efd6:
```suggestion
          if (file.attribute_ids == null) {
            continue;
          }
          for (def id : file.attribute_ids) {
            def type = params[id.toString()];
            if (type != null) {
              var.add(type);
            }
          }
```

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/pipeline_object_cybox.yml
```diff
@@ -572,27 +579,31 @@ processors:
...
+            if (file.containsKey('signature_value_ids') && file.signature_value_ids != null) {
+            def new_ids = [];
+            for (int j = 0; j < file.signature_value_ids.length; j++) {
+              if (file.signature_value_ids[j] != null) {
+                new_ids.add(file.signature_value_ids[j].toString());
+              }
+            }
+            file.signature_value_ids = new_ids;
+            }
+          }
```
efd6:
```suggestion
        for (int i = 0; i < ctx.ses.cybox.files.length; i++) {
          def file = ctx.ses.cybox.files[i];
          if (file.signature_value_ids == null || !file.containsKey('signature_value_ids')) {
            continue;
          }
          def new_ids = [];
          for (int j = 0; j < file.signature_value_ids.length; j++) {
            def value_id = file.signature_value_ids[j];
            if (value_id != null) {
              new_ids.add(value_id.toString());
            }
          }
          if (new_ids.length != 0) {
            file.signature_value_ids = new_ids;
          }
        }
```

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/pipeline_object_cybox.yml
```diff
@@ -702,10 +713,12 @@ processors:
...
-          if (params.containsKey(type_id.toString())) {
-            def type = params.get(type_id.toString());
-            var.add(type);
+          if (file.type_id != null) {
+            def type_id = file.type_id;
+            if (params.containsKey(type_id.toString())) {
+              def type = params.get(type_id.toString());
+              var.add(type);
+            }
           }
```
efd6:
```suggestion
          if (file.type_id == null) {
            continue;
          }
          def type = params[file.type_id.toString()];
          if (type != null) {
            var.add(type);
          }
```

packages/symantec_endpoint_security/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1166,23 +1166,31 @@ processors:
       if: ctx.ses?.category_id != null && ctx.ses.category_id == '6'
       tag: pipeline_category_diagnostic
       ignore_missing_pipeline: true
-  - foreach:
-      field: file.size
+
+  - script:
+      description: Convert size array to longs
+      tag: script_to_convert_file_size
       if: ctx.file?.size instanceof List
```
efd6:
```suggestion
      if: ctx.file?.size instanceof List && ctx.file.size.length != 0
```

PR #14068, opened 2025-05-29
Add `fips_compatible` flag to multiple packages

shmsr:
## Proposed commit message

Add a `fips_compatible` flag to some packages that we don't want to allow in FedRAMP env.

## Checklist


shmsr:
Although we want to disable sql_input but we can't as of now because of: https://github.com/elastic/package-spec/issues/901

shmsr:
Also, tests are failing as it's saying GA packages are using unreleased version: v3.4.0 of package-spec.

```
Error: checking package failed: linting package failed: found 4 validation errors:
  | 1. file "/opt/buildkite-agent/builds/bk-agent-prod-gcp-1748545681270973011/elastic/integrations/packages/mongodb/manifest.yml": package with GA version (1.20.0) is using an unreleased version of the spec (3.4.0-next) (PSR00001)
  | 2. file "/opt/buildkite-agent/builds/bk-agent-prod-gcp-1748545681270973011/elastic/integrations/packages/mongodb/data_stream/log/elasticsearch/ingest_pipeline/pipeline-json.yml" is invalid: field processors.11.remove: if is required
  | 3. file "/opt/buildkite-agent/builds/bk-agent-prod-gcp-1748545681270973011/elastic/integrations/packages/mongodb/data_stream/log/elasticsearch/ingest_pipeline/pipeline-json.yml" is invalid: field processors.11.remove.field: rename "message" to "event.original" processor requires remove "message" processor (JSE00001)
  | 4. file "/opt/buildkite-agent/builds/bk-agent-prod-gcp-1748545681270973011/elastic/integrations/packages/mongodb/manifest.yml" is invalid: field (root): Additional property fips_compatible is not allowed
  |  
```

packages/oracle/manifest.yml
```diff
@@ -12,6 +12,7 @@ conditions:
     version: "^8.13.0 || ^9.0.0"
   elastic:
     subscription: basic
+fips_compatible: false
```
ishleenk17: From which stack version onwards are we claiming FIPS compatibility?

shmsr: This is a package-spec thing. For stacks, versions 8.19.0 and 9.1.0 onwards are going to be FIPS 140-2 compliant but I think the change here doesn't matter regardless of the stack.

In case we fix the modules to be compliant in future, then we can bump the stack and make this flag true.

packages/postgresql/manifest.yml
```diff
@@ -1,7 +1,7 @@
-format_version: "3.0.2"
+format_version: "3.4.0"
```
ishleenk17: Is upgrading this format version required for supporting this flag ?

ishleenk17:
> Although we want to disable sql_input but we can't as of now because of: [elastic/package-spec#901](https://github.com/elastic/package-spec/issues/901)

@shmsr  Since the dependency is not on this as confirmed [here](https://github.com/elastic/package-spec/issues/901#issuecomment-2959298259), are we including this for disabling fips flag ?

shmsr:
> > Although we want to disable sql_input but we can't as of now because of: [elastic/package-spec#901](https://github.com/elastic/package-spec/issues/901)
> 
> @shmsr Since the dependency is not on this as confirmed [here](https://github.com/elastic/package-spec/issues/901#issuecomment-2959298259), are we including this for disabling fips flag ?

As pointed by Jaime, there's no version constraint for the flag. As soon as the latest spec releases, and CI passes will make the change for sql_input.

tommyers-elastic (approved):

shmsr:
@jsoriano / @mrodm  I am still getting the error:

>    3. file "/Users/subhamsarkar/go/src/github.com/elastic/integrations/build/packages/mysql-1.27.0.zip/manifest.yml" is invalid: field (root): Additional property fips_compatible is not allowed

I thought integrations repo has been updated with the latest version of ep and hence the 3.4.0 spec, so why am I getting this error. Can you please help?

jsoriano:
> @jsoriano / @mrodm I am still getting the error:
> 
> > 3. file "/Users/subhamsarkar/go/src/github.com/elastic/integrations/build/packages/mysql-1.27.0.zip/manifest.yml" is invalid: field (root): Additional property fips_compatible is not allowed
> 
> I thought integrations repo has been updated with the latest version of ep and hence the 3.4.0 spec, so why am I getting this error. Can you please help?

The flag was added at the package policy level, not at the root level of the manifest. Also, it was added only for integration packages, not for input packages.

I have prepared a PR with tests that illustrate where the flag is expected, and with support for input packages: https://github.com/elastic/package-spec/pull/914

shmsr:
> > @jsoriano / @mrodm I am still getting the error:
> > > 3. file "/Users/subhamsarkar/go/src/github.com/elastic/integrations/build/packages/mysql-1.27.0.zip/manifest.yml" is invalid: field (root): Additional property fips_compatible is not allowed
> > 
> > 
> > I thought integrations repo has been updated with the latest version of ep and hence the 3.4.0 spec, so why am I getting this error. Can you please help?
> 
> The flag was added at the package policy level, not at the root level of the manifest. Also, it was added only for integration packages, not for input packages.
> 
> I have prepared a PR with tests that illustrate where the flag is expected, and with support for input packages: [elastic/package-spec#914](https://github.com/elastic/package-spec/pull/914)

Thanks a lot @jsoriano!

@tommyers-elastic For now I am leaving out `sql_input` as the flag is not there for input packages; will add it later when it's available.

shmsr:
Can someone from @elastic/obs-ds-hosted-services also please review? Based on the modules that have been disabled in Beats, I have disabled them in Integrations as well.

SQL input still remains as the flag is not available as of yet for input packages.

ishleenk17:
@shmsr  : Changes look good. 
So looks like sql input we are not doing currently.
Also the ones that are not FIPS compatible. Are there some corresponding changes for it in beats. Or are they due to the way the beats code is fetching the metrics that makes it fips incompatible ?

shmsr:
> @shmsr : Changes look good. So looks like sql input we are not doing currently. Also the ones that are not FIPS compatible. Are there some corresponding changes for it in beats. Or are they due to the way the beats code is fetching the metrics that makes it fips incompatible ?

The modules are using code that contains FIPS non-approved cryptographic algorithms, either directly or through dependencies. While we could have kept some modules enabled since we're unlikely to execute the non-compliant code paths (as per manual review), keeping them enabled helped us identify the problematic libraries in our dependency tree (e.g., azure). Therefore, we've disabled all modules that could potentially execute non-FIPS-compliant code paths.

ishleenk17 (approved):

packages/postgresql/manifest.yml
jsoriano: I don't think so. We ~~could~~ can keep previous format versions.

jsoriano (approved):

PR #14070, opened 2025-05-30
[AWS] Add lambda logs dataset

agithomas:
- Enhancement

## Proposed commit message

Add a dataset for AWS lambda logs. 

## Checklist

## Author's Checklist

- [x] Added pipeline test
- [x] Perform integration testing
- [x] Add support for the different category of logs including JSON / plain-text across various programming language
- [x] Add support for fallback  

- `elastic-package build &&  elastic-package stack up -v -d --services package-registry` 

## Screenshots

No screenshots added as part of this PR.


packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
```diff
@@ -0,0 +1,194 @@
...
+    formats: ["UNIX_MS"]
+    ignore_failure: true
+
+# Flatten important fields from each log type
+
+- set: { field: aws.lambda.message, if: "ctx.parsed?.record != null", copy_from: parsed.record, ignore_failure: true }
+- set: { field: aws.lambda.message, if: "ctx.parsed?._aws != null", copy_from: parsed._aws, ignore_failure: true }
+- rename: { field: parsed.service, target_field: service.name, ignore_missing: true }
+- rename: { field: parsed.level, target_field: log.level, ignore_missing: true }
+- rename: { field: parsed.requestId, target_field: aws.lambda.request_id, ignore_missing: true }
```
efd6: I think this style is harder to read than the pure YAML. Same below.

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,48 @@
...
+        name: '{{ IngestPipeline "aws-lambda-json" }}'
+    - remove:
+        field: first_char
+        ignore_missing: true
+        description: 'Removes the `first_char` field used to determine the log format.'
+on_failure:
+- set:
+    field: event.kind
+    value: pipeline_error
+- set:
```
efd6: This being a `set` rather than `append` will clobber any error messages that already exist. I don't think this is a problem at the moment, but if more errorable processors are added it will become one.

efd6 (commented):
review as codeowner

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
agithomas: Thanks @efd6 for your input. 

I found this style (Flow style mapping) helpful to reduce the number of lines in the file. 

I understand that Flow style mapping is not usually followed, and I was not sure if this style would impact the readability. 

Now that you have raised the concern, I will change to the style we always follow - block style mapping.

efd6: Thanks

agithomas: Updated the YAML to follow block style. Marking as resolved.

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
```diff
@@ -0,0 +1,498 @@
...
+# 3. Platform report logs (time field)
+- date:
+    if: "ctx.parsed?.time != null"
+    field: parsed.time
+    target_field: "@timestamp"
+    formats: ["ISO8601"]
+    ignore_failure: true
+
+- date:
+    if: "ctx.parsed?._aws?.Timestamp != null"
```
ritalwar, line 27: This processor appears to be doing the same as what's done in the first date processor. Why are we handling it separately here instead of adding a condition in the original check?
```suggestion
    if: ctx.parsed?._aws != null && ctx.parsed?._aws?.Timestamp != null
```

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
```diff
@@ -0,0 +1,498 @@
...
+        for (int i = 0; i < ctx.parsed.stackTrace.length; i++) {
+            completionText.append(ctx.parsed.stackTrace[i]);
+            if (i != ctx.parsed.stackTrace.length - 1) {
+                completionText.append("\\n");
+            }
+        }
+        ctx.parsed.stack_trace_flattened = completionText.toString();
+
+- script:
+    description: "Join stackTrace array into single multiline string"
```
ritalwar: Is there a reason we're keeping two separate scripts for flattening the stackTrace? Could we simplify this with one script that checks both paths conditionally?

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
```diff
@@ -0,0 +1,498 @@
...
+        def completionText = new StringBuilder();
+        for (int i = 0; i < ctx.parsed.message.stackTrace.length; i++) {
+            completionText.append(ctx.parsed.message.stackTrace[i]);
+            if (i != ctx.parsed.message.stackTrace.length - 1) {
+                completionText.append("\\n");
+            }
+        }
+        ctx.parsed.message.stack_trace_flattened = completionText.toString();
+
+- rename:
```
ritalwar, line 89: Instead of renaming each field individually, can we use a script to handle the renaming more efficiently? I’ve noticed such scripts being used across multiple data streams in different packages. You can refer to some of the data streams in the o365_metrics package for examples.

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,48 @@
...
+        if: 'ctx.event?.original == null'
+        description: 'Renames the original `message` field to `event.original` to store a copy of the original message. The `event.original` field is not touched if the document already has one; it may happen when Logstash sends the document.'
+    - remove:
+        field: message
+        ignore_missing: true
+        if: ctx.event?.original != null
+        description: 'The `message` field is no longer required if the document has an `event.original` field.'
+    - set:
+        field: ecs.version
+        value: 8.11.0
```
ritalwar, line 16: Could we update this to the latest ECS version?

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
agithomas: Addressed as suggested.

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
agithomas: Addressed as suggested. Marking as resolved.

packages/aws/_dev/build/docs/lambda.md
```diff
@@ -54,6 +57,21 @@ When you configure the AWS integration, you can collect data from as many AWS se
 For step-by-step instructions on how to set up an integration, see the
 {{ url "getting-started-observability" "Getting started" }} guide.
 
+To enable AWS Lambda logs, ensure that your function's execution role includes the necessary permissions to write to Amazon CloudWatch Logs. Specifically, the role should have the following permissions: 
+- `logs:CreateLogGroup`
+- `logs:CreateLogStream`
+- `logs:PutLogEvents`.
```
Kavindu-Dodan: nit :) 

```suggestion
- `logs:PutLogEvents`
```

packages/aws/data_stream/lambda_logs/_dev/test/pipeline/test-awslambda-platform-json.log
```diff
@@ -0,0 +1,14 @@
+{"time":"2025-05-05T04:01:29.886Z","type":"platform.start","record":{"requestId":"01ad0e05-391e-4507-9233-ae7ad823dbc1","functionArn":"arn:aws:lambda:ap-south-1:627286350134:function:sam-app-powertools-python-HelloWorldFunction-snAkfaCNYJjx","version":"$LATEST","tracing":{"spanId":"4a16401155d9ea8d","type":"X-Amzn-Trace-Id","value":"Root=1-68183818-4eab0e4002398a2f375bd11a;Parent=03c383535c3c289b;Sampled=1"}}}
```
Kavindu-Dodan: nit : let's use dummy account ID similar to `123456789012`. Please see other occurrences.

Issue #13647, opened 2025-04-23
[meta] Update Rapid7 integration to Leverage Native Cloud Security Workflows


kcreddy:
As part of effort to leverage Cloud Security workflows such as Elastic CSPM and CNVM for 3rd party integrations, the vulnerabilities data from Rapid7 needs to be enriched just like previous enhancements for Wiz, AWS Security Hub, and Qualys VMDR.

For this work, the `rapid7_insightvm.vulnerability` data stream which ingests exported vulnerabilities of assets must be enriched to support Elastic CNVM workflow.

Tasks:

- [x] Get access to Tenable IO environment.
- [x] Setup Rapid7 data ingestion.
- [x] Analyse mappings for Rapid7 and get necessary clarifications.
- [ ] Implement mapping changes
- [ ] Implement transform

kcreddy:
Data ingestion is setup and mapping analysis and [discussion](https://github.com/elastic/security-team/issues/12534) is underway.

kcreddy:
Closing this meta issue as following has been implemented:

- Setup Rapid7 Insight VM data ingestion in SEI cluster.
- Approval of proposed mapping changes.
- Add new data stream `asset_vulnerability` with required mappings and transform for CDR workflow: https://github.com/elastic/integrations/pull/14079

andrewkroh:
> Closing this meta issue as following has been implemented:
> - Setup Rapid7 Insight VM data ingestion in SEI cluster.

@kcreddy Which cluster is this? I didn't see it in the place I expected.

kcreddy:
> > Closing this meta issue as following has been implemented:
> > 
> > * Setup Rapid7 Insight VM data ingestion in SEI cluster.
> 
> [@kcreddy](https://github.com/kcreddy) Which cluster is this? I didn't see it in the place I expected.

@andrewkroh, yes this wasn't setup in our cluster because the API Key was procured by Crest.
We tried to reach out to Rapid7 team: https://github.com/elastic/security-team/issues/12534#issuecomment-2901596037 to get our own license so it can be deployed into SEI demo cluster, but couldn't get the credentials/license yet.

cc: @cpascale43

PR #13372, opened 2025-04-01
Add new AWS Security Hub Findings Full Posture data stream and update misconfig transform to use it

maxcold:
## Proposed commit message

Similar to https://github.com/elastic/integrations/pull/12961, adding a new `AWS Security Hub Findings Full Posture` data stream, which in contrast to the existing `AWS Security Hub Findings` data stream, ingests all Security Hub Findings data every 24h to match the logic of the native Cloud Security Posture integration used by `cloudbeat`. Updated latest misconfigurations transform to use this new data stream as a source. Due to the change in the transform, this might be a breaking change for customers, they need to enable the new data stream when updating to still receive findings in their Findings > Misconfigurations view and other flows relying on cloud security posture data

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [x] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## How to test this PR locally

1. build package with `elastic-package build`
2. start the stack corresponding to the Kibana version from the manifest with `elastic-package stack up`
3. Install the integration and ingest AWS Security Hub Findings Full Posture data, findings from AWS should appear in Findings > Misconfiguration

## Related issues

Related to:
- https://github.com/elastic/kibana/issues/208933
- https://github.com/elastic/integrations/pull/12961

PR #14079, opened 2025-05-30
[rapid7_insightvm] Add asset_vulnerability data stream for Cloud Detection and Response (CDR) workflow

### LINKED ISSUE #9354: Change Rapid7 InsightVM integration to one doc per vulnerability
Similar to our recent update to[ Qualys VMDR update](https://github.com/elastic/integrations/issues/9261), we have similar feedback for our Rapid7 InisghtVM integration. Can we make a similar change to the Rapid7 integration, based on the user feedback as follows:

In our pursuit of refining our security posture through enhanced data analysis, the integration between Elastic and Rapid7 has surfaced areas ripe for optimization, particularly concerning the management of vulnerabilities data. Presently, the nightly indexing process for vulnerabilities not only encompasses all modified vulnerabilities but also tends to re-index vulnerabilities indiscriminately. This includes a broad spectrum of vulnerabilities scanned by Rapid7, rather than focusing solely on those relevant to our specific environment. Such an approach results in considerable data redundancy and inefficiency, straining our resources.

To address this, we propose the introduction of a new data category within Elastic, designated as 'state-*'. This category would advocate for the use of monolithic indices that leverage non-random document identifiers (not based on a timestamp), such as the unique ID of each vulnerability. Implementing this change would ensure that each vulnerability is represented by a single document within the dataset, markedly reducing the volume of data processed and stored daily. This adjustment would not only streamline the handling of vulnerability data but also significantly enhance the utility and performance of Rapid7 integrations within the Elastic ecosystem. We are keen to discuss this proposal further and explore the possibility of its implementation, hopeful that Elastic sees the value in optimizing the management of vulnerability data for mutual benefit.

### LINKED ISSUE #13775: Rapid7 Insight VM: Implement mappings for Cloud Security Workflows
Please see https://github.com/elastic/integrations/issues/13647 for more details

### LINKED ISSUE #13776: Implement transform

brijesh-elastic:
## Proposed commit message

```
rapid7_insightvm: Add asset_vulnerability data stream for Cloud Detection and Response (CDR) workflow

This adds a breaking change, as it involves deprecating asset data stream using HTTPJSON input and
adding new data_stream asset_vulnerability using CEL input.

The new data stream, asset_vulnerability, will retrieve a list of assets and the vulnerabilities contained within each.
Additionally, it will enrich the vulnerability details by querying vulnerability endpoint.
The agent will publish one document per vulnerability per host.

ECS mapping and transforms have also been added to facilitate
the Cloud Native Vulnerability Management (CNVM)[1] workflow.

Enabled agentless deployment for rapid7 insightvm
Upgraded the format_version to 3.3.5
Updated Kibana version constraints to ^8.19.0 || ^9.1.0

[1] https://www.elastic.co/guide/en/security/current/vuln-management-overview.html
```

> [!NOTE]  
> To Reviewers:
>  - Please refer to this guide: https://docs.elastic.dev/security-solution/cloud-security/cdr/3p-dev-guide which was used for proposed changes.

## Checklist

- Clone integrations repo.
- Move to integrations/packages/rapid7_insightvm directory.
> elastic-package test

## Related issues

- Closes #9354 
- Closes #13775 
- Closes #13776 
- Relates #13647


packages/rapid7_insightvm/manifest.yml
```diff
@@ -11,7 +11,7 @@ categories:
   - vulnerability_management
 conditions:
   kibana:
-    version: "^8.13.0 || ^9.0.0"
+    version: "^8.19.0 || ^9.0.0"
```
kcreddy: @maxcold, this is upgraded to support `8.19.0` Kibana UI changes similar to Tenable.

packages/rapid7_insightvm/elasticsearch/transform/latest_cdr_vulnerabilities/transform.yml
```diff
@@ -0,0 +1,31 @@
...
+retention_policy:
+  time:
+    field: "@timestamp"
+    max_age: 90d
+_meta:
+  managed: true
+  # Bump this version to delete, reinstall, and restart the transform during package.
+  # Version bump is needed if there is any code change in transform.
+  fleet_transform_version: 0.2.0
+  run_as_kibana_system: false
```
kcreddy:
```suggestion
```

This line will be removed once backport PRs for adding permissions are merged: https://github.com/elastic/elasticsearch/pull/128873. 

cc: @maxcold

packages/rapid7_insightvm/data_stream/asset_vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1111 @@
...
+  - set:
+      field: vulnerability.classification
+      tag: set_vulnerability_classification
+      value: CVSS
+  - set:
+      field: vulnerability.enumeration
+      tag: set_vulnerability_enumeration
+      value: CVE
+  # Remove cloud.* fields populated by beat.
+  # These fields correspond to EA rather than Tenable hosts and could be misleading.
```
maxcold: nit: comment mentions Tenable, should be Rapid7 I guess

packages/rapid7_insightvm/manifest.yml
maxcold: @kcreddy we need to update it to be `^8.19.0 || ^9.1.0` as 9.0.0 version doesn't contain the support for multiple/empty CVEs

packages/rapid7_insightvm/data_stream/asset_vulnerability/manifest.yml
```diff
@@ -0,0 +1,78 @@
+title: Asset Vulnerability Event
+type: logs
+streams:
+  - input: cel
+    title: Asset Vulnerability Event
+    description: Collecting asset vulnerability events via API.
+    template_path: cel.yml.hbs
+    enabled: false
+    vars:
+      - name: interval
```
maxcold, line 18: There is no initial interval as far as I can see. Do we do full scan periodically? In that case we need to set the @timestamp of the documents to the ingestion time, like we did for AWS SecurityHub full posture data stream https://github.com/elastic/integrations/pull/13372/files#diff-2aebbf76c4f4587a0d3701264a03075f4886df078bb04d4b41e40ec329902488R592 This way we won't have the problem of old documents outside of the retention period

brijesh-elastic: For Rapid7, it will collect all the vulnerabilities present in all the assets during the first interval. After subsequent intervals, it will collect only the new and remediated vulnerabilities that have introduced since the last interval.

cc: @kcreddy

packages/rapid7_insightvm/data_stream/asset_vulnerability/elasticsearch/ingest_pipeline/default.yml
kcreddy: Addressed in [9bd0b7b](https://github.com/elastic/integrations/pull/14079/commits/9bd0b7bfaf827e84201df8967f3194bccc291a79)

packages/rapid7_insightvm/manifest.yml
kcreddy: Updated in [9bd0b7b](https://github.com/elastic/integrations/pull/14079/commits/9bd0b7bfaf827e84201df8967f3194bccc291a79)

@brijesh-elastic, please update the commit message accordingly.

packages/rapid7_insightvm/data_stream/asset_vulnerability/manifest.yml
kcreddy: @maxcold, we are going with incremental load to keep lower memory footprint as @brijesh-elastic observed agentless agent restarts in Serverless environments.

But I see your point. We need to match the transform retention with `initial_interval`, otherwise even if we ingest 2 years worth of data if our transform retention is much lower, all the documents are deleted inside destination indices. Is this correct understanding?
If this is the case, we can add `initial_interval` like with other incremental integrations.

maxcold: Yes, if we go with incremental I think we need to follow other integraitons approach and have initial_interval. It will at least give users the option to ingest some initial data for the CDR flows. But I'm not sure I understand how do we then have the data from 2023 in the envs if you do incremental updates only?

kcreddy: > But I'm not sure I understand how do we then have the data from 2023 in the envs if you do incremental updates only?

In the absence of `initial_interval`, the first fetch ingests all historical data. All fetches after this will be incremental.

@brijesh-elastic is already working on adding `initial_interval` option, so that first fetch only has user-defined initial interval data instead of ingesting all historical data.

packages/rapid7_insightvm/elasticsearch/transform/latest_cdr_vulnerabilities/transform.yml
maxcold: @kcreddy can be removed already I think

kcreddy: Yes, I had some CI issues but all the backport PRs https://github.com/elastic/elasticsearch/pull/128661#issuecomment-2938637328 have been successfully merged now. 

@brijesh-elastic, please remove this.

packages/rapid7_insightvm/_dev/build/docs/README.md
```diff
@@ -26,6 +36,27 @@ This module uses **InsightVM Cloud Integrations API v4**.
...
+
+Version `2.0.0` of the Rapid7 InsightVM integration adds support for [Elastic Cloud Security workflow](https://www.elastic.co/docs/solutions/security/cloud/ingest-third-party-cloud-security-data#_ingest_third_party_security_posture_and_vulnerability_data). The enhancement enables the users of Rapid7 InsightVM integration to ingest their enriched asset vulnerabilities from Rapid7 InsightVM platform into Elastic and get insights directly from Elastic [Vulnerability Findings page](https://www.elastic.co/docs/solutions/security/cloud/findings-page-3).
+This update adds [Elastic Latest Transform](https://www.elastic.co/docs/explore-analyze/transforms/transform-overview#latest-transform-overview) which copies the latest vulnerability findings from source indices matching the pattern `logs-rapid7_insightvm.asset_vulnerability-*` into new destination indices matching the pattern `security_solution-rapid7_insightvm.vulnerability_latest-*`. The Elastic Vulnerability Findings page will display vulnerabilities based on the destination indices.
+
+For existing users of Rapid7 InsightVM integration, before upgrading to `4.0.0` please ensure following requirements are met:
```
chemamartinez:
```suggestion
For existing users of Rapid7 InsightVM integration, before upgrading to `2.0.0` please ensure following requirements are met:
```

packages/rapid7_insightvm/_dev/build/docs/README.md
```diff
@@ -10,13 +10,23 @@ Use the Rapid7 InsightVM integration to collect and parse data from the REST API
 
 The Rapid7 InsightVM integration collects two type of events: Asset and Vulnerability.
 
-**Asset** is used to get details related to inventory, assessment, and summary details of assets that the user has access to. See more details in the API documentation [here](https://help.rapid7.com/insightvm/en-us/api/integrations.html#operation/searchIntegrationAssets).
+**Asset (Deprecated)** is used to get details related to inventory, assessment, and summary details of assets that the user has access to. See more details in the API documentation [here](https://help.rapid7.com/insightvm/en-us/api/integrations.html#operation/searchIntegrationAssets). It is deprecated in version `2.0.0`. Instead, use the `Asset Vulnerability` data stream for enriched vulnerability documents and improved mappings.
 
 **Vulnerability** is used to retrieve all vulnerabilities that can be assessed. See more details in the API documentation [here](https://help.rapid7.com/insightvm/en-us/api/integrations.html#operation/searchIntegrationVulnerabilities).
 
+**Asset Vulnerability** is used to gather and aggregate data on assets and vulnerabilities to support Native CDR Workflows.
```
chemamartinez, line 15: Not too important, but I would move this above **Vulnerability** so the data streams description have the same order as they have in the **Logs Reference** section at the end of the documentation.

packages/rapid7_insightvm/_dev/build/docs/README.md
kcreddy:
```suggestion
For existing users of Rapid7 InsightVM integration, before upgrading to `2.0.0` please ensure following requirements are met:
```

packages/rapid7_insightvm/data_stream/asset/manifest.yml
```diff
@@ -1,8 +1,8 @@
-title: Collect Asset logs from Rapid7 InsightVM
+title: Collect Asset logs from Rapid7 InsightVM (Deprecated)
 type: logs
 streams:
   - input: httpjson
-    title: Asset logs
+    title: Asset logs (Deprecated)
```
chemamartinez, line 5: I'd add `enabled: false` for this as it will be deprecated. It won't affect upgrades when users have already enabled it but will appear disabled by default for new installations.

packages/rapid7_insightvm/data_stream/asset/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -15,9 +15,17 @@ processors:
...
       target_field: event.original
       ignore_missing: true
+      description: Renames the original `message` field to `event.original` to store a copy of the original message. The `event.original` field is not touched if the document already has one; it may happen when Logstash sends the document.
       if: ctx.event?.original == null
+  - remove:
+      field: message
+      tag: remove_message
+      ignore_missing: true
+      description: The `message` field is no longer required if the document has an `event.original` field.
+      if: ctx.event?.original != null
```
kcreddy, line 39: Also add this to list of `breaking-change`. Users could be using this field inside their `@custom` pipelines.

packages/rapid7_insightvm/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.0.0"
+  changes:
+    - description: Add asset_vulnerability datastream support for Cloud Detection and Response (CDR) vulnerability workflow.
+      type: enhancement
```
kcreddy:
```suggestion
      type: breaking-change
```

packages/rapid7_insightvm/data_stream/asset_vulnerability/manifest.yml
```diff
@@ -0,0 +1,86 @@
+title: Asset Vulnerability Event
+type: logs
+streams:
+  - input: cel
+    title: Asset Vulnerability Event
+    description: Collecting asset vulnerability events via API.
```
kcreddy:
```suggestion
    description: Collect enriched asset vulnerability events via API.
```

packages/rapid7_insightvm/manifest.yml
```diff
@@ -32,10 +32,22 @@ policy_templates:
...
     title: Rapid7 InsightVM logs
     description: Collect Rapid7 InsightVM logs.
+    deployment_modes: 
+      default:
+        enabled: true 
+      agentless:
+        enabled: true
+        organization: security
+        division: engineering
+        team: security-service-integrations
```
kcreddy, line 42: @chemamartinez, we are waiting on approval of https://github.com/elastic/integrations/pull/14172 to apply the approved fix here as well.

cc: @brijesh-elastic

packages/rapid7_insightvm/manifest.yml
```diff
@@ -32,10 +32,22 @@ policy_templates:
   - name: rapid7_insightvm
     title: Rapid7 InsightVM logs
     description: Collect Rapid7 InsightVM logs.
+    deployment_modes: 
+      default:
+        enabled: true 
+      agentless:
```
chemamartinez, line 38: As this integration will support Agentless from now on, it would be worthy to add the same change as https://github.com/elastic/integrations/pull/14172 for all ingest pipelines.

packages/rapid7_insightvm/data_stream/asset_vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,240 @@
...
+        (has(state.vulnerabilities) && state.is_current_vulnerabilities_fetched) ?
+          work.with({
+            "vulnerabilities": state.vulnerabilities,
+            "is_current_vulnerabilities_fetched": state.is_current_vulnerabilities_fetched
+          })
+        :
+          request(
+            "POST",
+            state.url.trim_right("/") + "/vm/v4/integration/vulnerabilities?" + {
+              "size": ["500"],
```
chemamartinez, line 134: Why is this hardcoded instead of `batch_size`?

packages/rapid7_insightvm/data_stream/asset_vulnerability/_dev/test/pipeline/test-asset-vulnerability.log-expected.json
```diff
@@ -0,0 +1,5116 @@
...
+                },
+                "type": "guest"
+            },
+            "observer": {
+                "product": "Rapid7 InsightVM",
+                "vendor": "Rapid7"
+            },
+            "package": {
+                "fixed_version": "latest",
+                "name": "Ubuntu Linux 22.04",
```
chemamartinez: Seems like this should be split into:
```
"name": "Ubuntu Linux",
"version": "22.04"
```

packages/rapid7_insightvm/changelog.yml
chemamartinez: Wondering why is this a breaking change as main changes are in a new data stream.

packages/rapid7_insightvm/manifest.yml
chemamartinez: Thanks, I added the same comment before seeing this one.

packages/rapid7_insightvm/data_stream/asset_vulnerability/agent/stream/cel.yml.hbs
brijesh-elastic: It was hardcoded to 500 (which is the maximum for the endpoint) because it is already a lower number for a very high number of vulnerabilities present in instance.

packages/rapid7_insightvm/data_stream/asset_vulnerability/_dev/test/pipeline/test-asset-vulnerability.log-expected.json
brijesh-elastic: We don't have the fix pattern for this to parse. Some examples are here from the live logs:
`Apache HTTPD`, `Microsoft Windows Server 2008 R2`, `Standard Edition SP1`, `FreeBSD 8.4-RELEASE`, `Mozilla Firefox 10.0.2 (C:\Program Files (x86)\Mozilla Firefox)`, `VMware ESXi Server 6.5.0 Update 2`, `Linux 2.6.23 - 2.6.38`, `Microsoft .NET Framework 3.5 SP1`, `Microsoft Internet Explorer 11.0.9600.19540`, `Linux 2.6.24 - 2.6.26 (Debian)`, `Linux 2.6.18 (Debian 4.0, x86)`, `Linux 2.6.15 - 2.6.26 (likely embedded)`

chemamartinez: Yeah, most of cases it could be split by looking at the first number but you are right it doesn't apply to all the cases.

efd6: Is this the value you that you want though? The asset vulnerability is not in the version of linux in this case (if it were we could get them from `rapid7_insightvm.asset_vulnerability.os.{system_name,version}`), it's in Oracle Java SE. There is no machine parsable version information in the CVE that describes the version of the package that is vulnerable.

In all of the cases in this test set, the values of these fields refer to the operating system, not the vulnerable package (none of the values refer to a package at all — and in none of the cases can I see any indication of the affected version of the package/software in any field other than the description). So filling these fields with these values is misinformative. In some cases below, the value of the `fixed_version` field is "latest". I don't think this is valuable since latest has no reference (although I guess you could get some notion of what it refers to by calculating a date from the timestamp).

All up, I'm not convinced that these fields should be filled.

packages/rapid7_insightvm/data_stream/asset_vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,240 @@
...
+        resp.Body.decode_json().as(body, {
+          "events": [{"message": "retry"}],
+          "batch_size": state.batch_size,
+          "initial_interval": state.initial_interval,
+          "api_key": state.api_key,
+          "cursor": {
+            ?"last_interval_time": state.?cursor.last_interval_time,
+          },
+          ?"next_asset_cursor": has(body.?metadata.cursor) && body.metadata.cursor != null ? optional.of(body.metadata.cursor) : optional.none(),
+          "is_all_assets_fetched": has(body.metadata.cursor) && body.metadata.cursor != null ? false : true,
```
efd6:
```suggestion
          "is_all_assets_fetched": !has(body.metadata.cursor) || body.metadata.cursor == null,
```
In general a ternary should never have its final evaluation type be a boolean given that its input is a boolean and so can be trivially mapped to the output without it.

packages/rapid7_insightvm/data_stream/asset_vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,240 @@
...
+          "assets": body.data,
+          "asset_vuln_ids": body.data.map(a,
+            a.?same.orValue([]).map(s, s.vulnerability_id) + a.?new.orValue([]).map(n, n.vulnerability_id) + a.?remediated.orValue([]).map(r, r.vulnerability_id)
+          ).flatten().as(vuln_ids,
+            vuln_ids.map(vuln_id, string(vuln_id)).as(str_vuln_ids, zip(str_vuln_ids, vuln_ids))
+          ).keys(),
```
efd6: This is intended to get a unique set of IDs. Is that correct? The order of operations here is not unambiguous to the reader, so I would like this clarified. I suspect that what you want is `(same + new + remediated).flatten`. On the basis of this assumption, I would prefer the following
```
body.data.map(a, (
  a.?same.orValue([]).map(s, s.id) +
  a.?new.orValue([]).map(n, n.id) +
  a.?remediated.orValue([]).map(r, r.id)
)).flatten().as(vuln_ids,
  zip(vuln_ids.as(id, string(id)), vuln_ids)
).keys()
```
If the IDs are always strings (it looks like they are), we can remove the `vuln_ids.as(id, string(id))` and replace it with `vuln_ids` saving allocations.

packages/rapid7_insightvm/data_stream/asset_vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,240 @@
...
+resource.url: {{url}}
+state:
+  api_key: {{api_key}}
+  batch_size: {{batch_size}}
+  initial_interval: {{initial_interval}}
+redact:
+  fields:
+    - api_key
+program: |
+  (
```
efd6, line 40: I'd like block comments in this program; it's large enough that there is significant state that needs to be understood when maintaining it.

packages/rapid7_insightvm/data_stream/asset/elasticsearch/ingest_pipeline/default.yml
brijesh-elastic: We're adding this to every integration, and it has also been included in the `elastic-package` tool validation. If we don't add it, it will result in an error during the `elastic-package build`. Should we still add this to the `changelog.yml`?

packages/rapid7_insightvm/data_stream/asset_vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,262 @@
...
+resource.url: {{url}}
+state:
+  api_key: {{api_key}}
+  batch_size: {{batch_size}}
+  initial_interval: {{initial_interval}}
+redact:
+  fields:
+    - api_key
+program: |
+
```
efd6:
```suggestion
program: |-
```

packages/rapid7_insightvm/data_stream/asset_vulnerability/_dev/test/pipeline/test-asset-vulnerability.log-expected.json
```diff
@@ -0,0 +1,5116 @@
...
+                "type": "guest"
+            },
+            "observer": {
+                "product": "Rapid7 InsightVM",
+                "vendor": "Rapid7"
+            },
+            "package": {
+                "name": "CentOS Linux 7.6.1810",
+                "version": "CentOS Linux 7.6.1810"
+            },
```
efd6: I don't think this is correct here; the package is MySQL (actually MariaDB).

packages/rapid7_insightvm/data_stream/asset_vulnerability/_dev/test/pipeline/test-asset-vulnerability.log-expected.json
```diff
@@ -0,0 +1,5116 @@
...
+                                "skill_level": "intermediate",
+                                "source": "metasploit"
+                            },
+                            {
+                                "description": "Microsoft Windows - 'srv2.sys' SMB Code Execution (Python) (MS09-050)",
+                                "id": "40280",
+                                "name": "Microsoft Windows - 'srv2.sys' SMB Code Execution (Python) (MS09-050)",
+                                "rank": "average",
+                                "skill_level": "expert",
+
```
efd6: I don't think this is right; the vulnerability is in BIND. Also the "latest" version value is not helpful.

efd6: I have no idea why GH put this comment here. It should be [here](https://github.com/elastic/integrations/pull/14079/files#diff-8a764ec3d41594bfc2803bca22007e5ec39381fd6492fe5a835938e0d4e1a4deR3696-R3700).

packages/rapid7_insightvm/changelog.yml
kcreddy: @chemamartinez, the breaking changes comes from the prerequisites of transforms and data duplication.
https://github.com/elastic/integrations/blob/471e7fe155e6dc0f6f4f8141202a0a50d5720ede/packages/rapid7_insightvm/_dev/build/docs/README.md?plain=1#L39-L58

It important to document users needing transform nodes to run a transform and also heads up on the data duplication to avoid unexpected costs.

packages/rapid7_insightvm/data_stream/asset_vulnerability/_dev/test/pipeline/test-asset-vulnerability.log-expected.json
kcreddy: `package.*` fields are categorised as `MUST HAVE` for the vulnerability CDR workflow, meaning without these fields the UI will break. The values needed for these fields are not directly provided by the API. Hence a best-effort is made to populate them from information within the vulnerabilities array.
Full analysis [here](https://github.com/elastic/security-team/issues/12534#issuecomment-2889800357) and PM approval followed by it.

kcreddy: > although I guess you could get some notion of what it refers to by calculating a date from the timestamp.

Yeah, that was the idea. Since populating a fix version helps analysts know if a fix is yet available or not.

efd6: > `package.*` fields are catego…

I'm not convinced, but OK.

chemamartinez (approved):
Regarding comments I added, LGTM.

packages/rapid7_insightvm/data_stream/asset/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -13,11 +13,30 @@ processors:
...
+      field:
+        - organization
+        - division
+        - team
+      ignore_missing: true
+      if: ctx.organization instanceof String && ctx.division instanceof String && ctx.team instanceof String
+      tag: remove_agentless_tags
+      description: >-
+        Removes the fields added by Agentless as metadata,
+        as they can collide with ECS fields.
```
kcreddy, line 26: Can you also add a changelog entry for this change (bugfix) similar to https://github.com/elastic/integrations/pull/14172

kcreddy (approved):
nit only. LGTM 👍🏼

packages/rapid7_insightvm/data_stream/asset_vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,262 @@
...
+      ).with({
+        "Header": {
+          "X-Api-Key": [state.api_key],
+          "Content-Type": ["application/json"],
+        },
+        "Body": {
+          ?"asset": has(state.?cursor.last_interval_time) ?
+            optional.none()
+          :
+            optional.of(("last_scan_end > " + string((timestamp(interval_time) - duration(state.initial_interval)).format(time_layout.RFC3339)))),
```
efd6:
```suggestion
            optional.of(("last_scan_end > " + (timestamp(interval_time) - duration(state.initial_interval)).format(time_layout.RFC3339))),
```
`timestamp.format` returns a string.

efd6 (approved):

packages/rapid7_insightvm/data_stream/asset_vulnerability/_dev/test/pipeline/test-asset-vulnerability.log-expected.json
maxcold: @kcreddy we discussed this with @nick-alayil yesterday and I think there is some confusion.
> Full analysis https://github.com/elastic/security-team/issues/12534#issuecomment-2889800357 and PM approval followed by it.

in this discussion we confirmed that for the lack of better source, parsing a string to get the package info (name and version) is acceptable. But I didn't find where we discussed having the OS instead of the vulnerable package in `package.*` fields
In the test data we have the following proof field `Vulnerable OS: Ubuntu Linux 22.04\n\n\n\nVulnerable software installed: Azul Systems JRE 17.54.22 (/root/infaagent/jdk/lib/jrt-fs.jar)`
which now converts into 
```
"package": {
                "fixed_version": "latest",
                "name": "Ubuntu Linux 22.04",
                "version": "Ubuntu Linux 22.04"
            },
```
but the vulnerable software information is available so it would have been better to have it as a package. Are there technical limitation to parsing out `Vulnerable software installed` or `Vulnerable package installed` and fallback to os only when there is no package/software information available?

kcreddy: @maxcold, thanks for this analysis. I don't think there's a technical limitation here. If `Vulnerable software installed` or `Vulnerable package installed` is present, it should take priority over OS. 

I believe the test data maybe old (from earlier commits) and this later commit from the PR: https://github.com/elastic/integrations/pull/14079/commits/132cb48a424e3ea89c8398f8e734fa0c385815da should've extracted it correctly. 

@brijesh-elastic can you please confirm this and also create new test data with latest changes?

brijesh-elastic: Yes, it's resolved in later commits. I've also created a new cluster with the latest changes.

PR #14095, opened 2025-05-30
[ti_threatconnect] Simplify API logic

### LINKED ISSUE #13336: ti_threatconnect: revise pagination and cursor logic in agent config
In #13235 an issue was fixed with a minimal change, but the logic that exists in cursor and pagination handling code is more complicated than it needs to be, and has been a source of confusion on a number of occasions. It should probably be revised to a simpler and more easily reasoned bout approach.

A proposed change was given in https://github.com/elastic/integrations/pull/13235#pullrequestreview-2716830914.

chrisberkhout:
## Proposed commit message

```
[ti_threatconnect] Simplify API logic

This simplifies the logic for requesting indicators as follows:

- Use the provided `next` URL instead of rebuilding it.
- Use `state.initial_interval` only once.
- Propagate only one timestamp.

The format of `state.cursor.last_timestamp` is unchanged, so it will
continue from an existing cursor position, if available.

This version has been manually tested against the live API.
```

## Checklist

## Related issues

- Closes #13336
- Relates #13235


efd6 (approved):
Very nice.

efd6:
Do we want to release this? (changelog and manifest entries) or just wait for a next release of the integration? I good either way, I also don't think that we need to have a changelog if we are waiting since it's implementation details that are not user-facing (which is an argument for waiting).

chrisberkhout:
> Do we want to release this? (changelog and manifest entries) or just wait for a next release of the integration? I good either way, I also don't think that we need to have a changelog if we are waiting since it's implementation details that are not user-facing (which is an argument for waiting).

I think release as an enhancement with a new patch version. If someone has a problem with current functionality I'd rather debug this version. If a later change introduces a problem, it'll be good to have some separation from this change.

PR #14103, opened 2025-06-02
[sysdig] Add support for vulnerability datastream

### LINKED ISSUE #12269: [Sysdig Secure] New data stream: Vulnerabilities
Add a new data stream to gather vulnerability information from Sysdig Secure. Read more [here](https://docs.sysdig.com/en/docs/sysdig-secure/vulnerabilities/).

brijesh-elastic:
## Proposed commit message

```
sysdig: add support for vulnerability datastream.

This vulnerability logs provides an overview of scan results from your
runtime, pipeline, and container registries to provide highly accurate views
of vulnerability risk, access to public exploits, and risk management.

Sanitized test case inputs were obtained from live Sysdig Secure instance
using the Sysdig Next Gen API.
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/sysdig directory.
> elastic-package test

## Related issues

- Closes #12269 

## Screenshots
vulnerability dashboard


packages/sysdig/data_stream/vulnerability/_dev/test/pipeline/test-common-config.yml
```diff
@@ -0,0 +1,3 @@
+fields:
+  tags:
+    - preserve_duplicate_custom_fields
```
chrisberkhout, line 3: Add this to fix the build:
```
dynamic_fields:
  "event.ingested": ".*"
```

packages/sysdig/_dev/build/docs/README.md
```diff
@@ -2,12 +2,14 @@
 This integration allows for the shipping of [Sysdig](https://sysdig.com/) logs to Elastic for security, observability and organizational awareness. Logs can then be analyzed by using either the dashboard included with the integration or via the creation of custom dashboards within Kibana.
 
 ## Data Streams
-The Sysdig integration collects two type of logs:
+The Sysdig integration collects three type of logs:
```
chrisberkhout:
```suggestion
The Sysdig integration collects three types of logs:
```

packages/sysdig/_dev/build/docs/README.md
```diff
@@ -2,12 +2,14 @@
...
 
 **Alerts** The Alerts data stream collected by the Sysdig integration is comprised of Sysdig Alerts. See more details about Sysdig Alerts in [Sysdig's Alerts Documentation](https://docs.sysdig.com/en/docs/sysdig-monitor/alerts/). A complete list of potential fields used by this integration can be found in the [Logs reference](#logs-reference)
 
 **Event** The event data stream collected through the Sysdig integration consists of Sysdig Security Events. See more details about Security Events in [Sysdig's Events Feed Documentation](https://docs.sysdig.com/en/docs/sysdig-secure/threats/activity/events-feed/).
 
+**Vulnerability** The vulnerability data stream collected through the Sysdig integration consists of Sysdig Vulnerability. See more details about Vulnerability in [Sysdig's Vulnerability Documentation](https://docs.sysdig.com/en/sysdig-secure/vulnerability-management/).
```
chrisberkhout:
```suggestion
**Vulnerability** The vulnerability data stream collected through the Sysdig integration consists of Sysdig vulnerability scan results. See more details about vulnerabilities in [Sysdig's Vulnerability Management documentation](https://docs.sysdig.com/en/sysdig-secure/vulnerability-management/).
```

packages/sysdig/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1516 @@
...
+      value: Sysdig
+  - set:
+      field: observer.product
+      tag: set_observer_product
+      value: Sysdig Secure
+  - rename:
+      field: json.assetType
+      tag: rename_assetType
+      target_field: sysdig.vulnerability.asset_type
+      ignore_missing: true
```
chrisberkhout: This ingest pipeline is 1516 lines now. I would like to see it shortened, as follows.

If the whole of `json` is renamed to `sysdig.vulnerability`, and a painless script is used to recursively rename all camel case fields, then many set/rename processors could be removed. There's an example script [here](https://github.com/elastic/integrations/blob/d5e75f590d27aabf588d050f055cc0dcf10e934e/packages/trend_micro_vision_one/data_stream/telemetry/elasticsearch/ingest_pipeline/default.yml#L47-L104).

Also, there are `date` processors below that only parse an ISO8601 date and put it in a new location. If the name change is handled by the camel to snake case script, such processors can be removed. An ISO8601 date/time can be passed directly to Elasticsearch as a value for a `date` field. They don't need to be converted first.

packages/sysdig/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1516 @@
+---
```
chrisberkhout, line 1: It would be helpful to use comments to break the pipeline up into sections: for:
- general processing
- extra processing of `sysdig.vulnerability.*` fields
- mapping to ECS

packages/sysdig/img/sysdig-vulnerability-screenshot.png
chrisberkhout, line 1: sysdig-vulnerability-screenshot-annotated

Regarding the global filters point, please ask to see the "SEI Integration Dashboard styling guide" if you haven't already.

A briefer overview could be like this:

> This dashboard provides an overview of the vulnerability data collected by the Sysdig integration.
> 
> It highlights trends in critical CVEs over time and shows counts of critical and high vulnerabilities.
> 
> Charts showing most affected namespaces and most frequent specific vulnerabilities can help to prioritize remediation efforts.
> 
> Select a suitable time range to avoid seeing partial results.

packages/sysdig/kibana/search/sysdig-81a3c734-23e4-4acb-83dd-ae05dd98536c.json
chrisberkhout, line 1: The change in this file can be removed from the PR.

packages/sysdig/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1516 @@
...
+  - convert:
+      field: json.package.vulnerability.cvssScore.score
+      tag: convert_package_vulnerability_cvssScore_score_to_double
+      target_field: sysdig.vulnerability.package.vulnerability.cvss_score.score
+      type: double
+      ignore_missing: true
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```
chrisberkhout: It looks like the input is already a double. Can conversions like this be removed?

packages/sysdig/data_stream/vulnerability/manifest.yml
```diff
@@ -0,0 +1,101 @@
...
+        default: registry
+        options:
+          - value: pipeline
+            text: Pipeline
+          - value: registry
+            text: Registry
+          - value: runtime
+            text: Runtime
+        required: true
+        show_user: true
```
chrisberkhout: Does the default of `registry` give us everything we want here?

It sounds like the same vulnerability information goes through multiple stages and we can choose to collect it earlier or later. Presumably it would have more information at later stages. Is this correct? Or would we ever want to collect from multiple stages.

I haven't been able to access the API documentation yet, but I think it would be good to have a bit stronger guidance in the description of this variable. If we choose the best stage by default and the others are for obscure use cases, it may be worth moving this to the advanced settings section (`show_user: false`).

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,214 @@
...
+                  "message": {
+                    "assetType": body.assetType,
+                    "metadata": body.metadata,
+                    "producer": body.producer,
+                    "stage": body.stage
+                  }.with(work.worklist.data[int(work.index)]).encode_json()
+                }]
+              :
+                result
+            ),
```
chrisberkhout: I think it would better to do all of this restructuring in Painless in the ingest pipeline, rather than in CEL.

It's easier if the events output by CEL that go into the ingest pipeline are very close to what the API returns. Then in the future, maintainers can work with just the pipeline and the API documentation, rather than having to go back to the CEL.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,214 @@
...
+            "batch_size": state.batch_size,
+            "api_token": state.api_token,
+            "stage": state.stage,
+            ?"worklist": int(work.index) + 1 < size(work.worklist.data) ? optional.of(work.worklist) : optional.none(),
+            ?"index": int(work.index) + 1 < size(work.worklist.data) ? optional.of(int(work.index) + 1) : optional.none(),
```
chrisberkhout: It's good to have clear separation of items ready to process, and other information for listing items. Currently, it's mixed together, with `state.worklist.data[0].resultId` being the first item ID to fetch, but `state.worklist.page.next` being information required for paging through listings of items.

Instead of using an index to go through a list of work items, a simpler pattern is to remove an item from the list, process that and return the remaining list.

Here's an example of:
[checking for a work item](https://github.com/elastic/integrations/blob/d5e75f590d27aabf588d050f055cc0dcf10e934e/packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs#L85)
[processing the first item](https://github.com/elastic/integrations/blob/d5e75f590d27aabf588d050f055cc0dcf10e934e/packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs#L93-L94)
[returning the rest of the work items](https://github.com/elastic/integrations/blob/d5e75f590d27aabf588d050f055cc0dcf10e934e/packages/trend_micro_vision_one/data_stream/telemetry/agent/stream/cel.yml.hbs#L110-L112)

The `tail()` function requires stack version 8.15.0 or greater.

Usually it's easier to process ready items first, then only fetch a list of new items when all the existing items have been processed.

chrisberkhout (changes requested):
Lots of good work. I think that some significant simplifications can be made before we merge it.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
brijesh-elastic: I agree that we should avoid making significant schema changes in CEL. However, in this case, we need to split and publish individual events from a single raw log, which is only feasible with CEL.

We could handle the splitting in CEL and then perform enriching and pre-processing (as explained [here](https://github.com/elastic/integrations/issues/12269#issuecomment-2820653565)) in the ingest pipeline. However, this approach would require sending a large amount of data with each event.

I favor using CEL for this purpose, but let me know your thoughts.

packages/sysdig/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
brijesh-elastic: It is there for safety purposes. If any other value type appears in the future, it will ensure that the field is converted to a double if possible, or it will remove that field.

If we eliminate this type checking, it could lead to a mapping conflict in Kibana in the future.

packages/sysdig/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
brijesh-elastic: If we eliminate this type checking, it could lead to a mapping conflict in Kibana in the future if other than ISO8601 format appears.

packages/sysdig/data_stream/vulnerability/manifest.yml
brijesh-elastic: Vulnerabilities will be collected at various stages, and they will differ for each stage. ([Documentation here](https://docs.sysdig.com/en/sysdig-secure/vulnerability-management/#key-concepts)) Also, we should remove the default value of stage, as it is not guaranteed that the final stage will contain all vulnerabilities.

packages/sysdig/data_stream/vulnerability/_dev/test/pipeline/test-common-config.yml
chrisberkhout: I see now `event.ingested` is populated by ES, not explicitly set in the pipeline. That's a better, and the dynamic fields config isn't necessary because it doesn't get set during the pipeline test run.

packages/sysdig/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: I see that reduced the length by more than a third. Thanks!

---

Regarding the date conversions...

The default `format` setting for a `date` field is `strict_date_optional_time||epoch_millis` ([ref](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/date#date-params)).

A document that doesn't match that will fail to be ingested.

So the current version is more robust. If the incoming date format changes, the value will be removed and the document will be indexed without it. To me it seems like it's low-value processing unless we have reason to believe the format is not stable.

Is this a documented recommendation, or something based on your experience? (We can go either way on this, I'm just curious)

packages/sysdig/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Thanks!

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
chrisberkhout: I see you're using the tail function. That's good.

packages/sysdig/_dev/build/docs/README.md
```diff
@@ -2,12 +2,14 @@
...
 
 ## Data Streams
-The Sysdig integration collects two type of logs:
+The Sysdig integration collects three types of logs:
 
 **Alerts** The Alerts data stream collected by the Sysdig integration is comprised of Sysdig Alerts. See more details about Sysdig Alerts in [Sysdig's Alerts Documentation](https://docs.sysdig.com/en/docs/sysdig-monitor/alerts/). A complete list of potential fields used by this integration can be found in the [Logs reference](#logs-reference)
 
 **Event** The event data stream collected through the Sysdig integration consists of Sysdig Security Events. See more details about Security Events in [Sysdig's Events Feed Documentation](https://docs.sysdig.com/en/docs/sysdig-secure/threats/activity/events-feed/).
 
+**Vulnerability** The vulnerability data stream collected through the Sysdig integration consists of Sysdig vulnerability scan results. See more details about vulnerabilities in [Sysdig's Vulnerability Management documentation](https://docs.sysdig.com/en/sysdig-secure/vulnerability-management/).
```
chrisberkhout, line 11: Either here or further down there are some important things we should say about the vulnerability data:

- On each interval we fetch all available scan results for the configured stage.
- You can only fetch scan results from one of the three stages.
- When we fetch the scan results, we break them into separate events for each package-vulnerability pair (or a package alone if it has no vulnerability, or just the scan if it has no packages).

packages/sysdig/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Resolving this one. It's the same issue as with the dates. I can see some value either way.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
chrisberkhout: Thanks, I caught up on the discussion in the issue about splitting. I agree that makes sense when we have huge responses like this.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,207 @@
...
+                    "message": {
+                      "assetType": body.assetType,
+                      "metadata": body.metadata,
+                      "producer": body.producer,
+                      "stage": body.stage
+                    }.with(worklist.data[0]).encode_json()
+                  }]
+                :
+                  dyn(result)
+              ),
```
chrisberkhout: The split logic here is important.

I had an idea for doing the splitting without changing the document structure.
It involves building a list of selections, where each selection says which IDs to keep in one event.
Each event can have the same structure as the full document, just removing items not relevant to the current event's package-vulnerability pair (although we do also add the fields from the listings endpoint).

We don't have to use this. It's a bit shorter, but I'm not sure how easy it is to read. It would also mean doing resolution in the ingest pipeline (although that would be fairly straightforward).

It would replace the highlighted section with something like this:
```
body.packages.map(packageRef,
  body.packages[packageRef].vulnerabilitiesRefs.as(vulnRefs,

    // make a selection for each package and vulnerability pair
    // (or package alone if there's one without any vulnerabilities)

    (vulnRefs == null ? [null] : vulnRefs).map(vulnRef, {
      "packageRef": packageRef,
      "vulnerabilityRef": vulnRef
    })

  )
).flatten().map(selection, selection.with({

  // if there's a vulnerability, add any risk accept refs it has

  "riskAcceptRefs": selection.vulnerabilityRef.as(vulnRef,
    vulnRef == null ? [] : body.vulnerabilities[vulnRef].riskAcceptRefs.as(refs, refs == null ? [] : refs)
  )

})).map(selection, selection.with({

  // add the package's layer ref

  "layerRef": body.packages[selection.packageRef].layerRef

})).map(selection, selection.with({

  // if there's a layer ref, add any base image refs it has

  "baseImageRefs": selection.layerRef == null ? [] :
    body.layers[selection.layerRef].as(layer, layer.baseImagesRef.as(refs, refs == null ? [] : refs))

})).as(selections,

  // now map each selection into a full document with that selection of objects
  // (or just the full document if there's no selection/package)

  size(selections) > 0 ?
    selections.map(selection,
      body.with({
        "packages": selection.packageRef.as(ref, { ref: body.packages[ref] }),
        "vulnerabilities": selection.vulnerabilityRef.as(ref, ref == null ? {} : { ref: body.vulnerabilities[ref] }),
        "riskAccepts": selection.riskAcceptRefs.as(refs, zip(refs, refs.map(ref, body.riskAccepts[ref]))),
        "layers": selection.layerRef.as(ref, ref == null ? {} : { ref: body.layers[ref] }),
        "baseImages": selection.baseImageRefs.as(refs, zip(refs, refs.map(ref, body.baseImages[ref]))),
      })
    )
  :
    [body]

).map(result,
  // merge the data from the listing request and the fetch request, add final wrapping

  { "message": worklist.data[0].with(result).encode_json() }
)
```

What do you think?

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,207 @@
...
+                  string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
+              ),
+            },
+          },
+          "want_more": false,
+        }
+      )
+  ).as(worklist,
+    has(worklist.events) ? state : // Exit early due to GET failure.
+      (worklist.?data[0].hasValue()) ?
```
chrisberkhout, line 65: I think there should be a comment for future developers about how we're turning scan results into package-vulnerability events.

Something like:

> Scan results include information about packages, layers, base images, vulnerabilities and risk accepts. These are related as follows:
> 
> A **package** is included in one **layer**, which can be built several **base images**.
> A **package** can have several **vulnerabilities**, which can have several **risk accepts**.
> 
> From the full scan result, we produce one event for each package-vulnerability pair ...

---

I think text as above or maybe ASCII boxes is the right way to describe this. I was thinking about [Mermaid](https://mermaid.js.org/), but I think that's too heavy for this case. Github renders mermaid diagrams in markdown, but we just need a comment in the CEL.

```
---
title: sysdig Vulnerability Results
---
erDiagram

  PACKAGE {
    ref layerRef
    ref[] vulnerabilitiesRefs
  }
  LAYER {
    ref[] baseImagesRef
  }
  VULNERABILITY {
    ref[] riskAcceptRefs
  }
  BASE_IMAGE {}
  RISK_ACCEPT {}

  PACKAGE 0+--1 LAYER : "included in"
  LAYER 0+--0+ BASE_IMAGE : "built on"
  PACKAGE 1+--0+ VULNERABILITY : "has"
  VULNERABILITY 0+--0+ RISK_ACCEPT : "has"
```

Untitled diagram _ Mermaid Chart-2025-07-04-153026

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,207 @@
...
+    - api_token
+program: |
+  (
+    (state.?worklist.data[0].hasValue()) ?
+      state.worklist
+    :
+      request("GET",
+        state.url.trim_right("/") + "/secure/vulnerability/v1/" + state.stage + "-results?" + {
+          "limit": [string(state.batch_size)],
+          ?"cursor": state.?next.page.optMap(v, [v]),
```
chrisberkhout, line 35: Regarding pagination, is it not possible to request scan results that are after scan results we have seen so far?

I wonder if the `next.page` cursor is long-lived, or if there are other filtering parameters. It would be good to avoid fetching everything each day, and also to avoid waiting a day to fetch new things.

I saw in the issue discussion:

> Workloads that are no longer running will be removed from the logs, as mentioned in the [runtime documentation](https://docs.sysdig.com/en/docs/sysdig-secure/vulnerabilities/#runtime).

But in the documentation it sounds like the old runtime workloads will be removed from a UI view, not that earlier scan results from such workloads will immediately become unavailable from the UI.

> * Runtime workloads will offer a live, auto-refreshing state. This means:
>   * Workloads that are no longer running will be removed from the runtime view

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,207 @@
...
+redact:
+  fields:
+    - api_token
+program: |
+  (
+    (state.?worklist.data[0].hasValue()) ?
+      state.worklist
+    :
+      request("GET",
+        state.url.trim_right("/") + "/secure/vulnerability/v1/" + state.stage + "-results?" + {
```
chrisberkhout, line 33: I read the [sysdig Vulnerability Management documentation](https://docs.sysdig.com/en/sysdig-secure/vulnerability-management/), and it seems like users might want to ingest scans from all 3 stages.
Although there would presumably be some overlap in detected vulnerabilities, it makes sense to me that we should keep any scan results available, unless we have a good reason not to.

Perhaps we should double check with product?

If we only allow one, I think we should be clear about it in the documentation.

packages/sysdig/_dev/build/docs/README.md
brijesh-elastic: Sure, I'll update.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
brijesh-elastic: I'll check the feasibility of Mermaid and update/add comments accordingly.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
brijesh-elastic: I agree.
We don't have a filter to collect only the latest scan results, so we have no choice but to gather all the results at every interval.

We can offer a filter option to users where they can apply filtering on supported fields (for example, `kubernetes.cluster.name="production" and kubernetes.pod.container.name="docker.internal.sysdig.tools"`). This will reduce the amount of data fetched and can be useful for users who need data based on specific criteria.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
brijesh-elastic: Agree, All the 3 stages are important in their way. We've added mappings for all 3 stages in ingest pipeline. currently, user can only select the one of the stage, but they always can add another integration and select the other stages.


> If we only allow one, I think we should be clear about it in the documentation.

I'll add the note/guide about it.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
brijesh-elastic: Mermaid templates are not visible on the integration overview page of Kibana; therefore, using plain text to display the associations.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
brijesh-elastic: The split logic you're suggesting is straightforward and easily readable/maintainable. I've tried it out, but it's taking too much time to complete the evaluation, and often, Filebeat/Mito are automatically killed or stop the execution.

I compared the execution times for both approaches. My current approach takes 22s, while your suggested approach takes 1m20s to execute.

packages/sysdig/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
brijesh-elastic: It's not documented; it's based on my experience. As you mentioned, it only serves to make the code more robust.

chrisberkhout: Okay, thanks. I still think it doesn't need to be done unless we expect the format to be unstable, but let's leave it in this time if you want.

packages/sysdig/data_stream/vulnerability/manifest.yml
chrisberkhout: Resolving this as there's a separate discussion of stage and I believe you've added extra explanation in the documentation.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
chrisberkhout: The comment in the CEL is good. Thanks!

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
chrisberkhout: Okay.

Here's what I meant this part:

> I wonder if the `next.page` cursor is long-lived, or if there are other filtering parameters. It would be good to avoid fetching everything each day, and also to avoid waiting a day to fetch new things.

In the system test I see cursor values like `abcdefgabcdefg01234567899876543210==`, which looks like a fake version of a base64-encoded value.

Often APIs have cursors like `dGltZXN0YW1wPjE3NDQ3NzI0NjExOTUxNDk4MDA=`, which can be base64-decoded to `timestamp>1744772461195149800`. A cursor like that is something we can reuse many times to keep checking for data after that point. We could even construct a cursor value like that ourselves using the timestamp of the last data we saw.

I haven't seen a real cursor value from the live API, so I'm not sure if this is one of the cases where it has a clearly identifiable expression or ID that could be used in that way.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
chrisberkhout: Thanks.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
chrisberkhout: Okay, thanks for checking that.

That execution time really is large. Maybe there are ways to optimize it, but then thing would start getting more complicated anyway.

I think in general it's good for the events from CEL to match the upstream data as closely as possible, but in this case some significant restructuring is necessary. So let's go with the current solution.

packages/sysdig/data_stream/vulnerability/agent/stream/cel.yml.hbs
brijesh-elastic: >Often APIs have cursors like dGltZXN0YW1wPjE3NDQ3NzI0NjExOTUxNDk4MDA=, which can be base64-decoded to timestamp>1744772461195149800. A cursor like that is something we can reuse many times to keep checking for data after that point. We could even construct a cursor value like that ourselves using the timestamp of the last data we saw.

Okay, I got it. The real cursor from the live API doesn't include the case you mentioned. I'm only receiving strings such as `NDAw`, `MzEw`, `MTQw`, etc.

Issue #13250, opened 2025-03-22
[CrowdStrike Falcon] Flattened field needs to have ignore_above setting to avoid dropping events.


leandrojmp:
Hello,

We are having some issues with some events being dropped because they have some immense field.

>Flattened field [crowdstrike.event.Attributes] contains one immense field whose keyed encoding is longer than the allowed max length of 32766 bytes. Key length: 25, value length: 114376 for key starting with [old_group_assignment_rule]

Looking at the mappings for this field, the parameter `ignore_above` is not present.

```
"crowdstrike": {
        "properties": {
          "event": {
            "properties": {
              ( .... )
              "Attributes": {
                "type": "flattened"
              }
              ( .... )
            }
          }
        }
      }
```
Some keys in this flattened field can have a lot of values, so the `ignore_above` parameter needs to be added to at least avoid droppping events.

In the example the value for the key `old_group_assignment_rule` has `114 KB`, so the `ignore_above` needs to be set to the maximum value allowed.

kcreddy:
Created PR: https://github.com/elastic/integrations/pull/14127 for adding `ignore_above` option, but it is not getting reflected in the created index pattern.
Raised a bug inside `elastic-package`: https://github.com/elastic/kibana/issues/223245

Alternate option to consider: 
- Truncate very long fields inside `Attributes` field when they cross 32766 character length. Example: https://github.com/elastic/integrations/blob/c8b50b249b04b6743c77b267ed1f8c8bd1e010c7/packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml#L3343-L3345

leandrojmp:
@kcreddy I'm not sure, but wouldn't truncate lead to a mapping conflict?

If the `flattened` expects a json object, truncate will break the json object, which would lead to a mapping conflict dropping the document. (not tested this, so I'm not sure).

kcreddy:
> [@kcreddy](https://github.com/kcreddy) I'm not sure, but wouldn't truncate lead to a mapping conflict?
> 
> If the `flattened` expects a json object, truncate will break the json object, which would lead to a mapping conflict dropping the document. (not tested this, so I'm not sure).

@leandrojmp, the `ignore_above` option for a [flattened](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/flattened) field is applied on each underlying leaf field. 
Similarly, we could iterate over `Attributes` and truncate each field's value inside it. We wont truncate `Attributes` field based on its size.
[...]

PR #14127, opened 2025-06-03
crowdstrike.falcon: Add `ignore_above` to Attributes flattened field

kcreddy:
## Proposed commit message
```
crowdstrike.falcon: Add `ignore_above` option to "Attributes" flattened field.

The "Attributes" field is a flattened data type and can contain 
fields longer than 32766 bytes. When this happens, the index
operation throws an "illegal_argument_exception" indicating 
that the flattened field contains atleast one immense field 
longer than 32766 bytes.

To avoid this, add "ignore_above"[1] option to "Attributes" 
flattened field mapping. This will ignore that field and continue 
to index the document.

[1]: https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/flattened
```

## Checklist

Added pipeline and system sample logs should pass.

## Related issues

- Closes https://github.com/elastic/integrations/issues/13250


kcreddy:
Looks like the system tests are failing because the option `ignore_above` is not being applied to `flattened` field types.

Raised bug in `elastic-package`: https://github.com/elastic/kibana/issues/223245

Alternate option to consider to get around this problem: 
- Truncate very long fields inside `Attributes` field when they cross 32766 character length. Example: https://github.com/elastic/integrations/blob/c8b50b249b04b6743c77b267ed1f8c8bd1e010c7/packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml#L3343-L3345

kcreddy:
Moved to Draft until https://github.com/elastic/kibana/issues/223245 is fixed. PR is currently blocked.

chrisberkhout (commented):
Should be good when https://github.com/elastic/kibana/issues/223245 is done and out.

Since the other issue will be fixed in fleet (not elastic-package) the min stack version of the package will need to be raised to the version with the fix.

A shorter-term solution seems difficult.

Perhaps the mapping change in this PR could be merged without the system test. It'd have no effect immediately, but once the stack catches up it'll start working.

To fix the errors before the fixed stack version I don't see any great option. Maybe optional truncation of long values and adding a flag to indicate when truncation happened.

Issue #14133, opened 2025-06-03
Standardize user name and user email fields across security integrations


chemamartinez:
There exists some inconsistencies on how ECS user fields are handled across different integrations.

Specifically, when usernames are formatted as email addresses (e.g., "[test.user@domain.com](mailto:test.user@domain.com)"), some integrations split the username and domain, while others do not. This inconsistency leads to varied results in dashboards and queries, causing confusion.

According to the [ECS user fields guidelines](https://www.elastic.co/docs/reference/ecs/ecs-user-usage#ecs-user-identifiers):

```
Here are a few pointers to help normalize some simple cases.
- When a system provides a composite value for the user name
  (e.g. DOMAINNAME\username), capture the domain name in user.domain
  and the user name (without the domain) in user.name.
- When a system uses an email address as the main identifier, populate
  both user.id and user.email with it.
```

Based on the guidelines above, the criteria to apply to all the integrations would be:
- When the user name contains an email address, this field is dissected into `<user.name>@<user.domain>`.
- Also the `user.email` field is populated with the full email address.
- Append `user.name` and `user.email` to related.user.

This is how most of the integrations already proceed.

PR #14085, opened 2025-05-30
[azure,o365,m365_defender] ECS mapping improvements

chrisberkhout:
## Proposed commit message

```
[azure,o365,m365_defender] ECS mapping improvements

Various suggested ECS field mapping updates to improve correlation of
ingested events with other security telemetry, for the integrations
azure, o365 and m365_defender.
```

## Discussion

Can be read commit-by-commit.

Suggested mappings done:

- Suggestion: `azure.signinlogs.properties.user_principal_name` -> `user.email`
  Done, whenever the value contains `@`.

- Suggestion: `azure.signinlogs.properties.service_principal_name` -> `service.name`
  Done, whenever it's populated.

- Suggestion: `azure.identityprotection.properties.user_principal_name` -> `user.email`
  Suggestion: `azure.identityprotection.properties.user_display_name` -> `user.full_name`
  Suggestion: `azure.identityprotection.properties.user_id` -> `user.id`
  All done. Matched to the handling in the `signinlogs` data stream.

- Suggestion: `o365.audit.OriginatingDomain` -> `url.domain`
  Suggestion: `o365.audit.Application` -> `process.name`
  Suggestion: `o365.audit.Sha1` -> `file.hash.sha1`
  Suggestion: `o365.audit.Sha256` -> `file.hash.sha256`
  Suggestion: `o365.audit.FileExtension` -> `file.extension`
  Suggestion: `o365.audit.Parameters.From` -> `email.from.address`
  All done. Field defintion and mapping logic added.

- Suggestion: `o365.audit.DeviceName` -> `host.name`
  Done. `DeviceName` is now the first source for `host.name` and two other sources will be tried if it’s empty.

- Suggestion: `o365.audit.TargetFilePath` -> `file.path`
  Done. Also `FilePath`.

- Suggestion: `o365.audit.FileSize` -> `file.size`
  Done. Field definition for `FileSize` added. New mapping logic tries `FileSizeBytes` first and falls back to `FileSize`.

- Suggestion: `o365.audit.Parameters.ForwardAsAttachmentTo` -> `email.to.address`
  Suggestion: `o365.audit.Parameters.ForwardTo` -> `email.to.address`
  Suggestion: `o365.audit.Parameters.RedirectTo` -> `email.to.address`
  All done. Mapping logic added.

Suggested mappings not done:

- Suggestion: `o365.audit.UserId` -> `user.email`
[...]

Issue #8836, opened 2024-01-08
[Check Point] Improve authentication logs normalization


AlessioPaternoster:
## Use Case

As a SOC analyst, authentication events (logon, logoff) are really important for security purposes. I want these logs to be categorized as authentication events. Moreover, I expect log properties, related to authentication information, to be copied / moved to the appropriate ECS fields. 

Elasticsearch fields like `event.type`, `event.outcome` or `event.action` should also be enriched according to information taken from the original event, collected from Check Point appliances.

## Overview

I did some analysis on firewall authentication events collected from our client's Check Point Manager console. We found some fields the current integration pipeline do not handle properly. Here is the list :

- `event.category` : must be _authentication_ for both logon and logoff events. Currently, logoff is categorized as _network_.

- `user.*` : must contain same values as `source.user.*` fields, as documented in ECS.

- `event.type` : should be populated with _start_ / _end_ and _denied_ / _allowed_ values based on `checkpoint.operation` and `checkpoint.audit_status` fields.

- `event.action` : must be populated with _logon-failed_, _logged-in_ or _logged-out_ (not _accept_) according to `checkpoint.operation` and `checkpoint.audit_status` fields.

- `event.outcome` : similary to `event.type` or `event.action`, this field must be populated with _success_ or _failure_ values according to the original logon / logoff event outcome.

- `event.code` : move `checkpoint.operation_number` value to this field (operation ID done by Administrator; each operation is represented by a number)

- `event.reason` : should contain information about a logon failure. Move the temporary field `checkpoint.additional_info` to this one.

- `host.name` : for SSH logons, the field should be populated with the value of `checkpoint.device_name`.

- `host.type` : for SSH logons, the field should be populated with the value of `checkpoint.type`.

- `network.protocol` : for SSH logons, the _ssh_ value must be copied into this field.

As a temporary solution, I created an additonal Elastic ingest pipeline, currently running after the integration one, which fixes problems I wrote above. I tested it out with a bunch of different syslogs and now it is running also on our production environment.

Below you will find the logs I used to create and test the pipeline and the pipeline itself.

### Before

![Before](https://github.com/elastic/integrations/assets/28502784/b0b6153f-826e-4ad1-9a51-0b01d02257de)

### After

![After](https://github.com/elastic/integrations/assets/28502784/9862e059-05b0-421c-ab2b-a6650e3578b3)

### Login Success

```
[...]

PR #14137, opened 2025-06-03
all: standardize user name and user email fields across integrations

chemamartinez:
## Proposed commit message

According to [ECS guidelines](https://www.elastic.co/docs/reference/ecs/ecs-user-usage#ecs-user-identifiers) about user fields usage:

- When a system provides a composite value for the user name (e.g. DOMAINNAME\username), capture the domain name in `user.domain` and the user name (without the domain) in `user.name`.
- When a system uses an email address as the main identifier, populate both `user.id` and `user.email` with it.

These changes try to align to the rest of the integrations on how to handle emails in user name fields. Changes applied include the following:

- When the user name is an email, this field is dissected into `<user.name>@<user.domain>`.
- Also the `user.email` field is populated with the email address.
- Both `user.name` and `user.email` are appended to `related.user`.
- Same for other variants as `source.user`, `destination.user`, etc.

> [!NOTE]  
> Each commit corresponds to an integration.

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/14133


packages/mongodb_atlas/data_stream/organization/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -87,6 +82,28 @@ processors:
       field: response.username
       target_field: user.name
       ignore_missing: true
+  - rename:
+       field: user.name
```
ishleenk17, line 86: Wouldn't this regress as we are renaming an existing field?

chemamartinez: This field is only renamed when `ctx.user.name.indexOf(“@”) > 0` (contains an email). If this rename succeeds, the subsequent processor will dissect that email and should always re-add content to `user.name`.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -689,6 +689,12 @@ processors:
   - lowercase:
       field: user.group.name
       ignore_missing: true
+  - lowercase:
```
mjwolf, line 692: Why is lowercase used here?

mjwolf (commented):
I reviewed the integrations owned by deployment-and-devices.

I have a question about the use of `lowercase` in checkpoint, it's changing the results of some non-email usernames in the pipeline tests. Otherwise I think the changes are good

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
chemamartinez: I have no clue why all user fields are lowercased, I did it to keep consistency with user fields as they are lowercased right above.

I'd remove all `lowercase` processors for user fields as I think it is better to keep the original form for them.

packages/mongodb_atlas/data_stream/organization/elasticsearch/ingest_pipeline/default.yml
ishleenk17: Got it, missed the readding part to user.name.
Then it should be good.

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1692,6 +1687,31 @@ processors:
...
+      target_field: user.email
+      tag: rename_user_email
+      if: ctx.user?.email == null && ctx.user?.name != null && ctx.user.name.indexOf("@") > 0
+  - dissect:
+      field: user.email
+      pattern: '%{user.name}@%{user.domain}'
+      tag: dissect_user_email
+      if: ctx.user?.name == null
+      ignore_missing: true
+      ignore_failure: true
```
zmoog: Looking at a CloudTrail log:

```json
      "userIdentity": {
        "type": "IAMUser",
        "principalId": "PRINCIPALID",
        "arn": "arn:aws:iam::[redacted]:user/maurizio.branca@elastic.co",
        "accountId": "[redacted]",
        "accessKeyId": "[redacted]",
        "userName": "maurizio.branca@elastic.co"
      },
```

I see from the `arn` and `userName` that the value `maurizio.branca@elastic.co` is the actual login name.

If I understand correctly, this change would modify the outcome from:

```json
            "user": {
                "id": "PRINCIPALID",
                "name": "maurizio.branca@elastic.co"
            }
```

To:

```json
            "user": {
                "domain": "elastic.co",
                "email": "maurizio.branca@elastic.co",
                "id": "PRINCIPALID",
                "name": "maurizio.branca"
            }
```

As a result, user.name would no longer contain the full login name.

The [ECS user.name](https://www.elastic.co/docs/reference/ecs/ecs-user#field-user-name) field docs only say:

> Short name or login of the user. [..] example: a.einstein
 
My concern here is that the user.name content for IAM users will no longer match the login name users would type in the AWS console login form.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
chemamartinez: @mjwolf what do you think? do you agree to remove the lowercase processor for all the user fields? or just the one added in this PR?

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
chemamartinez: Thanks for your feedback, I agree with your point.

We had this same conversation for the Okta integration [here](https://github.com/elastic/enhancements/issues/24531#issuecomment-2936305403), in case you want to get more context about the proposed changes.

This change is basically motivated by the attempt to follow the same criteria as other integrations currently do, in particular, the Azure integration that is which is homonymous to this one, currently [does the same thing](https://github.com/elastic/integrations/blob/main/packages/azure/data_stream/identity_protection/_dev/test/pipeline/test-rickyusers-raw.log-expected.json#L43-L48).

However, if the AWS integration has the same circumstances as the Okta integration and you consider the `user.name` should contain the user login, we could keep that behaviour and just populate the `user.email` as well. In that case, what do you think about doing the same for the Azure integration?

I think there is an ambiguity in ECS with these fields, so it is not clear what's better when a user name contains an email. Maybe adding a new field like `user.login` for these cases would help?

efd6: I think here the cloudtrail username should be the login name following the same reasoning as in the Okta case and for the same reasons.

zmoog: Thanks, @chemamartinez, for bringing in all the context. I missed a similar change to the Azure integration during the review of https://github.com/elastic/integrations/pull/14085; thanks for the heads-up.

It appears that https://github.com/elastic/integrations/pull/14085 has aligned the signin and identity protection logs to use the same logic. The signin logs' pipeline was already unpacking `azure.signinlogs.properties.user_principal_name` into `user.name` and `user.domain` 4 years ago. Now we're doing the same with `azure.identityprotection.properties.user_principal_name`.

We probably need to take a step back and agree on the semantics from ECS and the CSPs data model. 

- What's the semantic of `user.email`? Is the actual email address only, or the user login?
- What's the semantic of `user.name`? The "account name" or the user login name?

The `user.name` in ECS' [User fields](https://www.elastic.co/docs/reference/ecs/ecs-user#field-user-name) and OpenTelemetry's [User Attributes](https://www.elastic.co/docs/reference/ecs/ecs-user#field-user-name) seems to focus on the domainless account name, or at least is what they do in all examples.

On Azure Entra ID log categories like signin and identity protection logs, we have the **user principal name** (for example, `maurizio.branca@elastic.co`), which is the unique identifier for the user across tenants.

Maybe we should write down a table for each CSP with all user's attributes and find the mapping between each CSP and ECS/semconv attributes.

chemamartinez: In the case of the AWS integration is simple, we are only receiving `userName` which sometimes contains an email and sometimes not, so we could simplify by deciding what's best for this one.

For me it makes sense to map it as `user.name` for the reason you mentioned (it is the login name and the account part doesn't make sense by itself in AWS), but it also makes sense to fill `user.email` as in this case the user login is also the user email.

That's why I suggested to populate both, in case you think it is not appropriate to have a duplicated field, we should choose to follow the logical correspondence (AWS `userName` -> ECS `user.name`) or look for consistency with the rest of the integrations.

I think the same applies to Azure and the `userPrincipalName` field.

efd6: > The user.name in ECS' [User fields](https://www.elastic.co/docs/reference/ecs/ecs-user#field-user-name) and OpenTelemetry's [User Attributes](https://www.elastic.co/docs/reference/ecs/ecs-user#field-user-name) seems to focus on the domainless account name, or at least is what they do in all examples.

The problem with this approach is that in the context of a service like okta, where the login name is an email address and the email address is not an okta-specific address, we fold/collide users if we remove the domain. A user `the.boss@company.com` is not the same person as `the.boss@malware.org`. From an entity analytics perspective, we need to be able to differentiate these. The problem has arisen because the `user.name` field's definition is inadequately strong.

zmoog: > The problem with this approach is that in the context of a service like okta, where the login name is an email address and the email address is not an okta-specific address, we fold/collide users if we remove the domain.

Yeah, I agree. I feel this applies to AWS and Azure as well.

> A user the.boss@company.com is not the same person as the.boss@malware.org

Definitely.

In CSPs or large cloud services, the `user.name` should probably contain a name that uniquely identifies an entity.

chemamartinez: Thank you both for your feedback! @zmoog would you mind taking a look at this last commit? 961217dc11522c329800a8a9e3da692321de64c5

zmoog (approved):

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
mjwolf: Use of `lowercase` came from this requirement https://github.com/elastic/integrations/issues/8836. So I think it makes sense to continue to use it with the new fields.

PR #14150, opened 2025-06-04
[cisco_asa] Extract specific fields from message ids 717022, 751025

jrmolin:
## Proposed commit message

[cisco_asa] Add specific parsing for 717022, 751025
* Add message-id-specific parsing for the two message ids
* Further extract x509 fields, according to ECS
* no existing tests fail; new data passes

## Checklist

## Author's Checklist

- [ ] 

```
git clone --depth 1 https://github.com/elastic/integrations

cd integrations/packages/cisco_asa

elastic-package build && elastic-package test
```

## Related issues

-

### DIFF
--- a/packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
@@ -976,6 +976,13 @@ processors:
         - '^Group %{NOTSPACE:_temp_.cisco.webvpn.group_name} User( %{NOTSPACE:source.user.name})? IP %{NOTSPACE:destination.address} AnyConnect session (resumed connection|resumed. Connection) from( IP)? %{NOTSPACE:source.address}\.$'
       pattern_definitions:
         NOTBRACKET: "[^<>]+"
+  - grok:
+      if: "ctx._temp_.cisco.message_id == '717022'"
+      tag: parse_717022
+      field: message
+      description: "717022"
+      patterns:
+        - 'Certificate was successfully validated. serial number:\s*%{NOTSPACE:x509.serial_number}, subject name:\s*%{NOTSPACE:x509.subject.distinguished_name}\.'
   - dissect:
       if: "ctx._temp_.cisco.message_id == '721016'"
       tag: parse_721016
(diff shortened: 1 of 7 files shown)


packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1266,7 +1266,74 @@ processors:
...
+            def k = entry.getKey();
+            def _v = entry.getValue();
+            if (parts.containsKey(k)) {
+              def v = parts[k];
+              List values = (v instanceof List) ? v : [v];   // `[v]` is a Painless list literal
+              subject[_v["name"]] = values;
+            }
+          }
+          ctx._temp_.cisco.dn_parts = subject;
+      if: 'ctx._temp_.cisco?.dn_parts != null'
```
taylor-swanson: Seems like this can be simplified.

```suggestion
      tag: script_distinguished_name_parts
      description: >-
        Sets the parts to the correct names
      params:
        "ST": "state_or_province"
        "S": "state_or_province"
        "P": "state_or_province"
        "CN": "common_name"
        "C": "country"
        "L": "locality"
        "O": "organization"
        "OU": "organizational_unit"
      source: |
        if (ctx._temp_?.cisco?.dn_parts == null) {
          return;
        }
        def parts = [:];
        ctx._temp_.cisco.dn_parts.forEach((k, v) -> { if (params.containsKey(k)) { parts[params[k]] = v; } else return false; });
        ctx._temp_.cisco.dn_parts = parts;
```

jrmolin: yeah, i was doing other things at first. simplifying is good.

packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1248,7 +1266,66 @@ processors:
...
+          def parts = [:];
+          ctx._temp_.cisco.dn_parts.forEach((k,v) -> {
+            if (params.containsKey(k)) {
+              parts[params[k]] = (v instanceof List) ? v : [v];   // `[v]` is a Painless list literal
+            } else {
+              return false;
+            }
+          });
+          ctx._temp_.cisco.dn_parts = parts;
+      if: 'ctx._temp_.cisco?.dn_parts != null'
```
taylor-swanson: This line is redundant since the main script already checks for null. (if statements are painless scripts as well, so it's not like the if statement is going to be any faster).
```suggestion
```

packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1248,7 +1266,66 @@ processors:
...
+        "O": "organization"
+        "OU": "organizational_unit"
+      source: |
+          if (ctx._temp_?.cisco?.dn_parts == null) {
+            return;
+          }
+          def parts = [:];
+          ctx._temp_.cisco.dn_parts.forEach((k,v) -> {
+            if (params.containsKey(k)) {
+              parts[params[k]] = (v instanceof List) ? v : [v];   // `[v]` is a Painless list literal
```
taylor-swanson, line 1300: Oops, my bad, forgot that ECS wants these values to be an array.

jrmolin: yeah, that's why i kept it that way

packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
jrmolin: oh; i didn't know that about `if` statements. i thought it would be faster not to spin up an environment, but it definitely isn't. i understand now.

jrmolin: i made this change yesterday, but somehow forgot to push it after it passed my local testing.

PR #14153, opened 2025-06-05
[Microsoft SQLServer] Add support for list of Dynamic Counter values

### LINKED ISSUE #14083: [Microsoft SQLServer] Add support for List of Dynamic counters
### Summary

The integration currently supports dynamic counter value using LIKE operator which provides capability to add more counters based on the input provided.

Based on the recent research we try to use the Regex operator to add multiple counter values as input.

### Challenges in using Regex:

- As a user knowing the exact metrics, Do we need to build a regex pattern? 
- Using regex doesn't offer the exact match for a metric and pulls additional metrics.

### Observations

Is it possible to add list of counter values as an input? 
Yes, I have making the counter value as array and that which generates me the query to match the given input.

### Update Integration

Will changing the input parameter from text to an array ensure seamless integration upgrade experience for the end user? 
Yes / No
Yes, If the dynamic counter is black. This may not be the case as we have default value for this config.
No, The integration fails to upgrade and reports issue with the field type when the counter value exists.

#### How to fix the upgrade failure?
This can be achieved by removing the existing input value from the config and save the integration then upgrade the integration by providing the dynamic counter values.
We can check whether its possible to upgrade integration without modifying the input value.

#### Short video on upgrade scenario,

https://github.com/user-attachments/assets/18ff1916-485a-496c-9e61-58b40eb3f0db

muthu-mps:
- Breaking change

Please see this [issue](https://github.com/elastic/integrations/issues/14083) for more details.

This PR adds capability to pass list of dynamic counter values. This way user has more flexibility to add multiple values as input to fetch the desired metrics.

This change introduces a breaking update, which is why we are releasing a **new major version**. Users will need to manually remove the dynamic counter input from the UI, upgrade the integration, and then re-add the dynamic counters. This is necessary because the input parameter type has changed from text to array.

## Checklist

## Author's Checklist

- [ ] 

- Install the 2.12.0 integration version
- Validate the metrics collection.
- Upgrade the integration to 3.0.0. Follow the manual steps mentioned in the document on upgrade.
- Validate the metrics collection.

## Related issues

- Closes #14083

### DIFF
--- a/packages/microsoft_sqlserver/data_stream/performance/agent/stream/stream.yml.hbs
+++ b/packages/microsoft_sqlserver/data_stream/performance/agent/stream/stream.yml.hbs
@@ -43,8 +43,12 @@ sql_queries:
     response_format: table
   - query: "SELECT cntr_value As 're_compilations_per_sec' FROM sys.dm_os_performance_counters WHERE counter_name = 'SQL Re-Compilations/sec'"
     response_format: table
-  - query: "SELECT counter_name, cntr_value FROM sys.dm_os_performance_counters WHERE counter_name like '{{dynamic_counter_name}}'"
+{{#if dynamic_counter_name}}
+  {{#each dynamic_counter_name as | dynamic_counter i|}}
+  - query: "SELECT counter_name, cntr_value FROM sys.dm_os_performance_counters WHERE counter_name like '{{dynamic_counter}}'"
     response_format: variables
+  {{/each}}
+{{/if}}
 {{#if processors}}
 processors:
 {{processors}}
--- a/packages/microsoft_sqlserver/_dev/build/docs/README.md
+++ b/packages/microsoft_sqlserver/_dev/build/docs/README.md
@@ -2,6 +2,15 @@
 
 The Microsoft SQL Server integration package allows you to search, observe, and visualize the SQL Server audit logs, as well as performance and transaction log metrics, through Elasticsearch.
 
+## ⚠️ Important Note: Upgrading from version 2.12.0 to 3.0.0
+When upgrading from version 2.12.0 to 3.0.0, you must manually modify the dynamic counter configuration for performance metrics data stream. Follow these steps:
+ 1. Before upgrading, note your current `Dynamic Counter Name` values
+ 2. Remove the default value from the `Dynamic Counter Name` configurations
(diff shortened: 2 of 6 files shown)


packages/microsoft_sqlserver/data_stream/performance/agent/stream/stream.yml.hbs
```diff
@@ -43,8 +43,12 @@ sql_queries:
     response_format: table
   - query: "SELECT cntr_value As 're_compilations_per_sec' FROM sys.dm_os_performance_counters WHERE counter_name = 'SQL Re-Compilations/sec'"
     response_format: table
-  - query: "SELECT counter_name, cntr_value FROM sys.dm_os_performance_counters WHERE counter_name like '{{dynamic_counter_name}}'"
+{{#if dynamic_counter_name}}
+  {{#each dynamic_counter_name as | dynamic_counter i|}}
```
ishleenk17, line 47: Do we intend to use "i" anywhere, else remove it.

ishleenk17:
@muthu-mps : I think this is an important enhancement. And I would like to get this through.
I don't see any blocker. Can we take it to closure ?

muthu-mps:
> @muthu-mps : I think this is an important enhancement. And I would like to get this through. I don't see any blocker. Can we take it to closure ?

Yeah I agree there is no blocker, We need to agree upon the breaking upgrade change. cc: @lalit-satapathy

ishleenk17:
> > @muthu-mps : I think this is an important enhancement. And I would like to get this through. I don't see any blocker. Can we take it to closure ?
> 
> Yeah I agree there is no blocker, We need to agree upon the breaking upgrade change. cc: @lalit-satapathy

Can you point me to the exact point of contention here ?

muthu-mps:
> > > @muthu-mps : I think this is an important enhancement. And I would like to get this through. I don't see any blocker. Can we take it to closure ?
> > 
> > 
> > Yeah I agree there is no blocker, We need to agree upon the breaking upgrade change. cc: @lalit-satapathy
> 
> Can you point me to the exact point of contention here ?

Please see this [issue](https://github.com/elastic/integrations/issues/14083#:~:text=the%20given%20input.-,Integration%20Upgrade,-Will%20changing%20the) for more details.

PR #14167, opened 2025-06-06
[Cisco Meraki] Add filter in grok parsing for port events

moxarth-rathod:
## Proposed commit message

```
cisco_meraki: fix pipeline issue by adding a filter in grok parsing for port events

As the current Grok pattern in the ingest pipeline is designed to match log lines that indicate specific port actions, i.e. `status changes` or `STP role changes`, this PR adding a filter in grok parsing for port events to logs with actionable phrases to prevent errors from benign messages.
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/cisco_meraki directory.
> elastic-package test -v

### DIFF
--- a/packages/cisco_meraki/changelog.yml
+++ b/packages/cisco_meraki/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.28.2"
+  changes:
+    - description: Limit Grok parsing for port events to logs with actionable phrases to prevent errors from benign messages.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/1
 - version: "1.28.1"
   changes:
     - description: Extend the event pipeline with some ECS fields and a Grok pattern to improve DHCP event parsing.
--- a/packages/cisco_meraki/data_stream/log/elasticsearch/ingest_pipeline/events.yml
+++ b/packages/cisco_meraki/data_stream/log/elasticsearch/ingest_pipeline/events.yml
@@ -155,7 +155,7 @@ processors:
       SYSLOGHDR: '%{SYSLOGPRI}%{SYSLOGVER}'
       WORDORHOST: '(?:%{WORD}|%{HOSTNAME})'
       PORTACTION: '(?:changed stp role|status changed)'
-    if: ctx.event.original.startsWith('<') && ctx.cisco_meraki?.event_subtype == "port"
+    if: ctx.event.original.startsWith('<') && ctx.cisco_meraki?.event_subtype == "port" && (ctx.event.original.toLowerCase().contains('status changed') || ctx.event.original.toLowerCase().contains('changed stp role'))
 - gsub:
     field: _temp.port_action
     pattern: ' '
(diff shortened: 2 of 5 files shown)


packages/cisco_meraki/data_stream/log/elasticsearch/ingest_pipeline/events.yml
```diff
@@ -155,7 +155,7 @@ processors:
       SYSLOGHDR: '%{SYSLOGPRI}%{SYSLOGVER}'
       WORDORHOST: '(?:%{WORD}|%{HOSTNAME})'
       PORTACTION: '(?:changed stp role|status changed)'
-    if: ctx.event.original.startsWith('<') && ctx.cisco_meraki?.event_subtype == "port"
+    if: ctx.event.original.startsWith('<') && ctx.cisco_meraki?.event_subtype == "port" && (ctx.event.original.toLowerCase().contains('status changed') || ctx.event.original.toLowerCase().contains('changed stp role'))
```
ShourieG: Can we check if the following works - 

```
if: ctx.event.original.startsWith('<') && ctx.cisco_meraki?.event_subtype == 'port' && ctx.event.original.contains(/(?i)status changed|changed stp role/)
```

moxarth-rathod: @ShourieG this approach doesn't work, as `contains()` only expects a string, not a regex pattern.

ShourieG: ok, then I'm guessing something like this `ctx.event.original =~ /(?i)status changed|changed stp role/` or similar should work as I can see regex comparisons working in if conditionals [here](https://github.com/elastic/integrations/blob/4dc1ae5b1468102605dbcfe25a7f4a524f6293f8/packages/snyk/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml#L64). The idea is to use a regex pattern to avoid calling
`toLowerCase().contains()` twice. If the provided regex pattern fails to compile then something similar should be possible.

ShourieG: @moxarth-rathod, then lets do one optimization, let's store a lowercase value of ctx.event.original in a temp var and use it here and then remove it in the cleanup phase. That way toLowerCase() is called only once. Since the event.original can vary in size, this will keep the performance impact in check for very large events.

moxarth-rathod: Yes, right. That way it can be more optimised. I've made the changes as per your suggestion.

packages/cisco_meraki/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.28.2"
+  changes:
+    - description: Limit Grok parsing for port events to logs with actionable phrases to prevent errors from benign messages.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/1
```
ShourieG: Please fix pr number @moxarth-rathod

ShourieG:
```suggestion
      link: https://github.com/elastic/integrations/pull/14167
```

packages/cisco_meraki/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.28.2"
+  changes:
+    - description: Limit Grok parsing for port events to logs with actionable phrases to prevent errors from benign messages.
```
ShourieG:
```suggestion
    - description: Limit Grok parsing for port events to logs with actionable phrases to prevent benign error messages.
```

ShourieG (approved):
LGTM

PR #14170, opened 2025-06-06
[O365_metrics] Fix CEL evaluation error in teams_call_quality data stream.

ritalwar:
- Bug

## Proposed commit message

Fix cel code in teams_call_quality data stream to ensure state is populated after the case with empty events and want_more: false, so the program can continue evaluation without errors.

## Checklist


packages/o365_metrics/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.13.2"
+  changes:
+    - description: Fix CEL evaluation error in teams_call_quality data stream.
```
devamanv:
```suggestion
    - description: Fix CEL evaluation error in `teams_call_quality` data stream.
```

Linu-Elias (approved):
LGTM

packages/o365_metrics/data_stream/teams_call_quality/agent/stream/cel.yml.hbs
```diff
@@ -305,6 +305,12 @@ program: |
     :
     {
       "events": [],
-      "want_more": false
+      "want_more": false,
+      "base": state.base,
+      "sync_delay": state.sync_delay,
+      "worklist": [],
```
muthu-mps, line 311: Does it reset the worklist here as well?
https://github.com/elastic/integrations/blob/563f0e22f5977bc1307b4e1049fbd3a8a5661bb6/packages/o365_metrics/data_stream/teams_call_quality/agent/stream/cel.yml.hbs#L283

packages/o365_metrics/data_stream/teams_call_quality/agent/stream/cel.yml.hbs
```diff
@@ -305,6 +305,12 @@ program: |
     :
     {
       "events": [],
-      "want_more": false
+      "want_more": false,
```
muthu-mps, line 308: In each case want_more is marked as false at the end. Do you think this is redundant here?
https://github.com/elastic/integrations/blob/563f0e22f5977bc1307b4e1049fbd3a8a5661bb6/packages/o365_metrics/data_stream/teams_call_quality/agent/stream/cel.yml.hbs#L127

efd6 (commented):
This does not correctly handle the case that the response has a status code that is not 200. It is important to understand that an error does not result in an input restart, so in the error state, return also requires that parts of the config are also retained. This will continue to fail after a non-200 status code return.

I've re-looked at this and there is a `state.with` call at L133 that ensures that the fields are included.

Issue #14142, opened 2025-06-04
[Agentless Integrations] Agentless integrations 'organization' field pipeline failure


narph:
Pipeline failure is:
```
Processor 'rename' with tag '' failed with message 'cannot set [id] with parent object of type [java.lang.String] as part of path [organization.id]'
```

Where the field causing the failure is `organization` below:

```json
"division": "engineering",
"organization": "security",
"team": "security-service-integrations",
```

Agentless policies add the division, organization, and team as global data tags: 

https://github.com/elastic/kibana/blob/12b7429afe0fb470fa6c85168c9040a5957bbe8b/x-pack/platform/plugins/shared/fleet/public/applications/fleet/sections/agent_policy/create_package_policy_page/single_page_layout/hooks/setup_technology.ts#L195-L196

Global data tags get injected via add_fields processors when the policy is created:

https://github.com/elastic/kibana/blob/12b7429afe0fb470fa6c85168c9040a5957bbe8b/x-pack/platform/plugins/shared/fleet/server/services/agent_policies/package_policies_to_agent_inputs.ts#L181-L191

Until this will be fixed in https://github.com/elastic/kibana/issues/221312, a workaround should be applied in the current agentless integrations simialr to https://github.com/elastic/integrations/pull/14056

PR #14172, opened 2025-06-06
Agentless: remove organization field in ingest pipeline for Agentless integrations

chemamartinez:
## Proposed commit message

```
Fixes a potential ingestion error in integrations that support Agentless
because Agentless agents include a global processor to add the
organization field as a string, which collides with the ECS organization field
the integration expects.

This is a temporary workaround that removes the fields added by
Agentless from the ingested documents to avoid collision in integrations 
that populate that fields, as well as type conflict in searchs.
```

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/14142
- Relates https://github.com/elastic/kibana/issues/221312


ishleenk17:
@chemamartinez : How is it determined which integrations support Agentless ?
Also, whats the plan once the kibana fix is merged ? We will reintroduce these back ?

packages/aws/data_stream/securityhub_findings/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1,6 +1,13 @@
 ---
 description: Pipeline for processing AWS Security Hub Findings logs.
 processors:
+  - remove:
+      field: organization
+      if: ctx.organization != null && ctx.organization instanceof String
+      description: |
+        The organization field is added by Agentless as metadata,
+        which can collide with ECS organization fields.
+      tag: remove_organization_agentless
```
zmoog: Alternative: rename `organization` to `organization.name`, if instance of string?

chemamartinez:
> @chemamartinez : How is it determined which integrations support Agentless ?

@ishleenk17 I determined the integrations that support Agentless by looking at the [manifest files](https://github.com/elastic/integrations/blob/main/packages/armis/manifest.yml#L39-L43).

>  Also, whats the plan once the kibana fix is merged ? We will reintroduce these back ?

I am afraid the workaround could not be removed right after the Kibana fix is merged, it is planned to be fixed at 9.2 but still users running previous versions (8.18, 9.0, 9.1) will be still affected, so I think we would have to wait for a prudent period of time to be safe.

packages/aws/data_stream/securityhub_findings/elasticsearch/ingest_pipeline/default.yml
chemamartinez: In that case, we would be introducing internal data that is just used for Agentless internal purposes into an ECS field so it would be part of the event. In addition, Fleet or other components that use that field would not be able to read it as they expect `organization` instead of `organization.name`.

packages/armis/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: 0.1.1
+  changes:
+    - description: |
+        Add temporary processor to remove the organization field that
+        is populated by the Agentless policy.
+        This workaround won't be necessary once https://github.com/elastic/kibana/issues/221312 is solved.
```
efd6: Is this desirable user-facing text?

packages/armis/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -12,6 +12,13 @@ processors:
   - drop:
       if: ctx.message == 'retry'
       tag: drop_retry_events
+  - remove:
+      field: organization
+      if: ctx.organization != null && ctx.organization instanceof String
+      description: |
```
efd6:
```suggestion
      description: >-
```
?

(similar queries below)

packages/armis/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: 0.1.1
+  changes:
+    - description: |
```
efd6:
```suggestion
    - description: >-
```
?

I'm not convinced we need to preserve the line breaks here.

packages/armis/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -12,6 +12,13 @@ processors:
   - drop:
       if: ctx.message == 'retry'
       tag: drop_retry_events
+  - remove:
+      field: organization
+      if: ctx.organization != null && ctx.organization instanceof String
```
efd6: Do we ever expect a non-string value here by this time in the pipeline? Do we ever expect a value here by now at all except from the agentless agent code?

efd6 (commented):
Queries apply to all packages, but only asked for the first.

packages/armis/changelog.yml
chemamartinez: I agree probably it's better to omit that.

packages/armis/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
chemamartinez: I can't think of any situation where we can receive a value in this field that doesn't come from the Agentless code, and that field will always be a string value.

Do you have any in mind?

efd6: No, that's why I asked. I don't think this is a problem, but it could be simpler.
```suggestion
      if: ctx.organization instanceof String
```
or
```suggestion
      if: ctx.organization != null
```

chemamartinez: As we were not 100% sure I have searched for it and I think I found a couple of possibilities:

- It is possible that the [`decode_cef`](https://github.com/elastic/beats/blob/main/x-pack/filebeat/processors/decode_cef/keys.ecs.go#L42-L43) processor populate the ECS organization fields, and it is used by a considerable number of integrations.
- The [`Add Cloud Foundry metadata`](https://www.elastic.co/guide/en/beats/filebeat/8.18/add-cloudfoundry-metadata.html) processor is not currently used by any integration but it also populate organization fields.

It is a remote possibility that a collision happens but I'd let the full condition just in case.

kcreddy: How about the case when users have logstash in their pipeline and set `organization` fields inside logstash filters? This might be actually be rare though.
We could use this condition instead: `ctx.organization == 'security'` to avoid removing anything else.

chemamartinez: I think that could also be a problem as would collide with ECS on a type mismatch (string vs object), so potentially could generate the same pipeline error.

packages/cloud_security_posture/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 3.3.2
 name: cloud_security_posture
 title: "Security Posture Management"
-version: "2.0.0-preview03"
+version: "2.0.1"
```
seanrathier: The Cloud Security Posture team uses the `preview` postfix before releasing a new version.

seanrathier:
I am not sure if this will happen, but we may need to make some changes [here](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/public/applications/fleet/sections/agent_policy/create_package_policy_page/single_page_layout/hooks/setup_technology.ts#L284) to avoid problems when creating agentless agents.  The global tags are read and sent to the Agentless API [here](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/server/services/agents/agentless_agent.ts#L306).

https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/public/applications/fleet/sections/agent_policy/create_package_policy_page/single_page_layout/hooks/setup_technology.ts#L284

chemamartinez:
> I am not sure if this will happen, but we may need to make some changes [here](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/public/applications/fleet/sections/agent_policy/create_package_policy_page/single_page_layout/hooks/setup_technology.ts#L284) to avoid problems when creating agentless agents. The global tags are read and sent to the Agentless API [here](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/server/services/agents/agentless_agent.ts#L306).
> 
> https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/public/applications/fleet/sections/agent_policy/create_package_policy_page/single_page_layout/hooks/setup_technology.ts#L284

@seanrathier are these global tags being read from the ingested events? if these tags are needed to create agentless agents, they should be known before events are ingested I'd say.

I think the scope of https://github.com/elastic/kibana/issues/221312 is to fix this on the Agentless side, but we need this workaround on the integrations side in the meantime, as that fix won't be available in the near term.

packages/cloud_security_posture/manifest.yml
chemamartinez: Thanks for pointing that out! I am not sure if I should change the version to `2.0.0-preview04` or `2.0.1-preview01`. Could you help me with that?

seanrathier: You can change it to `2.0.0-preview04`

seanrathier:
> @seanrathier are these global tags being read from the ingested events?

I believe they are not being read from the ingested events.  We add the organization, division, and team to the agent policy's `global_tags` attribute and unpack them for the Agentless API `/deployment` endpoint `label` attribute.

The Agentless API/Controller then uses them to tag the agentless agent pod deployment.

AFAIK.

packages/cloud_security_posture/manifest.yml
chemamartinez: Done https://github.com/elastic/integrations/pull/14172/commits/c4a2a81de76a31a22c4d08e91490eac37c54a42a

chemamartinez:
> > @seanrathier are these global tags being read from the ingested events?
> 
> I believe they are not being read from the ingested events. We add the organization, division, and team to the agent policy's `global_tags` attribute and unpack them for the Agentless API `/deployment` endpoint `label` attribute.
> 
> The Agentless API/Controller then uses them to tag the agentless agent pod deployment.
> 
> AFAIK.

In that case, this PR shouldn't break anything as it just removes the organization field from the ingested events.

zmoog (approved):

packages/armis/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
efd6: The situation is messy, suggest `if: ctx.division instanceof String && ctx.organization instanceof String && ctx.team instanceof String` and remove all of them. This is not perfect, but it is unlikely that all three of the fields will have been set by a vendor data source.

packages/ess_billing/data_stream/billing/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -3,6 +3,13 @@ description: "Ingest pipeline for ESS billing data"
 processors:
   - drop:
       if: "ctx.ess?.billing == null && ctx.error?.message == null"
+  - remove:
+      field: organization
+      if: ctx.organization != null && ctx.organization instanceof String
```
3kt: I'm confused as to why this is even necessary: the integration doesn't expose `organization`, only `organization_id`: see [this](https://github.com/elastic/integrations/blob/main/packages/ess_billing/data_stream/billing/fields/fields.yml#L36-L38)

chemamartinez: When running on Agentless, the integration's events also include [these fields](https://github.com/elastic/integrations/blob/main/packages/ess_billing/manifest.yml#L40-L42), they are added by the Agentless policy as tags to identify the integration. See https://github.com/elastic/kibana/issues/221312.

packages/armis/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
chemamartinez: https://github.com/elastic/integrations/pull/14172/commits/c4d73944af889d12f4042759898dc9bff95408a8

packages/ess_billing/data_stream/billing/elasticsearch/ingest_pipeline/default.yml
3kt: Gotcha - thanks for clarifying <3

3kt (approved):
LGT @elastic/customer-architects for `ess_billing` integration

PR #14175, opened 2025-06-06
qualys_vmdr: prevent sending empty IDs in the knowledge base API request

navnit-elastic:
## Proposed commit message

```
qualys_vmdr: prevent sending requests with an empty 
ids parameter to the Knowledge Base API in the 
asset host detection datastream.

This change adds a check to ensure the request to the
Knowledge Base API is only made when HOST_QID_LIST 
contains at least one ID. Previously, the integration was 
sending requests without checking if the IDs were empty, 
which resulted in an API error.
```

## Checklist

## Author's Checklist

- [ ] 

System Tests:
```
--- Test results for package: qualys_vmdr - START ---
╭─────────────┬──────────────────────┬───────────┬───────────┬────────┬─────────────────╮
│ PACKAGE     │ DATA STREAM          │ TEST TYPE │ TEST NAME │ RESULT │    TIME ELAPSED │
├─────────────┼──────────────────────┼───────────┼───────────┼────────┼─────────────────┤
│ qualys_vmdr │ asset_host_detection │ system    │ ahd-kb    │ PASS   │ 2m17.289764466s │
╰─────────────┴──────────────────────┴───────────┴───────────┴────────┴─────────────────╯
--- Test results for package: qualys_vmdr - END   ---
Done
```
## Related issues

-

### DIFF
--- a/packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
+++ b/packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
@@ -156,7 +156,7 @@ program: |
       )
     ).as(state, state.with(
       !has(state.worklist) ? state : 
-        size(state.worklist) > 0 ?
+        has(state.worklist.HOST_QID_LIST) && size(state.worklist.HOST_QID_LIST[0]) > 0 ?
           request("GET", state.url.trim_right("/") + "/api/3.0/fo/knowledge_base/vuln/?" +
             {
               "ids": [state.worklist.HOST_QID_LIST[0].join(",")],
--- a/packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
+++ b/packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
@@ -277,6 +277,7 @@ program: |
             "events": [],
             "want_more": false,
             "next_page": {},
+            "worklist": {},
           }
       )
     )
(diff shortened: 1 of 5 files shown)


packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
```diff
@@ -156,7 +156,7 @@ program: |
       )
     ).as(state, state.with(
       !has(state.worklist) ? state : 
-        size(state.worklist) > 0 ?
+        has(state.worklist.HOST_QID_LIST) && size(state.worklist.HOST_QID_LIST[0]) > 0 ?
```
chrisberkhout, line 159: The change avoids a request if
```
has(state.worklist.HOST_QID_LIST) && size(state.worklist.HOST_QID_LIST[0]) == 0
```

which is good, but once it's in that state, does it get out?

In lines 276-280 I see an object returned that gets combined with state and leaves the cursor and its worklist intact.

Am I missing anything?

navnit-elastic: You are absolutely right @chrisberkhout, We need to reset worklist when above condition is false.
We don't need to reset cursor as we are not using it anywhere.

navnit-elastic: The else part must return:
```
{
  "events": [],
  "want_more": false,
  "next_page": {},
  "worklist": {},
}
```

chrisberkhout: Okay, thanks. Looks good.

I said cursor by mistake because I'm used to a worklist being there. 

Just for future work... Data in the cursor will survive restarts, while the rest of state will not. Sometimes you want pagination or cursor information to only be short-lived and it's okay if it is reset and repeated, but usually it's better to put that information in the cursor section.

PR #14191, opened 2025-06-10
Allow ISO8601 datetime format

Alphayeeeet:
Please label as enhancement

The officially as-well supported ISO8601 timestamp shall be added to the date processor.

### DIFF
--- a/packages/keycloak/changelog.yml
+++ b/packages/keycloak/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.28.1"
+  changes:
+    - description: Allow ISO8601 datetime format.
+      type: enhancement
+      link: TODO
 - version: "1.28.0"
   changes:
     - description: Set the ECS field `event.outcome` based on the value of `keycloak.login.type`.
--- a/packages/keycloak/manifest.yml
+++ b/packages/keycloak/manifest.yml
@@ -1,6 +1,6 @@
 name: keycloak
 title: Keycloak
-version: "1.28.0"
+version: "1.28.1"
 description: Collect logs from Keycloak with Elastic Agent.
 type: integration
 format_version: "3.0.3"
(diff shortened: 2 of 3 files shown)


efd6 (commented):
Is there a test for this and a demonstrated need?

Alphayeeeet:
I didn't add a test as I didn't find an official Keycloak doc of the supported date format, but we tested it internally for our instances. If necessary, I'm perfectly fine with adding a test-case.

Context: We use RHSSO, there are frequent parsing errors because of the date format, as it somehow uses ISO-like format without timezone offset. As this change should be only beneficial as multiple date formats are supported and checked sequentially by the processor, this would fix these issues. 

We might be able to change the log timestamp, however, I think it would be more beneficial to support proprietary distributions of Keycloak as well.

efd6:
Thanks. Please add a test.

Alphayeeeet:
@efd6 Test case has been added. Please run CI again.

efd6 (commented):
I suspect that you are using and older version of elastic-package. Can you update that to the most recent release and redo `elastic-package test pipeline -g`?

Alphayeeeet:
@efd6 Version was updated, however there were no significant changes in the expected test results.
Please run CI again and approve if applicable. Thanks

efd6:
Thanks. The change I was expecting was made.

packages/keycloak/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.28.1"
```
efd6:
```suggestion
- version: "1.29.0"
```
(enhancements bump minor)

packages/keycloak/manifest.yml
```diff
@@ -1,6 +1,6 @@
 name: keycloak
 title: Keycloak
-version: "1.28.0"
+version: "1.28.1"
```
efd6:
```suggestion
version: "1.29.0"
```

Alphayeeeet:
@efd6 Has been fixed

efd6 (approved):
Thanks

PR #14193, opened 2025-06-10
[ServiceNow] Fix missing date format and incorrect handling of non-terminal errors

StacieClark-Elastic:
## Proposed commit message
[ServiceNow] Fixes several issues.
Fixes issue where any datetime display_value field with AM/PM were not parsing. Updated every instance of the date processor for display_value to include a date format that handles AM/PM. 
Fixed a problem where an unencoded URL causes entire pipeline to fail. Added code to process the URL when its unencoded. 
Sets uri.full only if the original URI has a scheme to handle cases where the URL included only query related information

## Checklist

- Closes #14193


packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -3346,7 +3376,7 @@ processors:
           return src;
         }
         filterMassive(ctx);
-  - set:
+  - append:
```
efd6, line 3430: I'm not sure that we can do this. It's something that I have wanted to do, but there has been push back because it's not noted to be an array in ECS, though we could say exactly the same thing about `error.message` which we do treat this way.

/cc @andrewkroh

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -2938,14 +2966,15 @@ processors:
   - uri_parts:
       field: servicenow.event.url.display_value
       tag: uri_parts_url_full
-      ignore_missing: true
+      if: ctx.servicenow?.event?.url?.display_value != null && ctx.servicenow.event.url.display_value != ''
       on_failure:
         - append:
             field: error.message
-            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}  I was called from processor'
```
andrewkroh: The looks like leftover debug info that should be removed.

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
andrewkroh: Keeping `event.kind` as a scalar via `set` is correct based on a strict interpretation of ECS. Using `append` would provide better utility for the data.

There are 106 instances of `append(event.kind)` as of commit 090272918d0 (see [findings](https://gist.github.com/andrewkroh/a2d257fc3e386c63ee1d2e3b6ba89495#file-append-event-kind-csv)). Given the other usages of `append`, I feel confident that we won't break anything within Kibana.

However, this approach could lead to issues in the future if we need to integrate the data into a strictly typed system, such as a protobuf implementation of ECS. Therefore, avoiding `append` reduces the potential for new tech debt.

This relates to https://github.com/elastic/ecs/issues/2270, which didn't gain adoption due to being a breaking change for non-Elasticsearch data stores. If we need this then we should push on the ECS issue to change the schema.

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -3346,11 +3376,6 @@ processors:
           return src;
         }
         filterMassive(ctx);
-  - set:
-      field: event.kind
-      tag: set_pipeline_error_into_event_kind
-      value: pipeline_error
-      if: ctx.error?.message != null
```
efd6: I'm still not convinced that this is the correct answer here. These errors are informative and they can be avoided by expanding the tolerance of the pipeline for syntactically-correct-adjacent fields.

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -2938,14 +2966,15 @@ processors:
...
-      ignore_missing: true
+      if: ctx.servicenow?.event?.url?.display_value != null && ctx.servicenow.event.url.display_value != ''
       on_failure:
         - append:
             field: error.message
             value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
   - set:
       field: url.full
       tag: set_url_display_value_into_url_full
+      if: ctx.url != null && ctx.servicenow?.event?.url?.display_value =~ /^([a-z][a-z0-9+\-.]*):/
```
andrewkroh, line 3021: Can you please add a `description` field to the processor with the background of why it sets the url.full only when it has a scheme?

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: The errors go into the message. They just don't stop the pipeline

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
andrewkroh: Stacie and I talked about this today. She is going to push on the ECS change with the breaking change process. In the meantime, make it 107 instances of append(event.kind).

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: Andrew and I discussed this today. The script uses event.kind to mark the kind of event. We cannot stomp it when there is a pipeline error that we don't want to stop processing because it's existence is necessary. 

The existence of error.message should be enough for the user to look at the event and see that there were some data errors that require looking at. Errors that were not serious enough to stop processing.

However, if we want to keep using pipeline_error a  possible way forward is to change this back to append so that the event.kind is preserved and the pipeline_error is added. We know that 106 other integrations follow this same model. We would then put in a request to the breaking changes board to update ECS to handle an array so the use of this model is supported by the ECS schema. 

My concern here is the dashboards and if they are searching on event.kind  that may break some of the visualizations. Even if the ones we built work, I can't guarantee that custom dashboards will work

StacieClark-Elastic: Also, the use of the on_failure processor within a processor indicates that the original intent was to NOT stop processing. The stomping on the event.kind object is an unfortunate side effect.

efd6: There are two issues being discussed here. With Andrew's comment above about making the the new number 107, I'm OK with one of them (having a vector `event.kind` is my preferred solution and from memory elastic/ecs#2270 came out of discussion we had about the problem of orthogonal information clobbering each other in this field). I was sad that it got stopped.

The second impacts on the issue here, but is separate, and that is that irrespective of whether the error clobber a kind with utility to the user in a security setting, the ingest pipeline should not error unless the values are invalid, and they should error if they are. With the change here, they no longer error when the value is invalid. This may mask a problem with the data source that should be raised to the attention of the users. The solution to this second issue is to relax the constraints so that the [almost] valid values are converted to valid values and then the work continues. If the processors that are failing in the issue that prompted this change, then that's a real failure, and we should report it.

StacieClark-Elastic: The example they gave us was an unencoded URL that represented jus the query part of the URL. I've added logic to encode carets and spaces before trying to split into parts. I also check the original to see it starts with a single slash and remove the url.extension since it's incorrectly set in this case. 
I restored setting pipeline_error if anything is written to error.message

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -2935,19 +2963,38 @@ processors:
...
       copy_from: servicenow.event.url.display_value
+      description: Only copy the original URL if it has a scheme.
       ignore_empty_value: true
+  - remove:
+      field: url.extension
+      tag: remove_extension_if_url_only_contains_query
+      if: ctx.url?.extension != null && ctx.servicenow?.event?.url?.display_value =~ /^\/[a-z].*/
```
andrewkroh: From the test data I don't understand where this was needed. Given a display_value like `/somepath.do?sysparm_query=where^something`, what would be the reason to delete `url.extension: do`?

packages/servicenow/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.0.2"
+  changes:
+    - description: Added another date format for all display date times. Parses partial URLs
```
andrewkroh:
```suggestion
    - description: Add support for dates containing AM/PM, specifically support the `yyyy-MM-dd hh:mm:ss a` format.
```

packages/servicenow/changelog.yml
```diff
@@ -1,4 +1,9 @@
...
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/14193
```
andrewkroh, line 9:
```suggestion
      link: https://github.com/elastic/integrations/pull/14193
    - description: Improve the robustness of URL parsing.
      type: bugfix
      link: https://github.com/elastic/integrations/pull/14193
```

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -2935,19 +2963,38 @@ processors:
...
+      tag: encode_servicenow_event_url_display_value_carets
+      pattern: "\\^"
+      replacement: "%5E"
+      if: ctx.servicenow?.event?.url?.display_value != null && ctx.servicenow.event.url.display_value != ''
+  - gsub:
+      field: servicenow.event.url.display_value
+      tag: encode_servicenow_event_url_display_value_spaces
+      pattern: " "
+      replacement: "%20"
+      if: ctx.servicenow?.event?.url?.display_value != null && ctx.servicenow.event.url.display_value != ''
```
efd6:
```
  - script:
      tag: script_escape_urls
      if: ctx.servicenow?.event?.url?.display_value instanceof String
      params:
        should_encode:
          ' ': true
          '^': true
          # others?
      source: |-
        String encode(String path) {
          if (path.startsWith("http://") || path.startsWith("https://")) {
            // Not a path. Maybe some other test would be OK? !path.startsWith("/")
            return path;
          }
          StringBuilder sb = new StringBuilder();
          for (int i = 0; i < path.length(); i++) {
            char c = path.charAt(i);
            if (
              (char)'a' <= c && c <= (char)'z' ||
              (char)'A' <= c && c <= (char)'Z' ||
              (char)'0' <= c && c <= (char)'9' ||
              params.should_encode[String.valueOf(c)] != true
            ) {
              sb.append(c);
              continue;
            }
            sb.append('%');
            sb.append(Integer.toHexString((int)c))
          }
          return sb.toString();
        }
        ctx.servicenow.event.url.display_value = encode(ctx.servicenow.event.url.display_value);
```
This is more complex when there are only two, but I suspect there are more. Note that neither approach is completely correct, since that would depend on a complete URL parse, which we can't do (if we could, we wouldn't be here).

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: Because it's not an extension. It's part of a framework for queries. the single "/" marks the text as part of a query not a complete URL

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: I really didn't want to get into having to completely encode the string.

The data is bad. It's not a URL. It's part of a URL. The uri-parts parser will parse it when it begins with a single slash but then fails when it encounters unencoded data. The value of the data is low. 
We've designed ingest in a way that makes it impossible to indicate that low value data is bad without failing the entire pipeline.

My original fix was to ignore the failure to parse. If the data was good, we got it. If the data was bad we didn't. This is still the most robust fix but does not supply an error message.  An option would be to create a different field that captures url parsing errors without writing into error.message (which is what causes ingest to fail from the users perspective due to stomping on the event.kind value).

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: This no longer an issue as I am deferring to uri_parts to determine what parts go where. We still check for scheme before setting url.full as uri_parts will parse partial urls

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: Instead of trying to fix this, I am deferring back to uri_parts to determine what is a good or bad url. url.full will only be set if the original had a schema since uri_parts will parse incomplete URLS. Bad URLs will not cause a failure because after a team discussion, we've decided to use append for event.kind on non-terminal failures. This is known pattern in the integrations. We will revisit the issue of lists in event.kind with the breaking change board.

efd6: I expect that this will lead to support cases in the future.

packages/servicenow/changelog.yml
```diff
@@ -1,4 +1,11 @@
 # newer versions go on top
+- version: "1.0.2"
+  changes:
+    - description: |
+        Add support for dates containing AM/PM, specifically support the `yyyy-MM-dd hh:mm:ss a` format. 
+        Append pipeline_error to event.kind on non-terminal processor failures to not overwrite event.kind type.
```
efd6:
```suggestion
        Append pipeline_error to `event.kind` on non-terminal processor failures to not overwrite event.kind type.
```

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: It shouldn't. uri_parts recognizes what a good or bad url is. It's a bad URL. The actual problem is that the failing uri_parts was causing the event.kind to be overwritten and this integration relies on event.kind being set to the log type and it is not a static value. That is a failure it it's not there.  Now, it appends pipeline_error instead. The actual request was to ignore failures in uri_parts and not write any errors at all.

andrewkroh: If I test the messages that accompanied the original bug report, we'll still be producing errors with 1bc8149. If we know that ServiceNow is delivering a URL display_value that is not encoded (or not properly encoded), we should fall back to making a best effort recovery by applying proper encoding in on_failure and retrying.

If the URL values are known to behave badly and there's not much we or the user can do to address it, I think `ignore_failure` is fine. We'll set `url.original`, but the other url.* fields will be missing.

I did some testing and arrived at this...

<details>

```diff
diff --git a/packages/servicenow/data_stream/event/_dev/test/pipeline/test-event.log b/packages/servicenow/data_stream/event/_dev/test/pipeline/test-event.log
index 28bab48ba4..986e275ad3 100644
--- a/packages/servicenow/data_stream/event/_dev/test/pipeline/test-event.log
+++ b/packages/servicenow/data_stream/event/_dev/test/pipeline/test-event.log
@@ -43,3 +43,4 @@
 { "table_name":"task_ci","url":{"display_value":"//noscheme?query=values","value":"//noscheme?query=values"}}
 { "table_name":"task_ci","url":{"display_value":"/slash?query=values","value":"/slash?query=values"}}
 { "table_name":"task_ci","url":{"display_value":"noleadingslash?query=values","value":"noleadingslash?query=values"}}
+{ "table_name":"task_ci","url":{"display_value":"/slash.ext?query=fooISEMPTY^barIS LIKEbaz","value":"/slash.ext?query=fooISEMPTY^barIS LIKEbaz"}}
diff --git a/packages/servicenow/data_stream/event/_dev/test/pipeline/test-event.log-expected.json b/packages/servicenow/data_stream/event/_dev/test/pipeline/test-event.log-expected.json
index 9eddd96bb9..51432e61e2 100644
--- a/packages/servicenow/data_stream/event/_dev/test/pipeline/test-event.log-expected.json
+++ b/packages/servicenow/data_stream/event/_dev/test/pipeline/test-event.log-expected.json
@@ -6002,6 +6002,42 @@
                 "path": "noleadingslash",
                 "query": "query=values"
             }
+        },
+        {
+            "ecs": {
+                "version": "8.17.0"
+            },
+            "event": {
+                "category": [
+                    "configuration"
+                ],
+                "kind": "event",
+                "original": "{ \"table_name\":\"task_ci\",\"url\":{\"display_value\":\"/slash.ext?query=fooISEMPTY^barIS LIKEbaz\",\"value\":\"/slash.ext?query=fooISEMPTY^barIS LIKEbaz\"}}",
+                "timezone": "America/Los_Angeles",
+                "type": [
+                    "info"
+                ]
+            },
+            "servicenow": {
+                "event": {
+                    "table_name": "task_ci",
+                    "url": {
+                        "display_value": "/slash.ext?query=fooISEMPTY^barIS LIKEbaz",
+                        "value": "/slash.ext?query=fooISEMPTY^barIS LIKEbaz"
+                    }
+                }
+            },
+            "tags": [
+                "preserve_original_event",
+                "preserve_duplicate_custom_fields",
+                "hide_sensitive"
+            ],
+            "url": {
+                "extension": "ext",
+                "original": "/slash.ext?query=fooISEMPTY^barIS LIKEbaz",
+                "path": "/slash.ext",
+                "query": "query=fooISEMPTY^barIS LIKEbaz"
+            }
         }
     ]
 }
diff --git a/packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml b/packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
index c0f3c69e18..4f5aac2c6f 100644
--- a/packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
@@ -2967,13 +2967,46 @@ processors:
       copy_from: servicenow.event.time_zone.display_value
       ignore_empty_value: true
   - uri_parts:
+      if: >
+        ctx.servicenow?.event?.url?.display_value instanceof String &&
+        ctx.servicenow.event.url.display_value != ''
       field: servicenow.event.url.display_value
-      tag: uri_parts_url_full
-      if: ctx.servicenow?.event?.url?.display_value != null && ctx.servicenow.event.url.display_value != ''
-      on_failure:
-        - append:
-            field: error.message
-            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+      tag: uri_parts_url_display_value
+      on_failure:
+        - script:
+            description: >
+              URL values appear to be URI decoded, so attempt encoding the
+              necessary characters to give uri_parts the best chance at success.
+            tag: script_encode_url
+            source: |-
+              String encode(String path) {
+                StringBuilder sb = new StringBuilder();
+                for (int i = 0; i < path.length(); i++) {
+                  char c = path.charAt(i);
+                  if (
+                    (c >= (char)'a' && c <= (char)'z') ||
+                    (c >= (char)'A' && c <= (char)'Z') ||
+                    (c >= (char)'0' && c <= (char)'9') ||
+                    c != (char)' ' && c != (char)'^'
+                  ) {
+                    sb.append(c);
+                    continue;
+                  }
+                  sb.append('%');
+                  sb.append(Integer.toHexString((int)c));
+                }
+                return sb.toString();
+              }
+              field('_tmp.url').set(encode(ctx.servicenow.event.url.display_value));
+        - uri_parts:
+            tag: uri_parts_url_display_value_retry
+            field: _tmp.url
+            ignore_failure: true
+            keep_original: false
+        - set:
+            description: Set url.original to the unmodified value.
+            field: url.original
+            copy_from: servicenow.event.url.display_value
   - set:
       field: url.full
       tag: set_url_display_value_into_url_full
@@ -3335,8 +3368,8 @@ processors:
       if: ctx.tags == null || !ctx.tags.contains('preserve_duplicate_custom_fields')
   - remove:
       field:
-        - _conf.timestamp_field
-        - _conf.data_has_display_values
+        - _tmp
+        - _conf
         - servicenow.event.timestamp_field
       tag: remove_non_required_fields
       ignore_missing: true
@@ -3392,6 +3425,11 @@ processors:
       allow_duplicates: false
       if: ctx.error?.message != null
 on_failure:
+  - remove:
+      field:
+        - _tmp
+        - _conf
+      ignore_missing: true
   - append:
       field: error.message
       value: >-
```

</details>

StacieClark-Elastic: If we encode it and it still fails then we don't know why. If we write into error.message it won't cause a failure
 because event.kind will still be there.

packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -2969,11 +2969,52 @@ processors:
...
+                   value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+        - set:
+            tag: set_uri_original_if_url_decode_succeeds
+            if: ctx.url != null
+            description: Set url.original to the unmodified value.
+            field: url.original
+            copy_from: servicenow.event.url.display_value
+        - remove:
+            description: remove the temporary url field
+            field: _tmp.url
```
andrewkroh: I think it might better to write the temporary url value to `_tmp_url` and avoid the `_tmp` object, because `remove` won't clean up the empty `"_tmp": {}` object.

StacieClark-Elastic: I don't see an empty "_tmp": {} object in the test output.  I wonder why...
The very last out put in the expected log doesn't show it

andrewkroh: That is because of:

https://github.com/elastic/integrations/blob/7d3232e322975fb5bcee9cfe6479a85bdddcf086/packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml#L3317-L3319

but it's best to not rely on this.

andrewkroh (approved):

PR #14195, opened 2025-06-10
[Azure] signinlogs fix conditional_access_audience.application_id

### LINKED ISSUE #14178: [Azure Logs] Mapping conflict an dropped events because of field that can be both a string and an object.
Hello,

We just found out that some event for the Azure Signin Entra ID logs are being dropped because one of the fields can be sent by Microsoft as both a string or list and also as an object.

The resulting offending field is the field `azure.signinlogs.properties.conditional_access_audiences`, that is parsed from the field `conditionalAccessAudiences` in the siginlogs dataset. 

It seems that Microsoft can send events where this field has conflicting types, on some events it looks like this:

```
{
    "conditionalAccessAudiences": ["list-with-1-or-more-items"]
}
```

And in other events it looks like this

``` 
{
    "conditionalAccessAudiences": {
        "applicationId": "some-value",
        "audienceReasons": "some-values"
    }
}
```

It also seems that Microsoft made some changes on their side and stopped sending the information for the `audienceReasons` and also started to send the value of the `applicationId` directly on the `conditionalAccessAudiences` field, which leads to the conflict.

This field also does not have explict mapping on Elasticsearch side, so it will be dynamically mapped according to the first event it gets, so this lead to a conflict in Kibana as well in my case as one of the data streams got the mapping for this field as `keyword` instead of the object.

I was able to stop the events from being dropped by explictly mapping the field and using a custom ingest pipeline to rename the cases where `azure.signinlogs.properties.conditional_access_audiences` is a list into `azure.signinlogs.properties.conditional_access_audiences.application_id`.

This is the custom mapping I'm using:

```
PUT _component_template/logs-azure.signinlogs@custom
{
  "template": {
    "mappings": {
      "properties": {
        "azure": {
          "type": "object",
          "properties": {
            "signinlogs": {
              "type": "object",
              "properties": {
                "properties": {
                  "type": "object",
                  "properties": {
                    "conditional_access_audiences": {
                      "type": "object",
                      "properties": {
                        "application_id": {
                          "type": "keyword"
                        },
                        "audience_reasons": {
                          "type": "keyword"
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

And this is the custom ingest pipeline:

```
PUT _ingest/pipeline/logs-azure.signinlogs@custom
{
  "description": "custom signinlogs",
  "processors": [
    {
      "rename": {
        "ignore_failure": true,
        "ignore_missing": true,
        "field": "azure.signinlogs.properties.conditional_access_audiences",
        "target_field": "azure.signinlogs.properties.conditional_access_audiences.application_id",
        "if": "ctx?.azure?.signinlogs?.properties?.conditional_access_audiences instanceof List"
      }
    }
  ]
}
```

This should be applied to the Ingest pipeline and mapping for the integration.

Another issue is that this conflict may break a bult-in security rule: [Entra ID Device Code Auth with Broker Client](https://github.com/elastic/detection-rules/blob/main/rules/integrations/azure/credential_access_entra_id_device_code_auth_with_broker_client.toml)

This rule uses the field `azure.signinlogs.properties.conditional_access_audiences.application_id`, and since this can lead to a mapping conflict in Kibana the queries using this field may not work.

peterydzynski:
## Proposed commit message
This PR resolves the following issue: https://github.com/elastic/integrations/issues/14178

Microsoft apparently made a change to the format of the signinlogs and sometimes sends a list of items or a map. If a list is sent, the items in the list are application id's. This PR adds a processor to set the value of this list to the appropriate key under `azure.signinlogs.properties.conditional_access_audience.application_id`.

## Checklist

## Author's Checklist

N/A

## Related issues
- Closes #14178

-

### DIFF
--- a/packages/m365_defender/docs/README.md
+++ b/packages/m365_defender/docs/README.md
@@ -425,7 +425,6 @@ An example event for `alert` looks as following:
         ]
     }
 }
-
 ```
 
 **Exported fields**
@@ -1292,7 +1291,6 @@ An example event for `incident` looks as following:
         }
     }
 }
-
 ```
 
 **Exported fields**
@@ -1681,7 +1679,6 @@ An example event for `vulnerability` looks as following:
         "title": "An inappropriate implementation in the Downloads feature of Google Chrome versions prior to 135.0.7049.52 could allow a remote attacker to perform UI spoofing via a crafted HTML page. This vulnerability, classified with a low severity by Chromium, may also enable bypassing security restrictions when a victim visits a specially crafted website."
     }
 }
-
 ```
 
 **Exported fields**


packages/azure/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: "1.26.1"
+  changes:
+    - description: Fix bug in handling of conditional_access_audience field in signinlogs data stream.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/
```
efd6:
```suggestion
      link: https://github.com/elastic/integrations/pull/14195
```

packages/azure/data_stream/signinlogs/_dev/test/pipeline/test-signinlogs-raw.log
```diff
@@ -1,2 +1,3 @@
...
..."riskDetail":"none","riskEventTypes":[],"riskEventTypes_v2":[],"riskLevelAggregated":"none","riskLevelDuringSignIn":"none","riskState":"none","rngcStatus":0,"servicePrincipalId":"","sessionId":"665694e7-26fc-4216-bf7e-e5adddc7a2bf","sessionLifetimePolicies":[],"signInTokenProtectionStatus":"none","ssoExtensionVersion":"","status":{"additionalDetails":"MFA requirement satisfied by claim in the token","errorCode":0},"tenantId":"6cb7db5b-fc26-4548-8eae-ca52f13810d4","tokenIssuerName":"","tokenIssuerType":"AzureAD","tokenProtectionStatusDetails":{"signInSessionStatus":"unbound","signInSessionStatusCode":1002},"uniqueTokenIdentifier":"OTMzZjIwYzAtZWZkZi00NzdmLTk1ODYtZTVjYzY3NmYyZTAw","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36","userDisplayName":"Elastic Test","userId":"665694e7-26fc-4216-bf7e-e5adddc7a2bf","userPrincipalName":"mpliftrelastic20210901@outlook.com","userType":"Member"},"resourceId":"/tenants/665694e7-26fc-4216-bf7e-e5adddc7a2bf/providers/Microsoft.aadiam","resultSignature":"SUCCESS","resultType":"0","tenantId":"797f4846-ba00-4fd7-ba43-dac1f8f63013","time":"2025-06-10T19:52:50.4512146Z"}
```
efd6, line 3: I think we need a test for the object form of the data, as is shown in the second snippet in the [issue](https://github.com/elastic/integrations/issues/14178#issue-3124927877).

efd6 (commented):
Because there are changes to the field definitions, the README will need to be updates. This is done by running `elastic-package build`.

packages/azure/data_stream/signinlogs/_dev/test/pipeline/test-signinlogs-raw.log
peterydzynski: @efd6 darn was hoping I could sneak this through without 🙈 it's definitely a good idea though, I'll grab one and add.

peterydzynski:
This is waiting on a test event from me. Haven't had a chance to dig one up but will try to get to it this week. 👍🏿

PR #14210, opened 2025-06-11
[filestream]: fix fingerprint file identity options not taking effect

mauri870:
## Checklist

1. `gh pr checkout 14210`
2. `cd packages/filestream`
3. `elastic-package check`
4. `elastic-package install --zip ../../build/packages/filestream-1.1.5.zip -v`
5. Add filestream integration to an agent, enable Filestream file identity under Advanced settings, set offset and length
6. Capture diagnostics from the running agent, in the configuration it should reflect the correct prospector settings.

### DIFF
--- a/packages/filestream/data_stream/generic/_dev/test/system/test-filestream-fingerprint-config.yml
+++ b/packages/filestream/data_stream/generic/_dev/test/system/test-filestream-fingerprint-config.yml
@@ -1,4 +1,4 @@
-service: filestream-fingerprint-logfile
+service: filestream-logfile
 input: filestream
 data_stream:
   vars:


packages/filestream/data_stream/generic/agent/stream/filestream.yml.hbs
```diff
@@ -140,9 +140,8 @@ harvester_limit: {{harvester_limit}}
 
 {{#if fingerprint }}
 prospector.scanner.fingerprint.enabled: true
-file_identity.fingerprint.enabled: true
-file_identity.fingerprint.offset: {{ fingerprint_offset }}
-file_identity.fingerprint.length: {{ fingerprint_length  }}
+prospector.scanner.fingerprint.offset: {{ fingerprint_offset }}
```
mauri870, line 144: Q: I'm not sure if we need `file_identity.fingerprint: ~` here as well.

belimawr: Yes, we do. That is what enables fingerprint as file identity. `prospector.scanner.fingerprint.*` tells the prospector to generate a fingerprint for the files that is later used as file identity.

packages/filestream/data_stream/generic/manifest.yml
```diff
@@ -88,7 +88,7 @@ streams:
         type: bool
         title: Recursive Glob
         description: |
-          Enable expanding ** into recursive glob patterns. With this feature enabled, the rightmost ** in each path is expanded into a fixed number of glob patterns. For example: /foo/** expands to /foo, /foo/*, /foo/*/*, and so on. If enabled it expands a single ** into a 8-level deep * pattern.
+          Enable expanding **into recursive glob patterns. With this feature enabled, the rightmost** in each path is expanded into a fixed number of glob patterns. For example: /foo/** expands to /foo, /foo/*, /foo/*/*, and so on. If enabled it expands a single ** into a 8-level deep * pattern.
```
belimawr: In this line `**` is actually referring to the glob `**`, like in `/foo/**/bar`.

What do you think of this:
```suggestion
          Enable expanding `**` into recursive glob patterns. With this feature enabled, the rightmost `**` in each path is expanded into a fixed number of glob patterns. For example: `/foo/**` expands to `/foo`, `/foo/*`, `/foo/*/*`, and so on. If enabled it expands a single `**` into a 8-level deep `*` pattern.
```

mauri870: Ah, thanks for catching that! Yes, much better :)

belimawr (approved):

belimawr:
I'm wondering if we should add a test for the length/offset configuration.

We can set the variables, define the log file and how many events we expect, there is an example [here](https://github.com/elastic/integrations/blob/main/packages/filestream/data_stream/generic/_dev/test/system/test-filestream-config.yml)

mauri870:
> I'm wondering if we should add a test for the length/offset configuration.
> 
> We can set the variables, define the log file and how many events we expect, there is an example [here](https://github.com/elastic/integrations/blob/main/packages/filestream/data_stream/generic/_dev/test/system/test-filestream-config.yml)

I think that is totally valid, I'll work on adding a test.

mauri870:
@belimawr Added a test, can you check if it is what you had in mind?

belimawr:
> @belimawr Added a test, can you check if it is what you had in mind?

That's prefect! Thank you!!

PR #14211, opened 2025-06-11
entityanalytics_okta: map roles data to ECS

### LINKED ISSUE #13165: [Okta Entity Analytics]: Add the `user.roles` field to user entities containing their Okta admin role assignments
### Integration Name

Okta Entity Analytics [entityanalytics_okta]

### Dataset Name

entityanalytics_okta.user

### User Goal

We wish to know which users are privileged, in order to do this we need to know their Okta Admin roles.

For each user, we wish to know their okta admin roles assigned individually or through groups, e.g the content of this page:

Image

it looks like there is a [separate API](https://developer.okta.com/docs/api/openapi/okta-management/management/tag/RoleAssignmentAUser/#tag/RoleAssignmentAUser/operation/listUsersWithRoleAssignments) for this but it is not returned in the GET users API which we use today.

We would hope to have this in the [ECS user.roles field](https://www.elastic.co/guide/en/ecs/current/ecs-user.html#field-user-roles).

### Existing Features

We do not currently provide role information for users.

### What did you see?

n/a

### Anything else?

_No response_

efd6:
## Proposed commit message

See title.

> [!NOTE]
> Marking as bug since it fixes absent mapping and restricts the version to stacks that have the agent feature.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #13165


packages/entityanalytics_okta/data_stream/entity/elasticsearch/ingest_pipeline/user.yml
```diff
@@ -246,6 +246,55 @@ processors:
...
+      if: ctx.entityanalytics_okta?.roles instanceof List
+      processor:
+        date:
+          field: _ingest._value.last_updated
+          target_field: _ingest._value.last_updated
+          tag: date_user_roles.lastUpdated
+          formats:
+            - ISO8601
+          on_failure:
+            - rename:
```
chemamartinez, line 294: Why a rename here to the original form in the failure case?

packages/entityanalytics_okta/data_stream/entity/fields/fields.yml
```diff
@@ -218,3 +218,27 @@
...
+          description: Role status.
+        - name: created
+          type: date
+          description: When the role was created.
+        - name: last_updated
+          type: date
+          description: When the role was last updated.
+        - name: assignment_type
+          type: keyword
+          description: The Okta type the role can is assigned to.
```
chemamartinez:
```suggestion
          description: The Okta type the role can be assigned to.
```

packages/entityanalytics_okta/data_stream/user/fields/fields.yml
```diff
@@ -165,3 +165,27 @@
...
+          description: Role status.
+        - name: created
+          type: date
+          description: When the role was created.
+        - name: last_updated
+          type: date
+          description: When the role was last updated.
+        - name: assignment_type
+          type: keyword
+          description: The Okta type the role can is assigned to.
```
chemamartinez:
```suggestion
          description: The Okta type the role can be assigned to.
```

packages/entityanalytics_okta/data_stream/entity/elasticsearch/ingest_pipeline/user.yml
efd6: If the format cannot be parsed, this retains the value as a keyword. It prevents non-mapping while retaining data. The alternative is to write the error into error.message. Happy to do the latter if you would prefer.

andrewkroh (approved):
LGTM

PR #14212, opened 2025-06-12
crowdstrike: add support for multi-resource streams

### LINKED ISSUE #13953: [CrowdStrike Falcon] Streaming started processing all the backlog after agent upgrade.
Hello,

We just upgraded our agent from `8.17.3` to `8.18.1` and we have a policy where we use the Crowdstrike integration to get falcon logs using the `streaming` input.

After the upgrade we thougt that we stopped getting logs, but after some quick investigation we saw that the agent started getting the backlog again instead of resuming from the last know log.

Image

This is problematic because it may take some time to reprocess all the data and during this time we will have no visibility on this dataset, which is pretty important, it will also retrigger some alerts as they are configured using `event.ingested`, increasing the work for the operations team.

I'm also no sure if this is will lead to duplicate data or if it is getting events that it missed the first time, but both would be a problem.

Is this a know issue? Was this mentioned in any changelog? Could not find anything, but maybe I missed it.

Is there any way to set a start date for this integration?

efd6:
## Proposed commit message
```
crowdstrike: add support for multi-resource streams

elastic/beats#44548 added support for multi-resource stream. This
updates the CEL program that processes the events provided by the stream
so that it is able to handle the new cursor structure. It is able to
distinguish old agents from the new multi-resource aware agents by the
presence of the feed field in state.

When the agent is upgraded from the old state form to the new form, it is
expected that the integration will recollect the existing data since the
cursors are not compatible with each other.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #13953
- Relates elastic/beats#44548

### DIFF
--- a/packages/crowdstrike/changelog.yml
+++ b/packages/crowdstrike/changelog.yml
@@ -1,7 +1,12 @@
 # newer versions go on top
+- version: "1.75.1"
+  changes:
+    - description: Add support for multi-resource falcon hose streams.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/14212
 - version: "1.75.0"
   changes:
-    - description: Add additional parsed fields in alert datastream
+    - description: Add additional parsed fields in alert datastream.
       type: enhancement
       link: https://github.com/elastic/integrations/pull/14143
 - version: "1.74.0"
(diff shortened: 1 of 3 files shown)


chemamartinez (approved):
LGTM

andrewkroh (commented):
The program makes sense to me based on my understanding of the beats change.

I am wondering if it would be useful to have a mock test that has more than one feed? Given the only assertion we have available is hit count, I'm not sure this testing would catch anything more.

efd6:
I think that's worth looking into, but yeah, because we don't have any cursor-based testing, it's impossible to know that it failed.

PR #14223, opened 2025-06-16
entityanalytics_ad: add support for device entities

### LINKED ISSUE #13915: entityanalytics_ad: allow configuration of device entitly collection
Currently the provider collects human users only. Having device entities is an important feature that needs to have the input collect computer entries. With the input added, we will need to wire the integration into the new collection behaviour to allow these to be ingested by the integration.

efd6:
## Proposed commit message

```
entityanalytics_ad: add support for device entities

The data stream is split to do this, essentially in the same way that
the entityanalytics_okta integration move from v1 to v2 was managed.

The pipeline is split into three with calls to a common pipeline. The
third, non-user/non-device pipeline is intended to act as a fallback.
Currently, the differentiated pipelines are largely identical, but the
separation allows easier adjustment in the future.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #13915


nicpenning:
It would be a great add to have host.os.* and host.* fields as applicable such as host.name.

nicpenning:
Based on the output of the entityanalytics_ad integration/agent version prior to computer objects not being sent Here are fields for operating system, here are some fields worth using for host.os.*

entityanalytics_ad.user.operatingSystem : Windows 11 Enterprise
entityanalytics_ad.user.operatingSystemVersion : 10.0 (22631)


Notice how they have the sub-property user: entityanalytics_ad.**user**.operatingSystem - So perhaps this will be corrected in the new data stream.

Also, to get this field, you need to add these fields accordingly in the configuration or pull all fields to include them.

host.* fields can be derived from a combination of many fields, but here are some examples:

entityanalytics_ad.user.cn : COMPUTER1
entityanalytics_ad.user.name : COMPUTER1
entityanalytics_ad.user.sam_account_name : COMPUTER1$
entityanalytics_ad.user.dNSHostName : COMPUTER1.org.local
user.domain : org.local

Sorry this is a bit of a mess - With the upgrade of the agent/integration, we no longer receive these details. We would need to downgrade to an earlier working version to get more data as needed. Hopefully those fields may be enough to identify what is needed to at least get the computer object data stream out the door and can be improved later.

efd6:
With the new ingest pipeline, `entityanalytics_ad.user` is [renamed](https://github.com/elastic/integrations/pull/14223/files#diff-1f069cadec4f8a6ded58c12d4db53a634e49448cd515e7194eb2de8f3b4bf896R53-R59) to `entityanalytics_ad.device` for devices. So the `.user.` parts of the comment above are addressed.

For the second part, I think it might be best to wait until the new release is available and make a follow-up change to the integration here when we have a concrete idea of which fields are useful to use to populate for host details. Though I would imagine that give the ingest pipeline as it will exist, we would have
- `entityanalytics_ad.device.operatingSystem` → `host.os.full` (probably with distructuring to other fields)
- `entityanalytics_ad.device.operatingSystemVersion` → `host.os.version`
- `entityanalytics_ad.device.cn`/`entityanalytics_ad.user.name` -append unique?→ `host.hostname`
- `entityanalytics_ad.device.dNSHostName` →`lc`?→ `host.name`
- `entityanalytics_ad.device.domain` → `host.domain`

nicpenning:
Sounds good. Let's get this out the door then add the OS and other details later!

packages/entityanalytics_ad/_dev/build/docs/README.md
```diff
@@ -2,11 +2,18 @@
 
 This Active Directory Entity Analytics integration allows users to securely stream User Entities data to Elastic Security via the Active Directory LDAP look-ups. When integrated with Elastic Security, this valuable data can be leveraged within Elastic for risk-scoring scenarios (e.g., context enrichments) and detecting advanced analytics (UBA) use cases.
 
+## Upgrading to v0.15.0 from v0.14 and lower of the integration
+
+In v0.15.0 of the integration the user and device data was split into separate data streams. The data ingested into your index will be the same but you may need to update device searches if you were using them.
+
+**NOTE**: When you upgrade from v1 you will need to reconfigure the integration and enable it due to internal changes in the package. See [Resolve conflicts](https://www.elastic.co/guide/en/fleet/current/upgrade-integration.html#resolve-conflicts) in the Fleet documentation for details.
```
kcreddy: What does v1 mean in this context? Looks like integration is in beta.

packages/entityanalytics_ad/data_stream/device/fields/base-fields.yml
```diff
@@ -0,0 +1,20 @@
+- name: data_stream.type
```
kcreddy, line 1: Possible to use `external: ecs` on all of them?

packages/entityanalytics_ad/data_stream/entity/elasticsearch/ingest_pipeline/common.yml
```diff
@@ -1,27 +1,6 @@
 ---
 description: Pipeline for processing User logs.
```
kcreddy: Looks like this pipeline is not just for User data anymore. Possibly update description to reflect accordingly,

packages/entityanalytics_ad/data_stream/entity/elasticsearch/ingest_pipeline/entity.yml
```diff
@@ -0,0 +1,67 @@
+---
+description: Pipeline for processing User logs.
```
kcreddy:
```suggestion
description: Pipeline for processing Entity logs.
```

packages/entityanalytics_ad/_dev/build/docs/README.md
efd6: Changed to "a version prior to v0.15.0".

packages/entityanalytics_ad/data_stream/device/fields/base-fields.yml
efd6: Done.

packages/entityanalytics_ad/data_stream/entity/elasticsearch/ingest_pipeline/common.yml
efd6: Done.

packages/entityanalytics_ad/data_stream/entity/elasticsearch/ingest_pipeline/entity.yml
efd6: Done.

packages/entityanalytics_ad/data_stream/entity/elasticsearch/ingest_pipeline/device.yml
```diff
@@ -0,0 +1,67 @@
...
+      tag: set_event_kind
+      value: asset
+  - set:
+      field: event.category
+      tag: set_event_category
+      value: ['iam']
+  - set:
+      field: event.type
+      tag: set_event_type
+      value: ['user','info']
```
kcreddy: For device events, don't we need to set `host.*` fields unlike https://github.com/elastic/integrations/blob/main/packages/entityanalytics_entra_id/data_stream/entity/elasticsearch/ingest_pipeline/device.yml

packages/entityanalytics_ad/data_stream/entity/elasticsearch/ingest_pipeline/device.yml
```diff
@@ -0,0 +1,67 @@
...
+description: Pipeline for processing Device logs.
+processors:
+  - set:
+      field: event.kind
+      tag: set_event_kind
+      value: asset
+  - set:
+      field: event.category
+      tag: set_event_category
+      value: ['iam']
```
kcreddy: Should this be `host` for device events?

packages/entityanalytics_ad/data_stream/entity/elasticsearch/ingest_pipeline/device.yml
efd6: See https://github.com/elastic/integrations/pull/14223#issuecomment-3100103561

packages/entityanalytics_ad/data_stream/entity/elasticsearch/ingest_pipeline/device.yml
efd6: This is vexed. I've updated it, but because of the ActiveDirectory semantics, I don't think we can actually properly reflect the reality within ECS. Devices are not really hosts in this universe.

kcreddy: Thanks for clarification.

kcreddy (approved):
LGTM.

PR #14227, opened 2025-06-16
[jamf_pro] Tidy variable descriptions and input configuration

### LINKED ISSUE #13889: jamf_pro: inventory data stream uses non-standard logging layout
The jamf_pro.inventory data stream places request trace logs in [`logs/jamf_pro-versions/http-request-trace-*.ndjson`](https://github.com/elastic/integrations/blob/c0c77f7a797ae0c47324b8f37b04a5c88ba24a47/packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs#L7) rather than the canonical `logs/cel/http-request-trace-*.ndjson`

Additionally, URL construction/URL documentation result in an API target with a `//` in the path (which Jamf appears to tolerate). This happens because the documentation states that the URL [should have a final slash](https://github.com/elastic/integrations/blob/c0c77f7a797ae0c47324b8f37b04a5c88ba24a47/packages/jamf_pro/manifest.yml#L49), but the URLs used are constructed [here](https://github.com/elastic/integrations/blob/c0c77f7a797ae0c47324b8f37b04a5c88ba24a47/packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs#L3) and [here](https://github.com/elastic/integrations/blob/c0c77f7a797ae0c47324b8f37b04a5c88ba24a47/packages/jamf_pro/data_stream/inventory/agent/stream/cel.yml.hbs#L3) assuming that this does not exist. Due to a combination of Hyrum's Law, Handlebars weakness, and non-reading of documentation I am not sure that this should be fixed;
1. changing the documentation is unlikely to have any impact
2. changing the behaviour of the code is not possible in a consistent manner since while we can make the CEL code correctly handle incorrectly added trailing slashes, but the auth token URL is constructed by Handlebars, and it cannot do any work.
3. It works, though ugly, at the moment.

chrisberkhout:
## Proposed commit message

```
[jamf_pro] Tidy variable descriptions and input configuration

- Improve variable descriptions in the manifest.
- Use the standard CEL request tracer file name.
- Move API URL construction into the CEL so trimming has an effect.
```

## Checklist

## Related issues

- Closes #13889


chrisberkhout:
> I think it might be worth bumping the kibana.version so that we can add support for trace deletion at the same time. If we fix the path in a PR staged some time after that, then we can actually clean up any traces that are lying around (assuming users don't jump the intermediate version).

Done. Proposed commit message updated.

efd6 (approved):
Thanks

PR #14232, opened 2025-06-16
[O365_metrics] Add `active_users_services_user_counts` Dashboard

lucian-ioan:
## Proposed commit message
Add `active_users_services_user_counts` dashboard.

I think active/inactive graphs bar charts work fine. For it to have only one graph and a control at the top, the data structure would need to be a different format.

Having one bar chart/other vis for each specific service seems unnecessary since this data has a fixed number of only 6 services.

## Checklist

## Author's Checklist

- [ ] Update other dashboards descriptions hyperlinks.
- [ ] Push data into elasticsearch to simulate active counts and more than one report.

## Related issues

- 

## Screenshots

image


Linu-Elias (commented):
I think we should hide the y-axis  labels as we have multiple services
image

muthu-mps:
@lucian-ioan - The `Report Last Refresh Date` in the metrics panel, how this is helpful when the dashboard time range is selected for `last 15 mins` or `1 hour`? Can this be in a table view?

lucian-ioan:
@muthu-mps Report Last Refresh Date is showing how recent the data is. In other words, when was the data last updated which I think is useful to know.

Instead of a table view, perhaps a control on Report Refresh Date is more useful if we want users to filter data based on that while also seeing all of the dates the refreshes happened.

ritalwar:
> Instead of a table view, perhaps a control on Report Refresh Date is more useful if we want users to filter data based on that while also seeing all of the dates the refreshes happened.

I've noticed that the existing report based dashboards in the package include control over the report refresh date. I think we should maintain the same experience for other report-based dashboards as well.

PR #14235, opened 2025-06-17
abnormal_security: allow configuration of a grace period for ai_security_mailbox_not_analyzed

### LINKED ISSUE #14219: Abnormal Security: Add support for "AI Security Mailbox Not Analyzed" look-back offset
### Integration Name

Abnormal Security [abnormal_security]

### Dataset Name

ai_security_mailbox_not_analyzed

### Integration Version

1.6.0

### Agent Version

8.16.1

Ubuntu

### User Goal

Eliminate messages that only briefly show up in the not analyzed queue.

Due to delays in the amount of time it takes for Abnormal Security to process messages, messages will occasionally be available via the "not analyzed" endpoint and then no longer be available there after a short period of time.

Quote from Abnormal Security Support
> Generally our system aims to process 95% of messages within 60 seconds. However, messages that initially appear in the "Not Analyzed" state might take longer to be fully processed. Expected ETA is unconventional. 
> For your consideration about implementing a delay: Based on observed patterns, a delay of approximately 5-10 minutes would likely be sufficient to account for most messages transitioning from "Not Analyzed" to successfully scanned but sometime it may can go beyound when AI need more analysis to conclude the Analysis. 

The above can be achieved by adding a configurable parameter for "look-back offset" or similar so that the time range which the integration is retrieving logs for is not up to the current second.

### Existing Features

The configuration for the integration does not support a configurable duration to offset the look-back duration. Thus, messages are always captured up to what is available up to the current second.

### What did you see?

From my understanding, logs are retrieved up to the current second.
See below snippet where applicable:
https://github.com/elastic/integrations/blob/d976a3419a4ba85f1d83f41575956c6b1bd6b0f0/packages/abnormal_security/data_stream/ai_security_mailbox_not_analyzed/agent/stream/cel.yml.hbs#L27-L28

### Anything else?

_No response_

efd6:
## Proposed commit message
```
abnormal_security: allow configuration of a grace period for ai_security_mailbox_not_analyzed

The "Not Analyzed" data source may have a delayed availability for
messages, with advice from Abnormal Security being that 95% of messages
are ready by 60s, but that a 5-10 minute delay would likely be sufficient
to account for most messages transitioning from "Not Analyzed" to
successfully scanned. Given this range of expectations, provide a
mechanism for the user to tune the agent's response. The default
behaviour is to retain no grace period, the current behaviour.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #14219

### DIFF
--- a/packages/abnormal_security/data_stream/ai_security_mailbox_not_analyzed/manifest.yml
+++ b/packages/abnormal_security/data_stream/ai_security_mailbox_not_analyzed/manifest.yml
@@ -14,6 +14,13 @@ streams:
         show_user: true
         default: 2160h
         description: How far back to pull the AI Security Mailbox Not Analyzed messages from Abnormal Security API. Defaults to 90 days (2160h) before end. Supported units for this parameter are h/m/s.
+      - name: wait_interval
+        type: text
+        title: Recent Message Grace Interval 
+        multi: false
+        required: true
+        show_user: true
+        description: How long to wait before attempting to collect recent messages. This option allows the Abnormal Security API to complete analysis of messages before the agent attempts to collect them. This should not be greater than the initial interval. Supported units for this parameter are h/m/s.
       - name: interval
         type: text
         title: Interval
--- a/packages/abnormal_security/changelog.yml
+++ b/packages/abnormal_security/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.8.1"
+  changes:
+    - description: Prevent loss of recent unprocessed messages by the `ai_security_mailbox_not_analyzed` data stream.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/14235
 - version: "1.8.0"
   changes:
     - description: Enrich threat events with attachment and link details.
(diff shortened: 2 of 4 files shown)


packages/abnormal_security/data_stream/ai_security_mailbox_not_analyzed/manifest.yml
```diff
@@ -14,6 +14,13 @@ streams:
         show_user: true
         default: 2160h
         description: How far back to pull the AI Security Mailbox Not Analyzed messages from Abnormal Security API. Defaults to 90 days (2160h) before end. Supported units for this parameter are h/m/s.
+      - name: wait_interval
+        type: text
+        title: Recent Message Grace Interval 
+        multi: false
+        required: true
+        show_user: true
+        description: How long to wait before attempting to collect recent messages. This option allows the Abnormal Security API to complete analysis of messages before the agent attempts to collect them. This should not be greater than the initial interval. Supported units for this parameter are h/m/s.
```
efd6, line 23: I'm not wildly happy with the wording here. So I'd welcome any suggestions.

kcreddy: Thats comprehensible to me.

Issue #13500, opened 2025-04-10
[Cloudtrail] flattened fields response_elements, response_elements, additional_eventdata increase storage


norrietaylor:
After investigating storage costs for the cloud trail integration, it was shown that we are storing some large fields multiple times. 

`cloudtrail.response_elements`, `cloudtrail.request_elements` , `cloudtrail.additional_eventdata` fields.  

We store them five times: _source, keyword, text, and then flattened, which stores them twice (keyed and unkeyed)

After discussion with @efd6, the recommendation is to add a configuration value to disable the flattened fields.

strawgate:
I generally think these are useful because of the flattened fields -- I think the text field is somewhat questionable but somewhat makes sense, I'm struggling to understand however what someone does with 4kb of json in a keyword field

leandrojmp:
A feedback as user, I find them pretty useful and we use both the flattened and text field in security rules for different cases and needs.

If I understand correctly the proposal is to add a way for the user to disable it if wanted, similar to what is done if the user wants to keep the original event or duplicated fields in some integrations, right?

efd6:
I'd like to hear feedback. So far, what I have is an option to not retain the flattened field group. This is slightly more involved than you might expect, since it's not just deleting it at the end of the pipeline, so that we don't do fruitless work. If we need to have multiple of these, I'm less inclined to have multiple options, preferring to only have the one, since it avoids ingest work that's not otherwise needed, but give advice in the package documentation to add an `@custom` pipeline that deletes fields that are not wanted.

leandrojmp:
@efd6 @strawgate what exactly is being considered for removal/changes?

The 2 main fields from Cloudtrail are `request_parameters` and `response_elements`, those are present in two ways, one is a flattened form with the entire json like `aws.cloudtrail.flattened.request_parameters` and the other is in a multi-field as `aws.cloudtrail.request_parameters` for example.

Both are required, the `aws.cloudtrail.request_parameters`  we use more the `.text` field to search for strings inside of it, the `keyword` field is used sometimes when the value is just a single keyword in some searchs/dashboards.

In our use case both formats are required, the `aws.cloudtrail.request_parameters` could be changed to be just a text/match_only_text, but if it is entirely removed it would break a lot of things.

strawgate:
Do you by chance have an example of it containing a single value that is then usable as a keyword?

Right now the keyword stores up to 8kb of json.

[...]

Issue #14200, opened 2025-06-11
apache,aws.cloudtrail,nginx,windows,zeek: removed input has agent template remaining


efd6:
In #13246 HTTP JSON inputs were removed from:

- obs-infraobs-integrations
    - [ ] apache
    - [ ] aws.cloudtrail
    - [ ] nginx
- sec-windows-platform/elastic-agent-data-plane
    - [ ] windows
- sec-deployment-and-devices
    - [ ] zeek

However, the agent templates remain. These should be deleted.

PR #13246, opened 2025-03-21
#11767 - Remove deprecated httpjson from packages zeek, apache, system, windows, aws.cloudtrail and nginx

qcorporation:
## Parent Ticket
https://github.com/elastic/integrations/issues/11767

## Proposed commit message
Remove httpjson from packages zeek, apache, system, windows, aws.cloudtrail and nginx
- removed docker service if applicable
- removed system tests if they use the httpjson input that's been deprecated
- removed httpjson input from the manifest file
- update each integration to a new major version as this can potentially be a breaking change

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally
Ran system and pipeline test and validated no failures

## Exceptions
There is still the `windows` integration where all the system tests are based upon the `httpjson` input
Speaking with the teams with code ownership to get recommendations

## Screenshots
Screenshot 2025-03-21 at 3 24 56 PM
Screenshot 2025-03-21 at 3 03 27 PM
Screenshot 2025-03-21 at 2 49 48 PM
Screenshot 2025-03-21 at 2 11 15 PM

PR #14236, opened 2025-06-17
aws: allow user specification of fields to retain in the cloudtrail data stream

efd6:
## Proposed commit message
```
aws: allow user specification of fields to retain in the cloudtrail data stream

Storage of the response_elements, request_parameters and additional_eventdata
is a potentially significant cost, but different users have different
requirements for their present, so there is no ideal approach. Given
that it is likely that this optimisation will be a common desire,
provide a UI option to allow users to easily configure this behaviour
without the requirement of adding processors to remove the fields in an
@custom pipeline. Note also that there is a TODO in the pipeline
addition here to move from a remove after creation model, spending
fruitless work, to a non-creation model, which would not be possible to
implement in an @custom pipeline.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- For #13500


packages/aws/data_stream/cloudtrail/manifest.yml
```diff
@@ -189,6 +189,30 @@ streams:
...
+        type: select
+        title: Event Information Fields to Retain
+        multi: false
+        required: false
+        show_user: false
+        description: >-
+          Cloudtrail `response_elements`, `request_parameters` and `additional_eventdata` data can
+          be placed in keyword and text fields as JSON, and in flattened fields. Depending on requirements
+          This configuration determines which fields will be retained in the final document. The Minimal
+          option retains the minmal set of fields required for the Security Detection Engine rules.
```
kcreddy:
```suggestion
          Cloudtrail `response_elements`, `request_parameters` and `additional_eventdata` data can
          be placed in keyword and text fields as JSON, and in flattened fields. Depending on requirements
          this configuration determines which fields will be retained in the final document. The Minimal
          option retains the minimal set of fields required for the Security Detection Engine rules.
```

packages/aws/data_stream/cloudtrail/manifest.yml
```diff
@@ -325,3 +349,24 @@ streams:
...
+          This configuration determines which fields will be retained in the final document.
+        options:
+          - text: Keyword and Flattened
+            value: all
+          - text: Keyword
+            value: keyword
+          - text: Flattened
+            value: flattened
+          - text: Neither
+            value: none
```
kcreddy, line 373: Is `minimal` not applicable for this input?

efd6: No, I just forgot to add it; it came in later.

kcreddy (approved):
LGTM after resolving conflict. Thanks!

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1712,9 +1712,68 @@ processors:
...
+        required_flattened_fields:
+          - aws.cloudtrail.flattened.additional_eventdata.SSEApplied
+          - aws.cloudtrail.flattened.request_parameters.cidrIp
+          - aws.cloudtrail.flattened.request_parameters.dryRun
+          - aws.cloudtrail.flattened.request_parameters.fromPort
+          - aws.cloudtrail.flattened.request_parameters.includeDeprecated
+          - aws.cloudtrail.flattened.request_parameters.policyArn
+          - aws.cloudtrail.flattened.request_parameters.serialNumber
+          - aws.cloudtrail.flattened.request_parameters.withDecryption
+          - aws.cloudtrail.flattened.request_parameters.x-amz-server-side-encryption-customer-algorithm
```
romulets, line 1739: I personally worry about this duplicated list from the detection engine. This will be easily missed in future iterations. Do you have thoughts on process to avoid it? Or maybe can we automate fetching this list on pre-commit? Or automate tests to verify consistency?

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1712,9 +1712,68 @@ processors:
...
+      source: |-
+        def set(Map root, String path, def v) {
+          String[] elems = path.splitOnToken('.');
+          def dst = root;
+          for (int i = 0; i < elems.length-1; i++) {
+            dst = dst.computeIfAbsent(elems[i], _ -> new HashMap());
+          }
+          dst[elems[elems.length-1]] = v;
```
romulets, line 1747: have you tried just assigning the `.` style to a map and ingesting it as flattened? If I'm not mistaken that is what happens under the hood, maybe this map build is not needed

packages/aws/data_stream/cloudtrail/manifest.yml
```diff
@@ -189,6 +189,30 @@ streams:
...
+          This configuration determines which fields will be retained in the final document. The Minimal
+          option retains the minmal set of fields required for the Security Detection Engine rules.
+        options:
+          - text: Keyword and Flattened
+            value: all
+          - text: Keyword
+            value: keyword
+          - text: Flattened
+            value: flattened
+          - text: Minimal
```
romulets, line 210: Is there a way to explain to customers what minimal means? If I understood it retains the minimal set for flattened fields. How would customers make this decision to retain those fields?

If it's only for the detection engine, Maybe we can call it `detection_engine_fields` or something better that hints it's made for detection rules to still work

romulets (commented):
Looks good overall.

Question. I'm actually not sure what is the [httpjson](https://github.com/elastic/integrations/blob/main/packages/aws/data_stream/cloudtrail/agent/stream/httpjson.yml.hbs) stream. But why apply hbs changes to s3 and cloudwatch, but not to httpjson?

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
efd6: This is the consequence of the technical debt that has been built up by the unconstrained use of fields in the detection rules. Fixing this would require a significant refactor and I think is outside the scope of this change.

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
efd6: That would result in significant changes in the pipeline test outputs. I would rather not do that.

packages/aws/data_stream/cloudtrail/manifest.yml
efd6: This is defined in the UI text:

> The Minimal option retains the minimal set of fields required for the Security Detection Engine rules.

efd6:
> Question. I'm actually not sure what is the [httpjson](https://github.com/elastic/integrations/blob/main/packages/aws/data_stream/cloudtrail/agent/stream/httpjson.yml.hbs) stream. But why apply hbs changes to s3 and cloudwatch, but not to httpjson?

This was left over when the HTTP JSON input was removed in #13246. There is #14200 to track this.

ishleenk17 (approved):

zmoog (approved):

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
romulets: I agree fixing is outside. But what is the strategy to keep consistency in the list with the rules over time?

efd6:
Follow up issue: #14429

PR #14240, opened 2025-06-17
Update apache_tomcat README.md

rseldner:
this is simply a **tiny** PR to make it more clear the prometheus prerequisite applies to ingesting **metrics** instead of in general (i.e. prometheus is not needed for ingesting logs which use a `filestream` input)


ishleenk17:
@rseldner : You have to make change to the README under _docs and then build the package.
The main README will be auto generated.

PR #14251, opened 2025-06-18
fix(aws/securityhub_findings): event.kind as keyword

andrewkroh:
## Proposed commit message

Change event.kind to keyword instead of constant_keyword.
When a pipeline error occurs, event.kind was being set to
'pipeline_error' and this break ingestion because event.kind
was already assigned a value of 'state'.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/aws/data_stream/securityhub_findings/fields/agent.yml
+++ b/packages/aws/data_stream/securityhub_findings/fields/agent.yml
@@ -1,19 +1,11 @@
 - name: cloud
-  title: Cloud
-  group: 2
-  description: Fields related to the cloud or infrastructure the events are coming from.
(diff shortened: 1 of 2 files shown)


andrewkroh:
From https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/field-data-types

> Types in the same family have exactly the same search behavior but may have different space usage or performance characteristics.

Changing constant_keyword to keyword should not break anything, but the performance characteristics may differ.

efd6:
> Closes https://github.com/elastic/integrations/pull/14251

? Is there something else that it also closes?

PR #8115, opened 2023-10-06
Update elastic-package to use Package Spec 3.0.0-rc1 and fix v3 packages

jsoriano:
## Proposed commit message

Fix field definitions in packages that were updated before v3 GA release, and are failing now with the first RC (see https://github.com/elastic/integrations/pull/8109, https://github.com/elastic/integrations/pull/8120)

Changes applied:
- Fix mapping of ECS fields that are only "objects", by setting
  `type: group`. They don't generate any mapping, and they are detected as invalid objects without `object_type`.
  Child mappings are imported too. Some examples of these object fields are `dns.answer`, or `network.inner`.
- Objects without `object_type` that seem to fit as "tags" or "labels" use cases have been converted to `flattened`.
- Objects without `object_type` that seem to fit as metrics have been converted to `type: double`.
- Quote field names in test configuration files.
- Validations skipped for references founds in dashboards.
- Validations skipped for the required Kibana version needed for saved tags.

Packages with changes in fields include new changelog, so a new package will be released.

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).

## Author's Checklist

- [ ] 

## Related issues

-

PR #14256, opened 2025-06-18
[osquery manager] Fix mapping for action_data

gogochan:
## Proposed commit message

This PR fixes OS query manager kibana search by reverting the mapping changes introduced in [the previous PR](https://github.com/elastic/integrations/pull/8115).
Image

They contain the same data.
image

related https://github.com/elastic/sdh-beats/issues/6108

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/osquery_manager/data_stream/result/fields/primary-fields.yml
```diff
@@ -5,5 +5,25 @@
...
+      ignore_above: 1024
+    - name: query
+      type: keyword
+      ignore_above: 1024
+    - name: saved_query_id
+      type: keyword
+      ignore_above: 1024
+    - name: version
+      type: keyword
+      ignore_above: 1024
```
gogochan, line 29: these are copied from https://github.com/elastic/integrations/blob/main/packages/osquery_manager/data_stream/action_responses/fields/action-responses.yml#L1-L23

jsoriano: Are we sure these are all the mappings we need?

jsoriano (commented):
LGTM, having more specific mappings when possible uses to be better :+1:

jsoriano:
You can ignore SonarQube failure.

tomsonpl (approved):
Defend Workflows LGTM 👍 
Thanks for addressing this!

packages/osquery_manager/data_stream/result/fields/primary-fields.yml
gogochan: Not 100% sure. If we are missing mapping we can add them later. But this will give consistent experience for the user.

PR #14261, opened 2025-06-19
Add support for configuring  parameter.

### LINKED ISSUE #14259: [O365_metrics] Add support for max_executions as a configurable parameter in integration settings
The teams_call_quality data stream may require multiple pages of data to be processed, which can lead to exceeding the current default max_executions limit when using CEL input. To make the integration more flexible and avoid hitting execution limits in such scenarios, we should expose max_executions as a configurable parameter at the integration level.

ritalwar:
- Enhancement

## Proposed commit message
Adds support for setting the max_executions parameter through integration config,
enabling users to override the [default](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-cel#max_executions-cel) limit when handling large datasets.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #14259

### DIFF
--- a/packages/o365_metrics/manifest.yml
+++ b/packages/o365_metrics/manifest.yml
@@ -1,6 +1,6 @@
 name: o365_metrics
 title: Microsoft Office 365 Metrics
-version: "0.15.0"
+version: "0.16.0"
 description: Collect metrics from Microsoft Office 365 with Elastic Agent.
 type: integration
 format_version: "3.0.2"
@@ -181,6 +181,14 @@ policy_templates:
             show_user: false
             multi: false
             required: false
+          - name: max_executions
+            type: integer
+            title: Maximum executions
+            description: The maximum number of pages to fetch per cycle before waiting for the next interval.
+            multi: false
+            required: false
+            show_user: false
+            default: 1000
 owner:
   github: elastic/obs-infraobs-integrations
   type: elastic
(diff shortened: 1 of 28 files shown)


lucian-ioan:
Looks good, this should fix the [cel exceed error ](https://www.elastic.co/guide/en/beats/filebeat/8.18/filebeat-input-cel.html#:~:text=When%20the%20execution%20budget%20is%20exceeded%2C%20execution%20will%20be%20restarted%20at%20the%20next%20interval%20and%20a%20warning%20will%20be%20written%20into%20the%20logs)when there is too much data.

This parameter shouldn't be needed for data streams without pagination, for example `active_users_services_user_counts`. WDYT @ritalwar?

ritalwar:
> This parameter shouldn't be needed for data streams without pagination, for example `active_users_services_user_counts`. 

Yes, but since we're making this configurable at the package level, it's still appropriate to include this parameter in all data streams. That way, if there are any future changes to the CEL implementation of those data streams, we won't risk missing it.

packages/o365_metrics/manifest.yml
```diff
@@ -181,6 +181,14 @@ policy_templates:
             show_user: false
             multi: false
             required: false
+          - name: max_executions
+            type: integer
+            title: Maximum executions
+            description: The maximum number of pages to fetch per cycle before waiting for the next interval.
```
lucian-ioan: I think for this specific case we can be more verbose in the description so the user knows the integration will throw an error if they expect more than 1000 pages of API response and don't modify this default. WDYT @ritalwar ?

ritalwar: Updated.

muthu-mps (approved):

PR #14266, opened 2025-06-19
[sql_input] Update manifest format_version to 3.*

mykola-elastic:
## Proposed commit message

See title

- Change format_version to 3.3.2
- Change the type of `metrics.*` from `object` to `keyword` (we cannot have `type: object` without `object_type: ...` in format_version 3.*)
- Change system tests and sample_event.json to always have/return strings (`varchar`) in `metrics.*` (to match `keyword` type)

The type of `metrics.*` doesn't matter much because it is the responsibility of the user to define custom mappings. `metrics.*` only exists for system tests.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/obs-integration-team/issues/351


packages/sql_input/sample_event.json
```diff
@@ -60,31 +60,8 @@
...
-            "temp_bytes": 0,
-            "temp_files": 0,
-            "tup_deleted": 0,
-            "tup_fetched": 50,
-            "tup_inserted": 24,
-            "tup_returned": 127,
-            "tup_updated": 5,
-            "xact_commit": 4,
-            "xact_rollback": 0
+            "current_time": "2024-06-19T06:12:37.604443Z"
```
ishleenk17: the sample event does't have any fields from the query we are using now ?

ishleenk17 (approved):

PR #14280, opened 2025-06-20
[Synthetics] Release new version !!

shahzad31:
Release new version !!

### DIFF
--- a/packages/synthetics/changelog.yml
+++ b/packages/synthetics/changelog.yml
@@ -1,5 +1,5 @@
 # newer versions go on top
-- version: "1.4.1"
+- version: "1.4.2"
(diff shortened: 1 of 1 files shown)


packages/synthetics/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.4.1"
```
benakansara: should it be 1.4.2?

shahzad31: yes

shahzad31:
```suggestion
- version: "1.4.2"
```

PR #14288, opened 2025-06-23
[O365_Metrics] Add system tests

Linu-Elias:
## Proposed commit message
Add System tests for the following data stream
- mailbox_usage_quota_status
- outlook_activity

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/o365_metrics/_dev/deploy/docker/files/config.yml
+++ b/packages/o365_metrics/_dev/deploy/docker/files/config.yml
@@ -1679,3 +1679,41 @@ rules:
           "code": "AuthenticationFailed",
           "message": "Authentication failed."
           }
+  - path: /reports/getMailboxUsageQuotaStatusMailboxCounts(period='D7')
+    methods: ['GET']
+    request_headers:
+      Authorization:
+        - "Bearer xxxx"
+    responses:
+      - status_code: 200
+        headers:
+          Content-Type:
+              - "application/octet-stream"
+        body: '{{file "/files/mailboxUsageQuotaStatusMailboxCounts.csv"}}'           
+  - path: /reports/getEmailActivityCounts(period='D7')
+    methods: ['GET']
+    request_headers:
+      Authorization:
+        - "Bearer xxxx"
+    responses:
+      - status_code: 200
+        headers:
+          Content-Type:
+              - "application/octet-stream"
+        body: '{{file "/files/emailActivityCounts.csv"}}'
+      - status_code: 401
+        headers:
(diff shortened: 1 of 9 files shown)


stefans-elastic:
The PR description says the PR adds system tests but in changes I can see also addition of `input.type`. Could provide more info on `input.type` field (why it's added/what is its purpose)?

lucian-ioan (changes requested):
Please add a changelog entry and bump the manifest version.

Linu-Elias:
> The PR description says the PR adds system tests but in changes I can see also addition of `input.type`. Could provide more info on `input.type` field (why it's added/what is its purpose)?

@stefans-elastic The `input.type` field was added to ensure consistency between data we receive and showcase. While writing the system test, I noticed that this field was missing. Since it's a base field in the o365_metric integration, its value should be `cel`. I have now added this field to `base-fields.yml` to ensure no fields are missed and also the system test does not pass without it.

packages/o365_metrics/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.18.0"
```
stefans-elastic: is it worth bumping "minor" version here given that there was no functionality changes (just tests added)?

packages/o365_metrics/_dev/deploy/docker/files/config.yml
```diff
@@ -1678,4 +1678,217 @@ rules:
...
+            "error": {
+            "code": "InvalidAuthenticationToken",
+            "message": "Invalid Login Token",
+            "innerError": {
+                "date": "2025-06-20T06:20:05",
+                "request-id": "0f42494d-07ce-4f1e-ab26-e8483a7e8907",
+                "client-request-id": "0f42494d-07ce-4f1e-ab26-e8483a7e8907"
+              }
+            }
+          }
```
lucian-ioan, line 1866: Here we discussed to put another types of errors besides possibly user-related, such as token.

A generic 404 error would work, but at the same time, the functionality of the tests will be identical, up to you.

lucian-ioan (approved):
LGTM

packages/o365_metrics/changelog.yml
Linu-Elias: I think we only do PATCH version updates for bug fixes.

stefans-elastic (approved):
LGTM

muthu-mps (approved):

PR #14289, opened 2025-06-23
o365: fix handling of file size values and numeric IDs

efd6:
## Proposed commit message
```
o365: fix handling of file size values

O365 audit file size data may include decimal points and sizes may be
represented with e-notation. These break the convert processor. To work
around this, in the case of strings, non-digit characters were removed
to allow parsing, but this was incorrect since it would add, e.g.
significant zeros. Fix this all by using painless to conditionally
convert/parse the values of the size field and render them in long.

It appears that Microsoft will sometimes send IDs as numeric values so
be more careful when assessing for email-like IDs. Incidentally, also
take more care when checking for port delimiters in IP values.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
@@ -81,17 +81,25 @@ processors:
       copy_from: o365audit.FileSize
       ignore_empty_value: true
       override: false
-  - gsub:
-      field: file.size
-      pattern: '\D+'
-      replacement: ''
-      if: ctx.file?.size instanceof String
-      ignore_missing: true
-  - convert:
+  - script:
       tag: convert_file_size_to_long
-      field: file.size
-      type: long
-      ignore_missing: true
+      if: ctx.file?.size != null
+      source: |-
+        if (ctx.file.size instanceof long || ctx.file.size instanceof int) {
+          return;
+        }
+        if (ctx.file.size instanceof double) {
+          ctx.file.size = (long) ctx.file.size;
+          return;
+        }
+        try {
+          // Attempt to parse as long to prevent inexact integer conversion.
+          ctx.file.size = Long.parseLong(ctx.file.size);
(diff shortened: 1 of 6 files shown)


mmahacek:
I ran this update on a test instance, and it looks like it fails if the `file.size` value is an integer.
`"Processor script with tag convert_file_size_to_long in pipeline logs-o365.audit-2.18.2 failed with message: Cannot cast java.lang.Integer to java.lang.String"`

My test cases.  The first is currently failing, and the rest are passing.  Prior to this change, the first two would pass and the last three failed.
```json
POST /_ingest/pipeline/logs-o365.audit-TEST/_simulate
{"docs":[
  {"_source": {"o365audit": {"FileSizeBytes": 1249604}}},
  {"_source": {"o365audit": {"FileSizeBytes": "1249604"}}},
  {"_source": {"o365audit": {"FileSizeBytes": 1249604.0}}},
  {"_source": {"o365audit": {"FileSizeBytes": "1249604.0"}}},
  {"_source": {"o365audit": {"FileSizeBytes": "1.249604E6"}}}
]}
```

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -81,17 +81,25 @@ processors:
...
       tag: convert_file_size_to_long
-      field: file.size
-      type: long
-      ignore_missing: true
+      if: ctx.file?.size != null
+      source: |-
+        if (ctx.file.size instanceof long) {
+          return;
+        }
+        if (ctx.file.size instanceof double) {
```
mmahacek, line 91:
```suggestion
        if ((ctx.file.size instanceof double) || (ctx.file.size instanceof int)) {
```

efd6: Yeah

chrisberkhout (approved):

PR #14291, opened 2025-06-23
Fix table and edits on MySQL Integration 

alaudazzi:
This PR updates the https://www.elastic.co/docs/reference/integrations/mysql page to:

- Fix the compatibility table 
- Improve the text

Closes https://github.com/elastic/docs-content/issues/1824

### DIFF
--- a/packages/mysql/changelog.yml
+++ b/packages/mysql/changelog.yml
@@ -2,7 +2,7 @@
 - version: "1.26.2"
   changes:
     - description: Fix table and edits on MySQL Integration documentation.
-      type: enhancement
+      type: bugfix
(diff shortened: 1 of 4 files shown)


packages/mysql/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.26.2"
+  changes:
+    - description: Fix table and edits on MySQL Integration documentation.
+      type: enhancement
```
muthu-mps: Should we consider it as a bugfix? The description has a mention of Fixing the table content.

alaudazzi: It's only formatting, not content. But yes, we can change into a bug fix.

PR #14298, opened 2025-06-23
imperva_cloud_waf: handle case where cursor state cannot be found in …

efd6:
## Proposed commit message
```
imperva_cloud_waf: handle case where cursor state cannot be found in log file index

The previous code assumes that it is possible to determine the remainder
of work to be done by splitting on the last log file that was collected.
This fails when the last log file collected is not in the log files that
the index returns for whatever reason.

In that case, use the complete index to start again on the basis that
the cursor log file state has gone stale relative to the index returned
by the API.
```
> [!NOTE]
> Due to testing infrastructure limitations, the change here is not testable in CI.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/imperva_cloud_waf/changelog.yml
+++ b/packages/imperva_cloud_waf/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.12.2"
+  changes:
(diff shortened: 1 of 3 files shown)


ShourieG (approved):
LGTM

Issue #13031, opened 2025-03-10
[O11y][MySQL] Update documentation to add performance data stream details


harnish-crest-data:
As per latest MySQL integration, there is no information about performance data stream in the documentation.

![Image](https://github.com/user-attachments/assets/c0211208-781f-4943-add1-a5237cc4f8af)

 Therefore, need to add all the information about performance data stream.

PR #14303, opened 2025-06-24
Add performance data stream to MySQL Integration

alaudazzi:
This PR adds information about performance data stream.

Closes https://github.com/elastic/integrations/issues/13031


packages/mysql/_dev/build/docs/README.md
```diff
@@ -20,11 +20,12 @@ Metrics offer statistics that reflect the performance and health of MySQL. The `
```
shmsr, line 20: In this paragraph also, can we add just one liner for this `performance` data stream as other datastreams are mentioned?

Issue #14129, opened 2025-06-03
[Elastic Security]New Elastic Security Integration


narph:
# Integration release checklist

This checklist is intended for integrations maintainers to ensure consistency
when creating or updating a Package, Module or Dataset for an Integration.

### All changes

- [ ] Change follows the [contributing guidelines](https://github.com/elastic/integrations/blob/main/CONTRIBUTING.md)
- [ ] Supported versions of the monitoring target are documented
- [ ] Supported operating systems are documented (if applicable)
- [ ] Integration or [System tests](https://github.com/elastic/elastic-package/blob/master/docs/howto/system_testing.md) exist
- [ ] Documentation exists, [useful guidelines](https://github.com/elastic/integrations/blob/main/docs/documentation_guidelines.md) to follow
- [ ] Fields follow [ECS](https://github.com/elastic/ecs) and [naming conventions](https://www.elastic.co/guide/en/beats/devguide/master/event-conventions.html)
- [ ] At least a manual test with ES / Kibana / Agent has been performed.
- [ ] Required Kibana version set to:

mohitjha-elastic:
Here are the dashboard ideas:

**Control Panel (Filters):**
Index name
Source
Severity

**Visualizations:**

Alerts by Severity -> Pie Chart
Alerts by Host OS -> Pie Chart

Alerts by Rule name -> Table

Alerts by Risk Score -> Bar Chart
Alerts by Process Name -> Bar Chart
Alerts by Process User Name -> Bar Chart
Alerts by Host Name -> Bar Chart
Alerts by Event action -> Bar Chart
Alerts by User ID -> Bar Chart

kcreddy:
> **Control Panel (Filters):**
Index name
Source
Severity

It might be also worth include following Controls:
- Alert Name: Control as `Options list` for filtering on `kibana.alert.rule.name`
- User: Control as `Options list` for filtering on `user.name` 
- Host: Control as `Options list` for filtering on `host.name` 
- Risk Score: Control as `Range Slider` for filtering on `kibana.alert.rule.risk_score`
- Assignee: Control as `Options list` for filtering on `kibana.alert.workflow_assignee_ids` 
- Host Risk Level: Control as `Options list` for filtering on `host.risk.calculated_level`
- User Risk Level: Control as `Options list` for filtering on `user.risk.calculated_level`

PR #13494, opened 2025-04-10
[ti_recordedfuture] Add Support for Playbook and Triggered Alerts

mohitjha-elastic:
## Proposed Commit Message

```
ti_recordedfuture: add support for playbook and triggered alerts as two new data streams

This adds support for the Playbook and Triggered Alerts event type as two new data streams to
enhance the overall visibility of threat data in the Recorded Future integration.

This changes also includes updating ecs version to 8.17.0 in all existing data streams.

Sanitized test case inputs were obtained from the Recorded Future API for Playbook Alerts[1]
and from the Recorded Future Connect API for Triggered Alerts[2].

[1] https://api.recordedfuture.com/playbook-alert
[2] https://api.recordedfuture.com/v2/#!/Alerts/Alert_Notification_Search
```

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).

## How to test this PR locally
Clone integrations repo.
Install the elastic package locally.
Start the elastic stack using the elastic package.
Move to integrations/packages/ti_recordedfuture directory.
Run the following command to run tests.
`elastic-package test -v`

## Related issues

- Closes https://github.com/elastic/integrations/issues/13146

## Screenshots
playbook-alert-dashboard

trigg-alert-dashboard

PR #14305, opened 2025-06-24
[elastic_security] Initial release of Elastic Security

mohitjha-elastic:
## Proposed commit message
```
elastic_security: Initial release of the package.

Initial release with alert data stream, associated dashboards, and ingest pipelines. 
This package facilitates transferring security alert data from another Elasticsearch instance to your own. 
API integration was implemented as per the official documentation[1], and test samples 
were created using sanitized live data

[1] https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-search
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/elastic_security directory.

> elastic-package test -v

## Screenshots
image

-Logs-Elastic-Security-Alert-Elastic-06-30-2025_05_20_PM

event-id-1

event-id-2
config-params

## Related Issue

- Closes https://github.com/elastic/integrations/issues/14129


packages/elastic_security/_dev/build/docs/README.md
```diff
@@ -0,0 +1,52 @@
+# Elastic Security
+
+## Overview
+
+Elastic Security Alerts are triggered when detection rules identify suspicious or malicious activity. They provide detailed context like rule name, impacted entities, timestamps, and other necessary details. Alerts can be investigated in Kibana using tools like Timeline. They support custom actions such as notifications or automated responses. These alerts help prioritize and manage security threats efficiently.
+
+## Data streams
+
+This integration collects the following logs:
```
kcreddy:
```suggestion
The Elastic Security integration collects the following events:
```

packages/elastic_security/_dev/build/docs/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Agent-based deployment
+
+Elastic Agent must be installed. For more details, check the Elastic Agent [installation instructions](docs-content://reference/fleet/install-elastic-agents.md). You can install only one Elastic Agent per host.
+
+## Setup
+
+### To collect data from the Elastic API:
+
+To collect data from the Elastic API, you will need the following information:
```
kcreddy:
```suggestion
You will need the following information:
```
Redundant as per heading

packages/elastic_security/_dev/build/docs/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+1. The URL for the Elasticsearch instance.
+2. Authentication credentials such as username, password, API key, or bearer token depend on the selected authentication type.
```
kcreddy:
```suggestion
2. Authentication credentials such as username, password, API key, or bearer token depending on the selected authentication type.
```

packages/elastic_security/_dev/build/docs/README.md
```diff
@@ -0,0 +1,52 @@
+# Elastic Security
+
+## Overview
+
+Elastic Security Alerts are triggered when detection rules identify suspicious or malicious activity. They provide detailed context like rule name, impacted entities, timestamps, and other necessary details. Alerts can be investigated in Kibana using tools like Timeline. They support custom actions such as notifications or automated responses. These alerts help prioritize and manage security threats efficiently.
```
kcreddy: This section should be related to Elastic Security and not just the alerts.

packages/elastic_security/_dev/build/docs/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Data streams
+
+This integration collects the following logs:
+
+`alert`: - Retrieve alerts from Elastic Instance.
```
kcreddy:
```suggestion
`alert`: - Retrieve alerts from Elasticsearch Instance using Elasticsearch [_search](https://www.elastic.co/docs/api/doc/elasticsearch/group/endpoint-search) API.
```

packages/elastic_security/_dev/build/docs/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+1. The URL for the Elasticsearch instance.
+2. Authentication credentials such as username, password, API key, or bearer token depend on the selected authentication type.
+
+### Enable the integration in Elastic
+
+1. In Kibana navigate to **Management** > **Integrations**.
+2. In the search top bar, type **Elastic Security**.
+3. Select the **Elastic Security** integration and add it.
+4. Add all the required integration configuration parameters such as username, password, API key, or bearer token depend on the selected authentication type to enable data collection.
```
kcreddy:
```suggestion
4. Add all the required integration configuration parameters such as username, password, API key, or bearer token depending on the selected authentication type to enable data collection.
```

packages/elastic_security/_dev/build/docs/README.md
kcreddy: Can you add links how to create each of these auth types?

packages/elastic_security/data_stream/alert/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,109 @@
...
+resource.timeout: {{http_client_timeout}}
+{{/if}}
+resource.url: {{url}}
+state:
+  auth_type: {{auth_type}}
+  api_key: {{api_key}}
+  user: {{username}}
+  password: {{password}}
+  bearer_token: {{bearer_token}}
+  index: .internal.alerts-security.alerts*
```
kcreddy: Can you check if you can query the alrt index aliases instead as recommended?
https://www.elastic.co/docs/solutions/security/detect-and-alert/query-alert-indices#_alert_index_aliases

packages/elastic_security/_dev/build/docs/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Setup
+
+### To collect data from the Elastic API:
+
+To collect data from the Elastic API, you will need the following information:
+
+1. The URL for the Elasticsearch instance.
+2. Authentication credentials such as username, password, API key, or bearer token depend on the selected authentication type.
+
```
kcreddy, line 38: Also mention that users need `read` Index privileges on `.internal.alerts-security.alerts*` indices.

packages/elastic_security/data_stream/alert/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,109 @@
...
+program: |
+  state.with({
+    "start_time": state.?cursor.last_timestamp.orValue((now - duration(state.initial_interval)).format(time_layout.RFC3339)),
+  }).as(state, state.with(
+      request(
+        "GET",
+        state.url.trim_right("/") + "/" + state.index + "/_search?" + {
+          "sort": ["@timestamp:asc"],
+          "size": [string(state.batch_size)],
+          "q": ["@timestamp:{" + state.start_time + " TO now}"],
```
kcreddy: Can you use the request body `query` instead where you can use Elasticsearch Query DSL?
In case in future we need to support complex queries (or filters), we can do that using Query DSL.

packages/elastic_security/_dev/deploy/docker/files/config.yml
```diff
@@ -0,0 +1,2287 @@
...
+    responses:
+      - status_code: 200
+        body: |-
+          {{ minify_json `
+          {
+            "hits": {
+              "hits": [
+              {
+                "_index": "abcd-1234",
+                "_id": "fghiabcd",
```
kcreddy: Can you change the `_id` here as it should ideally be different?

packages/elastic_security/data_stream/alert/sample_event.json
```diff
@@ -0,0 +1,1122 @@
+{
+    "@timestamp": "2060-06-09T13:56:03.205Z",
```
kcreddy: Can you use past dates instead of future ones?

packages/elastic_security/data_stream/alert/_dev/test/pipeline/test-alert.log
```diff
@@ -0,0 +1 @@
...a.alert.rule.updated_by":"elastic","kibana.alert.rule.version":103,"kibana.alert.url":"https:\/\/release-app\/app\/security\/alerts\/redirect\/abcd?index=.alerts-security.alerts-default\u00d7tamp=2024-06-09T13:56:03.205Z","kibana.alert.uuid":"abcdef123","kibana.alert.workflow_tags":[],"kibana.alert.workflow_assignee_ids":[],"kibana.alert.rule.risk_score":47,"kibana.alert.rule.severity":"medium","kibana.alert.original_event.severity":73,"kibana.alert.original_event.code":"behavior","kibana.alert.original_event.risk_score":73,"kibana.alert.original_event.created":"2024-06-09T13:52:07.402627721Z","kibana.alert.original_event.kind":"alert","kibana.alert.original_event.module":"endpoint","kibana.alert.original_event.type":["info","allowed"],"kibana.alert.original_event.agent_id_status":"verified","kibana.alert.original_event.sequence":600720,"kibana.alert.original_event.ingested":"2024-06-09T13:52:08Z","kibana.alert.original_event.action":"rule_detection","kibana.alert.original_event.id":"Na7UF0\/abcddd++++3Y96","kibana.alert.original_event.category":["malware","intrusion_detection"],"kibana.alert.original_event.dataset":"endpoint.alerts","kibana.alert.original_event.outcome":"success"}
```
kcreddy: Can you use past dates instead of year `2060` or `2070`

packages/elastic_security/data_stream/alert/_dev/test/pipeline/test-alert.log-expected.json
```diff
@@ -0,0 +1,1103 @@
...
+                "category": [
+                    "malware",
+                    "intrusion_detection"
+                ],
+                "code": "behavior",
+                "created": "2024-06-09T13:52:07.402627721Z",
+                "id": "Na7UF0/g6Q++++3Y96",
+                "ingested": "2024-06-09T13:52:08Z",
+                "kind": "alert",
+                "original": "Malicious Behavior Prevention Alert: Empire Stager Execution",
```
kcreddy: This doesn't seem to be correct.

packages/elastic_security/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,117 @@
...
+      field: message
+      tag: remove_message
+      ignore_missing: true
+      description: The `message` field is no longer required if the document has an `event.original` field.
+      if: ctx.event?.original != null
+  - dot_expander:
+      field: "*"
+      ignore_failure: true
+  - uppercase:
+      field: host.mac
```
kcreddy, line 43: Add tag

packages/elastic_security/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,117 @@
...
+      if: ctx.event?.original == null
+  - remove:
+      field: message
+      tag: remove_message
+      ignore_missing: true
+      description: The `message` field is no longer required if the document has an `event.original` field.
+      if: ctx.event?.original != null
+  - dot_expander:
+      field: "*"
+      ignore_failure: true
```
kcreddy, line 41: Add tag

packages/elastic_security/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,117 @@
...
+      ignore_missing: true
+  - append:
+      field: tags
+      value: preserve_original_event
+      allow_duplicates: false
+      if: ctx.error?.message != null
+  - set:
+      field: event.kind
+      tag: set_event_kind
+      value: alert
```
kcreddy, line 59: We will be losing information about source `event.kind`. Can you store that first inside metadata?

packages/elastic_security/data_stream/alert/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,109 @@
...
+      }).do_request().as(resp, resp.StatusCode == 200 ?
+        resp.Body.decode_json().as(body,{
+          "events": (
+            has(body.?hits.hits) && body.hits.hits.size() > 0 ?
+              body.hits.hits.map(e,{
+                "source_metadata": {
+                   "_index": e.?_index,
+                   "_id": e.?_id,
+                },
+                "message": e._source.drop(['data_stream', 'event.dataset', 'event.module']).encode_json(),
```
kcreddy, line 74: Before deleting this source information, can you store inside `source_metadata`?

kcreddy: If you don't delete them and store everything in the root of the document itself, what is the error you get when  when building or running tests?

packages/elastic_security/data_stream/alert/manifest.yml
```diff
@@ -0,0 +1,161 @@
...
+          - text: Basic Auth
+            value: basic_auth
+          - text: Bearer Auth
+            value: bearer_auth
+        default: api_auth
+        description: Provide the type of authentication to be used for the request.
+      - name: url
+        type: url
+        title: URL
+        description: URL of Elasticsearch instance.
```
kcreddy: Please give an example for the URL.

packages/elastic_security/data_stream/alert/manifest.yml
```diff
@@ -0,0 +1,161 @@
...
+        type: select
+        options:
+          - text: API Auth
+            value: api_auth
+          - text: Basic Auth
+            value: basic_auth
+          - text: Bearer Auth
+            value: bearer_auth
+        default: api_auth
+        description: Provide the type of authentication to be used for the request.
```
kcreddy: Add link to https://www.elastic.co/docs/api/doc/elasticsearch#doc-authentication

kcreddy:
```suggestion
        description: Type of authentication to be used for the Elasticsearch API requests.
```

packages/elastic_security/data_stream/alert/manifest.yml
```diff
@@ -0,0 +1,161 @@
...
+        show_user: true
+        secret: true
+      - name: initial_interval
+        type: text
+        title: Initial Interval
+        multi: false
+        required: true
+        show_user: true
+        default: 24h
+        description: How far back to pull the logs from Elasticsearch API. Supported units for this parameter are h/m/s.
```
kcreddy:
```suggestion
        description: How far back to pull the events from Elasticsearch API. Supported units for this parameter are h/m/s. Example `72h`.
```

packages/elastic_security/data_stream/alert/fields/fields.yml
```diff
@@ -0,0 +1,1097 @@
+- name: elastic
+  type: group
+  fields:
+    - name: agent
+      type: group
+      fields:
+        - name: id
+          type: keyword
+- name: kibana
```
kcreddy, line 10: Is there any reference where this schema was taken? Can you add a comment at the top?

packages/elastic_security/data_stream/alert/fields/fields.yml
```diff
@@ -0,0 +1,1097 @@
...
+      type: keyword
+    - name: uuid
+      type: keyword
+    - name: workflow_status
+      type: keyword
+  - name: space_ids
+    type: keyword
+  - name: version
+    type: keyword
+- name: Endpoint
```
kcreddy, line 266: Is there any reference where this schema was taken? Can you add a comment at the top?
Same for all top level fields.

packages/elastic_security/data_stream/alert/fields/fields.yml
```diff
@@ -0,0 +1,1097 @@
...
+              type: keyword
+- name: process
+  type: group
+  fields:
+  - name: Ext
+    type: group
+    fields:
+    - name: ancestry
+      type: keyword
+  - name: args
```
kcreddy: Aren't some of these fields already part of ECS? Can you identify them and remove?
Same for other top level fields that are already in ECS.

packages/elastic_security/img/elastic-security-alert-dashboard.png
kcreddy, line 1: Add a search at the bottom of the dashboard.

packages/elastic_security/img/elastic-security-logo.svg
kcreddy, line 1: We could use the Elastic Security logo instead.

packages/elastic_security/manifest.yml
```diff
@@ -0,0 +1,44 @@
...
+        enabled: true
+      agentless:
+        enabled: true
+        organization: security
+        division: engineering
+        team: security-service-integrations
+    inputs:
+      - type: cel
+        title: Collect Elastic Security Alerts via API
+        description: Collect Alerts from Elastic instance via API.
```
kcreddy:
```suggestion
        title: Collect Elastic Security events via API
        description: Collect events from Elastic instance via API.
```

packages/elastic_security/data_stream/alert/sample_event.json
```diff
@@ -0,0 +1,1122 @@
...
+    "event": {
+        "action": "rule_detection",
+        "agent_id_status": "mismatch",
+        "category": [
+            "malware",
+            "intrusion_detection"
+        ],
+        "code": "behavior",
+        "created": "2024-06-09T13:52:07.402627721Z",
+        "dataset": "elastic_security.alert",
```
kcreddy, line 124: I don't see `event.module`. Can you please check why its missing?

packages/elastic_security/data_stream/alert/sample_event.json
mohitjha-elastic: Using a past date might not be appropriate in this case, as the entry point depends on the `initial_interval`. The data is fetched starting from that interval, and pagination happens based on the `@timestamp` of the incoming data. Therefore, the starting time should be in the future rather than the past.

A similar approach was used [here ](https://github.com/elastic/integrations/pull/13494/files/666267bdafa4e044ac9be936a1772488895d0a6d#r2072810347)for reference.

packages/elastic_security/data_stream/alert/agent/stream/cel.yml.hbs
mohitjha-elastic: `elastic-package` expects `data_stream.dataset` should be set as `<integration-name>.<data_stream_name>` else it throws the below error. For the same reason, `event.dataset` and `event.module` have also been dropped to avoid ambiguity.

image

Also the removed fields are already present inside the `kibana.alert.original_event` hence did not include in source_metadata.

packages/elastic_security/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
mohitjha-elastic: These values are already available under `kibana.alert.original_event`, so they weren’t included in `source_metadata`. Do you still think it's necessary to include them?

packages/elastic_security/data_stream/alert/fields/fields.yml
mohitjha-elastic: This has been taken from the live samples. Sure, will add comment.

packages/elastic_security/data_stream/alert/fields/fields.yml
mohitjha-elastic: https://github.com/elastic/integrations/pull/14305/files#r2171668104

packages/elastic_security/data_stream/alert/fields/fields.yml
mohitjha-elastic: Some fields are being received as string values instead of the expected array format, so we need to explicitly define them in `fields.yml`. 
This behavior has been observed consistently across events.

packages/elastic_security/data_stream/alert/sample_event.json
mohitjha-elastic: I checked and found that it’s not included in any of the packages where `event.module` is defined as a constant keyword in `base-fields.yml`. I'm not sure about the reason behind this, but it is available in Kibana.
image

packages/elastic_security/_dev/build/docs/README.md
```diff
@@ -0,0 +1,57 @@
...
+## Overview
+
+Elastic Security is a free and open solution that helps detect, investigate, and respond to threats using data from endpoints, cloud, and network sources. It offers SIEM and endpoint protection with powerful search, correlation, and visualization features in Kibana.
+It enables security teams to streamline investigations and strengthen their overall security posture.
+
+## Data streams
+
+The Elastic Security integration collects the following events:
+
+`alert`: - Retrieve alerts from Elasticsearch Instance using Elasticsearch [_search](https://www.elastic.co/docs/api/doc/elasticsearch/group/endpoint-search) API.
```
kcreddy: Can you link the specific API: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-search-2?

packages/elastic_security/_dev/build/docs/README.md
```diff
@@ -0,0 +1,57 @@
+# Elastic Security
+
+## Overview
+
+Elastic Security is a free and open solution that helps detect, investigate, and respond to threats using data from endpoints, cloud, and network sources. It offers SIEM and endpoint protection with powerful search, correlation, and visualization features in Kibana.
```
kcreddy: Add link to https://www.elastic.co/security

packages/elastic_security/_dev/build/docs/README.md
```diff
@@ -0,0 +1,57 @@
...
+### To collect data from the Elastic API:
+
+You will need the following information:
+
+1. The URL for the Elasticsearch instance.
+2. Authentication credentials such as username, password, API key, or bearer token depending on the selected authentication type.
+
+Note:
+1. Users must have read index privileges on the `.alerts-security.alerts*` indices to access and query security alerts.
+2. To learn how to create authentication credentials and use the appropriate authentication type, refer to the Elasticsearch Authentication [Documentation](https://www.elastic.co/docs/api/doc/elasticsearch/authentication).
```
kcreddy: The link in this line is not referring to how to create the authentication credentials, but how to use them inside the API.
Ideally you may want to add links to how to create auth credentials.

packages/elastic_security/_dev/build/docs/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+### To collect data from the Elastic API:
+
+You will need the following information:
+
+1. The URL for the Elasticsearch instance.
+2. Authentication credentials such as username, password, API key, or bearer token depending on the selected authentication type.
+
+Note:
+1. Users must have read index privileges on the `.alerts-security.alerts*` indices to access and query security alerts.
```
kcreddy:
```suggestion
1. Users must have `read` index privileges on the `.alerts-security.alerts*` indices to access and query security alerts.
```

packages/elastic_security/data_stream/alert/sample_event.json
kcreddy: > The data is fetched starting from that interval, and pagination happens based on the @timestamp of the incoming data. Therefore, the starting time should be in the future rather than the past.

If you set `@timestamp` inside the `config.yml` to reasonable past dates `2022`, `2023`, etc. you will still be able to achieve same result because your `cursor.last_timestamp` is based on those values.
https://github.com/elastic/integrations/blob/602ae72973cab718d12fc01796c2a2e7a60fd9b8/packages/elastic_security/data_stream/alert/agent/stream/cel.yml.hbs#L81.

packages/elastic_security/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
kcreddy: Okay, it may not be necessary.

packages/elastic_security/data_stream/alert/agent/stream/cel.yml.hbs
kcreddy: > Also the removed fields are already present inside the kibana.alert.original_event hence did not include in source_metadata.

Thanks for confirming.

packages/elastic_security/_dev/deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: '3.0'
+services:
+  elastic_security-cel:
+    image: docker.elastic.co/observability/stream:v0.17.1
```
efd6:
```suggestion
    image: docker.elastic.co/observability/stream:v0.18.0
```

packages/elastic_security/data_stream/alert/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,118 @@
...
+          ],
+        }
+      }).do_request().as(resp, resp.StatusCode == 200 ?
+        resp.Body.decode_json().as(body,{
+          "events": (
+            has(body.?hits.hits) && body.hits.hits.size() > 0 ?
+              body.hits.hits.map(e,{
+                "source_metadata": {
+                   "_index": e.?_index,
+                   "_id": e.?_id,
```
efd6: This is not safe. If either of `e._index` or `e._id` do not exist, when the value is finally serialised, this will fail. If they may be missing, the fields need to be marked as optional too.
```suggestion
                   ?"_index": e.?_index,
                   ?"_id": e.?_id,
```

packages/elastic_security/data_stream/alert/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,118 @@
...
+redact:
+  fields:
+    - user
+    - password
+    - api_key
+    - bearer_token
+program: |
+  state.with({
+    "start_time": state.?cursor.last_timestamp.orValue((now - duration(state.initial_interval)).format(time_layout.RFC3339)),
+  }).as(state, state.with(
```
efd6:
```suggestion
  state.?cursor.last_timestamp.orValue(
    (now - duration(state.initial_interval)).format(time_layout.RFC3339)
  ).as(start_time, state.with(
```

packages/elastic_security/data_stream/alert/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,118 @@
...
+        "GET",
+        state.url.trim_right("/") + "/" + state.index + "/_search?" + {
+          "sort": ["@timestamp:asc"],
+          "size": [string(state.batch_size)],
+        }.format_query(),
+        {
+          "query": {
+            "range": {
+              "@timestamp": {
+                "gt": state.start_time,
```
efd6:
```suggestion
                "gt": start_time,
```

packages/elastic_security/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,119 @@
...
+      tag: script_remove_message
+      if: ctx.event?.original != null && ctx.message != null
+      source: |-
+        if(ctx.event.original == ctx.message) {
+          ctx.remove('message');
+        }
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```
efd6: Why is this a script instead of a remove processor?

packages/elastic_security/data_stream/alert/fields/fields.yml
```diff
@@ -0,0 +1,1097 @@
...
+          type: keyword
+        - name: subtechnique
+          type: group
+          fields:
+            - name: id
+              type: keyword
+            - name: name
+              type: keyword
+            - name: reference
+              type: keyword
```
efd6: Add final new line.

packages/elastic_security/data_stream/alert/sample_event.json
efd6, line 1: For the sake of documentation, maybe the event that leads to this sample could be cut down a little.

efd6 (approved):
LGTM but please wait for @kcreddy.

mohitjha-elastic:
> LGTM for the review comments.
> 
> @mohitjha-elastic, can you also confirm the issues discussed via DMs, namely mismatched `event.severity`, incorrect `@timestamp`, and duplicate `event.id` are all fixed now?

@kcreddy  Sorry I left that issue unconcluded.

I investigated and confirmed that the logs at both the source and destination Elasticsearch instances are identical — same `timestamp`, `severity`, etc.
It turns out there are two entries for the same event ID in the source instance: one in the original index (e.g., `.ds-logs-endpoint.events.process*`) and another in the` .internal.alerts-security.alerts*` index. Since we're ingesting data only from the alert index, the ingested data at the destination matches what's present in that alert index.

Source Instance (From which we are ingesting the data) -
client-instance

Destination Instance (In which we are ingesting the data) -
our-instance

kcreddy (approved):
Thanks!

packages/elastic_security/data_stream/alert/fields/fields.yml
andrewkroh: > Some fields are being received as string values instead of the expected array format

I don't really understand this justification. What do you mean by "expected array" format? Is that specifically for `process.args`? Many of these fields can simply change to using `external: ecs` without affecting the field type.

<details>
<summary>Fields that should use external: ecs</summary>

|field|ECS type|
|--|--|
|data_stream.type|constant_keyword|
|data_stream.dataset|constant_keyword|
|data_stream.namespace|constant_keyword|
|@timestamp|date|
|process.args_count|long|
|process.entity_id|keyword|
|process.entry_leader.args|keyword|
|process.entry_leader.args_count|long|
|process.entry_leader.entity_id|keyword|
|process.entry_leader.entry_meta.type|keyword|
|process.entry_leader.executable|keyword|
|process.entry_leader.group.name|keyword|
|process.entry_leader.interactive|boolean|
|process.entry_leader.name|keyword|
|process.entry_leader.parent.entity_id|keyword|
|process.entry_leader.parent.pid|long|
|process.entry_leader.pid|long|
|process.entry_leader.real_group.name|keyword|
|process.entry_leader.real_user.name|keyword|
|process.entry_leader.same_as_process|boolean|
|process.entry_leader.user.name|keyword|
|process.entry_leader.working_directory|keyword|
|process.executable|keyword|
|process.group.name|keyword|
|process.group_leader.args|keyword|
|process.group_leader.args_count|long|
|process.group_leader.entity_id|keyword|
|process.group_leader.executable|keyword|
|process.group_leader.group.name|keyword|
|process.group_leader.interactive|boolean|
|process.group_leader.name|keyword|
|process.group_leader.pid|long|
|process.group_leader.real_group.name|keyword|
|process.group_leader.real_user.name|keyword|
|process.group_leader.same_as_process|boolean|
|process.group_leader.supplemental_groups.name|keyword|
|process.group_leader.user.name|keyword|
|process.group_leader.working_directory|keyword|
|process.hash.md5|keyword|
|process.hash.sha1|keyword|
|process.hash.sha256|keyword|
|process.interactive|boolean|
|process.name|keyword|
|process.parent.args|keyword|
|process.parent.args_count|long|
|process.parent.entity_id|keyword|
|process.parent.executable|keyword|
|process.parent.group.name|keyword|
|process.parent.interactive|boolean|
|process.parent.name|keyword|
|process.parent.pid|long|
|process.parent.real_group.name|keyword|
|process.parent.real_user.name|keyword|
|process.parent.supplemental_groups.name|keyword|
|process.parent.user.name|keyword|
|process.parent.working_directory|keyword|
|process.pid|long|
|process.previous.args|keyword|
|process.previous.args_count|long|
|process.previous.executable|keyword|
|process.real_group.name|keyword|
|process.real_user.name|keyword|
|process.session_leader.args|keyword|
|process.session_leader.args_count|long|
|process.session_leader.entity_id|keyword|
|process.session_leader.executable|keyword|
|process.session_leader.group.name|keyword|
|process.session_leader.interactive|boolean|
|process.session_leader.name|keyword|
|process.session_leader.pid|long|
|process.session_leader.real_group.name|keyword|
|process.session_leader.real_user.name|keyword|
|process.session_leader.same_as_process|boolean|
|process.session_leader.supplemental_groups.name|keyword|
|process.session_leader.user.name|keyword|
|process.session_leader.working_directory|keyword|
|process.supplemental_groups.name|keyword|
|process.user.name|keyword|
|process.working_directory|keyword|
|threat.tactic.id|keyword|
|threat.tactic.reference|keyword|
|threat.tactic.name|keyword|
|threat.technique.id|keyword|
|threat.technique.name|keyword|
|threat.technique.reference|keyword|
|threat.technique.subtechnique.id|keyword|
|threat.technique.subtechnique.name|keyword|
|threat.technique.subtechnique.reference|keyword|
</details>

This can be fixed by running

`go run github.com/andrewkroh/fydler@main -a useecs -fix packages/elastic_security/**/fields/*.yml`

However, the more concerning part are the fields that are declared in conflict with ECS. I think these need to be fixed to use `external: ecs` which will change their type. Most of these changes will be widening the data type except for the `keyword` to `date` changes. Here's a summary:

|field|type|ECS type|
|--|--|--|
|process.command_line|keyword|wildcard|
|process.entry_leader.group.id|long|keyword|
|process.entry_leader.parent.start|keyword|date|
|process.entry_leader.real_group.id|long|keyword|
|process.entry_leader.real_user.id|long|keyword|
|process.entry_leader.start|keyword|date|
|process.entry_leader.user.id|long|keyword|
|process.group.id|long|keyword|
|process.group_leader.group.id|long|keyword|
|process.group_leader.real_group.id|long|keyword|
|process.group_leader.real_user.id|long|keyword|
|process.group_leader.start|keyword|date|
|process.group_leader.supplemental_groups.id|long|keyword|
|process.group_leader.user.id|long|keyword|
|process.parent.command_line|keyword|wildcard|
|process.parent.group.id|long|keyword|
|process.parent.real_group.id|long|keyword|
|process.parent.real_user.id|long|keyword|
|process.parent.start|keyword|date|
|process.parent.supplemental_groups.id|long|keyword|
|process.parent.user.id|long|keyword|
|process.real_group.id|long|keyword|
|process.real_user.id|long|keyword|
|process.session_leader.group.id|long|keyword|
|process.session_leader.real_group.id|long|keyword|
|process.session_leader.real_user.id|long|keyword|
|process.session_leader.start|keyword|date|
|process.session_leader.supplemental_groups.id|long|keyword|
|process.session_leader.user.id|long|keyword|
|process.start|keyword|date|
|process.supplemental_groups.id|long|keyword|
|process.user.id|long|keyword|

mohitjha-elastic: @andrewkroh 
We're seeing two types of issues or behaviors here:

1. **Data type mismatches**: Some ECS fields are appearing with a different data type than expected. For example, `process.group.id` is coming in as an integer in the raw logs, whereas ECS expects it to be a string.

2. **Structure mismatches**: Some fields, like `threat.techniques`, are coming through as arrays of objects in the raw logs, while ECS expects them to be explicitly defined as `group` or `nested` types.

Example here -
```
{
  "threat": [
    {
      "framework": "MITRE ATT&CK",
      "technique": [
        {
          "reference": "https://attack.mitre.org/techniques/T1059/",
          "name": "Command and Scripting Interpreter",
          "subtechnique": [
            {
              "reference": "https://attack.mitre.org/techniques/T1059/004/",
              "name": "Unix Shell",
              "id": "T1059.004"
            },
            {
              "reference": "https://attack.mitre.org/techniques/T1059/006/",
              "name": "Python",
              "id": "T1059.006"
            }
          ],
          "id": "T1059"
        }
      ],
      "tactic": {
        "reference": "https://attack.mitre.org/tactics/TA0002/",
        "name": "Execution",
        "id": "TA0002"
      }
    },
    {
      "framework": "MITRE ATT&CK",
      "technique": [
        {
          "reference": "https://attack.mitre.org/techniques/T1132/",
          "name": "Data Encoding",
          "subtechnique": [
            {
              "reference": "https://attack.mitre.org/techniques/T1132/001/",
              "name": "Standard Encoding",
              "id": "T1132.001"
            }
          ],
          "id": "T1132"
        }
      ]
    }
  ]
}
```

For the first issue, even if we reference the `external:ecs` definition, we'll still encounter data type mismatch errors. To resolve this, we can add a script in the ingest pipeline to convert such fields (e.g., integers to strings) and then remove these ECS definitions from the `fields.yml` as definitions will be handled by the dynamic ECS imports.

For the second issue, the structure itself is incompatible. One possible solution is that we'll need to handle these cases in the pipeline as well—likely by flattening or reformatting the data to match the ECS schema.

Let me know your thoughts on this.

andrewkroh: > Data type mismatches: Some ECS fields are appearing with a different data type than expected. For example, process.group.id is coming in as an integer in the raw logs, whereas ECS expects it to be a string.

The Elasticsearch data type is the primary concern. The integration reads the event structure from [`_source`]( https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/mapping-source-field), which might not reflect how the data was indexed in the source cluster.

For instance, it's common for fields to be numbers in the JSON `_source` and stored as `keyword` types. However, in `elastic-package`, this causes a validation issue because we prefer `_source` to match the ES data type (e.g., JSON string for a keyword). To resolve this, you can either apply a `convert(type: string)` processor or configure `elastic-package` to ignore numeric fields using `numeric_keyword_fields`[^1].

[^1]: https://github.com/elastic/elastic-package/blob/f8f2f15a04bcc25eca00887fb147bd7f8a0f32b3/docs/howto/pipeline_testing.md#test-configuration

> then remove these ECS definitions from the fields.yml as definitions will be handled by the dynamic ECS imports.

Do not remove the static ECS definitions. Keep them. They are generally stronger (because they don't rely on `match_mapping_type`) and provide documentation.

------

> Structure mismatches: Some fields, like threat.techniques, are coming through as arrays of objects in the raw logs, while ECS expects them to be explicitly defined as group or nested types.

Again, I believe this relates to how the `_source` is structured versus how the data was actually indexed and how it will be indexed in the target cluster.

If you send the data as-is to an index that contains mappings for the ECS threat fields, it should be mapped correctly. This is because Elasticsearch automatically flattens[^2] arrays of objects. Did you encounter any errors while trying to use the ECS definitions with this JSON structure?

[^2]: https://www.elastic.co/guide/en/elasticsearch/reference/8.18/nested.html#nested-arrays-flattening-objects

mohitjha-elastic: Thanks for the suggestion. @andrewkroh 
Will ignore the numeric fields using `numeric_keyword_fields `and use `external: ecs` to resolve the data type mismatch issue.

Regarding the second issue on structure mismatch, using the ECS definitions with the provided JSON structure results in an error from elastic-package, as shown in the attached message.
ECS expects the threat.technique fields to be either group or nested but its coming as an array of objects.
image

andrewkroh: Thank you for sharing the error. That elastic-package validation seems counter productive given that it would force the pipeline to implement something that Elasticsearch does automatically. I don't see any way to disable the check. So I think a comment in the file explaining why the declaration for the threat.* fields exists, and why they are not using `external: ecs` is in order.

kcreddy: > If you send the data as-is to an index that contains mappings for the ECS threat fields, it should be mapped correctly. This is because Elasticsearch automatically flattens[2](https://github.com/elastic/integrations/pull/14305#user-content-fn-2-ff965d9188b5bc65d4ea83e54821305a) arrays of objects.

@andrewkroh, since this automatic flattening results in losing relationships/associations between fields, is it better to make them `nested` instead and keep the associations for `threat` fields?
@mohitjha-elastic, if you don't add `external: ecs` for any ECS fields, please manually add field's `description:`.

andrewkroh: > since this automatic flattening results in losing relationships/associations between fields, is it better to make them nested instead and keep the associations for threat fields?

I would defer to ECS on this. ECS does not indicate to use `nested` on the `threat`, only on the `threat.enrichments`. So it must be that the associations are not essential (if they are then we need to change ECS).

mohitjha-elastic: Thanks @andrewkroh and @kcreddy!!
Here is the [Issue ](https://github.com/elastic/integrations/issues/14637)and the [PR](https://github.com/elastic/integrations/pull/14638).

PR #14315, opened 2025-06-25
New httpcheck package based on the Otel collector httpcheck receiver

jsoriano:
Related to https://github.com/elastic/kibana/issues/224472.

This is an experiment about creating input packages that contain OTel collector configuration.

It uses what we have now, with some remarks:
- Input in the policy template is `input: otelcol`.
- Agent template still uses handlebars, but defines OTel collector configuration.

If this is enough to indicate Fleet that it should generate OTel collector configuration for Hybrid agent, we won't need changes in the package spec, but we may need to require a new spec version.


packages/httpcheck/agent/input/input.yml.hbs
```diff
@@ -0,0 +1,20 @@
+receivers:
```
cmacknz, line 1: We are going to need a way to include processors and compose pipelines. Possibly we'll want other component types as well.

Thinking about this a bit more, could we do something like the following:

* Instead of input.yml.hbs we functionally will pipeline.yml.hbs to define the integration pipeline in the collector.
* Each pipeline.yml.hbs defines a receiver, the processors and other components for it, and composes a pipeline with them in the expected order.
* The final exporter is omitted from the pipeline as it is controlled by Fleet.
  * Fleet would terminate the pipeline in this file with the forward connector and connect it to the configured exporter.
  * I say "final exporter" to leave using things like the routing connector to split into two pipelines in a configuration here, if that's what someone needed to do.

The direction would be to move from templating just the receiver, to being able to template an a pipeline that is wired into the correct exporter by Fleet.

Edit: update to suggest we should just omit the final pipeline exporter so that Fleet can control it. It will probably be the forward connector most of the time, but we don't want to lock ourselves into this.

jsoriano: I think this is more or less aligned with what we were thinking.

> Instead of input.yml.hbs we functionally will pipeline.yml.hbs to define the integration pipeline in the collector.

The name of the template is set by the developer [in the manifest](https://github.com/elastic/integrations/pull/14315/files#diff-891926bec56f4c067bf3e6615bf9f2785a9653a9e1fb0fa5cfc8788e8e340745R34), so it can really have any name.

> Each pipeline.yml.hbs defines a receiver, the processors and other components for it, and composes a pipeline with them in the expected order.

Yes, current files are expected to contain also processors, but by now we will focus on having a working receiver. Processing included in packages it not so relevant in the input use cases, but it will be in the integrations use cases.

In the integration receiver we also took the decision on allowing a single receiver per pipeline, I think we will end up with something pretty similar here.

> The final exporter is omitted from the pipeline as it is controlled by Fleet. 

Exactly, this is not expected to be defined in the input.

> Fleet would terminate the pipeline in this file with the forward connector and connect it to the configured exporter.

+1

> I say "final exporter" to leave using things like the routing connector to split into two pipelines in a configuration here, if that's what someone needed to do.

We may have multiple pipelines per package in the integrations use case, in a later stage. For inputs I would generate single pipelines.

On addition to what is defined in the package, we will also need to support user-defined custom processing. We will need to allow the user to somehow define processors in the policy, and append them in the correct place in the final pipeline.
We do this in general now with the `@custom` pipelines, and in some packages with a yaml setting for agent-side processing.

cmacknz: Thanks we are thinking in the same direction.

> Processing included in packages it not so relevant in the input use cases, but it will be in the integrations use cases.

> On addition to what is defined in the package, we will also need to support user-defined custom processing. We will need to allow the user to somehow define processors in the policy, and append them in the correct place in the final pipeline.

IMO our goal is to enable an OTel equivalent of integrations, which is going to require pairing generic inputs and processors together to get a complete solutions. There has to be a place for the processing to go in the first iteration of OTel packages/integrations we introduce or they'll be viewed as incomplete by users.

We also know we want the processing to be able to move between the edge and managed OTLP service though, so perhaps we'll need a way to define them separately so they can  be moved.

I like the idea of OTel "input" packages actually being "pipeline" packages that can define entire pipelines that terminate in a forward connector. This wouldn't give us a nice way to separate out processors though. The easiest solution I can think of is a "processors" package type where we allow connectors+receivers in one and processors in the other but that feels like kind of arbitrary division, and the other extreme is package per component which feels like too much.

cmacknz: Fleet could know to take the processors section of an input/package out and provide it to a separate API to put processing in the managed OTLP input.

There is no detailed design yet for service side OTTL so it is hard to predict what the right choice on the package side is. I do know we'll want processing on the edge to  match current integration capabilities and how to do this is a lot more obvious.

cmacknz: > We also know we want the processing to be able to move between the edge and managed OTLP service though, so perhaps we'll need a way to define them separately so they can be moved.

> Fleet could know to take the processors section of an input/package out and provide it to a separate API to put processing in the managed OTLP input.

I brought this up with @axw and it's not worth doing anything in the package spec to support moving processing around yet. It's too hard to know what the right choice is.

Let's direct the thinking into the best way to allow OTel integrations to include processors for processing that will happen on the edge.

jsoriano: Yes, I agree the goal is to have an equivalent to integration packages, but we considered having first the equivalent to input packages would help progressing, while giving something that is already useful. To support input packages we need to solve OTel collector config generation from policies, what will be also useful for integrations.

If we provide the same functionality as we have now for input packages, we will be able to use `@custom` pipelines for custom processing, though not on the edge.

Once we have input packages for OTel working at a similar level to our current input packages, we can elaborate on edge processing, and on full integrations.

jsoriano: > Let's direct the thinking into the best way to allow OTel integrations to include processors for processing that will happen on the edge.

Processors included in packages, and running in the edge, can be included in the template file, as we do now with beats-based inputs.

packages/httpcheck/_dev/test/policy/test-eprcheck.expected
```diff
@@ -0,0 +1,28 @@
...
+            indices: []
+        uuid-for-permissions-on-related-indices:
+            indices:
+                - names:
+                    - logs-*-*
+                  privileges:
+                    - auto_configure
+                    - create_doc
+receivers:
+    httpcheck/componentid:
```
jsoriano: Handling of ids in policy tests will need changes in elastic-package, something like https://github.com/elastic/elastic-package/pull/2799.

jsoriano:
Sorry for the noise! Closing, I will open another one.

PR #14318, opened 2025-06-25
[aws_billing] Add new `aws_billing` integration

ghost:
## Summary

This PR introduces a new integration for ingesting AWS Cost and Usage Report (CUR) 2.0 data from S3 buckets. The integration uses S3 polling with a minimum 24-hour interval (default) to avoid duplicate ingestion, as AWS CUR files are cumulative and updated multiple times daily.

The integration includes an hourly ES transform that groups billing data by key dimensions (service, account, usage type, region, tags) and aggregates unblended costs. This transform deduplicates the raw CUR data by creating uniquely identifiable data points.

For future upgrades, the transform configuration will require bumping both the `fleet_transform_version` and the `dest.index`. The `move_on_creation: true` setting is required as it ensures the destination index becomes the only member of the alias when created, automatically removing previous indices from the alias. This allows dashboards to continue displaying the data correctly.

I tested an actual upgrade by adding a new field to the transform's group by, then bumping the `fleet_transform_version` and `dest.index`. On upgrade, a new index `aws_billing.billing_latest-v2` was created and got the `aws_billing.billing_latest` alias. The old index `aws_billing.billing_latest-v1` lost the alias. The dashboards kept working fine because they use the alias, not the specific index names.

The integration uses defaults specific for the AWS CUR usecase:

- `bucket_list_interval` is set to 24h to avoid duplicate data
- `content_type` defaults to `text/csv`, the standard CUR format
- `file_selectors` defaults to `\.csv\.gz$` (this doesn't work as a default yet)

### Removed Configuration Options for `aws-s3`

I removed some config options that I thought are not relevant or applicable to this integration.

| Option |
|--------|
| [`expand_event_list_from_field`](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-aws-s3#_expand_event_list_from_field) |
| [`include_s3_metadata`](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-aws-s3#input-aws-s3-include_s3_metadata) |
| `max_number_of_messages` |
| [`queue_url`](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-aws-s3#_queue_url) |
| [`sqs_max_receive_count`](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-aws-s3#_sqs_max_receive_count) |
| [`sqs_wait_time`](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-aws-s3#_sqs_wait_time) |
| [`visibility_timeout`](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-aws-s3#_visibility_timeout) |

### Minor inconveniences

- Transform name starts with `logs-*` for some reason, this doesn't affect the functionality
- ~~Unable to set `file_selectors` default value to `\.csv\.gz$`~~

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Relates https://github.com/elastic/obs-integration-team/issues/202
- Closes https://github.com/elastic/obs-integration-team/issues/369


packages/aws_billing/data_stream/billing/manifest.yml
```diff
@@ -0,0 +1,177 @@
...
+        show_user: false
+        description: The file encoding to use for reading data that contains international characters. This only applies to non-JSON logs.
+      - name: file_selectors
+        type: yaml
+        title: File Selectors
+        multi: true
+        required: false
+        show_user: false
+        # default:
+        # - regex: '\.csv\.gz$'
```
ghost: this works if it's set in the UI but as a default it errors out when installing the integration in kibana @agithomas any ideas how to fix this?

agithomas: This dataset is closely associated with supporting only the `csv.gz` format. 

Hence, there may not be a need to make the settings configurable; instead can be statically assigned here

```
{{#if file_selectors}}
file_selectors:
{{file_selectors}}
{{/if}}
```

packages/aws_billing/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "0.1.0"
+  changes:
+    - description: Initial draft of the package
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/14318 # FIXME Replace with the real PR link
```
mykola-elastic: I guess this comment is no longer needed

mykola-elastic:
```suggestion
      link: https://github.com/elastic/integrations/pull/14318
```

packages/aws_billing/manifest.yml
```diff
@@ -0,0 +1,102 @@
...
+  license: "Elastic-2.0"
+description: Collect AWS CUR 2.0 billing data from S3 with Elastic Agent.
+type: integration
+categories:
+  - aws
+  - cloud
+  - custom
+conditions:
+  kibana:
+    version: "^8.18.0"
```
mykola-elastic: is it not meant to be supported on ^9.0.0?

ghost: the functionality we need is already available in version `8.18.0` (probably even in older versions), so unless there are significant bug fixes in the filebeat `awss3` input or other stack features in general, there's no strong justification for bumping the version :thinking: 

we could change it to `^8.16.5 || ^9.0.0"`

ghost:
Dashboards updated @agithomas @daniela-elastic 

aws_billing_all_time png
aws_billing_current_month

packages/aws_billing/data_stream/cur/sample_event.json
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+    },
+    "cloud": {
+        "provider": "aws"
+    },
+    "ecs": {
+        "version": "8.11.0"
+    },
+    "event": {
```
Linu-Elias, line 84: was the sample event generated or copied from elastic? I don't see few base fields like dataset and timestamp!

packages/aws_billing/changelog.yml
muthu-mps:
```suggestion
      link: https://github.com/elastic/integrations/pull/14318
```

packages/aws_billing/data_stream/cur/_dev/test/pipeline/test-aws-billing-without-resource-id.log-expected.json
```diff
@@ -0,0 +1,676 @@
+{
+    "expected": [
+        null,
+        {
+            "aws_billing": {
+                "cur": {
+                    "bill": {
+                        "billing_entity": "BillingEntity",
```
muthu-mps, line 8: Are we capturing the **cost_category** in this field? Does AWS represents this property as billing_entity? If no then we can change the field name with appropriate name.

packages/aws_billing/data_stream/cur/_dev/test/pipeline/test-aws-billing-without-resource-id.log-expected.json
```diff
@@ -0,0 +1,676 @@
+{
+    "expected": [
+        null,
```
muthu-mps, line 3: Why do we get `null` value here?

packages/aws_billing/data_stream/cur/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,780 @@
+---
+description: "X"
```
muthu-mps: nit: please fix this.

packages/aws_billing/data_stream/cur/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,780 @@
...
+processors:
+  - set:
+      field: ecs.version
+      value: '8.11.0'
```
muthu-mps: Can we set the ECS version to the latest?

packages/aws_billing/data_stream/cur/fields/fields.yml
```diff
@@ -0,0 +1,271 @@
+- name: aws_billing.cur
```
muthu-mps, line 1: Are we not adding field descriptions?

packages/aws_billing/data_stream/cur/fields/fields.yml
```diff
@@ -0,0 +1,271 @@
...
+    - name: cost_category
+      type: keyword
+    - name: cost
+      type: group
+      fields:
+        - name: amortized_cost
+          type: double
+        - name: net_amortized_cost
+          type: double
+    - name: discount
```
muthu-mps, line 50: Do we expect a field discount.discount? While looking into the [field format](https://github.com/elastic/integrations/pull/14318/files#diff-7419795195e9ce97705e09f796d3c6305a72fb96feefc218985312ad90260e4f:~:text=bill_payer_account_id%2Cbill_payer_account_name%2Ccost_category%2C-,discount,-%2Cdiscount_bundled_discount%2Cdiscount_total_discount%2Cidentity_line_item_id) in the pipeline tests it looks like, discount is different from bundled and total discount. IMO the discount should be a separate group and the total and bundled discount should not be grouped under discount. WDYT?

packages/aws_billing/manifest.yml
```diff
@@ -0,0 +1,106 @@
...
+title: "AWS Cost and Usage Report (CUR 2.0)"
+version: 0.1.0
+source:
+  license: "Elastic-2.0"
+description: Collect AWS CUR 2.0 billing data from S3 with Elastic Agent.
+type: integration
+categories:
+  - aws
+  - cloud
+  - custom
```
muthu-mps: Add **observability** category. Do we need **custom** or this could be replaced with **observability**?

packages/aws_billing/manifest.yml
```diff
@@ -0,0 +1,106 @@
...
+  kibana:
+    version: "^8.18.0"
+  elastic:
+    subscription: "basic"
+screenshots:
+  - src: /img/current_month.png
+    title: AWS Billing Current Month dashboard
+    size: 2522x1268
+    type: image/png
+  - src: /img/current_month.png
```
muthu-mps: The `current_month` image is added twice. Please update the name here.

packages/aws_billing/data_stream/cur/_dev/test/pipeline/test-aws-billing-without-resource-id.log-expected.json
ghost: because the first line in the csv file represents the column names, we have to drop it

packages/aws_billing/data_stream/cur/_dev/test/pipeline/test-aws-billing-without-resource-id.log-expected.json
ghost: The cost category values are stored in the `aws_billing.cur.cost_category` field. Some of these values are randomly generated, they are not real values from AWS.

packages/aws_billing/data_stream/cur/sample_event.json
ghost: It's mostly generated by me, not by a system test. The values are random.

packages/aws_billing/data_stream/cur/manifest.yml
```diff
@@ -0,0 +1,156 @@
...
+        default:
+          - forwarded
+      - name: processors
+        type: yaml
+        title: Processors
+        multi: false
+        required: false
+        show_user: false
+        description: >
+          Processors are used to reduce the number of fields in the exported event or to enhance the event with metadata. This executes in the agent before the logs are parsed. See [Processors](https://www.elastic.co/guide/en/beats/filebeat/current/filtering-and-enhancing-data.html) for details.
```
muthu-mps:
```suggestion
          Processors are used to reduce the number of fields in the exported event or to enhance the event with metadata. This executes in the agent before the logs are parsed. See [Processors](https://www.elastic.co/guide/en/fleet/current/elastic-agent-processor-configuration.html) for details.
```

packages/aws_billing/data_stream/cur/fields/fields.yml
ghost: I thought of renaming `_tmp.discount` to `aws_billing.cur.discount.discount`. We are not processing some of the fields from this list since in the first release of the integration we are focusing on the most important ones. Fields like `_tmp.discount` are usually maps that's why I left it out. I moved `bundled_discount` and `total_discount` outside of the `discount` group.

packages/aws_billing/_dev/build/docs/README.md
```diff
@@ -0,0 +1,96 @@
...
+Before using this integration, ensure the following:
+
+### AWS Cost and Usage Report (CUR) Export
+
+To use this integration, you must first create a **standard CUR 2.0 data export** in AWS. This will automatically create an S3 bucket if one does not already exist.
+
+#### Required CUR Export Configuration
+
+When creating your CUR export, the following **configuration is required** for compatibility:
+
```
agithomas, line 49: Lets mention about the need to selection the following options, explicitly

- How to navigate to the Data Exports options. It is not covered anywhere. 

- The user must select the Standard data export option
- The time granularity setting is not mentioned . (Hourly / Daily / Monthly). Can you test if all the above options are supported as part of this integration.?

packages/aws_billing/_dev/build/docs/README.md
```diff
@@ -0,0 +1,96 @@
...
+### Elastic Agent
+
+* You must install and configure [Elastic Agent](https://www.elastic.co/guide/en/fleet/current/elastic-agent-installation.html) to collect and ship data from the S3 bucket.
+* Only one Elastic Agent should be installed per host.
+
+## Setup
+
+1. In AWS, [create a CUR 2.0 export](https://docs.aws.amazon.com/cur/latest/userguide/cur-create.html) with the required settings.
+2. Ensure the report is stored in a reachable S3 bucket with proper permissions.
+3. Configure this integration in Elastic using the required AWS credentials or role.
```
agithomas, line 85: In other integrations, there is a mention of how to obtain AWS permissions , as shown below, to assist the user in generating the AWS credentials. Would it be better to mention the same here as well?

```
* **AWS Credentials** to connect with your AWS account.
* **AWS Permissions** to make sure the user you're using to connect has
  permission to share the relevant data.

For more details about these requirements, check the [AWS
integration
documentation](https://docs.elastic.co/integrations/aws#requirements).
```

packages/aws_billing/_dev/build/docs/README.md
ghost: > How to navigate to the Data Exports options. It is not covered anywhere.

```markdown
You can follow AWS documentation for setup:
🔗 [Creating Cost and Usage Reports in AWS](https://docs.aws.amazon.com/cur/latest/userguide/cur-create.html)
```

> The user must select the Standard data export option

```markdown
### AWS Cost and Usage Report (CUR) Export

To use this integration, you must first create a **standard CUR 2.0 data export** in AWS. This will automatically create an S3 bucket if one does not already exist.
```

> The time granularity setting is not mentioned . (Hourly / Daily / Monthly). Can you test if all the above options are supported as part of this integration.?

Data granularity does not matter for this integration, it's just the number of rows present in the report that will be affected.

packages/aws_billing/_dev/build/docs/README.md
ghost: Addressed in [2eec6f4](https://github.com/elastic/integrations/pull/14318/commits/2eec6f45160566fb2f803f3bbd8e81077f0a64c0)

packages/aws_billing/_dev/build/docs/README.md
agithomas: If i choose a monthly granularity, would the file be generated on a daily basis or on a monthly basis?

ghost: Report would be generated at the same frequency (mostly daily) it's just that higher granularity will have more rows in the report.

For example, if an EC2 instance runs continuously for 24 hours:
- With hourly granularity, the CUR will include 24 line items for that instance, one per hour.
- With daily granularity, it will include 1 aggregated line item for the entire day.
- With monthly granularity, it will have 1 aggregated line item for that same instance for the entire month (it can still change under specific conditions)

packages/aws_billing/data_stream/cur/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,780 @@
...
+      if: ctx.event?.original == null
+      description: 'Renames the original `message` field to `event.original` to store a copy of the original message. The `event.original` field is not touched if the document already has one; it may happen when Logstash sends the document.'
+  - remove:
+      field: message
+      ignore_missing: true
+      if: ctx.event?.original != null
+      description: 'The `message` field is no longer required if the document has an `event.original` field.'
+  - drop:
+      if: ctx.event?.original?.startsWith('bill_bill_type')
+      description: "Drop if logline contains header(s), which startswith `bill_bill_type`"
```
lucian-ioan, line 28: Are all headers removed here? Why is this needed instead of just storing the header information?

agithomas (approved):
LGTM!

muthu-mps (approved):
@gpop63 - Make sure the integration upgrade scenario is validated with transforms.
LGTM!

packages/aws_billing/data_stream/cur/elasticsearch/ingest_pipeline/default.yml
ghost: Since the reports are in CSV format, we have to drop the first line that contains the column names.

ghost:
@elastic/integrations-triaging hello :wave: can I get a review please? Code owner approval is required from this team.

packages/aws_billing/_dev/build/docs/README.md
```diff
@@ -0,0 +1,96 @@
+# AWS CUR 2.0 Billing
```
jsoriano, line 1: How does this package relate to the `billing` data stream in the AWS package? https://github.com/elastic/integrations/tree/main/packages/aws/data_stream/billing

packages/aws_billing/manifest.yml
jsoriano: @gpop63 do you want to support only 8.x, or also 9.x? If you want to support both, you need to add `|| ^9.0.0`.

packages/aws_billing/validation.yml
```diff
@@ -0,0 +1,4 @@
+errors:
+  exclude_checks:
+    - SVR00002
```
jsoriano, line 3: Couldn't the dashboards of this package contain filters? See https://github.com/elastic/elastic-package/blob/main/docs/howto/update_major_package_spec.md#expected-filter-in-dashboard-.

jsoriano (approved):
Approving to unblock, adding some questions and suggestions.

packages/aws_billing/validation.yml
ghost: We use a transform in this integration, the documents that we display are from that transform not from the raw data stream index, that's why I didn't add a dataset filter.

packages/aws_billing/_dev/build/docs/README.md
ghost: that one uses the Cost and Usage API this one uses Data Export CUR 2.0 reports, which provide much more granularity. CUR 2.0 provifdes detailed billing data that enables deeper and more precise cost analysis, whereas the Cost and Usage API offers more generalized estimates

Issue #12008, opened 2024-12-05
[New Integration] Qualys Web Application Scanning


clement-fouque:
## Description
In addition to Qualys VMDR integration, I would like to have an integration for the Qualys Web Application Scanning. This is one of the most used Qualys module (along with the VMDR one where we already have an integration). 

## Architecture
Qualys exposes a [JSON API](https://qualysguard.qg2.apps.qualys.com/portal-help/en/was/api_doc/online/index.htm) (not XML 🎉). 

## Dashboard Ideas
Will be updated later.

StacieClark-Elastic:
@clement-fouque what dashboards would you like to see for this integration?

clement-fouque:
@StacieClark-Elastic it's difficult to answer without knowing what data is available. FYI, here is the default Qualys Web Application Scanning dashboard with some hidden data.

![Image](https://github.com/user-attachments/assets/5cc3f9ed-230c-4cd9-aa80-0f5e6c1b0dcb)

PR #14322, opened 2025-06-25
[qualys_was] Added scripts to convert doubles to long. Added terminate Processor. Added IDs to web app tag list

StacieClark-Elastic:
## Proposed commit message
[qualys_was] Fixed several issues

- Added painless script to convert potential double values from cel scripts to longs.
- Changed web app tags to a list of tags with id and name instead of just a list of names. This is a breaking change
   since tags used to be a list of names and it's now a list of objects with id and name.
- Added a terminate processor to beginning of default.yml to fast fail on failing cel script.

## Checklist

- Relates #12008


packages/qualys_was/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "0.1.1"
+  changes:
+    - description: | 
+        Explicitly cast all long fields to long in case CEL sends them as doubles
+        Add terminate processor at start of default.yml
+        Changed web app tags from list of names to list of tag objects with name and id
```
efd6:
```suggestion
        Explicitly cast all long fields to long in case the agent sends them as doubles.
        Terminate ingest pipeline early if agent sends an error message.
        Change web app tags from list of names to list of tag objects with name and id.
```

packages/qualys_was/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -76,9 +75,28 @@ processors:
...
+  - script:
+      if: ctx.qualys_was?.vulnerability?.id != null
+      tag: vul_id_is_long
+      lang: painless
+      source: >
+        if (ctx.qualys_was.vulnerability.id instanceof String) {
+          ctx.qualys_was.vulnerability.id = Long.parseLong(ctx.qualys_was.vulnerability.id);
+        } else {
+          ctx.qualys_was.vulnerability.id = (long)ctx.qualys_was.vulnerability.id;
+        }
```
efd6, line 94: I think we probably want `on_failure`s for these.

packages/qualys_was/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -179,20 +201,37 @@ processors:
...
-          value:  "{{{ _ingest._value.Tag.name}}}"
-          allow_duplicates: false
-          on_failure:
-            - append:
-                field: error.message
-                value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - rename:
+      field: json.detection.webApp.id
+      tag: rename_webApp_url
+      target_field:  qualys_was.vulnerability.web_app.id
```
efd6, line 230: `ignore_missing`?

packages/qualys_was/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: done

packages/qualys_was/changelog.yml
StacieClark-Elastic: done

packages/qualys_was/docs/README.md
```diff
@@ -657,5 +663,6 @@ An example event for `vulnerability` looks as following:
 | qualys_was.vulnerability.wasc_references.url | WASC reference URL. Available in verbose mode. | keyword |
 | qualys_was.vulnerability.web_app.id | Web Application ID. | long |
 | qualys_was.vulnerability.web_app.name | Web Application name. | keyword |
-| qualys_was.vulnerability.web_app.tags | Web Application tags. Available in verbose mode. | keyword |
+| qualys_was.vulnerability.web_app.tags.id | ID of tag. | long |
```
efd6, line 666: Probably too late for this question, but why is `… .tags.id` a long? It should probably have been a keyword (same for `… .web_app.id` — which is why the question is too late).

efd6 (approved):
LGTM, but query.

PR #14323, opened 2025-06-26
amazon_security_lake: fix handling of JSON encoded flattened fields

efd6:
## Proposed commit message
```
amazon_security_lake: fix handling of JSON encoded flattened fields

The ocsf.api.{request,response}.data fields may be strings holding JSON
encoded data, but the fields are mapped as flattened, resulting in
failure to index documents. Conditionally JSON decode the string and if
this fails, put the value in a well known location in an object so that
the flattened mapping does not fail.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/amazon_security_lake/data_stream/event/_dev/test/pipeline/test-application-activity.log-expected.json
```diff
@@ -2210,7 +2210,9 @@
                     },
                     "response": {
                         "code": 96,
-                        "data": "phenomenon",
+                        "data": {
+                            "value": "phenomenon"
+                        },
```
ShourieG, line 2215: Could this lead to a mapping conflict since it's changing from string to object ?

efd6: The field is [defined as flattened](https://github.com/elastic/integrations/blob/e35e24a6d157a53bc73daff1cc2c644f3932e854/packages/amazon_security_lake/data_stream/event/fields/api-fields.yml#L121-L123). In reality, the test is incorrect if the field definition is correct. It's unfortunate that elastic-package does not check for this; the elastic-package code does have `isFieldTypeFlattened`, but it is not used for checking, instead being used to avoid checking children.

If there was a change that leads to this problem, it would have been #10405, but it needed to happen.

issue: elastic/elastic-package#2695

ShourieG (approved):
LGTM

PR #14336, opened 2025-06-26
[microsoft_dnsserver] Enrich dns.question.* fields

brijesh-elastic:
## Proposed commit message

```
microsoft_dnsserver: Enrich dns.question.* ECS fields for both datastream
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/microsoft_dnsserver directory.
> elastic-package test

## Related issues

- Related https://github.com/elastic/enhancements/issues/24796


kcreddy (approved):

kcreddy:
@elastic/sec-windows-platform, can we get a PR review for this minor change?

packages/microsoft_dnsserver/data_stream/audit/_dev/test/pipeline/test-events.json
```diff
@@ -118,7 +118,7 @@
                 "channel": "Microsoft-Windows-DNSServer/Audit",
                 "event_data": {
                     "Scope": "Default",
-                    "QNAME": "B.ROOT-SERVERS.NET.",
```
andrewkroh: Why is the test data being changed? 

The test data should reflect what Windows puts into these fields. Trailing dots are common in DNS data as this indicates it is a FQDN (rooted in the root-zone of (`.`)). So I would not be surprised if the trailing dots are supposed to be there.

andrewkroh: The screenshot here has `QNAME=example.com.`:

https://learn.microsoft.com/en-us/windows-server/networking/dns/dns-logging-and-diagnostics?tabs=desktop-experience

brijesh-elastic:
> I noticed the input data for one of our existing tests was modified. Usually, we try to keep these tests stable to catch regressions. Could you walk me through the reasoning for the change? My primary concern is that we want to ensure the new code handles the original data case, as it likely represents a real-world scenario we need to support.
> 
> What do you think about restoring the original test data and updating the implementation to make it pass (assuming it does not)?


Earlier, I thought the test case in the file was incorrect, and I modified it to parse out the current implementation. However, I now understand that the trailing `.` in the DNS data is a part of FQDN, and it is expected.
I will update the implementation to handle it and restore the original test data.

packages/microsoft_dnsserver/data_stream/analytical/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -364,6 +380,7 @@ processors:
       tag: append_dns_header_flag_TC
   - remove:
       field:
+        - _temp.question_name
```
andrewkroh: You want to remove the whole `_temp` object. If you only delete the `question_name` field it leaves an empty `_temp: {}` object.

```suggestion
        - _temp
```

PR #14354, opened 2025-06-27
[aws] Ingest managed insights from Security Hub

### LINKED ISSUE #9719: [aws] Security Hub integration to ingest managed insights
Our AWS integration can [collect insights and findings](https://docs.elastic.co/integrations/aws/securityhub) from AWS Security Hub.

For insights, we use the [`GetInsights` endpoint](https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_GetInsights.html) of the Security Hub API, and receive all **custom insights**, since we don't request specific ARNs. However, there are also **managed insights**, which will only be returned if they are requested by ARN.

There are 35 managed insight ARNs listed in the [AWS Security Hub documentation](https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-managed-insights.html). Our integration could make request for a hard-coded list of managed insight URNs. (The API doesn't provide any way to list managed insight ARNs).

Note that this would need to be implemented with separate requests for the managed insights (requested by ARN) and the custom insights (requested without ARNs).

chrisberkhout:
## Proposed commit message

```
[aws] Ingest managed insights from Security Hub

The AWS Security Hub API's GetInsights endpoint[1] returns only custom
insights. Managed insights[2] are built-in, and can only be retrieved
by ARN.

Fetch a hard-coded list of managed insights first, then fetch pages of
custom insights as before.

This has been tested against the live API.

[1]: https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_GetInsights.html
[2]: https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-managed-insights.html
```

## Checklist

## Related issues

- Closes #9719

### DIFF
--- a/packages/aws/data_stream/securityhub_insights/agent/stream/httpjson.yml.hbs
+++ b/packages/aws/data_stream/securityhub_insights/agent/stream/httpjson.yml.hbs
@@ -81,7 +81,9 @@ response.pagination:
       # This transform's job is to fail and end pagination when finished.
       # It will set a value if the last request was the first in the sequence
       # (which collects the managed insights), or if there is another page of
-      # custom insights to fetch.
+      # custom insights to fetch. Otherwise, there will be a value template
+      # error (.last_response.body.NextToken doesn't exist) and the pagination
+      # sequence will end (because fail_on_template_error is set).
       target: body.NextToken
       value: '[[if (eq (.last_response.page) 0)]][["placeholder"]][[else]][[.last_response.body.NextToken]][[end]]'
       fail_on_template_error: true


packages/aws/data_stream/securityhub_insights/agent/stream/httpjson.yml.hbs
```diff
@@ -23,15 +23,77 @@ request.transforms:
...
 {{!-- https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html --}}
 response.pagination:
+  - delete:
+      # Only request specific ARNs in the first request (for managed insights).
+      target: body.InsightArns
   - set:
+      # This transform's job is to fail and end pagination when finished.
+      # It will set a value if the last request was the first in the sequence
+      # (which collects the managed insights), or if there is another page of
+      # custom insights to fetch.
```
efd6: There's enough HTTPJSON magic here that I think the comment needs to explain more about how it is actually working.

efd6 (commented):
In commit message, s/retieved/retrieved/.

packages/aws/data_stream/securityhub_insights/agent/stream/httpjson.yml.hbs
chrisberkhout: I added a sentence. Here's the full new comment:

```
      # This transform's job is to fail and end pagination when finished.
      # It will set a value if the last request was the first in the sequence
      # (which collects the managed insights), or if there is another page of
      # custom insights to fetch. Otherwise, there will be a value template
      # error (.last_response.body.NextToken doesn't exist) and the pagination
      # sequence will end (because fail_on_template_error is set).
```

PR #14373, opened 2025-06-30
[Cloud Security] Bump Asset Discovery and CSPM version for 8.19 and 9.1 release

seanrathier:
## Proposed commit message

Bump Asset Discovery and CSPM version for 8.19 and 9.1 release

## Checklist


packages/cloud_asset_inventory/changelog.yml
```diff
@@ -1,6 +1,11 @@
 # newer versions go on top
 # version map:
 # 0.1.x - 8.15.x
+- version: "1.0.0"
```
maxcold: do we need this bump? we don't have any changes in the preview version, therefore we don't have a requirement to upgrade it really.

packages/cloud_security_posture/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.0.0"
```
maxcold, line 2: We need to remove all the preview versions from changelog, and have all the changes under `2.0.0` . See https://github.com/elastic/integrations/pull/13053/files for example
also there is a version map comment which for some reason was moved after 2.0.0, let's make sure it's on the top.

packages/cloud_asset_inventory/changelog.yml
seanrathier: @opauloh ?

opauloh: Just cross-posting here what we discussed in Slack: Asset Discovery is not yet released, so there's no bump needed as it is still in tech preview.

packages/cloud_security_posture/changelog.yml
opauloh: We are going to wait for the QA cycle to be completed before merging the bump, correctly?

maxcold: I remember we changed when we release and I think our docs reflect that `The release process of the cloud_security_posture integration aligns with the stack release. Still, we release the package version a bit earlier (before out team's 2nd QA cycle) to give ample time to test the released integration version with the yet-to-be-released stack version.` . @kfirpeled you probably should remember, is it the current process, release integration after 1st QA cycle?

opauloh: That "release in between QAs" workflow makes sense to me, so we would perform a first round of QA before releasing the integration, and the second QA cycle would have the published version.

packages/cloud_security_posture/changelog.yml
```diff
@@ -1,9 +1,21 @@
...
+    - description: Remove project_id and organization_id from validation
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/13866
+    - description: Fix the Azure credentials validation
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/13714
+    - description: Populate event.outcome field with value from result.evaluation
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/13582
 # version map:
```
maxcold, line 19: @seanrathier can you move this version map comment above 2.0.0 changelog item?

maxcold:
We need to make sure that https://github.com/elastic/kibana/issues/226037 is resolved before releasing 2.0.0 version of our integration

maxcold:
and we also need to figure out the situation with the support for both 8.19 and 9.1 versions, it looks like we have some changes that would only work in 9.1 and some that work in both 8.19 and 9.1. It seems like we will need to release two versions using the backport branch. One supporting 8.19 and another one supporting 9.1

seanrathier:
Closing as the version bump was already merged

PR #14381, opened 2025-07-01
wiz: fix description of client ID and secret fields

JordanSh:
Simple change to the helper text in the interrogation's form input fields for better readability

fixes these texts:

image


JordanSh:
> This will need a changelog entry and a manifest version bump.

Thank you :) I wasn't sure if I should update the version for a minor change. I've now updated the changelog and the manifest version as well

packages/wiz/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.4.1"
+  changes:
+    - description: Update texts for the input fields helpers.
+      type: enhancement
```
efd6:
```suggestion
      type: bugfix
```
(if you think it's a bugfix — I think it probably is since the previous text was ungrammatical — otherwise bump minor; enhancements bump minor, bugfixes bump patch, breaking changes bump major)

efd6 (approved):
Thanks

PR #14384, opened 2025-07-01
Update the Cloudflare Logpush integration data streams intro

alaudazzi:
This PR updates the Data streams section with a link to the [Logs reference](https://www.elastic.co/docs/reference/integrations/cloudflare_logpush#logs-reference) section.

Closes https://github.com/elastic/integration-docs/issues/406


packages/cloudflare_logpush/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.38.4"
```
efd6:
```suggestion
- version: "1.39.0"
```

packages/cloudflare_logpush/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: "3.0.2"
 name: cloudflare_logpush
 title: Cloudflare Logpush
-version: "1.38.3"
+version: "1.38.4"
```
efd6:
```suggestion
version: "1.39.0"
```

packages/cloudflare_logpush/docs/README.md
```diff
@@ -164,16 +169,17 @@ curl --location --request POST 'https://api.cloudflare.com/client/v4/zones/<ZONE
...
+1. In Kibana, go to **Management** > **Integrations**.
 2. In the integrations search bar type **Cloudflare Logpush**.
 3. Click the **Cloudflare Logpush** integration from the search results.
 4. Click the **Add Cloudflare Logpush** button to add Cloudflare Logpush integration.
 5. Enable the Integration with the HTTP Endpoint, AWS S3 input or GCS input.
 6. Under the AWS S3 input, there are two types of inputs: using AWS S3 Bucket or using SQS.
 7. Configure Cloudflare to send logs to the Elastic Agent via HTTP Endpoint, or any R2, AWS or GCS Bucket following the specific guides above.
 
-## Logs reference
+## Logs reference [logs-reference]
```
kcreddy: This shows up like this in the integrations page inside the elastic stack.

image


Not sure if this is okay or how it would render inside published docs.

alaudazzi: Thanks for flagging this, @kcreddy. This is the format we use to create internal links for the public docs, but it doesn't work in Kibana. I'll remove it for now.

chrisberkhout (approved):
The changes look good (after removing the link anchor).

One other issue I notice here is that it talks about collecting from "AWS S3 Bucket" and "AWS SQS" as if they are different sources, when it's really the same source, with different configuration for how new data is detected.

packages/cloudflare_logpush/docs/README.md
alaudazzi: Removed.

PR #14389, opened 2025-07-02
cisco_meraki: fix ECS mapping for translated IP and port

### LINKED ISSUE #14385: [cisco_meraki]: Align NAT field mapping with ECS standards in IP flows pipeline
### Integration Name

Cisco Meraki [cisco_meraki]

### Dataset Name

cisco_meraki.log

### Integration Version

1.29.1

### Agent Version

8.15.5

Ubuntu 24.04 LTS (x86_64)

### User Goal

I need the Cisco Meraki integration's ipflow ingest pipeline to correctly map NAT fields according to the Elastic Common Schema (ECS) standards. Specifically, I would like translated IP addresses and ports to be stored in the appropriate NAT fields (source.nat.ip and source.nat.port) while preserving the original source information in the standard fields (source.ip and source.port).

### Existing Features

The current implementation causes two issues:

- Data loss due to conditional logic: When `translated_src_ip` exists, the original source IP in `src` is not stored anywhere, resulting in loss of the internal/private IP address information. This hinders our ability to make targeted rule exceptions for alerts.

- Incorrect ECS mapping: The pipeline places `translated_src_ip` (the NAT'd public IP) into `source.ip`, which contradicts ECS guidelines. NAT'd addresses should be stored in `source.nat.ip`, while the original source IP `src` should remain in `source.ip`.

### What did you see?

Example of a IP flow event:
```
<134>1 1751379284.245040794 FW_01 ip_flow_start src=10.140.40.72 dst=8.8.8.8 protocol=udp sport=18212 dport=53 translated_src_ip=1.128.3.4 translated_port=13710
```

### Current behavior

- `source.ip`: 1.128.3.4 (translated/NAT'd IP)
- `source.port`: 13710 (translated port)
- Original IP (10.140.40.72) and port (18212) are not stored

### Expected behavior
- `source.ip`: 10.140.40.72 (original internal IP)
- `source.port`: 18212 (original port)
- `source.nat.ip`: 1.128.3.4 (NAT'd public IP)
- `source.nat.port`: 13710 (NAT'd port)

### Anything else?

This is the part of the ingest pipeline that is responsible for the behaviour: https://github.com/elastic/integrations/blob/08dd8a22376437fc37f332983467aeec2f175c3f/packages/cisco_meraki/data_stream/log/elasticsearch/ingest_pipeline/ipflows.yml#L19-L39.

### Proposed solution
The pipeline should be updated to always map fields as follows:

- `src` → `source.ip` (always)
- `sport` → `source.port` (always)
- `translated_src_ip` → `source.nat.ip` (when present)
- `translated_port` → `source.nat.port` (when present)

This change should also be extended to destination NAT fields (`translated_dst_ip` → `destination.nat.ip`, etc.) if they exist in the pipeline for consistency.

This change would:

- Preserve all network information (both internal and external IPs/ports)
- Align with ECS standards for NAT field mapping
- Enable better network traffic analysis, troubleshooting and more precise rule exclusions for alerts
- Maintain consistency with how other integrations handle NAT IPs

### References

- [ECS NAT Fields Documentation](https://www.elastic.co/guide/en/ecs/current/ecs-source.html#field-source-nat-ip)
- Similar NAT handling can be seen in other integrations like [Palo Alto Networks](https://github.com/elastic/integrations/blob/main/packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/traffic.yml) and [Fortinet FortiGate](https://github.com/elastic/integrations/blob/main/packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/traffic.yml)

efd6:
## Proposed commit message
```
cisco_meraki: fix ECS mapping for translated IP and port

Previously these were conditionally mapped to the {source,destination}.{ip,port}
fields. Instead, map them to {source,destination}.nat.{ip,port}.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #14385


packages/cisco_meraki/data_stream/log/_dev/test/pipeline/test-ip-flow.log-expected.json
```diff
@@ -569,6 +517,62 @@
...
+            "event": {
+                "category": [
+                    "network"
+                ],
+                "original": "<134>1 1751379284.245040794 FW_01 ip_flow_start src=10.140.40.72 dst=81.2.69.144 protocol=udp sport=18212 dport=53 translated_src_ip=1.128.3.4 translated_port=13710",
+                "type": [
+                    "info"
+                ]
+            },
+            "message": "src=10.140.40.72 dst=81.2.69.144 protocol=udp sport=18212 dport=53 translated_src_ip=1.128.3.4 translated_port=13710",
```
kcreddy, line 530: `translated_port=13710` is being set into both `source.nat.port` and `destination.nat.port`. Just wanted to double-check if this is correct.

kcreddy (approved):
Just a clarification. LGTM otherwise.

packages/cisco_meraki/data_stream/log/_dev/test/pipeline/test-ip-flow.log-expected.json
efd6: No, that's not right, but it looks like it was the original behaviour. I'll fix it.

PR #14424, opened 2025-07-04
gitlab: fix null check condition and field type for production data stream

chemamartinez:
## Proposed commit message

```
Fixing two issues for the production data stream:
- Fix a pipeline error when url.path is null.
- Update the field type for gitlab.production.params.variables from
  keyword to flattened, this field could either contain strings or objects.
  This change avoids an ingestion error when the field contains an object.
```

## Checklist


packages/gitlab/data_stream/production/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -237,7 +237,7 @@ processors:
     value:
     - web
     allow_duplicates: false
-    if: ctx.url?.path.contains("/search/")
+    if: ctx.url?.path != null && ctx.url.path.contains("/search/")
```
chrisberkhout, line 237: An alternative here is
```diff
-    if: ctx.url?.path.contains("/search/")
+    if: ctx.url?.path?.contains("/search/") == true
```

packages/gitlab/data_stream/production/fields/fields.yml
```diff
@@ -168,7 +168,7 @@
             - name: search
               type: keyword
             - name: variables
-              type: keyword
+              type: flattened
```
chrisberkhout, line 171: The system test fails to ingest the event with 
```
"params":[...,{"key":"variables","value":"[FILTERED]"},...
```

Flattened doesn't take any JSON value (like a string), it specifically needs an object.

The test passes if the data is modified like this:

```diff
-{"key":"variables","value":"[FILTERED]"}
+{"key":"variables","value":{"[FILTERED]":null}}
```

There's a similar issue with pipeline test data (but it doesn't fail because it doesn't actually try to ingest it).

packages/gitlab/data_stream/production/elasticsearch/ingest_pipeline/default.yml
chrisberkhout, line 1: For `gitlab.production.params`, rather than turning an array of key-value pairs into an array of single-key objects, it seems more natural to me to turn them into one object:

```diff
diff --git a/packages/gitlab/data_stream/production/elasticsearch/ingest_pipeline/default.yml b/packages/gitlab/data_stream/production/elasticsearch/ingest_pipeline/default.yml
index c7779dc8f8..8ecb37ec49 100644
--- a/packages/gitlab/data_stream/production/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/gitlab/data_stream/production/elasticsearch/ingest_pipeline/default.yml
@@ -141,14 +141,11 @@ processors:
     lang: painless
     description: Move the objects in `params.value` to their own key value pairs
     source: |
-      def keyValuePairs = [];
+      Map map = [:];
       for (item in ctx.gitlab.production.params) {
-        def key = item.key;
-        def value = item.value;
-        def keyValueObject = [key: value];
-        keyValuePairs.add(keyValueObject)
+        map[item.key] = item.value;
       }
-      ctx.gitlab.production.params = keyValuePairs;
+      ctx.gitlab.production.params = map;
 - geoip:
     field: source.ip
     tag: geoip_source_ip
```

packages/gitlab/docs/README.md
```diff
@@ -1197,7 +1197,7 @@ Collect logs for Rails controller requests received from GitLab. Check out the [
 | gitlab.production.params.operationName |  | keyword |
 | gitlab.production.params.query |  | keyword |
 | gitlab.production.params.search |  | keyword |
-| gitlab.production.params.variables |  | keyword |
+| gitlab.production.params.variables |  | flattened |
```
chrisberkhout, line 1200: I notice there are other fields called `*.variables` that are mapped to `keyword`.

Should they be handled similarly?

packages/gitlab/data_stream/production/fields/fields.yml
chemamartinez: Yes... the problem is that I couldn't find any log sample apart from the one reported in the issue and the log samples here were the field value was redacted. It also seems like we need a dev licence to generate these logs and docs don't give a clue either. So I am not sure if that field must contain always objects or can also contain strings (as it was mapped as keyword in first place).

So I don't know if we better assume they should always be an object and update the test logs, or keep the field as keyword and convert it when it contains objects (https://github.com/elastic/integrations/pull/14424/commits/f2c5bfb339a30bbc44d0665033d1ae7a852a32d8).

packages/gitlab/docs/README.md
chemamartinez: I'd say it is logical to think they would follow the same format, but we don't have any examples of these fields containing objects.

Right now the issue is with this particular field, and we have a log sample with that field being an object, I'd keep the things as they were designed until we get some proves of these fields should be flattened as well.

chrisberkhout: Looking at https://docs.gitlab.com/administration/logs/#production_jsonlog, I think that in general these logs are API request logs, and `params` is usually the parameters from a query string or URL-encoded request body.

However, the examples from the other issue are GraphQL requests. GraphQL requests often have a generic query and some variables for the specific case ([example](https://graphql.org/learn/queries/#variables)).

The gitlab documentation also has an GraphQL example where a `variables` key gets an object value: https://docs.gitlab.com/administration/logs/#graphql_jsonlog.

So I think the params values will often be strings, but if it's a GraphQL request and param key is `variables` the value will probably be an object.

Those are the values to expect, but flattened or serialized as keyword are both fine. Keyword seems more flexible.

packages/gitlab/data_stream/production/fields/fields.yml
chrisberkhout: I think either way is fine. I put some stuff I found in the documentation in another comment.

packages/gitlab/data_stream/production/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -145,7 +145,13 @@ processors:
       for (item in ctx.gitlab.production.params) {
         def key = item.key;
         def value = item.value;
-        def keyValueObject = [key: value];
+        def valueString;
+        if (key == "variables") {
+          valueString = value != null ? value.toString() : "";
```
chrisberkhout: Since this will be JSON data, I think it's better to serialize it with `Json.dump(value)`. It could also use `instanceof` to check.

packages/gitlab/docs/README.md
chemamartinez: Agree. Thanks for expanding this. I'd say in this case to go for flattened type for all these variables fields for simplicity and also for better formatting on final events.

chrisberkhout (approved):

PR #14430, opened 2025-07-07
snyk: add attributes.status and attributes.updated_at field values to fingerprint

### LINKED ISSUE #14356: [Snyk]: Fingerprint should include the status
### Integration Name

Snyk [snyk]

### Dataset Name

snyk.issues

### Integration Version

2.0.0

### Agent Version

9.0.1

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.0.1

Docker image (elastic-package)

### Software/API Version

Snyk SaaS

### Error Message

No error message

### Event Original

No event original

### What did you do?

I ingested Snyk issues into the stack. Based on the way the fingerprint is set up, there should be 

### What did you see?

https://github.com/elastic/integrations/blob/7d3232e322975fb5bcee9cfe6479a85bdddcf086/packages/snyk/data_stream/issues/elasticsearch/ingest_pipeline/default.yml#L28-L31

Add a field (like the vulnerability status) as part of the fingerprint.

### Anything else?

This was never tested/confirmed. It's based of my analysis of the integration and a review from @SpencerLN and @christophercutajar. As the GitHub Secret Scanning contains multiple fields, it confirms our analysis. 

https://github.com/elastic/integrations/blob/7d3232e322975fb5bcee9cfe6479a85bdddcf086/packages/github/data_stream/secret_scanning/elasticsearch/ingest_pipeline/default.yml#L31-L37

efd6:
## Proposed commit message

See title.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #14356

### DIFF
--- a/packages/snyk/changelog.yml
+++ b/packages/snyk/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.2.0"
+  changes:
+    - description: Add `attributes.status` and `attributes.updated_at` field values to fingerprint.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/14430
 - version: "2.1.0"
   changes:
     - description: Add maximum executions configuration option for data streams.
--- a/packages/snyk/changelog.yml
+++ b/packages/snyk/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "2.2.0"
   changes:
-    - description: Add `attributes.status` and `attributes.updated_at` field values to fingerprint.
+    - description: Add `attributes.ignored`, `attributes.status` and `attributes.updated_at` field values to fingerprint.
       type: enhancement
       link: https://github.com/elastic/integrations/pull/14430
 - version: "2.1.0"
--- a/packages/snyk/data_stream/issues/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/snyk/data_stream/issues/elasticsearch/ingest_pipeline/default.yml
@@ -28,6 +28,7 @@ processors:
   - fingerprint:
       fields:
         - snyk.issues.id
+        - snyk.issues.attributes.ignored
         - snyk.issues.attributes.status
         - snyk.issues.attributes.updated_at
       target_field: _id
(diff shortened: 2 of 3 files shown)


efd6:
Waiting for @clement-fouque.

packages/snyk/data_stream/issues/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -28,7 +28,10 @@ processors:
   - fingerprint:
       fields:
         - snyk.issues.id
+        - snyk.issues.attributes.status
```
clement-fouque, line 32:
```suggestion
        - snyk.issues.attributes.ignored
        - snyk.issues.attributes.status
```

packages/snyk/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.2.0"
+  changes:
+    - description: Add `attributes.status` and `attributes.updated_at` field values to fingerprint.
```
clement-fouque:
```suggestion
    - description: Add `attributes.ignored`, `attributes.status` and `attributes.updated_at` field values to fingerprint.
```

clement-fouque (changes requested):
I reviewed the API and we have 2 options:
- Either we add the `ignored` field along with `status` and `updated_at`
- Either we remove the `status` as the `updated_at` _should_ be updated anyway

I'm proposing the first option (all fields) as it's clearer to understand what is affecting fingerprint and it protects from issues on Snyk API (if status changes and `updated_at` field is not updated).

packages/snyk/data_stream/issues/elasticsearch/ingest_pipeline/default.yml
efd6: Why is this necessary?

clement-fouque: Please see my comment in the PR

efd6: So the position is that it may be the case that ignored changes but updated does not? Looking at the API, I don't think that is the case. The reason I included status in the fingerprint is to avoid timestamp collision even though that would be very unlikely. I have added the ignored field though.

efd6:
Not prioritised.

PR #14434, opened 2025-07-07
[auth0] Improve webhook setup instructions

chrisberkhout:
## Proposed commit message

See title.

## Checklist

## Screenshots

Here's what the policy editor UI form looks like with the manifest changes:

Screenshot 2025-07-07 at 11-18-14 Add integration - Auth0 - Integrations - Elastic

### DIFF
--- a/packages/auth0/_dev/build/docs/README.md
+++ b/packages/auth0/_dev/build/docs/README.md
@@ -26,7 +26,7 @@ This integration method is also documented on the Auth0 webpage for the [Elastic
 3. Enter value for **Authorization Token**. This must match the value entered when configuring the "Custom Webhook" in Auth0 cloud.
 4. In the "Advanced options" section, enter settings for **SSL Configuration**. Auth0 requires that webhook requests use HTTPS. So you must either configure the integration with a valid TLS certificate here, or use a separarately configured reverse proxy in front of the agent.
 
-Using the external address of the agent to which the integration is added, and the configured "Listen Port" and "Webhook Path", make a note of the full endpoint URL. It will have the form `https://{AGENT_ADDRESS}:{LISTEN_PORT}{WEBHOOK_PATH}` (for example, `https://agent01.external.example.com:8383/auth0/logs`).
+Using the external address of the agent to which the integration is added, and the configured "Listen Port" and "Webhook Path", make a note of the full endpoint URL. It will have the form `https://{AGENT_ADDRESS}:{LISTEN_PORT}/{WEBHOOK_PATH}` (for example, `https://agent01.external.example.com:8383/auth0/logs`).
 
 ### Creating the stream in Auth0
 
--- a/packages/auth0/_dev/build/docs/README.md
+++ b/packages/auth0/_dev/build/docs/README.md
@@ -34,7 +34,7 @@ Using the external address of the agent to which the integration is added, and t
 2. Choose **Custom Webhook**.
 3. Name the new **Event Stream** appropriately (e.g. Elastic) and click **Create**.
 4. In **Payload URL**, enter the endpoint URL set up in the **Configure the Auth0 integration** section.
-5. In **Authorization Token**, enter the **Authorization Token**. This must match the value entered in Step 3 of **Configure the Auth0 integration** section.
+5. In **Authorization Token**, enter the **Authorization Token**. This must match the value entered in Step 3 of the **Configure the Auth0 integration** section.
 6. In **Content Type**, choose  **application/json**.
 7. In **Content Format**, choose **JSON Lines**.
 8. Click **Save**.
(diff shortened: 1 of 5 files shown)


packages/auth0/_dev/build/docs/README.md
```diff
@@ -17,22 +17,24 @@ The package collects log events either sent via log stream webhooks, or by API r
...
 
 ### Creating the stream in Auth0
 
 1. From the Auth0 management console, navigate to **Logs > Streams** and click **+ Create Stream**.
 2. Choose **Custom Webhook**.
 3. Name the new **Event Stream** appropriately (e.g. Elastic) and click **Create**.
-4. In **Payload URL**, paste the **Endpoint URL** collected during Step 1 of **Configure the Auth0 integration** section.
-5. In **Authorization Token**, paste the **Authorization Token**. This must match the value entered in Step 2 of **Configure the Auth0 integration** section.
+4. In **Payload URL**, enter the endpoint URL set up in the **Configure the Auth0 integration** section.
+5. In **Authorization Token**, enter the **Authorization Token**. This must match the value entered in Step 3 of **Configure the Auth0 integration** section.
```
alaudazzi:
```suggestion
5. In **Authorization Token**, enter the **Authorization Token**. This must match the value entered in Step 3 of the **Configure the Auth0 integration** section.
```

packages/auth0/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.23.0"
+  changes:
+    - description: Improve webhook setup instructionsn
```
efd6:
```suggestion
    - description: Improve webhook setup instructions.
```

packages/auth0/_dev/build/docs/README.md
```diff
@@ -17,22 +17,24 @@ The package collects log events either sent via log stream webhooks, or by API r
...
-3. Enter value for "Secret value". This must match the "Authorization Token" value entered when configuring the "Custom Webhook" from Auth0 cloud.
-4. Enter values for "TLS". Auth0 requires that the webhook accept requests over HTTPS. So you must either configure the integration with a valid TLS certificate or use a reverse proxy in front of the integration.
+2. Enter values for **Listen Address**, **Listen Port** and **Webhook Path**.
+3. Enter value for **Authorization Token**. This must match the value entered when configuring the "Custom Webhook" in Auth0 cloud.
+4. In the "Advanced options" section, enter settings for **SSL Configuration**. Auth0 requires that webhook requests use HTTPS. So you must either configure the integration with a valid TLS certificate here, or use a separarately configured reverse proxy in front of the agent.
+
+Using the external address of the agent to which the integration is added, and the configured "Listen Port" and "Webhook Path", make a note of the full endpoint URL. It will have the form `https://{AGENT_ADDRESS}:{LISTEN_PORT}{WEBHOOK_PATH}` (for example, `https://agent01.external.example.com:8383/auth0/logs`).
```
efd6, line 29:
```suggestion
Using the external address of the agent to which the integration is added, and the configured "Listen Port" and "Webhook Path", make a note of the full endpoint URL. It will have the form `https://{AGENT_ADDRESS}:{LISTEN_PORT}/{WEBHOOK_PATH}` (for example, `https://agent01.external.example.com:8383/auth0/logs`).
```
? (arguable depending on the details of WEBHOOK_PATH; we don't explicitly say that it should have a leading slash, but maybe it should in which case the text here would be fine — note that the URL rfc does not assume that a path has a leading slash, it is explicitly included in the grammar)

chrisberkhout: Actually, I reverted this one, because the default value for Webbook Path does include a leading slash. The example `https://agent01.external.example.com:8383/auth0/logs` is what you get with default settings and `https://{AGENT_ADDRESS}:{LISTEN_PORT}{WEBHOOK_PATH}`.

efd6: I think the instructions should explicitly note that the path is expected to have a leading slash. Yes, it's in the example, but a single character in an example is very easy for a user to ignore.

efd6: @alaudazzi On the basis of what Chris has written here, I don't think the slash in the templatish is the right way to achevi this documentation and doing it there will lead to confusion; we will see people with `https://host.name.tld:port//path/to/endpoint` which may or may not work. What we should have is some text following the templatish with a statement explaining that the WEBHOOK_PATH value must have a leading slash.

alaudazzi (approved):
LGTM

packages/auth0/_dev/build/docs/README.md
chrisberkhout: To be extra explicit, what if I add "It must include a leading slash" to the form they will be looking at as they follow the instructions?

leading-slash

efd6: SGTM

chrisberkhout: Okay. That's done.

efd6:
Needs an `elastic-package build`.

efd6 (approved):

PR #14456, opened 2025-07-08
System auth log fixes

Tacklebox:
## Proposed commit message

This adds a grok pattern and a few replacements to handle previously unmatched fields in some auth log messages.

## Checklist

### DIFF
--- a/packages/system/data_stream/auth/_dev/test/pipeline/test-auth.log-expected.json
+++ b/packages/system/data_stream/auth/_dev/test/pipeline/test-auth.log-expected.json
@@ -569,6 +569,53 @@
             "user": {
                 "name": "test"
             }
+        },
+        {
+            "ecs": {
+                "version": "8.11.0"
+            },
+            "event": {
+                "category": [
+                    "authentication"
+                ],
+                "outcome": "failure",
+                "kind": "event",
+                "timezone": "+0000"
+            },
+            "host": {
+                "hostname": "host.name"
+            },
+            "input": {
+                "type": "log"
+            },
+            "process": {
+                "name": "sudo",
+                "pid": 654321
+            },
+            "related": {
+                "hosts": [
+                    "host.name"
+                ],
+                "user": [
+                    "alice",
+                    "root"
+                ]
+            },
+            "source": {
+                "address": "",
+                "domain": ""
+            },
+            "system": {
+                "auth": {}
+            },
+            "user": {
+                "effective": {
+                    "id": "0",
+                    "name": "root"
+                },
+                "name": "alice",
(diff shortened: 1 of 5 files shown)


packages/system/data_stream/auth/elasticsearch/ingest_pipeline/log.yml
```diff
@@ -34,11 +34,69 @@ processors:
...
+      target_field: user.id
+      ignore_missing: true
+      ignore_failure: true
+  - rename:
+      tag: rename_rhost
+      field: pam.data.rhost
+      target_field: source.address
+      ignore_missing: true
+      ignore_failure: true
+  - rename:
```
nicholasberlin: this `rename` looks like a repeat of line 82's .. intentional? if so, why?

Tacklebox: Unintentional, thanks!

packages/system/data_stream/auth/_dev/test/pipeline/test-secure-rhel7.log-expected.json
```diff
@@ -218,16 +218,51 @@
             "input": {
                 "type": "log"
             },
-            "message": "pam_unix(sshd:auth): authentication failure; logname= uid=0 euid=0 tty=ssh ruser= rhost=89.160.20.156  user=root",
```
andrewkroh, line 221: Why is the `message` going away? The `message` should remain IMO to allow for searches on the string (it's indexed as a `text`-family data type).

It looks like this was already the behavior based on the some other samples, but I think deleting `message` is the wrong thing to do.

packages/system/data_stream/auth/_dev/test/pipeline/test-auth.log-expected.json
```diff
@@ -569,6 +569,53 @@
...
+                    "host.name"
+                ],
+                "user": [
+                    "alice",
+                    "root"
+                ]
+            },
+            "source": {
+                "address": "",
+                "domain": ""
```
andrewkroh, line 606: Let's remove these if when the values are empty.

packages/system/data_stream/auth/_dev/test/pipeline/test-secure-rhel7.log-expected.json
andrewkroh: I have opened a fix to address this and consolidate the processing logic for journald and logfile inputs. That will make incorporating these PAM parsing enhancements easier for you.

https://github.com/elastic/integrations/pull/16542

PR #14464, opened 2025-07-09
[cisco_ios] Add Dissect and Grok patterns for IPACCESSLOGP logs with Event Code Trimming

mohitjha-elastic:
## Proposed commit message
```
cisco_ios: add dissect and grok patterns for IPACCESSLOGP logs with event code trimming.

Added dissect and grok patterns to support multiple log formats for IPACCESSLOGP logs.
Trimmed whitespace around the event code field to ensure accurate parsing and normalization of log entries.
These changes enhance compatibility with varied log formats and improve field extraction reliability.
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/cisco_ios directory.

> elastic-package test -v

## Related Issue

- Enhancement issue 22830


packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -226,11 +227,22 @@ processors:
...
+      ignore_missing: true
   - dissect:
       field: message
       tag: dissect_gp
       pattern: "list %{cisco.ios.access_list} %{_temp_.event.action} %{network.transport} %{source.address}(%{source.port}) %{} %{destination.address}(%{destination.port}), %{source.packets} packet"
       if: "['IPACCESSLOGP', 'ACCESSLOGP', 'IPV6ACCESSLOGP'].contains(ctx.event?.code)"
+      ignore_failure: true
+  - dissect:
+      field: message
+      tag: dissect_gp
```
jrmolin: i don't think this should be the same tag as the previous dissect.

this is definitely easier to read than converting it to a `grok`, and is likely faster.

packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -68,6 +68,7 @@ processors:
       tag: grok_message
       patterns:
         - '^%%{GREEDYDATA:message}$'
+        - '^%{GREEDYDATA}%%{GREEDYDATA:message}$'
```
jrmolin: what is this supposed to do? yaml supports comments, so please put a comment on this, since regular expressions are hard to understand even when they are straightforward

jrmolin (commented):
questions

packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
mohitjha-elastic: Thanks for pointing that out. I have updated the tag.

packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
mohitjha-elastic: This is actually extracting part after `%` in message field so that it could be further parsed by the next grok processor and the relevant information could be extracted (with `tag: grok_message`).

packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -65,9 +65,10 @@ processors:
         CISCO_TZ: '[a-zA-Z]{1,7}([+-]\d{1,2}|[+-]\d{2}:\d{2})?'
   - grok:
       field: _temp_.message
-      tag: grok_message
+      tag: grok_message_at_top_level
```
jrmolin, line 68: i wish there were an easier way to do this in painless, but i know that there just isn't. i guess this is the best we can hope for. you could also add a step before this to split on the `%` and then return the last part. maybe that would handle all the cases?

packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -226,11 +227,22 @@ processors:
...
   - dissect:
       field: message
       tag: dissect_gp
       pattern: "list %{cisco.ios.access_list} %{_temp_.event.action} %{network.transport} %{source.address}(%{source.port}) %{} %{destination.address}(%{destination.port}), %{source.packets} packet"
       if: "['IPACCESSLOGP', 'ACCESSLOGP', 'IPV6ACCESSLOGP'].contains(ctx.event?.code)"
+      ignore_failure: true
+  - dissect:
+      field: message
+      tag: dissect_gp_access_list
+      pattern: "access-list %{cisco.ios.access_list} %{_temp_.event.action} %{network.transport} %{source.address}(%{source.port}) %{} %{destination.address}(%{destination.port}), %{source.packets} packet"
```
jrmolin, line 243: have you ever seen multiple packets? so it would be `%{source.packets} packets`?

mohitjha-elastic: Yes, there could be multiple packets. The provided dissect pattern will still match the string correctly in such cases and function as expected.
There are several other test cases related to multiple packets are already present in the test folder for other kind of events.

packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
mohitjha-elastic: Yes, I think it's possible to split the string on `%` before applying `Grok`. However, that would require introducing an additional processor, which may not be ideal in this case. Given the trade-off, I think sticking with the current Grok pattern is a more efficient and manageable approach.

mohitjha-elastic:
@jrmolin Let me know if everything looks good to you so we can proceed with merging the PR.

jrmolin:
looks fine

PR #14493, opened 2025-07-10
[zscaler_zpa] Cleanup empty field from json to avoid errors

SimonKoetting:
Zscaler uses a template language to send logs. As a result, it's possible for a
field in the JSON we receive to exist but have an empty string as its value.

For date processors in the pipeline, this issue has already been addressed
using conditions like
`ctx.json?.LogTimestamp != null && ctx.json.LogTimestamp != ''`.

However, we encountered a problem with a user where a non-date field failed
to convert to an IP address due to the error message:
`'' is not an IP string literal`.

To prevent this from happening with any field, I propose adding a cleanup step
at the beginning of the pipeline in this PR. This would handle all empty fields
directly, allowing us to rely on the `ignore_missing` option instead of adding
multiple conditions like `[field] != null && [field] != ''` for each processor.


packages/zscaler_zpa/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.23.2"
+  changes:
+    - description: Cleanup empty field from json to avoid errors.
```
andrewkroh:
```suggestion
    - description: Remove fields with empty values from the incoming JSON document to avoid errors.
```

packages/zscaler_zpa/data_stream/app_connector_status/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -13,6 +13,23 @@ processors:
       field: event.original
       target_field: json
       ignore_failure: true
+  - script:
```
andrewkroh, line 16: Add a `tag`.

https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#script-processors-should-have-a-tag-and-description

packages/zscaler_zpa/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -13,11 +13,28 @@ processors:
       field: event.original
       target_field: json
       ignore_failure: true
+  - script:
```
andrewkroh, line 16: Add a `tag`.

https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#script-processors-should-have-a-tag-and-description

packages/zscaler_zpa/data_stream/browser_access/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -12,13 +12,30 @@ processors:
   - json:
       field: event.original
       target_field: json
+  - script:
```
andrewkroh, line 15: Add a `tag`.

https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#script-processors-should-have-a-tag-and-description

packages/zscaler_zpa/data_stream/user_activity/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -12,13 +12,30 @@ processors:
   - json:
       field: event.original
       target_field: json
+  - script:
```
andrewkroh, line 15: Add a `tag`.

https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#script-processors-should-have-a-tag-and-description

packages/zscaler_zpa/data_stream/user_status/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -13,13 +13,30 @@ processors:
       field: event.original
       target_field: json
       ignore_failure: true
+  - script:
```
andrewkroh, line 16: Add a `tag`.

https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#script-processors-should-have-a-tag-and-description

packages/zscaler_zpa/data_stream/app_connector_status/elasticsearch/ingest_pipeline/default.yml
andrewkroh: Similar code already exists in the pipeline (~L305). Can the later one be removed now?

SimonKoetting: added the tags, also to the second script processor that was already there to be consistent.
Yes the code is basically identical just the first one i added cleans up the json before the parsing starts to avoid any errors in the pipeline.
The one at the end of the pipeline cleans up the whole event after the parsing to make sure that no empty field is getting stored in the index.

So we should keep that script

efd6 (commented):
I'm inclined to think that adding the guards on the affected processors would be a better approach rather than traversing the full document an extra time.

SimonKoetting:
Hey @efd6 ,
for performance reasons?

I would be fine with that as well, my thought was just that it will be a bit easier to maintain in the future if there maybe are new fields added or similar.

efd6:
> for performance reasons?

Yes.

PR #14504, opened 2025-07-11
[Cisco Nexus] Improve timezone and timestamp handling

P1llus:
## Proposed commit message

The PR introduces tz_map as an option in the configuration of the integration, similar to tz_map this is specifically for usecases where the vendor provides a timezone format in their logs which is not supported by Java, and is the only way for a user to map these unsupported formats to a proper IANA format on ingest time.

While implementing this there was also several unecessary additions between syslog and vendor timestamps, so I changed that up a bit and updated the system and pipeline tests. Older documents now get missing ECS fields and the dates are parsed correctly also on older logs (which is why older pipeline test output have changes in them).

## Checklist


packages/cisco_nexus/data_stream/log/manifest.yml
```diff
@@ -23,6 +23,19 @@ streams:
...
         show_user: true
         default: 9506
+      - name: tz_map
+        type: yaml
+        title: Timezone Map
+        multi: false
+        required: false
+        show_user: false
+        description: >-
+          A collectiom of timezones found in Cisco Nexus logs (as defined in each `tz_short`), and the replacement value (as defined in each `tz_long`) which should be the full proper IANA Timezone format. This is used to override vendor provided timezone formats that is not supported by Elasticsearch [Date Processors](https://www.elastic.co/docs/reference/enrich-processor/date-processor#date-processor-timezones)
```
taylor-swanson: Could we also include offsets as well?

This is a snippet of the description from `cisco_asa`:

> proper IANA time zone or offset (for example, Australia/Sydney or +10:00).

packages/cisco_nexus/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.4.1"
+  changes:
+    - description: Fix bug that did not recognize timestamps to use tz_map override.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/14458
```
dwhyrock: PR link needs to be updated

packages/cisco_nexus/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -3,126 +3,101 @@ description: Pipeline for processing Cisco Nexus logs.
 processors:
   - set:
       field: ecs.version
-      value: '8.17.0'
-      tag: 'set_ecs_version'
+      value: "8.17.0"
```
dwhyrock, line 6: Curious why you're updating the single quotes to double quotes?  I do prefer the doubles, but curious if there's some yaml-specific reason to do this.

packages/cisco_nexus/data_stream/log/sample_event.json
```diff
@@ -1,72 +1,55 @@
...
+        "id": "13671cfa-49ce-4139-8d65-90166401d5f5",
+        "name": "elastic-agent-83602",
         "type": "filebeat",
-        "version": "8.10.2"
+        "version": "9.0.3"
     },
     "cisco_nexus": {
         "log": {
-            "description": "EARL 0  NF ASIC: Uncorrectable Parity error in Netflow Table.",
-            "facility": "EARL",
```
dwhyrock, line 13: Is there a reason why all of these fields are being deleted?  Is this part of the processing changes in `default.yml`, or something else?

packages/cisco_nexus/data_stream/log/_dev/test/pipeline/test-nexus.log-expected.json
```diff
@@ -775,16 +790,16 @@
...
                 "log": {
                     "description": "last message repeated 5 time",
                     "ip_address": "10.100.0.34",
                     "priority_number": 187,
                     "sequence_number": 6031594,
                     "syslog_time": "2025-05-03T13:20:50.000Z",
-                    "time": "2025-05-03T03:20:48.739Z",
-                    "timezone": "AEST"
+                    "time": "2025-05-03T13:20:48.739+10:00",
+                    "timezone": "Australia/Sydney"
```
taylor-swanson: syslog_time is UTC but has the same time (ignoring the zone) as the `time` and `@timestamp`. This is incorrect.

The syslog_time can't be interpreted as UTC, since that timestamp is zone-less. Absent of zone info, either `tz_offset` needs to be used, or in this case, the zone provided by the more precise timestamp.

As of right now, `syslog_time` and `@timestamp` are 10 hours off, when they are only 2 seconds off from each other.

What we could do is set a "global" time zone (`event.timezone`) after we've done the necessary processing the of time zone and use that for all subsequent date processing.

packages/cisco_nexus/data_stream/log/_dev/test/pipeline/test-nexus.log-expected.json
```diff
@@ -3843,9 +3897,10 @@
                     "priority_number": 187,
                     "severity": 3,
                     "switch_name": "ac45ce-sr1",
+                    "syslog_time": "2025-06-14T11:34:35.000Z",
                     "threshold_value": "-1.0 dBm",
-                    "time": "2025-06-14T01:34:35.000Z",
-                    "timezone": "AEST",
+                    "time": "2025-06-14T11:34:35.000Z",
```
taylor-swanson: Incorrect time zone

dwhyrock:
@P1llus do you have a timeline as to when you'll address @taylor-swanson 's comments?  It'd be nice to get this resolved to help the SDH customer.

packages/cisco_nexus/data_stream/log/manifest.yml
dwhyrock: I added that in as well.  I tested it, and the timestamps were the same (+10:00 offset), and only the `timezone` string changed from `Australia/Sydney` to `+10:00`, so it does work.

dwhyrock:
@taylor-swanson I believe I addressed all of your comments in this PR since Marius is on PTO right now.  Can you have another look, please?

packages/cisco_nexus/data_stream/log/manifest.yml
```diff
@@ -23,6 +23,19 @@ streams:
...
         show_user: true
         default: 9506
+      - name: tz_map
+        type: yaml
+        title: Timezone Map
+        multi: false
+        required: false
+        show_user: true
+        description: >-
+          A collectiom of timezones found in Cisco Nexus logs (as defined in each `tz_short`), and the replacement value (as defined in each `tz_long`) which should be the full proper IANA Timezone format (for example, Australia/Sydney or +10:00). This is used to override vendor provided timezone formats that is not supported by Elasticsearch [Date Processors](https://www.elastic.co/docs/reference/enrich-processor/date-processor#date-processor-timezones)
```
taylor-swanson, line 33: This looks good, just needs to be applied to the other two inputs.

dwhyrock: Updated

PR #14529, opened 2025-07-14
[cloudflare_logpush] Fix audit data stream null handling

chrisberkhout:
## Proposed commit message

```
[cloudflare_logpush] Fix audit data stream null handling

Fix misuse of the Elvis operator that was causing the error:

    Cannot invoke "Object.getClass()" because "value" is null

Also add a tag to every processor in the audit data stream and include include
it in the on_failure message. Fix some indentation.
```

## Checklist


efd6 (approved):
Thanks

PR #14537, opened 2025-07-14
amazon_security_lake: accept UNIX_MS in *time_dt fields

efd6:
## Proposed commit message
```
amazon_security_lake: accept UNIX_MS in *time_dt fields

AWS send documents with time_dt fields populated with Unix millis. This
does not appear to be specified in the OCSF spec which says that time_dt
fields are in timestamp_t[1], which is RFC3339[2].

[1]https://github.com/ocsf/examples/tree/c56f79d2bab9095f1d9d7b986997bb8468054bc5/encodings/json
[2]https://github.com/ocsf/ocsf-schema/blob/92b69f508b883ff7175003923e0c462f2fbed93c/dictionary.json#L6343-L6349
```

> [!NOTE]
> As with the previous version of this PR, I have misgivings. I think that by doing this, we are incorrectly following someone who is breaking the spec. However, it may be that it is something that we need to do because of the mass of the at other person.
>
> Postel was wrong.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


efd6:
Thanks for picking that up. Amended.

PR #14550, opened 2025-07-15
[FireEye] Add missing ECS fields

### LINKED ISSUE #13756: [FireEye Network Security]: ECS Fields Missing
### Integration Name

FireEye Network Security [fireeye]

### Dataset Name

fireeye.nx

### Integration Version

1.25.0

### Agent Version

8.17.5

### Agent Output Type

logstash

### Elasticsearch Version

8.17.1

Linux

### Software/API Version

_No response_

### Error Message

None

### Event Original

_No response_

### What did you do?

Switched FireEye NX from manual logstash pipelines to Elastic Agent Integration 

### What did you see?

Missing ECS Fields. 

ECS Fields

### Anything else?

From raw log > ECS

meta_cbname > observer.hostname

meta_sip4 > observer.ip

event_type > event.type/event.category? 
   I run this as event.type but it could be a few different event fields.  This contains information about the flow like... 
flow, smb, tls, fileinfo.  Doesn't fit perfectly for the event.type/event.category with the restrictions you place on it.  But, it also doesn't really fit the network fields. 

fireeye.nx.fileinfo.md5 > file.hash

fireeye.nx.fileinfo.filename > file.name

tls.client.server_name > destination.hostname

I'm sure there are others, it's hard to parse through when the event fields are missing.

moxarth-rathod:
## Proposed commit message

```
fireeye: add missing ECS fields

This PR has added the missing ECS fields like event, file, IP related fields, etc.
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/fireeye directory.

> elastic-package test -v

## Related issues

- Closes #13756

### DIFF
--- a/packages/fireeye/changelog.yml
+++ b/packages/fireeye/changelog.yml
@@ -1,6 +1,6 @@
 - version: "1.27.0"
   changes:
-    - description: Add support for `event.outcome`, `event.kind`, `destination.domain`, `file.hash.md5`, `file.path`, `observer.hostname` and `observer.ip` ECS fields.
+    - description: Add support for `event.kind`, `destination.domain`, `file.hash.md5`, `file.path`, `observer.hostname` and `observer.ip` ECS fields.
       type: enhancement
       link: https://github.com/elastic/integrations/pull/14550
 - version: "1.26.0"
--- a/packages/fireeye/changelog.yml
+++ b/packages/fireeye/changelog.yml
@@ -1,3 +1,8 @@
+- version: "1.27.0"
+  changes:
+    - description: Add support for `event.outcome`, `event.kind`, `destination.domain`, `file.hash.md5`, `file.path`, `observer.hostname` and `observer.ip` ECS fields.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
 - version: "1.26.0"
   changes:
     - description: Add Overview Dashboard.
--- a/packages/fireeye/changelog.yml
+++ b/packages/fireeye/changelog.yml
@@ -2,7 +2,7 @@
   changes:
     - description: Add support for `event.outcome`, `event.kind`, `destination.domain`, `file.hash.md5`, `file.path`, `observer.hostname` and `observer.ip` ECS fields.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/1
+      link: https://github.com/elastic/integrations/pull/14550
 - version: "1.26.0"
   changes:
     - description: Add Overview Dashboard.
--- a/packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
+++ b/packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
@@ -12,7 +12,10 @@ processors:
   - rename:
       field: rawmsg.app_proto
       target_field: network.protocol
-      if: ctx.rawmsg.app_proto != null && !['failed','succeded'].contains(ctx.rawmsg.app_proto.toLowerCase())
+      if: >-
+        ctx.rawmsg?.app_proto != null
+        && ctx.rawmsg.app_proto.toLowerCase() != 'failed'
+        && ctx.rawmsg.app_proto.toLowerCase() != 'succeeded'
       ignore_missing: true
   - set:
       field: event.outcome
@@ -21,7 +24,7 @@ processors:
   - set:
       field: event.outcome
       value: success
-      if: ctx.rawmsg?.app_proto != null && ['succeded'].contains(ctx.rawmsg.app_proto.toLowerCase())
+      if: ctx.rawmsg?.app_proto?.equalsIgnoreCase('succeeded') == true
   - set:
       field: event.outcome
       value: unknown
--- a/packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
+++ b/packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
@@ -12,23 +12,7 @@ processors:
   - rename:
       field: rawmsg.app_proto
       target_field: network.protocol
-      if: >-
-        ctx.rawmsg?.app_proto != null
-        && ctx.rawmsg.app_proto.equalsIgnoreCase('failed') == false
-        && ctx.rawmsg.app_proto.equalsIgnoreCase('succeeded') == false
       ignore_missing: true
-  - set:
-      field: event.outcome
-      value: failure
-      if: ctx.rawmsg?.app_proto?.equalsIgnoreCase('failed') == true
-  - set:
-      field: event.outcome
-      value: success
-      if: ctx.rawmsg?.app_proto?.equalsIgnoreCase('succeeded') == true
-  - set:
-      field: event.outcome
-      value: unknown
-      if: ctx.event?.outcome == null
   - rename:
       field: rawmsg.flow_id
       target_field: fireeye.nx.flow_id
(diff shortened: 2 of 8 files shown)


packages/fireeye/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: "1.27.0"
+  changes:
+    - description: Add support for `event.outcome`, `event.kind`, `destination.domain`, `file.hash.md5`, `file.path`, `observer.hostname` and `observer.ip` ECS fields.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
```
andrewkroh:
```suggestion
      link: https://github.com/elastic/integrations/pull/14550
```

packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
```diff
@@ -12,7 +12,20 @@ processors:
...
+      if: ctx.rawmsg.app_proto != null && !['failed','succeded'].contains(ctx.rawmsg.app_proto.toLowerCase())
       ignore_missing: true
+  - set:
+      field: event.outcome
+      value: failure
+      if: ctx.rawmsg?.app_proto != null && ['failed'].contains(ctx.rawmsg.app_proto.toLowerCase())
+  - set:
+      field: event.outcome
+      value: success
+      if: ctx.rawmsg?.app_proto != null && ['succeded'].contains(ctx.rawmsg.app_proto.toLowerCase())
```
andrewkroh: Is `succeded` correct? It's not spelled correctly. Do we have a sample showing an example `app_proto` value that contains the misspelling? It looks like the samples either have `failed` or the name of a protocol.

packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
```diff
@@ -12,7 +12,20 @@ processors:
   - rename:
       field: rawmsg.app_proto
       target_field: network.protocol
+      if: ctx.rawmsg.app_proto != null && !['failed','succeded'].contains(ctx.rawmsg.app_proto.toLowerCase())
       ignore_missing: true
+  - set:
+      field: event.outcome
+      value: failure
+      if: ctx.rawmsg?.app_proto != null && ['failed'].contains(ctx.rawmsg.app_proto.toLowerCase())
```
andrewkroh: Why did you use Collection.contains() to implement this? It seems like a string comparison using `equalsIgnoreCase` would be more clear.

```suggestion
      if: ctx.rawmsg?.app_proto?.equalsIgnoreCase('failed') == true
```

packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
```diff
@@ -12,7 +12,20 @@ processors:
   - rename:
       field: rawmsg.app_proto
       target_field: network.protocol
+      if: ctx.rawmsg.app_proto != null && !['failed','succeded'].contains(ctx.rawmsg.app_proto.toLowerCase())
```
andrewkroh: I think this can be written out as "if foo equals failed || if foo equals succeeded" to avoid allocating a new list in each expression eval.

packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
andrewkroh: I see you that you have made changes. But I'm still lacking evidence that `succeeded` (correctly spelled) is a valid `app_proto` value? Where did this value originate? Is it from docs, a sample, an assumption?

packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
```diff
@@ -12,16 +12,19 @@ processors:
   - rename:
       field: rawmsg.app_proto
       target_field: network.protocol
-      if: ctx.rawmsg.app_proto != null && !['failed','succeded'].contains(ctx.rawmsg.app_proto.toLowerCase())
+      if: >-
+        ctx.rawmsg?.app_proto != null
+        && ctx.rawmsg.app_proto.toLowerCase() != 'failed'
+        && ctx.rawmsg.app_proto.toLowerCase() != 'succeeded'
```
andrewkroh: These can use the same equalsIgnoreCase() function as below.

packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
moxarth-rathod: @andrewkroh we've got the `failed` value from the test data so this was an assumption. And i think that `succeeded` is grammatically correct, do you think there should be any other value here?

andrewkroh: > so this was an assumption

I don't think we should be making the assumption that there is a `succeeded` value. 

If this behaves anything like Suricata, then `failed` would mean that it couldn't detect the layer 7 protocol. That's what it looks like this is doing. And if that's true, then I don't think this equates to an [event.outcome](https://www.elastic.co/docs/reference/ecs/ecs-event#field-event-outcome) disposition, so we should not be determining the outcome on the basis of the L7 protocol decoding success/failure.

moxarth-rathod: Got it @andrewkroh, I've now removed this from the pipeline.

packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
```diff
@@ -58,9 +66,15 @@ processors:
       target_field: fireeye.nx.deviceid
       ignore_missing: true
```
kcreddy, line 67: Use `json.` for `meta_oml` and `deviceid` as well.

packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
```diff
@@ -58,9 +66,15 @@ processors:
...
   - rename:
-      field: meta_cbname
-      target_field: fireeye.nx.hostname
+      field: json.meta_cbname
+      target_field: observer.hostname
       ignore_missing: true
```
kcreddy: Without removing custom field `fireeye.nx.hostname`, can you set `observer.hostname` after?
I think this could cause a breaking-change for users if we remove custom fields.

packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml
```diff
@@ -158,11 +182,21 @@ processors:
...
   - rename:
       field: rawmsg.fileinfo.size
       target_field: fireeye.nx.fileinfo.size
       if: ctx?.event?.type == 'fileinfo'
       ignore_missing: true
+  - set:
+      field: file.size
+      tag: set_file_size_from_fileinfo_size
+      copy_from: fireeye.nx.fileinfo.size
+      ignore_empty_value: true
```
kcreddy, line 203: As per the issue, user also expects `fireeye.nx.fileinfo.filename > file.name`.

moxarth-rathod:
> For issue suggestion: `tls.client.server_name > destination.hostname`, the correct mapping should be into [`destination.domain`](https://www.elastic.co/docs/reference/ecs/ecs-tls#field-tls-client-server-name)

Yes, i've mapped `tls.client.server_name` to `destination.domain` here -
https://github.com/elastic/integrations/blob/595c3c9e386f4e4191b56743e6aebebaeb7869f4/packages/fireeye/data_stream/nx/elasticsearch/ingest_pipeline/renaming-raws.yml#L485-L489

kcreddy (approved):

Issue #5799, opened 2023-04-05
[Kubernetes Audit Logs] [AWS Cloudwatch] Create a new datastream for Ingesting EKS Audit Logs


imays11:
Problem
----
I used the Kubernetes Audit Logs Integration to create a ruleset for detecting threat behaviors in Kubernetes using the API Server Audit Logs. As of right now that Integration can only access audit logs from local K8s environments who have access to the control plane. In order to get audit logs from Cloud managed K8s providers (AWS EKS, GCP GKE, Azure AKS), you have to utilize one of our cloud specific Integrations to gather those logs from the respective cloud platform's logging solution.  In order to be searchable, those audit log events need to be further parsed because they come in as a large json blob in a single `message` field. In order to be aligned with our current Kubernetes ruleset, those fields need to be appended with `kubernetes.audit` to match the mapping from our Kubernetes Audit Logs Integration, which is what the ruleset was based off of. 

I worked with @andrewkroh a while back to create this solution. It's specifically designed to ingest K8s audit logs from EKS via AWS Cloudwatch Integration and align the field mapping to the current Kubernetes Audit Logs Integration. The goal being to eventually do something similar to this across all the Cloud managed K8s platforms so that we can use 1 single ruleset for all K8s Audit Logs.

Request
----
Below is that documented solution for AWS, while simple in design, it's not very user friendly. My request is for a new datastream that would essentially do this for the user. This could be housed within the AWS Integration or the Kubernetes Integration. I'm not sure which is a better long-term solution as the goal is to do this for all the major cloud platforms K8s offerings. So we'd either need to create a datastream like this in each of those cloud platform Integrations, or create something like this inside the Kubernetes Integration itself that could allow for distinction between each cloud provider. With the upcoming launch of Defend for Containers (cc: @learhy), having a good solution to showcase K8s audit log ingestion via our current Kubernetes Integration and pre-built K8s audit log detection rules, for at least an EKS cluster, would be great timing since these are all important aspects of cloud workload protection. 

Solution
----
0. Within the AWS Integration, setup Cloudwatch logs datastream to collect  Log Group ARN containing EKS Audit Logs.

1. Place this processor, pulled from the ```kubernetes.audit_logs``` Integration, directly in the Cloudwatch Integration Policy.
```
- decode_json_fields:
    fields: [message]
    target: kubernetes.audit
    
# Copied from the kubernetes.audit_logs integration.
[...]

PR #14554, opened 2025-07-15
Kubernetes.audit_logs: add support for cloud providers

chemamartinez:
## Proposed commit message

Extend the Kubernetes audit_logs data stream to support collecting audit logs from managed Kubernetes clusters in major cloud providers:
- AWS EKS via CloudWatch Logs
- Azure AKS via Event Hub
- Google GKE via Pub/Sub

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/5799

## Screenshots

Add-integration-Kubernetes-Integrations-Elastic-07-15-2025_06_49_PM

Edit-integration-Elastic-Agent-elastic-package-Agent-policies-Fleet-Elastic-07-15-2025_06_52_PM

Screenshot 2025-07-15 at 18 52 48

Screenshot 2025-07-15 at 18 53 07


packages/kubernetes/_dev/build/docs/audit-logs.md
```diff
@@ -1,8 +1,16 @@
 # audit-logs
 
-audit-logs integration collects and parses Kubernetes audit logs.
+Audit logs integration collects and parses Kubernetes audit logs.
```
gizas:
```suggestion
Audit-logs integration collects and parses Kubernetes audit logs.
```

packages/kubernetes/data_stream/audit_logs/agent/stream/aws-cloudwatch.yml.hbs
```diff
@@ -0,0 +1,103 @@
+{{#unless log_group_name}}
```
gizas, line 1: Wondering if those files should be placed under relevant aws-cloudwatch integration (and similarly under azure and gcp). I am thinking that it would be difficult for our users to figure out that we have audit logs support and to check on k8s when initially are on a CSP integration. 

@zmoog wdyt on this?

packages/kubernetes/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -61,10 +104,56 @@ processors:
...
       field: event.kind
       value: pipeline_error
+  - append:
+      field: tags
+      value: preserve_original_event
+      allow_duplicates: false
   - append:
       field: error.message
-      value: '{{{ _ingest.on_failure_message }}}'
+      value: >
```
gizas: Use this example: https://github.com/elastic/integrations/blob/main/packages/aws/data_stream/cloudwatch_logs/elasticsearch/ingest_pipeline/default.yml#L25

packages/kubernetes/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1,6 +1,19 @@
 ---
 description: Pipeline for processing Kubernetes audit logs.
 processors:
+  - rename:
+      field: message
```
gizas, line 5: I think we should do it with set: https://github.com/elastic/integrations/blob/main/packages/aws/data_stream/cloudwatch_logs/elasticsearch/ingest_pipeline/default.yml#L9

In order not to loose message field right?

packages/kubernetes/data_stream/audit_logs/manifest.yml
```diff
@@ -6,10 +6,322 @@ elasticsearch:
     mappings:
       dynamic: false
 streams:
+  - input: aws-cloudwatch
```
gizas, line 9: Similar as above comment https://github.com/elastic/integrations/pull/14554/files#r2218628097

packages/kubernetes/docs/audit-logs.md
```diff
@@ -1,8 +1,16 @@
 # audit-logs
 
-audit-logs integration collects and parses Kubernetes audit logs.
+Audit logs integration collects and parses Kubernetes audit logs.
```
gizas:
```suggestion
Audit-logs integration collects and parses Kubernetes audit logs.
```

packages/kubernetes/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
chemamartinez: What would be the default criteria for this in Obs integrations? We usually keep the original raw message in `event.original` and `message` gets removed so it is not duplicated.

packages/kubernetes/data_stream/audit_logs/agent/stream/aws-cloudwatch.yml.hbs
chemamartinez: After some discussions (see https://github.com/elastic/integrations/issues/5799#issuecomment-2939856915 and further comments), that option was on the table but it shows some inconveniences in terms of maintenance mainly).

Happy to hear your thoughts on this point.

terrancedejesus: @chemamartinez - This is amazing and thank you for putting in the PR.

@gizas - Definitely a great question. My two cents:

For AKS at least, most Azure data sources are tied to the single Azure integration. Entra ID, Activity, Diagnostic, Platform, etc. logs are all in the Azure integration as a separate data stream, rather than having their own integrations. However, Azure may be unique in this perspective since all logs, regardless of service, are forwarded to an Event Hub, thus putting them in a centralized location for us to ingest from. From a detection rule perspective, if we wanted to correlate activity between - for instance - Entra ID sign-ins and K8s, two separate integration installations are required instead of one. However, for AWS, most rules are written on CloudTrail audit logs, so we still would require a separate integration, CloudWatch, for this.

On the flip side - K8s as a separate integration may make more sense to isolate the data streams (both local and CSP-based) as being done here and point users with K8s requirements to a single integration and pick which provider. We know that K8s is a very popular integration as-is so it may be good to roll this out with what is already adopted heavily.

imays11:
This looks good to me, thank you for the changes. I agree with Terrance it's probably best to modify the K8s intregration which is already heavily used rather than modify each individual CSP integration. This will also help us maintain a single K8s ruleset, and will make K8s audit log ingest more user friendly for everyone.

packages/kubernetes/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -61,10 +104,58 @@ processors:
...
+        append:
+          tag: append_related_source_ip
+          field: related.ip
+          value: '{{{_ingest._value}}}'
+          allow_duplicates: false
+
+  - script:
+      lang: painless
+      description: Drops null/empty values recursively.
+      tag: drop_null_empty_values
```
kcreddy: Can you use this instead? https://github.com/elastic/integrations/blob/main/packages/mimecast/data_stream/archive_search_logs/elasticsearch/ingest_pipeline/default.yml#L130-L156

There were marginal gains using this.

packages/kubernetes/manifest.yml
```diff
@@ -246,8 +246,17 @@ policy_templates:
...
-        title: Collect Kubernetes audit logs
-        description: Collect Kubernetes audit logs
+        title: Collect Kubernetes audit logs from local files
+        description: Collect Kubernetes audit logs from local nodes with Elastic Agent.
+      - type: aws-cloudwatch
+        title: Collect Kubernetes audit logs from AWS CloudWatch
+        description: Collect Kubernetes audit logs from AWS EKS with Elastic Agent.
+      - type: azure-eventhub
+        title: Collect Kubernetes audit logs from Azure Event Hub
+        description: Collect Kubernetes audit logs from with Azure AKS Elastic Agent.
```
kcreddy:
```suggestion
        description: Collect Kubernetes audit logs from Azure AKS with Elastic Agent.
```

packages/kubernetes/manifest.yml
```diff
@@ -246,8 +246,17 @@ policy_templates:
...
+      - type: gcp-pubsub
+        title: Collect Kubernetes audit logs logs from Google Cloud Pub/Sub
```
kcreddy:
```suggestion
        title: Collect Kubernetes audit logs from Google Cloud Pub/Sub
```

packages/kubernetes/data_stream/audit_logs/fields/ecs.yml
```diff
@@ -32,3 +32,7 @@
   name: event.action
 - external: ecs
   name: event.outcome
+- external: ecs
+  name: related.ip
+- external: ecs
+  name: related.user
```
kcreddy, line 38: Could possibly add `orchestrator` fields, but its beyond scope of this PR.

packages/kubernetes/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -61,10 +104,58 @@ processors:
       field: client.ip
       copy_from: source.ip
       ignore_empty_value: true
+
+  - append:
```
kcreddy, line 170: Do you think its worth adding `cloud.*` fields depending on `input.type`?

packages/kubernetes/data_stream/audit_logs/fields/ecs.yml
chemamartinez: I think is a good time to add them, as some related fields are added here.

packages/kubernetes/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
chemamartinez: Cloud metadata is not present in events, do you mean using the [`add_cloud_metadata`](https://www.elastic.co/docs/reference/beats/filebeat/add-cloud-metadata) processor?

kcreddy: You are right. The metadata is not present in all inputs and only see it in `Cloudwatch` input.
https://github.com/elastic/beats/blob/main/x-pack/filebeat/input/awscloudwatch/processor.go#L62-L65

`Eventhub` defines some, but that cannot be used in `cloud.*` fields. https://github.com/elastic/beats/blob/main/x-pack/filebeat/input/azureeventhub/v2_input.go#L483-L495

packages/kubernetes/data_stream/audit_logs/_dev/test/pipeline/test-audit.log-expected.json
```diff
@@ -125,6 +136,17 @@
...
             },
+            "orchestrator": {
+                "api_version": "audit.k8s.io/v1",
+                "cluster": {
+                    "name": "leases"
+                },
+                "namespace": "kube-system",
+                "resource": {
+                    "name": "elastic-agent-cluster-test"
+                },
```
kcreddy: Just checking if `resource.name` and `cluster.name` are mapped reversely. 
It seems the values here makes sense if the mapping is reversed.

packages/kubernetes/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
kcreddy: We can set `cloud.provider` based on `ctx.input.type`.

```yaml
  - set:
      field: cloud.provider
      value: aws
      if: ctx.input?.type == "aws-cloudwatch"
```

packages/kubernetes/data_stream/audit_logs/_dev/test/pipeline/test-audit.log-expected.json
chemamartinez: I have doubted about the same.

However, I am pretty sure the resource name is correct. Breaking down the request URI: `/apis/coordination.k8s.io/v1/namespaces/kube-system/leases/elastic-agent-cluster-test`

The resource name is always the last part of the URI. I think the confusion comes from a bad naming when setting the names for generating the logs.

Getting deeper into this, I think the problem is that the `cluster.name` is actually the `resource.type`.

From this [article](https://nordicapis.com/best-practices-for-hosting-an-api-on-kubernetes/):

> In Kubernetes, requests are organized using different HTTP requests. All Kubernetes requests begin with one of two prefixes. Core APIs begin with the prefix /api/. Groups of APIs begin with the prefix /apis/.
> When hosting APIs on Kubernetes, it’s highly recommended you follow this convention. Individual APIs should be grouped under /api/, and groups should be filed under /apis/.
> This naming convention is also convenient for standard API formatting like versioning. A path for an API hosted with Kubernetes might look like this:
> /api/v1/\<resource-type-name>/\<resource-name>
> /apis/\<api-group>/\<api-version>/\<resource-type-name>/\<resource-name>

I'll change that.

kcreddy (approved):
LGTM. Thanks Chema.

leandrojmp:
Hello @chemamartinez,

With this change I understand that now I can get the AKS logs using this integration instead of using the Cloudwatch integration and a reroute processor on a custom pipeline? Is that right?

chemamartinez:
@leandrojmp yes, now you should be able to configure this integration to directly collect the EKS Audit logs as the same way you did for the AWS Cloudwatch integration, and data will automatically be processed by the Kubernetes Audit logs pipeline with no need to reroute.

PR #14567, opened 2025-07-16
[o365] Stricter enforcement of maximum age limits

chrisberkhout:
## Proposed commit message

```
[o365] Stricter enforcement of maximum age limits

In order to avoid requesting data more than 7 days old:

- When calculating the start of the allowed time range, use `now()` (a
  function that returns the current time) rather than `now` (which is
  the time the current CEL evaluation began), as the latter will become
  too old if a singe CEL evaluation runs for more than a few minutes.

- Make the `167h55m`  limit used throughout the CEL expression
  configurable, in the advanced settings, so that it can be shortened if
  necessary.

This integration's CEL expression is one that may run for a long time.
For each of a list of content types it will register a subscription,
list data and fetch data, and this has been observed to take many
minutes.
```

## Checklist

### DIFF
--- a/packages/o365/data_stream/audit/manifest.yml
+++ b/packages/o365/data_stream/audit/manifest.yml
@@ -69,7 +69,7 @@ streams:
       - name: initial_interval
         type: text
         title: Initial Interval
-        description: Initial interval for the first API call. Default starts fetching events from 167h55m, i.e., 7 days ago, and can not go further back than that. Supports following suffixes - "h" (hour), "m" (minute), "s" (second), "ms" (millisecond), "us" (microsecond), and "ns" (nanosecond)
+        description: Initial interval for the first API call. Default starts fetching events from 167h55m, i.e., 7 days ago, and must not go further back than that. Supports following suffixes - "h" (hour), "m" (minute), "s" (second), "ms" (millisecond), "us" (microsecond), and "ns" (nanosecond)
         show_user: true
         required: true
         default: 167h55m
--- a/packages/o365/data_stream/audit/manifest.yml
+++ b/packages/o365/data_stream/audit/manifest.yml
@@ -83,7 +83,7 @@ streams:
       - name: maximum_age
         type: text
         title: Maximum Age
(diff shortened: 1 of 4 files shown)


packages/o365/data_stream/audit/manifest.yml
```diff
@@ -69,7 +69,7 @@ streams:
       - name: initial_interval
         type: text
         title: Initial Interval
-        description: Initial interval for the first API call. Default starts fetching events from 167h55m, i.e., 7 days ago. This value should not be more than this and will be cut to 167h55m if it is. Supports following suffixes - "h" (hour), "m" (minute), "s" (second), "ms" (millisecond), "us" (microsecond), and "ns" (nanosecond)
+        description: Initial interval for the first API call. Default starts fetching events from 167h55m, i.e., 7 days ago, and can not go further back than that. Supports following suffixes - "h" (hour), "m" (minute), "s" (second), "ms" (millisecond), "us" (microsecond), and "ns" (nanosecond)
```
efd6:
```suggestion
        description: Initial interval for the first API call. Default starts fetching events from 167h55m, i.e., 7 days ago, and must not go further back than that. Supports following suffixes - "h" (hour), "m" (minute), "s" (second), "ms" (millisecond), "us" (microsecond), and "ns" (nanosecond)
```

packages/o365/data_stream/audit/manifest.yml
```diff
@@ -80,6 +80,13 @@ streams:
         show_user: true
         required: true
         default: 1h
+      - name: maximum_age
+        type: text
+        title: Maximum Age
+        description: A hard maximum age limit for data that can be requested. It defaults to 5 mins less than the API's documented limit but can be shortened as a workaround for errors related to expired data. Supports following suffixes - "h" (hour), "m" (minute), "s" (second), "ms" (millisecond), "us" (microsecond), and "ns" (nanosecond)
```
efd6:
```suggestion
        description: A hard maximum age limit for data that can be requested. It defaults to 5 mins less than the API's documented limit but may be shortened as a workaround for errors related to expired data. Supports following suffixes - "h" (hour), "m" (minute), "s" (second), "ms" (millisecond), "us" (microsecond), and "ns" (nanosecond)
```

PR #14571, opened 2025-07-16
[Enhancement] Update categories for packages

trisch-me:
We are using categories of packages in telemetry and it seems that categories of packages are not reliable as I found a few of security packages not marked as such, for example `aws` etc.

I took advantage of LLM to update the list of categories according to the readme in each integration and using this [list](https://github.com/elastic/package-spec/blob/a50d2bc6a419ac15cfeda61fb1d3b7f7f1b15100/spec/integration/manifest.spec.yml#L21-L96) of available categories

I have verified it as much as I could but as it touches a lot of integrations it would be great to have more eyes on it. 
Every new category has justification

p.s. I am new to the integrations development so it might be I’m missing some of the usages of categories and current changes might break something. If so - I am happy to update the PR 

FYI @vglagoleva @andrewkroh

### DIFF
--- a/packages/osquery_manager/_dev/build/build.yml
+++ b/packages/osquery_manager/_dev/build/build.yml
@@ -1,3 +1,3 @@
 dependencies:
   ecs:
-    reference: git@v8.10.0
+    reference: git@v9.1.0
--- a/packages/osquery_manager/changelog.yml
+++ b/packages/osquery_manager/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.19.0"
+  changes:
+    - description: Update osquery to version 5.18.1 and ECS to 9.10.0
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/15321
 - version: "1.18.0"
   changes:
     - description: Adjust result mapping to match action responses
--- a/packages/osquery_manager/data_stream/result/fields/ecs.yml
+++ b/packages/osquery_manager/data_stream/result/fields/ecs.yml
@@ -1,6 +1,6 @@
 # This file is generated with osqgen (https://github.com/aleksmaus/osqgen) tool from the official ECS fields yml
 # To regenerate use:
-# osqgen --schema "./schema/ecs/fields.ecs_8.10.0.yml" ecs > ecs.yml
+# osqgen --schema "./schema/ecs/fields.ecs_9.1.0.yml" ecs > ecs.yml
 - external: ecs
   name: client.as.number
 - external: ecs
@@ -498,8 +498,6 @@
 - external: ecs
   name: process.parent.pe.sections.virtual_size
 - external: ecs
-  name: process.parent.pgid
-- external: ecs
   name: process.parent.pid
 - external: ecs
   name: process.parent.start
@@ -531,8 +529,6 @@
   name: process.pe.sections.var_entropy
 - external: ecs
   name: process.pe.sections.virtual_size
-- external: ecs
-  name: process.pgid
 - external: ecs
   name: process.pid
 - external: ecs
@@ -853,6 +849,12 @@
   name: user.risk.static_score
 - external: ecs
   name: user.risk.static_score_norm
+- external: ecs
+  name: volume.removable
+- external: ecs
+  name: volume.size
+- external: ecs
+  name: volume.writable
 - external: ecs
   name: vulnerability.score.base
 - external: ecs
--- a/packages/osquery_manager/data_stream/result/fields/osquery.yml
+++ b/packages/osquery_manager/data_stream/result/fields/osquery.yml
@@ -332,7 +332,9 @@
           norms: false
           default_field: false
     - name: app_name
-      description: windows_firewall_rules.app_name - Friendly name of the application to which the rule applies
+      description: |-
+        homebrew_packages.app_name - Name of the installed App (for Casks)
+        windows_firewall_rules.app_name - Friendly name of the application to which the rule applies
       type: keyword
       ignore_above: 1024
       multi_fields:
@@ -662,6 +664,14 @@
           default_field: false
     - name: auto_update
       description: lxd_images.auto_update - Whether the image auto-updates (1) or not (0)
+      type: keyword
+      ignore_above: 1024
+      multi_fields:
+        - name: number
+          type: long
+          default_field: false
+    - name: auto_updates
+      description: homebrew_packages.auto_updates - 1 if the cask auto-updates otherwise 0
       type: keyword
       ignore_above: 1024
       multi_fields:
@@ -2570,6 +2580,7 @@
     - name: datetime
       description: |-
         crashes.datetime - Date/Time at which the crash occurred
+        dns_lookup_events.datetime - Event timestamp in DATETIME format
         powershell_events.datetime - System time at which the Powershell script event occurred
         process_etw_events.datetime - Event timestamp in DATETIME format
         syslog_events.datetime - Time known to syslog
@@ -3296,6 +3307,7 @@
         bpf_process_events.eid - Event ID
         bpf_socket_events.eid - Event ID
         disk_events.eid - Event ID
+        dns_lookup_events.eid - Event ID
         es_process_events.eid - Event ID
         es_process_file_events.eid - Event ID
         file_events.eid - Event ID
@@ -3349,6 +3361,15 @@
       multi_fields:
         - name: number
           type: long
+          default_field: false
+    - name: emulated_cpu_type
+      description: system_info.emulated_cpu_type - Emulated CPU type
+      type: keyword
+      ignore_above: 1024
+      multi_fields:
+        - name: text
+          type: text
+          norms: false
           default_field: false
     - name: enable_admin_account
       description: security_profile_info.enable_admin_account - Determines whether the Administrator account on the local computer is enabled
@@ -4004,6 +4025,7 @@
         file.filename - Name portion of file path
         lxd_images.filename - Filename of the image file
         prefetch.filename - Executable filename.
+        recent_files.filename - The name of the file
         xprotect_entries.filename - Use this file name to match
       type: keyword
       ignore_above: 1024
@@ -6778,6 +6800,14 @@
         - name: number
           type: long
           default_field: false
+    - name: memory_inactive_file
+      description: docker_container_stats.memory_inactive_file - Memory inactive file
+      type: keyword
+      ignore_above: 1024
+      multi_fields:
+        - name: number
+          type: long
+          default_field: false
     - name: memory_limit
       description: |-
         docker_container_stats.memory_limit - Memory limit
@@ -6798,6 +6828,14 @@
           default_field: false
     - name: memory_total
       description: memory_info.memory_total - Total amount of physical RAM, in bytes
+      type: keyword
+      ignore_above: 1024
+      multi_fields:
+        - name: number
+          type: long
+          default_field: false
+    - name: memory_total_inactive_file
+      description: docker_container_stats.memory_total_inactive_file - Memory total inactive file
       type: keyword
       ignore_above: 1024
       multi_fields:
@@ -7222,6 +7260,7 @@
         gatekeeper_approved_apps.mtime - Last modification time
         process_events.mtime - File modification in UNIX time
         quicklook_cache.mtime - Parsed version date field
+        recent_files.mtime - Last modification time of the shortcut (usually corresponds to last opened time for the file)
         registry.mtime - timestamp of the most recent registry write
       type: keyword
       ignore_above: 1024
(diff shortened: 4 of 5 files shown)


leehinman (approved):
I just reviewed the changes to "windows" package since that is the only one owned by elastic-agent-data-plane.

susan-shu-c (approved):
The addition of the `network_security` category in `packages/beaconing` should work; cc @sodhikirti07

packages/bbot/manifest.yml
```diff
@@ -6,6 +6,8 @@ description: "BBOT is a recursive internet scanner inspired by Spiderfoot, but d
 type: integration
 categories:
   - security
+  # Added threat_intel category as BBOT is an OSINT tool that provides external intelligence about attack surfaces
+  - threat_intel
```
andrewkroh: I'm not sure about this one. I don't view BBOT the same as the traditional threat intel packages like ti_abusech.

packages/blacklens/manifest.yml
```diff
@@ -8,6 +8,10 @@ description: "Collect logs from blacklens.io with Elastic Agent"
 type: integration
 categories:
   - security
+  # Added threat_intel category as Blacklens provides Darknet Monitoring and external threat intelligence
+  - threat_intel
```
andrewkroh: This might be true for blacklens.io, but I think the scope of the integration only covers alerts and not their emerging threats data.

packages/ti_eset/manifest.yml
```diff
@@ -7,6 +7,10 @@ type: integration
 categories:
   - security
   - threat_intel
+  # Added siem category as ESET Threat Intelligence provides data that can be used for security monitoring and incident response
+  - siem
```
andrewkroh: With that justification you could apply `siem` to most of the packages. I think this category should be removed.

packages/ti_eset/manifest.yml
```diff
@@ -7,6 +7,10 @@ type: integration
...
+  # Added network_security category as ESET Threat Intelligence includes IP and domain indicators for network protection
+  - network_security
```
andrewkroh: This would also be true of the other TI providers too. I would omit this category, and reserve it for network based monitors.

This applies to the other TI providers as well (the next few also add these categories).

packages/ti_util/manifest.yml
```diff
@@ -5,6 +5,8 @@ description: Prebuilt Threat Intelligence dashboard for Elastic Security
 categories:
   - security
   - threat_intel
+  # Added siem category as TI Util provides dashboards for security monitoring and threat intelligence visualization
+  - siem
```
andrewkroh: Assuming `siem` is meant for integrations with other SIEM-like things, I don't think this category applies here.

```suggestion
```

packages/unifiedlogs/manifest.yml
```diff
@@ -12,6 +12,8 @@ conditions:
 categories:
```
andrewkroh, line 12: macos unified logs are also used for security purposes.

packages/windows/manifest.yml
```diff
@@ -6,6 +6,10 @@ type: integration
 categories:
   - os_system
   - security
+  # Added observability category as Windows integration provides comprehensive system metrics and logs for monitoring
+  - observability
+  # Added siem category as Windows integration provides security event data for monitoring and incident response
+  - siem
```
andrewkroh:
```suggestion
```

Assuming siem is meant for integrations with other SIEM-like things, I don't think this category applies here.

packages/zeek/manifest.yml
```diff
@@ -9,7 +9,11 @@ icons:
     size: 214x203
     type: image/svg+xml
 format_version: "3.0.3"
-categories: [network, security]
+categories: 
+  - network
+  - security
+  # Added siem category as Zeek provides security event data for monitoring and incident response
+  - siem
```
andrewkroh: Assuming siem is meant for integrations with other SIEM-like things, I don't think this category applies here.
```suggestion
```

lalit-satapathy:
Added @daniela-elastic to help review the o11y package changes.

packages/universal_profiling_agent/manifest.yml
florianl, line 1: What about universal_profiling_collector and universal_profiling_symbolizer?

packages/tomcat/manifest.yml
```diff
@@ -3,7 +3,13 @@ name: tomcat
 title: Tomcat NetWitness Logs (Deprecated)
 version: "1.14.0"
 description: Collect and parse logs from Apache Tomcat servers with Elastic Agent.
-categories: ["web", "observability"]
+categories: 
+  - web
+  - observability
+  # Added application_observability category as Tomcat is a web application server that provides observability data
+  - application_observability
+  # Added websphere category as Tomcat is a web application server similar to WebSphere
```
shmsr: Websphere is not required. Can we tweak the prompt to exclude software categories that are similar?

trisch-me: i think the best way would be to add description to categories themselves. Because it’s not clear and ambiguous. I am not even sure that the list of categories is comprehensive and we need there all those categories

trisch-me: But for this one I will just manually remove it

packages/universal_profiling_agent/manifest.yml
trisch-me: you suggest to update categories to include `observability`?

packages/ti_eset/manifest.yml
trisch-me: I have removed from other TI providers as well

packages/universal_profiling_agent/manifest.yml
trisch-me: I have updated it

packages/activemq/manifest.yml
daniela-elastic, line 1: Could we say that ActiveMQ is part of the infrastructure stack and what is the definition of infra stack? Asking because we have an integration Kafka which one could argue is an alternative to ActiveMQ (minus the streaming part) however today Kafka integration doesn't have infra category. Perhaps we should look into conformity between the various similar family of services to make sure they are categorized in the same way.

packages/airflow/manifest.yml
daniela-elastic, line 1: Airflow is primarily designed for batch processing and is not ideal for real-time streaming workflows.  Hence I don't think we should apply category stream_processing

packages/apache_tomcat/manifest.yml
daniela-elastic, line 1: Not sure it makes sense to put Apache Tomcat under category Websphere (owned by IBM) just because it is a similar product. If we decided to do that then we'd have to do it for all services that are similar.

packages/o365/manifest.yml
daniela-elastic, line 1: Could we also add observability category. There's no reason why o365 logs can't be used for observability purposes

packages/system/manifest.yml
daniela-elastic, line 1: Should we also add observability category? To me it seems a host (aka system) is the quintessential entity to observe

trisch-me:
@daniela-elastic I have updated code according to your comments

gizas (approved):

trisch-me:
@daniela-elastic @shmsr could you please re-check your review? thanks

packages/aws/manifest.yml
```diff
@@ -7,6 +7,8 @@ type: integration
 categories:
   - aws
   - cloud
+  # Added security category as AWS integration collects security-relevant data like CloudTrail logs, GuardDuty findings, and other security monitoring data
+  - security
```
daniela-elastic, line 13: Can we also add "observability" category

packages/cloud_defend/manifest.yml
```diff
@@ -9,6 +9,8 @@ type: integration
 categories:
   - containers
   - kubernetes
+  # Observability is a parent category for kubernetes
+  - observability
```
daniela-elastic, line 13: I'm not sure that cloud defend is an observability category. Shouldn't this be "security"?

packages/falco/manifest.yml
```diff
@@ -8,6 +8,8 @@ categories:
   - containers
   - kubernetes
   - monitoring
+  # Observability is a parent category for kubernetes and monitoring
+  - observability
```
daniela-elastic, line 12: I'm not sure that Falco integration is used for observability

packages/fortinet_fortiproxy/manifest.yml
```diff
@@ -6,6 +6,8 @@ description: "Collect logs from Fortinet FortiProxy with Elastic Agent."
 type: integration
 categories:
   - network
+  # Observability is a parent category for web
+  - observability
```
daniela-elastic, line 10: I think we should revisit adding "observability" to some (but not all) of the security integrations. If we do decide that everything that collects signals is under the "observability" umbrella, then we need to be consistent and add observability to every integration. But if this is not the case, I'm struggling to understand why some security integrations are  tagged as "observability"

packages/gigamon/manifest.yml
```diff
@@ -9,6 +9,8 @@ categories:
   - security
   - network
   - application_observability
+  # Observability is a parent category for application_observability
+  - observability
```
daniela-elastic, line 13: Same comment regarding tagging some security integrations as "observability"

packages/menlo/manifest.yml
```diff
@@ -9,6 +9,8 @@ type: integration
 categories:
   - monitoring
   - network
+  # Observability is a parent category for monitoring
+  - observability
```
daniela-elastic, line 13: Same comment regarding tagging some security integrations as "observability"

packages/miniflux/manifest.yml
```diff
@@ -7,6 +7,8 @@ source:
 description: Collect RSS feed content from the Miniflux API with Elastic Agent.
 type: integration
 categories:
+  # Observability is a parent category for web
+  - observability
```
daniela-elastic, line 11: Same comment regarding tagging some security integrations as "observability"
Also, I can't see the security tag here (maybe it's in a line further down)

packages/squid/manifest.yml
```diff
@@ -4,6 +4,8 @@ title: Squid Proxy
 version: "1.3.0"
 description: Collect and parse logs from Squid devices with Elastic Agent.
 categories:
+  # Observability is a parent category for web
+  - observability
```
daniela-elastic, line 8: Same comment regarding tagging some security integrations as "observability"

packages/sysdig/manifest.yml
```diff
@@ -8,6 +8,8 @@ categories:
   - containers
   - kubernetes
   - monitoring
+  # Observability is a parent category for kubernetes and monitoring
+  - observability
```
daniela-elastic, line 12: Same comment regarding tagging some security integrations as "observability"

packages/teleport/manifest.yml
```diff
@@ -9,6 +9,8 @@ type: integration
 categories:
   - monitoring
   - network
+  # Observability is a parent category for monitoring
+  - observability
```
daniela-elastic, line 13: Same comment regarding tagging some security integrations as "observability"

packages/tetragon/manifest.yml
```diff
@@ -12,6 +12,8 @@ categories:
   - cloud
   - cloudsecurity_cdr
   - kubernetes
+  # Observability is a parent category for kubernetes
+  - observability
```
daniela-elastic, line 16: Same comment regarding tagging some security integrations as "observability"

daniela-elastic (commented):
I left a few comments / questions. It's possible that I'm missing some context on the decision to apply "observability' tag to some (but not all) of the security integrations. I'm happy to have a wider discussion if necessary.

trisch-me:
hey @daniela-elastic I agree with you, it was automated update based on this [structure](https://github.com/elastic/package-registry/blob/main/categories/categories.yml) where for example Kubernetes or web is a subcategory for observability. Should we revisit then this tree as well?

packages/cloud_defend/manifest.yml
trisch-me: Security is a already a category there, I have added observability because it is a parent category for Kubernetes

packages/falco/manifest.yml
trisch-me: I have added observability because it is a parent category for Kubernetes and monitoring

packages/fortinet_fortiproxy/manifest.yml
trisch-me: I have put as a comment a link to category structure, where some categories are main and some are sub-categories. So I have checked all integrations and updated all that have subcategories to have main categories as well

Please let me know if this is wrong. I believe we should update then tree structure and remove some sub categories from parent categories

packages/miniflux/manifest.yml
trisch-me: I don’t think miniflux is used by security. Do you have another opinion?

trisch-me:
@shmsr @daniela-elastic could you please re-review this PR?

daniela-elastic:
> hey @daniela-elastic I agree with you, it was automated update based on this [structure](https://github.com/elastic/package-registry/blob/main/categories/categories.yml) where for example Kubernetes or web is a subcategory for observability. Should we revisit then this tree as well?

Yes, agree this should be revisited, especially if it drives decisions on how to categorise each integration

trisch-me:
@daniela-elastic should I remove latest changes then in this PR for it to be merged?

packages/cloud_defend/manifest.yml
daniela-elastic: is kubernetes a parent of cloud_defend?

packages/falco/manifest.yml
daniela-elastic: Are Kubernetes and/or monitoring parent(s) of falco category?

packages/miniflux/manifest.yml
daniela-elastic: Given that Miniflux is an RSS feed reader for content,  is likely that this integration is for content.  Tbf I haven't looked at what data it collects exactly so I might be off. Currently the codeowners for this integration are security and the obs team doesn't work on it

packages/mysql_enterprise/manifest.yml
```diff
@@ -7,6 +7,8 @@ type: integration
 categories:
```
daniela-elastic, line 7: we have marked the mysql integration as obs category. can we also add observability to this mysql integration for consistency?

trisch-me:
Resolved both of the comments (by adding requested category)

trisch-me:
@shmsr could please update your review?

trisch-me:
@elastic/obs-ux-management-team @elastic/ml-ui @elastic/obs-ds-intake-services @elastic/search-extract-and-transform @elastic/integration-experience 

could I get your review for this PR?

peteharverson (approved):
ml-ui owned change to beaconing package categories LGTM.

PR #14577, opened 2025-07-17
[statsd_input] update statsd_input package manifest.yml format_version to 3.4.0.

giorgi-imerlishvili-elastic:
## Proposed commit message

Update package statsd_input's manifest.yml format_version to 3.4.0
Update  files packages/statsd_input/fields/input.yml and packages/statsd_input/manifest.yml accordingly 

## Checklist

## Related issues

Closes https://github.com/elastic/obs-integration-team/issues/387


packages/statsd_input/fields/input.yml
```diff
@@ -8,3 +8,4 @@
 - name: statsd.*.*
   description: Default mapping for metric fields. You need to add your own custom mappings when storing other kinds of information.
   type: object
+  object_type: keyword
```
mykola-elastic:
```suggestion
  object_type: double
```
it seems the system tests only use numbers, so we can try with `double`.

mykola-elastic (approved):
LGTM!

mykola-elastic:
@ishleenk17 system tests pass and it was tested using local stack by @giorgi-imerlishvili-elastic

packages/statsd_input/manifest.yml
```diff
@@ -1,9 +1,11 @@
-format_version: 2.9.0
+format_version: 3.4.0
 name: statsd_input
 title: StatsD Input
-version: "0.5.1"
+version: "0.6.0"
 description: StatsD Input Package
 type: input
+source:
```
ishleenk17: Was it errroing out that we had to add this ?

giorgi-imerlishvili-elastic: I just looked other packages with version upper than 3.0 and used those packages as example
But I think without that property it should work either, because it's not defined as required property in spec: https://github.com/elastic/package-spec/blob/v3.4.0/spec/input/manifest.spec.yml

mykola-elastic: @ishleenk17 shall `source.license` be removed?

stefans-elastic (approved):
LGTM

packages/statsd_input/manifest.yml
ishleenk17: If it's not a mandate, we can remove it

ishleenk17 (approved):

PR #14593, opened 2025-07-18
Fix go.mod, update all dependencies except elastic-package

jsoriano:
There was a conflict between dependencies that was producing failures with dependabot, try to fix it, and update all dependencies.

`elastic-package` is intentionally not updated here because we need to make changes in pipelines before updating to 0.113.0.


go.mod
```diff
@@ -18,101 +18,101 @@ require (
...
-	github.com/elastic/go-windows v1.0.1 // indirect
+	github.com/elastic/go-sysinfo v1.15.3 // indirect
+	github.com/elastic/go-windows v1.0.2 // indirect
 	github.com/elastic/gojsonschema v1.2.1 // indirect
 	github.com/elastic/kbncontent v0.1.4 // indirect
-	github.com/elastic/package-spec/v3 v3.4.0 // indirect
-	github.com/emicklei/go-restful/v3 v3.12.1 // indirect
-	github.com/envoyproxy/go-control-plane v0.13.1 // indirect
-	github.com/envoyproxy/protoc-gen-validate v1.1.0 // indirect
+	github.com/elastic/package-spec/v3 v3.4.1 // indirect
```
mrodm: Should we keep this dependency to be 3.4.0 , so it gets updated with `elastic-package v0.113.0` ?

jsoriano: Yes, that would be better, good catch.

PR #14594, opened 2025-07-18
Update elastic-package and check links when checking sources

jsoriano:
Update elastic-package to 0.113.0. This version includes support for links, add a new check when checking sources to check their consistency.


.buildkite/scripts/check_sources.sh
```diff
@@ -10,3 +10,6 @@ with_mage
 mage -v check
 
 check_git_diff
+
+use_elastic_package
+${ELASTIC_PACKAGE_BIN} links check
```
jsoriano, line 15: @mrodm not sure if adding it here is the best place... would you have other suggestions?

mrodm: Following the documentation https://github.com/elastic/elastic-package/blob/0ae85732ad409e95dccb703b815820b1086d373b/docs/howto/links.md, I've run some tests locally. It looks like that the error shown here it could be improved to help debugging.

Scenario:
- File created as a links reference in `packages/elastic_package_registry/data_stream/metrics/_dev/shared/myfields.yml`
- `packages/elastic_package_registry/data_stream/metrics/fields/shared.yml.link` contains a link to `../../_dev/shared/myfields.yml` (missing to add another `../`)

>  $ elastic-package links update
Update linked files checksums if needed.
Error: updating linked files checksums failed: updating linked files checksums failed: could not initialize linked file /home/mariorodriguez/Coding/work/integrations/packages/elastic_package_registry/data_stream/metrics/fields/shared.yml.link: path /home/mariorodriguez/Coding/work/integrations/packages/elastic_package_registry/data_stream/metrics/_dev/shared/myfields.yml escapes the repository root


In that scenario, the link file did not contain the right path to the file, and therefore it did not exist the target link but it does not escape the repository root.

In any case, as this command can be executed out of each package folder (it can check sub-directories), LGTM to add here this command :+1:

jsoriano: I am taking a look to the issue with the error messages https://github.com/elastic/elastic-package/issues/2751

I will merge in the meantime, as it should still detect problematic links.

PR #14609, opened 2025-07-21
[o365] No delays until up-to-date

chrisberkhout:
## Proposed commit message

```
[o365] No delays until up-to-date

Instead of fetching history continuously until reaching the current day
then waiting the interval time between requests, keep fetching until
within one batch length of the current time.

With the default settings, waiting 3 minutes between requests for one
hour batches would significantly extend the time required to get
up-to-date. For example, it's 45 minutes of waiting to cover 15 hours.
It's particularly disruptive to the user experience if there is no
historical data and new data has such a delay.
```

## Checklist

### DIFF
--- a/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
+++ b/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
@@ -171,13 +171,17 @@ program: |-
   														(list_contents_resp.Header.Nextpageuri[0])
   													:
   														"",
-  													// keep fetching more if (nextpageuri exists) or (max time returned date != today's date)
+  													// keep fetching more if (nextpageuri exists) or (the request time range ends within one batch duration of now)
   													"want_more_content": has(list_contents_resp.Header) &&
   													(
   														has(list_contents_resp.Header.NextPageUri) && list_contents_resp.Header.NextPageUri.size() > 0 ||
-  														has(list_contents_resp.Header.Nextpageuri) && list_contents_resp.Header.Nextpageuri.size() > 0
-  													) || {"temp": list_contents_resp_body}.collate("temp.contentCreated").max().split("T").as(t, t.size() > 1 &&
-  													t[0] != now().format("2006-01-02")),
+  														has(list_contents_resp.Header.Nextpageuri) && list_contents_resp.Header.Nextpageuri.size() > 0 ||
+  														list_contents_resp.Request.URL.parse_url().RawQuery.parse_query().as(reqQuery,
+  														  reqQuery.?endTime[?0].optMap(endTime,
+  														    (now() - endTime.parse_time(time_layout.RFC3339)) >= duration(state.base.batch_interval)
+  														  ).orValue(false)
+  														)
+  													)
   												}
   											)
   										:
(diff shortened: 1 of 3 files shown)


packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -171,13 +171,17 @@ program: |-
...
+  													// keep fetching more if (nextpageuri exists) or (the request time range ends within one batch duration of now)
   													"want_more_content": has(list_contents_resp.Header) &&
   													(
   														has(list_contents_resp.Header.NextPageUri) && list_contents_resp.Header.NextPageUri.size() > 0 ||
-  														has(list_contents_resp.Header.Nextpageuri) && list_contents_resp.Header.Nextpageuri.size() > 0
-  													) || {"temp": list_contents_resp_body}.collate("temp.contentCreated").max().split("T").as(t, t.size() > 1 &&
-  													t[0] != now().format("2006-01-02")),
+  														has(list_contents_resp.Header.Nextpageuri) && list_contents_resp.Header.Nextpageuri.size() > 0 ||
+  														list_contents_resp.Request.URL.parse_url().RawQuery.parse_query().as(reqQuery,
+  														  reqQuery.?endTime[?0].optMap(endTime,
```
efd6:
```suggestion
  														  reqQuery.?endTime[0].optMap(endTime,
```
We are already optional, so the second `?` is not required.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -171,13 +171,17 @@ program: |-
...
+  														    (now() - endTime.parse_time(time_layout.RFC3339)) >= duration(state.base.batch_interval)
+  														  ).orValue(false)
+  														)
```
efd6:
```suggestion
  														list_contents_resp.Request.URL.parse_url().RawQuery.parse_query().?endTime[0].optMap(endTime,
														    (now() - endTime.parse_time(time_layout.RFC3339)) >= duration(state.base.batch_interval)
														  ).orValue(false)
```
I'm not sure about whether this is better, but do we need the `as` since there is only one use?

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -171,13 +171,17 @@ program: |-
   														(list_contents_resp.Header.Nextpageuri[0])
   													:
   														"",
-  													// keep fetching more if (nextpageuri exists) or (max time returned date != today's date)
+  													// keep fetching more if (nextpageuri exists) or (the request time range ends within one batch duration of now)
```
efd6: "within"? The logic below is `now()-query_endtime ≥ batch_interval`. For "within" shouldn't that be `≤`. Do you mean "before"?

chrisberkhout: yeah, the words are backwards. Should have been "not within", but I've corrected it to:
> the end of the request range is one batch interval or longer before the current time

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: Removed.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: True. I think it is better.

efd6 (approved):
Thanks

PR #14613, opened 2025-07-21
use links panels instead of markdown for dashboard navigation

tommyers-elastic:
This PR replaces the markdown navigation section at the top of most K8s dashboards with a native 'Links' panel which maintains filters between navigation.

_Copilot wrote most of these changes. I visually verified they look correct, and the package still builds. But as a note to the reviewer, please double check you're happy that nothing else changed here._ 

I was not able to include the panel as a reference due to 'links' not being allowed in package-spec, so the json is repeated in every dashboard. I know we generally prefer inline visualisations, but in this case it seems like a reference would be significantly better from a maintanence POV. We should potentially resolve this by allowing links JSON in packages under kibana/links.

The existing markdown panels are replaced without modification to sizing or position, so you end up with some looking kinda weird. e.g.

This existing markdown panel:

Screenshot 2025-07-21 at 11 35 38

becomes like this:

Screenshot 2025-07-21 at 11 35 41

This sucked in the first place, so I don't think anything got worse.

Because the links panel retains query filters as you navigate between dashboards, I also had to modify the API server dashboard to use per-panel filtering instead of at the top level. I also enabled margins on this one, since it's a simple boolean toggle, and it was only one not using them. 

Before:

Screenshot 2025-07-21 at 11 36 29

After:

Screenshot 2025-07-21 at 11 55 05

Another sucky part of this change is that you have to scroll the links panel to get to the links 'further down' (on the right). If we want to be able to have the links fill more vertical space we need to take this up with the dashboards team.

Fixes https://github.com/elastic/kibana/issues/202471.

### DIFF
--- a/packages/kubernetes/kibana/dashboard/kubernetes-0a672d50-bcb1-11ec-b64f-7dd6e8e82013.json
+++ b/packages/kubernetes/kibana/dashboard/kubernetes-0a672d50-bcb1-11ec-b64f-7dd6e8e82013.json
@@ -36,41 +36,116 @@
     },
     "panelsJSON": [
       {
+        "type": "links",
+        "title": "Kubernetes Dashboards [Metrics Kubernetes]",
         "embeddableConfig": {
-          "enhancements": {},
-          "savedVis": {
-            "data": {
-              "aggs": [],
-              "searchSource": {
-                "filter": [],
-                "query": {
-                  "language": "kuery",
-                  "query": ""
-                }
-              }
-            },
+          "attributes": {
+            "title": "Kubernetes Dashboards [Metrics Kubernetes]",
             "description": "",
-            "params": {
-              "fontSize": 10,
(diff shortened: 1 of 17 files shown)


packages/kubernetes/kibana/dashboard/kubernetes-0a672d50-bcb1-11ec-b64f-7dd6e8e82013.json
```diff
@@ -36,41 +36,116 @@
...
-            "uiState": {}
-          }
+            "layout": "horizontal",
+            "links": [
+              {
+                "label": "Kubernetes Overview",
+                "type": "dashboardLink",
+                "id": "6416766e-26ec-4394-b02d-399b42ca9a6e",
+                "order": 0,
+                "destinationRefName": "link_6416766e-26ec-4394-b02d-399b42ca9a6e_dashboard"
```
gizas, line 52: My only concern is that we change the ids of the dashbaords and this is a breaking change.
So for example this should align with kubernetes-f4dc26db-1b53-4ea2-a78b-1bfab8ea267c that it was before.

Also does this add the `/view/` suffix in front? 
Overall the new panels should create this url for example for the Kubernetes overview 
`<url>/app/dashboards#/view/kubernetes-f4dc26db-1b53-4ea2-a78b-1bfab8ea267c`

roshan-elastic:
Hey @daniela-elastic @bturquet,

If we want to get this merged, do you know who needs to do it?

We probably need a couple of UX tweaks before it can be merged though.

cc @smith

gizas:
@roshan-elastic (because @bturquet is on PTO) this is just an integration PR so we dont have a problem to merge it and this will be available to the latest version of the kubernetes package.

roshan-elastic:
Cheers @gizas - I'll pull this one back as I think it needs some UX changes. I'll make a note to see if we want to prioritise making those changes or not

Issue #14437, opened 2025-07-07
Wiz: Vulnerability Datastream only retrieves OPEN vulnerabilities event


BipoGrande:
### Integration Name

Wiz [wiz]

### Dataset Name

wiz.vulnerabilities

### Integration Version

3.3.0

### Agent Version

9.0.2

### OS Version and Architecture

agentless

### User Goal

Currently only events where the vulnerabilities are in the 'OPEN' status are retrieved, we would like to see when those vulnerabilities are updated to other statuses.

### Existing Features

Without fetching other statuses it doesn't allow to see any actions performed on those vulnerabilities such as when they are closed.

### What did you see?

In the GraphQL query the filterby field needs to override the default status filter and add the other status types: https://github.com/elastic/integrations/blob/main/packages/wiz/data_stream/vulnerability/agent/stream/cel.yml.hbs#L142

`
filterBy: {
      status: [RESOLVED, OPEN, REJECTED, IN_PROGRESS],
 }
`

PR #14664, opened 2025-07-23
[wiz] Add Filters to Retrieve all Types of Vulnerabilities

mohitjha-elastic:
## Proposed commit message
```
wiz: add filters to retrieve all types of vulnerabilities in the vulnerability dataset

This change introduces filters to the vulnerability dataset, allowing
retrieval of all types of vulnerabilities. It enhances the dataset's
flexibility by ensuring comprehensive coverage across different
vulnerability categories.
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/wiz directory.

> elastic-package test -v

## Related Issue

- Closes https://github.com/elastic/integrations/issues/14437

### DIFF
--- a/packages/wiz/data_stream/vulnerability/agent/stream/cel.yml.hbs
+++ b/packages/wiz/data_stream/vulnerability/agent/stream/cel.yml.hbs
@@ -25,6 +25,7 @@ state:
   initial_interval: {{initial_interval}}
   want_more: false
   batch_size: {{batch_size}}
+  include_all_vulnerability: {{include_all_vulnerability}}
   query: >-
     query VulnerabilityFindingsPage($filterBy: VulnerabilityFindingFilters $first: Int $after: String $orderBy: VulnerabilityFindingOrder) {
        vulnerabilityFindings(filterBy: $filterBy first: $first after: $after orderBy: $orderBy) {
@@ -140,7 +141,12 @@ program: |
           "first": state.batch_size,
           "after": state.?end_cursor.value.orValue(null),
           "filterBy": {
-            "status": ["RESOLVED", "OPEN", "REJECTED", "IN_PROGRESS"],
+            "status": (
+              state.?include_all_vulnerability.orValue(false) ?
+                ["RESOLVED", "OPEN", "REJECTED", "IN_PROGRESS"]
+              :
+                ["OPEN"]
+            ),
             "updatedAt": {
               "after": state.want_more ?
                   state.?cursor.first_timestamp.orValue(null)
--- a/packages/wiz/data_stream/vulnerability/agent/stream/cel.yml.hbs
+++ b/packages/wiz/data_stream/vulnerability/agent/stream/cel.yml.hbs
@@ -25,7 +25,13 @@ state:
   initial_interval: {{initial_interval}}
   want_more: false
   batch_size: {{batch_size}}
-  include_all_vulnerability: {{include_all_vulnerability}}
+  {{#if include_all_vulnerability}}
+  want_status:
+    - RESOLVED
+    - OPEN
+    - REJECTED
+    - IN_PROGRESS
+  {{/if}}
   query: >-
     query VulnerabilityFindingsPage($filterBy: VulnerabilityFindingFilters $first: Int $after: String $orderBy: VulnerabilityFindingOrder) {
        vulnerabilityFindings(filterBy: $filterBy first: $first after: $after orderBy: $orderBy) {
@@ -141,12 +147,7 @@ program: |
           "first": state.batch_size,
           "after": state.?end_cursor.value.orValue(null),
           "filterBy": {
-            "status": (
-              state.?include_all_vulnerability.orValue(false) ?
-                ["RESOLVED", "OPEN", "REJECTED", "IN_PROGRESS"]
-              :
-                ["OPEN"]
-            ),
+            ?"status": state.?want_status,
             "updatedAt": {
               "after": state.want_more ?
                   state.?cursor.first_timestamp.orValue(null)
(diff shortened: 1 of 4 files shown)


packages/wiz/data_stream/vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -140,6 +140,7 @@ program: |
           "first": state.batch_size,
           "after": state.?end_cursor.value.orValue(null),
           "filterBy": {
+            "status": ["RESOLVED", "OPEN", "REJECTED", "IN_PROGRESS"],
```
efd6: I'm wondering whether we want this to be configurable?

cc @maxcold

mohitjha-elastic: We typically fetch all types of data that could potentially be useful, allowing users to apply Kibana filters later based on their needs. However, I understand this approach might lead to unnecessary index space consumption for users who are not interested in viewing resolved vulnerabilities.

maxcold: сс @nick-alayil

maxcold: after talking with @nick-alayil we think the best way forward is to make it configurable with OPEN being the default to keep the existing behaviour. Thanks for flagging it @efd6 !

efd6: @mohitjha-elastic Please add a manifest entry to allow this to be configurable.

@maxcold What kind of configuration would you prefer? The story around this kind of option is not a happy one; we either have a set of four booleans which do not show obvious grouping in the UI, 2^4 menu options, or multi-text (which allows typo-based misconfiguration) — or YAML. None of the available options are good.

nick-alayil: My $0.02. Can we've a simple toggle to control this configuration?

**Default (toggle OFF): Active vulnerabilities only**
- Sends only "OPEN" status vulnerabilities  
- Covers majority of use cases since most teams primarily care about actionable items
- Simple, obvious, can't be misconfigured

**Toggle ON: All vulnerability statuses**
- Includes RESOLVED, OPEN, REJECTED, IN_PROGRESS
- For users who want comprehensive data for reporting/analytics
- Still just one boolean setting. Impossible to mess up

efd6: That would work for me.

mohitjha-elastic: Thanks! Added a toggle option to control this configuration-

image

packages/wiz/data_stream/vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -140,6 +141,12 @@ program: |
           "first": state.batch_size,
           "after": state.?end_cursor.value.orValue(null),
           "filterBy": {
+            "status": (
+              state.?include_all_vulnerability.orValue(false) ?
+                ["RESOLVED", "OPEN", "REJECTED", "IN_PROGRESS"]
+              :
+                ["OPEN"]
```
maxcold: btw I didn't find any docs in the Wiz partner portal that says that by default it queries only OPEN vulnerabilities. What concerns me is that we are changing the behaviour, as the query didn't have any status filter before, and we assumed that by default it filtered in only OPEN vulnerabilities. Do we know that for sure?

maxcold: @mohitjha-elastic

efd6: I agree. This should not set the parameter in the case that the option is not selected.

I would suggest instead of using a boolean here, use
```
{{#if include_all_vulnerability}}
  want_status:
    - RESOLVED
    - OPEN
    - REJECTED
    - IN_PROGRESS
{{/if}}
```
in the state definition. And then here
```
?"status": state.?want_status,
```

mohitjha-elastic: Thanks @efd6 @maxcold 
The suggestions have been implemented.

packages/wiz/data_stream/vulnerability/manifest.yml
```diff
@@ -7,6 +7,16 @@ streams:
     description: Collect Vulnerability logs from Wiz.
     template_path: cel.yml.hbs
     vars:
+      - name: include_all_vulnerability
+        required: true
+        show_user: true
+        title: Include All Vulnerability Statuses
+        description: >-
+          Enable this toggle to fetch all types of vulnerabilities, including those with statuses RESOLVED, OPEN, REJECTED, and IN_PROGRESS.
+          By default, only active vulnerabilities with the OPEN status are retrieved.
```
efd6: I would leave this as "By default, the API's default filter is used." This protects us from future changes in API behaviour.

/cc @maxcold for opinion.

mohitjha-elastic: I agree with the suggestion, as the documentation does not explicitly mention that only open vulnerabilities are fetched by default.

efd6 (approved):
Will wait for @maxcold or @nick-alayil to comment on the documentation before merge.

packages/wiz/data_stream/vulnerability/manifest.yml
maxcold: yes, let's not make guesses and have it as `By default, the API's default filter is used.` , thanks for checking!

PR #14665, opened 2025-07-23
Mapping of Gigamon Metadata Attributes to ECS

apps-elastic-gigamon:
- Enhancement
-  Have mapped the Gigamon metadata attributes which are required for creating AMI starter Pack dashboards to ECS fields 
-  Tested the changes by adding the relevant metadata fields to test-ami.json and validating the corresponding expected results json.

### DIFF
--- a/packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
@@ -465,7 +465,7 @@ processors:
             ignore_missing: true
         - append:
             field: error.message
-            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'  
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
   - script:
       lang: painless
       description: Gigamon AMI lookup mappings
@@ -1199,14 +1199,14 @@ processors:
       if: ctx.gigamon?.ami?.dst_packets != null
   - set:
       field: service.version
-      tag: set_snmp_version
-      copy_from: gigamon.ami.snmp_version
-      if: ctx.gigamon?.ami?.snmp_version != null
-  - set:
-      field: service.version
       tag: set_smb_version
       copy_from: gigamon.ami.smb_version_value
       if: ctx.gigamon?.ami?.smb_version_value != null
+  - set:
+      field: service.version
+      tag: set_snmp_version
+      copy_from: gigamon.ami.snmp_version
+      if: ctx.gigamon?.ami?.snmp_version != null
   - set:
       field: event.id
       tag: set_event_id
@@ -1228,7 +1228,7 @@ processors:
      description: Replace ":" in destination mac with "-"
      source: >-
        ctx.destination.mac = ctx.gigamon.ami.dst_mac.replace(":", "-").toUpperCase();
-     if: ctx.gigamon?.ami?.dst_mac != null
+     if: ctx.gigamon?.ami?.dst_mac != null  
   - remove:
       field:
         - ts
--- a/packages/gigamon/data_stream/ami/_dev/test/pipeline/test-ami.json
+++ b/packages/gigamon/data_stream/ami/_dev/test/pipeline/test-ami.json
@@ -1083,7 +1083,7 @@
                 "generator": "gs_apps_appInst16_423722da-33ec-1556-b24b-cda2e74a53f6",
                 "dst_mac": "0c:c4:7a:f8:0d:c4",
                 "src_mac": "00:50:56:9f:7f:ff",
-                "src_ip": "10.115.83.4",
+                "src_ip": "89.160.20.112",
                 "dst_ip": "10.115.81.118",
                 "protocol": "6",
                 "src_port": "50694",
--- a/packages/gigamon/data_stream/ami/_dev/test/pipeline/test-ami.json-expected.json
+++ b/packages/gigamon/data_stream/ami/_dev/test/pipeline/test-ami.json-expected.json
@@ -2612,8 +2612,26 @@
                 "type": "https"
             },
             "source": {
+                "as": {
+                    "number": 29518,
+                    "organization": {
+                        "name": "Bredband2 AB"
+                    }
+                },
                 "bytes": 399,
-                "ip": "10.115.83.4",
+                "geo": {
+                    "city_name": "Linköping",
+                    "continent_name": "Europe",
+                    "country_iso_code": "SE",
+                    "country_name": "Sweden",
+                    "location": {
+                        "lat": 58.4167,
+                        "lon": 15.6167
+                    },
+                    "region_iso_code": "SE-E",
+                    "region_name": "Östergötland County"
+                },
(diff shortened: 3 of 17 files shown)


packages/gigamon/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.8.0"
+  changes:
+    - description: Mapped Gigamon Metadata attributes to ECS fields
+      type: enhancement
+      link:
```
kcreddy:
```suggestion
      link: https://github.com/elastic/integrations/pull/14665
```

packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -827,10 +853,522 @@ processors:
...
+        - gigamon.ami.tcp_wrong_crc
+        - gigamon.ami.dst_mac
+        - gigamon.ami.src_mac
+        - gigamon.ami.tcp_retransmission_bytes
+        - gigamon.ami.smb_command_string
+        - gigamon.ami.smb_filename
+        - gigamon.ami.smb_host
+        - gigamon.ami.smb_path
       if: ctx.tags == null || !(ctx.tags.contains('preserve_duplicate_custom_fields'))
       ignore_missing: true
```
kcreddy: Don't remove any of the `gigamon.ami.*` custom fields before updating dashboards, because these are still being used inside dashboards and the default value of `preserve_duplicate_custom_fields` is `false`. So, all these custom fields will be deleted by default, which we don't want to happen yet resulting in empty dashboards.

packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -827,10 +853,522 @@ processors:
...
+  - set:
+      field: dns.type
+      tag: set_dns_type
+      copy_from: gigamon.ami.dns_message_type
+      if: ctx.gigamon?.ami?.dns_message_type != null
+  - set:
+      field: gigamon.ami.dns.tunneling
+      tag: set_dns_tunneling
+      copy_from: gigamon.ami.dns_tunneling
+      if: ctx.gigamon?.ami?.dns_tunneling != null
```
kcreddy: Both `gigamon.ami.dns_tunneling` and `gigamon.ami.dns.tunneling` are custom fields. There is no difference between them other than extra nesting (`_` vs `.`). If you use a `set` processor here, you will have both fields.
Why to create 2 custom fields here populating the same value?

If you want to only have the field named `gigamon.ami.dns.tunneling` (i.e., with nesting), you could use a [`rename`](https://www.elastic.co/docs/reference/enrich-processor/rename-processor) processor instead. Rename processor renames the existing field itself, instead of copying value into another field (like `set` processor).

Example of `rename` processor:
```yaml
  - rename:
      field: gigamon.ami.dns_tunneling
      tag: rename_dns_tunneling
      target_field: gigamon.ami.dns.tunneling
      ignore_missing: true
```

Also, I suggest if you take this approach you (extra nesting with `.`), then do this consistently. You shouldn't have some custom fields with `gigamon.ami.dns_xyz` and others with `gigamon.ami.dns.xyz` and same for other apps also `http`, `ssl`, etc.

packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -827,10 +853,522 @@ processors:
...
+  - set:
+      field: tls.client.subject
+      tag: set_tls_client_subject
+      copy_from: gigamon.ami.ssl_certificate_subject_cn
+      if: ctx.gigamon?.ami?.ssl_certificate_subject_cn != null 
+  - set:
+      field: source.ip
+      tag: set_source_ip
+      copy_from: gigamon.ami.src_ip
+      if: ctx.gigamon?.ami?.src_ip != null
```
kcreddy, line 1099: After populating `source.ip`, apply [these 4 processors](https://github.com/elastic/integrations/blob/main/packages/aws/data_stream/firewall_logs/elasticsearch/ingest_pipeline/default.yml#L95-L114) (2 `geoip`, 2 `rename`) so that appropriate [ECS geo](https://www.elastic.co/docs/reference/ecs/ecs-geo) fields are also populated.

packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -827,10 +853,522 @@ processors:
...
+  - set:
+      field: destination.ip
+      tag: set_destination_ip
+      copy_from: gigamon.ami.dst_ip
+      if: ctx.gigamon?.ami?.dst_ip != null
```
kcreddy, line 1104: After populating destination.ip, apply [these 4 processors](https://github.com/elastic/integrations/blob/main/packages/aws/data_stream/firewall_logs/elasticsearch/ingest_pipeline/default.yml#L139-L158) (2 geoip, 2 rename) so that appropriate [ECS geo](https://www.elastic.co/docs/reference/ecs/ecs-geo) fields are also populated.

packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -827,10 +853,522 @@ processors:
...
+      if: ctx.gigamon?.ami?.end_time != null
+  - set:
+      field: source.ip
+      tag: set_source_ip
+      copy_from: gigamon.ami.src_ipv6
+      if: ctx.gigamon?.ami?.src_ipv6 != null
+  - set:
+      field: destination.ip
+      tag: set_destination_ip
+      copy_from: gigamon.ami.dst_ipv6
```
kcreddy, line 1138: What if both `dst_ip` and `dst_ipv6` are available in data? Which one should take preference for populating `destination.ip`? 
Similarly for `src_ip` and `src_ipv6` .

packages/gigamon/data_stream/ami/fields/ecs-custom.yml
```diff
@@ -0,0 +1,54 @@
...
+  fields:
+    - name: response.content_encoding
+      type: keyword
+    - name: response.headers.set_cookie
+      type: keyword
+- name: snmp
+  type: group
+  fields:
+    - name: version
+      type: keyword
```
kcreddy: This is exactly how we _shouldn't_ name the custom fields. If ECS fields are created with these exact name in future, there could be mapping conflicts.
What if you just put them all under `gigamon.ami` just like other custom fields?

packages/gigamon/data_stream/ami/fields/ecs-custom.yml
```diff
@@ -0,0 +1,54 @@
+- name: gigamon.ami
+  type: group
+  fields:
+    # base fields
+    - name: dns.tunneling
+      type: keyword
```
kcreddy: I am curious with the approach here. If we are nesting custom fields inside `gigamon.ami` anyway, then is there any reason to call the field `dns.tunneling` over `dns_tunneling` (i.e., what we already have)?
I just don't want some custom fields to be called `gigamon.ami.http_xyz` and some other `gigamon.ami.http.xyz`. We need consistent naming whichever approach we choose (`_` underscore vs `.` dot). 

All fields inside `fields.yml` uses `_` underscore, and all fields inside `ecs-custom.yml` uses `.` dot. This is highly inconsistent. I am starting to believe you don't need to add file `ecs-custom.yml` at all. 
I suggest you continue to use underscore `_` (instead of `.` dot) inside custom fields names and add any missing custom fields into `fields.yml` itself.

apps-elastic-gigamon:
Messed up with commits and amend. Will raise a new PR addressing these comments

PR #14671, opened 2025-07-24
[PANW ] Add conditional to processors populating _temp_ fields for internal and external zones

leandrojmp:
## Proposed commit message

The first 2 processors in the logic to populate the field `network.direction` relies on the existence of the fields `_conf.internal_zones` and `_conf.external_zones` if the user does not configure those fields in the integration, or if the source data does not have this field, those processors would still try to process every message and would fail because the field does not exist, this can add some extra time in the overall processing time of the ingest pipeline.

These 2 processors needs to be conditionally executed to avoid this, we detected this issue while troubleshooting the performance of the ingest pipeline.

 - Bug

### DIFF
--- a/packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/traffic.yml
+++ b/packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/traffic.yml
@@ -249,10 +249,12 @@ processors:
   - set:
       field: _temp_.external_zones
       copy_from: _conf.external_zones
+      if: ctx._conf?.external_zones != null
       ignore_failure: true
   - set:
       field: _temp_.internal_zones
       copy_from: _conf.internal_zones
+      if: ctx._conf?.internal_zones != null
       ignore_failure: true
   - set:
       field: network.direction
--- a/packages/panw/changelog.yml
+++ b/packages/panw/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "5.3.2"
+  changes:
+    - description: Add conditional to processors populating _temp_ fields for internal and external zones.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/XXXX
 - version: "5.3.1"
   changes:
     - description: Update documentation
--- a/packages/panw/manifest.yml
+++ b/packages/panw/manifest.yml
@@ -1,6 +1,6 @@
 name: panw
 title: Palo Alto Next-Gen Firewall
-version: "5.3.1"
+version: "5.3.2"
 description: Collect logs from Palo Alto next-gen firewalls with Elastic Agent.
 type: integration
 format_version: "3.0.3"


packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/traffic.yml
```diff
@@ -249,10 +249,12 @@ processors:
   - set:
       field: _temp_.external_zones
       copy_from: _conf.external_zones
+      if: ctx._conf?.external_zones != null
```
taylor-swanson, line 252:
```suggestion
      ignore_empty_value: true
```

packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/traffic.yml
```diff
@@ -249,10 +249,12 @@ processors:
...
       ignore_failure: true
   - set:
       field: _temp_.internal_zones
       copy_from: _conf.internal_zones
+      if: ctx._conf?.internal_zones != null
```
taylor-swanson, line 257:
```suggestion
      ignore_empty_value: true
```

packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/traffic.yml
leandrojmp: Is `ignore_empty_value` less expensive than the conditional in this case?

leandrojmp: Just tested here, adding `ignore_empty_value: true` will make the processor to be executed, it will just not populate the field because the value is empty, this still waste time on the processor, with the conditional there is no time spend on the processor.

leandrojmp: this is after removing the conditional and adding the `ignore_empty_value: true`, before it was zero as no event was passing through the processor, since we are trying to make this ingest pipeline faster, I think we need to use the conditional.

image

leandrojmp: hello @taylor-swanson adding `ignore_empty` does not solve the issue regarding performance as the processor is still executed wasting some time on it.

taylor-swanson: I'm not too familiar with how the benchmark works. It doesn't record the time for a processor if the conditional evaluates to false, but it is definitely not zero cost. The conditional is its own painless script that gets executed. The Elasticsearch team would know more on the performance implications of the conditional vs. `ignore_empty_value`.

My suggestion for using `ignore_empty_value` was not for performance, rather it was for readability/maintainability. Looking at the code for the set processor, it does have to do some work before it can stop once it realizes the source value is null. Whether or not this is more work than an if statement, I don't know.

Did you noticed a drop in the overall performance of the pipeline? As I mentioned, the conditional isn't going to be recorded against the processor, but it should show up in the overall performance of the pipeline (assuming the report isn't just accumulating the individual processor times).

leandrojmp: I'm having performance issues with this pipeline, have a support ticket that went nowhere, so I'm trying to see what can be change to improve performance, looking at the pipeline monitoring we saw that those 2 processors in the traffic pipeline were spending a considerable time doing nothing with all events being marked as failure, once we changed it to be conditionally executed, like all the other processors after that, we saw a small performance increase.

I also do not know if `ingore_empty_value` is more or less expensive than using a conditional, or if it makes no difference, but since I do not want this processor to be executed, I used the conditional.

leandrojmp: Would you be able to get any insight about what is less expensive in this case or if does not make any difference?

taylor-swanson: I can ask the elasticsearch team to see if they have any insights.

Functionally, they should accomplish the same thing. If the conditional appears to be slightly faster, I'm fine with going with that. Main thing is we don't want either processor failing, since that will have a much bigger impact on performance (and will also kill the pipeline midway through processing).

taylor-swanson: Had a chat with a dev from the elasticsearch team, `if` is indeed faster (for a variety of reasons, one of them being that the field lookup is much slower in the the processor, even for the `ignore_empty_value` part of it). In most cases, "it doesn't usually matter", but for panw and the amount of data it handles, it can definitely make a difference.

So with that figured out, if statement it is.

packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/traffic.yml
taylor-swanson: Disregard, if statement is going to be more performant.

taylor-swanson (approved):

packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/traffic.yml
leandrojmp: Thanks!

PR #14672, opened 2025-07-24
[microsoft_dnsserver] Adapt to support new ETW events.

marc-gr:
## Proposed commit message

Adapt to support new ETW events:
- Avoid doing unnecessary work when events are already rendered
- Add mappings for new fields
- Add more test samples

## Checklist

## Related issues

- Depends on https://github.com/elastic/beats/pull/44664


packages/microsoft_dnsserver/data_stream/analytical/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -307,6 +307,15 @@ processors:
         ctx.microsoft_dnsserver.analytical = renameKeys(ctx.microsoft_dnsserver.analytical, params)
 
 # ECS fields
+  - rename:
+      field: microsoft_dnsserver.analytical.extended_data.SID
+      target_field: user.id
+      ignore_missing: true
+      tag: rename_user_id
+  - remove:
```
chemamartinez, line 315: If the new `enable_property` option is added to the integration, it is possible that `extended_data` contains more fields, depending on the properties (https://learn.microsoft.com/en-us/windows/win32/api/evntrace/ns-evntrace-enable_trace_parameters).

Should we consider that as a valid possibility and support these extra fields?

chemamartinez (commented):
Based on new changes at the input, do you think we should add support for the new config options?
- enable_property
- buffer_size
- minimum_buffers
- maximum_buffers

that would mean to bump minimum required version to 9.2.0 (keeping this PR blocked until it is released).

marc-gr:
> Based on new changes at the input, do you think we should add support for the new config options?
> 
> * enable_property
> * buffer_size
> * minimum_buffers
> * maximum_buffers
> 
> that would mean to bump minimum required version to 9.2.0 (keeping this PR blocked until it is released).

I think this is something that can be pulled in a different PR, there is no reason to prevent users who do not upgrade from using the new pipelines. We can then do a separate PR just addint the new options and bumping the minimum version if that makes sense

packages/microsoft_dnsserver/data_stream/analytical/elasticsearch/ingest_pipeline/default.yml
chemamartinez: This doesn't apply if `enable_property` is not added in this PR as mentioned [here](https://github.com/elastic/integrations/pull/14672#issuecomment-3173604439).

chemamartinez (approved):

gogochan (approved):

PR #14679, opened 2025-07-24
[crowdstrike] Support various Tags formats in FDR data

chrisberkhout:
## Proposed commit message

```
[crowdstrike] Support various Tags formats in FDR data

Process formats seen in the falcon data stream's data, and elsewhere, to
prepare them for indexing as `keyword`.
```

## Checklist

### DIFF
--- a/packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
@@ -3049,14 +3049,14 @@ processors:
 
           } else if (ctx.crowdstrike.Tags instanceof Map) {
             for (def entry : ctx.crowdstrike.Tags.entrySet()) {
-              result.add(entry.getKey() + ": " + entry.getValue());
+              result.add(entry.getKey() + ":" + entry.getValue());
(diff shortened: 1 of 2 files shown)


packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -3030,11 +3030,41 @@ processors:
...
+            for (def part : parts) {
+              def trimmed = part.trim();
+              if (trimmed != "") {
+                result.add(trimmed);
+              }
+            }
+
+          } else if (ctx.crowdstrike.Tags instanceof Map) {
+            for (def entry : ctx.crowdstrike.Tags.entrySet()) {
+              result.add(entry.getKey() + ": " + entry.getValue());
```
efd6: Nit: Do we want the space here? ISTM this is a single token. Also below.

chrisberkhout: I went with the space because I thought it looked better.

But without is fine, maybe better. I'll remove it.

efd6 (approved):
Thanks

PR #14681, opened 2025-07-24
[TENABLE_IO] Use script to convert strings to numeric in asset datastream 

StacieClark-Elastic:
The tenable io API for fetching asset data changed the response definition to return decimal values in strings that are defined by the API as integers. The convert processor for string to long fails on decimals. We replaced the convert processor with a painless script for fields exposure_score which becomes tenable_io.asset.exposure_score and acr_score which becomes tenable_io.asset.acr_score.

## Proposed commit message
Fixes issue where string values that represent longs in fields acr_score and exposure score started to be returned from the tenable io asset API as strings with decimals. This caused the convert processor for strings to long to fail. Replaced the convert processor with a painless sctipt that handles strings as longs or decimals as well as primtive long, integer and double values. The issue is a due to the source changing returned values and violating the API contract

## Checklist


kcreddy (approved):

packages/tenable_io/changelog.yml
```diff
@@ -1,4 +1,11 @@
 # newer versions go on top
+- version: "4.2.1"
+  changes:
+    - description: |
```
efd6, line 4: We don't need to preserve new lines so
```suggestion
    - description: >-
```

packages/tenable_io/data_stream/asset/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -569,26 +569,62 @@ processors:
...
       target_field: tenable_io.asset.network_interfaces
       ignore_missing: true
-  - convert:
-      field: json.acr_score
-      target_field: tenable_io.asset.acr_score
-      type: long
-      if: ctx.json?.acr_score != ''
-      ignore_missing: true
+  - script:
+      if: ctx.json?.acr_score != null && ctx.json?.acr_score != ''
```
efd6, line 573:
```suggestion
      if: ctx.json?.acr_score != null && ctx.json.acr_score != ''
```

PR #14692, opened 2025-07-25
Mapping of Gigamon Metadata Attributes to ECS fields

apps-elastic-gigamon:
Proposed commit message :

This is an enhancement change for the mapping of Gigamon Metadata Attributes to corresponding ECS fields

## Checklist

## Related issues

- Supersedes #14665


packages/gigamon/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.7.0"
```
efd6:
```suggestion
- version: "1.8.0"
```

packages/gigamon/data_stream/ami/fields/fields.yml
```diff
@@ -107,7 +115,7 @@
     - name: dns_reply_code_value
       type: keyword
     - name: dns_response_time
-      type: double
+      type: float
```
efd6: Why is this being done? Same below for `double` to `float` changes.

packages/gigamon/data_stream/ami/fields/fields.yml
```diff
@@ -122,7 +130,7 @@
     - name: http_server_agent
       type: keyword
     - name: http_rtt
-      type: keyword
+      type: float
```
efd6: This is not OK. Please revert this.

packages/gigamon/data_stream/ami/_dev/test/pipeline/test-ami.json
efd6, line 1: Rather than mutating existing tests, please add new tests that cover the additions here.

packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -865,3 +1267,5 @@ on_failure:
       field: tags
       value: preserve_original_event
       allow_duplicates: false
+
+
```
efd6:
```suggestion
```

packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -827,6 +889,346 @@ processors:
...
+      if: ctx.gigamon?.ami?.dns_query != null
+  - set:
+      field: dns.question.registered_domain
+      tag: set_dns_question_registered_domain
+      copy_from: gigamon.ami.dns_name
+      if: ctx.gigamon?.ami?.dns_name != null
+  - append:
+      field: host.ip
+      tag: set_host_ip
+      value: "{{gigamon.ami.dns_host_addr}}"
```
efd6:
```suggestion
      value: "{{{gigamon.ami.dns_host_addr}}}"
```
(throughout)

packages/gigamon/changelog.yml
```diff
@@ -1,4 +1,9 @@
...
+  changes:
+    - description: Mapping of Gigamon Attributes to ECS fields
```
efd6: This should be a sentence; suggest "Improve mapping of Gigamon attributes to ECS fields."

packages/gigamon/data_stream/ami/fields/fields.yml
```diff
@@ -7,7 +7,7 @@
     - name: seq_num
       type: long
     - name: app_id
-      type: long
+      type: keyword
```
efd6, line 10: This is unfortunate, but cannot be changed without a breaking change. Please revert this.

packages/gigamon/data_stream/ami/fields/fields.yml
```diff
@@ -152,6 +168,22 @@
     # tcp_ fields
     - name: tcp_flags
       type: keyword
+    - name: tcp_rtt
+      type: float
```
efd6:
```suggestion
      type: double
```
(throughout)

efd6 (commented):
Please use the PR template. Can you provide a proposed commit message to repair that?

kcreddy (commented):
CI is failing on error: `Error: checking package failed: checking readme files are up-to-date failed: files do not match`

To fix this CI error, you will need to run `elastic-package build` and commit the generated `docs/README.md`.

packages/gigamon/changelog.yml
apps-elastic-gigamon: Changed the version to "1.8.0"

packages/gigamon/data_stream/ami/fields/fields.yml
apps-elastic-gigamon: Changed the data type to float for fields dns_response_time, http_rtt, tcp_rtt, and tcp_rtt_app as their values typically fall within the precision range of a float (i.e., up to 6 decimal places). Based on the comments, I have changed to double.

packages/gigamon/data_stream/ami/fields/fields.yml
apps-elastic-gigamon: For a recent use case from customer, we need to perform math operation for this attribute. So, changed the data type from keyword to float

packages/gigamon/data_stream/ami/_dev/test/pipeline/test-ami.json
apps-elastic-gigamon: Added new JSON records as per the comments

packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
apps-elastic-gigamon: Modified as per comments

packages/gigamon/changelog.yml
apps-elastic-gigamon: We are currently mapping the Gigamon attributes with the appropriate ECS fields, rather than making enhancements to the existing mappings.

packages/gigamon/data_stream/ami/fields/fields.yml
apps-elastic-gigamon: We have mapped service.id ECS field to this app_id. Hence the data type has to be changed. Also, in the backend we haven't done any math operations using this app_id

packages/gigamon/data_stream/ami/fields/fields.yml
apps-elastic-gigamon: Changed the data type to float for fields dns_response_time, http_rtt, tcp_rtt, and tcp_rtt_app as their values typically fall within the precision range of a float (i.e., up to 6 decimal places). Based on the comments, I have changed to double.

packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
apps-elastic-gigamon: Removed the extra line

packages/gigamon/data_stream/ami/fields/fields.yml
efd6: Doing this is a breaking change.

packages/gigamon/changelog.yml
efd6: The current text is not a valid English sentence since it has no verb. Please apply the change or something equivalent. The suggested change adds an appropriate verb since this is an improvement.

packages/gigamon/data_stream/ami/fields/fields.yml
efd6: Yes, it's unfortunate that the field was originally defined as a long rather than a keyword as is conventional for an ID field, but changing this from a long to a keyword is a breaking change, so making such a change would require a significant benefit to counter the cost of breaking people.

It _does_ need to be a keyword when it is copied to `event.id`, but that can either be done as a `convert` to a `string`, or just by a copy since numeric values can be mapped as keywords.

packages/gigamon/data_stream/ami/fields/fields.yml
kcreddy: @apps-elastic-gigamon, any datatype changes not within the same family, for example: [keyword family](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/keyword), is a breaking-change because it leads to mapping conflicts for existing users.

If this is must, you will need to update changelog version as `2.0.0` with `type: breaking-change`. Example: 
https://github.com/elastic/integrations/blob/ed4eea92060964cfdadf72f160b3525d0a1f416e/packages/ti_abusech/changelog.yml#L11-L17

packages/gigamon/data_stream/ami/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -827,6 +889,346 @@ processors:
...
+  - set:
+      field: source.ip
+      tag: set_source_ip
+      copy_from: gigamon.ami.src_ipv6
+      if: ctx.gigamon?.ami?.src_ipv6 != null
+  - set:
+      field: destination.ip
+      tag: set_destination_ip
+      copy_from: gigamon.ami.dst_ipv6
+      if: ctx.gigamon?.ami?.dst_ipv6 != null
```
kcreddy, line 1139: Much above, you are copying `gigamon.ami.dst_ip` into `destination.ip`. Later you are copying `gigamon.ami.dst_ipv6` into `destination.ip`.

By default, the [`set` processor ](https://www.elastic.co/docs/reference/enrich-processor/set-processor)overrides the field, you will end up having `destination.ip` always copying from `dst_ipv6` whenever it is not null.
Is this intended? Same for `source.ip`

apps-elastic-gigamon: In each JSON record, only one of the fields — either gigamon.ami.src_ip or gigamon.ami.src_ipv6 — will be present, not both. The same applies to gigamon.ami.dst_ip and gigamon.ami.dst_ipv6. Therefore, the destination.ip and source.ip fields will always be populated with either an IPv4 or IPv6 address, depending on which is available.

packages/gigamon/data_stream/ami/fields/fields.yml
apps-elastic-gigamon: Code changes done as per the comments.
- version: "2.0.0"
  changes:
    - description: Update mapping of Gigamon attributes to align with ECS fields
      type: breaking-change
      link: https://github.com/elastic/integrations/pull/14692

packages/gigamon/changelog.yml
apps-elastic-gigamon: Changed the description as per comments

packages/gigamon/data_stream/ami/fields/fields.yml
apps-elastic-gigamon: Updated the changelog.yml accordingly

kcreddy (approved):
LGTM, to be merged after @efd6 comments are addressed.

packages/gigamon/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.0.0"
+  changes:
+    - description: Update mapping of Gigamon attributes to align with ECS fields
```
efd6:
```suggestion
    - description: Update mapping of Gigamon attributes to align with ECS fields.
```

packages/gigamon/changelog.yml
```diff
@@ -1,4 +1,9 @@
...
+      type: breaking-change
```
efd6, line 5: I do not believe the balance of benefits/costs supports a breaking change here.

packages/gigamon/data_stream/ami/fields/fields.yml
efd6: I still do not believe this change is justified.

Issue #14684, opened 2025-07-24
[tenable_io] Main branch fix for asset API returning decimals instead of longs


StacieClark-Elastic:
Create fix described in parent issue for head of main branch

PR #14700, opened 2025-07-25
[TENABLE_IO] Use script to convert strings to numeric in asset datastream - backport

### LINKED ISSUE #14685: [tenable_io] backport fix for asset API returning decimals instead of longs
Create a backport between versions 3.10.1 and 4.0.0. 4.0.0 is a breaking change.
3.10.1 is the most recent version for 8.18.0.

StacieClark-Elastic:
The tenable io API for fetching asset data changed the response definition to return decimal values in strings that are defined by the API as integers. The convert processor for string to long fails on decimals. We replaced the convert processor with a painless script for fields exposure_score which becomes tenable_io.asset.exposure_score and acr_score which becomes tenable_io.asset.acr_score.

## Proposed commit message
Fixes issue where string values that represent longs in fields acr_score and exposure score started to be returned from the tenable io asset API as strings with decimals. This caused the convert processor for strings to long to fail. Replaced the convert processor with a painless sctipt that handles strings as longs or decimals as well as primtive long, integer and double values. The issue is a due to the source changing returned values and violating the API contract

## Checklist

## Related issues

- Closes #14685
- Relates #14684


packages/tenable_io/data_stream/asset/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -558,22 +558,58 @@ processors:
...
+          return;
+        }
+        if (ctx.json.acr_score instanceof int || ctx.json.acr_score instanceof long) {
+          ctx.tenable_io.asset.acr_score = (long) ctx.json.acr_score;
+          return;
+        }
+        if (ctx.json.acr_score instanceof double) {
+          ctx.tenable_io.asset.acr_score = (long) ctx.json.acr_score;
+          return;
+        }
```
andrewkroh, line 583: To cover all your bases I'd use `instanceof Number`.

```
        if (ctx.json.acr_score instanceof Number) {
          ctx.tenable_io.asset.acr_score = ((Number) ctx.json.acr_score).longValue();
          return;
        }
```

The instanceof checks with primitives looked odd to me because this is not done in Java.

And the number case is not tested by any of the test data.

StacieClark-Elastic: This is a backport. Let's keep it as it is so it matches what is on main.

andrewkroh (approved):

PR #14712, opened 2025-07-28
[Jolokia_input] Fix `path` field name

Linu-Elias:
## Proposed commit message
Changed `metrics_path` to `path` to align with Beats and other integrations.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/jolokia_input/manifest.yml
+++ b/packages/jolokia_input/manifest.yml
@@ -31,6 +31,8 @@ policy_templates:
         multi: true
         required: true
         show_user: true
+        description: |
+          Full hosts for Jolokia Input (http://host_address:jolokia_port).
         default:
           - http://127.0.0.1:8778
       - name: period
@@ -39,6 +41,7 @@ policy_templates:
         multi: false
         required: true
         show_user: true
+        description: The period for fetching metrics, e.g. every 1s, 1m, 1h.
         default: 10s
       - name: path
         type: text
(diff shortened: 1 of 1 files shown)


packages/jolokia_input/manifest.yml
```diff
@@ -40,9 +40,9 @@ policy_templates:
         required: true
         show_user: true
         default: 10s
-      - name: metrics_path
+      - name: path
```
ishleenk17, line 46: Can we add descriptions to all the config opetions of jolokia input package. 
So that its easy for the user to know what does each config mean

packages/jolokia_input/manifest.yml
```diff
@@ -31,6 +31,8 @@ policy_templates:
         multi: true
```
ishleenk17, line 31: Can we have multi host ?
If not lets make multi: false

Linu-Elias: I think we do support multiple hosts. for testing the same I added two hosts in the Integration, and I can see documents for both in Elastic (both show errors due to incorrect IPs, but that shouldn’t affect the validation of multi-host support).
 Also, in the Beats code, the host field is defined as a `[]string`

ishleenk17:
Does Jolokia has support for authentication ?
Do we have some authentication mechanism for it ?

Linu-Elias:
> Does Jolokia has support for authentication ? Do we have some authentication mechanism for it ?

yes, we do support basic auth and bearer token authentication mechanisms

ishleenk17 (approved):

PR #14717, opened 2025-07-28
Change TargetResponseTime.avg type from `long` to `double`

### LINKED ISSUE #14715: [AWS]: Application ELB TartgetResponseTime metric type incorrect
### Integration Name

AWS [aws]

### Dataset Name

aws.elb_metrics

### Integration Version

3.13.2

### Agent Version

8.12.2

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.18.0

Ubuntu 20.04 LTS

### Software/API Version

AWS API

### Error Message

_No response_

### Event Original

_No response_

### What did you do?

I am using the AWS integration for pulling in ELB metrics and visualizing them with the [Metrics AWS] ALB Overview dashboard.

### What did you see?

When looking at the average response time in AWS Cloudwatch, the values being reported are subsecond values ( ex. 0.011 ).

Image

But in the [Metrics AWS] ALB Overview dashboard, the TargetResponseTime is always 0. ( See below )

Image

When the metrics goes above 1, it does get reported as "1" in the dashboard and documents that are ingested.

Image

I expected to see data similar to what is reported by Cloudwatch Metrics.

### Anything else?

I found the [pull request](https://github.com/elastic/integrations/pull/10181) where this feature was added. The PR description has test output that correctly shows the `TargetResponseTime` being returned as a float. In the PR comments, there was a request to change the type from `double` to `long` which would only report whole numbers. I think we just need to change this back to `double` and we should be good.

jkomara:
## Proposed commit message

Change TargetResponseTime.avg type from `long` to `double`

The values reported by AWS are floats and we lose a lot of valuable information by processing this as an int. This change will give us the granularity that we are looking for in dashboards and alerts.

## Checklist

## Author's Checklist 

- [ ] 

## Related issues

Closes: #14715

### DIFF
--- a/packages/aws/changelog.yml
+++ b/packages/aws/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "8.0.0"
+  changes:
+    - description: Change TargetResponseTime.avg from `long` to `double`.
+      type: breaking-change
+      link: https://github.com/elastic/integrations/pull/14717
 - version: "7.4.0"
   changes:
     - description: |
--- a/packages/aws/changelog.yml
+++ b/packages/aws/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "8.0.0"
+  changes:
+    - description: Change TargetResponseTime.avg from `long` to `double`.
+      type: breaking-change
+      link: https://github.com/elastic/integrations/pull/14717
 - version: "7.1.1"
   changes:
     - description: Add `data_stream.namespace` to the Amazon Inspector vulnerability latest transform's unique key so findings are tracked per namespace, preventing findings ingested into non-default namespaces from being dropped or conflated in the latest index. Bump transform's destination suffix to `-v2`.
--- a/packages/aws/manifest.yml
+++ b/packages/aws/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.6.1
 name: aws
 title: AWS
-version: 7.4.0
+version: 8.0.0
 description: Collect logs and metrics from Amazon Web Services (AWS) with Elastic Agent.
 type: integration
 categories:
--- a/packages/aws/manifest.yml
+++ b/packages/aws/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.6.1
 name: aws
 title: AWS
-version: 7.1.1
+version: 8.0.0
 description: Collect logs and metrics from Amazon Web Services (AWS) with Elastic Agent.
 type: integration
 categories:
--- a/packages/aws/data_stream/elb_metrics/fields/fields.yml
+++ b/packages/aws/data_stream/elb_metrics/fields/fields.yml
@@ -162,7 +162,7 @@
               metric_type: gauge
               description: The number of load balancer capacity units (LCU) used by your load balancer.
             - name: TargetResponseTime.avg
-              type: long
+              type: double
               metric_type: gauge
               unit: s
               description: The time elapsed after the request leaves the load balancer until the target starts to send the response headers.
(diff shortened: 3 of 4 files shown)


packages/aws/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.13.3"
+  changes:
+    - description: Remove Include Linked Accounts config option from AWS Health.
```
efd6: This is not right.

packages/aws/changelog.yml
```diff
@@ -1,4 +1,9 @@
...
+      type: bugfix
```
efd6, line 443: I'm not convinced that this is a bugfix; it might be, but it could also be deemed an enhancement. In either case, this is likely a breaking change.

packages/aws/changelog.yml
jkomara: Thank you for pointing that out. I updated the description.

packages/aws/changelog.yml
jkomara: I am not familiar with this flow so if I misclassified something I apologize. It was originally working as intended until the type was changed to `long` so the charts aren't very useful. If that falls more into an enhancement, is there a way that I could change this?

This is my first time I have submitted a PR here so I am not sure if I am approaching everything the right way. It looked like a straightforward change so I took a stab at it. Happy to make any changes needed here or close it out and let someone else take a look.

kaiyan-sheng: @efd6 I agree it's probably a breaking change but I think we should get it fixed. What's the process for this kind of breaking change? 
I thought about adding a new field called `TargetResponseTime.avg_double` to avoid making it a breaking change. But that's just not clean...
Maybe we can bump the version? and kibana version?

efd6: Bumping the `kibana.version` does not protect users from being impacted by this. If this is worth doing, then it can be marked as "breaking-change", bumping the version to 4.0.0.

jkomara: As a user, I feel that it is a bad user experience to see everything reporting properly in AWS and then Kibana showing 0. Technically we can still alert and see large fluctuations in the response time, but it is not ideal. I don't know who gets to decide if this is worth doing, but I would definitely vote for getting this change in. Thank you both for taking a look. 🙇

efd6: Please mark the change as a breaking change.

jkomara: Sorry I missed your comment. How do I mark the change as a breaking change? I added the label but not sure if I need to do anything else.

efd6: Instead of `type: bugfix` it should be `type: breaking-change`. The version should be "4.0.0".

jkomara: Thank you. That has been updated. This PR only addresses one of the metrics. Should someone else go and review all of them so that we can fix anything else that may be typed incorrectly so we aren't potentially making multiple breaking changes?

packages/aws/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 3.3.2
 name: aws
 title: AWS
-version: 3.13.2
+version: 3.13.3
```
ishleenk17: Incorrect,should be 4.0.0

packages/aws/changelog.yml
ishleenk17: > The version should be "4.0.0"

@efd6 : Do we usually follow this practice of incrementing the major version in these cases ?
I see both type of instances for breaking change. Either Minor or major version is being changed.

efd6: We follow semver, so if we are not v0, then major is bumped for a breaking change. If we are in v0, then either minor (for enhancements), or patch (for bugfixes) since any version change in v0 is allowed to be a breaking change. See also [this](https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#versioning-changes).

jkomara:
This is still an issue.

jkomara:
This is still an issue.

jkomara:
I can't reopen this issue.

mrg-elastic (approved):

jkomara:
Thank you for approving this. Is there anything I need to do to get this merged? I didn't know if someone else was going to drive this or if I needed to do something so I wanted to ask.

efd6:
@jkomara Please resolve the conflicts.

ishleenk17 (approved):
Please resolve the conflicts. Else looks good

jkomara:
@efd6 / @ishleenk17 this should be good to go.

PR #14744, opened 2025-07-30
fix(ess_billing): Improve reliability and prevent API errors

### LINKED ISSUE #14743: [ESS Billing]: CEL Configuration Issue
### Integration Name

Elasticsearch Service Billing [ess_billing]

### Dataset Name

ess_billing.billing

### Integration Version

1.4.2

### Agent Version

9.0.1

### Agent Output Type

elasticsearch

### Elasticsearch Version

Elastic Cloud

Agentless

### Software/API Version

_No response_

### Error Message

These are the logs for an execution that fails. It starts, executes 1000 times very quickly, then is stopped by CEL when the execution budget is exhausted.

| message                                    | error.message                                                                                                                                                                                                          | count |
|--------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------|
| process repeated request                   |                                                                                                                                                                                                                        | 1     |
| single event object returned by evaluation | GET  https://billing.elastic-cloud.com/api/v2/billing/organizations/ <org> /costs/instances?from=2025-07-18T00%3A00%3A00Z&to=2025-07-19T00%3A00%3A00Z: 403: User does not have 'billing-costs:get: <org> ' permissions | 1000  |
| exceeding maximum number of CEL executions |                                                                                                                                                                                                                        | 1     |

### Event Original

N/A

### What did you do?

Configured the ESS Billing integration with an API key that is not authorized to read billing data.

### What did you see?

The input makes 1000 API requests to the billing API over about 20 seconds.

The only thing that stops the loop is the `max_executions` protections built into the CEL input. The problem appears to be that on a non-200 response, the program sets `want_more` to true. I believe the program should set `want_more` to false for these non-retryable error codes. The input will attempt to retry at the next periodic interval.


It should stop making requests after an HTTP 403 error. It will try again on the next periodic interval.

### Anything else?

The relevant source code is at https://github.com/elastic/integrations/blob/9510edafbedd269436cb91dfdac142cf2af81b9a/packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs#L114.

Note that the CEL input automatically retries requests for error codes that are retryable (default is 5 retries). 

------

You can see in this image that the first two executions behave like we want because `want_more` evaluates to `false` on failure because the `to` is within 24h from `now`. But after the first 24h it goes into API DoS mode.

<img width="1394" height="215" alt="Image" src="https://github.com/user-attachments/assets/5ee6cfe0-3e78-41bc-a594-588ad2043d62" />

### LINKED ISSUE #14755: [ESS Billing] Invalid 'from' query param
### Integration Name

Elasticsearch Service Billing [ess_billing]

### Dataset Name

ess_billing.billing

### Integration Version

Unknown

### Agent Version

8.18.3

### Agent Output Type

elasticsearch

### Elasticsearch Version

Unknown

Agentless

### Software/API Version

_No response_

### Error Message

`GET https://billing.elastic-cloud.com/api/v2/billing/organizations/1111/costs/instances?from=2001-08-04T00%3A00%3A00Z&to=2001-08-05T00%3A00%3A00Z: {"error": "invalid_input", "message": "An error occurred with input: {'from': '2001-08-04T00:00:00Z', 'to': '2001-08-05T00:00:00Z'}", "detail": {"messages": {"_schema": ["'from' must not be earlier than 2021-01-01T00:00:00+00:00"]}}, "success": false}`

### Event Original

_No response_

### What did you do?

N/A (I only observed the logs.)

### What did you see?

Repeated Bad Request errors every 24 hours.

No errors. Or at most one error, followed by a recovery.

### Anything else?

I'm not how it was possible to get into a state where the `from` values have year 2001 dates. Either we need to ensure that it cannot happen or that if this occurs that the integration is able to recover.

I think the cause would be either be a problem with cursor and making it into a timestamp OR the user configured a really large `lookbehind` value.

-----

The problem is occurring for 8 users. And because of #14743 it happens 1000 times per user.

Image

andrewkroh:
NOTE: A commit-by-commit review is recommended because of the celfmt changes.

## Proposed commit message

```
fix(ess_billing): Improve reliability and prevent API errors

This commit introduces several improvements to the ess_billing integration to
enhance its reliability and prevent common API errors.

- To prevent repeated failing requests on non-200 HTTP status codes, the CEL
  program now sets want_more to false. This allows the input to retry at the
  next periodic interval instead of exhausting the execution budget.
- A validation rule has been added to enforce a minimum 'from' date of
  2021-01-01. This clamps the calculated timestamp to the API's minimum allowed
  value, preventing 'Bad Request' errors caused by lookbehind configurations
  creating excessively early dates. This uses the max() function which requires
  Elastic Agent 8.18 or greater so the Kibana constraint was raised as a proxy.
- All instances of the now() function in the CEL program have been replaced
  with the now variable.  This ensures a stable time reference throughout a
  single execution, leading to more consistent and predictable time-based
  calculations.

  A system test has been added to cover these scenarios.

  Fixes #14743
  Fixes #14755
```

## Checklist

`elastic-package -C packages/ess_billing test system -v` 

## Related issues

- Fixes #14743
- Fixes #14755

### DIFF
--- a/packages/ess_billing/data_stream/billing/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/ess_billing/data_stream/billing/elasticsearch/ingest_pipeline/default.yml
@@ -1,6 +1,10 @@
 ---
 description: "Ingest pipeline for ESS billing data"
 processors:
+  # NOTE(2025-08-15): This 'drop' processor is no longer necessary because
+  # there is a Beat processor to drop events with 'fake: true'. It is kept here
+  # to allow users a chance to upgrade their Agent configurations first, given
+  # that package upgrades are not atomic (pipelines upgrade before agents).
   - drop:
       if: "ctx.ess?.billing == null && ctx.error?.message == null"
   - remove:


efd6:
This all seems sane, though it looks like the system test needs attention since it's failing. I _would_ be nice to understand where the invalid timestamp in the bugged case came from, but this will fix it.

andrewkroh:
Looks like the CI problem is my usage of `max` with 8.15.0.

`fleet-server-1  | {"log.level":"info","@timestamp":"2025-07-31T03:37:26.000Z","message":"applying new components data","component":{"binary":"fleet-server","dataset":"elastic_agent.fleet_server","id":"fleet-server-default","type":"fleet-server"},"log":{"source":"fleet-server-default"},"service.name":"fleet-server","req.Components":[{"id":"cel-default","message":"Healthy: communicating with pid '316'","status":"HEALTHY","type":"cel","units":[{"id":"cel-default-cel-ESS Billing-4f94630c-28ef-4096-a8ba-2306872c53e2","message":"failed to check program: failed compilation: ERROR: <input>:11:8: found no matching overload for 'max' applied to '(timestamp, timestamp)'\n |     max(from, timestamp(\"2021-01-01T00:00:00Z\"))\n | .......^ accessing config","payload":{"streams":{"cel-ess_billing.billing-4f94630c-28ef-4096-a8ba-2306872c53e2":{"error":"","status":"STARTING"}}},"status":"FAILED","type":"input"},{"id":"cel-default","message":"Healthy","status":"HEALTHY","type":"output"}]}],"service.type":"fleet-server","http.request.id":"01K1F8WQZJQW54PFJT18VCHVNX","server.address":"","fleet.agent.id":"64fa78d6-ba93-421e-b6bf-f44e31225444","fleet.access.apikey.id":"wx2JXpgBcJYJrjvG18FE","@timestamp":"2025-07-31T03:37:26Z","ecs.version":"1.6.0","ecs.version":"1.6.0"}`

andrewkroh:
@3kt would you be able to review my changes (since you have contributed and are a CODEOWNER)? Thanks!

packages/ess_billing/data_stream/billing/_dev/deploy/docker/files/stream-config.yml
3kt, line 1: I didn't know we could write tests like this <3
Great stuff, thanks for adding it!

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
```diff
@@ -83,22 +85,22 @@ program: |-
               "last_to": req.to,
             },
             // Are we more than 1 day behind?
-            "want_more": req.to < now() - duration("24h"),
+            "want_more": req.to < now - duration("24h"),
```
3kt, line 88: Out of curiosity, what's the difference between `now()` and `now` (apart from the obvious that one is a function and the other a variable)?

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
```diff
@@ -119,7 +121,7 @@ program: |-
           "cursor": {
             "last_to": req.to,
           },
-          "want_more": req.to < now() - duration("24h"),
+          "want_more": false,
```
3kt, line 121: Why would we stop there?
Let's say we have a 429 or 503 - shouldn't we retry if we're not caught up?

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
```diff
@@ -139,7 +141,8 @@ tags:
 publisher_pipeline.disable_host: true
 {{/contains}}
 
-{{#if processors}}
 processors:
+- drop_event.when.equals.fake: true
```
3kt, line 142: The `drop` processor then needs to be removed from the ingest pipeline as well:

```
  - drop:
      if: "ctx.ess?.billing == null && ctx.error?.message == null"
```

Its only purpose was to drop this fake event, but I got heavy-handed and stated "anything that doesn't have an error or the `billing` namespace should be dropped".

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
andrewkroh: https://pkg.go.dev/github.com/elastic/mito/lib#Time

`now` is global var set at execution time. `now()` is function that returns the current time. By invoking `now()` you could have a different view of the current time in all of the places that it is used.

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
andrewkroh: It can be removed, but to be safe it should probably be removed at a later time to give Agents a chance to upgrade and get the local beat drop_event processor. The package upgrade and the agent policy upgrade are not an atomic operation. If the package upgrades, but the related Agent policies are not immediately upgraded then this could offer a window where `fake: true` events sneak through.

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
andrewkroh: This is the crux of the problem. For instance, by returning `want_more: true` on HTTP 500, it will trigger 999 more CEL program executions in just a few seconds (assuming the server keeps responding with non-200 status codes). There is already a default built-in retry request mechanism that attempts the request up to 5 times.

My goal is to stop the program and have the next iteration retry the same request. However, it has been some time since I made this change, so I need to revisit it to ensure the cursor isn't advanced. I might need to remove the `cursor` from this failure case to prevent advancing it. I'll test this and provide an update.

3kt: The other issue is that we'll wait 24 hours before retrying I guess? 24h being the execution schedule.

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
3kt: This is a very good point actually, I forgot the local and remote parts aren't necessarily atomically tied.
Maybe worth commenting in the pipeline that this should be removed at a later date then?

andrewkroh: Comment added in f9c66e2

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
andrewkroh: To prevent the cursor from updated on HTTP errors, it was necessary to remove the `cursor` from the returned data. See 650ff09

PR #14760, opened 2025-07-31
[proofpoint_on_demand] add keep_alive and keep_alive_interval config options

ShourieG:
# Type of change
- Enhancement

## Proposed commit message
**WHAT:** Added support for  "keep_alive" and  "keep_alive_interval" configuration options at the data stream level.
**WHY:** This will help in scenarios where low volume connections run into idle timeouts. This option has been added at the data stream level and not the package level because it is meant to be used as a last resort option for a specific data stream in scenarios when intermittent timeouts/disconnects keep occurring.

**Should only be merged after release of 8.19 and 9.1**

**Commit Message:**

```
proofpoint_on_demand: add "keep_alive" and "keep_alive_interval" configuration options

This will help in cases where low volume connections run into idle
timeouts.

It has been added at the data stream level rather than the package level
as it is meant to be used as a last resort option for a specific data
stream in cases when intermittent timeouts/disconnects keep occurring.
```

## Tests
System tests have been disabled atm, please follow issue: https://github.com/elastic/integrations/issues/14854

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/integrations/issues/14761


packages/proofpoint_on_demand/manifest.yml
```diff
@@ -1,14 +1,14 @@
...
-version: "1.6.0"
+version: "1.7.0"
 description: Collect logs from Proofpoint On Demand with Elastic Agent.
 type: integration
 categories:
   - security
 conditions:
   kibana:
-    version: "^8.16.3 || ^8.17.1 || ^9.0.0"
+    version: "^8.19.0 || ^9.1.0"
```
ShourieG, line 11: this is because of the feature requirement in beats.

packages/proofpoint_on_demand/data_stream/audit/manifest.yml
```diff
@@ -6,6 +6,22 @@ streams:
...
+        show_user: false
+        default: false
+      - name: keep_alive_interval
+        type: text
+        title: Keep Alive Interval
+        description: The interval at which the agent will send a keep alive message to the websocket endpoint. This is a time duration value. Examples, 30s, 1m, 1h.
+        multi: false
+        required: false
+        show_user: false
+        default: 30s
```
efd6, line 24: Could we roll this into one configuration knob, `keep_alive_interval`? and then set enable to true in cases where it is non zero?

efd6 (commented):
Suggest the following commit message (pending decision on user-facing option simplification).
```
proofpoint_on_demand: add "keep_alive" and "keep_alive_interval" configuration options

This will help in cases where low volume connections run into idle
timeouts.

It has been added at the data stream level rather than the package level
as it is meant to be used as a last resort option for a specific data
stream in cases when intermittent timeouts/disconnects keep occurring.
```

packages/proofpoint_on_demand/data_stream/audit/manifest.yml
ShourieG: We could do that, but from a config standpoint it looses out on clarity. I'd rather have a clear distinction between enablement and duration. I know this creates an extra hop for users but it also provides extra guidance from a users perspective so they do not misconfigure things.  Also having a **`0` value**  deciding enablement of a config option wont be a good UI exp.

ShourieG: If our templates allowed a feature where we could have a toggle and a text box in a single attribute, where the text box would become configurable once the toggle was enabled, this would be a different story.

efd6: OK

efd6 (approved):

efd6:
This is still failing with health check updates:
```
Unit state changed streaming-default-websocket-proofpoint_on_demand-4a53b480-c799-43d0-bcbe-86df104ee0ba (HEALTHY->DEGRADED): websocket connection encountered an error: websocket: close 1006 (abnormal closure): unexpected EOF
```

ShourieG:
> This is still failing with health check updates:
> 
> ```
> Unit state changed streaming-default-websocket-proofpoint_on_demand-4a53b480-c799-43d0-bcbe-86df104ee0ba (HEALTHY->DEGRADED): websocket connection encountered an error: websocket: close 1006 (abnormal closure): unexpected EOF
> ```

Locally tests seem to be passing, let me check by updating the branch. Maybe some infra or container issue caused the websocket server to shutdown in the middle of execution.

ShourieG:
@efd6, I've temporarily disabled the system tests until we resolve the metrics state degraded issue. I believe we can merge the updates to address the SDH meanwhile. The issue seems to be related to the test environment and our mock server but further investigation is required to find the root cause.

ShourieG:
> Please file an issue to fix the test server and link that in the skip.

I've already linked the issue that's tracking the problems, if we require a fix in the mock server or something in docker container, that is yet not derived and I'll be investigating that today.

PR #14762, opened 2025-07-31
crowdstrike: Remove _id-Based Deduplication when `Data Deduplication` option is disabled in fdr data stream

kcreddy:
## Proposed commit message
```
AWS S3 input does "_id-Based Deduplication"[1] and 
generates "_id" by default.
"Data Deduplication" integration setting was introduced 
to allow users to deduplicate data at the cost of indexing[2].
When users disable this setting, the "_id" generated by the 
input should also be deleted. This PR does that.

[1]: https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-aws-s3#_document_id_generation
[2]: https://github.com/elastic/integrations/pull/13109
```

## Checklist


packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -422,6 +422,18 @@ processors:
...
+  ## When "Data Deduplication" is not enabled, this field must be removed.
+  ## https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-aws-s3#_document_id_generation
+  - remove:
+      field: _id
+      tag: remove_id_based_deduplication
+      description: >
+        When data deduplication is disabled, even the _id-Based Deduplication
+        needs to be removed.
+      if: ctx._conf?.enable_deduplication == false
+      ignore_missing: true
```
kcreddy, line 435: I am inclined to call this a `breaking-change` because users who currently disabled `Data Deduplication` option still experience benefit of deduplication from `_id-Based Deduplication` of AWS s3 input. 
Removing this based on `Data Deduplication` default value (`false`) could increase duplicates for users using default settings. 

@andrewkroh, please let me know your thoughts.

andrewkroh: I think you are totally right, good catch. This has the potential to break users because the default is off when it used to always happen in the input.

kcreddy: Changelog updated to `breaking-change` with modified description in [df08323](https://github.com/elastic/integrations/pull/14762/commits/df083239b78042131d8ae2f8ce2492b5eaf3fff9).

packages/crowdstrike/data_stream/fdr/manifest.yml
```diff
@@ -106,7 +106,7 @@ streams:
         title: Enable Data Deduplication
         required: true
         show_user: false
-        description: If data deduplication is enabled, it ensures no duplicate events are received. A [fingerprint processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/fingerprint-processor.html) is used to calculate hash from a set of crowdstrike fields that uniquely defines the event.
+        description: If data deduplication is enabled, it ensures no duplicate events are received by generating a custom document `_id` for each event using [fingerprint processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/fingerprint-processor.html), rather than relying on Elasticsearch to automatically generate one. If the option is disabled, the AWS-S3 input's [_id-Based Deduplication](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-aws-s3#_document_id_generation) feature is also disabled, thus relying on Elasticsearch to automatically generate document `_id`.
```
chrisberkhout:
```suggestion
        description: If data deduplication is enabled, it ensures that no duplicate events are indexed. This is achieved by generating an `_id` value based on the content of each event using the [fingerprint processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/fingerprint-processor.html). If the option is disabled, the AWS-S3 input's [_id-Based Deduplication](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-aws-s3#_document_id_generation) feature is also disabled and every received event will be indexed, with a unique `_id` generated by Elasticsearch.
```

chrisberkhout: This is hard to explain. Here's another attempt, de-emphasizing the AWS-S3 input's functionality, since that isn't used in either case.

```suggestion
        description: If data deduplication is enabled, it ensures that no duplicate events are indexed. This is achieved by generating an `_id` value based on the content of each event using the [fingerprint processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/fingerprint-processor.html). If the option is disabled, every received event will be indexed with a unique `_id` generated by Elasticsearch (overriding the the AWS-S3 input's [_id-Based Deduplication](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-aws-s3#_document_id_generation) feature).
```

chrisberkhout (approved):
Changes looks good. Some suggestions for alternative variable descriptions you can consider.

packages/crowdstrike/data_stream/fdr/manifest.yml
kcreddy: Going with the last description in [f9f1d37](https://github.com/elastic/integrations/pull/14762/commits/f9f1d37e05d922899b7632aa7cb3df0300287039).

packages/crowdstrike/changelog.yml
```diff
@@ -1,4 +1,13 @@
 # newer versions go on top
+- version: "2.0.0"
+  changes:
+    - description: | 
+        Remove _id-Based Deduplication when `Data Deduplication` option is disabled in fdr data stream.
+        If duplicates are not preferred, enable the integration setting `Data Deduplication`.
+        Disabling `Data Deduplication` option makes Elasticsearch automatically generate document `_id`, 
+        but has potential for creating duplicates at a faster indexing rate.
```
andrewkroh: I'm worried that users won't understand the nuance of that first sentence. I tried revising it to add an emphasis on the trade-off and also be direct about what users need to do to retain their old behavior. WDYT?

> Data deduplication is now disabled by default for the FDR data stream when configured with the aws-s3 input. Previously, the FDR data stream automatically handled deduplication by computing an Elasticsearch document `_id` using the aws-s3 input. To prevent duplicate documents, you must now explicitly enable the Data Deduplication setting. While enabling this setting prevents duplicates, it may result in a lower indexing rate because Elasticsearch must check for existing documents before indexing.

kcreddy: Looks good to me, yes this is clearer. Updated in [524d97b](https://github.com/elastic/integrations/pull/14762/commits/524d97bba6c07fdc88f5e7d42680a058cddde2a1)

PR #14764, opened 2025-07-31
Write System Tests for RabbitMQ

giorgi-imerlishvili-elastic:
## Proposed commit message
Write system tests for RabbitMQ

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


ishleenk17:
@giorgi-imerlishvili-elastic : Are there no sample json files generated here ?

giorgi-imerlishvili-elastic:
> @giorgi-imerlishvili-elastic : Are there no sample json files generated here ?

No there is no sample json files. Do we need those?

giorgi-imerlishvili-elastic:
> > @giorgi-imerlishvili-elastic : Are there no sample json files generated here ?
> 
> No there is no sample json files. Do we need those?

Just updated PR with newly generated: sample_event.json for each data stream
The command used for generating those files: `elastic-package test system --generate`

packages/rabbitmq/_dev/deploy/docker/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+# Copy the Debian stretch APT fix script to the container's /usr/local/bin directory
+COPY fix_debian_stretch_apt.sh /usr/local/bin/fix_debian_stretch_apt.sh
+
+# Make fix_debian_stretch_apt.sh executable and run it to fix apt issues
+# Make the Debian stretch APT fix script executable and run it to fix APT issues for Debian 9
+RUN chmod +x /usr/local/bin/fix_debian_stretch_apt.sh && \
+    /usr/local/bin/fix_debian_stretch_apt.sh
+# Make the entrypoint and async consumer scripts executable
+RUN chmod +x /usr/local/bin/entrypoint.sh
+RUN chmod +x /usr/local/bin/async_rabbitmq_consumer.sh
```
mykola-elastic: why not setting `+x` on the files in the repo? so that we don't have these `RUN`s

packages/rabbitmq/_dev/deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+# This docker-compose file defines the services required for the RabbitMQ integration test environment
+version: '2.3'
```
mykola-elastic: `version` is obsolete: https://docs.docker.com/reference/compose-file/version-and-name/

mykola-elastic: it still remains a lot in the code, but maybe we should not use it in the new files we create. What do you think?

packages/rabbitmq/_dev/deploy/docker/fix_debian_stretch_apt.sh
mykola-elastic, line 1: @lucian-ioan I think you did fixes like this in other Dockerfiles?

packages/rabbitmq/data_stream/exchange/fields/fields.yml
```diff
@@ -39,3 +39,8 @@
       metric_type: gauge
       description: |
         How much the exchange publish-out count has changed per second in the most recent sampling interval.
+    - name: type
+      type: keyword
+      description: Exchange type
+      # Holds information about the type of exchange.
```
mykola-elastic:
```suggestion
```
`description` might be enough

packages/rabbitmq/data_stream/queue/fields/package-fields.yml
```diff
@@ -7,3 +7,14 @@
...
+      type: group
+      # Reason for adding is to pass system tests.
+      fields:
+        - name: name
+          description: Node name
+          # Reason for adding is to pass system tests.
+          fields:
+            - name: text
+              type: text
+              description: Full text of node name
```
mykola-elastic: to be consistent, please add a `.` in the end of description.

Also, this thing gets to the README (see line 188). Does it look good enough?

mykola-elastic: @giorgi-imerlishvili-elastic can you explain a bit more why `node.name.text` needed?

packages/rabbitmq/_dev/deploy/docker/fix_debian_stretch_apt.sh
lucian-ioan: @mykola-elastic Can't say I had with scripting. In my case I just had to fetch the images from the archive in the Dockerfile.

PR: https://github.com/elastic/integrations/pull/14687

ishleenk17:
@giorgi-imerlishvili-elastic : I see that there are multiple scripts being used. Also the debian script looks overly complex. Could you pls add explanations to the utility of each of these ?

packages/rabbitmq/_dev/deploy/docker/entrypoint.d/simulate_queue_connection.py
ghost, line 1: is this python script necessary for the system test? without simulating a connection and sending messages some metrics won't be generated?

giorgi-imerlishvili-elastic: I haven't found any way around, system test wasn't getting hit while testing connection data stream, for that reason I created that python script

packages/rabbitmq/data_stream/queue/fields/package-fields.yml
giorgi-imerlishvili-elastic: While running system tests I discovered it was failing and were logging error something like:
 `rabbitmq.node.name and node.name.text is not defined` after investigation I found that it wasn't really declared in queue data stream  for that reason I added those fields
 same goes for connection data stream

giorgi-imerlishvili-elastic:
> @giorgi-imerlishvili-elastic : I see that there are multiple scripts being used. Also the debian script looks overly complex. Could you pls add explanations to the utility of each of these ?

sure

- packages/rabbitmq/_dev/deploy/docker/entrypoint.d/entrypoint.sh
is wrapper of original entrypoint.sh, the script before running original entrypoint.sh runs connection simulation script
- packages/rabbitmq/_dev/deploy/docker/entrypoint.d/async_rabbitmq_consumer.sh
wrapper of connection simulator script. Before running the script first it checks if Python is installed and also waits until RabbitMQ starts
- packages/rabbitmq/_dev/deploy/docker/entrypoint.d/simulate_queue_connection.py
Connection simulation python script. It creates queue publishes number of messages and consumes as well this way it simulates connection to the server.

packages/rabbitmq/_dev/deploy/docker/fix_debian_stretch_apt.sh
giorgi-imerlishvili-elastic: @lucian-ioan it little bit different case. I don't think just coping that solution will work here.
Here I have two versions of Debian where only in the old version I'm doing that repo replacement stuff, so here I needed to have those OS version checks which is written in the shell script.

packages/rabbitmq/_dev/deploy/docker/docker-compose.yml
giorgi-imerlishvili-elastic: agree, if it's obsolete it's better not to use in at least new files 🙌

packages/rabbitmq/_dev/deploy/docker/entrypoint.d/async_rabbitmq_consumer.sh
```diff
@@ -0,0 +1,26 @@
...
+set -e
+
+# This script waits for RabbitMQ to be fully started and then runs a Python consumer script
+# to simulate a long-lived connection. It is compatible with RabbitMQ 3.7.4 and 4.x.
+
+# Wait for RabbitMQ to be fully started (compatible with RabbitMQ 3.7.4 and 4.x)
+# The loop checks the output of 'rabbitmqctl status' for several possible indicators of readiness.
+# It breaks out of the loop once RabbitMQ is ready, otherwise it waits and retries.
+while true; do
+  if gosu rabbitmq rabbitmqctl status 2>&1 | grep -Eiq 'pid,|Pid|Status of node|RabbitMQ'; then
```
shmsr:
```suggestion
  if gosu rabbitmq rabbitmqctl status 2>&1 | grep -Eiq 'pid|Pid|Status of node|RabbitMQ'; then
```

Is the comma extra here; typo?

packages/rabbitmq/_dev/deploy/docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+      gosu rabbitmq rabbitmqctl status >/dev/null 2>&1
+
+# Copy entrypoint and helper scripts
+# Copy the main entrypoint script to the container's /usr/local/bin directory
+COPY entrypoint.d/entrypoint.sh /usr/local/bin/entrypoint.sh
+# Copy the async RabbitMQ consumer script to the container's /usr/local/bin directory
+COPY entrypoint.d/async_rabbitmq_consumer.sh /usr/local/bin/async_rabbitmq_consumer.sh
+# Copy the Debian stretch APT fix script to the container's /usr/local/bin directory
+COPY fix_debian_stretch_apt.sh /usr/local/bin/fix_debian_stretch_apt.sh
+
```
shmsr: Too many comments; each step is self explanatory imo.

packages/rabbitmq/_dev/deploy/docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# run it to fix APT issues for Debian 9
+RUN /usr/local/bin/fix_debian_stretch_apt.sh
+
+# Install Python 3 and pika for consumer simulation
+# Install Python 3 and the pika library for simulating a RabbitMQ consumer
```
shmsr: Keep 1 comment.

packages/rabbitmq/_dev/deploy/docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+      gosu rabbitmq rabbitmqctl node_health_check >/dev/null 2>&1 || \
+      gosu rabbitmq rabbitmqctl status >/dev/null 2>&1
+
+# Copy entrypoint and helper scripts
+# Copy the main entrypoint script to the container's /usr/local/bin directory
+COPY entrypoint.d/entrypoint.sh /usr/local/bin/entrypoint.sh
+# Copy the async RabbitMQ consumer script to the container's /usr/local/bin directory
+COPY entrypoint.d/async_rabbitmq_consumer.sh /usr/local/bin/async_rabbitmq_consumer.sh
+# Copy the Debian stretch APT fix script to the container's /usr/local/bin directory
+COPY fix_debian_stretch_apt.sh /usr/local/bin/fix_debian_stretch_apt.sh
```
shmsr: Can be combined in one COPY directive imo:

https://docs.docker.com/reference/dockerfile/#copy

As all are being kept in `/usr/local/bin/`; avoids layers.

packages/rabbitmq/_dev/deploy/docker/entrypoint.d/async_rabbitmq_consumer.sh
giorgi-imerlishvili-elastic: It was suggested by chatGPT  to catch strings like this: {pid,12345}
but agree in this case having comma there is not necessary

packages/rabbitmq/data_stream/queue/fields/package-fields.yml
```diff
@@ -7,3 +7,14 @@
       dimension: true
       description: |
         Virtual host name with non-ASCII characters escaped as in C.
+    - name: node
+      type: group
+      # Reason for adding is to pass system tests.
```
ishleenk17, line 12: Its ok to not have this comment as I supose these fields would add value that whyy they are a must by stsem tests

ishleenk17 (approved):
Nit, otherwise looks good!

PR #14779, opened 2025-08-02
[ti_threatstream] Allow setting an advanced search query

chrisberkhout:
## Proposed commit message

```
[ti_threatstream] Allow setting an advanced search query

An advanced setting that lets the user specify a value for the `q`
parameter of the request, such as `confidence>=80 and severity=high`.
```

## Notes

Relevant details from the Anomali ThreatStream API Reference Guide:

> ### Advanced Search Queries
> Most filtering criteria can be specified using filter operators described in this document. However, if
you need to specify a complex filter, you can use an advanced search query. This method allows you
to specify queries similar to those you formulate on the Advanced search screen on the
ThreatStream user interface. To specify a filter language query, use this syntax in the API call:
`&q=<filter_language_query>`. The filter operators used for the filter language query are the
symbolic form (`=`, `<`, `>`, and so on) and not the descriptive form (exact, lt, gt, and so on) and the query
must be URL encoded. See "To retrieve intelligence using an advanced search query:" on the next
page for an example.

## Checklist

## Screenshots

Screenshot 2025-08-02 at 09-39-24 Add integration - Anomali - Integrations - Elastic

### DIFF
--- a/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
+++ b/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
@@ -29,6 +29,7 @@ state:
   initial_interval: {{initial_interval}}
   want_more: false
   preserve_original_event: {{preserve_original_event}}
+  query: {{query}}
 redact:
   fields:
     - api_key
@@ -43,6 +44,7 @@ program: |-
   			"limit": [string(state.page_size)],
   			?"update_id__gt": state.?cursor.last_update_id.optMap(id, [string(int(id))]),
   			?"remote_api": state.remote_api_true ? optional.of(["true"]) : optional.none(), // never set remote_api=false, only true or absent
+  			?"q": state.query != null && state.query != "" ? optional.of([state.query]) : optional.none(),
   		}.format_query()
   	).with(
   		{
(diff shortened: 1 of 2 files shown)


efd6:
When the manifest query var is not set, `state.query` is absent. So suggest instead a more explicit
```diff
diff --git a/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs b/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
index 043b99a07a..6453aef251 100644
--- a/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
+++ b/packages/ti_anomali/data_stream/intelligence/agent/stream/cel.yml.hbs
@@ -29,7 +29,9 @@ state:
   initial_interval: {{initial_interval}}
   want_more: false
   preserve_original_event: {{preserve_original_event}}
+{{#if query}}
   query: {{query}}
+{{/if}}
 redact:
   fields:
     - api_key
@@ -44,7 +46,7 @@ program: |-
                        "limit": [string(state.page_size)],
                        ?"update_id__gt": state.?cursor.last_update_id.optMap(id, [string(int(id))]),
                        ?"remote_api": state.remote_api_true ? optional.of(["true"]) : optional.none(), // never set remote_api=false, only true or absent
-                       ?"q": state.query != null && state.query != "" ? optional.of([state.query]) : optional.none(),
+                       ?"q": state.?query.orValue("") != "" ? optional.of([state.query]) : optional.none(),
                }.format_query()
        ).with(
                {
```
Depending on the syntax of the query, it may be worth using `escape_string` in the template.
```
{{#if query}}
  query: {{escape_string query}}
{{/if}}
```

chrisberkhout:
@efd6 Thanks.

> When the manifest query var is not set, `state.query` is absent. So suggest instead a more explicit

Right. I thought about that at the time and expected the template logic to set make `state.query: null`. I've run into this before. I think the template part is correct but the key gets stripped by fleet or something before it gets to the agent.

I added the condition and escaping and check.

efd6 (approved):

PR #14809, opened 2025-08-05
{m365_defender,microsoft_defender_endpoint}: Add mapping and transform for CDR workflows

kcreddy:
## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/base-fields.yml
kcreddy, line 1: @maxcold, for `microsoft_defender_endpoint` integration, I had to shorten the transform name as `latest_cdr_vuln` because if not, the system tests would throw an error during package installation:

```
Error: failed to setup system runner: can't install the package: there was an apply error: installation failed: can't install the package: could not zip-install package; API status code = 500; response body = {"statusCode":500,"error":"Internal Server Error","message":"action_request_validation_exception\n\tRoot causes:\n\t\taction_request_validation_exception: Validation Failed: 1: The id cannot contain more than 64 characters.;"}
```

Its coming from Kibana (logs below):

```
2025-08-05 19:18:07 [2025-08-05T13:48:07.229+00:00][WARN ][plugins.fleet] Error during execution of state "install_transforms" with status "failed": action_request_validation_exception
2025-08-05 19:18:07     Root causes:
2025-08-05 19:18:07             action_request_validation_exception: Validation Failed: 1: The id cannot contain more than 64 characters.;
2025-08-05 19:18:07 [2025-08-05T13:48:07.246+00:00][DEBUG][plugins.fleet] Created transform: logs-microsoft_defender_endpoint.latest_action-default-1.0.0
2025-08-05 19:18:07 [2025-08-05T13:48:07.311+00:00][DEBUG][plugins.fleet] Started transform: logs-microsoft_defender_endpoint.latest_action-default-1.0.0
2025-08-05 19:18:08 [2025-08-05T13:48:08.138+00:00][WARN ][plugins.fleet] Failure to install package [microsoft_defender_endpoint]: [ResponseError: action_request_validation_exception
2025-08-05 19:18:08     Root causes:
2025-08-05 19:18:08             action_request_validation_exception: Validation Failed: 1: The id cannot contain more than 64 characters.;]
2025-08-05 19:18:09 [2025-08-05T13:48:09.163+00:00][ERROR][plugins.fleet] Uninstalling microsoft_defender_endpoint-3.0.0 after error installing: [ResponseError: action_request_validation_exception
2025-08-05 19:18:09     Root causes:
2025-08-05 19:18:09             action_request_validation_exception: Validation Failed: 1: The id cannot contain more than 64 characters.;] with install type: install
```

Hope the shortened name is okay.

packages/m365_defender/manifest.yml
```diff
@@ -11,7 +11,7 @@ conditions:
   elastic:
     subscription: basic
   kibana:
-    version: "^8.18.0 || ^9.0.0"
+    version: "^8.19.0 || ^9.1.0"
```
kcreddy: Currently both integrations are not added to permissions list in Elasticsearch.
I created the PR: https://github.com/elastic/elasticsearch/pull/132445 to add them. Once approved and backported, I will update these versions accordingly. Until then the system tests will fail because of permissions error inside transform. 

cc: @maxcold

kcreddy: I added `run_as_kibana_system: false` for now to test. Will send you the credentials for testing.

packages/microsoft_defender_endpoint/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -150,8 +150,8 @@ processors:
   - set:
       field: event.id
       tag: set_event_id_from_microsoft_defender_endpoint_vulnerability_affected_machine_id
-      value: '{{{microsoft_defender_endpoint.vulnerability.affected_machine.id}}}-{{{_ingest.timestamp}}}'
-      if: ctx.microsoft_defender_endpoint?.vulnerability?.affected_machine?.id != null
+      copy_from: microsoft_defender_endpoint.vulnerability.affected_machine.id
+      ignore_empty_value: true
```
chrisberkhout, line 154: Seems odd to set the vulnerability event ID to the affected machine ID.
I see it's set to a vulnerability ID below so I guess this is just the fallback.

packages/microsoft_defender_endpoint/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -551,6 +570,17 @@ processors:
       tag: rename_cvssVector
       target_field: microsoft_defender_endpoint.vulnerability.cvss_vector
       ignore_missing: true
+  - grok:
+      field: microsoft_defender_endpoint.vulnerability.cvss_vector
+      tag: grok_to_extract_vulnerability_score_version
+      patterns:
+        - '^CVSS:%{DATA:vulnerability.score.version}/%{GREEDYDATA}$'
+        - '^%{GREEDYDATA}$'
```
chrisberkhout: With the 2nd pattern, won't it match and ignore any string that didn't match the first?

If that's intentional, maybe it can get `if: ctx.microsoft_defender_endpoint?.vulnerability?.cvss_vector instanceof String` and avoid the `on_failure`.

packages/microsoft_defender_endpoint/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -646,18 +676,25 @@ processors:
...
+      ignore_empty_value: true
+      if: ctx.vulnerability?.id != null && ctx.vulnerability.id.toUpperCase().contains('CVE') == true
   - set:
       field: event.id
       tag: set_event_id_from_vulnerability_id
-      value: '{{{vulnerability.id}}}-{{{_ingest.timestamp}}}'
-      if: ctx.event?.id == null && ctx.vulnerability?.id != null
+      copy_from: microsoft_defender_endpoint.vulnerability.id
+      ignore_empty_value: true
+      if: ctx.event?.id == null
```
chrisberkhout, line 705: If the vulnerability ID is a CVE, I guess we may get many events with the same vulnerability ID.

What's the intention with removing the timestamp from `event.id`?

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/base-fields.yml
chrisberkhout, line 1: The field mapping files in the transform that match what's in the data stream can removed by using the new [`elastic-package links`](https://github.com/elastic/elastic-package/?tab=readme-ov-file#elastic-package-links) functionality.

In case a file currently has both common and unique content, it'd be good to split that file.

chrisberkhout: The full links documentation: https://github.com/elastic/elastic-package/blob/main/docs/howto/links.md

chrisberkhout (commented):
Looks good. I made a few comments in microsoft_defender_endpoint, but they apply to both.

packages/microsoft_defender_endpoint/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
kcreddy: Actually, this field indicates a finding id of sorts, example: `86c0491db8ff7e8dcad520288b7759fa27793ce1-_-CVE-2024-11168-_-red_hat-_-python-unversioned-command_for_linux-_-0:3.9.18-3.el9_4.6-_-`.
It includes machineId, vulnerabilityId, packageName, packageVersion, etc.
https://github.com/elastic/integrations/blob/main/packages/m365_defender/data_stream/vulnerability/_dev/test/pipeline/test-vulnerability.log-expected.json#L60.

packages/microsoft_defender_endpoint/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
kcreddy: This is clarified with Cloud team, whether we really want uniqueness (on source indices) for this field, and it seems that we don't. Essentially the idea for this field is that we don't need to append timestamp as the same finding should have the same id over time.

packages/m365_defender/_dev/build/docs/README.md
```diff
@@ -96,10 +99,11 @@ Elastic Agent must be installed. For more details, check the Elastic Agent [inst
   - Client Secret
   - Tenant ID
 
-### Data Retention and ILM Configuration
-A full sync pulls in a large volume of data, which can lead to storage issues or index overflow over time. To avoid this, we’ve set up an Index Lifecycle Management (ILM) policy that automatically deletes data older than 7 days. This helps keep storage usage under control.
+#### Data Retention and ILM Configuration
+
+A full sync pulls in a large volume of data, which can lead to storage issues or index overflow over time. To avoid this, we've set up an Index Lifecycle Management (ILM) policy that automatically deletes data older than 7 days. This helps keep storage usage under control.
```
efd6:
```suggestion
A full sync pulls in a large volume of data, which can lead to storage issues or index overflow over time. To avoid this, we have set up an Index Lifecycle Management (ILM) policy that automatically deletes data older than 7 days. This helps keep storage usage under control.
```

packages/m365_defender/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -570,6 +589,17 @@ processors:
       tag: rename_cvssVector
       target_field: m365_defender.vulnerability.cvss_vector
       ignore_missing: true
+  - grok:
+      field: m365_defender.vulnerability.cvss_vector
+      tag: grok_to_extract_vulnerability_score_version
+      patterns:
+        - '^CVSS:%{DATA:vulnerability.score.version}/%{GREEDYDATA}$'
+        - '^%{GREEDYDATA}$'
```
efd6: The first of these can be `^CVSS:%{DATA:vulnerability.score.version}/` since the `GREEDYDATA` just consumes the string to throw it away.

It's not clear to me what the second one is for, perhaps just to prevent an error from a non-match, but why given there is an `on_failure` below which this will prevent from firing. There are a few ways to write this, the second can be more cheaply written as `'^.?'` or `''`, or preferably, just omitted and instead use `ignore_failure: true`.

packages/m365_defender/data_stream/vulnerability/fields/base-fields.yml
```diff
@@ -1,20 +1,16 @@
...
   value: m365_defender
 - name: event.dataset
+  external: ecs
   type: constant_keyword
-  description: Event dataset.
   value: m365_defender.vulnerability
 - name: '@timestamp'
-  type: date
-  description: Event timestamp.
+  external: ecs
```
efd6: Final new line.

packages/m365_defender/data_stream/vulnerability/fields/ecs.yml
```diff
@@ -0,0 +1,7 @@
+# Define ECS constant fields as constant_keyword
+- name: observer.vendor
+  type: constant_keyword
+  external: ecs
+- name: vulnerability.scanner.vendor
+  type: constant_keyword
+  external: ecs
```
efd6: Final new line.

packages/m365_defender/elasticsearch/transform/latest_cdr_vulnerabilities/fields/ecs.yml
```diff
@@ -0,0 +1,7 @@
+# Define ECS constant fields as constant_keyword
+- name: observer.vendor
+  type: constant_keyword
+  external: ecs
+- name: vulnerability.scanner.vendor
+  type: constant_keyword
+  external: ecs
```
efd6: Final new line.

packages/m365_defender/elasticsearch/transform/latest_cdr_vulnerabilities/fields/package.yml
```diff
@@ -0,0 +1,11 @@
...
+  type: group
+  fields:
+    - name: fixed_version
+      type: keyword
+    - name: name
+      type: keyword
+      external: ecs
+    - name: version
+      type: keyword
+      external: ecs
```
efd6: Final new line.

packages/m365_defender/elasticsearch/transform/latest_cdr_vulnerabilities/fields/resource.yml
```diff
@@ -0,0 +1,7 @@
+- name: resource
+  type: group
+  fields:
+    - name: id
+      type: keyword
+    - name: name
+      type: keyword
```
efd6: Final new line.

packages/m365_defender/elasticsearch/transform/latest_cdr_vulnerabilities/fields/vulnerability.yml
```diff
@@ -0,0 +1,10 @@
+- name: vulnerability
+  type: group
+  fields:
+    - name: cve
+      type: keyword
+      description: The CVE id of the vulnerability.
+    - name: published_date
+      type: date
+    - name: title
+      type: keyword
```
efd6: Final new line.

packages/microsoft_defender_endpoint/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
efd6: Similar comment, as for m365_defender. Also time complexity of the first pattern can be reduced.

packages/microsoft_defender_endpoint/data_stream/vulnerability/fields/base-fields.yml
```diff
@@ -1,20 +1,16 @@
...
   value: microsoft_defender_endpoint
 - name: event.dataset
+  external: ecs
   type: constant_keyword
-  description: Event dataset.
   value: microsoft_defender_endpoint.vulnerability
 - name: '@timestamp'
-  type: date
-  description: Event timestamp.
+  external: ecs
```
efd6: Final new line.

packages/microsoft_defender_endpoint/data_stream/vulnerability/fields/ecs.yml
```diff
@@ -0,0 +1,7 @@
+# Define ECS constant fields as constant_keyword
+- name: observer.vendor
+  type: constant_keyword
+  external: ecs
+- name: vulnerability.scanner.vendor
+  type: constant_keyword
+  external: ecs
```
efd6: Final new line.

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/ecs.yml
```diff
@@ -0,0 +1,7 @@
+# Define ECS constant fields as constant_keyword
+- name: observer.vendor
+  type: constant_keyword
+  external: ecs
+- name: vulnerability.scanner.vendor
+  type: constant_keyword
+  external: ecs
```
efd6: Final new line.

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/package.yml
```diff
@@ -0,0 +1,11 @@
...
+  type: group
+  fields:
+    - name: fixed_version
+      type: keyword
+    - name: name
+      type: keyword
+      external: ecs
+    - name: version
+      type: keyword
+      external: ecs
```
efd6: Final new line.

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/resource.yml
```diff
@@ -0,0 +1,7 @@
+- name: resource
+  type: group
+  fields:
+    - name: id
+      type: keyword
+    - name: name
+      type: keyword
```
efd6: Final new line.

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/vulnerability.yml
```diff
@@ -0,0 +1,10 @@
+- name: vulnerability
+  type: group
+  fields:
+    - name: cve
+      type: keyword
+      description: The CVE id of the vulnerability.
+    - name: published_date
+      type: date
+    - name: title
+      type: keyword
```
efd6: Final new line.

kcreddy: Done in [e608bbb](https://github.com/elastic/integrations/pull/14809/commits/e608bbb5341442e3e2a63d2fb2066e0fa9669351)

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/resource.yml
kcreddy: Done in [e608bbb](https://github.com/elastic/integrations/pull/14809/commits/e608bbb5341442e3e2a63d2fb2066e0fa9669351)

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/package.yml
kcreddy: Done in [e608bbb](https://github.com/elastic/integrations/pull/14809/commits/e608bbb5341442e3e2a63d2fb2066e0fa9669351)

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/ecs.yml
kcreddy: Done in [e608bbb](https://github.com/elastic/integrations/pull/14809/commits/e608bbb5341442e3e2a63d2fb2066e0fa9669351)

packages/microsoft_defender_endpoint/data_stream/vulnerability/fields/ecs.yml
kcreddy: Done in [e608bbb](https://github.com/elastic/integrations/pull/14809/commits/e608bbb5341442e3e2a63d2fb2066e0fa9669351)

packages/m365_defender/elasticsearch/transform/latest_cdr_vulnerabilities/fields/package.yml
kcreddy: Done in [e608bbb](https://github.com/elastic/integrations/pull/14809/commits/e608bbb5341442e3e2a63d2fb2066e0fa9669351)

packages/m365_defender/elasticsearch/transform/latest_cdr_vulnerabilities/fields/resource.yml
kcreddy: Done in [e608bbb](https://github.com/elastic/integrations/pull/14809/commits/e608bbb5341442e3e2a63d2fb2066e0fa9669351)

packages/m365_defender/elasticsearch/transform/latest_cdr_vulnerabilities/fields/ecs.yml
kcreddy: Done in [e608bbb](https://github.com/elastic/integrations/pull/14809/commits/e608bbb5341442e3e2a63d2fb2066e0fa9669351)

packages/m365_defender/data_stream/vulnerability/fields/ecs.yml
kcreddy: Done in [e608bbb](https://github.com/elastic/integrations/pull/14809/commits/e608bbb5341442e3e2a63d2fb2066e0fa9669351)

packages/m365_defender/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
kcreddy: Added `ignore_failure: true` in [e608bbb](https://github.com/elastic/integrations/pull/14809/commits/e608bbb5341442e3e2a63d2fb2066e0fa9669351)

packages/m365_defender/_dev/build/docs/README.md
kcreddy: Done in [e608bbb](https://github.com/elastic/integrations/pull/14809/commits/e608bbb5341442e3e2a63d2fb2066e0fa9669351)

efd6:
CI seems to be having problems with docker; I've been seeing problems in beats too.

kcreddy:
> CI seems to be having problems with docker; I've been seeing problems in beats too.

Actually I thought it was because images for `8.19.2` are not available yet (maybe soon). 
I need `8.19.2` for supporting permissions added to Elasticsearch recently: https://github.com/elastic/elasticsearch/pull/132629

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/base-fields.yml
kcreddy: I tested this today and this _links_ feature doesn't seem to be working as expected for me.

[Here](https://github.com/kcreddy/integrations/compare/defender_xdr_endpoint-cdr...kcreddy:integrations:test-links?expand=1) is the commit that I tested against.

When I build the integration using `elastic-package build` and check the `build` folder, the linked files are actually generated inside package root directory.
Screenshot 2025-08-12 at 2 09 26 PM

I checked the mappings in Kibana as well. All the linked mappings are missing. Only explicitly defined mapping exists.
Screenshot 2025-08-12 at 2 00 35 PM

Also, the transform produced an error which also indicates the mapping doesn't work.
Screenshot 2025-08-12 at 2 05 36 PM

packages/m365_defender/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -665,18 +691,25 @@ processors:
       tag: set_vulnerability_id_from_vulnerability_id
       copy_from: m365_defender.vulnerability.id
       ignore_empty_value: true
+  - set:
+      field: vulnerability.cve
```
alexreal1314, line 695: @kcreddy it is supposed to be `field: vulnerability.id`, this it were we store cve data.

packages/m365_defender/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -692,6 +725,21 @@ processors:
             return;
           }
         }
+  - set:
+      field: vulnerability.title
```
alexreal1314, line 729: @kcreddy I wonder what is the added value to the user seeing a title in this format:
'Vulnerability found in {{{package.name}}} {{{package.version}}}'

Since we already have package.name and package.version fields which appear in the flyout and data grid already and describe basically the same. This is kinda a duplication of data.
I understand that the title might be quite long - i'll check how we handle it today in the UI.

cc @maxcold 

image

packages/m365_defender/data_stream/vulnerability/sample_event.json
```diff
@@ -155,6 +156,7 @@
     ],
     "vulnerability": {
         "classification": "CVSS",
+        "cve": "CVE-2025-3074",
```
alexreal1314, line 159: should be "id": "CVE-2025-3074",

packages/m365_defender/docs/README.md
```diff
@@ -1670,6 +1675,7 @@ An example event for `vulnerability` looks as following:
     ],
     "vulnerability": {
         "classification": "CVSS",
+        "cve": "CVE-2025-3074",
```
alexreal1314, line 1678: should be "id": "CVE-2025-3074",

packages/m365_defender/elasticsearch/transform/latest_cdr_vulnerabilities/fields/vulnerability.yml
```diff
@@ -0,0 +1,10 @@
+- name: vulnerability
+  type: group
+  fields:
+    - name: cve
```
alexreal1314, line 4: - name: id

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/transform.yml
```diff
@@ -0,0 +1,40 @@
...
+        - exists:
+            field: resource.id
+dest:
+  index: "security_solution-microsoft_defender_endpoint.vulnerability_latest-v1"
+  aliases:
+    - alias: "security_solution-microsoft_defender_endpoint.vulnerability_latest"
+      move_on_creation: true
+latest:
+  unique_key:
+    - vulnerability.id
```
alexreal1314, line 19: @kcreddy seems like event.id as described [here](https://github.com/elastic/integrations/pull/14809/files#r2267698553) already includes vulnerability.id, package.name and package.version in that case and as we did in other integrations (rapid7, qualys, tenable) lets use this instead:
```
  unique_key:
    - event.id
    - resource.id
    - data_stream.namespace
```

packages/m365_defender/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
kcreddy: @alexreal1314, we also store CVE ids inside `vulnerability.id` field right above this processor. 

https://github.com/elastic/integrations/blob/2f2742031dadb6a684d9e4a652e1351aba8c0fb2/packages/m365_defender/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml#L689-L693

In here, we are populating `vulnerability.cve` whenever `ctx.vulnerability.id.toUpperCase().contains('CVE')`. This is discussed in the spreadsheet

packages/microsoft_defender_endpoint/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Looks good now

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/base-fields.yml
chrisberkhout: Okay, me too. Same after updating to `elastic-package v0.114.0 version-hash 6e962930 (build time: 2025-08-08T13:28:38+02:00)`.

I guess we shouldn't wait the fix for that.

packages/m365_defender/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
alexreal1314: @kcreddy Thanks, I didn't see it as it was not part of the changes.

Why are we adding vulnerability.cve it only for some of the 3p integrations? I see it in aws inspectr, google scc, microsoft defender and here.

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/transform.yml
kcreddy: I was going for consistency with CDR guide on this. Let me know if you want me to change to `event.id`.

packages/m365_defender/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
kcreddy: We should definitely be adding for all 3p integrations where applicable. It may have been missed because it wasn't Must Have. But will take a look for other integrations and add accordingly. Its also possible that for those other integrations, the `vulnerbaility.id` isn't given as a CVE.

packages/m365_defender/data_stream/vulnerability/sample_event.json
kcreddy: Same explanation as https://github.com/elastic/integrations/pull/14809#discussion_r2273183889.

packages/m365_defender/docs/README.md
kcreddy: Same explanation as https://github.com/elastic/integrations/pull/14809#discussion_r2273183889

packages/m365_defender/elasticsearch/transform/latest_cdr_vulnerabilities/fields/vulnerability.yml
kcreddy: Same explanation as https://github.com/elastic/integrations/pull/14809#discussion_r2273183889

packages/m365_defender/data_stream/vulnerability/elasticsearch/ingest_pipeline/default.yml
kcreddy: There is no field from the raw data to derive the `title` from. The integration previously added it from `Summary:` part of `vulnerability.description` field. But this alone is quite huge. 

Sample value: `The bzip2 utility, up to version 1.0.6, contains a vulnerability (CVE-2019-12900) in the BZ2_decompress function within decompress.c. This flaw involves an out-of-bounds write when processing files with numerous selectors, potentially leading to data integrity errors during decompression.`

After discussing with @maxcold and @nick-alayil inside the spreadsheet, we updated it to avoid large multiline titles.

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/transform.yml
alexreal1314: I think yes. Would like to get @maxcold opinion regarding it also.

alexreal1314 (approved):

packages/microsoft_defender_endpoint/data_stream/log/agent/stream/httpjson.yml.hbs
```diff
@@ -43,7 +43,8 @@ response.split:
     keep_parent: true
 cursor:
   lastUpdateTime:
-    value: "[[.last_response.body.lastUpdateTime]]"
+    value: '[[with $f := (index .last_response.body "lastUpdateTime")]][[if (ne $f "")]][[.last_response.body.lastUpdateTime]][[end]][[end]]'
```
efd6: I think this can be simpler.
```suggestion
    value: '[[with (index .last_response.body "lastUpdateTime")]][[.]][[end]]'
```
From ["text/template"](https://pkg.go.dev/text/template#hdr-Actions)
> ```
> {{with pipeline}} T1 {{end}}
>	If the value of the pipeline is empty, no output is generated;
>	otherwise, dot is set to the value of the pipeline and T1 is
>	executed.
> ```

Probably (untested), it could actually be `[[index .last_response.body "lastUpdateTime"]]`, but I don't think it's worth it since it relies on clear knowledge of the details of the template handler in the input; the `with` implementation is clearer.

kcreddy: Thanks, that makes sense. I updated as per your suggestion [0e47471](https://github.com/elastic/integrations/pull/14809/commits/0e474713a875286a98f128776c0fbe5ca942d746)

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/base-fields.yml
chrisberkhout: I looked at fixing it. There's a bit more too it than I thought. Here's an issue to track the problem: https://github.com/elastic/elastic-package/issues/2830

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/transform.yml
kcreddy: > I think yes. Would like to get @maxcold opinion regarding it also.

@maxcold, let me know if you like the unique combination changed to using `event.id` or the current combination (as per CDR guide) is okay.

maxcold: My thinking: as in our native integrations, we rely on this combination of fields, we went with this set of fields for 3p integrations as well. Then we discovered that for some integrations where some of these fields can be multivalue, we looked for a workaround and decided to use the event.id, but before we made sure it doesn't break our flows, eg. it's not a generated value that changes for the same vulnerability.id, package.name and package.version as we would see the duplication in that case. But in my opinion, in integrations where these fields are not multivalue, it's better to use the combination instead of `event.id` even if `event.id` covers exactly these fields for the following reasons:
- It's much clearer from looking at the current transform what the unique logic is. Because `event.id` can be anything and differ from provider to provider. So one needs to look into the ingest pipeline or even in the provider docs to understand what this event.id is, if it's unique and so on
- `event.id` implementation (if it's created via some logic in the ingest pipeline) can change and then it can stop providing the necessary qualities to be used in the unique setting. Again, with the combination of fields, it's much more transparent

@alexreal1314 do you have any specific reason that you think `event.id` would be a better fit here?

packages/microsoft_defender_endpoint/elasticsearch/transform/latest_cdr_vuln/fields/base-fields.yml
maxcold: yes, it's ok I think. Don't know if we need to come up with a common way to name it or not. We had the same issue with our native transform https://github.com/elastic/integrations/blob/main/packages/cloud_security_posture/elasticsearch/transform/misconfiguration/manifest.yml and named it simply `misconfiguration`, don't know if it has some effect on the UX for the users

kcreddy: Great, thanks for the confirmation.

PR #14814, opened 2025-08-05
Enhance backport_branch.sh to handle the case when package name and path are not identical

mykola-elastic:
## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


docs/extend/developer-workflow-support-old-package.md
```diff
@@ -74,6 +74,7 @@ Follow these detailed steps to release a fix for a given package version:
 
     * **BASE_COMMIT** (default: "") - enter the commit from the previous step (8cb321075afb9b77ea965e1373a03a603d9c9796)
     * **PACKAGE_NAME** (default: "") - enter the package name, for example aws
```
mrodm: It could be added a note here that this name should be the one defined in the manifest.yml file.

.buildkite/pipeline.backport.yml
```diff
@@ -35,6 +35,10 @@ steps:
       key: "PACKAGE_NAME"
```
mrodm, line 35: I think it would be helpful for users to add this note for this input field:
```yaml
    - text: "Enter package name (as defined in manifest.yml)"
      key: "PACKAGE_NAME"
      required: true
      default: ""
```

mykola-elastic: good one!

PR #14853, opened 2025-08-07
Vectra RUX: fix integer conversion for query parameter

chemamartinez:
## Proposed commit message

Fixed the handling of `from` query parameter that was being converted to scientific notation when high integer values, which made the API fail with a `Invalid request parameter` error.

Tested with Mito when `state.from == 3008842`:
```
string(state.?from.orValue(0)) => "3.008842e+06"
string(int(state.?from.orValue(0))) => "3008842"
```

Rules in `config.yml` test file have been ordered to match the order they are hit. Last rule for `audits` endpoint in that file tests the fix, before the CEL change it failed with error:
```
vectra_rux-1  | {"level":"debug","ts":"2025-08-07T07:39:31.957Z","caller":"httpserver/httpserver.go:58","msg":"request did not match with any rule: Request path: GET /api/v3.4/events/audits?event_timestamp_gte=2025-08-06T07%3A39%3A29Z&from=3.008842e%2B06&limit=2, Request Headers: 'User-Agent: [Go-http-client/1.1]' 'Authorization: [Bearer xxxx]' 'Accept-Encoding: [gzip]' 'Connection: [close]' , Request Body: ","address":":8090"}
```

## Checklist

### DIFF
--- a/packages/vectra_rux/changelog.yml
+++ b/packages/vectra_rux/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "0.3.2"
   changes:
-    - description: Fix CEL code to properly handle scientific notation in `from` query parameter.
+    - description: Fix CEL code to properly handle `from` query parameter.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/14853
 - version: "0.3.1"


packages/vectra_rux/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.3.2"
+  changes:
+    - description: Fix CEL code to properly handle scientific notation in `from` query parameter.
```
efd6:
```suggestion
    - description: Fix CEL code to properly handle `from` query parameter.
```
The notion of scientific notation here is not informative to users and the issue here is the result of an implementation detail, rather than an intrinsic aspect of the system.

efd6 (approved):
LGTM after changelog entry change.

PR #14882, opened 2025-08-08
[cisco_asa] Support IPv6 addresses in 750002 and 750003 messages

### LINKED ISSUE #14576: [Cisco ASA]: Parsing of port fails due to IPv6 Addresses
### Integration Name

Cisco ASA [cisco_asa]

### Dataset Name

cisco_asa.log

### Integration Version

2.43.5

### Agent Version

8.18.3

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.18.3

Debian 12

### Software/API Version

Cisco ASA 9.18(4)47

### Error Message

Error Message of Pipeline:
`Processor convert with tag convert_source_port in pipeline logs-cisco_asa.log-2.43.5 failed with message: For input string: "c6a0:1000:11::3:4500"`

source.port contains parts of the original IPv6 address and the port

### Event Original

`Jul 17 2025 07:53:41: %ASA-5-750002: Local:3a04:c6a0:1001:11::6:500 Remote:3a04:c6a0:3091:234:9c2f:3ce2:3996:6b6a:50329 Username:Unknown IKEv2 Received a IKE_INIT_SA request`

`Jul 17 2025 07:53:41: %ASA-5-750003: Local:3a04:c6a0:1001:11::6:4500 Remote:3a03:c6a0:3091:234:9c2f:3ce2:3996:6b6a:50330 Username:Unknown IKEv2 Negotiation aborted due to ERROR: Auth exchange failed`

### What did you do?

nothing, stock integration

### What did you see?

see above, source.ip and destination.ip are incomplete, source.port and destination.port cannot be parsed.

full IP address (v4/v6) and port

### Anything else?

Use GROK instead of DISSECT, because source/destination Address can be IPv4 or IPv6. Unfortunately Cisco just adds the port directly to the original IP without using standard IPv6 formatting ([2001:db8::1]:8080).

DISSECT fails in this case because the first part is not greedy.

```
  {
    "dissect": {
      "pattern": "Local:%{source.address}:%{source.port} Remote:%{destination.address}:%{destination.port} Username:%{user.name} %{event.reason}",
      "tag": "parse_750002",
      "field": "message",
      "if": "ctx._temp_.cisco.message_id == '750002'"
    }
  },  
```
```
{
    "dissect": {
      "pattern": "Local:%{source.address}:%{source.port} Remote:%{destination.address}:%{destination.port} Username:%{user.name} %{event.reason} ERROR:%{+event.reason}",
      "tag": "parse_750003",
      "field": "message",
      "if": "ctx._temp_.cisco.message_id == '750003'"
    }
  },
```

taylor-swanson:
## Proposed commit message

- Add support for IPv6 addresses in 750002 and 750003 messages.
- Required changing dissect processors to grok

## Checklist

```
cd packages/cisco_asa
elastic-package test
```

## Related issues

- Closes #14576

### DIFF
--- a/packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
@@ -1183,7 +1183,7 @@ processors:
       tag: parse_750002
       field: "message"
       patterns:
-        - "Local:%{DATA:source.address}:%{NUMBER:source.port} Remote:%{DATA:destination.address}:%{NUMBER:destination.port} Username:%{DATA:user.name} %{GREEDYDATA:event.reason}"
+        - "Local:%{IP:source.address}:%{NUMBER:source.port} Remote:%{IP:destination.address}:%{NUMBER:destination.port} Username:%{DATA:user.name} %{GREEDYDATA:event.reason}"
   - dissect:
       if: "ctx._temp_.cisco.message_id == '713120'"
       tag: parse_713120
@@ -1207,7 +1207,7 @@ processors:
       tag: parse_750003
       field: "message"
       patterns:
-        - "Local:%{DATA:source.address}:%{NUMBER:source.port} Remote:%{DATA:destination.address}:%{NUMBER:destination.port} Username:%{DATA:user.name} %{GREEDYDATA:event.reason}"
+        - "Local:%{IP:source.address}:%{NUMBER:source.port} Remote:%{IP:destination.address}:%{NUMBER:destination.port} Username:%{DATA:user.name} %{GREEDYDATA:event.reason}"
   - grok:
       if: '["713905", "713904", "713906", "713902", "713901"].contains(ctx._temp_.cisco.message_id)'
       tag: parse_713901-713906


packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1178,11 +1178,12 @@ processors:
...
       field: "message"
       pattern: "%{network.type}: An %{network.direction} %{_temp_.cisco.tunnel_type} SA (SPI= %{}) between %{source.address} and %{destination.address} (user= %{user.name}) has been %{}."
-  - dissect:
+  - grok:
       if: "ctx._temp_.cisco.message_id == '750002'"
       tag: parse_750002
       field: "message"
-      pattern: "Local:%{source.address}:%{source.port} Remote:%{destination.address}:%{destination.port} Username:%{user.name} %{event.reason}"
+      patterns:
+        - "Local:%{DATA:source.address}:%{NUMBER:source.port} Remote:%{DATA:destination.address}:%{NUMBER:destination.port} Username:%{DATA:user.name} %{GREEDYDATA:event.reason}"
```
qcorporation: do we have to use `grok`? is it the trailing message, hence GREEDYDATA, the reason why we can't use `dissect` in this use case?

packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1201,11 +1202,12 @@ processors:
...
         - "Authentication: rejected, group = %{NOTSPACE:source.user.group.name} user = %{USER:source.user.name} , Session Type: %{NOTSPACE:_temp_.cisco.session_type}"
         - "Group <%{DATA:source.user.group.name}> User <%{DATA:source.user.name}> IP <%{IP:source.address}> Authentication: rejected, Session Type: %{NOTSPACE:_temp_.cisco.session_type}\\."
-  - dissect:
+  - grok:
       if: "ctx._temp_.cisco.message_id == '750003'"
       tag: parse_750003
       field: "message"
-      pattern: "Local:%{source.address}:%{source.port} Remote:%{destination.address}:%{destination.port} Username:%{user.name} %{event.reason} ERROR:%{+event.reason}"
+      patterns:
+        - "Local:%{DATA:source.address}:%{NUMBER:source.port} Remote:%{DATA:destination.address}:%{NUMBER:destination.port} Username:%{DATA:user.name} %{GREEDYDATA:event.reason}"
```
qcorporation: is `source.address` always an IP address or could it be a uri?
if it's only IP then we should use `%{IP:source.address}` so we can handle both IPv4 and IPv6 validation

packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: In this case, we do. There's too much ambiguity with IPv6 addresses and how they've formatted the port here. Normally, you're supposed to wrap the IPv6 address in square brackets, then add `:PORT`, but Cisco didn't do that here. Dissect fails here because it can't handle multiple colons in the IPv6 address. We got away with this for IPv4 because there are no colons in an IPv4 address.

packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: I was afraid that the `IP` pattern wouldn't be able to handle the `:PORT` at the end, but it actually seems to handle it fine. I can change it to use `IP`.

It will only ever be an IPv4 or IPv6 address as far as we know.

qcorporation (approved):
thanks for the changes

PR #14923, opened 2025-08-13
[Cloud Security] Add `aws` prefix to CSPM vars

seanrathier:
## Proposed commit message
Add `aws` prefix to CSPM vars

This change is needed so we can use the same input variables for CSPM and Cloud Asset Inventory in order to make integration with the CloudSetup component.

## Checklist

## Related issues
- https://github.com/elastic/security-team/issues/13207
- https://github.com/elastic/kibana/pull/231343

### DIFF
--- a/packages/cloud_security_posture/changelog.yml
+++ b/packages/cloud_security_posture/changelog.yml
@@ -15,6 +15,11 @@
 # 1.4.x - 8.9.x
 # 1.3.x - 8.8.x
 # 1.2.x - 8.7.x
+- version: "4.0.0-preview01"
+  changes:
+    - description: Changed AWS input vars with aws. prefix
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/14791
 - version: "3.0.1"
   changes:
     - description: Save GCP Project ID as string
--- a/packages/cloud_security_posture/data_stream/findings/agent/stream/aws.yml.hbs
+++ b/packages/cloud_security_posture/data_stream/findings/agent/stream/aws.yml.hbs
@@ -23,20 +23,38 @@ config:
         {{#if access_key_id}}
         access_key_id: {{access_key_id}}
         {{/if}}
+        {{#if aws.access_key_id}}
+        access_key_id: {{aws.access_key_id}}
+        {{/if}}
         {{#if secret_access_key}}
         secret_access_key: {{secret_access_key}}
+        {{/if}}
+        {{#if aws.secret_access_key}}
+        secret_access_key: {{aws.secret_access_key}}
         {{/if}}
         {{#if session_token}}
         session_token: {{session_token}}
         {{/if}}
+        {{#if aws.session_token}}
+        session_token: {{aws.session_token}}
+        {{/if}}
         {{#if shared_credential_file}}
         shared_credential_file: {{shared_credential_file}}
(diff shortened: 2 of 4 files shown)


packages/cloud_security_posture/changelog.yml
```diff
@@ -15,6 +15,11 @@
 # 1.4.x - 8.9.x
 # 1.3.x - 8.8.x
 # 1.2.x - 8.7.x
+- version: "4.0.0-preview01"
```
kfirpeled, line 18: what's the reasoning behind major version upgrade? 
So far, we kept it in sync with major stack versions

packages/cloud_security_posture/data_stream/findings/agent/stream/aws.yml.hbs
```diff
@@ -23,21 +23,39 @@ config:
         {{#if access_key_id}}
         access_key_id: {{access_key_id}}
         {{/if}}
+        {{#if aws.access_key_id}}
```
kfirpeled, line 26: what happens if both `access_key_id` and `aws.access_key_id` are available?

I believe only one should be used to resolve source of truth

`access_key_id` is deprecated and its content is migrated to `aws.access_key_id`.  then I'd use `aws.access_key_id` when available, and when not I'll check for `access_key_id` for compatibility

kfirpeled (changes requested):

packages/cloud_security_posture/changelog.yml
```diff
@@ -15,6 +15,11 @@
...
+  changes:
+    - description: Changed AWS input vars with aws. prefix
+      type: enhancement
```
andrewkroh, line 21: I noticed the GH PR is labeled with `breaking change` but that's not in agreement with the changelog `type`. Assuming it's a breaking change that affects end users then I would expect to have a `changes` item with `type: breaking-change` explaining the impact and how users can handle the breaking change when they upgrade.

seanrathier: @andrewkroh, I was pretty sure that this would be a breaking eventually, however I managed to find a different path in Kibana to fix this issue so I will be closing this.

I will add the correct type next time.

packages/cloud_security_posture/data_stream/findings/agent/stream/aws.yml.hbs
seanrathier: I was doing that in the `aws.hbs` file,  the change checks for both values with and without the `aws` prefix.  Or so I thought that was the solution you were referring to in DMs.

However,  this change does not address showing the Fleet extension in Kibana for older versions of the integration; therefore, I am going to close this PR and have a [CloudSetup mapping](https://github.com/elastic/kibana/pull/231343/files#diff-91a83d3a0d7b6dd928735193df48c22e180e4c207c25d65262bcb2573fc3f8aeR36)

packages/cloud_security_posture/changelog.yml
seanrathier: We are at version 3. I perceived this to be a breaking change, warranting a major version change.

seanrathier:
I am closing this PR because there is a path in Kibana that is less risky and to support older versions of the CSPM integration.

Issue #14735, opened 2025-07-30
o365: rewrite agent CEL collector code to use work lists


efd6:
The current CEL input implementation uses an atomic approach to collecting data from the API; it attempts to do all of the work required to collect the current set of data in a single periodic evaluation.

This means that:
1. any failure during the evaluation results in the failure of collecting all events that would be published
2. the full API response, all intermediate eval node values, and the final results are kept in memory during the evaluation

In practice, we have not seen any impact arising from partial failures resulting in elevated rates of complete failure (1.), but we do regularly see user's with OoM failures when using the integration. This is exacerbated by the fact that the O365 API does not provide a mechanism for API callers to specify a limit for response sizes that the API returns.

The fix to both of the issues above is to re-implement the CEL program to make use of work lists rather than to operate atomically. The work list approach has been used in a number of other CEL-based integrations and involves the agent collector getting a set of work items (filling the work list) if it has none, otherwise doing work on the first element of the work list and returning the result and the rest of the work list to be worked on in the next `want_more` iteration. By doing this, at most one collection will fail and at most one event will be in memory at a time.

The change would have no impact on users or configuration otherwise.

PR #14932, opened 2025-08-14
o365: reduce code complexity, repeated work and garbage generation

efd6:
## Proposed commit message
```
o365: reduce code complexity, repeated work and garbage generation

This is the first step in making the integration safer to use in a
memory-constrained setting. The current code does the full set of
requests for each element of a set of subscription results, causing a
heavy memory load. The intention is to make use of work lists to reduce
the individual evaluation memory load when the CEL program is running.
The existing code is relatively over-complex, so this reduces that
complexity, while replacing heavily allocating operations with lighter
calls, and omitting completely calls that are not required, and so
eliding the allocation that those calls fruitlessly make.

Rather than making this change a commit in a pull request to convert to
work lists, this optimised version is made a version release since it is
a less risky change, while still improving performance and memory use
characteristics.

The changes that are made include:

* using sprintf to construct strings
* using the exists macro rather than the size of a filter result
* retaining work rather than repeating it
* reducing copying large values
* using optional types to avoid ternaries
* removing redundant calls: flatten and drop_empty on flat, all
  non-empty lists
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- For #14735

### DIFF
--- a/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
+++ b/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
@@ -150,9 +150,9 @@ program: |-
           ).do_request().as(list_resp,
             list_resp.Body.decode_json().as(list_body,
               (
-                type(list_body) == list && list_body[?0].optMap(e,
-                  e.contentUri != "" && e.contentCreated != ""
-                ).orValue(false)
+                type(list_body) == list && list_body[?0].as(e,
+                  e.contentUri.orValue("") != "" && e.contentCreated.orValue("") != ""
+                )
               ) ?
                 // contents exist to consume
                 list_body.map(l1,
--- a/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
+++ b/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
@@ -75,244 +75,227 @@ redact:
     - base.tenant_id
 
 program: |-
-  state.base.content_types.split(",").map(content_type_raw,
-  	content_type_raw.trim_space()
-  ).map(content_type,
-  	request(
-  		"POST",
-  		state.url.trim_right("/") + "/api/v1.0/" + state.base.tenant_id + "/activity/feed/subscriptions/start?" +
-  		{
-  			"contentType": [content_type],
-  			"PublisherIdentifier": [state.base.tenant_id],
-  		}.format_query()
-  	).do_request().as(start_subs_resp,
-  		start_subs_resp.Body.decode_json().as(start_subs_resp_body,
-  			(
-  				has(start_subs_resp_body.status) && start_subs_resp_body.status == "enabled" ||
-  				has(start_subs_resp_body.error) && has(start_subs_resp_body.error.code) &&
-  				start_subs_resp_body.error.code == "AF20024"
-  			) ?
-  				// When start-subscription API returns success or if already started subscription,
-  				duration(state.base.batch_interval).as(batch_interval,
-  					min(duration("24h"), batch_interval).as(batch_interval,
-  						request(
-  							"GET",
-  							(
-  								state.want_more && has(state.?cursor.content_types_state_as_list) &&
-  								size(state.cursor.content_types_state_as_list.filter(e, e.content_type == content_type)) > 0 &&
-  								state.cursor.content_types_state_as_list.filter(e, e.content_type == content_type)[0].next_page != ""
-  							) ?
-  								// if NextPageUri exists
-  								state.cursor.content_types_state_as_list.filter(e, e.content_type == content_type)[0].next_page
-  							:
-  							(
-  								has(state.?cursor.content_types_state_as_list) &&
-  								size(state.cursor.content_types_state_as_list.filter(e, e.content_type == content_type)) > 0
-  							) ?
-  								// if NextPageUri does not exist, but content_type_state_created_at exists in state
-  								state.cursor.content_types_state_as_list.filter(e, e.content_type == content_type).as(content_type_state,
-  									content_type_state[0].content_created_at.as(content_type_state_created_at,
-  										// if saved time inside state is more than 7 days old, then change it to 7 days.
-  										max(
-                        // Enforce the maximum age limit.
-  											now() - duration(state.base.maximum_age),
-  											content_type_state_created_at.parse_time(time_layout.RFC3339)
-  										).as(state_created_at_calc,
-  											state.url.trim_right("/") + "/api/v1.0/" + state.base.tenant_id + "/activity/feed/subscriptions/content?" +
-  											{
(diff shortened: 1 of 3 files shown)


packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -74,244 +74,227 @@ redact:
...
+                      "endTime": [string(min(now(), now() - duration(state.base.list_contents_start_time) + batch_interval))],
+                    }.format_query(),
+                  ]
+                )
+            )
+          ).do_request().as(list_resp,
+            list_resp.Body.decode_json().as(list_body,
+              (
+                type(list_body) == list && list_body[?0].optMap(e,
+                  e.contentUri != "" && e.contentCreated != ""
```
kcreddy: Does `contentUri` and `contentCreated` needs to be checked for `!= null` as well?

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -74,244 +74,227 @@ redact:
...
+                // contents does not exist, or is empty array
+                list_resp.Request.URL.parse_url().RawQuery.parse_query().as(r_query,
+                  [
+                    {
+                      "events_per_content_type": (size(list_body) == 0) ? [] : [list_body],
+                      "content_type": content_type,
+                      "content_created_at": (list_resp.StatusCode == 200 && (r_query.?endTime.optMap(t, t.size() > 0)).orValue(false)) ?
+                        (r_query.endTime[0])
+                      :
+                        r_query.?startTime[0].orValue(
```
kcreddy, line 221: Shouldn't `r_query.?startTime[0].orValue(` also have a condition on `.size() > 0` just like `endTime` has `(r_query.?endTime.optMap(t, t.size() > 0)).orValue(false))` ?

Just curious why we handle the same previous size conditions differently here.

kcreddy (approved):
LGTM overall. Just minor clarifications.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -74,244 +74,227 @@ redact:
...
+                  "error": {
+                    "code": string(sub_resp.StatusCode),
+                    "id": string(sub_resp.Status),
+                    "message": content_type + ": " + (
+                      sub_body.?error.code.optFlatMap(code,
+                        sub_body.?error.message.optMap(message,
+                          code + " " + message
+                        )
+                      )
+                    ).orValue(sub_body),
```
chrisberkhout, line 254: Would be good to add the content type prefix for the `sub_body` case. I forgot that last time.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -74,244 +74,227 @@ redact:
...
+          state.base.tenant_id,
+          {
+            "contentType": [content_type],
+            "PublisherIdentifier": [state.base.tenant_id],
+          }.format_query(),
+        ]
+      )
+    ).do_request().as(sub_resp,
+      sub_resp.Body.decode_json().as(sub_body,
+        (sub_body.?status == optional.of("enabled") || sub_body.?error.code == optional.of("AF20024")) ?
```
chrisberkhout, line 94: Nice.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: I think all the checking is redundant. The API requires either no times (and returns a default time range like last 15 mins or something) or it needs both `startTime` and `endTime`. I'm pretty sure we always provide the times.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: I agree. It looks like it based on the code. I'm not sure if they keys will always be there.

chrisberkhout (approved):
Nice refactoring.

I agree with Krishna about the null check probably being needed.

Other points are some minor potential improvements.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: `null` != absence; does O365 use `null` to indicate absence? Depending on whether they do, the implementation would be different. I'm going to assume that they don't since the previous code does not tolerate them doing so.

This then becomes
```
                type(list_body) == list && list_body[?0].as(e,
                  e.contentUri.orValue("") != "" && e.contentCreated.orValue("") != ""
                )
```

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: This is sort of clever, so I was a little reluctant to use it, but the syntax saving I think is worth it. If I missed this elsewhere, I'll fix it.

To clarify the behaviour here, the index into startTime is against the value as an optional type, so the index is allowed event if it would fail, if it's zero-length it would fail, returning an `optional.none()`, so we don't need to check that the length is safe (the expression here is returning a `string` and the types agree). In the case of the `endTime`, we have the same thing, but the types don't agree, on existence, we are a `string` and on absence we are a `bool`. This is why we need the comparison. I had briefly thought I'd missed some, but all the "missed" cases are the same shape and so cannot be simplified.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: AFAICS that's already there; there is `content_type +` before the total expression here, which includes this value

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: I've gone over the code with Chris' comment in mind, and I believe that is true. Do we want to remove all the checks? While I agree, I'd like to keep the risk level down on this change, so I will not remove them now.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: That's right. Actually it was right in the previous version too.

I'd seen logs where the combined error messages look like this:
```
"The subscription was disabled.; Audit.Exchange: StartSubscription IDS AND STACK TRACE...; The subscription was disabled.; DLP.All: StartSubscription IDS AND STACK TRACE..."
```

It looked like the "The subscription was disabled" part was missing the content type prefix, but now I realize it's not coming from a subscription request. It's there because one of the data requests resulted in an event that coincidentally has the `{ "error": { "message": ... }` structure.

PR #14948, opened 2025-08-15
[iptables] added optional space to process vyos logs

### LINKED ISSUE #10881: [iptables]: not properly parsing VyOS logs
### Integration Name

Iptables [iptables]

### Dataset Name

_No response_

### Integration Version

1.16.1

### Agent Version

8.15.0

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.15.0

Ubuntu 22.04.4 LTS

### Software/API Version

_No response_

### Error Message

_No response_

### Event Original

[7598791.530254] [RULE_IN-2001-A] IN=eth0 OUT=eth1 MAC=00:1b:21:bc:c7:ba:52:54:00:4a:b3:d0:08:00 SRC=10.10.10.10 DST=10.11.11.11 LEN=84 TOS=0x00 PREC=0x00 TTL=62 ID=5751 DF PROTO=UDP SPT=49704 DPT=53 LEN=64 

### What did you do?

I forward IPTables logs from a VyOS firewall to an agent where I use the Custom UDP input, with Syslog parsing enabled.
Then in the custom ingestion pipeline for that dataset, I tell it to call the logs-iptables.log-1.16.1 pipeline.

### What did you see?

some of these logs, at least some DROP logs, seem to be parsed fine with the UBIQUITY_LABEL, but others, at least what I saw now, ACCEPT or REJECT logs, have a "space" between the [UBIQUITY_LABEL] and the following IN=.... but the 4th grok pattern in the ingest pipeline Grok "Extracts values from "message" that match a grok pattern", looks alike: 

```
%{GREEDYDATA}\[%{UBIQUITI_LABEL}\]%{IPTABLES}%{SPACE}
 ```

due to the missing space, it didn't matched, and moved on with the more general pattern below. So I only ended up with event.action: drop events, and the others don't have that field set.

Since VyOS more or less the same naming convention for its rules, the IPTables integration should parse them same way as Ubiquity IPTables logs. So that event.action will properly be set to accept, drop, reject.

### Anything else?

updating above mentioned grok pattern to:

```
%{GREEDYDATA}\[%{UBIQUITI_LABEL}\]%{SPACE}?%{IPTABLES}%{SPACE}
 ```

did the trick for me. now in discover, I can see "drop" and "accept", for some reason, when there are reject rules I only see "r", haven't yet found where the event.action is expanded from the letters to the long names.

robester0403:
## Proposed commit message
```
iptables: Added an optional space to grok patterns as to process vyos logs properly without operator error.
```

## Checklist

~~- [] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).~~

Run and observe the new pipeline test to see expected corrected results.

## Related Issues
Closes #10881

### DIFF
--- a/packages/iptables/changelog.yml
+++ b/packages/iptables/changelog.yml
@@ -1,8 +1,8 @@
 # newer versions go on top
 - version: "1.21.3"
   changes:
-    - description: Added optional space to properly process UBIQUITI_LABEL.
-      type: enhancement
+    - description: Added optional space to properly process UBIQUITI_LABEL. Removed unneeded trailing space from grok patterns.
+      type: bugfix
       link: https://github.com/elastic/integrations/pull/14948
 - version: "1.21.2"
   changes:
--- a/packages/iptables/data_stream/log/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/iptables/data_stream/log/elasticsearch/ingest_pipeline/default.yml
@@ -54,11 +54,11 @@ processors:
   - grok:
       field: message
       patterns:
-      - '%{IPTABLES_HOSTNAME}%{GREEDYDATA}\[%{UBIQUITI_LABEL}\]%{IPTABLES}%{SPACE}'
-      - '%{IPTABLES_ACTION}%{GREEDYDATA}%{IPTABLES}%{SPACE}'
+      - '%{IPTABLES_HOSTNAME}%{GREEDYDATA}\[%{UBIQUITI_LABEL}\]%{IPTABLES}'
+      - '%{IPTABLES_ACTION}%{GREEDYDATA}%{IPTABLES}'
       - '%{IPTABLES_HOSTNAME}%{SPACE}%{UDM_LOGS}%{IPTABLES_IP_PAYLOAD}'
-      - '%{GREEDYDATA}\[%{UBIQUITI_LABEL}\](%{SPACE})?%{IPTABLES}%{SPACE}'
-      - '%{GREEDYDATA}%{IPTABLES}%{SPACE}'
+      - '%{GREEDYDATA}\[%{UBIQUITI_LABEL}\](%{SPACE})?%{IPTABLES}'
+      - '%{GREEDYDATA}%{IPTABLES}'
       pattern_definitions:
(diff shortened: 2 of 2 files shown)


packages/iptables/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -57,7 +57,7 @@ processors:
       - '%{IPTABLES_HOSTNAME}%{GREEDYDATA}\[%{UBIQUITI_LABEL}\]%{IPTABLES}%{SPACE}'
       - '%{IPTABLES_ACTION}%{GREEDYDATA}%{IPTABLES}%{SPACE}'
       - '%{IPTABLES_HOSTNAME}%{SPACE}%{UDM_LOGS}%{IPTABLES_IP_PAYLOAD}'
-      - '%{GREEDYDATA}\[%{UBIQUITI_LABEL}\]%{IPTABLES}%{SPACE}'
+      - '%{GREEDYDATA}\[%{UBIQUITI_LABEL}\](%{SPACE})?%{IPTABLES}%{SPACE}'
```
andrewkroh: Why the trailing `%{SPACE}`? Is that necessary? The pattern is not anchored with `$` so IIUC it will effectively ignore trailing space at the end. It's not covered by the test case either.

packages/iptables/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.21.3"
+  changes:
+    - description: Added optional space to properly process UBIQUITI_LABEL.
+      type: enhancement
```
andrewkroh: You labeled the PR with `bugfix`. The changelog type and labels should be in agreement.

packages/iptables/data_stream/log/elasticsearch/ingest_pipeline/default.yml
robester0403: Regarding the trailing space at the end of these patterns, I think it's not necessary. I tested it and it made no difference to the expected logs & current test cases. I will remove it from the patterns.

andrewkroh (approved):
LGTM

PR #14959, opened 2025-08-18
fix(azure.eventhub): handles properties as string and drops empty fields

zmoog:
### WHAT

- Drops empty/null fields
- Renames `azure.eventhub.properties` as `azure.eventhub.properties.raw` when `azure.eventhub.properties` is a string instead of an object.

### WHY

**Drops empty/null fields**

Sometimes Azure services emit log event like the following:

```json
{
  "category": "NetworkSecurityGroupEvent",
  "operationName": "NetworkSecurityGroupEvents",
  "properties": {
    "conditions": {
      "": "",
      "destinationPortRange": "0-65535",
      "sourcePortRange": "0-65535"
    },
    "direction": "Out",
    "macAddress": "00-11-22-33-4444",
    "primaryIPv4Address": "10.0.4.6",
    "primaryIPv6Address": "ace:ace:dead:beef::9",
    "priority": 65000,
    "ruleName": "DefaultRule_AllowVnetOutBound",
    "subnetPrefix": "10.0.4.0/25",
    "type": "allow",
    "vnetResourceGuid": "{a08d316f-3c0a-428d-84ec-2977078852a5}"
  }
}
```

Elasticsearch cannot index `"": ""` field inside `properties.condition`, so we need to clean it up.

**Renames `azure.eventhub.properties` as `azure.eventhub.properties.raw` when `azure.eventhub.properties` is a string instead of an object**

Sometimes Azure sends `azure.eventhub.properties` as a string, this field should really be a an `object` instead, causing mapping errors like:

```text
object mapping for [azure.eventhub.properties] tried to parse field [properties] as object, but found a concrete value
```

By renaming `azure.eventhub.properties` as `azure.eventhub.properties.raw`, we avoid the mapping error and give users the opportunity to handle the original value using a custom pipeline.

## Checklist


packages/azure/data_stream/eventhub/elasticsearch/ingest_pipeline/parsed-message.yml
```diff
@@ -39,6 +45,23 @@ processors:
...
+      source: |
+        boolean dropEmptyFields(Object object) {
+          if (object == null || object == "") {
+            return true;
+          } else if (object instanceof Map) {
+            ((Map) object).values().removeIf(value -> dropEmptyFields(value));
+            return (((Map) object).size() == 0);
+          } else if (object instanceof List) {
+            ((List) object).removeIf(value -> dropEmptyFields(value));
+            return (((List) object).length == 0);
```
Kavindu-Dodan: It should be `(((List) object).size() == 0)` isn't it ?

zmoog: I see it works both ways:

CleanShot 2025-08-18 at 23 38 01@2x

However `size()` seems like the right approach since both `List` and `Map` extend `Collection`.

Kavindu-Dodan: Interesting, I thought painless is Java, but I think `.length` is an addon by painless.

packages/azure/data_stream/eventhub/elasticsearch/ingest_pipeline/parsed-message.yml
```diff
@@ -15,6 +15,12 @@ processors:
   - json:
       field: event.original
       target_field: azure.eventhub
+  - rename:
+      field: azure.eventhub.properties
+      if: "ctx.azure?.eventhub?.properties instanceof String"
+      target_field: azure.eventhub.properties.raw
+      ignore_missing: true
+      description: 'Rename the field to `properties.raw` to avoid parse errors when `properties` is a string.'
```
kaiyan-sheng:
```suggestion
      description: Rename the field to `properties.raw` to avoid parse errors when `properties` is a string.       
```

muthu-mps:
The sample event [here](https://github.com/elastic/integrations/pull/14959#:~:text=%22%2C%0A%20%20%22-,properties,-%22%3A%20%7B%0A%20%20%20%20%22conditions) has the properties with fields and the **conditions** property has empty string and also fields with values. Can we include this format as well to the pipeline tests?

muthu-mps (approved):
Left a comment on adding additional events to the pipeline tests. Change looks good to me otherwise!

chemamartinez (approved):
LGTM! In case it’s useful to know you can also apply `subobjects: false` to solve the conflict. Done for example at https://github.com/elastic/integrations/pull/13925.

zmoog:
> The sample event [here](https://github.com/elastic/integrations/pull/14959#:~:text=%22%2C%0A%20%20%22-,properties,-%22%3A%20%7B%0A%20%20%20%20%22conditions) has the properties with fields and the **conditions** property has empty string and also fields with values. Can we include this format as well to the pipeline tests?

Yeah, but this would require to add mappings for these fields not to make test fail, which is something I am trying to avoid in a generic integration like `azure.eventhub`.

zmoog:
> LGTM! In case it’s useful to know you can also apply `subobjects: false` to solve the conflict. Done for example at #13925.

Great point, I didn't think about using `subobjects: false`.

Since this is a generic integration, where users are supposed to customize pipeline and mapping, I'm trying to leave all the options on the table.

PR #14961, opened 2025-08-18
fix(platformlogs) workaround the invalid field names from Azure Native ISV

zmoog:
## Proposed commit message

Renames the field names coming from the Azure Native ISV from the current names to the expected names:

| Current | Expected |
|--------|--------|
| `dataStream` | `data_stream` |
| `azureLogForwarder` | `azure_log_forwarder` | 
| `timestamp` | `@timestamp` |

It seems that Azure updated their Elasticsearch client, but they probably missed mapping the above field to the expected names we were using before.

While we wait for a proper fix on the Azure side, we are shipping this workaround.

## Checklist

### DIFF
--- a/packages/azure/changelog.yml
+++ b/packages/azure/changelog.yml
@@ -1,3 +1,8 @@
+- version: "1.28.4"
+  changes:
+    - description: Workaround the invalid field names from Azure Native ISV
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/14961
 - version: "1.28.3"
   changes:
     - description: Handle invalid log events (empty fields and properties as string) from Azure services.
--- a/packages/azure/manifest.yml
+++ b/packages/azure/manifest.yml
@@ -1,6 +1,6 @@
 name: azure
 title: Azure Logs
-version: "1.28.3"
+version: "1.28.4"
 description: This Elastic integration collects logs from Azure
 type: integration
 icons:


muthu-mps:
@zmoog - Can we add changelog entry?

packages/azure/data_stream/platformlogs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -13,6 +13,38 @@ processors:
...
+      ignore_missing: true
+      override: false
+      description: 'Renames the `azureLogForwarder.serviceProvider` field to `azure_log_forwarder.service_provider` to workaround the issue with invalid field names in the Azure Native ISV logs.'
+  - rename:
+      field: dataStream
+      target_field: data_stream
+      ignore_missing: true
+      override: false
+      description: 'Renames the `dataStream` field to `data_stream` to workaround the issue with invalid field names in the Azure Native ISV logs.'
+  - rename:
```
chemamartinez: How sure are we that the `timestamp` field contains a valid format? would it be worth it to use the `date` processor just in case?

zmoog: Good point.

This is a temporary workaround to rename the invalid timestamp field name Azure is sending since they updated the ES client. But you're right, we should use a `date` processor to be on the safe side.

zmoog:
> @zmoog - Can we add changelog entry?

Added.

packages/azure/data_stream/platformlogs/elasticsearch/ingest_pipeline/default.yml
zmoog: Parsing `timestamp` with a `date` processor.

chemamartinez: Sounds good, thanks!

zmoog:
> * As we are using sample event from `FunctionappLogs` category for validations. Is there a possibility to get different response format for other categories?

I confirm Azure is sending the same invalid field names (`timestamp`, `azureLogForwarder`, etc) for other categories as well.

image

PR #14962, opened 2025-08-19
o365: re-write CEL code to use work lists

### LINKED ISSUE #14735: o365: rewrite agent CEL collector code to use work lists

efd6:
## Proposed commit message
```
o365: re-write CEL code to use work lists

This reduces the per-execution maximum memory requirements since only
one content collection is held in memory at a time. It also reduces API
request failure brittleness since only a single content collection will
fail in the case that the request fails.

In practice, it also reduces network load by apparently around 90%;
experiments with the previous code show that it repeatedly recollects
data from the API multiple times — the test mock in the package shows
five of the events are collected 12 time, two are collected 24 times and
one 36 times. This does not happen nearly to the same extent with the
new code, seven events are collected once and one is collected three
times. That it happens at all is concerning, and should be revisited.
```

> [!NOTE]
> ### Ordering of events:
> The current code in this change assumes an ascending ordering of content creation times. I don't know if this is an invariant that we can rely on, but the assumption makes the code more efficient. If it is not something that we can rely on the change is reasonably straightforward (untried at this stage, but I feel it in my giblets).
>
> ### Pathological recollection:
> ```
> $ mito -data data.json src.cel|jq -c '.events|.[]'|sort|uniq -c|awk '{print $1}'|sort|uniq -c
>       5 12
>       2 24
>       1 36
> $ mito -data data.json src_new.cel|jq -c '.events|.[]'|sort|uniq -c|awk '{print $1}'|sort|uniq -c
>       7 1
>       1 3
> ```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #14735


packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,226 +75,344 @@ redact:
...
+                          "contentType": [work.curr_type],
+                          "PublisherIdentifier": [state.base.tenant_id],
+                        }.format_query(),
+                      ]
+                    )
+                  ).do_request().as(sub_resp, (sub_resp.StatusCode == 200) ?
+                    sub_resp.Body.decode_json().as(sub,
+                      {
+                        "type": sub.contentType,
+                        "enabled": sub.?status == optional.of("enabled") || sub.?error == optional.of("AF20024"),
```
kcreddy: In the case of `AF20024`, the response is `400 Bad Request`. Also it should check for `?error.code` based on the response:
```json
{
    "error": {
        "code": "AF20024",
        "message": "The subscription is already enabled. No property change."
    }
}
```

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,226 +75,344 @@ redact:
...
+                        {
+                          "contentType": [work.curr_type],
+                          "PublisherIdentifier": [state.base.tenant_id],
+                        }.format_query(),
+                      ]
+                    )
+                  ).do_request().as(sub_resp, (sub_resp.StatusCode == 200) ?
+                    sub_resp.Body.decode_json().as(sub,
+                      {
+                        "type": sub.contentType,
```
kcreddy: Do we need this, because we already seem to carry the content type inside `work.curr_type`.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,226 +75,344 @@ redact:
...
-        [
-          state.url.trim_right("/"),
-          state.base.tenant_id,
-          {
-            "contentType": [content_type],
-            "PublisherIdentifier": [state.base.tenant_id],
-          }.format_query(),
-        ]
+  state.with(
+    state.?work.todo_content[0].hasValue() ?
```
kcreddy, line 79: Just for my understanding, is this functionally similar to:
`has(state.?work.todo_content[?0]) ?` or simply `has(state.?work.todo_content[0]) ?`

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,226 +75,344 @@ redact:
...
+              {
+                "cursor": state.cursor.drop("content_types_state_as_list").with(
+                  {
+                    // Use zip to avoid requiring a higher agent version.
+                    "last_for": zip(
+                      // Lowercase key since test data has this mixed. w00t!1
+                      state.cursor.content_types_state_as_list.map(e, e.content_type.to_lower()),
+                      state.cursor.content_types_state_as_list.map(e, e.content_created_at)
+                    ),
+                  }
```
kcreddy, line 332: Should the `.drop("content_types_state_as_list")` happen after the calculation of `last_for` as I see `content_types_state_as_list` is being used in its calculation ?

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: In `state.?work.todo_content[?0]` you only need the first `?`, the other one is automatic.

The `.hasValue()` call checks if there's a value in the option. The `has()` macro does a different type of field checking. There are cases where you can use either, but it's not always the case.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,226 +75,344 @@ redact:
...
-            "PublisherIdentifier": [state.base.tenant_id],
-          }.format_query(),
-        ]
+  state.with(
+    state.?work.todo_content[0].hasValue() ?
+      // Do any outstanding content collection to do.
+      state.work.todo_content[0].as(l,
+        // This depends on state.work.todo_content being ordered ascending by contentCreated.
+        // If this is not the case, this needs to be replaced with a linear scan for the
+        // oldest, putting the oldest in the link and the rest in the remaining todo list.
```
chrisberkhout: I think `state.work.todo_content` doesn't necessarily have to be exactly in order. As long as the `batch_interval` (1h) windows are in order, there's time to get it all. They should live for 7 days.

However, getting the exact right `state.cursor.last_for` value is important to avoid duplicates.

The [doc](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference#list-available-content) says:
> The content will be listed in the order in which the aggregations become available

Probably we shouldn't assume that `contentCreated` and "became available" are the same.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,226 +75,344 @@ redact:
...
+          e.?contentExpiration.optMap(t, t.parse_time(time_layout.RFC3339) >= now()).orValue(true)
+        ).as(list,
+          {
+            "cursor": state.?cursor.orValue({}).with(
+              {
+                "last_for": state.?cursor.last_for.orValue({}).with(
+                  dyn(
+                    (
+                      list[?0].optMap(l,
+                        {l.contentType.to_lower(): l.contentCreated}
```
chrisberkhout: I think this would be a bit more robust if it handled the case of the link leading to an empty page.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,226 +75,344 @@ redact:
...
+              ).do_request().as(list_resp, (list_resp.StatusCode == 200) ?
+                list_resp.Body.decode_json().filter(e,
+                  e.?contentExpiration.optMap(t, t.parse_time(time_layout.RFC3339) >= now()).orValue(true)
+                ).as(list,
+                  {
+                    "cursor": state.?cursor.orValue({}).with(
+                      {
+                        "last_for": state.?cursor.last_for.orValue({}).with(
+                          // Use now to ensure we bracket the collection.
+                          {state.work.curr_type.to_lower(): now}
```
chrisberkhout: If it runs and checks the first hour (7 days ago), won't this finish with `last_for` saying it's seen everything up to `now`? (I see if there's a link to follow it will update this based on data returned from that on the next run)

I'm concerned about it's ability to advance hour by hour when there's no data.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: The code should be a direct translation of [this](https://github.com/elastic/integrations/blob/771a844e525392520907c0cfd3099ff4e23aeff6/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs#L94). It should be
```
"enabled": sub.?status == optional.of("enabled") || sub.?error.code == optional.of("AF20024")
```
Thanks.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: There is more state, and more redundant state retained in the new implementation to aid debugging. This is intentional. Even with this additional work we are winning in terms of work done compared to the old implementation. We can remove things in the future if we feel it is worth it.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: Chris' answer is the case.

Facetiously, yes, the two of `has(state.?work.todo_content[?0])` and `has(state.?work.todo_content[0])` are the same, they both are an error (because `has` only checks for field presence, not the more general value existence). Less facetiously, `state.?work.todo_content[?0]` and `state.?work.todo_content[0]` are the same since once a optional type is constructed using a `.?` operator, the type continues to be an optional type (the model differs from the Java `?.` operator which evaluates to the field or `null`, because Java — incorrectly — conflates `null` with absence). Indexing into an optional type holding an array is safe and will return either `optional.of(e)` (where `e` is the element indexed to) or `optional.none()`.

So here uses `.hasValue()` to see whether the optional type can be dereferenced to a value.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: No, the `state` here in the `zip` call is the the `state` value labelled at [L270](https://github.com/elastic/integrations/pull/14962/files#diff-562ac389f34f5d7b13b142acb0eaa2282105f9932df5253a6cf7e9e43570e20aR270).

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: Yeah, this was my concern.

What is that link intended to point to; it points to this PR currently.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: I am not convinced that this is the case; in the situation that the page is empty, no event will be published and so there will be no cursor update. On the next periodic evaluation the old (possibly staled-out-of-range) cursor will be clamped to the API's query constraints. The logic that corresponds to what you are asking for in the original code is required because it bundles successes (which will update cursors) with empty sets (which won't have correct cursors).

I may be wrong here, so I'd like to hear what you think based on the position above.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: This should be conformed to the API's constraints at the time of the call. If I haven't done that, I should.

efd6: No, I see what you mean. I should hoist the `time_range` so that it is visible here and then use `time_range.end` instead of `now`.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: This is the primitive that I felt would work:
```
-- src.cel --
state.work.todo_content.as(list,
	list.map(e,
		timestamp(e.contentExpiration) > now(),
		timestamp(e.contentCreated)
	).min().as(min,
		list.filter(e, timestamp(e.contentCreated) == min)[0]
	).as(oldest,
		{
			"head": oldest,
			"tail": list.filter(e, e.contentId != oldest.contentId || e.contentUri != oldest.contentUri),
		}
	)
)
-- data.json --
{
	"work": {
		"todo_content": [
			{
				"contentCreated": "2021-08-18T22:06:51.076390384Z",
				"contentExpiration": "2055-04-30T17:35:00.000Z",
				"contentId": "1",
				"contentType": "Audit.Sharepoint",
				"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
			},
			{
				"contentCreated": "2020-08-18T22:06:51.076390384Z",
				"contentExpiration": "2055-04-30T17:35:00.000Z",
				"contentId": "0",
				"contentType": "Audit.Sharepoint",
				"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
			},
			{
				"contentCreated": "2023-08-18T22:06:51.076390384Z",
				"contentExpiration": "2055-04-30T17:35:00.000Z",
				"contentId": "3",
				"contentType": "Audit.Sharepoint",
				"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
			},
			{
				"contentCreated": "2022-08-18T22:06:51.076390384Z",
				"contentExpiration": "2055-04-30T17:35:00.000Z",
				"contentId": "2",
				"contentType": "Audit.Sharepoint",
				"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
			}
		]
	}
}
-- out.json --
{
	"head": {
		"contentCreated": "2020-08-18T22:06:51.076390384Z",
		"contentExpiration": "2055-04-30T17:35:00.000Z",
		"contentId": "0",
		"contentType": "Audit.Sharepoint",
		"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
	},
	"tail": [
		{
			"contentCreated": "2021-08-18T22:06:51.076390384Z",
			"contentExpiration": "2055-04-30T17:35:00.000Z",
			"contentId": "1",
			"contentType": "Audit.Sharepoint",
			"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
		},
		{
			"contentCreated": "2023-08-18T22:06:51.076390384Z",
			"contentExpiration": "2055-04-30T17:35:00.000Z",
			"contentId": "3",
			"contentType": "Audit.Sharepoint",
			"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
		},
		{
			"contentCreated": "2022-08-18T22:06:51.076390384Z",
			"contentExpiration": "2055-04-30T17:35:00.000Z",
			"contentId": "2",
			"contentType": "Audit.Sharepoint",
			"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
		}
	]
}
```

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
kcreddy: Sorry I didn't link the line above it in my review. 

What I meant is that the conditional: `sub.?error.code == optional.of("AF20024")` is reached after the conditional on L245 i.e., `(sub_resp.StatusCode == 200) ?` succeeds. But they are mutually exclusive because `AF20024` error happens with `StatusCode` `400`.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
kcreddy: Thanks for the clarifications 😄

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
kcreddy: > What is that link intended to point to; it points to this PR currently.

The link: https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference#list-available-content

chrisberkhout: > What is that link intended to point to; it points to this PR currently.

[Office 365 Management Activity API reference > List available Content](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference#list-available-content)

chrisberkhout: > This is the primitive that I felt would work:

The one issue I see in that code is that it would accumulate expired items. The expired items should be filtered out of `list` not just when working out the `min`.

chrisberkhout: Actually, I think this isn't necessary, because `last_for` doesn't need to be updated here. It should just get the `time_range.end` value.

If that's the case, the only issue is whether it's doing the oldest thing in the work list first, which is not critical because it won't affect query time ranges. Tiny differences in created vs available won't matter and even if things are out of order within an hour window it won't really matter.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: I was thinking it would update `last_for` when there's data and clear it otherwise. I see now that if there's no data it will keep the existing value, so that concern is resolved.

But now I think that `last_for` should only be updated with `time_range.end` (and in the once-off rewriting of old cursor values), so not here and not above during content collection. It's enough to use `time_range.end`, and it avoids potential issues with ordering.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,226 +75,343 @@ redact:
...
-          ]
-      )
-    )
-  ).flatten().drop_empty().as(events_list,
-    events_list.collate("events_per_content_type").as(events,
-      events.filter(e, has(e.?error.message)).as(errors,
-        (errors.size() > 0) ?
-          // If there are errors, return a single merged error event and discard any other results
-          {
-            "url": state.url,
```
chrisberkhout, line 275: Currently, the most work it will do in one evaluation is to subscribe to a content type, list content in a time range for that content type, then fetch the first chunk of that content.

That can produce no events but still have more work to do.

Normally it's simpler because we aren't restricted to a max 24h time range. The time range can be open-ended and will either get data or be done. In this case if we had an open-ended time range, we'd have more content types to handle anyway.

So I think we need to keep working until we're really done, regardless of whether a particular step produces events or not, which means if we want more and don't have real events, we need to add a fake event that will be dropped by the pipeline, to make sure it continues immediately.

If that's done, it's also not necessary to do multiple types of requests in one evaluation, and I think breaking those out would make it a bit easier to read. Each run could do one of: fetch content, follow listing link, make listing request, subscribe.

We could also do fewer of the subscription requests, maybe with a flag in state so they are done once on startup.

chrisberkhout: Also it should keep going on a content type until it is up to date. I don't think that's currently happening.

For users it might be slightly nicer if it does one time range per content type and then they all advance together, but I think it's okay to get one type fully up to date before doing the next if the implementation is easier.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,226 +75,343 @@ redact:
...
-                      :
+                // Calculate start and end time stamps.
+                (has(state.?cursor.last_for) && state.work.curr_type.to_lower() in state.cursor.last_for) ?
+                  // Clamp to 7 days, the maximum age limit.
+                  max(
+                    now() - duration(state.base.maximum_age),
+                    state.cursor.last_for[state.work.curr_type.to_lower()].parse_time(time_layout.RFC3339)
+                  ).as(created_at,
+                    {
+                      "start": created_at + duration("1s"),
```
chrisberkhout: I think the `+ duration("1s")` is to avoid overlap with the last request. Maybe it can be a millisecond or microsecond?

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: > That can produce no events but still have more work to do.

This is true, in the final block, we should test whether events is empty when we want to do more work, and if it is, fill it with a retry dummy.

> Also it should keep going on a content type until it is up to date. I don't think that's currently happening.

Why do you say that? It's not mandated by the code, but that behaviour emerges. Either way, I'm not convinced that that is something that matters.

> For users it might be slightly nicer if it does one time range per content type and then they all advance together, but I think it's okay to get one type fully up to date before doing the next if the implementation is easier.

I believe the complexity (and likely the pathological behaviour) of the previous implementation comes from trying to do this.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: > The one issue I see in that code is that it would accumulate expired items. The expired items should be filtered out of `list` not just when working out the `min`.

This is the primitive. Where `tail` goes is up to the context of the primitive's use. This is a list consumer, so it is guaranteed to completely consume the list eventually so long as the data source is finite. This is the same as the simpler implementation that is in the change as it stands, mod work order.

> Actually, I think this isn't necessary, because `last_for` doesn't need to be updated here. It should just get the `time_range.end` value.

That would also work, but it would result in potentially recollecting data in the case of failures.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: See https://github.com/elastic/integrations/pull/14962#discussion_r2289311614 re recollecting data.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: Are you happy with
```diff
diff --git a/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs b/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
index 80e7f45fb2..64e6ec7c06 100644
--- a/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
+++ b/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
@@ -242,7 +242,7 @@ program: |-
                         }.format_query(),
                       ]
                     )
-                  ).do_request().as(sub_resp, (sub_resp.StatusCode == 200) ?
+                  ).do_request().as(sub_resp, (sub_resp.StatusCode == 200 || sub_resp.StatusCode == 400) ?
                     sub_resp.Body.decode_json().as(sub,
                       {
                         "type": sub.contentType,
```

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: The primitive in place
```diff
diff --git a/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs b/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
index 64e6ec7c06..877981b102 100644
--- a/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
+++ b/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
@@ -78,27 +78,37 @@ program: |-
   state.with(
     state.?work.todo_content[0].hasValue() ?
       // Do any outstanding content collection to do.
-      state.work.todo_content[0].as(l,
-        // This depends on state.work.todo_content being ordered ascending by contentCreated.
-        // If this is not the case, this needs to be replaced with a linear scan for the
-        // oldest, putting the oldest in the link and the rest in the remaining todo list.
+      state.work.todo_content.as(list,
+        list.map(e,
+          // Filter out stale elements.
+          timestamp(e.contentExpiration) > now(),
+          timestamp(e.contentCreated)
+        ).min().as(min,
+          list.filter(e, timestamp(e.contentCreated) == min)[0]
+        ).as(oldest,
+          {
+            "link": oldest,
+            "todo_content": list.filter(e, e.contentUri != oldest.contentUri),
+          }
+        )
+      ).as(w,
         {
           "cursor": state.?cursor.orValue({}).with(
             {
               "last_for": state.?cursor.last_for.orValue({}).with(
-                {l.contentType.to_lower(): l.contentCreated}
+                {w.link.contentType.to_lower(): w.link.contentCreated}
               ),
             }
           ),
           "work": state.work.with(
             {
-              "link": l, // debugging value only
-              "todo_content": tail(state.work.todo_content),
+              "link": w.link, // debugging value only
+              "todo_content": w.todo_content,
             }
           ),
           "events": request(
             "GET",
-            l.contentUri
+            w.link.contentUri
           ).do_request().as(content_resp, (content_resp.StatusCode == 200) ?
             dyn(
               content_resp.Body.decode_json().map(e,
@@ -113,7 +123,7 @@ program: |-
               "error": {
                 "code": string(content_resp.StatusCode),
                 "id": string(content_resp.Status),
-                "message": "GET " + l.contentUri + ": " + (
+                "message": "GET " + w.link.contentUri + ": " + (
                   (size(content_resp.Body) != 0) ?
                     string(content_resp.Body)
                   :
@@ -133,13 +143,32 @@ program: |-
         list_resp.Body.decode_json().filter(e,
           e.?contentExpiration.optMap(t, t.parse_time(time_layout.RFC3339) >= now()).orValue(true)
         ).as(list,
+          (list + state.work.todo_content).as(list,
+            list.map(e,
+              // Filter out stale elements.
+              timestamp(e.contentExpiration) > now(),
+              timestamp(e.contentCreated)
+            ).min().as(min,
+              list.filter(e, timestamp(e.contentCreated) == min)[?0]
+            ).as(oldest,
+              {
+                "link": oldest,
+                "todo_content": (
+                  oldest.optMap(oldest,
+                    list.filter(e, e.contentUri != oldest.contentUri)
+                  )
+                ).orValue(list),
+              }
+            )
+          )
+        ).as(w,
           {
             "cursor": state.?cursor.orValue({}).with(
               {
                 "last_for": state.?cursor.last_for.orValue({}).with(
                   dyn(
                     (
-                      list[?0].optMap(l,
+                      w.link.optMap(l,
                         {l.contentType.to_lower(): l.contentCreated}
                       )
                     ).orValue({})
@@ -149,16 +178,16 @@ program: |-
             ),
             "work": state.work.drop(["link", "next_list"]).with(
               {
-                ?"link": list[?0], // debugging value only
+                ?"link": w.link, // debugging value only
                 ?"next_list": [
                   // Work around non-conformance with HTTP header MIME standards.
                   ?list_resp.Header.?Nextpageuri[0],
                   ?list_resp.Header.?NextPageUri[0],
                 ][?0],
-                "todo_content": state.work.todo_content + tail(list),
+                "todo_content": w.todo_content,
               }
             ),
-            "events": list[?0].optMap(l,
+            "events": w.link.optMap(l,
               request(
                 "GET",
                 l.contentUri
```
Note that it's not necessary to do this in the initial collection because of the documented behaviour.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
kcreddy: Yes

kcreddy (approved):
Thanks!

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: Not sure we'll need it but...

If I adjust the primitive like this to run in a loop and have an expired item, it will never consume the expired item. It fails where there is no non-expired item left from which to select a `contentCreated` value.

```diff
@@ -1,82 +1,129 @@
 -- src.cel --
-state.work.todo_content.as(list,
+state.tail.as(list,
 	list.map(e,
 		timestamp(e.contentExpiration) > now(),
 		timestamp(e.contentCreated)
 	).min().as(min,
 		list.filter(e, timestamp(e.contentCreated) == min)[0]
 	).as(oldest,
 		{
 			"head": oldest,
 			"tail": list.filter(e, e.contentId != oldest.contentId || e.contentUri != oldest.contentUri),
-		}
+		}.as(result, result.with({"want_more": size(result.tail) > 0}))
 	)
 )
 -- data.json --
 {
-	"work": {
-		"todo_content": [
+	"tail": [
 			{
 				"contentCreated": "2021-08-18T22:06:51.076390384Z",
-				"contentExpiration": "2055-04-30T17:35:00.000Z",
+				"contentExpiration": "2005-04-30T17:35:00.000Z",
 				"contentId": "1",
 				"contentType": "Audit.Sharepoint",
 				"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
 			},
 			{
 				"contentCreated": "2020-08-18T22:06:51.076390384Z",
 				"contentExpiration": "2055-04-30T17:35:00.000Z",
 				"contentId": "0",
 				"contentType": "Audit.Sharepoint",
 				"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
 			},
 			{
 				"contentCreated": "2023-08-18T22:06:51.076390384Z",
 				"contentExpiration": "2055-04-30T17:35:00.000Z",
 				"contentId": "3",
 				"contentType": "Audit.Sharepoint",
 				"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
 			},
 			{
 				"contentCreated": "2022-08-18T22:06:51.076390384Z",
 				"contentExpiration": "2055-04-30T17:35:00.000Z",
 				"contentId": "2",
 				"contentType": "Audit.Sharepoint",
 				"contentUri": "http://localhost:9001/api/v1.0/celsp1/activity/feed/audit/celid2b"
 			}
 		]
-	}
 }
```

It works if the filter is applied to the list in general, not just when choosing a min `contentCreated`:

```diff
@@ -1,129 +1,103 @@
 -- src.cel --
-state.tail.as(list,
+state.tail.filter(e, timestamp(e.contentExpiration) > now()).as(list,
        list.map(e,
-               timestamp(e.contentExpiration) > now(),
                timestamp(e.contentCreated)
        ).min().as(min,
                list.filter(e, timestamp(e.contentCreated) == min)[0]
        ).as(oldest,
                {
                        "head": oldest,
                        "tail": list.filter(e, e.contentId != oldest.contentId || e.contentUri != oldest.contentUri),
                }.as(result, result.with({"want_more": size(result.tail) > 0}))
        )
 )
```

chrisberkhout: > > Actually, I think this isn't necessary, because last_for doesn't need to be updated here. It should just get the time_range.end value.
> 
> That would also work, but it would result in potentially recollecting data in the case of failures.

I don't see why it would recollect data if there's one place where the query time range is set.

If I list data that became available in 2pm-3pm and save 3pm in `last_for`, then my next listing request should start from 3pm, regardless of what times are in the fetched data.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: > > That can produce no events but still have more work to do.
> 
> This is true, in the final block, we should test whether events is empty when we want to do more work, and if it is, fill it with a retry dummy.

:+1: 

> > For users it might be slightly nicer if it does one time range per content type and then they all advance together, but I think it's okay to get one type fully up to date before doing the next if the implementation is easier.
> 
> I believe the complexity (and likely the pathological behaviour) of the previous implementation comes from trying to do this.

Yeah, that may be the case. I think it's okay to finish one content type before starting the next.

> > Also it should keep going on a content type until it is up to date. I don't think that's currently happening.
> 
> Why do you say that? It's not mandated by the code, but that behaviour emerges. Either way, I'm not convinced that that is something that matters.

... (to discuss on call)

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -131,35 +134,57 @@ program: |-
...
+            list.map(e,
+              timestamp(e.endTime)
+            ).min().as(min,
+              list.filter(e, timestamp(e.endTime) == min)[?0]
+            ).as(oldest,
+              {
+                "curr_content": oldest,
+                "todo_content": list.filter(e, e.contentUri != oldest.contentUri),
+              }
+            )
```
efd6: We need to do this here since we have a mixed list. We do not need to do this at L382 since all elements of the list have the same `endTime`.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,227 +75,413 @@ redact:
...
+      // even if we did not get any content for the query period.
+      timestamp(state.cursor.last_for[state.work.curr_type.to_lower()]) < now - duration("1h") &&
+      !state.work.todo_type.exists(t, t == state.work.curr_type)
+    ) ?
+      state.with(
+        {
+          "work": state.work.with({"todo_type": [state.work.curr_type] + state.work.todo_type}),
+        }
       )
+    :
```
chrisberkhout, line 467: I think this should take `next_list` into account.

I ran it against my mock server and it checked enough time windows, but didn't get the last page of the last one. It finished like this:

```
{
	"base": {
		"batch_interval": "1h",
		"content_types": "Audit.AzureActiveDirectory, Audit.Exchange",
		"list_contents_start_time": "12h",
		"maximum_age": "167h55m",
		"tenant_id": "sometenantid"
	},
	"cursor": {
		"last_for": {
			"audit.azureactivedirectory": "2025-08-26T16:12:04.443773725Z",
			"audit.exchange": "2025-08-26T16:12:50.249959671Z"
		}
	},
	"events": [
		{
			"event": {
				"original": "{\"CreationTime\":\"2025-08-21T07:17:07.010667854Z\",\"Id\":\"1SYhkg0hcg6D\"}"
			},
			"o365audit": {
				"CreationTime": "2025-08-21T07:17:07.010667854Z",
				"Id": "1SYhkg0hcg6D"
			}
		},
		{
			"event": {
				"original": "{\"CreationTime\":\"2025-08-22T21:32:28.132642086Z\",\"Id\":\"chNYbFKonjEm\"}"
			},
			"o365audit": {
				"CreationTime": "2025-08-22T21:32:28.132642086Z",
				"Id": "chNYbFKonjEm"
			}
		}
	],
	"url": "http://localhost:8888",
	"want_more": false,
	"work": {
		"curr_content": {
			"contentCreated": "2025-08-26T16:04:38.09546882Z",
			"contentExpiration": "2025-09-02T23:04:38.09546882Z",
			"contentId": "7U4MYYzyYLqZ",
			"contentType": "Audit.Exchange",
			"contentUri": "http://localhost:8888/api/v1.0/sometenantid/activity/feed/audit/7U4MYYzyYLqZ",
			"endTime": "2025-08-26T16:12:50.249959671Z"
		},
		"curr_type": "Audit.Exchange",
		"next_list": "http://localhost:8888/api/v1.0/sometenantid/activity/feed/subscriptions/content?PublisherIdentifier=sometenantid&contentType=Audit.Exchange&endTime=2025-08-26T16%3A12%3A50.249959671Z&nextPage=20&startTime=2025-08-26T15%3A12%3A50.250959671Z",
		"sub": {
			"enabled": true,
			"type": "Audit.Exchange"
		},
		"todo_content": [],
		"todo_type": []
	}
}
```

chrisberkhout (commented):
The [mock server](https://gist.github.com/chrisberkhout/4d19d5c53ff82021e939908f7a6838c2) will now serve a lot of data and do a lot of checks.

I ran it with random data over the last 12h and its logging looked like this (I filtered out the fetch calls):

```
2025/08/26 18:12:02 RunStart StartTime=2025-08-26T16:12:02.895535488Z, scenarioName=2_types_random_250_to_500_last_12h_pages_of_20, seed=1756224722895535488
2025/08/26 18:12:04 Subscribe Audit.AzureActiveDirectory, result 400
2025/08/26 18:12:04 List Audit.AzureActiveDirectory [2025-08-26T04:12:04.443770925Z--2025-08-26T05:12:04.443773725Z] (1h0m0.0000028s) nextPage=, returning 20, remaining 11
2025/08/26 18:12:05 List Audit.AzureActiveDirectory [2025-08-26T04:12:04.443770925Z--2025-08-26T05:12:04.443773725Z] (1h0m0.0000028s) nextPage=20, returning 11, remaining 0
2025/08/26 18:12:06 Subscribe Audit.AzureActiveDirectory, result 400
2025/08/26 18:12:06 List Audit.AzureActiveDirectory [2025-08-26T05:12:04.444773725Z--2025-08-26T06:12:04.443773725Z] (59m59.999s) nextPage=, returning 20, remaining 9
2025/08/26 18:12:09 List Audit.AzureActiveDirectory [2025-08-26T05:12:04.444773725Z--2025-08-26T06:12:04.443773725Z] (59m59.999s) nextPage=20, returning 9, remaining 0
2025/08/26 18:12:10 Subscribe Audit.AzureActiveDirectory, result 400
2025/08/26 18:12:10 List Audit.AzureActiveDirectory [2025-08-26T06:12:04.444773725Z--2025-08-26T07:12:04.443773725Z] (59m59.999s) nextPage=, returning 20, remaining 17
2025/08/26 18:12:13 List Audit.AzureActiveDirectory [2025-08-26T06:12:04.444773725Z--2025-08-26T07:12:04.443773725Z] (59m59.999s) nextPage=20, returning 17, remaining 0
2025/08/26 18:12:15 Subscribe Audit.AzureActiveDirectory, result 400
2025/08/26 18:12:15 List Audit.AzureActiveDirectory [2025-08-26T07:12:04.444773725Z--2025-08-26T08:12:04.443773725Z] (59m59.999s) nextPage=, returning 20, remaining 5
2025/08/26 18:12:17 List Audit.AzureActiveDirectory [2025-08-26T07:12:04.444773725Z--2025-08-26T08:12:04.443773725Z] (59m59.999s) nextPage=20, returning 5, remaining 0
2025/08/26 18:12:18 Subscribe Audit.AzureActiveDirectory, result 400
2025/08/26 18:12:18 List Audit.AzureActiveDirectory [2025-08-26T08:12:04.444773725Z--2025-08-26T09:12:04.443773725Z] (59m59.999s) nextPage=, returning 20, remaining 3
2025/08/26 18:12:21 List Audit.AzureActiveDirectory [2025-08-26T08:12:04.444773725Z--2025-08-26T09:12:04.443773725Z] (59m59.999s) nextPage=20, returning 3, remaining 0
2025/08/26 18:12:21 Subscribe Audit.AzureActiveDirectory, result 400
2025/08/26 18:12:21 List Audit.AzureActiveDirectory [2025-08-26T09:12:04.444773725Z--2025-08-26T10:12:04.443773725Z] (59m59.999s) nextPage=, returning 20, remaining 9
2025/08/26 18:12:24 List Audit.AzureActiveDirectory [2025-08-26T09:12:04.444773725Z--2025-08-26T10:12:04.443773725Z] (59m59.999s) nextPage=20, returning 9, remaining 0
2025/08/26 18:12:24 Subscribe Audit.AzureActiveDirectory, result 400
2025/08/26 18:12:24 List Audit.AzureActiveDirectory [2025-08-26T10:12:04.444773725Z--2025-08-26T11:12:04.443773725Z] (59m59.999s) nextPage=, returning 20, remaining 10
2025/08/26 18:12:27 List Audit.AzureActiveDirectory [2025-08-26T10:12:04.444773725Z--2025-08-26T11:12:04.443773725Z] (59m59.999s) nextPage=20, returning 10, remaining 0
2025/08/26 18:12:28 Subscribe Audit.AzureActiveDirectory, result 400
2025/08/26 18:12:28 List Audit.AzureActiveDirectory [2025-08-26T11:12:04.444773725Z--2025-08-26T12:12:04.443773725Z] (59m59.999s) nextPage=, returning 20, remaining 17
2025/08/26 18:12:31 List Audit.AzureActiveDirectory [2025-08-26T11:12:04.444773725Z--2025-08-26T12:12:04.443773725Z] (59m59.999s) nextPage=20, returning 17, remaining 0
2025/08/26 18:12:33 Subscribe Audit.AzureActiveDirectory, result 200
2025/08/26 18:12:33 List Audit.AzureActiveDirectory [2025-08-26T12:12:04.444773725Z--2025-08-26T13:12:04.443773725Z] (59m59.999s) nextPage=, returning 20, remaining 19
2025/08/26 18:12:36 List Audit.AzureActiveDirectory [2025-08-26T12:12:04.444773725Z--2025-08-26T13:12:04.443773725Z] (59m59.999s) nextPage=20, returning 19, remaining 0
2025/08/26 18:12:38 Subscribe Audit.AzureActiveDirectory, result 400
2025/08/26 18:12:38 List Audit.AzureActiveDirectory [2025-08-26T13:12:04.444773725Z--2025-08-26T14:12:04.443773725Z] (59m59.999s) nextPage=, returning 20, remaining 12
2025/08/26 18:12:41 List Audit.AzureActiveDirectory [2025-08-26T13:12:04.444773725Z--2025-08-26T14:12:04.443773725Z] (59m59.999s) nextPage=20, returning 12, remaining 0
2025/08/26 18:12:42 Subscribe Audit.AzureActiveDirectory, result 400
2025/08/26 18:12:42 List Audit.AzureActiveDirectory [2025-08-26T14:12:04.444773725Z--2025-08-26T15:12:04.443773725Z] (59m59.999s) nextPage=, returning 20, remaining 14
2025/08/26 18:12:45 List Audit.AzureActiveDirectory [2025-08-26T14:12:04.444773725Z--2025-08-26T15:12:04.443773725Z] (59m59.999s) nextPage=20, returning 14, remaining 0
2025/08/26 18:12:46 Subscribe Audit.AzureActiveDirectory, result 400
2025/08/26 18:12:46 List Audit.AzureActiveDirectory [2025-08-26T15:12:04.444773725Z--2025-08-26T16:12:04.443773725Z] (59m59.999s) nextPage=, returning 20, remaining 8
2025/08/26 18:12:49 List Audit.AzureActiveDirectory [2025-08-26T15:12:04.444773725Z--2025-08-26T16:12:04.443773725Z] (59m59.999s) nextPage=20, returning 8, remaining 0
2025/08/26 18:12:50 Subscribe Audit.Exchange, result 400
2025/08/26 18:12:50 List Audit.Exchange [2025-08-26T04:12:50.249957722Z--2025-08-26T05:12:50.249959671Z] (1h0m0.000001949s) nextPage=, returning 20, remaining 6
2025/08/26 18:12:52 List Audit.Exchange [2025-08-26T04:12:50.249957722Z--2025-08-26T05:12:50.249959671Z] (1h0m0.000001949s) nextPage=20, returning 6, remaining 0
2025/08/26 18:12:53 Subscribe Audit.Exchange, result 400
2025/08/26 18:12:53 List Audit.Exchange [2025-08-26T05:12:50.250959671Z--2025-08-26T06:12:50.249959671Z] (59m59.999s) nextPage=, returning 20, remaining 9
2025/08/26 18:12:55 List Audit.Exchange [2025-08-26T05:12:50.250959671Z--2025-08-26T06:12:50.249959671Z] (59m59.999s) nextPage=20, returning 9, remaining 0
2025/08/26 18:12:56 Subscribe Audit.Exchange, result 200
2025/08/26 18:12:56 List Audit.Exchange [2025-08-26T06:12:50.250959671Z--2025-08-26T07:12:50.249959671Z] (59m59.999s) nextPage=, returning 20, remaining 4
2025/08/26 18:12:59 List Audit.Exchange [2025-08-26T06:12:50.250959671Z--2025-08-26T07:12:50.249959671Z] (59m59.999s) nextPage=20, returning 4, remaining 0
2025/08/26 18:12:59 Subscribe Audit.Exchange, result 200
2025/08/26 18:12:59 List Audit.Exchange [2025-08-26T07:12:50.250959671Z--2025-08-26T08:12:50.249959671Z] (59m59.999s) nextPage=, returning 20, remaining 13
2025/08/26 18:13:02 List Audit.Exchange [2025-08-26T07:12:50.250959671Z--2025-08-26T08:12:50.249959671Z] (59m59.999s) nextPage=20, returning 13, remaining 0
2025/08/26 18:13:03 Subscribe Audit.Exchange, result 200
2025/08/26 18:13:03 List Audit.Exchange [2025-08-26T08:12:50.250959671Z--2025-08-26T09:12:50.249959671Z] (59m59.999s) nextPage=, returning 20, remaining 12
2025/08/26 18:13:06 List Audit.Exchange [2025-08-26T08:12:50.250959671Z--2025-08-26T09:12:50.249959671Z] (59m59.999s) nextPage=20, returning 12, remaining 0
2025/08/26 18:13:07 Subscribe Audit.Exchange, result 200
2025/08/26 18:13:07 List Audit.Exchange [2025-08-26T09:12:50.250959671Z--2025-08-26T10:12:50.249959671Z] (59m59.999s) nextPage=, returning 20, remaining 16
2025/08/26 18:13:10 List Audit.Exchange [2025-08-26T09:12:50.250959671Z--2025-08-26T10:12:50.249959671Z] (59m59.999s) nextPage=20, returning 16, remaining 0
2025/08/26 18:13:12 Subscribe Audit.Exchange, result 400
2025/08/26 18:13:12 List Audit.Exchange [2025-08-26T10:12:50.250959671Z--2025-08-26T11:12:50.249959671Z] (59m59.999s) nextPage=, returning 20, remaining 18
2025/08/26 18:13:14 List Audit.Exchange [2025-08-26T10:12:50.250959671Z--2025-08-26T11:12:50.249959671Z] (59m59.999s) nextPage=20, returning 18, remaining 0
2025/08/26 18:13:17 Subscribe Audit.Exchange, result 400
2025/08/26 18:13:17 List Audit.Exchange [2025-08-26T11:12:50.250959671Z--2025-08-26T12:12:50.249959671Z] (59m59.999s) nextPage=, returning 20, remaining 17
2025/08/26 18:13:19 List Audit.Exchange [2025-08-26T11:12:50.250959671Z--2025-08-26T12:12:50.249959671Z] (59m59.999s) nextPage=20, returning 17, remaining 0
2025/08/26 18:13:21 Subscribe Audit.Exchange, result 400
2025/08/26 18:13:21 List Audit.Exchange [2025-08-26T12:12:50.250959671Z--2025-08-26T13:12:50.249959671Z] (59m59.999s) nextPage=, returning 20, remaining 20
2025/08/26 18:13:24 List Audit.Exchange [2025-08-26T12:12:50.250959671Z--2025-08-26T13:12:50.249959671Z] (59m59.999s) nextPage=20, returning 20, remaining 0
2025/08/26 18:13:27 Subscribe Audit.Exchange, result 200
2025/08/26 18:13:27 List Audit.Exchange [2025-08-26T13:12:50.250959671Z--2025-08-26T14:12:50.249959671Z] (59m59.999s) nextPage=, returning 20, remaining 8
2025/08/26 18:13:30 List Audit.Exchange [2025-08-26T13:12:50.250959671Z--2025-08-26T14:12:50.249959671Z] (59m59.999s) nextPage=20, returning 8, remaining 0
2025/08/26 18:13:30 Subscribe Audit.Exchange, result 400
2025/08/26 18:13:30 List Audit.Exchange [2025-08-26T14:12:50.250959671Z--2025-08-26T15:12:50.249959671Z] (59m59.999s) nextPage=, returning 20, remaining 19
2025/08/26 18:13:33 List Audit.Exchange [2025-08-26T14:12:50.250959671Z--2025-08-26T15:12:50.249959671Z] (59m59.999s) nextPage=20, returning 19, remaining 0
2025/08/26 18:13:35 Subscribe Audit.Exchange, result 200
2025/08/26 18:13:35 List Audit.Exchange [2025-08-26T15:12:50.250959671Z--2025-08-26T16:12:50.249959671Z] (59m59.999s) nextPage=, returning 20, remaining 6
2025/08/26 18:14:01 received signal terminated, shutting down
2025/08/26 18:14:01 RunStop 2025-08-26T16:14:01.460620769Z
2025/08/26 18:14:01 Report subscriptions for content type Audit.AzureActiveDirectory: 12
2025/08/26 18:14:01 Report subscriptions for content type Audit.Exchange: 12
2025/08/26 18:14:01 Report requested range spans [2025-08-26T04:12:04.443770925Z--2025-08-26T16:12:04.443773725Z] for Audit.AzureActiveDirectory
2025/08/26 18:14:01 Report requested range spans [2025-08-26T04:12:50.249957722Z--2025-08-26T16:12:50.249959671Z] for Audit.Exchange
2025/08/26 18:14:01 Report max contentCreated value 2025-08-26T16:04:38.09546882Z for Audit.Exchange
2025/08/26 18:14:01 Report max contentCreated value 2025-08-26T16:10:01.651026482Z for Audit.AzureActiveDirectory
2025/08/26 18:14:01 Report served NextPageUri requested never: 1
2025/08/26 18:14:01 Report served NextPageUri requested once: 23
2025/08/26 18:14:01 Report served NextPageUri requested repeatedly: 0
2025/08/26 18:14:01 Report served and fetched never: 0
2025/08/26 18:14:01 Report served and fetched once: 733
2025/08/26 18:14:01 Report served and fetched repeatedly: 23
```

So it looks good. I see one issue that I noted in another comment.

I'd like to try a few more scenarios with empty space, but so far so good.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,227 +75,413 @@ redact:
...
+              }
+            ),
+            "work": state.work.drop(["curr_content", "next_list"]).with(
+              {
+                ?"curr_content": w.curr_content,
+                "todo_content": w.todo_content,
+                ?"next_list": [
+                  // Work around non-conformance with HTTP header MIME standards.
+                  ?list_resp.Header.?Nextpageuri[0],
+                  ?list_resp.Header.?NextPageUri[0],
```
chrisberkhout, line 196: I was thinking about this. I suspect the API uses `NextPageUri` consistently but it can be rewritten by things in between.

efd6: It's a MIME header, it should be the first format. Any Go stdlib handing of the headers will correct them. MS should not be sending the second form, but I'm sure they do.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: Maybe make `want_more` also conditional on the presence of that field?

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,227 +75,413 @@ redact:
...
+              )
+          ).as(list, (size(list) != 0) ?
+            list.map(e,
+              timestamp(e.endTime)
+            ).min().as(min,
+              list.filter(e, timestamp(e.endTime) == min)[?0]
+            ).as(oldest,
+              {
+                "curr_content": oldest,
+                "todo_content": list.filter(e, e.contentUri != oldest.contentUri),
```
chrisberkhout, line 167:
```suggestion
                "todo_content": list.filter(e, e.contentUri != oldest.value().contentUri),
```
This is a minimal change to fix double fetching the first item of every non-initial page.

From the code above, it seems that the optional `oldest` will always have a value. If the optional access is removed (`[?0]` becomes `[0]`), then the later usage of `w.curr_content` as an option would also need to change.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,227 +75,413 @@ redact:
...
-                            e.content_type == content_type
-                          )[?0].content_created_at.orValue(
-                            string(now() - duration(state.base.list_contents_start_time))
-                          )
-                        ),
-                      "next_page": "",
-                      "want_more_content": list_resp.StatusCode == 200 && (
-                        r_query.?endTime[0].optMap(t,
-                          now() - t.parse_time(time_layout.RFC3339) >= duration(state.base.batch_interval)
+                      "start": created_at + duration("1ms"),
```
chrisberkhout:
```suggestion
                      "start": created_at,
```

I just reread the [api doc](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference#list-available-content) and it says:

> The time range is inclusive with respect to startTime (startTime <= contentCreated) and exclusive with respect to endTime (contentCreated < endTime), so that non-overlapping, incrementing time intervals can used to page through available content.

Since it doesn't use `contentCreated` anymore, only `endTime` populated from the request range, the increment isn't needed. We can use the end of the last request range for the start of the new one and have no overlap.

chrisberkhout: A good thing about the end time being exclusive and not using `contentCreated` is that it doesn't matter if the times we send are truncated to a lower resolution.

chrisberkhout (changes requested):
I added two more suggestions, but then I think it's good to go.

Shall we merge and release directly? I think any remaining issues would be so minor it'd be hard to get feedback on them.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: I've removed the use of optional types here since we can prove the element always exists.

PR #14986, opened 2025-08-19
feat: add endpoint suffix override support for azure

Kavindu-Dodan:
## Proposed commit message

Introduce `endpoint_suffix` configurations for Azure integrations so that the `EndpointSuffix` of the Connection String can be overridden based on the deployment environment.

Default - `core.windows.net`
Possible overdding options : 
   - US Gov : `core.usgovcloudapi.net`
  

## Checklist

- [] I have verified that all data streams collect metrics or logs.

## Screenshots

image


zmoog (changes requested):
Looks good overall!

We need to make some changes because the "endpoint suffix" only works with the processor v2, and the only integration that allows using the processor v2 is the one named “Collect all Azure Logs (v2 preview)”.

Here's a screenshot with a few comments.

Image

Kavindu-Dodan:
@zmoog thanks for the review. I have now focused this change only for Azure V2 stream. Please have a look when you have time.

packages/azure/_dev/build/docs/adlogs.md
```diff
@@ -54,6 +54,10 @@ The name of the storage account the state/offsets will be stored and updated.
 _string_
 The storage account key, this key will be used to authorize access to data in your storage account.
 
+`endpoint_suffix` :
+_string_
+Override the default endpoint suffix used to construct the connection string. Default is set to `core.windows.net`. For example, US Government Cloud users should set this to `core.usgovcloudapi.net`.
+
```
zmoog: We should remove all these `endpoint_suffix` config docs — except the one at `packages/azure/data_stream/events`

zmoog (commented):
It's okay to add the parameter to the `azure_logs` input package because it supports the v2 processor.

zmoog (commented):
Sorry for not being more precise in the previous comments.

We can add the `endpoint_suffix` to all integrations that support the v2 processor.

For the `azure` package, it's only needed in `packages/azure/data_stream/events`. But adding it to `azure_logs` is a great idea—I hadn't thought of that initially.

packages/azure/_dev/build/docs/adlogs.md
Kavindu-Dodan: Thanks, fixed with the latest :)

Kavindu-Dodan:
@zmoog thanks for the review 🙏  

Agree on the suggestion to add this to `Custom Azure Logs` as it utilizes V2 stream.

packages/azure_logs/manifest.yml
```diff
@@ -212,6 +199,26 @@ policy_templates:
...
+          The checkpoint migration converts the checkpoint information from the v1 format to the v2 format.
+
+          Default is `false`, which means the processor will not perform the checkpoint migration.
+      - name: endpoint_suffix
+        type: text
+        default: core.windows.net
+        title: Storage account endpoint suffix
+        show_user: false
+        description: >-
+          Override the default storage account endpoint suffix.
```
zmoog:
```suggestion
          (Processor v2 only) Override the default storage account endpoint suffix.
```

packages/azure/data_stream/events/manifest.yml
```diff
@@ -155,6 +142,26 @@ streams:
...
+          The checkpoint migration converts the checkpoint information from the v1 format to the v2 format.
+
+          Default is `false`, which means the processor will not perform the checkpoint migration.
+      - name: endpoint_suffix
+        type: text
+        default: core.windows.net
+        title: Storage account endpoint suffix
+        show_user: false
+        description: >-
+          Override the default storage account endpoint suffix.
```
zmoog:
```suggestion
          (Processor v2 only) Override the default storage account endpoint suffix.
```

Kavindu-Dodan:
> Thanks! Taking a test drive on my local env.

Thank you and I have enabled auto-merge :) I will loop back if you see the need for more changes.

packages/azure/data_stream/events/agent/stream/stream.yml.hbs
```diff
@@ -19,7 +19,11 @@ storage_account: {{storage_account}}
 {{/if}}
 {{#if storage_account_key}}
 storage_account_key: {{storage_account_key}}
+{{$if endpoint_suffix}}
```
zmoog:
```suggestion
{{#if endpoint_suffix}}
```

packages/azure_logs/agent/input/input.yml.hbs
```diff
@@ -19,7 +19,11 @@ storage_account: {{storage_account}}
 {{/if}}
 {{#if storage_account_key}}
 storage_account_key: {{storage_account_key}}
+{{$if endpoint_suffix}}
```
zmoog: Same here.

```suggestion
{{#if endpoint_suffix}}
```

Kavindu-Dodan: Thanks and fixed by https://github.com/elastic/integrations/pull/14986/commits/1236019b2b2810218365ad1a22a5e22bd31c74ea 🙏

packages/azure/data_stream/events/agent/stream/stream.yml.hbs
Kavindu-Dodan: Thanks and fixed by https://github.com/elastic/integrations/pull/14986/commits/1236019b2b2810218365ad1a22a5e22bd31c74ea 🙏

packages/azure/data_stream/events/agent/stream/stream.yml.hbs
```diff
@@ -19,7 +19,11 @@ storage_account: {{storage_account}}
 {{/if}}
 {{#if storage_account_key}}
 storage_account_key: {{storage_account_key}}
+{{#if endpoint_suffix}}
+storage_account_connection_string: DefaultEndpointsProtocol=https;AccountName={{storage_account}};AccountKey={{storage_account_key}};EndpointSuffix={{endpoint_suffix}}
+{{else}}
```
muthu-mps: As the default value is set already for the endpoint_suffix config. Is it okay to remove the else condition? WDYT?
Consider making the **endpoint_suffix** `required:true` when removing the else loop.
This is applicable to the azure_logs as well.

Kavindu-Dodan: @muthu-mps good point. I added the default value to show what we use internally. So it makes sense to make it mandatory and keep the default in the UI along with simple logic internally. 

Added this change with https://github.com/elastic/integrations/pull/14986/commits/309ff31eaec23cccdf819d824b03117c529cfeae :)

muthu-mps (approved):
LGTM!

Issue #14985, opened 2025-08-19
[Beyond Insight Password Safe Managed Account]: 401 - User Not Authenticated


muralidharanvenkataraman:
### Integration Name

BeyondInsight and Password Safe [beyondinsight_password_safe]

### Dataset Name

beyondinsight_password_safe.managedaccount

### Integration Version

0.5.0

### Agent Version

9.0.4

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.0.4

### OS Version and Architecture

Cloud

### Software/API Version

0.5.0

### Error Message

User not authenticated (in http trace) and "Authentication expired or incorrect. Clearing and retrying..." in error.message

### Event Original

Not Applicable

### What did you do?

Used Beyondinsight.password_safe integration
Setup the URL, ApiKey, user, password
enabled integrations.

### What did you see?

We were receiving logs till yesterday 3:07 pm (est).  After that just authentication errors.  Do not see any Authorization parameters in any one of the trace logs.  But do see consistent SessionId in all http traces.

### What did you expect to see?

New Post Request with to reinitialize the cookie?

andrewkroh:
Fix: https://github.com/elastic/integrations/pull/14987

andrewkroh:
I'm working through cleanup of the data streams and adding more complete test scenarios. 
 
- https://github.com/elastic/integrations/pull/14999
- https://github.com/elastic/integrations/pull/15007
- https://github.com/elastic/integrations/pull/15010
- https://github.com/elastic/integrations/pull/15091
- https://github.com/elastic/integrations/pull/15103
- #15307
- #15327

TODO:

- [x] useraudit data stream
- [x] asset data stream
- [x] ingest pipeline consistency pass (format yaml, ensure best practices are followed, general consistency, maybe `subobjects: false`, disable dynamic mappings because schema is well-defined, set `ecs.version` before `terminate`, remove the `ctx.message == null` check from `terminate`) _DRAFT is ready in  ca01e51d9412d6315f4a5c2a0ff6f08799bdfb9a, waiting for other PRs to merge first._
- [x] Remove root level deploy/ directory. Each data stream has its own test setup now.
- [x] readme/doc improvements
- [x] Move enable_request_tracer to data_stream
- ~Use `duration` variable type, correct description to specify the support unit types~
- [x] Use `url_allowed_schemes` on URL variables

muralidharanvenkataraman:
@andrewkroh - Thank you :).

PR #14987, opened 2025-08-19
fix(beyondinsight_password_safe): remove cookies on auth failure

andrewkroh:
## Proposed commit message

```
fix(beyondinsight_password_safe): remove cookies on auth failure

When authentication fails with 401, remove the explicit 'cookies: null' assignment
and use an empty cursor object instead. This allows the orValue() fallback logic
to work properly by checking for key existence rather than null values.

Otherwise the cookie value would never be refreshed, leading to persistent auth
failures.

Fixes: #14985
```

## Checklist

[beyondinsight_password_safe-0.5.1.zip](https://github.com/user-attachments/files/21868782/beyondinsight_password_safe-0.5.1.zip)

## Related issues

- Fixes: #14985

### DIFF
--- a/packages/beyondinsight_password_safe/manifest.yml
+++ b/packages/beyondinsight_password_safe/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.3.0
 name: beyondinsight_password_safe
 title: BeyondInsight and Password Safe
-version: "0.5.0"
+version: "0.5.1"
 source:
   license: "Elastic-2.0"
 description: Ingest privileged access management (PAM) data from BeyondTrust's BeyondInsight PAM Reporting Platform and Password Safe, using Elastic Agent.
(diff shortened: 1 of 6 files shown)


packages/beyondinsight_password_safe/manifest.yml
```diff
@@ -69,11 +69,7 @@ policy_templates:
             required: false
             show_user: false
             description: >-
-              The request tracer logs requests and responses to the agent's local file-system for debugging configurations.
-              Enabling this request tracing compromises security and should only be used for debugging. Disabling the request
-              tracer will delete any stored traces.
-              See [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-cel.html#_resource_tracer_enable)
-              for details.
+              The request tracer logs requests and responses to the agent's local file-system for debugging configurations. Enabling this request tracing compromises security and should only be used for debugging. Disabling the request tracer will delete any stored traces. See [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-cel.html#_resource_tracer_enable) for details.
```
andrewkroh, line 72: Change was made by `elastic-package changelog add`.

andrewkroh:
This issue mentioned `beyondinsight_password_safe.managedaccount` only. I did not review the other data streams to see if they suffer the same issue. That should probably be done before merging this.

andrewkroh:
I reviewed the other data streams, and all but the `asset` had the same problem.

PR #15041, opened 2025-08-26
splunk: add search data stream

efd6:
## Proposed commit message

See title.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/splunk/_dev/deploy/docker/files/config.yml
+++ b/packages/splunk/_dev/deploy/docker/files/config.yml
@@ -535,11 +535,11 @@ rules:
             "messages": [],
             "fields": [
               {
-                "name": "@ecs.host.name",
+                "name": "'@ecs.host.name'",
                 "type": "str"
               },
               {
-                "name": "@ecs.splunk.search.result.important_user_field",
+                "name": "'@ecs.splunk.search.result.important_user_field'",
                 "type": "str"
               },
               {
@@ -584,8 +584,8 @@ rules:
             ],
             "results": [
               {
-                "@ecs.host.name": "ip-186-86-82-48",
-                "@ecs.splunk.search.result.important_user_field": "test",
+                "'@ecs.host.name'": "ip-186-86-82-48",
+                "'@ecs.splunk.search.result.important_user_field'": "test",
                 "_bkt": "main~0~266C23A8-75B5-4A00-8732-4A6065281018",
                 "_cd": "0:10",
                 "_indextime": "1740148016",
@@ -629,11 +629,11 @@ rules:
             "messages": [],
             "fields": [
               {
-                "name": "@ecs.host.name",
+                "name": "'@ecs.host.name'",
                 "type": "str"
               },
               {
-                "name": "@ecs.splunk.search.result.important_user_field",
+                "name": "'@ecs.splunk.search.result.important_user_field'",
                 "type": "str"
               },
               {
@@ -678,8 +678,8 @@ rules:
             ],
             "results": [
               {
-                "@ecs.host.name": "ip-186-86-82-48",
-                "@ecs.splunk.search.result.important_user_field": "test",
+                "'@ecs.host.name'": "ip-186-86-82-48",
+                "'@ecs.splunk.search.result.important_user_field'": "test",
                 "_bkt": "main~0~266C23A8-75B5-4A00-8732-4A6065281018",
                 "_cd": "0:10",
                 "_indextime": "1740148017",
@@ -698,8 +698,8 @@ rules:
                 "splunk_server": "ip-186-86-82-49"
               },
               {
-                "@ecs.host.name": "ip-186-86-82-48",
-                "@ecs.splunk.search.result.important_user_field": "test",
+                "'@ecs.host.name'": "ip-186-86-82-48",
+                "'@ecs.splunk.search.result.important_user_field'": "test",
                 "_bkt": "main~0~266C23A8-75B5-4A00-8732-4A6065281018",
                 "_cd": "0:10",
                 "_indextime": "1740148018",


packages/splunk/data_stream/search/manifest.yml
```diff
@@ -0,0 +1,169 @@
...
+        type: yaml
+        title: Field mappings
+        multi: false
+        required: false
+        show_user: true
+        description: >-
+          Field mappings is a YAML map for CIM field names to ECS field names for the query.
+          Splunk field names may appear only once and ECS fields should not appear more than
+          once. Field names may not include "=" or ",". The Splunk `_time` field is always
+          used to set `@timestamp` and remapping this field is ignored by the integration.
```
kcreddy: It might be helpful for users to add link to ECS fields.

packages/splunk/data_stream/search/manifest.yml
```diff
@@ -0,0 +1,169 @@
...
+        type: yaml
+        title: Default field values
+        multi: false
+        required: false
+        show_user: true
+        description: >-
+          Default values for ECS fields. Field values will be set if not otherwise found in
+          the search results after applying field mappings.
+        default: |
+          #"event.module": splunk
```
kcreddy, line 60: Same here. A link to ECS fields might help.

packages/splunk/_dev/build/docs/README.md
```diff
@@ -56,3 +56,13 @@ This is the `alert` dataset.
 {{event "alert"}}
 
 {{fields "alert"}}
+
+### Search
+
+This is the `search` dataset.
```
kcreddy, line 62: Could also add it after https://github.com/elastic/integrations/blob/main/packages/splunk/_dev/build/docs/README.md?plain=1#L13

packages/splunk/_dev/build/docs/README.md
```diff
@@ -56,3 +56,13 @@ This is the `alert` dataset.
 {{event "alert"}}
 
 {{fields "alert"}}
+
+### Search
```
kcreddy, line 60: Can we add link to the API documentation somewhere in README?

packages/splunk/data_stream/search/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,182 @@
...
+        - splunk.alert
+      tag: remove_custom_duplicate_fields
+      ignore_missing: true
+      if: ctx.tags == null || !(ctx.tags.contains('preserve_duplicate_custom_fields'))
+  - remove:
+      field: tmp
+      tag: remove_temp
+      ignore_missing: true
+  - script:
+      tag: script_to_drop_null_values
```
kcreddy: Should we go with updated script processor instead?

https://github.com/elastic/integrations/blob/main/packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml#L3491-L3516

packages/splunk/data_stream/search/manifest.yml
efd6: Which part? The use of ECS here is suggestive only, what I really mean is fields in the resulting document.

packages/splunk/data_stream/search/manifest.yml
efd6: As above.

packages/splunk/_dev/build/docs/README.md
efd6: This is already in the [README](https://github.com/elastic/integrations/blob/main/packages/splunk/_dev/build/docs/README.md?plain=1#L7) since this uses the same API as the existing data stream, just differently.

packages/splunk/data_stream/search/elasticsearch/ingest_pipeline/default.yml
efd6: Is there a reason to choose that over what is here?

kcreddy: I took it from https://github.com/elastic/integrations/pull/13325 where we saw marginal gains changing the null cleaning script.

packages/splunk/data_stream/search/manifest.yml
kcreddy: For users new to Elastic (who may not know what ECS is), it might benefit to add a link to our [ECS fields](https://www.elastic.co/docs/reference/ecs/ecs-field-reference) that can be used in this setting.

(Suggestion only, please ignore if you believe its not the right place).

efd6: I'll add it.

kcreddy (approved):

efd6:
I'll wait for @andrewkroh, particularly for review of [this](https://github.com/elastic/integrations/pull/15041/files#diff-3c71f1864048145c6cdd060dd6337234abb978b7d859eee25d6700a1a2afb72bR61-R75) and the subsequent logic that depends on that (query construction based on user input), and for @jamesspi for general comments on the user-facing aspects.

packages/splunk/data_stream/search/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,21 @@
...
+    url: http://{{Hostname}}:{{Port}}
+    username: search_user
+    password: password
+    query: "search *"
+    interval: 5m
+    initial_interval: 24h
+    batch_size: 2
+    preserve_original_event: true
+    preserve_duplicate_custom_fields: true
+    enable_request_tracing: true
```
andrewkroh:
```suggestion
    enable_request_tracer: true
```

packages/splunk/data_stream/search/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,315 @@
...
+                        // Copy fields to new names and delete originals.
+                        // Dynamic query construction issues obviously exist;
+                        // fields cannot contain "=" or ",". There may be
+                        // other constraints.
+                        // We do not include "-" which is invalid or "_time"
+                        // which we need.
+                        sprintf(
+                          "|eval %s|fields - %s",
+                          [
+                            state.field_mapping.map(f, f != "_time" && f != "-", "@ecs." + state.field_mapping[f] + "=" + f).join(","),
```
andrewkroh: I think we need to apply single-quoting around the field names. Here's what I was tinkering with:

```js
{
	"host": "host.name",
	"field1": "splunk.search.result.important_user_field",
	"user,type": "user.type",
	"status=active": "event.outcome",
	"response|time": "event.duration",
	"client(app)": "user_agent.name",
	"user name": "user.name",   
    "drive-C:\\": "file.device",
    "field's name": "labels.field_name",

}.transformMapEntry(k, v,
	{
		"'" + k.replace_all("\\", "\\\\").replace_all("'", "\\'") + "'": "'@ecs." + v.replace_all("\\", "\\\\").replace_all("'", "\\'") + "'",
	}
)
```

Sources:

https://help.splunk.com/en/splunk-cloud-platform/search/spl2-search-manual/wildcards-quotes-and-escape-characters/quotation-marks

- Field names that begin with anything other than a-z, A-Z, or the underscore ( _ ) character must be enclosed in single quotation marks ( ' ).
- Field names that contain anything other than a-z, A-Z, 0-9, or the underscore ( _ ) character must be enclosed in single quotation marks ( ' ). This includes the wildcard ( * ) character, the dash ( - ), and the space character.

https://help.splunk.com/en/splunk-cloud-platform/search/spl2-search-manual/wildcards-quotes-and-escape-characters/when-to-escape-characters

- If a field name contains a single quotation mark ( ' ), that single quotation mark must be escaped. Otherwise, the search will misinterpret where the field name ends.

efd6: Given where we are, that seems reasonable.

efd6: New code tested against a real tenant. It returns the fields quoted, so the second part is to reverse these changes when collating into the `ecs_result` field.

andrewkroh: > It returns the fields quoted

That's an interesting finding.

> New code tested against a real tenant.

Was the old code tested? I'm just curious to know if the unquoted `@ecs.` fields actually caused problems.

efd6: > Was the old code tested?

Yes.

> I'm just curious to know if the unquoted `@ecs.` fields actually caused problems.

No problems with that at all; everything just worked. Having the quotes though now means that we can have cool things like `=`-containing fields.

Issue #14239, opened 2025-06-17
[filestream]: Add support for ingesting GZIP logs


AndersonQ:
### Integration Name

Custom Filestream Logs [filestream]

### Dataset Name

filestream.gzip

### Integration Version

1.1.5

### Agent Version

9.1.0

### OS Version and Architecture

all

### User Goal

With the addition of GZIP support to the `filestream` input (https://github.com/elastic/beats/issues/44866),, the Custom Logs (filestream) integration can be updated to expose and leverage this new capability. This will ensure users can easily enable GZIP log collection.
The GZIP support on filestream is on tech preview and is an opt-in feature.

### Existing Features

Currently it isn't possible to ingest GZIP files.

### What did you see?

Currently it isn't possible to ingest GZIP files.

PR #15047, opened 2025-08-26
WIP: 14239 filestream gzip support

AndersonQ:
## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/integrations/issues/14239


packages/filestream/_dev/deploy/docker/sample_logs/test-filestream-fingerprint.log
```diff
@@ -1,2 +0,0 @@
-28.20.184.203 - toy3210 [12/Jun/2025:11:28:10 +0000] "POST /leverage/applications HTTP/1.0" 406 22116
-145.92.68.176 - tromp7278 [12/Jun/2025:11:28:10 +0000] "DELETE /scale HTTP/1.1" 502 13209
```
belimawr, line 2: Why was this file removed?

packages/filestream/data_stream/generic/manifest.yml
```diff
@@ -14,6 +14,15 @@ streams:
         show_user: true
         default:
           - /var/log/*.log
+      - name: gzip_experimental
```
belimawr: [Question]
Why the `_experimental` suffix? It's not shown to any user and at some point it will be GA....

AndersonQ: The idea is to remove the flag completely and have GZIP decompressed as a default behaviour. It's defined on the RFC

belimawr: It's not a blocker, but I'd still name it only `gzip` as we've done with all other beta/experimental features.

AndersonQ: It's to match the config property, that's why I used `gzip_experimental`.
I thought you're talking about the config on filebeat

PR #15035, opened 2025-08-25
[postgresql] OTel Integration

devamanv:
## Proposed commit message
The PR  adds PostgreSQL OpenTelemetry assets. 
For more details, refer the related issue down below. 

## Author's Checklist
- [x] Add a README.md explaining the purpose of the package and clear instructions to setup and sample configuration
- [x] Add a screenshot of the dashboard

## Related issues

-  Closes https://github.com/elastic/obs-integration-team/issues/466

## Screenshots
image

PR #15082, opened 2025-08-28
[mysql_otel] [docs] remove "content pack" and "EDOT collector" mentions

mykola-elastic:
The EDOT OTel collector is not yet supported, the "content pack" is not an officially defined thing.

Context: comments from PostgreSQL OTel assets PR review:
- https://github.com/elastic/integrations/pull/15035/files#r2300449003
- https://github.com/elastic/integrations/pull/15035/files#r2300455794

### DIFF
--- a/packages/mysql_otel/docs/README.md
+++ b/packages/mysql_otel/docs/README.md
@@ -6,7 +6,7 @@ The MySQL OpenTelemetry assets provide a visual representation of MySQL metrics 
 
 ## Compatibility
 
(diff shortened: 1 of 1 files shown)


ishleenk17:
What is the Integration version that will be seen on 9.1 if we don't have that in our manifest ?
Does it pick the latest one by default ?

mykola-elastic:
@ishleenk17 updated the changelog/manifest so that it is visible now (as version 0.1.1)

PR #15093, opened 2025-08-29
rapid7_insightvm: use terminate function to stop pagination in HTTPJSON templates

navnit-elastic:
## Proposed commit message

```
rapid7_insightvm: use terminate function to stop pagination in HTTPJSON templates

Previously, it relied on a template error to stop pagination, which resulted in
the fleet's health status changing to degraded.

The "terminate" processor is available starting from stack version 8.19.3 and 9.1.3.
Reference: https://github.com/elastic/beats/pull/45810.
```
## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Relates #14330
- Relates #14301


efd6 (commented):
I think this needs `assert.hit_count` in the system tests (looking at this, I'm not sure that the tests would actually pass in the old state; AFAICS there should be 3 events in the assets test and 7 for the vulnerability test; but they give only 2 each).

navnit-elastic:
Hi @efd6, by looking at the agent logs, It appears that the template execution is not getting terminated when it should be. The message `template execution terminated` appears multiple time in logs. Did I make the correct use of the `terminate` function?

efd6:
@navnit-elastic The `terminate` helper prevents any error from being emitted, so the pagination will not stop. It terminates the evaluation of the template, nothing else. The option that is needed here is `ignore_empty_value`, but this is only on the cursor assignment templates. I do not think we have a way of suppressing the same thing when the destination is not the cursor.

navnit-elastic:
Thanks @efd6 for the clarification, The confusion is arises from the template error and fleet health status being degraded (which is the root cause of flaky tests failures) representing different aspects.
We need a way to stop pagination without updating the fleet health status to degraded, which I believe be addressed after PR https://github.com/elastic/beats/pull/46332 is merged. This PR should be closed.

efd6:
Yeah. When the fix in beats is merged, the fix here becomes a matter of just bumping the `kibana.version`.

kcreddy:
@navnit-elastic, the backport PRs are also merged.
You can reopen this PR to use `do_not_log_failure` option as per the PR.

navnit-elastic:
@kcreddy, the changes will be available starting from the next release. I'll raise a separate PR to bump the Kibana version after 8.19.4 and 9.1.4 are released.

PR #15101, opened 2025-08-29
Add warning note in RabbitMQ package readme about possible discrepancy after RabbitMQ upgrade

giorgi-imerlishvili-elastic:
## Proposed commit message

Add warning message in RabbitMQ package readme which warns customers that upgrading RabbitMQ from tested version 3.7.4 to recent version 4.1.3 will cause minor discrepancy.
In recent version of RabbitMQ 4.1.3 api endpoint /api/nodes no more returns data `rabbitmq.node.queue.index.journal_write` which was used in **Queue Index Operations [Metrics RabbitMQ]** dashboard so after upgrading to 4.1.3 or later, this metric and related visualizations in the dashboard will be missing or incomplete.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/rabbitmq/docs/README.md
```diff
@@ -14,6 +14,12 @@ the management plugin (which needs to be installed and enabled). Exchange datase
 
 The application logs dataset parses single file format introduced in 3.7.0.
 
+### Note on RabbitMQ 4.1.3 and above
```
ishleenk17: You need to update the _README under _dev. 
This one is autogenerated.

packages/rabbitmq/docs/README.md
```diff
@@ -14,6 +14,12 @@ the management plugin (which needs to be installed and enabled). Exchange datase
 
 The application logs dataset parses single file format introduced in 3.7.0.
```
ishleenk17, line 14: The version of RabbitMQ needs to be updated here as well

packages/rabbitmq/docs/README.md
```diff
@@ -14,6 +14,12 @@ the management plugin (which needs to be installed and enabled). Exchange datase
...
 
+### Note on RabbitMQ 4.1.3 and above
+
+When upgrading RabbitMQ from version 3.7.4 to 4.1.3, the `/api/nodes` endpoint no longer returns the `rabbitmq.node.queue.index.journal_write` metric in its JSON response. This metric is used in the Kibana dashboard "Queue Index Operations [Metrics RabbitMQ]". As a result, after upgrading to 4.1.3 or later, this metric and related visualizations in the dashboard will be missing or incomplete.
```
ishleenk17:
```suggestion
When upgrading RabbitMQ to version 4.1.3, the `/api/nodes` endpoint no longer returns the `rabbitmq.node.queue.index.journal_write` metric in its response. This metric is used in the Kibana dashboard "Queue Index Operations [Metrics RabbitMQ]". As a result, after upgrading to 4.1.3 or later, this metric and related visualizations in the dashboard will be missing or incomplete.
```

giorgi-imerlishvili-elastic:
@ishleenk17 updated readme under _dev folder, also made suggested changes.
Merged warning note to Compatibility section and updated versions there as well.

packages/rabbitmq/_dev/build/docs/README.md
```diff
@@ -9,11 +9,12 @@ in this integration configuration.
 
 ## Compatibility
 
-The RabbitMQ integration is fully tested with RabbitMQ 3.7.4 and it should be compatible with any version supporting
-the management plugin (which needs to be installed and enabled). Exchange dataset is also tested with 3.6.0, 3.6.5 and 3.7.14.
+The RabbitMQ integration is fully tested with RabbitMQ 3.7.4 and 4.1.3, and it should be compatible with any version supporting the management plugin (which needs to be installed and enabled). It has also been tested with 3.6.0, 3.6.5, and 3.7.14.
```
ishleenk17: It has also been tested with 3.6.0, 3.6.5, and 3.7.14.

Why do we need this separately ?

ishleenk17: @giorgi-imerlishvili-elastic You tried it with 3.7.14 and 4.1.3.
Is that right ?

giorgi-imerlishvili-elastic: > @giorgi-imerlishvili-elastic You tried it with 3.7.14 and 4.1.3. Is that right ?

yes right.



> It has also been tested with 3.6.0, 3.6.5, and 3.7.14.
> 
> Why do we need this separately ?

Previously it was written like this: "Exchange dataset is also been tested with 3.6.0, 3.6.5, and 3.7.14."
 I removed this part "Exchange dataset is" that's why I think now it looks little bit confusing

ishleenk17: We can mention just 3.7.14 and 4.1.3
Having so many 3.x versions doesn't make sense to me.

@daniela-elastic @alaudazzi  : Do you have any issue if we keep just these 2 ?

daniela-elastic: I think users will like to know if the integration will work with 3.7.x and 4.1.x. If you have tested it only with 3.6.0, 3.6.5, etc, does this answer the user's question? If so, we can mention what we support. If not or we don't know then we can mention what we've tested it with (the long list) and also add that we are reasonably sure that it should work for all patch versions of 3.6, 3.7 and 4.1

packages/rabbitmq/_dev/build/docs/README.md
```diff
@@ -9,11 +9,12 @@ in this integration configuration.
...
 
 The application logs dataset parses single file format introduced in 3.7.0.
 
+When upgrading RabbitMQ to version 4.1.3 or above, the `/api/nodes` endpoint no longer returns the `rabbitmq.node.queue.index.journal_write` metric in its response. This metric is used in the Kibana dashboard "Queue Index Operations [Metrics RabbitMQ]". As a result, after upgrading to 4.1.3 or later, this metric and related visualizations in the dashboard will be missing or incomplete.
```
daniela-elastic: this endpoint is the RabbitMQ endpoint, right? Maybe specify that so there's no confusion that it's an Elastic endpoint's fault

packages/rabbitmq/_dev/build/docs/README.md
giorgi-imerlishvili-elastic:
```suggestion
The RabbitMQ integration is fully tested with RabbitMQ 3.7.4 and 4.1.3, and it should be compatible with any version supporting the management plugin (which needs to be installed and enabled). Exchange dataset is also tested with 3.6.0, 3.6.5 and 3.7.14.
```
I think will be better if i bring back how it was written before, I don't know why `Exchange dataset` were mentioned separately that's why I removed it but looks like removing it causes more confusion.

alaudazzi: @ishleenk17 we need to mention the relevant versions only.

alaudazzi:
```suggestion
The RabbitMQ integration is fully tested with RabbitMQ versions 3.7.4 and 4.1.3, and it should be compatible with any version supporting the management plugin, which must be installed and enabled. It has also been tested with 3.6.0, 3.6.5, and 3.7.14.
```
Sounds like this info is relevant, might not deserve brackets...

packages/rabbitmq/_dev/build/docs/README.md
```diff
@@ -9,11 +9,12 @@ in this integration configuration.
 
 ## Compatibility
 
-The RabbitMQ integration is fully tested with RabbitMQ 3.7.4 and it should be compatible with any version supporting
-the management plugin (which needs to be installed and enabled). Exchange dataset is also tested with 3.6.0, 3.6.5 and 3.7.14.
+The RabbitMQ integration is fully tested with RabbitMQ 3.7.4 and 4.1.3, and it should be compatible with any version supporting the management plugin (which needs to be installed and enabled). It has also been tested with 3.6.0, 3.6.5, and 3.7.14.
 
 The application logs dataset parses single file format introduced in 3.7.0.
 
+When upgrading RabbitMQ to version 4.1.3 or above, the `/api/nodes` endpoint (from RabbitMQ's management API) no longer returns the `rabbitmq.node.queue.index.journal_write` metric in its response. This metric is used in the Kibana dashboard "Queue Index Operations [Metrics RabbitMQ]". As a result, after upgrading to 4.1.3 or later, this metric and related visualizations in the dashboard will be missing or incomplete.
```
alaudazzi:
```suggestion
When upgrading RabbitMQ to version 4.1.3 or higher, the `/api/nodes` endpoint (from RabbitMQ's management API) no longer returns the `rabbitmq.node.queue.index.journal_write` metric in its response. This metric is used in the Kibana dashboard "Queue Index Operations [Metrics RabbitMQ]". As a result, after upgrading to 4.1.3 or higher, this metric and related visualizations in the dashboard will be missing or incomplete.
```

alaudazzi (approved):
Left a few minor suggestions, otherwise LGTM.

PR #15114, opened 2025-09-02
Add json log formatter support for RabbitMQ

giorgi-imerlishvili-elastic:
## Proposed commit message

Add json log formatter support for RabbitMQ

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


muthu-mps:
@giorgi-imerlishvili-elastic - Can we add the pipeline tests for the new log format?

agithomas:
@giorgi-imerlishvili-elastic , we have Amazon MQ integration for RabbitMQ. Could you verify if the mentioned log format is supported for Amazon MQ managed RabbitMQ as well?

giorgi-imerlishvili-elastic:
> @giorgi-imerlishvili-elastic , we have Amazon MQ integration for RabbitMQ. Could you verify if the mentioned log format is supported for Amazon MQ managed RabbitMQ as well?

Checked and currently it's not supported, in order to make it support it will require similar changes there as well

agithomas:
> Checked and currently it's not supported, in order to make it support it will require similar changes there as well

Thanks.Kindly create a issue for this against Amazon MQ managed RabbitMQ

muthu-mps:
@alaudazzi - Can you help review the REAME changes?

packages/rabbitmq/data_stream/log/_dev/test/pipeline/test-rabbitmq.log-expected.json
```diff
@@ -4081,6 +4081,29 @@
...
+          "type": [
+            "info"
+          ]
+        },
+        "log": {
+          "level": "warning"
+        },
+        "message": "HTTP access denied: user 'sdfds' - invalid credentials",
+        "rabbitmq": {
+          "log": {}
```
muthu-mps: The log field is empty. `pid` and `domain` values are missing in the extracted document.

giorgi-imerlishvili-elastic:
> > Checked and currently it's not supported, in order to make it support it will require similar changes there as well
> 
> Thanks.Kindly create a issue for this against Amazon MQ managed RabbitMQ

Created: https://github.com/elastic/integrations/issues/15300

packages/rabbitmq/data_stream/log/_dev/test/pipeline/test-rabbitmq.log
```diff
@@ -166,4 +166,5 @@ database dir   : /var/lib/rabbitmq/mnesia/rabbit@af6809c8510d
 * rabbitmq_prometheus
 * rabbitmq_management
 * rabbitmq_web_dispatch
-* rabbitmq_management_agent
\ No newline at end of file
+* rabbitmq_management_agent
+{"time":"2025-09-10 14:52:44.657199+00:00","level":"warning","msg":"HTTP access denied: user 'sdfds' - invalid credentials","pid":"<0.997.0>","domain":"rabbitmq"}
```
muthu-mps: - Is this the only format RabbitMQ logs emit? The official document has more examples [here](https://www.rabbitmq.com/docs/logging#json:~:text=Before%20demonstrating%20an%20example%2C%20here%20is%20a%20message%20with%20the%20default%20mapping%3A). 
- The example format added in the [issue](https://github.com/elastic/obs-integration-team/issues/492#:~:text=This%20is%20example%20what%20json%20log%20message%20looks%20like%3A) is different from what I see in test logs.
- There is a way to [customize the logging](https://www.rabbitmq.com/docs/logging#json:~:text=Now%2C%20an%20example%20that%20uses%20JSON%20logging%20with%20a%20custom%20field%20mapping%3A) field names. This should be called out explicitly in the README that the JSON logging field name should be of default format.
- Are you able to test the JSON format by running RabbitMQ?

mykola-elastic: Would it be better if we make a separate test case for json logs?

in two separate files, for example, `test-rabbitmq-json.log` and `test-rabbitmq-json.log-expeceted.json`?

Mixing JSON and "normal" log in one file looks slightly uncomfortable to me

giorgi-imerlishvili-elastic: * No, RabbitMQ supports multiple log formats. The default JSON log format uses a predefined set of field names, but RabbitMQ allows customization of these field names and the addition of extra fields.
* The example in the issue demonstrates the default fields, sometimes with additional custom fields. Our integration currently focuses on extracting information from the default set of fields, similar to how we handle the TEXT format.
* Yes, we are aware of this. The README has been updated to explicitly state that the integration supports parsing only the default field names as output by RabbitMQ. If the field names are customized, parsing may not work as expected.
* Yes, I have tested the integration locally with RabbitMQ emitting JSON logs. All fields currently defined in the integration are correctly extracted from the default JSON format.

packages/rabbitmq/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -13,28 +13,62 @@ processors:
   - set:
       field: event.type
       value: ["info"]
+  # Try to parse as JSON log first
+  - json:
+      field: message
+      target_field: rabbitmq.log.json
+      add_to_root: false
```
muthu-mps: This is set to [false by default](https://www.elastic.co/docs/reference/enrich-processor/json-processor#:~:text=will%20be%20overwritten.-,add_to_root,-no). Can we remove this flag?

packages/rabbitmq/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -13,28 +13,62 @@ processors:
   - set:
       field: event.type
       value: ["info"]
+  # Try to parse as JSON log first
```
muthu-mps, line 16: When applying the processor without `IF` condition this will hit every time for the older version non JSON logs. Is there a possibility to add instanceOF operator?

giorgi-imerlishvili-elastic: @muthu-mps you mean checking if string is json before trying to parse it?

muthu-mps: yeah, If condition fails then skips parsing. The current way has to parse each non Json input. 
@ishleenk17 - Do you have any comments?

giorgi-imerlishvili-elastic: Not sure but following condition might work:
```suggestion
  # Try to parse as JSON log first
  - json:
      field: message
      target_field: rabbitmq.log.json
      ignore_failure: true
      if: "ctx.message != null && ctx.message.startsWith('{')"
```

PR #15128, opened 2025-09-02
Add GH workflow for making docs edit easier

shmsr:
## Proposed commit message

See https://github.com/elastic/obs-integration-team/issues/431 for more details. This PR adds a workflow to make edits to the docs easier than ever.

### DIFF
--- a/.github/workflows/docs-edit-automation.yml
+++ b/.github/workflows/docs-edit-automation.yml
@@ -137,7 +137,7 @@ git commit -m \"docs: update changelogs and build documentation\"
 git push
 
 
-Prerequisite: \`go install github.com/elastic/elastic-package@v0.114.0\`"
+Prerequisite: \`go install github.com/elastic/elastic-package\`"
 
           # Add info about packages already updated
           if [ -n "${{ steps.analyze.outputs.packages_already_done }}" ]; then


shmsr:
Hey @elastic/ecosystem! I am adding a workflow that runs only on specific branch patterns. This workflow makes edits to the docs easier.

Can you please review and approve? I want to test this out fast.

mykola-elastic:
Should this be documented somewhere as well? Will there be a PR on adding some description of this? The folks should somehow find out that they need to name branches like `docs-enhancement/**`, `docs-bugfix/**` and the `changelog: ` thing

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,158 @@
...
+          
+          echo "change_type=$CHANGE_TYPE" >> $GITHUB_OUTPUT
+          echo "label=$LABEL" >> $GITHUB_OUTPUT
+          
+          # Get PR body for changelog description
+          PR_BODY=$(gh pr view $PR_NUMBER --json body -q '.body // ""')
+          
+          CHANGELOG_DESC=""
+          if echo "$PR_BODY" | grep -qi "^changelog:"; then
+            CHANGELOG_DESC=$(echo "$PR_BODY" | grep -i "^changelog:" | sed 's/^changelog://i' | xargs)
```
mykola-elastic: so, the users need to put a line beginning with `changelog: ` to edit the description but this is optional, right?

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,158 @@
...
+            
+            PROCESSED="$PROCESSED $PACKAGE"
+            cd ../..
+            echo "::endgroup::"
+          done <<< "$CHANGED_DOCS"
+          
+          echo "processed=$PROCESSED" >> $GITHUB_OUTPUT
+          
+          # Add PR label
+          gh pr edit $PR_NUMBER --add-label "$LABEL" || true
```
mykola-elastic: should we add `documentation` label too?

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,158 @@
...
+            # Update changelog if tool exists
+            if command -v elastic-package-changelog &> /dev/null; then
+              elastic-package-changelog add-next \
+                --type "$CHANGE_TYPE" \
+                --description "$CHANGELOG_DESC" \
+                --pr "$PR_NUMBER" || echo "::warning::Failed to update changelog for $PACKAGE"
+            fi
+            
+            # Build documentation
+            elastic-package build || echo "::warning::Failed to build $PACKAGE"
```
mykola-elastic: Why is warning used instead of just failing?

shmsr:
> Should this be documented somewhere as well? Will there be a PR on adding some description of this? The folks should somehow find out that they need to name branches like `docs-enhancement/**`, `docs-bugfix/**`

Yes, I will document this one. First, I'd like to test it. Then will publish the docs for the final set of changes.

.github/workflows/docs-edit-automation.yml
shmsr: Exactly; to override the default text.

.github/workflows/docs-edit-automation.yml
shmsr: Ah, right. Will add.

.github/workflows/docs-edit-automation.yml
shmsr: So as things handle one ore more packages, to continue the loop, it was like that. However, improved it further to show error labels instead of warnings and clearly display the failed packages. Nevertheless, we are not failing the script if the build or changelog addition fails as want it to go through the full set of packages.

shmsr:
> Hey @elastic/ecosystem! I am adding a workflow that runs only on specific branch patterns. This workflow makes edits to the docs easier.
> 
> Can you please review and approve? I want to test this out fast.

Friendly ping again. I'd like to test this; might require more changes after I am done with testing. Could you please approve for now?

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,180 @@
...
+              if ! elastic-package-changelog add-next \
+                --type "$CHANGE_TYPE" \
+                --description "$CHANGELOG_DESC" \
+                --pr "$PR_NUMBER"; then
+                echo "::error::Failed to update changelog for $PACKAGE"
+                PACKAGE_FAILED=true
+              fi
+            else
+              echo "::warning::elastic-package-changelog tool not found"
+            fi
```
mrodm: `elastic-package` offers also the command `elastic-package changelog` command to help adding new changelog entries.
Could it be used that ?

```shell
 $ elastic-package changelog add -h
Use this command to add an entry to the changelog file.

The entry added will include the given description, type and link. It is added on top of the
last entry in the current version

Alternatively, you can start a new version indicating the specific version, or if it should
be the next major, minor or patch version.

Usage:
  elastic-package changelog add [flags]

Flags:
      --description string   description for the changelog entry
  -h, --help                 help for add
      --link string          link to the pull request or issue with more information about the changelog entry
      --next major           changelog entry is added in the next major, `minor` or `patch` version
      --type string          type of change (bugfix, enhancement or breaking-change) for the changelog entry
      --version string       changelog entry is added in the given version

Global Flags:
  -C, --change-directory string   change to the specified directory before running the command
  -v, --verbose count             verbose mode
```

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,180 @@
+name: Documentation edit automation
+
+on:
+  push:
+    branches:
+      - 'docs-enhancement/**'
+      - 'docs-bugfix/**'
```
mrodm: IIRC this implies that branches should be created in the upstream. Usually, Pull Requests are created using forks in this repository.

Could it be used the event `pull_request` instead of `push` ? To just run these actions once the pull requests are opened ? In that way, I think they will be executed for PRs created from forks.

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-the-head-or-base-branch-of-a-pull-request

Another option is to add a new step in the buildkite pipeline to add this commits.
I think Kibana uses Buildkite to push new commits to the pull requests branches using this function:
https://github.com/elastic/kibana/blob/ca1df75db329e64315a60c970ad581feb71e4ee8/.buildkite/scripts/common/util.sh#L29C1-L29C24

WDYT @elastic/ecosystem ? Any preference to use GH Actions or Buildkite for this?

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,180 @@
...
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.24'
+
+      - name: Install tools
+        run: |
+          go install github.com/elastic/elastic-package@latest
+          go install github.com/andrewkroh/go-examples/elastic-package-changelog@main
+          echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
```
mrodm: Instead of latest, it would be better to use the same version defined in "go.mod" as buildkite builds do. This will ensure the same output or results.

For the other package, if it cannot be used `elastic-package changelog add` command, it looks like it would be better to pin the version too in [`tools.go` file](https://github.com/elastic/integrations/blob/d7c2620a3b0e00b8634f49ca6817c16b02a3f425/tools.go) if possible.

.github/workflows/docs-edit-automation.yml
mrodm: @v1v do you think there is any inconvenience to add a GitHub action that creates new commits in branches ?

v1v: > IIRC this implies that branches should be created in the upstream. Usually, Pull Requests are created using forks in this repository.

That's the expected behaviour for upstream branches only.

> Could it be used the event pull_request instead of push ? To just run these actions once the pull requests are opened ? In that way, I think they will be executed for PRs created from forks.

That won't work as a `pull-request` event does not allow `write` access, while the alternative is `pull-request-target`, but that's a no-go.

> Another option is to add a new step in the buildkite pipeline to add this commits.
I think Kibana uses Buildkite to push new commits to the pull requests branches using this function:
[elastic/kibana@ca1df75/.buildkite/scripts/common/util.sh#L29C1-L29C24](https://github.com/elastic/kibana/blob/ca1df75db329e64315a60c970ad581feb71e4ee8/.buildkite/scripts/common/util.sh#L29C1-L29C24)

I think, the same should apply as the `pull-request-target` for GH actions.

----------

What's the reason this is not a post-merge PR creation linked to the original PR in the PR description? If `main` can only by updated through PRs, then a `on: push: main` should work with `write` permissions and then gather the PR that created the commit in main.

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,180 @@
...
+  auto-update-docs:
+    # Skip if pushed by bot to avoid loops
+    if: github.actor != 'github-actions[bot]'
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
v1v: `v5` has been released recently, token is not needed when usin gthe default GH token.

```suggestion
        uses: actions/checkout@v5
```

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,180 @@
...
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.24'
```
v1v: See https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file


```suggestion
          go-version-file: 'go.mod'
```

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,180 @@
...
+          echo "::notice::Using changelog description: $CHANGELOG_DESC"
+        
+          
+          # Find changed documentation files
+          CHANGED_DOCS=$(git diff --name-only origin/main...HEAD | grep '^packages/.*/_dev/build/docs/README\.md$' || true)
+          
+          if [ -z "$CHANGED_DOCS" ]; then
+            echo "::notice::No documentation changes found"
+            exit 0
+          fi
```
v1v: IIUC, this means that if the PR does not contain any docs changes, if so, I think you can filter with the path in the `on` event:

- https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-excluding-paths

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,180 @@
...
+            exit 1
+          fi
+          
+          echo "processed=$PROCESSED" >> $GITHUB_OUTPUT
+          
+          # Add PR labels
+          gh pr edit $PR_NUMBER --add-label "$LABEL,documentation"
+
+      - name: Auto-commit changes
+        uses: stefanzweifel/git-auto-commit-action@v5
```
v1v: If you use third-party github actions, please follow the guidelines below:

- https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

.github/workflows/docs-edit-automation.yml
mrodm: > What's the reason this is not a post-merge PR creation linked to the original PR in the PR description? If `main` can only by updated through PRs, then a `on: push: main` should work with `write` permissions and then gather the PR that created the commit in main.

IIUC the aim of these branches is to update the documentation of the packages. As the documentation is updated in the branch, this requires to publish a new version of the required packages (adding also new changelog entries). And all those changes are required to be in the same Pull Request so new versions of the packages can be released when the PR is merged.

I think it shouldn't be updated the documentation (in main or backport branches) without updating the versioning.

shmsr: While GitHub provides the [feature](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) (which is enabled by default for PRs to elastic/integrations, elastic/beats, etc.), GitHub Actions still cannot push commits back to fork branches due to security restrictions.

This PR addresses a real pain point for documentation contributors. Currently, the process for getting documentation changes merged involves https://github.com/elastic/obs-integration-team/issues/431#issue-3259529491, which creates friction for doc writers who should be focusing on content rather than tooling.

Given the technical constraints, I am thinking to do this:

- Automated Analysis: The workflow detects packages needing changelog/manifest updates
- Generate Commands: Create a ready-to-run command that handles all necessary updates
- PR Comment (or logs): Post a comment with clear instructions that the author of the PR can run locally and push the changes to origin

So it is still the same workflow but one extra manual step at the end. Do you think there's a better way?

.github/workflows/docs-edit-automation.yml
shmsr: Yes, sure. I will use this one. Thanks!

.github/workflows/docs-edit-automation.yml
shmsr: Thanks!

.github/workflows/docs-edit-automation.yml
shmsr: Ok.

.github/workflows/docs-edit-automation.yml
shmsr: Ok, I will try this one. Thanks!

shmsr: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-pull_request_target-workflow-based-on-files-changed-in-a-pull-request

Will use this.

.github/workflows/docs-edit-automation.yml
shmsr: After this comment: https://github.com/elastic/integrations/pull/15128#discussion_r2329265059 I don't think we would need commits so not using this action anymore. Resolving this comment.

.github/workflows/docs-edit-automation.yml
shmsr: @v1v @mrodm For now, I have simplified the script. Now in the workflow it only analyzes what packages needs a changelog update and creates a custom command for the author of the PR to run locally and push.

v1v: Thanks @shmsr. I wonder whether we could enable `GitHub copilot` to do the trick if needed. I'll talk with my team internally and see whether that's something feasible and secured. But that's something for the future if we move in that direction, instead of this PR

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,164 @@
...
+Prerequisite: \`go install github.com/elastic/elastic-package@latest\`"
+
+          # Add info about packages already updated
+          if [ -n "${{ steps.analyze.outputs.packages_already_done }}" ]; then
+            COMMENT="$COMMENT
+
+Note: These packages already updated: \`${{ steps.analyze.outputs.packages_already_done }}\`"
+          fi
+          
+          gh pr comment $PR_NUMBER --body "$COMMENT"
```
v1v, line 149: This won't work, `pull_request` event does not have any `write` permissions

.github/workflows/docs-edit-automation.yml
shmsr: Ok, cool! Let me know if Copilot can do this. Thanks a lot!

.github/workflows/docs-edit-automation.yml
shmsr: Ok

shmsr: For now, let me echo the command to run, so that one can pick the command to run from the logs? This is not the ideal solution, but still. Our documentation writers are facing a lot of issues when they do a bulk change across multiple packages.

.github/workflows/docs-edit-automation.yml
```diff
@@ -1,180 +1,164 @@
...
+  elastic-package changelog add --type $CHANGE_TYPE --description \"$CHANGELOG_DESC\" --link \"https://github.com/${{ github.repository }}/pull/$PR_NUMBER\" --next $VERSION_BUMP
+  elastic-package build
+  cd ../..
+done
+git add -u
+git commit -m \"docs: update changelogs and build documentation\"
+git push
+\`\`\`
+
+Prerequisite: \`go install github.com/elastic/elastic-package@latest\`"
```
mrodm: It would be better to use the same `elastic-package` version that it is specified in go.mod, to avoid any possible failure due to different versions between CI and local executions.

Assuming these commands are executed within the working copy of the integrations repository, this command should install the same version defined in `go.mod` (without `@latest`)
```
Prerequisite: \`go install github.com/elastic/elastic-package\`"
```

.github/workflows/docs-edit-automation.yml
```diff
@@ -1,180 +1,164 @@
...
+            MESSAGE="All packages with documentation changes already have updated changelogs and manifests."
+          fi
           
-          gh pr comment $PR_NUMBER --body "$COMMENT"
\ No newline at end of file
+          gh pr comment ${{ github.event.pull_request.number }} --body "## 📚 Documentation Status
+
+$MESSAGE
+
+Your PR is ready for review! 🎉"
```
mrodm: Per this comment, it looks like this command to add a comment in the github PR would not work: 
https://github.com/elastic/integrations/pull/15128/files#r2329597753

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,164 @@
...
+Note: These packages already updated: \`${{ steps.analyze.outputs.packages_already_done }}\`"
+          fi
+          
+          gh pr comment $PR_NUMBER --body "$COMMENT"
+
+      - name: Add labels
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr edit ${{ github.event.pull_request.number }} --add-label "${{ steps.analyze.outputs.label }},documentation"
```
mrodm: Per this comment, probably this command to add a label to the github PR would not work: 
https://github.com/elastic/integrations/pull/15128/files#r2329597753

shmsr: Yes, I will just echo it for now that add this label.

.github/workflows/docs-edit-automation.yml
shmsr: Yes, I will just echo it for now.

.github/workflows/docs-edit-automation.yml
shmsr: Sorry my bad. Missed it. Had planned it but forgot.

shmsr:
@mrodm @v1v I have now echoed the commands that the user has to run. Have addressed all your comments.

.github/workflows/docs-edit-automation.yml
```diff
@@ -0,0 +1,172 @@
...
+  elastic-package changelog add --type $CHANGE_TYPE --description \"$CHANGELOG_DESC\" --link \"https://github.com/${{ github.repository }}/pull/$PR_NUMBER\" --next $VERSION_BUMP
+  elastic-package build
+  cd ../..
+done
+git add -u
+git commit -m \"docs: update changelogs and build documentation\"
+git push
+
+
+Prerequisite: \`go install github.com/elastic/elastic-package@v0.114.0\`"
```
mrodm: As mentioned in https://github.com/elastic/integrations/pull/15128#discussion_r2335975496, assuming that  the user is at the root of the working copy at this moment of the script, it is better to run this command rather than setting a version:
```
```suggestion
Prerequisite: \`go install github.com/elastic/elastic-package\`"
```

This ensures to install v0.112.0, v0.114.0, v0.115.0 ... or any other elastic-package version set in `go.mod` in the corresponding branch. This will help us to not being required to update this workflow for each elastic-package version.

For instance, running that command in an old commit:
```bash
 $ pwd
/home/user/Coding/work/integrations-backport
 $ grep elastic-package go.mod 
	github.com/elastic/elastic-package v0.111.0
 $ go install github.com/elastic/elastic-package
 $ elastic-package version 
2025/09/15 16:53:35  INFO New version is available - v0.114.0. Download from: https://github.com/elastic/elastic-package/releases/tag/v0.114.0
elastic-package v0.111.0 version-hash undefined (build time: unknown)
```

Or running that command in the current main branch:
```shell
 $ pwd
/home/user/Coding/work/integrations-main
 $ grep elastic-package go.mod
	github.com/elastic/elastic-package v0.114.0
 $ go install github.com/elastic/elastic-package
 $ elastic-package version 
2025/09/15 16:56:33  WARN CommitHash is undefined, in both /home/mariorodriguez/.elastic-package/version and the compiled binary, config may be out of date.
elastic-package v0.114.0 version-hash undefined (build time: unknown)
```

shmsr: My bad. I will take of this.

shmsr: Done!

mrodm: Thanks!

mrodm (approved):

shmsr:
I will merge the PR and see if it working. And will create a follow up PR with fixes, if any.

PR #15130, opened 2025-09-02
[SOPHOS UTM] Add GeopIP conditions

SimonKoetting:
Prevent pipeline errors if source/destionation .ip is a empty string

### DIFF
--- a/packages/sophos/changelog.yml
+++ b/packages/sophos/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.15.2"
+  changes:
+    - description: Add conditions to GeoIP processors to prevent failure on empty strings
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/14753
 - version: "3.15.1"
   changes:
     - description: Changed owners.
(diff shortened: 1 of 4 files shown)


packages/sophos/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.15.2"
+  changes:
+    - description: Add conditions to GeoIP processors to prevent failure on empty strings
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/14753
```
qcorporation: please change this to https://github.com/elastic/integrations/pull/15130

qcorporation (commented):
The changes look OK.
Question: are you not seeing these type of behaviour within any of the other geoip: process.
I see similar processing within 

- utm/../dhcp.yml
- utm/../dns.yml
- xg/../default.yml

packages/sophos/changelog.yml
SimonKoetting: was already commited. just forgot to push.. :D

SimonKoetting:
I observed the issue so far just on the http & packet filter, but you're right, it's the same on the other geoip processors. I added that as well

PR #15131, opened 2025-09-02
[Zscaler ZPA] Add Convert conditions

SimonKoetting:
Add condition to convert processors as it can happens that the publicip is a empty string

### DIFF
--- a/packages/zscaler_zpa/changelog.yml
+++ b/packages/zscaler_zpa/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "1.23.2"
   changes:
-    - description: Add conditions to convert processors to prevent failure on empty strings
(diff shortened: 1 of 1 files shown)


packages/zscaler_zpa/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.23.2"
+  changes:
+    - description: Add conditions to convert processors to prevent failure on empty strings
```
efd6:
```suggestion
    - description: Add conditions to convert processors to prevent failure on empty strings.
```

efd6 (commented):
LGTM after nit is addressed.

SimonKoetting:
Thanks for approving, @efd6. Could you please also merge? I'm not allowed to :/

PR #15149, opened 2025-09-03
fix(kubernetes): rename orchestrator.cluster.id to cloud.project.id, set cloud.region for GKE audit logs

lyoung-confluent:
Currently the Kubernetes audit log ingest pipeline extracts the `project_id` label from the event and _incorrectly_ assigns the value to `orchestrator.cluster.id `. This is wrong, the `project_id` is actually the `cloud.project.id` where the event occurred, not the unique identifier of the source Kubernetes cluster. Additionally, I set `cloud.region` based on the `location` label (ex: `us-east4`) that is present on GKE audit logs.

### DIFF
--- a/packages/kubernetes/changelog.yml
+++ b/packages/kubernetes/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.85.0"
+  changes:
+    - description: Add cloud.region and cloud.project.id fields to the audit logs data stream.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/15149
 - version: "1.84.0"
   changes:
     - description: Add kubernetes.container.status.last_terminated_exitcode field to the state_container data stream.
--- a/packages/kubernetes/manifest.yml
+++ b/packages/kubernetes/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.1.2
 name: kubernetes
 title: Kubernetes
-version: 1.84.0
+version: 1.85.0
 description: Collect logs and metrics from Kubernetes clusters with Elastic Agent.
 type: integration
 categories:


lyoung-confluent:
FYI @andrewkroh I got the CLA signed by our legal folks, apologies for the delay

lyoung-confluent:
Adding a 👍 comment for the bot to confirm this PR is not stale, still relevant

gizas:
Yes please update the [changelog](https://github.com/elastic/integrations/blob/main/packages/kubernetes/changelog.yml#L2) file by adding a new netry and also update the version of the [manifest](https://github.com/elastic/integrations/blob/main/packages/kubernetes/manifest.yml#L4) with your new version

chemamartinez:
@lyoung-confluent I hope you don't mind I took the liberty to update the branch and add the pending changes by myself.

@gizas it is ready for review now.

packages/kubernetes/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -214,11 +214,6 @@ processors:
       field: orchestrator.cluster.name
       copy_from: kubernetes.audit.resource.labels.cluster_name
       ignore_empty_value: true
-  - set:
```
gizas, line 217: Why to delete this. This is also an ECS field https://www.elastic.co/docs/reference/ecs/ecs-orchestrator#field-orchestrator-cluster-id

I dont like the idea to introduce a breaking change here. You can have both orchestrator.cluster.id and cloud.project.id

chemamartinez: According to the PR description, it was wrongly assigned to `orchestrator.cluster.id`.

Maybe @lyoung-confluent can provide any docs reference that supports that statement.

Issue #14839, opened 2025-08-06
[elastic_security] Remove Auth Type Dropdown


mohitjha-elastic:
**Description:**

Currently, the elastic_security integration includes a configuration parameter named `auth_type`, which is a dropdown menu with the following options:

- api_auth
- basic_auth
- bearer_auth

However, users often skip selecting the `auth_type `explicitly and instead directly enter credentials (such as username/password or tokens), which results in authentication errors due to a mismatch between the input and the selected (or unselected) auth type.

**Proposed Change**
To improve user experience and prevent misconfiguration:

Remove the auth_type configuration parameter from the integration UI.
Handle authentication type selection programmatically in CEL (backend) with the following priority:
- Use api_auth if API credentials are provided.
- Fallback to basic_auth if username/password is detected.
- Fallback to bearer_auth if only bearer token is available.
This automatic handling will remove the need for manual selection and ensure smoother authentication handling.

**Additional Updates**

- Update the README documentation to reflect this change.
- Clearly mention the priority order used for authentication (api_auth → basic_auth → bearer_auth).

**Benefits**

- Simplifies configuration for users.
- Prevents frequent auth-related errors due to missing or mismatched config values.
- Reduces confusion around which authentication method to select.

narph:
https://github.com/elastic/kibana/pull/211863

andrewkroh:
I tested the `required_vars` package-spec feature in 9.2.0-SNAPSHOT. (The [specification](https://github.com/elastic/package-spec/blob/f602e78226cf590c24fe8b9a164bbaed5e7bffac/spec/integration/data_stream/manifest.spec.yml#L35-L56) leaves a lot to the imagination of the reader 🫤.)

With this added configuration, it validates, for example, that `username` and `password` are set if Basic Auth is selected. I didn't investigate which versions of Kibana support this, but when I first tested on 9.0.4, there wasn't enough error information provided.

From my testing in 9.2.0-SNAPSHOT, as long as one of the required variables is satisfied, it won't cause an error if another is set For instance, if Auth Type is Basic Auth and you set `username`, `password`, and `api_key`, it will not error because the `api_key` being set (i.e. it does not force mutual exclusivity).

<img width="661" height="786" alt="Image" src="https://github.com/user-attachments/assets/a0bd841e-9ecd-4494-be3e-35e2a03f598d" />


```diff
[...]

PR #15253, opened 2025-09-09
[elastic_security] Add Support for Conditional Required Fields to Handle the Auth Type

mohitjha-elastic:
## Proposed commit message

```
elastic_security: add support for conditional required fields to handle the auth type.

This update introduces support for conditional required fields in the elastic_security module to better 
handle different authentication types.
Added logic to dynamically enforce required fields based on the selected auth_type.
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/elastic_security directory.

> elastic-package test -v

## Related issues

- Closes https://github.com/elastic/integrations/issues/14839

elastic_security-ss

### DIFF
--- a/packages/elastic_security/data_stream/alert/manifest.yml
+++ b/packages/elastic_security/data_stream/alert/manifest.yml
@@ -5,11 +5,24 @@ streams:
     title: Elastic Security Alerts
     description: Collect Alerts from Elastic Security.
     template_path: cel.yml.hbs
+    required_vars:
+      api_key:
+        - name: auth_type
+          value: api_auth
+        - name: api_key
+      basic_auth:
+        - name: auth_type
+          value: basic_auth
+        - name: username
+        - name: password
+      bearer_auth:
+        - name: auth_type
+          value: bearer_auth
+        - name: bearer_token
     vars:
       - name: auth_type
         title: Authentication Type
         show_user: true
-        required: true
         multi: false
         type: select
         options:
--- a/packages/elastic_security/changelog.yml
+++ b/packages/elastic_security/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.4.0"
+  changes:
+    - description: Add support for conditional required fields to handle the authentication type.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
 - version: "0.3.2"
   changes:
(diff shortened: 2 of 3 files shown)


packages/elastic_security/data_stream/alert/manifest.yml
```diff
@@ -5,11 +5,24 @@ streams:
     title: Elastic Security Alerts
     description: Collect Alerts from Elastic Security.
     template_path: cel.yml.hbs
+    required_vars:
+      api_key:
+        - name: auth_type
+          value: api_auth
+        - name: api_key
```
chrisberkhout:
```suggestion
      api_auth:
        - name: auth_type
          value: api_auth
        - name: api_key
```

packages/elastic_security/data_stream/alert/manifest.yml
```diff
@@ -5,11 +5,24 @@ streams:
...
+        - name: password
+      bearer_auth:
+        - name: auth_type
+          value: bearer_auth
+        - name: bearer_token
     vars:
       - name: auth_type
         title: Authentication Type
         show_user: true
-        required: true
```
chrisberkhout, line 12: Can this stay `required: true`?

In the screenshot the "Authentication Type" box has "Optional" on the top right. I understand the `required_vars` settings will indirectly require it to be set, but if we can make that "Optional" label go away it's clearer for the user.

chrisberkhout (changes requested):
In the proposed commit message, please use "package" or "integration" instead of "module", because [modules](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-modules) are different.

Modules have some of the functionality of integrations but in they are included directly inside Auditbeat/Filebeat/Metricbeat/Winlogbeat instead of being managed by Elastic Agent.

packages/elastic_security/data_stream/alert/manifest.yml
mohitjha-elastic: Nope. Elastic package does not allow this to be set as `required: true`.
It throws the below error -
`required var "auth_type" in optional group is defined as always required.`

chrisberkhout: Thanks

chrisberkhout (approved):

PR #15267, opened 2025-09-10
Kafka dashboards for consumer producer datasets

stefans-elastic:
## Checklist

## Author's Checklist

- [ ] 

## Related issues

- 

## Screenshots

metricbeat-kafka-consumer
metricbeat-kafka-producer

### DIFF
--- a/packages/kafka/kibana/dashboard/kafka-32a191b0-c8ac-4ea1-812d-0e3d406f9a41.json
+++ b/packages/kafka/kibana/dashboard/kafka-32a191b0-c8ac-4ea1-812d-0e3d406f9a41.json
@@ -2190,5 +2190,5 @@
         }
     ],
     "type": "dashboard",
-    "typeMigrationVersion": "10.3.0"
+    "typeMigrationVersion": "8.9.0"
 }


agithomas:
Please find the review comments here


Screenshot 2025-10-07 at 10 23 45 AM

Can we use a more suitable aggregation function here : 

average() for bytes_consumed, records_consumed, in.bytes_per_sec, fetch_rate

maximum() for max_lag

stefans-elastic:
> In the `fields.yml` description, the description of the field is - The average number of records consumed per second for a specific topic.
> In this case, why the metrics is grouped by clients? Shouldn't it be topic?

The main problem here is that the records don't contain topic. Here is an example of such event's data:
```
 "kafka": {
      "consumer": {
        "bytes_consumed": 38220,
        "client_id": "client1",
        "fetch_rate": 2.179592,
        "mbean": "kafka.consumer:type=consumer-fetch-manager-metrics,client-id=client1",
        "metric_fingerprint": "r4AVV0cN6yZcEtu0qJvZDM6yLSw=",
        "records_consumed": 766
      }
    },
```

And event if it had the data the dashboard is called "Consumer" so I think it makes sense that data shown here is focused on consumer. If we wanted to show something like this I think it should have been in "Topic" dashboard.

agithomas:
This comment is for the consumer dashboard.

In some of the groupBy, the top 3 values are considered. Can you make it either top 5 or top 10?

Recommendation: Keep legends below the X-axis.  Other dashboards in Kafka follow this approach.

agithomas:
Feedback for the producer dashboard

image
Screenshot 2025-10-07 at 1 42 00 PM

agithomas:
**Important addition for all the dashboards**

Can you please additionally include a dropdown filter for `service.address` in all the existing dashboard, other than Replica manager (It already exists in Replica manager) ? Also, please remember to change the label to `Service address` (similar to replica manager dashboard) so that the dashboard can handle multiple instances of Kafka, without using the inline dashboard filters.Please keep this as the first input control - before the Topic name filter ,  Partition filter etc.

PR #15268, opened 2025-09-10
Change stack version requirements due to the breaking changes

rdner:
The metric name `queue.filled.pct.events` was changed to `queue.filled.pct` in v8.15.0

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [ ] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## Related issues

-

PR #15273, opened 2025-09-10
[Backport] Change stack version requirements due to the breaking changes

rdner:
## Proposed commit message

The metric name `queue.filled.pct.events` was changed to `queue.filled.pct` in v8.15.0 by this commit https://github.com/elastic/beats/commit/f8aedce388312b782a19b053508541a0901dbf18 (PR https://github.com/elastic/beats/pull/39774).

We've fixed the metric name in the package fields by this PR https://github.com/elastic/integrations/pull/15244

However, the minimal stack version requirements remained unchanged. This would break any customers running on stack versions 8.11.2 – 8.14.3 and upgrading their Elastic Agent integration to 2.5.2.

## Checklist

## Related issues

- Relates https://github.com/elastic/integrations/pull/15268

### DIFF
--- a/packages/elastic_agent/changelog.yml
+++ b/packages/elastic_agent/changelog.yml
@@ -1,10 +1,5 @@
 # newer versions go on top
 
-- version: "2.5.2"
-  changes:
-    - description: Update `queue.filled.pct.events` to `queue.filled.pct` metric name 
-      type: enhancement
-      link: "https://github.com/elastic/integrations/pull/15244"
 - version: "2.5.1"
   changes:
     - description: Change Agent metrics dashboard memory source fields
--- a/packages/elastic_agent/changelog.yml
+++ b/packages/elastic_agent/changelog.yml
@@ -1,5 +1,15 @@
 # newer versions go on top
 
+- version: "2.5.3"
+  changes:
+    - description: Revert metric name changes, so the package is still usable on stacks < 8.15.0
+      type: bugfix
+      link: "https://github.com/elastic/integrations/pull/15273"
+- version: "2.5.2"
+  changes:
+    - description: Update `queue.filled.pct.events` to `queue.filled.pct` metric name
+      type: enhancement
+      link: "https://github.com/elastic/integrations/pull/15244"
 - version: "2.5.1"
   changes:
     - description: Change Agent metrics dashboard memory source fields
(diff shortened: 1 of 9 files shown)


VihasMakwana (approved):

packages/elastic_agent/manifest.yml
```diff
@@ -1,13 +1,13 @@
 name: elastic_agent
 title: Elastic Agent
-version: 2.5.2
+version: 2.5.3
```
VihasMakwana, line 3: This should be `2.6.0` I guess?

rdner: No, the previous PR was to introduce a new minor version 2.6.0. This time I backport changes from 2.6.0 to 2.5.x, so it needs to be 2.5.3.

This was all suggested by @jsoriano in https://github.com/elastic/integrations/pull/15268#discussion_r2336341263

VihasMakwana: Got it. Thanks!

packages/elastic_agent/manifest.yml
```diff
@@ -1,13 +1,13 @@
...
 description: Collect logs and metrics from Elastic Agents.
 type: integration
 format_version: 3.1.4
 categories: ["elastic_stack"]
 conditions:
   kibana:
-    version: "^8.11.2 || ^9.0.0"
+    version: "^8.15.0 || ^9.0.0"
```
jsoriano: This branch is intended for `^8.11.2 || ^9.0.0`, right?

packages/elastic_agent/changelog.yml
```diff
@@ -1,8 +1,13 @@
 # newer versions go on top
 
+- version: "2.5.3"
+  changes:
+    - description: Update version requirements due to the breaking change in metric names
+      type: bugfix
+      link: "https://github.com/elastic/integrations/pull/15273"
```
jsoriano: We should just revert the breaking-change, not changing the kibana constraints.

jsoriano (commented):
We should revert the breaking change on this branch, and keep the previous kibana constraint, so the fix reaches affected installations.

rdner:
After discussing this with @cmacknz we make a decision to close this PR.

The stack constraint wouldn't prevent the original issue. This stack constraint does not control the Elastic Agent version and the breaking change took place in Elastic Agent, not in the stack itself.

PR #15279, opened 2025-09-10
Prefer set with copy_from

joegallo:
## Proposed commit message

Prefer the `copy_from` option of the `set` processor for high certain high volume integrations.

When the field to be copied is already just a string, and so the set processor with mustache isn't being used for the side effect of converting to a string, then it's quite a bit faster to use `copy_from` rather than `value` (with mustache templating).

For example, in a large cluster that I was looking at a few minutes ago, the `set` most expensive single `set` processor is this one:

```
{
  "set": {
    "field": "cloud.account.id",
    "if": "ctx.aws?.vpcflow?.account_id != null",
    "value": "{{aws.vpcflow.account_id}}"
  }
}
```

It's taking 2.8 microseconds per doc, as compared to the average of all set processor invocations for the same pipeline which is only .6 microseconds per doc. The cluster in question is processing billions and billions of documents per hour, though, so microseconds add up (and this particularly-expensive set processor is the eighth-most expensive processor for the entire pipeline).

-----

I'm marking this a draft because I'm not 100% absolutely sure about my changes or the correct process for PRs on this repo.

### DIFF
--- a/packages/aws/data_stream/vpcflow/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/aws/data_stream/vpcflow/elasticsearch/ingest_pipeline/default.yml
@@ -284,11 +284,11 @@ processors:
   - set:
       if: ctx.aws?.vpcflow?.account_id != null
       field: cloud.account.id
-      value: '{{aws.vpcflow.account_id}}'
+      copy_from: aws.vpcflow.account_id
   - set:
       if: 'ctx?.aws?.vpcflow?.instance_id != null && ctx.aws.vpcflow.instance_id != "-"'
       field: cloud.instance.id
-      value: '{{aws.vpcflow.instance_id}}'
+      copy_from: aws.vpcflow.instance_id
   - convert:
      field: aws.vpcflow.packets_lost_no_route
      type: long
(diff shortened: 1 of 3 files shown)


packages/aws/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "4.3.3"
```
efd6:
```suggestion
- version: "4.4.0"
```
enhancements bump minor, also update manifest.

packages/gcp/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.44.1"
```
efd6:
```suggestion
- version: "2.45.0"
```
enhancements bump minor, also update manifest.

packages/aws/changelog.yml
joegallo: 👍, 908c362bbd25ff58f09facf1a9d0dda367a8274b

packages/gcp/changelog.yml
joegallo: 👍, 908c362bbd25ff58f09facf1a9d0dda367a8274b

PR #9765, opened 2024-04-30
Add panel to dashboard, add percent field

fearful-symmetry:
closes: https://github.com/elastic/beats/issues/38708
For more context: https://github.com/elastic/beats/pull/39205

This PR adds support in integrations for the new queue percentage full metric. As part of this, we:

- Add a `Queue Usage` panel to the agent metrics dashboard
- Add a `queue.filled.pct.events` field to the integrations

I haven't tinkered with integrations in ages, so...I hopefully haven't broken anything.

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).

Screenshot 2024-05-01 at 1 49 16 PM

PR #15280, opened 2025-09-10
[elastic_agent] Support new pipeline fields from beats logs

jsoriano:
## Proposed commit message

Support fields added and fixed for pipeline metrics in Beats 8.15.0.

* Definitions added for new fields.
* Ingest pipeline to adapt renamed field.
* Revert kibana version constraint to `^8.11.2 || ^9.0.0`.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Beats change in https://github.com/elastic/beats/pull/39774

### DIFF
--- a/packages/elastic_agent/data_stream/auditbeat_logs/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/elastic_agent/data_stream/auditbeat_logs/elasticsearch/ingest_pipeline/default.yml
@@ -1,12 +1,11 @@
 ---
 description: "Pipeline for adapting Elastic Agent fields"
 processors:
-  - script:
-      lang: painless
-      description: Rename field from older versions of Beats.
-      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map && ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events != null
-      source: |
-          ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct = ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events;
+  - remove:
+      description: "Remove object that contained invalid data and can cause mapping conflicts"
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
+      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map
+      ignore_missing: true
 
 on_failure:
   - set:
--- a/packages/elastic_agent/data_stream/cloudbeat_logs/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/elastic_agent/data_stream/cloudbeat_logs/elasticsearch/ingest_pipeline/default.yml
@@ -1,12 +1,11 @@
 ---
 description: "Pipeline for adapting Elastic Agent fields"
 processors:
-  - script:
-      lang: painless
-      description: Rename field from older versions of Beats.
-      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map && ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events != null
-      source: |
-          ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct = ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events;
+  - remove:
+      description: "Remove object that contained invalid data and can cause mapping conflicts"
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
+      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map
+      ignore_missing: true
 
 on_failure:
   - set:
--- a/packages/elastic_agent/data_stream/filebeat_logs/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/elastic_agent/data_stream/filebeat_logs/elasticsearch/ingest_pipeline/default.yml
@@ -1,12 +1,11 @@
 ---
 description: "Pipeline for adapting Elastic Agent fields"
 processors:
-  - script:
-      lang: painless
-      description: Rename field from older versions of Beats.
-      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map && ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events != null
-      source: |
-          ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct = ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events;
+  - remove:
+      description: "Remove object that contained invalid data and can cause mapping conflicts"
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
+      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map
+      ignore_missing: true
 
 on_failure:
   - set:
--- a/packages/elastic_agent/data_stream/heartbeat_logs/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/elastic_agent/data_stream/heartbeat_logs/elasticsearch/ingest_pipeline/default.yml
@@ -1,12 +1,11 @@
 ---
 description: "Pipeline for adapting Elastic Agent fields"
 processors:
-  - script:
-      lang: painless
-      description: Rename field from older versions of Beats.
-      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map && ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events != null
-      source: |
-          ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct = ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events;
+  - remove:
+      description: "Remove object that contained invalid data and can cause mapping conflicts"
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
+      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map
+      ignore_missing: true
 
 on_failure:
   - set:
--- a/packages/elastic_agent/data_stream/metricbeat_logs/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/elastic_agent/data_stream/metricbeat_logs/elasticsearch/ingest_pipeline/default.yml
@@ -1,12 +1,11 @@
 ---
 description: "Pipeline for adapting Elastic Agent fields"
 processors:
-  - script:
-      lang: painless
-      description: Rename field from older versions of Beats.
-      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map && ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events != null
-      source: |
-          ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct = ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events;
+  - remove:
+      description: "Remove object that contained invalid data and can cause mapping conflicts"
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
+      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map
+      ignore_missing: true
 
 on_failure:
   - set:
--- a/packages/elastic_agent/data_stream/osquerybeat_logs/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/elastic_agent/data_stream/osquerybeat_logs/elasticsearch/ingest_pipeline/default.yml
@@ -1,12 +1,11 @@
 ---
 description: "Pipeline for adapting Elastic Agent fields"
 processors:
-  - script:
-      lang: painless
-      description: Rename field from older versions of Beats.
-      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map && ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events != null
-      source: |
-          ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct = ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events;
+  - remove:
+      description: "Remove object that contained invalid data and can cause mapping conflicts"
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
+      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map
+      ignore_missing: true
 
 on_failure:
   - set:
--- a/packages/elastic_agent/data_stream/packetbeat_logs/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/elastic_agent/data_stream/packetbeat_logs/elasticsearch/ingest_pipeline/default.yml
@@ -1,12 +1,11 @@
 ---
 description: "Pipeline for adapting Elastic Agent fields"
 processors:
-  - script:
-      lang: painless
-      description: Rename field from older versions of Beats.
-      if: ctx.monitoring?.metrics?.libbeat?.pipeline?.queue?.filled?.pct instanceof Map && ctx.monitoring.metrics.libbeat.pipeline.queue.filled.pct.events != null
-      source: |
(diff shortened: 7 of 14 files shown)


packages/elastic_agent/data_stream/auditbeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
+---
+description: "Pipeline for adapting Elastic Agent fields"
+processors:
+  # Rename field from older versions of Beats.
+  - rename:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
+      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```
rdner:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```

The `queue.filled.pct.events` was renamed to `queue.filled.pct`. I think this is the right rename then?

packages/elastic_agent/data_stream/cloudbeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
+---
+description: "Pipeline for adapting Elastic Agent fields"
+processors:
+  # Rename field from older versions of Beats.
+  - rename:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
+      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```
rdner:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```

packages/elastic_agent/data_stream/filebeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
+---
+description: "Pipeline for adapting Elastic Agent fields"
+processors:
+  # Rename field from older versions of Beats.
+  - rename:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
+      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```
rdner:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```

packages/elastic_agent/data_stream/heartbeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
+---
+description: "Pipeline for adapting Elastic Agent fields"
+processors:
+  # Rename field from older versions of Beats.
+  - rename:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
+      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```
rdner:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```

packages/elastic_agent/data_stream/metricbeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
+---
+description: "Pipeline for adapting Elastic Agent fields"
+processors:
+  # Rename field from older versions of Beats.
+  - rename:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
+      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```
rdner:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```

packages/elastic_agent/data_stream/osquerybeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
+---
+description: "Pipeline for adapting Elastic Agent fields"
+processors:
+  # Rename field from older versions of Beats.
+  - rename:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
+      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```
rdner:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```

packages/elastic_agent/data_stream/packetbeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
+---
+description: "Pipeline for adapting Elastic Agent fields"
+processors:
+  # Rename field from older versions of Beats.
+  - rename:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
+      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```
rdner:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct.events
      target_field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```

packages/elastic_agent/data_stream/auditbeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
...
+      ignore_missing: true
+  - remove:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```
rdner, line 6:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```

packages/elastic_agent/data_stream/cloudbeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
...
+      ignore_missing: true
+  - remove:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```
rdner, line 6:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```

packages/elastic_agent/data_stream/filebeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
...
+      ignore_missing: true
+  - remove:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```
rdner, line 6:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```

packages/elastic_agent/data_stream/heartbeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
...
+      ignore_missing: true
+  - remove:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```
rdner, line 6:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```

packages/elastic_agent/data_stream/metricbeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
...
+      ignore_missing: true
+  - remove:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```
rdner, line 6:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```

packages/elastic_agent/data_stream/osquerybeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
...
+      ignore_missing: true
+  - remove:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```
rdner, line 6:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```

packages/elastic_agent/data_stream/packetbeat_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,11 @@
...
+      ignore_missing: true
+  - remove:
+      field: monitoring.metrics.libbeat.pipeline.queue.filled.pct
```
rdner, line 6:
```suggestion
      field: monitoring.metrics.libbeat.pipeline.queue.filled.events
```

packages/elastic_agent/data_stream/auditbeat_logs/fields/fields.yml
```diff
@@ -70,8 +70,44 @@
     - name: queue.max_events
       type: long
       metric_type: counter
-      description: Maximum number of events in a queue
+      description: Maximum number of events in a queue if it has one, otherwise zero.
```
rdner, line 73: Why are there so many changes in field files? Were we missing so many fields before?

jsoriano: According to the PR in Beats all these fields have been added: https://github.com/elastic/beats/pull/39774

Haven't been able to confirm.

packages/elastic_agent/data_stream/auditbeat_logs/elasticsearch/ingest_pipeline/default.yml
jsoriano: I assumed that being called `events`, it was a count of events, but I would like to confirm with real metrics.

faec: It's a ratio, and the name was a mistake (my fault). It is a slightly less accurate version of the current `queue.filled.pct`. But I also thought `pct.events` was fixed before it hit an actual release, I will check on that.

jsoriano: Ok, if it is a ratio I will apply Denis changes, thanks!

faec:
This is a meta-comment to clarify what the situation was as of ~3 days ago and how it got that way.

Timeline:
- April 30, 2024: `pipeline.queue.filled.pct.events` is first defined in https://github.com/elastic/beats/pull/39205, released in 8.15.0. The name is a mistake (my fault).
- May 2, 2024: The `elastic_agent` package is updated to list `queue.filled.pct.events` as one of its fields. https://github.com/elastic/integrations/pull/9765
- June 11, 2024: Existing queue metrics are fixed and several new ones are added. `pipeline.queue.filled.pct.events` is renamed to `pipeline.queue.filled.pct`. https://github.com/elastic/beats/pull/39774, released in 8.15.0.
- August 9, 2024: A bug that prevented the queue metrics from being registered properly is fixed. https://github.com/elastic/beats/pull/40480, released in 8.15.1.
- January 29, 2025: The new queue metrics are added to Metricbeat's schema for the `beat/stats` metricset, so they will be ingested by Metricbeat monitoring. https://github.com/elastic/beats/pull/42439, released in 8.17.2 and 8.18.0.
- February 4, 2025: The tracking issue to properly display the new metrics in the Agent dashboards is closed, seemingly due to a misunderstanding about the scope of what had been completed so far https://github.com/elastic/beats/issues/42093 (I think @strawgate was working on some of that? But whether because of this issue closure or something else, it looks like that followup was incomplete.)


Commentary:
- The queue `filled` metric never existed under _any_ name until 8.15.0, which included both the initial definition (https://github.com/elastic/beats/pull/39205) and the naming fix (https://github.com/elastic/beats/pull/39774); 8.14.3 and earlier included neither.
- Neither the `filled` metric nor any of the other new ones were ingested by Metricbeat monitoring until v8.17.2 and later.
- Until today, I believed no released version ever reported metrics under the mistaken name `queue.filled.pct.events`, because the name was corrected before it hit a release. This was _almost_ true, however:
  * There was an "inert" occurrence of the old name, with no references, that was missed in the initial fix, and stayed there until the Metricbeat schema update (https://github.com/elastic/beats/pull/42439/files#diff-558176dd2327daca24e089ce7ec773653e787ef7438af3d3df257e655b383e65L133).
  * The schema update PR cleaned up a few "redundant" metrics variable definitions that had no code references, since the same variables were created and used elsewhere, not noticing that the "redundant" `filled` metric had used the old / incorrect name. [But of course, it was unreferenced and absent from the Metricbeat schema, so there was no way for it to actually show up anywhere except the logs, right?]
  * The real queue metrics also cleared their registry on initialization, so the old name only existed for a short window during pipeline initialization when the queue had not yet been created.
  * _Except_, the bug fixed in https://github.com/elastic/beats/pull/40480 prevented the old name from being removed in 8.15.0.
  * If 8.15.0 was run with self-monitoring (self-monitoring doesn't enforce the schema restriction of Metricbeat monitoring, so it includes all variables), `queue.filled.pct.events` would be ingested, always with a value of zero.
  * Also, for completeness: Between versions 8.15.1 and 8.17.1 (inclusive), if self-monitoring is enabled, _and_ a metrics scan interval triggers in the brief window before the queue is created (I'm not sure this is actually possible, but if so it's probably extremely rare), a (zero) `queue.filled.pct.events` field would be ingested.
- The upshot is: running with self-monitoring on _precisely_ version 8.15.0, then upgrading to a later version with self-monitoring, or to 8.17.2 or later with any monitoring, can cause a mapping conflict where the earlier `queue.filled.pct.events` field interferes with the `queue.filled.pct` field used by all other versions.
- One important consequence of all this: in trying to untangle this mess, it is _always_ safe to delete any field with the name `queue.filled.pct.events`. To the extent it ever appeared at all, it was always zero.
- This still leaves the field definitions in the Agent package, which still use the old field name, and would seem to conflict with the real monitoring values since 8.17.2, although I'm not sure exactly how a nested example like this would be handled -- if that did cause a mapping conflict, I would expect it to be an extremely common error, which doesn't seem to have been the case? So that should be addressed, but our recent issues might be an example of the former corner case instead.

jsoriano:
Thanks for the explanation!

> * The upshot is: running with self-monitoring on _precisely_ version 8.15.0, then upgrading to a later version with self-monitoring, or to 8.17.2 or later with any monitoring, can cause a mapping conflict where the earlier `queue.filled.pct.events` field interferes with the `queue.filled.pct` field used by all other versions.

Then this PR and https://github.com/elastic/integrations/pull/15244 should fix the issue in the integration, right?
* https://github.com/elastic/integrations/pull/15244 fixes the mapping.
* https://github.com/elastic/integrations/pull/15280 (this one) ensures that only the correct mapping is used, and adds additional mappings for the other fields added in https://github.com/elastic/beats/pull/39774.

> Between versions 8.15.1 and 8.17.1 (inclusive), if self-monitoring is enabled, and a metrics scan interval triggers in the brief window before the queue is created (I'm not sure this is actually possible, but if so it's probably extremely rare), a (zero) queue.filled.pct.events field would be ingested.

The pipelines I am adding in this PR would be intended to handle these cases. If this is so rare, is it worth it to add them?

> * The queue `filled` metric never existed under _any_ name until 8.15.0

This explains why I didn't manage to reproduce the issue with the oldest version supported by the package (8.11.2).

cmacknz:
> If this is so rare, is it worth it to add them?

IMO yes we already have a support case about this, let's just put in the fixes and then not have to think about or explain this again.

faec:
> If this is so rare, is it worth it to add them?

In addition to Craig's comment: I'd expect that the fix for that case is the same as the (presumably more common) one of running 8.15.0 with self-monitoring -- the correct behavior with any version of Beats is always to delete the `queue.filled.pct.events` field if it exists, since in any case where it could have existed it didn't contain real data.

jsoriano:
> > If this is so rare, is it worth it to add them?
> 
> In addition to Craig's comment: I'd expect that the fix for that case is the same as the (presumably more common) one of running 8.15.0 with self-monitoring -- the correct behavior with any version of Beats is always to delete the `queue.filled.pct.events` field if it exists, since in any case where it could have existed it didn't contain real data.

Ok, so instead of renaming it I will remove it.

jsoriano:
Btw, another follow up issue created, to have system tests on this package: https://github.com/elastic/elastic-package/issues/2903

PR #15287, opened 2025-09-11
[airflow] Make Airflow package GA

mykola-elastic:
Work done in this PR so far (in progress):
- bump version to `1.0.0`
- update `ecs` version to `8.17.0` https://github.com/elastic/integrations/issues/11952
- update Readme (+ check that links work)
- test with Airflow 2.4.3 and 3.0.6
- test with Elastic Stack 8.13, 8.17, 9.0, 9.1
- deal with `undefined` on the integration configuration page
- updated system tests (filled metrics log file with 100 rows of data recorded from Airflow 3.0)
- remake dashboard

## Proposed commit message

See title.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- 

## Screenshots

Screenshot 2025-10-07 at 07 43 54

Screenshot 2025-10-06 at 10 04 10

### DIFF
--- a/packages/airflow/_dev/build/docs/README.md
+++ b/packages/airflow/_dev/build/docs/README.md
@@ -12,7 +12,7 @@ Use the Airflow integration to:
 
 ## Compatibility
 
-The Airflow package is tested with Airflow `2.4.0` and `3.0.6`. It will work with any `2.*` and `3.*` versions, but uses a smaller subset of metrics to be backwards-compatible with Apache Airflow 2.0.
+The Airflow package is tested with Airflow `2.4.3`. It should work with any `2.*` version.
 
 ## What data does this integration collect?
 
--- a/packages/airflow/docs/README.md
+++ b/packages/airflow/docs/README.md
@@ -12,7 +12,7 @@ Use the Airflow integration to:
 
 ## Compatibility
 
-The Airflow package is tested with Airflow `2.4.0` and `3.0.6`. It will work with any `2.*` and `3.*` versions, but uses a smaller subset of metrics to be backwards-compatible with Apache Airflow 2.0.
+The Airflow package is tested with Airflow `2.4.3`. It should work with any `2.*` version.
 
 ## What data does this integration collect?
 


mykola-elastic:
@alaudazzi PR is still in progress, but README can already be reviewed. Can you add reviewing the README file to your task list?

packages/airflow/_dev/build/docs/README.md
```diff
@@ -23,17 +27,11 @@ Data stream:
...
 ## What do I need to use this integration?
 
 Users require Elasticsearch to store and search user data, and Kibana to visualize and manage it. They can utilize the hosted Elasticsearch Service on Elastic Cloud, which is recommended, or self-manage the Elastic Stack on their own hardware.
 
-To ingest data from Airflow, users must have [StatsD](https://github.com/statsd/statsd) to receive the same.
-
-## Setup
+## How do I deploy this integration?
 
 For step-by-step instructions on how to set up an integration, see the [Getting started](https://www.elastic.co/guide/en/starting-with-the-elasticsearch-platform-and-its-solutions/current/getting-started-observability.html) guide.
```
alaudazzi:
```suggestion
For step-by-step instructions on how to set up an integration, refer to the [Getting started](https://www.elastic.co/guide/en/starting-with-the-elasticsearch-platform-and-its-solutions/current/getting-started-observability.html) guide.
```

alaudazzi (approved):
Left a very minor suggestion, otherwise LGTM.

mykola-elastic:
Original dashboard for comparison (**BEFORE**):

Screenshot 2025-10-06 at 10 07 15


**AFTER**:
Screenshot 2025-10-07 at 07 43 54

Screenshot 2025-10-06 at 10 04 10

CC @daniela-elastic

mykola-elastic:
Fixed the CI and marked ready for review
CC @ishleenk17

packages/airflow/manifest.yml
```diff
@@ -10,40 +10,31 @@ categories:
   - process_manager
 conditions:
   kibana:
-    version: "^8.13.0 || ^9.0.0"
+    version: "^8.17.0 || ^9.0.0"
```
ishleenk17: Why the kibana version bump ?

packages/airflow/manifest.yml
```diff
@@ -10,40 +10,31 @@ categories:
...
+  - src: /img/dashboard_continued.png
+    title: Airflow Overview Metrics screenshot (continued)
+    size: 600x600
+    type: image/png
 icons:
   - src: /img/airflow.svg
     title: Airflow logo
     size: 32x32
     type: image/svg+xml
-vars:
```
ishleenk17, line 21: Not that it would matter much, but any specific readson to move it to datastream ?
since there is a single datastream

packages/airflow/docs/README.md
```diff
@@ -10,7 +10,11 @@ Use the Airflow integration to:
 - Create informative visualizations to track usage trends, measure key metrics, and derive actionable business insights.
 - Monitor your workflows' performance and status in real-time.
 
-## Data streams
+## Compatibility
+
+The Airflow package is tested with Airflow `2.4.0` and `3.0.6`. It should work with any `2.*` and `3.*` versions.
```
ishleenk17: Hoping no backward compatibility breakages here between 2.0 and 3.0 ?

ishleenk17:
In dashboard, the DAG BagSize, DAG's on the y axis is the DAG count ?

packages/airflow/manifest.yml
mykola-elastic: to remove `undefined` from the configuration page in 9.0 and 9.1 (I'll add a screenshot soon - how it looked like before the change)

mykola-elastic:
> In dashboard, the DAG BagSize, DAG's on the y axis is the DAG count ?

Yes, it is measured in DAGs. I didn't include the word "Count" as it is considered redundant in the new dashboard guidelines

packages/airflow/docs/README.md
mykola-elastic: no, I am using only metrics available in 2.0

packages/airflow/manifest.yml
mykola-elastic: That's how it looked like before (screenshot is from 9.1, in old kibanas it looks fine)
Screenshot 2025-10-08 at 08 00 07

packages/airflow/manifest.yml
mykola-elastic: I used newer Kibanas to remake the dashboard from scratch, so I dropped those older ones. If the 8.13-8.16 needed, I can backport the dashboard (remake it in older kibana).

We also can only do bugfixes to beats 8.17 and newer, if something arises in future SDHs

mykola-elastic: Do we need to support them?

packages/airflow/manifest.yml
mykola-elastic: also, future-wise. If logs is to be added to this integration at some point, this change would be needed anyway, as there is no use from UDP port/host for logs

mykola-elastic: How it looks now:
Screenshot 2025-10-08 at 08 10 25

packages/airflow/manifest.yml
mykola-elastic: btw, I also updated ECS to 8.17 because of https://github.com/elastic/integrations/issues/11952
Even though, ECS versioning matches semantic versioning of Elastic Stack, I didn't find anything about its compatibility with older stacks/kibanas

ishleenk17: If there is no specific requirement to upgrade the version, we should be ok with keeping it as it is. So that any future bugs reach as many stacks as possible. 

What do you mean by dashboard backport ? You can make dashboard in any stack version, just check that all panels work as it is in older stacks too.

packages/airflow/manifest.yml
ishleenk17: MAkes sense. Lets go ahead with this

packages/airflow/docs/README.md
ishleenk17: Are there new metrics added in 3.0 ?

packages/airflow/manifest.yml
mykola-elastic: I made and exported the dashboard from 8.17, installing it in 8.13 fails

packages/airflow/docs/README.md
mykola-elastic: yes, there are few metrics added in some minor versions too

ishleenk17: Do we need to support them here ? 
Since we are writing we support 3.x but are not supporting the newer metrics ?
This GA activity is also actually doing the service version upgrade activity as well :) 
Which is good if we can address the question if we want to include the nwer metrics as well as we are clamining support for 3.x here.

mykola-elastic: The was no 3.0 when the package was initially created. I remade the dashboard and also tested it with 3.0 too. So, the support means that nothing breaks.
So, is this a phrasing problem?

ishleenk17: In service version uograde when we say that we support a newer service version also, there are 2 aspects we check:

1. Nothing existing breaks
2. We support any new important metrics that were added between the 2 service versions . In this case 2.0, 3.0.
If we think there are newer metrics added and we are not taking care of that in this PR. We should keep support as 2.0 and add support for new metrics as a task under [service version upgrade process](https://github.com/elastic/obs-integration-team/issues/384). 


cc: @lalit-satapathy

packages/airflow/manifest.yml
mykola-elastic: Remade the dashboard in 8.13. Now it works on all `^8.13 || ^9.0` (verified)

packages/airflow/docs/README.md
mykola-elastic: if we add new metrics to the dashboard, then the panels using them will break if using older airflows

ishleenk17: @mykola-elastic : Since there are newer metrics added in 3.x and we are not taking care of them as part of GA. Lets keep compatibility for 2.x only.
We will move to 3.x as part of the service version upgrade, when we properly adhere to the process and handle the new metrics properly.

mykola-elastic: updated readme: https://github.com/elastic/integrations/pull/15287/commits/bcdeda8f654f9c70d8bb4cb5f298beedaab1b7c9

ok?

packages/airflow/_dev/build/docs/README.md
```diff
@@ -10,7 +10,11 @@ Use the Airflow integration to:
 - Create informative visualizations to track usage trends, measure key metrics, and derive actionable business insights.
 - Monitor your workflows' performance and status in real-time.
 
-## Data streams
+## Compatibility
+
+The Airflow package is tested with Airflow `2.4.0` and `3.0.6`. It will work with any `2.*` and `3.*` versions, but uses a smaller subset of metrics to be backwards-compatible with Apache Airflow 2.0.
```
ishleenk17: Lets just mention compatibility with 2.x. Lets remove mention of 3.x completely from here. 
Smaller subset line will be confusing for users. 

So basically use what was existing before in compatibility.

mykola-elastic: done, "The Airflow package is tested with Airflow `2.4.3`. It should work with any `2.*` version."

PR #14077, opened 2025-05-30
zscaler_zpa: standardize user name processing

chemamartinez:
## Proposed commit message

```
According to [ECS documentation](https://www.elastic.co/docs/reference/ecs/ecs-user-usage#ecs-user-identifiers) about user fields usage:

- When a system provides a composite value for the user name (e.g. DOMAINNAME\username), capture the domain name in user.domain and the user name (without the domain) in user.name.
- When a system uses an email address as the main identifier, populate both user.id and user.email with it.

This changes try to align to the rest of the integrations on how to handle emails in user name fields. Changes applied include the following:

- When the user name is an email, this field is dissected into `<user.name>@<user.domain>`.
- Also the `user.email` field is populated with the full email address.
- Append `user.name` and `user.email` to `related.user`.
```

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Related issues

- Relates https://github.com/elastic/enhancements/issues/24531

PR #15292, opened 2025-09-11
zscaler_zpa: unify treatment of user fields

chemamartinez:
## Proposed commit message

Same as done [here](https://github.com/elastic/integrations/pull/14077) for the `audit` data stream, it has been extended to the other data streams of the integration.

## Checklist

## Related issues

- Relates https://github.com/elastic/integrations/pull/14077

### DIFF
--- a/packages/zscaler_zpa/data_stream/app_connector_status/sample_event.json
+++ b/packages/zscaler_zpa/data_stream/app_connector_status/sample_event.json
@@ -1,11 +1,11 @@
 {
     "@timestamp": "2019-07-03T05:17:22.000Z",
     "agent": {
-        "ephemeral_id": "2b9e8c22-3a98-4cfd-bbb2-a20949d0801b",
-        "id": "6d6e7f83-d964-4fed-a813-9abde122c632",
-        "name": "elastic-agent-25332",
+        "ephemeral_id": "a668f81e-2848-4d79-95ef-1ecd5a6d332c",
+        "id": "85fbad05-ff47-41ab-b16e-b299c3b10b39",
+        "name": "elastic-agent-11095",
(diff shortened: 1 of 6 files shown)


efd6 (commented):
The sample events need to be regenerated. I'm not entirely sure where they come from, since running `elastic-package test system -g` gives different values for the failing fields.

For example:
```diff
diff --git a/packages/zscaler_zpa/data_stream/app_connector_status/sample_event.json b/packages/zscaler_zpa/data_stream/app_connector_status/sample_event.json
index da1e65abc8..952e36a90f 100644
--- a/packages/zscaler_zpa/data_stream/app_connector_status/sample_event.json
+++ b/packages/zscaler_zpa/data_stream/app_connector_status/sample_event.json
@@ -63,11 +63,8 @@
     "observer": {
         "geo": {
             "location": {
-                "coordinates": [
-                    -122.0000000204891,
-                    46.99999996460974
-                ],
-                "type": "Point"
+                "lat": 47,
+                "lon": -122
             }
         },
         "ip": [
```

chemamartinez:
> The sample events need to be regenerated. I'm not entirely sure where they come from, since running elastic-package test system -g gives different values for the failing fields.

I don't know either where these fields come from, I am just building the package with the latest official version of `elastic-package` (v0.114.0) and running the same command. I will look for a explanation here.

efd6:
You are running the stack at 9.1.0, but the manifest claims 8.13.0. The lower version is the one that will be run in CI, so you will need to use that.

9.1 has added a whole bunch of geometry handling that makes this happen. I think it should probably leave fields alone that have been explicitly set by a non-geo processor, but it doesn't seem to.

PR #15296, opened 2025-09-11
Add data retention for elastic agent status_change_logs data stream

MichelLosier:
## Proposed commit message

Resolves: https://github.com/elastic/ingest-dev/issues/5692

This adds a 90 day data retention limit for the `elastic_agent.status_change` data stream. 

Prior to this change, this data stream did not have a retention limit set. Kibana has a task that writes to this `elastic_agent.status_change` every minute only if there is a change in status. 

The primary use case for this data is for setting up alert rule monitoring of elastic agents. 90 days would allow users to analyze trends with status changes that may occur with an agent over the span of weeks or months. This also aligns with similar retention of our per 1m metric default policies for APM: https://www.elastic.co/docs/solutions/observability/apm/index-lifecycle-management.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes: https://github.com/elastic/ingest-dev/issues/5692

### DIFF
--- a/packages/elastic_agent/changelog.yml
+++ b/packages/elastic_agent/changelog.yml
@@ -1,5 +1,10 @@
 # newer versions go on top
-- version: "2.6.1"
+- version: "2.6.2"
+  changes:
+    - description: Add data retention limit for status_change_logs data stream
+      type: enhancement
+      link: "https://github.com/elastic/integrations/pull/15296"
+- version: "2.6.1" 
   changes:
     - description: Remove otelconsumer from Agent metrics dashboard
       type: bugfix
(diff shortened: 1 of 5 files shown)


jlind23:
@MichelLosier if this datastream only exist from 9.2 onwards, shouldn't we update the Kibana version constraint of this package?

nchaulet:
> @MichelLosier if this datastream only exist from 9.2 onwards, shouldn't we update the Kibana version constraint of this package?

The datastream exist as part of the elastic-agent package, so both the datastream definition and the template are contained in that version, and could be installed in kibana < 9.2

MichelLosier:
The datastream was already introduced in version 2.3.0 of this package, if we were to kibana version constraint this maybe a breaking change for the package.

PR #15307, opened 2025-09-12
feat!(beyondinsight_password_safe): standardize pipelines, disable dynamic mappings

andrewkroh:
## Proposed commit message

```
Disable dynamic mappings for beyondinsight_password_safe.* to prevent
unknown fields from receiving automatic mappings. This approach ensures
that API changes introducing new fields can be handled safely with
explicit field mappings and minimal risk of breaking existing users.

Standardize pipeline structures across all data streams with consistent
processing order:

- ecs.version
- terminate
- json decoding
- drop null/empty
- convert numberic IDs to strings
- convert camel-case to snake-case
- timestamps
- general ECS mappings
- append to ECS related.* fields

Sort fields.yml files alphabetically by name.

Remove unused package-level _dev/deploy. Each data stream has its own
mock test config.

BREAKING: The dsskey_rule_id field was renamed to dss_key_rule_id in
the managedsystem data stream to follow the snake-case naming pattern.

Related: #14985
```

## Checklist

## Related issues

- Relates #14985


packages/beyondinsight_password_safe/changelog.yml
```diff
@@ -1,3 +1,14 @@
+- version: "0.11.0"
+  changes:
+    - description: Renamed dsskey_rule_id field to dss_key_rule_id in managedsystem data stream to follow snake-case naming convention.
```
efd6:
```suggestion
    - description: Renamed `dsskey_rule_id` field to `dss_key_rule_id` in managedsystem data stream to follow snake-case naming convention.
```
Also, the movement of `beyondinsight_password_safe.asset.operating_system` field values from `os.name` to `host.os.full` is breaking for people depending on that value in that location.

packages/beyondinsight_password_safe/data_stream/useraudit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1,189 +1,161 @@
...
       value: "{{{user.email}}}"
-      if: ctx.user?.email != null
       allow_duplicates: false
-      ignore_failure: true
-
+  - append:
+      if: ctx.beyondinsight_password_safe.useraudit?.user_id == "-1"
+      field: related.user
+      value: "{{{beyondinsight_password_safe.useraudit.user_id}}}"
+      allow_duplicates: false
```
efd6: Is this intentional? If it is, `value: "-1"`, but I suspect it's not; most likely `s/==/!=/`, or obtain it from `user.id` if it's present.

Issue #15036, opened 2025-08-25
[Elastic Agent]: Endpoint process memory is reported incorrectly on MacOS


ycombinator:
### Integration Name

Elastic Agent [elastic_agent]

### Dataset Name

endpoint_security_metrics: process.memory

### Integration Version

2.5.0

### Agent Version

9.0.3

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.0.3

### OS Version and Architecture

MacOS Sonoma 14.5 / Apple M2 Pro

### What did you do?

No special configuration.  Just added the Elastic Defend integration to the Agent's policy, waited a couple of minutes, then viewed the Agent Metrics dashboard,

### What did you see?

Overall memory for Elastic Agent shot up from a few hundred MB to a few hundred GB after the Elastic Defend integration was added to the Agent's policy:

<img width="663" height="288" alt="Image" src="https://github.com/user-attachments/assets/1fe6d270-163f-4be4-9def-c6301ca5d4a2" />

Likewise, the component-wise breakdown of Elastic Agent memory also shot up, with the `endpoint_security` component reporting memory around 390GB while each of the other components were < 200 MB each.

<img width="722" height="332" alt="Image" src="https://github.com/user-attachments/assets/f6476cb7-96d5-49f0-849c-efde3f0eab7a" />

### What did you expect to see?

That the memory reported by the `endpoint_security` component was also on the order of a couple hundred MB, at most.

### Anything else?

Checking the raw data in the `metrics-*` data stream for the `endpoint_security` component, it appears that each document contains a high value of `system.process.memory.size` — around 390 GB. 

```json
POST metrics-*/_search?filter_path=hits.hits._source
{
  "_source": [ "system.process.memory" ],
  "sort": { "@timestamp": "desc" },
  "query": {
    "bool": {
      "filter": [
        { "term": { "elastic_agent.process": "endpoint_security" } },
        { "exists": { "field": "system.process.memory.size" } }
      ]
    }
  }
}
```

```json
{
  "hits": {
    "hits": [
      {
        "_source": {
          "system": {
            "process": {
              "memory": {
                "rss": {
                  "bytes": 274956288
                },
                "size": 421678710784
              }
            }
          }
        }
      },
      {
        "_source": {
          "system": {
            "process": {
              "memory": {
                "rss": {
                  "bytes": 228245504
                },
                "size": 421678710784
              }
            }
          }
        }
      },
      {
        "_source": {
          "system": {
            "process": {
[...]

PR #15315, opened 2025-09-15
[elastic_agent] Prefer RSS for memory usage charts

AndersonQ:
## Proposed commit message

```
[elastic_agent] prefer RSS for memory usage charts

The overall memory chart and the by component memory chart now prefer to use RSS memory (`system.process.memory.rss.bytes`) instead of virtual memory (`system.process.memory.size`).

When `system.process.memory.rss.bytes` is available for a component, it's used, if not available it falls back to `system.process.memory.size`. Currently only Elastic Defend (endpoint) has RSS memory available.
```

## Checklist

 - run a cloud stack
 - clone this repo
 - set the following env vars for elastic-package
```shell
export ELASTIC_PACKAGE_ELASTICSEARCH_HOST=https://elasticsearch.elastic-cloud.com
export ELASTIC_PACKAGE_ELASTICSEARCH_KIBANA_HOST=https://kibana.elastic-cloud.com
export ELASTIC_PACKAGE_KIBANA_HOST=https://kibana.elastic-cloud.com
export ELASTIC_PACKAGE_KIBANA_USERNAME=elastic
export ELASTIC_PACKAGE_KIBANA_PASSWORD=changeme
export ELASTIC_PACKAGE_ELASTICSEARCH_USERNAME=elastic
export ELASTIC_PACKAGE_ELASTICSEARCH_PASSWORD=changeme
```
 - install an elastic agent with Elastic Defend integration 
 - check the _[Elastic Agent] Agent metrics_ dashboard. It should be as described in the issue
 - install the new package version
```shell
elastic-package install -v
```
 - check the _[Elastic Agent] Agent metrics_ dashboard again. Now the memory reported for endpoint should be within reason

## Related issues

- Closes https://github.com/elastic/integrations/issues/15036

#### updated dashboard
Screenshot from 2025-09-15 09-31-16

#### comparison Old vs New
Screenshot from 2025-09-15 08-52-58
Screenshot from 2025-09-15 08-53-18

#### changes made
Screenshot from 2025-09-15 09-53-28

Screenshot from 2025-09-15 09-54-22

### DIFF
--- a/packages/elastic_agent/kibana/dashboard/elastic_agent-f47f18cc-9c7d-4278-b2ea-a6dee816d395.json
+++ b/packages/elastic_agent/kibana/dashboard/elastic_agent-f47f18cc-9c7d-4278-b2ea-a6dee816d395.json
@@ -7881,6 +7881,6 @@
         }
     ],
     "type": "dashboard",
-    "typeMigrationVersion": "10.3.0",
+    "typeMigrationVersion": "8.9.0",
     "updated_by": "u_849892341_cloud"
 }


belimawr (approved):

cmacknz:
Does the the memory calculation in the agent details view in Fleet also need to be updated to account for this change?

Reading the description and dashboard files it seems like this only applies to endpoint security which is what we want, but mainly because we don't collect system/metrics process metrics for anything else.

AndersonQ:
> Does the the memory calculation in the agent details view in Fleet also need to be updated to account for this change?

No, it uses another metric, I don't remember on top of my head which one, but it does not match the dashboard, see below. I thought it came from the agent doc, but it's not there.

Screenshot from 2025-09-18 10-22-00

it's close to the agent's virtual memory, but don't quite match:

Screenshot from 2025-09-18 10-40-38


> Reading the description and dashboard files it seems like this only applies to endpoint security which is what we want, but mainly because we don't collect system/metrics process metrics for anything else.

exactly, if there is a rss metric available, the dashboard will use it. Right now only endppoint has it, thus it only apply to endpoint.

AndersonQ:
@cmacknz is it good to merge?

cmacknz:
> No, it uses another metric, I don't remember on top of my head which one, but it does not match the dashboard, see below. I thought it came from the agent doc, but it's not there.

This dashboard needs to match the way Fleet calculates the memory usage, either this dashboard is correct now and Fleet is wrong and needs a separate update, or this dashboard should be adjusted to match what Fleet does because that is correct.

Without addressing this we will immediately have another bug that two parts of the product do not agree on how to calculate memory usage.

I would like you investigate what Fleet is doing before merging this.

AndersonQ:
> > No, it uses another metric, I don't remember on top of my head which one, but it does not match the dashboard, see below. I thought it came from the agent doc, but it's not there.
> 
> This dashboard needs to match the way Fleet calculates the memory usage, either this dashboard is correct now and Fleet is wrong and needs a separate update, or this dashboard should be adjusted to match what Fleet does because that is correct.
> 
> Without addressing this we will immediately have another bug that two parts of the product do not agree on how to calculate memory usage.
> 
> I would like you investigate what Fleet is doing before merging this.

ok, so I'll open another issue as to track it better as it's a different thing.

cmacknz:
Sure, as long as we actually do that investigation we can merge this first, it will make the MacOS graphs not be totally incorrect.

AndersonQ:
> Sure, as long as we actually do that investigation we can merge this first, it will make the MacOS graphs not be totally incorrect.

here is the issue: https://github.com/elastic/integrations/issues/15434

I agree, merging it already will bring value regardless if it's consistent with fleetUI or not.

just to be sure. I can merge it right?

cmacknz:
Yes, LGTM

PR #15334, opened 2025-09-16
[kafka] add system tests to consumer and producer data streams

stefans-elastic:
## Proposed commit message

See title

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/kafka/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.23.1"
+  changes:
+    - description: Add system tests for consumer and producer data streams.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/15334
```
mykola-elastic, line 6: why enhancement with patch version update?

stefans-elastic: because there is no changes to the functionality

agithomas:
Can you regenerate the sample events for the producer dataset and update it as part of this PR?

packages/kafka/data_stream/consumer/_dev/deploy/docker/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM apache/kafka:4.0.0
+
```
Linu-Elias: @stefans-elastic, The changes look good, but I was just wondering, what if we add the setup script to the parent `_dev` directory of this integration? That way, we can avoid repeating the script in each data stream, especially since there are quite a few data streams in Kafka that use JMX and share the same setup. Only the data stream–specific details might differ. WDYT?

stefans-elastic:
> Can you regenerate the sample events for the producer dataset and update it as part of this PR?

done

packages/kafka/data_stream/consumer/_dev/deploy/docker/Dockerfile
stefans-elastic: hmmm, let me try this

stefans-elastic: it worked (at least locally), thanks for the suggestion!
Lets wait for the CI to complete to see if it works in CI tests

stefans-elastic: I've actually reverted the change for `raft` data stream because when later on when we'll introduce `variants` for the tests we'll need a different set of variants for `raft` stream (because it would work only if kafka is run in KRaft mode)

agithomas (approved):
LGTM!

PR #15346, opened 2025-09-16
[AWS] Introduce initial alert rule templates

ghost:
## Overview

This PR introduces the first set of alert rule templates for key AWS data streams. For each stream, we selected the two most critical metrics to monitor.

### `ec2_metrics`

- **High CPU Utilization**
  - This alarm is used to detect high CPU utilization.
- **Status Check failed**
  - This alarm is used to detect the underlying problems with instances, including both system status check failures and instance status check failures.

### `lambda`

- **High Number of Throttles**
  - The alarm helps detect a high number of throttled invocation requests for a Lambda function.
- **High Number of Errors**
  - The alarm helps detect high error counts in function invocations.

### `sqs`

- **Oldest Message Name is Too High**
  - This alarm is used to detect whether the age of the oldest message in the QueueName queue is too high. **Threshold depends on situation.**
- **High Number of Visible Messages**
  - This alarm is used to detect whether the message count of the active queue is too high and consumers are slow to process the messages or there are not enough consumers to process them. **Threshold depends on situation.**

### `sns`

- **Any Message Delivery Fails**
  - This alarm helps you proactively find issues with the delivery of notifications and take appropriate actions to address them. **Threshold depends on situation.**
- **Number of Notifications Filtered Out - Invalid Attributes**
  - The alarm is used to detect if the published messages are not valid or if inappropriate filters have been applied to a subscriber.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes elastic/obs-integration-team/issues/536


packages/aws/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.17.0"
+  changes:
+    - description: Add ec2_metrics, lambda, sqs and sns alert rule templates.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/999
```
efd6:
```suggestion
      link: https://github.com/elastic/integrations/pull/15346
```

ishleenk17:
@gpop63 : The template will be usable from 9.2 onwards .
If yes, lets mark the PR as DON'T MERGE.

Can you please share a screenshot of how a particular alert looks like. Also, are we not adding any information about alert support in the README's ?

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
```diff
@@ -0,0 +1,37 @@
+{
+  "id": "b6513de4-6c36-499a-8f0a-98431cd4dbee",
+  "type": "alerting_rule_template",
+  "attributes": {
+    "name": "EC2 High CPU Utilization",
+    "tags": [],
```
muthu-mps: Can we add tags?

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
```diff
@@ -0,0 +1,37 @@
+{
+  "id": "b6513de4-6c36-499a-8f0a-98431cd4dbee",
```
muthu-mps: Should the id match with the file name of the rule_template?
`Error: defines non-matching ID`

ghost:
@ishleenk17 right now the support is not fully there we only see them under assets and in saved objects

image

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
```diff
@@ -0,0 +1,37 @@
...
+      "thresholdComparator": ">",
+      "size": 100,
+      "esqlQuery": {
+        "esql": "FROM metrics-aws.ec2_metrics-default\n| STATS cpuutilization=avg(aws.ec2.metrics.CPUUtilization.avg) by cloud.account.id, cloud.region, aws.dimensions.InstanceId\n| WHERE cpuutilization >= 80"
+      },
+      "aggType": "count",
+      "groupBy": "all",
+      "termSize": 5,
+      "sourceFields": [],
+      "timeField": "event.ingested",
```
muthu-mps: Can the time field be @timestamp? Is there a reason for choosing event.ingested instead of @timestamp?

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
```diff
@@ -0,0 +1,37 @@
...
+      "threshold": [
+        0
+      ],
+      "thresholdComparator": ">",
+      "size": 100,
+      "esqlQuery": {
+        "esql": "FROM metrics-aws.ec2_metrics-default\n| STATS cpuutilization=avg(aws.ec2.metrics.CPUUtilization.avg) by cloud.account.id, cloud.region, aws.dimensions.InstanceId\n| WHERE cpuutilization >= 80"
+      },
+      "aggType": "count",
+      "groupBy": "all",
```
muthu-mps: Is this groupBy not applicable while using ESQL query?

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
```diff
@@ -0,0 +1,37 @@
...
+      "searchType": "esqlQuery",
+      "timeWindowSize": 5,
+      "timeWindowUnit": "m",
+      "threshold": [
+        0
+      ],
+      "thresholdComparator": ">",
+      "size": 100,
+      "esqlQuery": {
+        "esql": "FROM metrics-aws.ec2_metrics-default\n| STATS cpuutilization=avg(aws.ec2.metrics.CPUUtilization.avg) by cloud.account.id, cloud.region, aws.dimensions.InstanceId\n| WHERE cpuutilization >= 80"
```
muthu-mps: Applying dataset filter help fetch only the specific data for the alerting metrics. WDYT?

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
```diff
@@ -0,0 +1,37 @@
...
+    "tags": [],
+    "ruleTypeId": ".es-query",
+    "schedule": {
+      "interval": "1m"
+    },
+    "params": {
+      "searchType": "esqlQuery",
+      "timeWindowSize": 5,
+      "timeWindowUnit": "m",
+      "threshold": [
```
muthu-mps: Similar to `groupby`. Check whether the threshold value is applied directly from ESQL query and not from here.

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
ghost: I tried using the @timestamp field but it wasn't generating alerts. For some AWS data streams @timestamp is when the actual metric happened in AWS.

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
ghost: The threshold is set in the esql query, this is a different property of the alert.

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
ghost: How do we do that? also this esql query targets documents from a specific data stream/index (metrics-aws.ec2_metrics-default)

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
ghost: The group by of actual data happens in the esql query itself, this has to be a property of the alert.

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
daniela-elastic: What tags were you thinking of, Muthu?

packages/aws/kibana/alerting_rule_template/aws-ec2-high-cpu-utilization.json
daniela-elastic, line 1: Where do we declare which service (entity) this alert template applies to? Something like resource : aws.ec2

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
muthu-mps: The tags can have the service name and the Alert metrics name. Similar to what I have added [here](https://github.com/muthu-mps/integrations/blob/d7ad961aeb0969ca0e9733bd13b33c3ef35e94ed/packages/azure_ai_foundry/kibana/alerting_rule_template/latency_spike.json#L6) in Azure AI Foundry.
e.g., [`AWS EC2`, `AWS EC2 CPU Utilization`].

packages/aws/kibana/alerting_rule_template/aws-ec2-high-cpu-utilization.json
ghost: I have included the service name in the name of the alert rule template. I suppose Kibana should allow us to filter by tags or by partial matches on the title of the alert rule template.

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
```diff
@@ -0,0 +1,37 @@
...
+        0
+      ],
+      "thresholdComparator": ">",
+      "size": 100,
+      "esqlQuery": {
+        "esql": "FROM metrics-aws.ec2_metrics-default\n| STATS cpuutilization=avg(aws.ec2.metrics.CPUUtilization.avg) by cloud.account.id, cloud.region, aws.dimensions.InstanceId\n| WHERE cpuutilization >= 80"
```
muthu-mps: The field `aws.ec2.metrics.CPUUtilization.avg` is renamed [here](https://github.com/elastic/integrations/blob/ce98afae516c2d077551708248daf3d49ffeeaa9/packages/aws/data_stream/ec2_metrics/elasticsearch/ingest_pipeline/default.yml#L20). Do you think this field should be changed to `host.cpu.usage` field?

ghost: Both metrics are present but switching to the ECS field seems like the better option. Switched to `host.cpu.usage` in [d28dd85](https://github.com/elastic/integrations/pull/15346/commits/d28dd851c2b82e2634fe8a88b80e4f1eada61dcc). I had to multiply it by 100 to use percentages.

MichelLosier (approved):
Configuration looks good to me!

packages/aws/kibana/alerting_rule_template/ec2-high-cpu-utilization.json
muthu-mps: We can ignore this as we directly target against specific datastream.

muthu-mps (approved):

packages/aws/kibana/alerting_rule_template/aws-sqs-messages-visible.json
```diff
@@ -0,0 +1,28 @@
+{
+  "id": "sqs-messages-visible",
+  "type": "alerting_rule_template",
+  "attributes": {
+    "name": "[AWS SQS] Messages Visible",
+    "tags": ["AWS SQS"],
+    "ruleTypeId": ".es-query",
+    "schedule": {
+      "interval": "1m"
+    },
```
agithomas, line 10: This is applicable for all the configurations. 

Should we keep this so frequently? I suggest, this be equal to the default period value for metrics ingestion. Following so, it helps to avoid any no-data found alert (when user decides to extend the configuration)

ghost: Should we set `timeWindowSize` to match the integration period? That way, for example, every 5 minutes we’d check for alerts in documents from the past 5 minutes.

agithomas: I think, thats a resonable thing to do. The impact I assume here will be that instead of an alert being notified at the period + 1m interval, the alert will be notified at 2 x period internal. Here period is 5m for most AWS servies. 

@tommyers-elastic , what would be your recommendation?

packages/aws/manifest.yml
```diff
@@ -15,7 +15,7 @@ conditions:
   elastic:
     subscription: basic
   kibana:
-    version: "^8.19.0 || ^9.1.0"
+    version: "^9.2.1"
```
muthu-mps: @elastic/security-service-integrations team, This feature is supported starting from **9.2.1 release version**. The minimum stack version gets upgraded to 9.2.1. Since AWS integrations involve co-ownership, Could you confirm if the stack version upgrade is fine with the integrations managed by security team?

packages/aws/kibana/alerting_rule_template/aws-sqs-messages-visible.json
tommyers-elastic: i don't think we have any way to couple configs in agent policy templates with these rule configurations, so whatever we choose will have to be always added by hand. 

my only thinking here is that it doesn't make sense to run a rule more frequently than the integration collection period. matching the rule frequency with the collection period seems sensible to me.

tommyers-elastic: it's a shame there's no way to put hints in the form such that we could have something that shows up and says "should match the integration collection period" or something. if we think it's worthwhile we could suggest this as a feature.

agithomas:
Sharing a suggestion here:

Could we follows a more structured comment style - helping user to identify the purpose, default value, condition, group-by information with easily and make suggestions easier? When followed, these descriptions could be combined with other platform capabilities including AI assistant (in future, if & when needed)

Examples:

```
1. 
// Alert triggers when the maximum number of visible messages in an SQS queue
// reaches or exceeds the defined threshold (default: 1000) within the lookback window.
//
// The alert is grouped by cloud account, region, and queue name to identify
// which specific queue is experiencing backlog.
//
// To adjust sensitivity, change the `msgsvisible` threshold value in the WHERE clause.
```

```
2.
// Alert triggers when SNS notifications are being filtered out by subscription filter policies.
// A non-zero value usually indicates a mismatch between published messages and subscriber
// filter rules, which may result in messages not reaching intended consumers.
//
// The alert is grouped by cloud account, region, and topic name to identify the affected topic.
//
// To adjust sensitivity, change the `notificationsfilteredout` threshold in the WHERE clause.

```

packages/aws/manifest.yml
chrisberkhout: Ad discussed elsewhere, I think this version constraint should be left unchanged. The alerting rule template files will be built into the package and they will be installed and used on stack versions that support them

muthu-mps:
@agithomas - Apart from version dependency, Can you help with the review and approval if everything looks good?

agithomas (approved):
> @agithomas - Apart from version dependency, Can you help with the review and approval if everything looks good?

LGTM from the alerts configuration. Looking forward to have a common agreement on the version dependency before proceeding.

PR #15353, opened 2025-09-16
[CI] Move test dependencies to tools.go

mrodm:
## Proposed commit message

Ensure that test dependencies used in CI will not fail due to a mismatch of required Golang versions.

Related failure seen in backport branches:
```
Setting up the Go environment...
GVM v0.5.2 (platform linux arch amd64
go version go1.23.4 linux/amd64
/root/.gvm/versions/go1.23.4.linux.amd64/bin/go
go: downloading github.com/magefile/mage v1.15.0
go: downloading github.com/jstemmer/go-junit-report v1.0.0
go: downloading gotest.tools/gotestsum v1.13.0
go: downloading gotest.tools v2.2.0+incompatible
go: gotest.tools/gotestsum@latest: gotest.tools/gotestsum@v1.13.0 requires go >= 1.24.0 (running go 1.23.4; GOTOOLCHAIN=local)
🚨 Error: The command exited with status 1
```

Examples of builds failing:
- https://buildkite.com/elastic/integrations/builds/31461#01995328-df87-4570-93eb-e7da7bdbe102
- https://buildkite.com/elastic/integrations/builds/31463#01995329-404c-4755-87db-3dff541dff29

## Author's Checklist

- [x] Run `mage -v check` locally.

### DIFF
--- a/.buildkite/scripts/common.sh
+++ b/.buildkite/scripts/common.sh
@@ -125,7 +125,7 @@ with_mage() {
     create_bin_folder
     with_go
 
-    go install github.com/magefile/mage@latest
+    go install "github.com/magefile/mage@${SETUP_MAGE_VERSION:-"latest"}"
 
     mage --version
 }
--- a/.buildkite/scripts/common.sh
+++ b/.buildkite/scripts/common.sh
@@ -125,14 +125,8 @@ with_mage() {
     create_bin_folder
     with_go
 
-    local install_packages=(
-            "github.com/magefile/mage"
-            "github.com/jstemmer/go-junit-report"
-            "gotest.tools/gotestsum"
-    )
-    for pkg in "${install_packages[@]}"; do
-        go install "${pkg}@latest"
-    done
+    go install github.com/magefile/mage@latest
+
     mage --version
 }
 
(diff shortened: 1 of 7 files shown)


tools.go
andrewkroh, line 1: There's a new option in Go for manage tool dependencies. Of course you can still use the tools.go method, but just thought I'd mention it:

https://go.dev/doc/modules/managing-dependencies#tools

mrodm: Thanks for the heads up about this new option @andrewkroh !! 

One question about this new option, would dependabot update those packages being defined as tools in go.mod ?

mrodm: I think at least in this PR I would continue with the tools.go method. This new option (tool directive) could be checked in a follow-up PR.

andrewkroh: Renovate does, but I think dependabot may need a change before it supports updating them automatically.

- https://github.com/dependabot/dependabot-core/issues/12050
- https://github.com/renovatebot/renovate/pull/35078

mrodm: Thanks for checking this out 
As in this repository it is used dependabot, I'll keep with the tools.go file then.

.buildkite/scripts/common.sh
```diff
@@ -125,14 +125,9 @@ with_mage() {
     create_bin_folder
     with_go
 
-    local install_packages=(
-            "github.com/magefile/mage"
-            "github.com/jstemmer/go-junit-report"
-            "gotest.tools/gotestsum"
```
mrodm, line 131: AFAIK those two dependencies were not used.

I've removed `go-junit-report` dependency, and I've set `gotestsum` to run the tests in the `magefile.go` adding the parameter to create the jUnit file.

PR #15359, opened 2025-09-16
sailpoint_identity_sc: pass events as serialised json strings

efd6:
## Proposed commit message
```
sailpoint_identity_sc: pass events as serialised json strings

This makes it possible to users who are passing events through logstash
to use the agent.

Also improve ingest pipeline error handling.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/sailpoint_identity_sc/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1,8 +1,21 @@
 ---
 description: Pipeline for processing sailpoint identity security cloud events
 processors:
+  - set:
+      field: ecs.version
+      value: "8.11.0"
+
```
andrewkroh, line 11: Should we add in the `terminate` processor now too while doing this clean up?

packages/sailpoint_identity_sc/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1,8 +1,21 @@
...
+  - remove:
+      field: message
+      ignore_missing: true
+      if: ctx.event?.original != null
   - rename:
       field: message
+      target_field: event.original
+      ignore_missing: true
+      if: ctx.event?.original == null
```
andrewkroh, line 20: Would this logic be unnecessary if the CEL program were made to populate only `event.original`?

packages/sailpoint_identity_sc/data_stream/events/elasticsearch/ingest_pipeline/default.yml
efd6: I thought about that and the CEL does not gracefully handle non-200 status codes. I'll add that too.

packages/sailpoint_identity_sc/data_stream/events/elasticsearch/ingest_pipeline/default.yml
efd6: Considering the context of the bug that this is fixing, that could be a breaking change (it's not for the op of the issue, but it could be for someone else with a similar data flow).

andrewkroh (approved):

PR #15360, opened 2025-09-16
jamf_protect: respect preserve_original_event in alerts

efd6:
## Proposed commit message

See title.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/jamf_protect/data_stream/alerts/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -4,10 +4,21 @@ processors:
...
+        field: event.original
         target_field: jamf_protect.alerts
-        if: ctx.message != null
+        if: ctx.event?.original != null
+    - script:
+        description: Retain the JSON in event.original if it came in that way and we have preserve_original_event.
+        if: ctx.event?.original == null && ctx.json != null && ctx.tags != null && ctx.tags.contains('preserve_original_event')
+        source: |-
+          ctx.event = ctx.event ?: [:];
+          ctx.event.original = Json.dump(ctx.json);
```
andrewkroh: What's the use case for this? IIUC, the input sets `event.original` when `preserve_original_event` is enabled, so this should never be needed.

efd6: ¯\\\_(ツ)\_/¯

I've reached out to the original author to find out since it wasn't clear from the PR description that introduced the handling of the `json` field. My suspicion is that it comes into play with either the http_endpoint or awss3 inputs somehow.

I don't plan to merge this until I have the answer from them.

andrewkroh: I had not considered that there is an AWS-S3 input in the data stream too, but even so, that input should be populating only `message` (with stringified JSON), and therefore the `.json` field would not exist in that case either.

efd6: When I hear back I'll sort out a correction.

efd6: I've figured it out. This is happening because http_endpoint is putting the data in the `json` field (the default). It could be put somewhere else, but that would not fix the issue that this is addressing since we would still need to marshal the JSON (http_endpoint sends JSON, not a message, and there is no support for JSON marshaling in the flavour of CEL that http_endpoint provides, so it cannot be done there). I agree, this is not ideal, but I think it's the best we can do.

andrewkroh: The [preserve_original_event http_endpoint input option](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-http_endpoint.html#_preserve_original_event) should take care of the marshaling when the same name integration variable is enabled.

https://github.com/elastic/integrations/blob/9431cb5106c89ebb19e65646f54ecbf92dba673d/packages/jamf_protect/data_stream/alerts/agent/stream/http_endpoint.yml.hbs#L10-L11

The processor has some use in case preserve_original_event was not enabled and the pipeline fails, because then the `Json.dump` version of event.original will at least be in the event.

efd6: Looking at that, I think that should always be set to true, rather than conditional on the UI setting. Leaving it with the possibility of being false allows us to be in the situation that there is no `event.original` to be retained in the case of an error.

andrewkroh: SGTM. If we do that then I think this processor can be removed from the pipeline.

efd6: I've left it in because we will undoubtedly have people who are sending this via logstash at some point with only the `event.original`.

andrewkroh (approved):

PR #15378, opened 2025-09-17
[citrix_adc] Parsing issue for HTTPREQUEST & UDPFLOWSTAT

robester0403:
## Checklist

`run elastic-package stack up and elastic-package test pipeline`


packages/citrix_adc/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.17.2"
+  changes:
+    - description: "Fix the Grok processing for HTTPREQUEST and UDPFLOWSTAT in sslvpn_and_aaatm_feature pipeline with optional patterns."
```
andrewkroh: I don't think `Grok` needs to be given proper noun status. `s/Grok/grok/`

robester0403:
> When adding the new test samples, how did you choose between adding a .json file and a .log file?

Good call! I corrected both logs should be in a .log file and after re-reading the raw data the logs were observed by the firewall. I moved the logs to test-citrix-waf-native-14-1.log after seeing that's where these logs are from.

packages/citrix_adc/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.17.2"
+  changes:
+    - description: "Fix Grok processing for HTTPREQUEST and UDPFLOWSTAT in sslvpn_and_aaatm_feature pipeline with optional patterns."
```
andrewkroh:
```suggestion
    - description: "Fix grok processing for HTTPREQUEST and UDPFLOWSTAT in sslvpn_and_aaatm_feature pipeline with optional patterns."
```

robester0403:
> The additions here look fine.
> 
> This isn't new, but I noticed it while reviewing the generated files - there is a lot of duplication in these events. In particular between
> 
> * `event.original` (keyword)
> * `citrix.detail` (keyword)
> * `citrix.extended.message` (part of a `flattened`)
> * `citrix_adc.log.message` (keyword).
> 
> `citrix.detail` is a duplicate of event.original. This could be turned into an `alias` to retain the behavior when `preserve_original_event` is enabled, but I would probably omit the alias given other parts will be a breaking change.
> 
> `citrix.extended.message` moves to `message` so that the event has the core ECS [message](https://www.elastic.co/docs/reference/ecs/ecs-base#field-message) field which is `match_only_text` for search.
> 
> `citrix_adc.log.message` is removed. The same info will be in `message`.
> 
> This would reduce the event size, make search work better, and better follow ECS. We should probably have an issue to track this cleanup for a future breaking change.

I will make an issue for this in our backlog

PR #15392, opened 2025-09-18
m365_defender: ensure $skip parameter is correctly formatted

efd6:
## Proposed commit message
```
m365_defender: ensure $skip parameter is correctly formatted

The values used to populate this parameter can be greater than the
cutover to e-notation for doubles, so convert to int before converting
to string.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/m365_defender/changelog.yml
+++ b/packages/m365_defender/changelog.yml
@@ -4,6 +4,9 @@
     - description: Ensure large `$skip` API parameter values are correctly formatted in `vulnerability` data stream.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/15392
+    - description: Supress expected empty template health updates.
+      type: bugfix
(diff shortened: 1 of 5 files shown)


efd6:
The failures here are both unrelated and cosmetic, and unfixable until 8.19.4 is released since it depends on elastic/beats#46332. However, this change is fixing a real issue. So it would be good if on approval this could be merged despite the failures in alert and incident (I have added changes that will pick up the fix in those data streams when the stack is updated).

kcreddy:
> hope for the future

Curious how the CI passed after adding `do_not_log_failure: true` when the fix was added to `8.19.4` but integration's min stack version is [`8.19.2`](https://github.com/elastic/integrations/blob/main/packages/m365_defender/manifest.yml#L14).

PR #15396, opened 2025-09-18
docs(sentinel_one): Changes H1 for increasing SEO

szabosteve:
## Overview

This PR updates the page title to SentinelOne integration. Since this page is a child of the main SentinelOne page in the generated docs, using the same title for both negatively impacts SEO.

### DIFF
--- a/packages/sentinel_one/docs/README.md
+++ b/packages/sentinel_one/docs/README.md
@@ -1,4 +1,4 @@
-# SentinelOne
+# SentinelOne integration
 
 The [SentinelOne](https://www.sentinelone.com/) integration collects and parses data from SentinelOne REST APIs. This integration also offers the capability to perform response actions on SentinelOne hosts directly through the Elastic Security interface (introduced with v8.12.0). Additional configuration is required; for detailed guidance, refer to [documentation](https://www.elastic.co/guide/en/security/current/response-actions-config.html).
 


colleenmcginnis (commented):
@szabosteve  you'll need to follow this process for any docs updates:

1. Manually update the content in the Markdown file in the `packages/{package_name}/_dev/build/docs` directory (for example, [`packages/sentinel_one/_dev/build/docs/monitor.md`](https://github.com/elastic/integrations/blob/main/packages/sentinel_one/_dev/build/docs/monitor.md)). 
2. Add a new changelog entry to the `packages/{package_name}/changelog.yml` file (for example, [`packages/sentinel_one/changelog.yml`](https://github.com/elastic/integrations/blob/main/packages/sentinel_one/changelog.yml)). Docs-only updates should add a patch version, the `type` should be `enhancement`, and the `link` should be a link to this PR. 
3. Bump the `version` in the `packages/{package_name}/manifest.yml` file (for example, [`packages/sentinel_one/manifest.yml`](https://github.com/elastic/integrations/blob/main/packages/sentinel_one/manifest.yml)) to the new version you added to the changelog.
4. Install `elastic-package` ([docs](https://github.com/elastic/elastic-package?tab=readme-ov-file#getting-started)).
5. Change into the package directory `cd packages/{package_name}` (for example, `cd packages/sentinel_one`).
6. Run `elastic-package build`.
7. Commit both the manually and generated files.

chrisberkhout (commented):
> Since this page is a child of the main SentinelOne page in the generated docs

[This page](https://www.elastic.co/docs/reference/integrations/sentinelone-intro), right?

Rather than changing the title of the integration's README in a way that doesn't match the integration name and the way we do other integration names, I think it might be better to change the title of that parent page like this:

Screenshot From 2025-11-10 16-19-42 (Copy)

What do you think?

Issue #14980, opened 2025-08-19
[ssi_some]Add support to affected integrations - Avoid updating Fleet health status to degraded when HTTPJSON template value evaluation is empty


narph:
Support was added to avoid updating Fleet health status to degraded when HTTPJSON template value evaluation is empty in **https://github.com/elastic/beats/issues/45664**. 
To enable this feature some changes need to be made to the following integrations:
- 1password
- google_workspace
- google_scc
- rapid7_insightvm

navnit-elastic:
The fleet health status degraded issue has been resolved for the mentioned integrations in https://github.com/elastic/integrations/pull/15415. hence, closing this issue.

PR #15415, opened 2025-09-22
ssi_some: prevent updating fleet health status to degraded

navnit-elastic:
## Proposed commit message

Prevent updating fleet health status to degraded for following packages
when HTTPJSON template value evaluation is empty.
* 1password
* google_scc
* m365_defender
* rapid7_insightvm

This change bumps minimum Kibana version to 8.19.4 and 9.1.4 and
adds do_not_log_filure: true in set processors to avoid updating
fleet health status to degraded.
This change also fixes the 1password package's system tests.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Relates #14330 
- Relates #14954 
- Relates #15105 
- For #14980

### DIFF
--- a/packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs
+++ b/packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs
@@ -45,7 +45,7 @@ response.split:
 response.pagination:
   - set:
       target: body.cursor
-      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[else]][[/*This is an empty value to stop pagination.*/]][[end]]'
+      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[end]]'
       fail_on_template_error: true
       do_not_log_failure: true
   - delete:
--- a/packages/1password/data_stream/item_usages/agent/stream/httpjson.yml.hbs
+++ b/packages/1password/data_stream/item_usages/agent/stream/httpjson.yml.hbs
@@ -46,7 +46,7 @@ response.split:
 response.pagination:
   - set:
       target: body.cursor
-      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[else]][[/*This is an empty value to stop pagination.*/]][[end]]'
+      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[end]]'
       fail_on_template_error: true
       do_not_log_failure: true
   - delete:
--- a/packages/1password/data_stream/signin_attempts/agent/stream/httpjson.yml.hbs
+++ b/packages/1password/data_stream/signin_attempts/agent/stream/httpjson.yml.hbs
@@ -46,7 +46,7 @@ response.split:
 response.pagination:
   - set:
       target: body.cursor
-      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[else]][[/*This is an empty value to stop pagination.*/]][[end]]'
+      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[end]]'
       fail_on_template_error: true
       do_not_log_failure: true
   - delete:


packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs
```diff
@@ -27,23 +27,33 @@ request.transforms:
       value: 'Bearer {{token}}'
   - set:
       target: body.cursor
-      value: '[[if not (eq (len .cursor) 0)]][[.cursor.last_cursor]][[end]]'
+      value: >-
```
navnit-elastic: @efd6, If the terminate helper evaluates to an empty value, then we don't need to use terminate here, right? Also, I'm not sure about using "do_not_log_failure: true" without "fail_on_template_error: true". Does this scenario prevent changing health status to degraded?

efd6: > If the terminate helper evaluates to an empty value, then we don't need to use terminate here, right?

That's correct. Note that `terminate` is an error-free termination of the template expansion, not a termination of the surrounding environment. Using terminate to stop pagination is incorrect.

> Also, I'm not sure about using "do_not_log_failure: true" without "fail_on_template_error: true". Does this scenario prevent changing health status to degraded?

It should.

packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs
```diff
@@ -27,23 +27,33 @@ request.transforms:
...
+        [[if not (eq (len .cursor) 0)]]
+          [[.cursor.last_cursor]]
+        [[else]]
+          [[terminate "cursor is empty"]]
+        [[end]]
```
efd6: We don't need to do this, just `fail_on_template_error: true` with the `do_not_log_failure: true`.

(comment applies throughout)

packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs
```diff
@@ -27,23 +27,33 @@ request.transforms:
...
+      do_not_log_failure: true
   - set:
       target: body.limit
       value_type: int
-      value: '[[if eq (len .cursor) 0]]{{limit}}[[end]]'
+      value: >-
+        [[if eq (len .cursor) 0]]{{limit}}[[else]][[terminate "cursor is not empty"]][[end]]
```
efd6: Same here.

packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs
navnit-elastic: @efd6, Thank you for your answer.

packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs
navnit-elastic: Updated in https://github.com/elastic/integrations/pull/15415/commits/c9ccf9d8059fe4513b77ded11bd5ee376ced97bb. Not enabling `fail_on_template_error` because the error is expected here.

kcreddy: @efd6 @navnit-elastic , just FYI we have an open [SDH](https://github.com/elastic/sdh-beats/issues/6460) on this.

efd6 (commented):
I've checked locally that this resolves the issue in 1password on 9.1.4 and that the current tip fails on that version of the stack.

I would like the change to be simplified though
```diff
diff --git a/packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs b/packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs
index 621af5d15d..1a6101567f 100644
--- a/packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs
+++ b/packages/1password/data_stream/audit_events/agent/stream/httpjson.yml.hbs
@@ -45,7 +45,7 @@ response.split:
 response.pagination:
   - set:
       target: body.cursor
-      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[else]][[/*This is an empty value to stop pagination.*/]][[end]]'
+      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[end]]'
       fail_on_template_error: true
       do_not_log_failure: true
   - delete:
diff --git a/packages/1password/data_stream/item_usages/agent/stream/httpjson.yml.hbs b/packages/1password/data_stream/item_usages/agent/stream/httpjson.yml.hbs
index 36b285bf19..06cb4c0889 100644
--- a/packages/1password/data_stream/item_usages/agent/stream/httpjson.yml.hbs
+++ b/packages/1password/data_stream/item_usages/agent/stream/httpjson.yml.hbs
@@ -46,7 +46,7 @@ response.split:
 response.pagination:
   - set:
       target: body.cursor
-      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[else]][[/*This is an empty value to stop pagination.*/]][[end]]'
+      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[end]]'
       fail_on_template_error: true
       do_not_log_failure: true
   - delete:
diff --git a/packages/1password/data_stream/signin_attempts/agent/stream/httpjson.yml.hbs b/packages/1password/data_stream/signin_attempts/agent/stream/httpjson.yml.hbs
index d12967a150..c5cd14d84d 100644
--- a/packages/1password/data_stream/signin_attempts/agent/stream/httpjson.yml.hbs
+++ b/packages/1password/data_stream/signin_attempts/agent/stream/httpjson.yml.hbs
@@ -46,7 +46,7 @@ response.split:
 response.pagination:
   - set:
       target: body.cursor
-      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[else]][[/*This is an empty value to stop pagination.*/]][[end]]'
+      value: '[[if eq .last_response.body.has_more true]][[.last_response.body.cursor]][[end]]'
       fail_on_template_error: true
       do_not_log_failure: true
   - delete:
```
The else branch is not needed since in the absence of a true condition we already resolve to the empty string, and the use of the `fail_on_template_error` option indicates the rationale for the condition.

efd6 (approved):
Thanks

Issue #14094, opened 2025-05-30
[trend_micro_vision_one]: missing required field accessing 'resource.url'


ipsec:
### Integration Name

Trend Micro Vision One [trend_micro_vision_one]

### Integration Version

2.2.0

### Agent Version

8.18.1

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.18.1

### OS Version and Architecture

SUSE Linux Enterprise Server 15 SP5

### Error Message

missing required field accessing 'resource.url'

### What did you do?

Upgrade from 2.1.0 to 2.2.0.

### What did you see?

![Image](https://github.com/user-attachments/assets/7df3e5fb-9751-42ce-99da-69300f10d0ff)

### What did you expect to see?

No error.

### Anything else?

Was working in 2.1.0

efd6:
~@ipsec Which data stream is this for?~ By exclusion it must be telemetry.

efd6:
@ipsec Please take a look at the pre-config.yaml in a diagnostic bundle from the agent that is complaining of this. You should see something like
```
    - data_stream:
        namespace: "59948"
      id: cel-trend_micro_vision_one-<UUID>
      meta:
        package:
            name: trend_micro_vision_one
            version: 2.2.0
      name: trend_micro_vision_one-telemetry-59948
      package_policy_id: <UUID>
      revision: 1
      streams:
        - config_version: 2
          data_stream:
            dataset: trend_micro_vision_one.telemetry
            type: logs
          fields_under_root: true
          id: cel-trend_micro_vision_one.telemetry-<UUID>
          interval: 15m
          keep_null: true
          program: |4
<SNIP>
          publisher_pipeline:
            disable_host: true
          redact:
            fields:
                - api_token
          resource:
            ssl: null
            timeout: 30s
            tracer:
                enabled: true
                filename: ../../logs/cel/http-request-trace-*.ndjson
                maxbackups: 5
            url: http://svc-trend_micro_vision_one:8080
          state:
            api_token: <REDACTED>
            initial_interval: 2h
            managed_pipeline_description: managed-by-elastic-integrations-trend_micro_vision_one-telemetry
            max_recency: 30s
            max_time_range: 1h
            min_time_range: 10m
            registration_count: 0
            top: 500
            want_more: false
          tags:
            - preserve_original_event
            - preserve_duplicate_custom_fields
            - forwarded
            - trend_micro_vision_one-telemetry
```
[...]

PR #15427, opened 2025-09-22
trend_micro_vision_one: relocate common required options

chemamartinez:
## Proposed commit message

```
To avoid bug described at https://github.com/elastic/kibana/issues/235242,
common required settings such as URL and API Token have been located
under input level, as they were in previous versions.

This forces users to solve conflicts in the integration configuration during
upgrades from any version, avoiding the integration to gets broken once upgraded.

Request tracing option has been moved to each data stream at the same time
to allow for more granular control of request tracing.
```

## Checklist

### 2.4.0 fresh install

All options are shown as expected:

image

image

### Upgrade from < 2.2.0

When clicking on `Upgrade to latest version` the integration is upgraded but the policy upgrade fails due to conflicts.

image

Once the policy is manually upgraded and conflicts are solved, the integration works as expected and all config options appear in the UI for future editions. This is expected in a breaking change according to [our docs](https://www.elastic.co/guide/en/fleet/8.18/upgrade-integration.html#resolve-conflicts).

image

### Upgrade from > 2.2.0

Same situation as for previous case. The breaking change forces users to update policies manually.

image

After the upgrade, everything works fine.

## Related issues

- Closes https://github.com/elastic/integrations/issues/14094
- Relates https://github.com/elastic/kibana/issues/235242

### DIFF
--- a/packages/trend_micro_vision_one/data_stream/alert/manifest.yml
+++ b/packages/trend_micro_vision_one/data_stream/alert/manifest.yml
@@ -71,3 +71,11 @@ streams:
         show_user: false
         description: >-
           Processors are used to reduce the number of fields in the exported event or to enhance the event with metadata. This executes in the agent before the logs are parsed. See [Processors](https://www.elastic.co/guide/en/beats/filebeat/current/filtering-and-enhancing-data.html) for details.
+      - name: enable_request_tracer
+        type: bool
+        title: Enable request tracing
+        default: false
+        multi: false
+        required: false
+        show_user: false
+        description: The request tracer logs requests and responses to the agent's local file-system for debugging configurations. Enabling this request tracing compromises security and should only be used for debugging. See [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-httpjson.html#_request_tracer_filename) for details.
--- a/packages/trend_micro_vision_one/data_stream/detection/manifest.yml
+++ b/packages/trend_micro_vision_one/data_stream/detection/manifest.yml
@@ -79,3 +79,11 @@ streams:
         show_user: false
         description: >-
           Processors are used to reduce the number of fields in the exported event or to enhance the event with metadata. This executes in the agent before the logs are parsed. See [Processors](https://www.elastic.co/guide/en/beats/filebeat/current/filtering-and-enhancing-data.html) for details.
+      - name: enable_request_tracer
+        type: bool
+        title: Enable request tracing
+        default: false
+        multi: false
+        required: false
+        show_user: false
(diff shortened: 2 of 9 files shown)


packages/trend_micro_vision_one/data_stream/alert/manifest.yml
```diff
@@ -71,3 +71,11 @@ streams:
...
         description: >-
           Processors are used to reduce the number of fields in the exported event or to enhance the event with metadata. This executes in the agent before the logs are parsed. See [Processors](https://www.elastic.co/guide/en/beats/filebeat/current/filtering-and-enhancing-data.html) for details.
+      - name: enable_request_tracer
+        type: bool
+        title: Enable request tracing
+        default: false
+        multi: false
+        required: false
+        show_user: false
+        description: The request tracer logs requests and responses to the agent's local file-system for debugging configurations. Enabling this request tracing compromises security and should only be used for debugging. See [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-httpjson.html#_request_tracer_filename) for details.
```
efd6: Add final new line.

packages/trend_micro_vision_one/data_stream/detection/manifest.yml
```diff
@@ -79,3 +79,11 @@ streams:
...
         description: >-
           Processors are used to reduce the number of fields in the exported event or to enhance the event with metadata. This executes in the agent before the logs are parsed. See [Processors](https://www.elastic.co/guide/en/beats/filebeat/current/filtering-and-enhancing-data.html) for details.
+      - name: enable_request_tracer
+        type: bool
+        title: Enable request tracing
+        default: false
+        multi: false
+        required: false
+        show_user: false
+        description: The request tracer logs requests and responses to the agent's local file-system for debugging configurations. Enabling this request tracing compromises security and should only be used for debugging. See [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-httpjson.html#_request_tracer_filename) for details.
```
efd6: Add final new line.

efd6 (commented):
In the upgrade path we see what is apparently the secret added twice (it's not). We could conceivably work around this by naming it with an input tag as well, but I think this would make the more common interaction with the option worse in order to improve this uncommon interaction. I think that probably fleet should do better when it sees that the names are the same and qualify them with the part of the configuration that they come from.

Screenshot from 2025-09-23 08-10-03

cc @nimarezainia ?

chemamartinez:
> Have you confirmed that this fixes the upgrade issue by taking it through the problematic upgrade path manually?

@efd6 yes, I have tested the upgrade path `< 2.2.0 -> 2.3.0 -> 2.4.0` where 2.3.0 gets broken with the missing settings in the UI. The result is the same as the upgrade from `> 2.2.0 described above`.

The integration gets upgraded and conflicts in the policy have to be manually solved, after that, the integration is not affected by the bug anymore.

efd6 (approved):

Issue #15265, opened 2025-09-10
Sentinel One: Dashboard Improvements and fixes


kcreddy:
1. Currently there are only 2 screenshots, but 5 dashboards. Add more screenshots to [manifest.yml](https://github.com/elastic/integrations/blob/main/packages/sentinel_one/manifest.yml#L13-L21), one per dashboard.

    <img width="1632" height="445" alt="Image" src="https://github.com/user-attachments/assets/f27cd4e2-9443-4d66-bcbc-30e6d21ef428" />

2. Update Threat dashboard screenshot. It is quite old.
3. Add navigation panels to all dashboards. Use link panels instead of markdown: https://github.com/elastic/integrations/issues/15075
4. Remove `[Logs SentinelOne]` text from all visualisation titles inside dashboards: https://www.elastic.co/docs/extend/integrations/dashboard-guidelines#_titles_in_visualisations_matter
5. Use new Metrics visualisation instead of Legacy metrics in all dashboards.
6. Add controls to all dashboards.

Issue #15075, opened 2025-08-28
ssi: migrate dashboard navigation to using links panel widgets


efd6:
Kibana now provides a [links panel widget](https://www.elastic.co/docs/explore-analyze/visualize/link-panels) for handling navigation links. Using this widget ensures that navigation styling is consistent across integration dashboards and also prevents issues with [breaking links on serverless](https://github.com/elastic/integrations/issues/13630).

The security-service-integrations dashboards make heavy use of markdown-based navigation panels. These should be migrated to using links panels.

- [ ] admin_by_request_epm
- [ ] abnormal_security
- [ ] amazon_security_lake
- [ ] armis
- [ ] authentik
- [ ] azure_network_watcher_nsg
- [ ] azure_network_watcher_vnet
- [ ] beyondinsight_password_safe
- [ ] beyondtrust_pra
- [x] canva - https://github.com/elastic/integrations/pull/15298
- [ ] checkpoint_email
- [ ] claroty_ctd
- [x] claroty_xdome - https://github.com/elastic/integrations/pull/15298
- [ ] cloudflare_logpush
- [ ] cloudflare
- [ ] corelight
- [ ] cribl
- [ ] crowdstrike
- [ ] cyberark_epm
- [ ] cybereason
- [ ] digital_guardian
- [ ] eset_protect - #15910
- [ ] f5_bigip
- [ ] gigamon
- [x] github - https://github.com/elastic/integrations/pull/15095
- [ ] google_scc
- [ ] google_secops
- [ ] google_workspace
- [ ] infoblox_threat_defense
- [x] jamf_pro - https://github.com/elastic/integrations/pull/15298
- [ ] keycloak
- [ ] m365_defender
- [ ] microsoft_sentinel
- [ ] nozomi_networks
- [ ] panw_cortex_xdr
- [ ] ping_federate
- [ ] prisma_access
- [ ] prisma_cloud
- [ ] proofpoint_on_demand
- [ ] qualys_vmdr
- [ ] servicenow
- [ ] spycloud
- [ ] sublime_security
- [ ] symantec_endpoint_security
- [ ] sysdig
- [x] tenable_ot_security - https://github.com/elastic/integrations/pull/15298
- [ ] ti_abusech
- [ ] ti_anomali
- [ ] ti_cif3
- [ ] ti_crowdstrike
- [ ] ti_cybersixgill
- [ ] ti_eset
- [ ] ti_misp
- [ ] ti_opencti
- [ ] ti_otx
[...]

PR #15016, opened 2025-08-22
Don't migrate the filters when changing the dashboard using the links

TimBosman:
## Proposed commit message

By default the links panel take all the filters from the host dashboard to the destination dashboard. The Cassandra dashboards have a dashboard wide filter that filters the data and gets taken to the other dashboard. Therefore, as a temporary solution I disable that setting, [as discussed here](https://github.com/elastic/integrations/issues/14502#issuecomment-3195811479)

## Checklist

- [X] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [X] I have verified that all data streams collect metrics or logs.
- [X] I have added an entry to my package's `changelog.yml` file.
- [X] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [X] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices)

PR #15437, opened 2025-09-23
[sentinel_one] Update Deprecated Visualizations and Enhanced Dashboards

mohitjha-elastic:
## Proposed commit message
```
sentinel_one: Update dashboards with new screenshots, navigation panels, and visualizations.

Update all dashboards by replacing outdated screenshots and adding navigation panels for
improved usability. Migrated from legacy metrics visualizations to the new Metrics visualization
for consistency and better performance. Additionally, added controls to all dashboards to enhance 
interactivity and user experience.
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/sentinel_one directory.

> elastic-package test -v

## Related Issue

- Closes https://github.com/elastic/integrations/issues/15265


packages/sentinel_one/img/sentinel-one-activities-dashboard.png
efd6, line 1: We should move the filters from the dashboard to the visualisations so that users do not remove them accidentally.

packages/sentinel_one/img/sentinel-one-agents-dashboard.png
kcreddy, line 1: The width of Navigation and Table of Contents in this dashboard is not same as alerts dashboard. Can you make it uniform across all dashboards for improved experience?

packages/sentinel_one/kibana/dashboard/sentinel_one-acd472d9-a6b7-4a53-a58d-06c315764c8d.json
```diff
@@ -402,52 +402,80 @@
...
+                    "attributes": {
+                        "layout": "vertical",
+                        "links": [
+                            {
+                                "destinationRefName": "link_81cc8355-af43-42be-a636-6650086f4e79_dashboard",
+                                "id": "81cc8355-af43-42be-a636-6650086f4e79",
+                                "label": "Activities",
+                                "order": 0,
+                                "type": "dashboardLink"
+                            },
```
kcreddy, line 582: We would need to apply fix from: https://github.com/elastic/integrations/pull/15016 as we don't want to carry the filters/queries from previous dashboards, it may even break them.
By default the links panel carries the filters, so disable the option and re-export dashboards.

Image

cc: @efd6, I think this could be applicable for all integrations when migrating to links panels: https://github.com/elastic/integrations/issues/15075

efd6: This is a very good reason why filters should be on the visualisation and not on the dashboard. I think that we _should_ use the filters and queries from the origin, but that those filters should _only be_ filters that the user has applied and the filters that are required for correctness are not affected.

kcreddy: If we enable it, in scenario when users apply the filter/query on a custom field, say `sentinel_one.threat.fieldA: true` on Threat dashboard, and when the user navigates to alert dashboard the experience would be broken (because this custom field doesn't exist in alert data stream). 
Is that expected correctly?

efd6: Sorry, can you clarify? If the user applies a filter and it doesn't work, they are the person who has applied it and so should be aware of its presence, so they can rectify it. ISTM that it _should not_ break, but if it does the solution would be to remove it. On the other hand, if an analyst is working on a train of thought which involves a specific filter, and wants to correlate two dashboards, then keeping the filter in place makes sense. Now, this may not be the situation always, but it is easier to remove a filter than it is to have to replace it when the system has removed it. I don't think there is definitively correct default, but IMO leaving them in place makes more sense.

kcreddy: I haven't thought from an analyst POV, it makes sense that way. Thanks!
We can let it be at default (carry filters/queries) for now.

kcreddy: @mohitjha-elastic, can you please revert the change for links panel and allow them to carry the filters?

mohitjha-elastic: Thanks! @efd6 @kcreddy 
Reverted the changes related to the nav links panel carrying the filter. Please take a look.

PR #15441, opened 2025-09-23
[kubernetes] adds beta support for rotated logs (including GZIP)

### LINKED ISSUE #14238: [kubernetes]: Add GZIP support for logs ingested by filestream
### Integration Name

Kubernetes [kubernetes]

### Dataset Name

kubernetes.container_logs

### Integration Version

1.80.2

### Agent Version

9.1.0

All supported

### User Goal

With the addition of GZIP support to the `filestream` input (https://github.com/elastic/beats/issues/44866), the Kubernetes integration can be updated to expose and leverage this new capability. This will ensure users can easily enable GZIP log collection.
The GZIP support on filestream is on tech preview and is an opt-in feature

### Existing Features

Currently it isn't possible to ingest GZIP files

### What did you see?

Currently it isn't possible to ingest GZIP files

### Anything else?

_No response_

AndersonQ:
# DO NOT MERGE BEFORE 9.2 IS RELEASED

## Proposed commit message

```
[kubernetes] adds beta support for rotated logs

Enables the collection of rotated log files, including GZIP-compressed files, for Kubernetes container logs.

This enhancement introduces new configuration options to control the ingestion of rotated logs, including specifying paths for the rotated logs and enabling GZIP decompression.
```

## Checklist

 - install this integration to an ES stack
```
cd packages/kubernetes
elastic-package install -v
```

 - start a kind cluster with the follwing config

<details>
  <summary>node-cfg.yaml</summary>

```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
  - role: control-plane
    kubeadmConfigPatches:
      - |
        kind: KubeletConfiguration
        containerLogMaxSize: "10Mi"
        containerLogMaxFiles: 10
```
</details>   

```
kind create -v cluster --config ./node-cfg.yaml
```

 - run a flog pod, wait the logs to be generated.

<details>
  <summary>flog.yaml</summary>

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: flog-log-generator
spec:
  template:
    spec:
      containers:
        - name: flog
          image: mingrammer/flog
#          too small "-d" won't give kubelet time to rotate the files
          args: ["-t", "stdout", "-f", "json", "-n", "185000", "-d", "1ms"]
      restartPolicy: Never
  backoffLimit: 4
```
</details>  

```
kubectl apply -f flog.yaml
```

 - add the kubernets integration enabling to ingest rotated logs. Ideally only enable container logs to avoid errors as kind is really minimal and not all metrics are available
 - install a 9.2.0-SNAPSHOT elastic-agent using the instructions for deploying it on k8s
 - ensure all `185000` were ingested and the ingested files include the rotated plain text and gzip files
 

## Related issues

- Closes #14238

## Screenshots

Screenshot from 2025-09-30 15-07-07

### DIFF
--- a/packages/kubernetes/docs/container-logs.md
+++ b/packages/kubernetes/docs/container-logs.md
@@ -12,17 +12,47 @@ By default, only [container parser](https://www.elastic.co/guide/en/beats/filebe
 stack: beta 9.2.0
 ```
 
+The integration can monitor and ingest rotated Kubernetes container logs, including 
+on-the-fly decompression of GZIP archives. To enable this, `add gzip_experimental: true` 
+under _Advanced options > Custom configurations_ and set the path to the rotated 
+log files in the Kubernetes container log path configuration. Refer to
+[filestream documentation on reading GZIP files](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-filestream#reading-gzip-files)
+
+### Which path to add
+
+Kubernetes stores logs on `/var/log/pods` and uses symlinks on `/var/log/containers`
+for active log files. For full details, refer to the official 
+[Kubernetes documentation on log rotation](https://kubernetes.io/docs/concepts/cluster-administration/logging/#log-rotation).
+
+The path you add depends on whether this is a new or existing integration.
+
+#### For new integrations
+
+For new installations, instead of monitoring both paths, you can monitor only 
+`/var/log/pods`, which includes active and rotated files.
+
+Add the following path to the Kubernetes container log path configuration: 
+`/var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log*`
+This wildcard (`*.log*`) matches both active (`.log`) and rotated (`.log.*`) 
+files.
+
+#### For existing integrations
+By default, existing integrations monitor active logs via `/var/log/containers`. 
+To add rotated logs, you must add the path to the rotated log files:
+`/var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log.*`
+
+This wildcard (`*.log.*`) specifically targets only the rotated archive files.
+
 ::::{warning}
-**Potential Data Duplication**: When you add a path to an existing integration, the integration reads all existing files in that directory from the beginning. This action causes a one-time re-ingestion of all previously rotated logs, which results in duplicate data.
+Potential Data Duplication: When you add this path to an existing integration, 
+the integration reads all existing files in that directory from the beginning. 
+This action causes a one-time re-ingestion of all previously rotated logs, which 
+results in duplicate data.
+
+After the initial scan, the integration tracks files normally and will only 
+ingest new log data. Subsequent file rotations are handled correctly without 
+further data duplication.
 ::::
-
-The integration can monitor and ingest rotated Kubernetes container logs, including on-the-fly decompression of GZIP archives.
-
-To enable this, use the following configurations:
-* **Include rotated log files**: Set to `true`.
-* **Kubernetes container rotated log path**: in case of a custom (non-default) logging configuration in Kubernetes, use this parameter to specify your custom path where rotated log files are stored.
-
-After this initial scan, the integration tracks files normally and will only ingest new log data. Subsequent file rotations are handled correctly without further data duplication.
 
 ## Rerouting and preserve original event based on pod annotations
 


packages/kubernetes/_dev/build/docs/container-logs.md
```diff
@@ -5,8 +5,24 @@ container-logs integration collects and parses logs of Kubernetes containers.
 It requires access to the log files in each Kubernetes node where the container logs are stored.
 This defaults to `/var/log/containers/*${kubernetes.container.id}.log`.
 
-By default only {{ url "filebeat-input-filestream-parsers" "container parser" }} is enabled. Additional log parsers can be added as an advanced options configuration.
+By default, only [container parser](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream.html#_parsers) is enabled. Additional log parsers can be added as an advanced options configuration.
```
rdner: I don't think this change was necessary. This file is a template and we use helpers. However, I don't have a strong opinion here.

packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -6,9 +6,25 @@ streams:
...
+        show_user: true
+        default: true
+      - name: data_stream.dataset
+        required: true
+        default: kubernetes.container_logs
+        title: Dataset name
+        description: >
+          Set the name for your dataset. Changing the dataset will send the data to a different index. For more info look at [data_stream field](https://www.elastic.co/guide/en/ecs/master/ecs-data_stream.html).
+        type: text
+
```
rdner: Could you please move it back where it was, so the diff is cleaner for future references? Otherwise "git blame" does not account for the right changes.

packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -6,9 +6,25 @@ streams:
...
       - name: paths
         type: text
         required: true
+        show_user: false
```
rdner: why did we change that?

packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -20,21 +36,32 @@ streams:
...
+      - name: rotated_logs
         type: bool
-        title: Use Symlinks
+        title: Include rotated log files
         multi: false
+        required: false
+        show_user: false
+        default: false
+        description: |
+          Warning: it might lead to data re-ingestion. When enabled, Elastic-Agent will also ingest the rotated log files for each pod, including GZIP-compressed files (.gz), decompressing them as they are read. This is beta feature. For full details about GZIP decompression, see the [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream#reading-gzip-files). Available for Elastic Agent 9.2.0 and newer.
```
rdner: In my opinion all the warnings should be in the very beginning to reduce the chance of them being ignored. Also, just "re-ingestion" is not enough, we should be clear about data duplication here.

```suggestion
          It's a beta feature. Warning: changing this parameter leads to a temporary data duplication due to one time re-ingestion. When enabled, Elastic-Agent will also ingest the rotated log files for each pod, including GZIP-compressed files (.gz), decompressing them as they are read. For full details about GZIP decompression, see the [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream#reading-gzip-files). Available for Elastic Agent 9.2.0 and newer.
```

packages/kubernetes/_dev/build/docs/container-logs.md
```diff
@@ -5,8 +5,24 @@ container-logs integration collects and parses logs of Kubernetes containers.
...
+
+::::{warning}
+**Potential Data Duplication**: When you first add a path, the integration reads all existing files in that directory from the beginning. This action causes a one-time re-ingestion of all previously rotated logs, which results in duplicate data.
+::::
+
+The integration can monitor and ingest rotated Kubernetes container logs, including on-the-fly decompression of GZIP archives.
+
+To enable this, use the following configurations:
+* **Include rotated log files**: Set to `true`.
+* **Kubernetes container rotated log path**: Specify the path where rotated log files are stored.
```
rdner: We already set a reasonable default. The current statement is confusing and tells the user to always set this parameter even if they don't need it by default.

```suggestion
* **Kubernetes container rotated log path**: in case of a custom (non-default) logging configuration in Kubernetes, use this parameter to specify your custom path where rotated log files are stored.
```

packages/kubernetes/_dev/build/docs/container-logs.md
```diff
@@ -5,8 +5,24 @@ container-logs integration collects and parses logs of Kubernetes containers.
...
+
+After this initial scan, the integration tracks its state and will only ingest new log data. Subsequent file rotations are handled correctly without further duplication.
```
rdner:
```suggestion
After this initial scan, the integration tracks files normally and will only ingest new log data. Subsequent file rotations are handled correctly without further data duplication.
```

packages/kubernetes/docs/container-logs.md
```diff
@@ -5,8 +5,24 @@ container-logs integration collects and parses logs of Kubernetes containers.
...
+**Potential Data Duplication**: When you first add a path, the integration reads all existing files in that directory from the beginning. This action causes a one-time re-ingestion of all previously rotated logs, which results in duplicate data.
+::::
+
+The integration can monitor and ingest rotated Kubernetes container logs, including on-the-fly decompression of GZIP archives.
+
+To enable this, use the following configurations:
+* **Include rotated log files**: Set to `true`.
+* **Kubernetes container rotated log path**: Specify the path where rotated log files are stored.
+
+After this initial scan, the integration tracks its state and will only ingest new log data. Subsequent file rotations are handled correctly without further duplication.
```
rdner: This file needs to be regenerated once the template is updated.

packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -20,21 +36,32 @@ streams:
...
-        required: true
-        default: kubernetes.container_logs
-        title: Dataset name
-        description: >
-          Set the name for your dataset. Changing the dataset will send the data to a different index. For more info look at [data_stream field](https://www.elastic.co/guide/en/ecs/master/ecs-data_stream.html).
+        show_user: false
+        title: Kubernetes container rotated log path
+        multi: true
+        default:
+          - /var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log.*
```
rdner: I find it suspicious that for the `paths` our default is `/var/log/containers/*${kubernetes.container.id}.lo` but for the `rotated_logs_paths` it's a completely different `/var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log.*`.

I have a couple of questions here:

1. Why do we even need the new `rotated_logs_paths` parameters? Why can't we just add a second line to the `paths` default?
2. How did we determine that the rotated container logs are stored in `/var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log.*`? Can we put a documentation reference as a comment near this default?

packages/kubernetes/_dev/build/docs/container-logs.md
belimawr: I agree with @rdner, keeping the template helper is a good idea. Maybe something like this:
```suggestion
By default, only [container parser]({{ url "filebeat-input-filestream-parsers" "container parser" }}) is enabled. Additional log parsers can be added as an advanced options configuration.
```

I did not manage to build the docs to test :/.

How did you test the docs?

packages/kubernetes/_dev/build/docs/container-logs.md
```diff
@@ -5,8 +5,24 @@ container-logs integration collects and parses logs of Kubernetes containers.
...
+By default, only [container parser](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream.html#_parsers) is enabled. Additional log parsers can be added as an advanced options configuration.
 
+## Ingesting Rotated Container Logs
+```{applies_to}
+stack: beta 9.2.0
+```
+
+::::{warning}
+**Potential Data Duplication**: When you first add a path, the integration reads all existing files in that directory from the beginning. This action causes a one-time re-ingestion of all previously rotated logs, which results in duplicate data.
+::::
```
belimawr: [Suggestion]

Add something to make extra clear that this happens only when enabling this to an existing integration
```suggestion
::::{warning}
**Potential Data Duplication**: When you add a path to an existing integration, the integration reads all existing files in that directory from the beginning. This action may cause a one-time re-ingestion of all previously rotated logs, which results in duplicate data.
::::
```

packages/kubernetes/_dev/build/docs/container-logs.md
belimawr: I agree with this paragraph, it is a good warning for people enabling this to an existing integration, however it is not clear this is only a concert when enabling this to an existing integration. 

Honestly, do we even need this paragraph here? Isn't the warning at the beginning enough?

packages/kubernetes/data_stream/container_logs/manifest.yml
AndersonQ: > Why do we even need the new rotated_logs_paths parameters?

I added some context [on the issue](https://github.com/elastic/integrations/issues/14238#issuecomment-3322776076), but the tl;dr is that the paths serve different purposes. The `/var/log/containers/` is deprecated and is just a collection of symlinks pointing to the active log file. To get the rotated logs, we have to target the source directory, `/var/log/pods/`, directly.

As for why this is a new parameter, it was the cleanest and safest approach for a few reasons:

**Isolating a beta feature**: Since support for rotated/gzipped logs is new, it's safer to keep it separate.

**Preventing data re-ingestion**: Changing the existing default paths could cause users to re-ingest logs, which we must avoid.

**Simplicity/limitations**: With a different path, it can be added or not depending on `rotated_logs` toggle, without interfering with the default to active log files.

> How did we determine that the rotated container logs are stored in /var/log/pods/${kubernetes.namespace}_${kubernetes.pod.name}_${kubernetes.pod.uid}/${kubernetes.container.name}/*.log.*? 

A mix of docs and reading k8s code. Unfortunately I did not find authoritative and detailed docs on this path, just the logs are in `/var/log/pods/`. I also explain [on the issue](https://github.com/elastic/integrations/issues/14238#issuecomment-3322776076) and link the docs/source code from where I got this path. I'll add some comment here

packages/kubernetes/data_stream/container_logs/manifest.yml
AndersonQ: I moved so it reflects the UI. As this is shown by default at the top, on the UI I put it on top.
I should have moved the other that is shown by default as well.

belimawr: Yeah, I'm also confused about this move. I believe it's better to keep the original order, so the paths are the first option in the advanced configuration.

packages/kubernetes/data_stream/container_logs/manifest.yml
AndersonQ: zen of python, "explicit is better than implicit" and to be consistent with the other path related entries. I can add it to the other properties for container logs. Right now it's inconsistent

packages/kubernetes/data_stream/container_logs/manifest.yml
belimawr: I agree with @rdner, but I'd word it a little different to state that data duplication only happens on existing integrations.
```suggestion
          It's a beta feature. Warning: enabling this on an existing integration might cause data duplication due to one time re-ingestion of existing rotated files in the newly added paths. When enabled, Elastic-Agent will also ingest the rotated log files for each pod, including GZIP-compressed files (.gz), decompressing them as they are read. For full details about GZIP decompression, see the [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream#reading-gzip-files). Available for Elastic Agent 9.2.0 and newer.
```

packages/kubernetes/_dev/build/docs/container-logs.md
AndersonQ: my bad. I edited the rendered file, then I discovered this one, so I copied it from there, thus the rendered link. Thanks for spotting that.

packages/kubernetes/data_stream/container_logs/manifest.yml
rdner: Yes, this phrasing is indeed better.

packages/kubernetes/_dev/build/docs/container-logs.md
AndersonQ: I like the 1st part, but it'll always re ingest the rotated log files. Thus I'd rather not add the _may cause_

packages/kubernetes/data_stream/container_logs/manifest.yml
AndersonQ: I organised it better, are you really against it reflecting the UI, if so I can move it back. However I believe now it's easier to maintain.

packages/kubernetes/data_stream/container_logs/manifest.yml
AndersonQ: I added `show_user: false` to all others, so it clear what is and what isn't shown by default.

packages/kubernetes/docs/container-logs.md
AndersonQ: oh, `install` only build the integration, but not the docs. Good to know

AndersonQ:
@rdner, @belimawr I also updated the screenshot in the PR description. please re-review at your earliest convenience

packages/kubernetes/data_stream/container_logs/manifest.yml
belimawr: If it is the same order as the UI, then it's fine.

packages/kubernetes/data_stream/container_logs/manifest.yml
rdner: @AndersonQ thanks for the clarification!

My concern is our future plan for this feature. How are we going to take it out of beta? Would this extra parameter `rotated_logs_paths` stay?

While it's still in Beta it might be totally fine to hard-code `/var/log/pods/${kubernetes.namespace}${kubernetes.pod.name}${kubernetes.pod.uid}/${kubernetes.container.name}/.log` as a part of `paths` in the template if `rotated_logs` is switched to true without adding a new `rotated_logs_paths` parameter. How common is it to need to configure `rotated_logs_paths`?

If we add this new parameter `rotated_logs_paths`, we commit to having it in the future too.

I'd like to see @cmacknz's take on it.

cmacknz: > My concern is our future plan for this feature. How are we going to take it out of beta? Would this extra parameter rotated_logs_paths stay?

No, we would make this the default, we just have to do it without data duplication in this case.

I think my overall preference would be to have a feature toggle for something like "ingest rotated gzip logs (beta)" that just makes all the changes necessary without exposing the individual pieces that are changed.

I think the toggles now are technically correct but way to technical for a lot of users to potentially understand. The integration is supposed to abstract away complexity and this just putting the underlying Kubernetes log structure in everyones view directly.

cmacknz: For reference this is what filelog does:

```yaml
filelog:
  include:
    - /var/log/pods/*/*/*.log
  exclude:
    # Exclude logs from all containers named otel-collector
    - /var/log/pods/*/otel-collector/*.log
  start_at: end
  include_file_path: true
  include_file_name: false
  operators:
    # parse container logs
    - type: container
      id: container-parser
```

You will note they do not template an input per pod UID and they have an obvious way to exclude their own logs. 

This something we should also do we are blocked by requiring data re-ingestion to make changes like this. We need some way to correct our logging configurations and change them at will without re-ingesting every log in a cluster. Several improvements are stuck behind this problem.

cmacknz: https://opentelemetry.io/docs/platforms/kubernetes/collector/components/#filelog-receiver

AndersonQ: hey folks, 

> How are we going to take it out of beta? Would this extra parameter `rotated_logs_paths` stay?

This parameter exists only on the template. So we should be able to remove it without much problem. If ingesting rotated logs become the default, we can remove the parameter and add the path as part of the default `paths` variable.

> How common is it to need to configure `rotated_logs_paths`?

I don't know. But I'd expect we want to allow users to do so, even if it's just a few. I can imagine users requesting it if we do not allow at first.

> You will note they do not template an input per pod UID and they have an obvious way to exclude their own logs.

the _per pod UID_ templating was just to be consistent with how we template the path for the active log file under `/var/log/container/`. I'm happy to use a glob for `/var/log/pod/`. Also because k8s docs don't really document either pattern, the one for `/var/logs/container/` and `/var/log/pod/`.


> This something we should also do we are blocked by requiring data re-ingestion to make changes like this. We need some way to correct our logging configurations and change them at will without re-ingesting every log in a cluster. Several improvements are stuck behind this problem.

+1
yeah, we could go to something simpler with a single path glob, but if filebeat considers the old path deleted, it might re-ingest logs and we don't want it.

> For reference this is what filelog does:

this is for the active log file. It isn't tracking the rotated logs. The rotated logs are like `*.log.*`. Anyway, Ideally we'd change to `var/log/pod` as the `/var/log/container/` is deprecated.

belimawr: > I'm happy to use a glob for /var/log/pod/. Also because k8s docs don't really document either pattern, the one for /var/logs/container/ and /var/log/pod/.

Because the input is templated using the container ID in the input ID, for each container we have one input, which currently works well with the paths matching a single file. However, if we add a glob in the paths, then for each container we will have one input instance ingesting files from all containers, that's gonna generate massive data duplication.

> For reference this is what filelog does:

It's interesting that they set `start_at: end`, their default is to consider every existing file as fully ingested and only ingest new data. That helps a lot to avoid data duplication, but do not ingest old data like we do.

belimawr: > This something we should also do we are blocked by requiring data re-ingestion to make changes like this. We need some way to correct our logging configurations and change them at will without re-ingesting every log in a cluster. Several improvements are stuck behind this problem.

We already can migrate state from one Filestream input to another. The current implementation is very conservative and requires us to set the ID from the old input we're getting the state from, however we can loose this requirement and just have input look up any state in Filestream's registry, regardless of the input it belongs to.

The more we discuss about it, the more I believe we need to re-think how we store and scope file states in Filestream.

packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -6,9 +6,25 @@ streams:
     title: Collect Kubernetes container logs
     description: Collect Kubernetes container logs
     vars:
+      - name: symlinks
+        type: bool
+        title: Use Symlinks
```
cmacknz: This shouldn't be called "Use Symlinks". I don't think any end user will understand what this means or the implications of it.

You should have a single setting called something like "Enable ingestion of rotated GZIP logs (beta)" that toggles between the old and new paths with a warning that it causes the first set of rotated logs to be re-ingested.

The intent is to allow people to turn this on if they need it while we work out how to migrate things nicely under the hood.

packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -20,45 +36,54 @@ streams:
           the paths contain no variable! Refer to the [integration
           documentation](https://www.elastic.co/guide/en/integrations/current/kubernetes.html)
           for more details.
-      - name: symlinks
+      - name: rotated_logs
         type: bool
-        title: Use Symlinks
+        title: Include rotated log files
```
cmacknz: This needs to have GZIP in the name to explain why you'd want to do this.

packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -20,45 +36,54 @@ streams:
...
         multi: false
+        required: false
+        show_user: false
+        default: false
+        description: |
+          It's a beta feature. Warning: changing this parameter leads to a temporary data duplication due to one time re-ingestion. When enabled, Elastic-Agent will also ingest the rotated log files for each pod, including GZIP-compressed files (.gz), decompressing them as they are read. Refer to the [integration documentation](https://www.elastic.co/docs/reference/integrations/kubernetes/container-logs) for more details and refer to the [GZIP decompression documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream#reading-gzip-files) for more details about GZIP decompression. Available for Elastic Agent 9.2.0 and newer.
+      - name: rotated_logs_paths
```
cmacknz: Doesn't the path here also include the live file? Or is the `log.*` the equivalent of the regular expression `log\.*` where we need to match a literal `.` and it isn't the regular `.` "match zero or more times" regex expression?

packages/kubernetes/data_stream/container_logs/manifest.yml
AndersonQ: it was already like that. The Symlinks are used to ingest the active log file. I just put it up so the yaml reflects the UI ordering

packages/kubernetes/data_stream/container_logs/manifest.yml
AndersonQ: it's a glob-pattern, a `.` is a `.`: https://pkg.go.dev/path/filepath#Glob

packages/kubernetes/data_stream/container_logs/manifest.yml
AndersonQ: > The more we discuss about it, the more I believe we need to re-think how we store and scope file states in Filestream.

Oh yeah

AndersonQ: > > I'm happy to use a glob for /var/log/pod/. Also because k8s docs don't really document either pattern, the one for /var/logs/container/ and /var/log/pod/.
> 
> Because the input is templated using the container ID in the input ID, for each container we have one input, which currently works well with the paths matching a single file. However, if we add a glob in the paths, then for each container we will have one input instance ingesting files from all containers, that's gonna generate massive data duplication.
> 

Oh, that's a good pint. Lat me think about it

AndersonQ: It's fine. The variables are substituted all together. In a simplified way:

the variables of an input are only substituted if all variables are returned by the provider 
the k8s provider returns all the variables in a set with a single value for each key. In other words, for a pod with 2 containers there would be 2 sets like:
```
namespace: default
pod: 
  uid: 1
  name: pod-1
container:
  id: cont-1
  name cont-name-1

namespace: default
pod:
  uid: 1
  name: pod-1
container:
  id: cont-2
  name cont-name-2
```
that way, there is no risk of creating a combination that mixes 2 pods or containers from the same pod as each set of values are all valid and belonging to the same hierarchy

AndersonQ:
@cmacknz, @rdner is it ok now?

rdner:
@AndersonQ I don't think this [comment](https://github.com/elastic/integrations/pull/15441/files#r2402418466) from @cmacknz is addressed, is it?

My initial question remains unanswered: why do we need the new configuration parameter `rotated_logs_paths` when we can just add values in `paths` based on one feature flag/toggle `rotated_logs`?

If I understood correctly, @cmacknz and I are aligned on this. We should minimize the amount of new configuration fields we introduce, especially when we know this parameter won't be necessary in the future. I believe it should be just one toggle.

AndersonQ:
> @AndersonQ I don't think this [comment](https://github.com/elastic/integrations/pull/15441/files#r2402418466) from @cmacknz is addressed, is it?

@rdner, I thought my explanation [here](https://github.com/elastic/integrations/pull/15441/files#r2405104402) addressed it.

> My initial question remains unanswered: why do we need the new configuration parameter `rotated_logs_paths` when we can just add values in `paths` based on one feature flag/toggle `rotated_logs`?
> 
> If I understood correctly, @cmacknz and I are aligned on this. We should minimize the amount of new configuration fields we introduce, especially when we know this parameter won't be necessary in the future. I believe it should be just one toggle.

it's about allowing users to configure it if they have another path for it. I can definitely go with just a toggle, but then it won't be configurable.
I don't know how often it's needed, I'd guess pretty much as much as one needs to change the default path for logs. That minus the amount of users which change it to collect directly from `/var/log/pod`

Also, if for some reason the users have the logs in a custom location, other than `/var/log/pod`, I doubt the rotated logs would be there, thus, they need to configure it.

It's a sane default, just as we have for the active log. I don't see why we should not allow it to be configurable as well.

AndersonQ:
I'm happy to remove it if you 2 want it removed. However I'd not be surprised if we get requests to make it configurable.

rdner:
@AndersonQ It's not about what we two want, it's about what makes sense and how we're going to present this beta feature now and take it out of beta in the future.

I see what we do here in this PR as temporary – to let our customers try it out. When we figure out how to migrate from one path to another without data duplication, we just make the path we toggle now by the `rotated_logs: true` a new default.

What are you going to do with the `rotated_logs_paths` then?

AndersonQ:
> how we're going to present this beta feature now and take it out of beta in the future.
> 
> I see what we do here in this PR as temporary – to let our customers try it out. When we figure out how to migrate from one path to another without data duplication, we just make the path we toggle now by the `rotated_logs: true` a new default.
> 
> What are you going to do with the `rotated_logs_paths` then?

If, once we can add new paths and we migrate the default log path from the integration to `/var/log/pods`, it might not be necessary and we can remove it if we don't want to allow users to modify it anymore.

There is at least the windows case, the path for windows is different, so we might always need it just as we allow to modify the active logs path.
Or if we move to even removing the toggle to ingest rotated logs, again, we remove the toggle and the `rotated_logs_paths`.

For now, I believe we should let it be configurable. It's possible to change it later, so I don't see it as a problem.

@rdner, @cmacknz, what do you think?

rdner:
@AndersonQ the problem is that in this PR we allow to configure the same Filebeat configuration parameter `paths` with 2 integration configuration parameters: `paths` and `rotated_logs_paths`. Furthermore, we now allow to configure `rotated_logs_paths`, our users will start to rely on it and then we remove it. This is a breaking change. 

I'd rather avoid introducing something that is unnecessary and would lead to a breaking change later.

If we don't support the Windows case in Beta and document it – it's fine. The current objective is to give early access to the GZip ingestion, not to solve the entire use-case.

I'm going to leave the final decision to @cmacknz since it looks like my argumentation alone is not sufficient here.

cmacknz:
When I first thought of this toggle approach I was only considering the `gzip_experimental: true` part which we expect to eventually become the default, so we could remove the toggle with no consequences. The additional need to change the paths complicates this quite a bit.

I also see that we can solve having people enable this purely via a documentation, for example if a user adds `gzip_experimental: true` in the `{{custom}}` block and changes the paths appropriately then we don't have to make any changes here. So my preference would be to do that.

We can change approach from this PR into instead making a PR to the beats+elastic agent documentation (because the same change works for standalone not just this integration) and include the same instructions in the relevant changelogs.

Then we don't have to worry about trying to remove these temporary settings later.

AndersonQ:
> We can change approach from this PR into instead making a PR to the beats+elastic agent documentation (because the same change works for standalone not just this integration) and include the same instructions in the relevant changelogs.
> 
> Then we don't have to worry about trying to remove these temporary settings later.

ok then.

@cmacknz, do you have a particular agent docs in mind or you meant the k8s integration docs? On top of my head I don't know which agent docs you're referring to. 

Also I believe we can update the integrations doc to either mention it or to point to the agent's doc. Thus, so far I believe the following docs can be updated to mention ingesting GZIP files:

 - k8s [integration docs](packages/kubernetes/_dev/build/docs/container-logs.md)
 - filebeat ["run on k8s#log-rotation"](https://www.elastic.co/docs/reference/beats/filebeat/running-on-kubernetes#_logrotation)
 - filebeat [run on k8s#autodiscover snippet](https://www.elastic.co/docs/reference/beats/filebeat/running-on-kubernetes#_red_hat_openshift_configuration) - I think it should not be under OpenShift. I don't know if it's intentional or a mistake.
 - filebeat [deploy on k8s](https://github.com/elastic/beats/tree/9da0f31ee2cc0f57de0b16a69bdd120e25b5cb76/deploy/kubernetes/filebeat)
 - filebeat [Log rotation results in lost or duplicate events](https://www.elastic.co/docs/reference/beats/filebeat/file-log-rotation) (this is unrelated to k8s, but can profit from at least us mentioning we can read gzip files)

do you have anything else in mind?

cmacknz:
> Also I believe we can update the integrations doc to either mention it or to point to the agent's doc. 

Yes if we can document this in one single place, perhaps the filestream docs, and then just reference the new section everywhere that would be best.

The only places I see in addition we would want to mention this are:
- The Beats and Agent changelog, which can also reference the new docs section
- The Elastic Agent Helm chart in the readme probably, we need to confirm there is a way to configure this there.https://github.com/elastic/elastic-agent/tree/main/deploy/helm/elastic-agent

AndersonQ:
as per https://github.com/elastic/integrations/pull/15441#issuecomment-3434122796 it'll be just a doc change. So to better track it and keep the discussions focused, I'll open another PR and use https://github.com/elastic/integrations/issues/15792 to track the task

PR #15476, opened 2025-09-26
o365: Fix 429 due to multiple subscription start attempts.

kcreddy:
## Proposed commit message
```
o365: Fix 429 due to multiple subscription start attempts.

Right now, the code makes multiple requests to API:
"/subscriptions/start?contentType=" which causes the error:
"Too many frequent subscription start requests. Please retry again after 14m 56s"

Update CEL program to call "/subscriptions/start?contentType=" only
when the contentType is not already started. This was checked
using presence of contentType inside "cursor.last_for".
```

## Checklist

### DIFF
--- a/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
+++ b/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
@@ -276,6 +276,7 @@ program: |-
               "work": work.with(
                 {
                   "sub": (has(state.?cursor.last_for) && work.curr_type.to_lower() in state.cursor.last_for) ?
+                    // We have started this content type before, so set "enabled: true".
                     {
                       "type": work.curr_type,
                       "enabled": true,
--- a/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
+++ b/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
@@ -275,40 +275,46 @@ program: |-
               "base": state.base,
               "work": work.with(
                 {
-                  "sub": request(
-                    "POST",
-                    sprintf(
-                      "%s/api/v1.0/%s/activity/feed/subscriptions/start?%s",
-                      [
-                        state.url.trim_right("/"),
-                        state.base.tenant_id,
-                        {
-                          "contentType": [work.curr_type],
-                          "PublisherIdentifier": [state.base.tenant_id],
-                        }.format_query(),
-                      ]
-                    )
-                  ).do_request().as(sub_resp, (sub_resp.StatusCode == 200 || sub_resp.StatusCode == 400) ?
-                    sub_resp.Body.decode_json().as(sub,
-                      {
-                        "type": work.curr_type, // If we get an AF20024, sub.contentType will be absent.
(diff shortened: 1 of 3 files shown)


packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -275,40 +275,47 @@ program: |-
...
                         {
-                          "contentType": [work.curr_type],
-                          "PublisherIdentifier": [state.base.tenant_id],
-                        }.format_query(),
-                      ]
-                    )
-                  ).do_request().as(sub_resp, (sub_resp.StatusCode == 200 || sub_resp.StatusCode == 400) ?
-                    sub_resp.Body.decode_json().as(sub,
+                          "type": work.curr_type, // If we get an AF20024, sub.contentType will be absent.
+                          "enabled": sub.?status == optional.of("enabled") || sub.?error.code == optional.of("AF20024"),
```
kortschak, line 302: With the new logic, we should not see AF20024 errors anymore unless the subscription has held over from a previous state that did not make it into the cursor, correct?

Also, what is the life cycle of a subscription according the the API? Do we know for certain that if we have a `last_for` mark in the cursor for the type we are querying that there will be an active subscription? That is, in L278-L283, can we guarantee that the final value of `enabled` actually reflects the subscription state for the type.

kcreddy: > With the new logic, we should not see AF20024 errors anymore unless the subscription has held over from a previous state that did not make it into the cursor, correct?

Yes, `AF20024` shouldn't happen unless the edgecase scenario you specified above occurs.

> Also, what is the life cycle of a subscription according the the API?

Based on [`/start`](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference#start-a-subscription) subscription call documentation, I didn't find expiration for the API polling method. There is optional user-configurable expiration when using webhooks, but not for the API polling. My gut says it doesn't expire unless called with [`/stop`](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference#stop-a-subscription).

There are few ways we can handle, in case the presence of the `cursor.last_for` is not enough.
1. Call [`/list`](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference#list-current-subscriptions) current subscriptions before [`/content`](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference#list-available-content) listing of a `contentType` and check if subscription exists and is enabled.
2. Check for common [error codes](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference#errors) that [`/content`](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference#list-available-content) can return. When subscription is stopped using `/stop`, the following is returned:
    ```json
    {
        "error": {
            "code": "AF20023",
            "message": "The subscription was disabled."
        }
    }
    ```
    I also think `AF20022: No subscription found for the specified content type` could be possible, but not tested.

PR #15487, opened 2025-09-28
Add dashboards for log categories Application Logs, Audit Logs, Console Logs, HTTP Logs, IPsec Audit Logs and Platform Logs.

giorgi-imerlishvili-elastic:
## Proposed commit message
Add dashboards for log categories Application Logs, Audit Logs, Console Logs, HTTP Logs, IPsec Audit Logs and Platform Logs.

## Checklist

- [ X] I have added an entry to my package's `changelog.yml` file.

## Author's Checklist

- [ ] 

## Related issues

-


packages/azure_app_service/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: "3.0.2"
 name: azure_app_service
 title: "Azure App Service"
-version: "0.7.0"
+version: "0.10.0"
 source:
   license: "Elastic-2.0"
 description: "Collect logs from Azure App Service with Elastic Agent."
```
Linu-Elias, line 7: You can add the `screenshots` attribute in the manifest file and include references to the dashboard image, similar to how it's done in other integrations.

packages/azure_app_service/changelog.yml
```diff
@@ -1,4 +1,19 @@
 # newer versions go on top
+- version: "0.10.0"
```
Linu-Elias: Instead of having multiple changelog entries for the same PR, you can consolidate them into a single entry with a combined description.
example for reference: https://github.com/elastic/integrations/blob/d8a5e6d790772a0cc837fc5313ca4abbc7a6270c/packages/ti_abusech/changelog.yml#L4

giorgi-imerlishvili-elastic: I wanted to have in one change log as well. I like combined description idea, but what about `type` it can only have `enhancement` or `bugfix` value as I know. This PR contains enhancement and also bugfixes...

mykola-elastic: will it be hard to split the changes into 2 separate PRs? or even 3 as I see 3 changelogs?

Is it OK to have several changelogs in one PR @ishleenk17 ?

packages/azure_app_service/manifest.yml
mykola-elastic: Yes, please make sure screenshots contain the whole dashboard, and also please add those same screenshots to the PR page (right now the screenshot in the PR description is truncated)

packages/azure_app_service/changelog.yml
muthu-mps: @giorgi-imerlishvili-elastic - Could we create a separate PR for the ingest pipeline changes? Since we're adding a new dashboard to the integrations, it would be better to keep these changes distinct.

giorgi-imerlishvili-elastic: Removed bugfixes from this PR and created new one: https://github.com/elastic/integrations/pull/15591

packages/azure_app_service/manifest.yml
giorgi-imerlishvili-elastic: done

mykola-elastic (approved):

muthu-mps:
@giorgi-imerlishvili-elastic - Why the `last 30 mins` filter is applied on `Log Level` panels. Do you think we can remove this filter?

giorgi-imerlishvili-elastic:
> @giorgi-imerlishvili-elastic - Why the `last 30 mins` filter is applied on `Log Level` panels. Do you think we can remove this filter?

I just saw one integration dashboard had such filter and I did same. It can be easily removed

muthu-mps:
> > @giorgi-imerlishvili-elastic - Why the `last 30 mins` filter is applied on `Log Level` panels. Do you think we can remove this filter?
> 
> I just saw one integration dashboard had such filter and I did same. It can be easily removed

We can remove this filter to show data for the entire time range selected.

lucian-ioan:
A few comments:
- Are we sure all of the data streams actually have log level? For example for IPSecAuditLogs, I don't see log level in the [documentation](https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/appserviceipsecauditlogs), nor is it in the sample events.

- It might help to include a few key fields for each data stream alongside the Activity (ex. Top client/event IP). It could make things a bit too crowded though.

WDYT @giorgi-imerlishvili-elastic, @muthu-mps ?

giorgi-imerlishvili-elastic:
> A few comments:
> 
> * Are we sure all of the data streams actually have log level? For example for IPSecAuditLogs, I don't see log level in the [documentation](https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/appserviceipsecauditlogs), nor is it in the sample events.
> * It might help to include a few key fields for each data stream alongside the Activity (ex. Top client/event IP). It could make things a bit too crowded though.
> 
> WDYT @giorgi-imerlishvili-elastic, @muthu-mps ?

Yeah I confirm, just double checked and IPSecAuditLogs and AuditLogs doesn't have log level field 😕, I'll remove panels for those

giorgi-imerlishvili-elastic:
Removed log level panels from IPSecAudit and Audit logs section and updated screenshots

Any ideas, what else information/panel can be added instead of those? @lucian-ioan @muthu-mps

muthu-mps:
Can we update the dashboard image with real time log data. The current one shows `0` in all the panels.

lucian-ioan:
> Yeah I confirm, just double checked and IPSecAuditLogs and AuditLogs doesn't have log level field 😕, I'll remove panels for those.

I believe it's also the case for [AppServiceHTTPLogs](https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/appservicehttplogs).

> Any ideas, what else information/panel can be added instead of those? @lucian-ioan @muthu-mps

The goal is to try to provide value for the users. For example for the ones without Log Level:

1. IpSecAuditLogs -> Fields: `client_ip` and `result`.  Result can only be "Allowed" or "Denied". Top IPs denied would help understand which IPs keep trying to gain access without authorization.

2. AppServiceAuditLogs -> Fields: `user` and `protocol`. Top Active Users and Top Protocol which tells the most popular auth methods (ex. SSH, OAuth2) 

3. AppServiceHTTPLogs -> Fields:  `sc_status` = HTTP response code, `time_taken` =  latency, `cs_uri_stem` = endpoint. Very useful fields here, starting from Average Latency, Status Codes Pie Chart, Top Endpoints Used all the way to aggregations such as highest average latency by endpoint.

giorgi-imerlishvili-elastic:
All feedbacks were addressed please review and let me know if I missed something @lucian-ioan, @muthu-mps, @Linu-Elias

muthu-mps:
@daniela-elastic - Can you take a look into the dashboard changes?

lucian-ioan (changes requested):
1. I'd rename any "Top 10 values of" with just the actual field title and any other raw fields with simpler names.
2. Average Response Time should have a suffix (ms) to be more readable.

ghost:
@giorgi-imerlishvili-elastic here are some dashboard guidelines you can follow https://docs.google.com/document/d/1aTqkl0BSYPa1WeKXkKCChQaIqM0J5w4vY-8flHDUj8A/edit?tab=t.0#heading=h.qt35o6gmkwy8

ishleenk17:
Reviewing from the screenshot in the PR:
1. The total errors is such elongated box and others are all v compact. Can we have a better alignment there.
 Something like this screenshot. 1 under the other ?
Screenshot 2025-10-29 at 8 35 29 AM

2. Can we have the (i) icon next to log level panels ?

> 2\. X-axis name - "per 12 hours" - can you check with @ishleenk17 if this is how we want to describe the x-axis? ALso, even though you say it's per 12 hours, i can see that you're actually showing 24 hours so it's a bit confusing

@giorgi-imerlishvili-elastic : Is your x axis the @timestamp field ? the per 12 hours, per 24 hours changes as per what time range we choose above. We usually do something like below.
Screenshot 2025-10-29 at 8 40 33 AM

giorgi-imerlishvili-elastic:
> The total errors is such elongated box and others are all v compact. Can we have a better alignment there.
Something like this screenshot. 1 under the other ?

I'll try make some changes
> @giorgi-imerlishvili-elastic : Is your x axis the @timestamp field ?  

Yes it's `@timestamp` field

muthu-mps:
> 1. I'd rename any "Top 10 values of" with just the actual field title and any other raw fields with simpler names.
> 2. Average Response Time should have a suffix (ms) to be more readable.

@lucian-ioan - If this comment is addressed, Can you approve it?

ishleenk17 (approved):
Log level panel doesn't have mention of @timestamp field like other panels/ OTherwsise looks good

PR #15491, opened 2025-09-29
Proxysg update bytes fields

jrmolin:
## Description

The pipeline was keeping the client-to-server and server-to-client bytes counts in ECS as `keyword`, which is not a very efficient use of the field. The pipeline already converted the data into integer values and stored them in `server.bytes` and `client.bytes`, which is valid ECS and the preferred way of keeping the data. 

This PR simply removes the old fields.

## Proposed commit message

[ProxySG] Remove duplicate and incorrect byte-count fields
- Remove the `proxysg.client_to_server.bytes` field
- Remove the `proxysg.server_to_client.bytes` field
- Ensure the ECS field descriptions are updated
- No tests fail

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


taylor-swanson:
Should we be removing these fields? I doubt we use them in rules, but a user might. I wonder if it would be better to just leave them in there.

If we decide to remove it, this would constitute a breaking change and a major version bump. We may also have to engage with the breaking change committee as well.

jrmolin:
i don't have to remove the fields. it's the only actual change. the "bug" doesn't make sense, since we're already outputting the fields correctly in ECS. either i leave them (don't change to `long`) or remove them.

PR #15498, opened 2025-09-30
Github: update audit dashboards

chemamartinez:
## Proposed commit message

Audit dashboards in Github integration didn't have any filters so controls showed not only fields related to `github.audit` but also any possible value for fields like `user.name` or `event.action` even they belong to different indexes.

Added the filter `data_stream.dataset == github.audit` to both dashboards.

Also replaced deprecated controls panel in both dashboards.

## Checklist

## Author's Checklist

- Tested new installation in oldest supported 8.17.1
- Tested new installation in newest supported version 9.1.3
- Tested integration upgrade in 8.17.1

In all tests, checked that dashboards show data properly and new filters and controls work.

### Before

GitHub-Audit-Log-Activity-Elastic-09-30-2025_09_59_AM

GitHub-User-Change-Audit-Elastic-09-30-2025_10_00_AM

### After

github-audit-dashboard

github-user-dashboard

### DIFF
--- a/packages/github/changelog.yml
+++ b/packages/github/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Update deprecated controls and filters in Audit dashboards.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/99999
+      link: https://github.com/elastic/integrations/pull/15498
 - version: "2.15.1"
   changes:
     - description: Added missing oauth2 toggle in audit data stream.


efd6 (commented):
Rather than having the filter on the dashboard, can we have the filtering done at the visualisation level? This prevents users from accidentally removing it.

chemamartinez:
> Rather than having the filter on the dashboard, can we have the filtering done at the visualisation level? This prevents users from accidentally removing it.

Cool. I'll make the change, although I will have to do the same for all dashboards of the integration for consistency.

@efd6 what do you think about this? in managed dashboards where nothing is editable for users, why predefined filters can be removed? if these filters are needed for the dashboard to work fine, and can be accidentally removed with just one click, shouldn't they also be immutable?

efd6:
> if these filters are needed for the dashboard to work fine, and can be accidentally removed with just one click, shouldn't they also be immutable?

I'm not sure how that would be achievable while still allowing the users to add their own refining filters.

chemamartinez:
> I'm not sure how that would be achievable while still allowing the users to add their own refining filters.

I think having two kind of filters, the default ones managed by Elastic where typically set the data stream corresponding to the dashboard, and user filters.

Anyway, there is a problem with removing the filters on the dashboard. When clicking in controls they don't filter by values from events of the data stream, they show every possible value of that field in the data view (typically `logs-*`).

That is what the SDH is originally about so I don't think we could remove them. Controls cannot add filters like visualisations.

I've checked and every visualisation has their own filter so if users remove any global filter by mistake it would still work fine.

PR #15524, opened 2025-10-03
[arista_ngfw] Generate processor tags and normalize error handler

taylor-swanson:
## Proposed commit message

## Checklist

### DIFF
--- a/packages/arista_ngfw/data_stream/log/elasticsearch/ingest_pipeline/admin_login.yml
+++ b/packages/arista_ngfw/data_stream/log/elasticsearch/ingest_pipeline/admin_login.yml
@@ -8,14 +8,17 @@ processors:
       field: arista.reason
       target_field: event.reason
       ignore_missing: true
+      tag: rename_07f5af7a
   - rename:
       field: arista.login
       target_field: user.name
       ignore_missing: true
+      tag: rename_3b52aae3
   - rename:
       field: arista.clientAddress
       target_field: source.ip
       ignore_missing: true
+      tag: rename_1f127b6f
   - script:
       lang: painless
       params:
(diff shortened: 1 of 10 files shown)


packages/arista_ngfw/data_stream/log/elasticsearch/ingest_pipeline/admin_login.yml
```diff
@@ -53,7 +56,7 @@ processors:
         ctx.event.type = params.get((ctx.arista.succeeded).toString()).get('type').clone();
         ctx.event.provider = params.get((ctx.arista.succeeded).toString()).get('provider');
         ctx.arista.remove('succeeded');
-  
+      tag: script_cef61fad
```
qcorporation: I still think we need a space after for the comment

taylor-swanson:
> In general I like it - the tag however after the script processors are a bit hard to read due to spacing, it is at all possible to add them at the top of the processor instead of at the end? would that complicate things greatly?

Not sure, but I'll look into it.

I'm putting this one back into draft, I forgot to run the duplicate tag checker and 6 of the generated tags are duplicates (likely because the processors themselves were repeated). Either way, looking into that as well.

Edit: these are the 6 duplicated tags, all resulting in duplicated processors.

```
duplicate tag "convert_b57304ca" found in processor
duplicate tag "convert_0c5a9349" found in processor
duplicate tag "geoip_da2e41b2" found in processor
duplicate tag "geoip_28d69883" found in processor
duplicate tag "rename_a917047d" found in processor
duplicate tag "rename_f1362d0b" found in processor
```

I initially had the index of the processor influencing the hash, but I removed it. I'm glad I did, because it's finding these duplicated processors.

PR #15540, opened 2025-10-03
[fortinet_fortigate] Generate processor tags and normalize error handler

taylor-swanson:
## Proposed commit message

## Checklist

### DIFF
--- a/packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/default.yml
@@ -4,17 +4,21 @@ processors:
   - set:
       field: ecs.version
       value: '8.17.0'
+      tag: set_f5923549
   - set:
       field: event.original
       copy_from: message
+      tag: set_b9758751
   - remove:
       field: message
+      tag: remove_7144efd2
   - grok:
       field: event.original
       ecs_compatibility: v1
       patterns:
         - "^(?:%{SYSLOG5424PRI}%{NONNEGINT} )+(?:%{TIMESTAMP_ISO8601}|-) +(?:%{HOSTNAME:syslog5424_host}|-) +(-|%{SYSLOG5424PRINTASCII}) +(-|%{SYSLOG5424PRINTASCII}) +(-|%{SYSLOG5424PRINTASCII}) +(-|%{SYSLOG5424PRINTASCII}) +(?:%{GREEDYDATA:syslog5424_msg}|-|)"
         - "^(?:%{SYSLOG5424PRI} *)?%{GREEDYDATA:syslog5424_msg}$"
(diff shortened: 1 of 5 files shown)


packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -563,89 +650,95 @@ processors:
...
       value: "{{{source.nat.ip}}}"
       if: ctx.source?.nat?.ip != null
       allow_duplicates: false
+      tag: append_53b62ed8
   - append:
       field: related.ip
       value: "{{{destination.nat.ip}}}"
       if: ctx.destination?.nat?.ip != null
       allow_duplicates: false
-  - append:
```
haetamoudi, line 594: out of curiosity, would this duplication set the field twice? can this trigger an error because of `allow_duplicates: false`?

taylor-swanson: No, because `allow_duplicates` is false, it won't set it twice. As far as I know, it shouldn't show up as an error (but then again I don't know, as ignore_failure still yields an increase in the error count 🤷‍♂️)

The issue here is that the processor does have to actually check every element of related.ip, do a string comparison, and fail out when it finds the existing value in there already.

packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -106,45 +114,55 @@ processors:
...
       field: _temp.time
       value: "{{{fortinet.firewall.date}}} {{{fortinet.firewall.time}}} {{{event.timezone}}}"
       if: ctx.fortinet?.firewall?.date != null && ctx.fortinet?.firewall?.time != null && ctx.event?.timezone != null
   - set:
+      tag: set__temp_time_41a88693
       field: _temp.time
       value: "{{{fortinet.firewall.date}}} {{{fortinet.firewall.time}}}"
       if: ctx.fortinet?.firewall?.date != null && ctx.fortinet?.firewall?.time != null && ctx.event?.timezone == null
   - date:
+      tag: date__temp_time_to_@timestamp_745b440f
```
haetamoudi: This one has an `@`

packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -155,6 +173,7 @@ processors:
       timezone: "{{{event.timezone}}}"
       if: ctx._temp?.time != null && ctx.event?.timezone != null
   - date:
+      tag: date__temp_time_to_@timestamp_978d46f3
```
haetamoudi: This one has an `@` too

PR #15563, opened 2025-10-06
[modsecurity] Generate processor tags and normalize error handler

taylor-swanson:
## Proposed commit message

- Ran elastic-package format

## Checklist

### DIFF
--- a/packages/modsecurity/data_stream/auditlog/elasticsearch/ingest_pipeline/apache-modsec.yml
+++ b/packages/modsecurity/data_stream/auditlog/elasticsearch/ingest_pipeline/apache-modsec.yml
@@ -5,6 +5,7 @@ processors:
       field: json.transaction.time
       target_field: _temps.date
       ignore_missing: true
+      tag: rename_3f87beb3
   - grok:
       field: _temps.date
       patterns:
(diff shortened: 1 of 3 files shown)


packages/modsecurity/data_stream/auditlog/elasticsearch/ingest_pipeline/apache-modsec.yml
```diff
@@ -66,98 +75,122 @@ processors:
...
       field: json.transaction.remote_port
       target_field: source.port
       ignore_missing: true
   - grok:
+      tag: grok_json_request_request_line_e2746bce
       field: json.request.request_line
       patterns:
         - "%{NOTSPACE:http.request.method} %{URIPATHPARAM:url.original}(?: HTTP/%{NUMBER:http.version})"
   - rename:
+      tag: rename_json_transaction_request_headers_host_to_json_transaction_request_headers_Host_50a00924
```
bhapas, line 107: Just curious.. Hope there is no limitation on the tag character length?

taylor-swanson: That's a good question... I don't think there is? Not aware of anything in elasticsearch itself (aside from maybe the default limit we put on fields, I can't recall what it was but I think it's 1024 or higher). I don't think BigQuery will impose any limits either. Data types should be variable length, and don't think we impose limits there.

taylor-swanson: @haetamoudi, are you aware of limits in BigQuery or other places?

haetamoudi: should be fine. Here is a valid tag name that is already present in stats tables in bigquery https://github.com/elastic/integrations/blob/2b577707165221ed6215d2ee77ad96fbf2cfbbbf/packages/watchguard_firebox/data_stream/log/elasticsearch/ingest_pipeline/pipeline_diagnostic.yml#L642

PR #15601, opened 2025-10-08
aws.cloudtrail: improve CloudTrail user identity processing

chemamartinez:
## Proposed commit message

For CloudTrail events, it has been updated how IAM users are handled.

In particular, for the user identity `IAMUser` type, the `user.name` and `user.id` are
populated with the user fields that made the action/request.

For the user identity `AssumedRole` type, AWS SIEM rules need roles to be treated as IAMUsers in order to work fine. So the role identifies inside `sessionIssuer` populate `user.*` fields. Then, the session name is being mapped as `user.changes.name` as it can be interpreted as the name the user is taking for that particular session, and it's the closest approach in ECS.

References:
- https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-user-identity.html
- https://www.elastic.co/docs/reference/ecs/ecs-user-usage#ecs-user-usage-combining

## Checklist

## Related issues

- Relates https://github.com/elastic/enhancements/issues/25241

### DIFF
--- a/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-assume-role-json.log-expected.json
+++ b/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-assume-role-json.log-expected.json
@@ -68,6 +68,7 @@
                             "session_issuer": {
                                 "account_id": "111111111111",
                                 "arn": "arn:aws:iam::111111111111:role/JohnRole1",
+                                "principal_id": "AROAIN5ATK5U7KEXAMPLE",
                                 "type": "Role"
                             }
                         },
@@ -249,6 +250,7 @@
                             "session_issuer": {
                                 "account_id": "111111111111",
                                 "arn": "arn:aws:iam::111111111111:role/JohnRole1",
+                                "principal_id": "AROAIN5ATK5U7KEXAMPLE",
                                 "type": "Role"
                             }
                         },
--- a/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-converse-json.log-expected.json
+++ b/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-converse-json.log-expected.json
@@ -32,6 +32,7 @@
                             "session_issuer": {
                                 "account_id": "00000000000",
                                 "arn": "arn:aws:iam::00000000000:role/private-ec2-instance-role",
+                                "principal_id": "PRINCIPALID",
                                 "type": "Role"
                             }
                         },
--- a/packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
@@ -1072,11 +1072,9 @@ processors:
       tag: rename_user_name_assumed_role
   - rename:
       field: json.userIdentity.sessionContext.sessionIssuer.principalId
-      target_field: user.id
+      target_field: aws.cloudtrail.user_identity.session_context.session_issuer.principal_id
       ignore_missing: true
-      override: true
-      if: ctx.user?.id == null && (ctx.aws?.cloudtrail?.user_identity?.type == 'AssumedRole' || ctx.aws?.cloudtrail?.user_identity?.type == 'FederatedUser')
-      tag: rename_user_id_assumed_role
+      ignore_failure: true
   - grok:
       field: aws.cloudtrail.user_identity.arn
       patterns:
--- a/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-attach-user-policy-json.log-expected.json
+++ b/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-attach-user-policy-json.log-expected.json
@@ -33,6 +33,7 @@
                             "session_issuer": {
                                 "account_id": "000000000",
                                 "arn": "arn:aws:iam::000000000:role/ec2-instance-role",
+                                "principal_id": "PRINCIPALID",
                                 "type": "Role"
                             }
                         },
--- a/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-authorize-security-group-egress-json.log-expected.json
+++ b/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-authorize-security-group-egress-json.log-expected.json
@@ -68,6 +68,7 @@
                             "session_issuer": {
                                 "account_id": "000000000",
                                 "arn": "arn:aws:iam::000000000:role/ec2-instance-role",
+                                "principal_id": "PRINCIPALID",
                                 "type": "Role"
                             }
                         },
--- a/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-console-login-json.log-expected.json
+++ b/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-console-login-json.log-expected.json
@@ -281,6 +281,7 @@
                             "session_issuer": {
                                 "account_id": "123456789012",
                                 "arn": "arn:aws:iam::123456789012:role/RoleToBeAssumed",
+                                "principal_id": "AROAIDPPEZS35WEXAMPLE",
                                 "type": "Role"
                             }
                         },
--- a/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-create-control-channel-json.log-expected.json
+++ b/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-create-control-channel-json.log-expected.json
@@ -35,6 +35,7 @@
                             "session_issuer": {
                                 "account_id": "00000000000",
                                 "arn": "arn:aws:iam::00000000000:role/bedrock_ec2_role",
+                                "principal_id": "PRINCIPALID",
                                 "type": "Role"
                             }
                         },
--- a/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-delete-bucket-json.log-expected.json
+++ b/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-delete-bucket-json.log-expected.json
@@ -30,6 +30,7 @@
                             "session_issuer": {
                                 "account_id": "777788889999",
                                 "arn": "arn:aws:iam::777788889999:role/AssumeNothing",
+                                "principal_id": "AIDAQRSTUVWXYZEXAMPLE",
                                 "type": "Role"
                             }
                         },
--- a/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-get-bucket-policy-json.log-expected.json
+++ b/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-get-bucket-policy-json.log-expected.json
@@ -50,6 +50,7 @@
                             "session_issuer": {
                                 "account_id": "0000000000",
                                 "arn": "arn:aws:iam::0000000000:role/aws-service-role/config.amazonaws.com/AWSServiceRoleForConfig",
+                                "principal_id": "PRINCIPALID",
                                 "type": "Role"
                             }
                         },
--- a/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-get-policy-json.log-expected.json
+++ b/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-get-policy-json.log-expected.json
@@ -32,6 +32,7 @@
                             "session_issuer": {
                                 "account_id": "00000000000",
                                 "arn": "arn:aws:iam::00000000000:role/Elastic-Cloud-Security-Posture-Man-ElasticAgentRole-EdUmKXybQxe9",
+                                "principal_id": "PRINCIPALID",
                                 "type": "Role"
                             }
                         },
--- a/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-open-control-channel-json.log-expected.json
+++ b/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-open-control-channel-json.log-expected.json
@@ -45,6 +45,7 @@
                             "session_issuer": {
                                 "account_id": "000000000000",
                                 "arn": "arn:aws:iam::000000000000:role/ec2-instance-role",
+                                "principal_id": "PRINCIPALID",
                                 "type": "Role"
                             }
                         },
--- a/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-publish-json.log-expected.json
+++ b/packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-publish-json.log-expected.json
@@ -44,6 +44,7 @@
                             "session_issuer": {
                                 "account_id": "00000000000",
                                 "arn": "arn:aws:iam::00000000000:role/private-ec2-instance-role",
+                                "principal_id": "PRINCIPALID",
                                 "type": "Role"
                             }
                         },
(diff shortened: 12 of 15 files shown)


packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1734,6 +1756,13 @@ processors:
       allow_duplicates: false
       ignore_failure: true
 
+  - append:
+      field: related.user
+      value: '{{{user.changes.name}}}'
```
Kavindu-Dodan, line 2012: nit : Given we extract this from sessionContext arn with following description,

> arn
The Amazon Resource Name (ARN) of the principal that made the call. The last section of the arn contains the user or role that made the call.

Wouldn't a name like `user.arn.principal` better match this?

Kavindu-Dodan: Anyway not a blocker to get this merged

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-assume-role-json.log-expected.json
```diff
@@ -112,8 +111,9 @@
                     "arn:aws:iam::111111111111:role/JohnRole2"
                 ],
                 "user": [
-                    "AROAIN5ATK5U7KEXAMPLE:JohnRole1",
-                    "JohnDoe"
+                    "AROAIN5ATK5U7KEXAMPLE",
+                    "JohnDoe",
+                    "JohnRole1"
```
efd6, line 117: It seems odd to me to map a role as a user.

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-converse-json.log-expected.json
```diff
@@ -70,8 +69,9 @@
                     "arn:aws:sts::00000000000:assumed-role/private-ec2-instance-role/i-03cd6b2a7eb4bf3ae"
                 ],
                 "user": [
-                    "PRINCIPALID:i-03cd6b2a7eb4bf3ae",
-                    "private-ec2-instance-role"
+                    "PRINCIPALID",
+                    "private-ec2-instance-role",
+                    "i-03cd6b2a7eb4bf3ae"
```
efd6, line 82: This seems more odd, since this is a session now.

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
chemamartinez: We are trying to stick to current ECS available fields, unfortunately `user.arn` doesn't exist in ECS.

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-assume-role-json.log-expected.json
chemamartinez: This come from [this conversation](https://github.com/elastic/enhancements/issues/25241#issuecomment-3410632346) with the trade team, where they need to treat the IAM user name and the role name as the `user.name` for the AWS SIEM rules.

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-converse-json.log-expected.json
chemamartinez: From this [same conversation](https://github.com/elastic/enhancements/issues/25241#issuecomment-3427426240), the goal is to consider the session name as temporary user name, that's why it is being mapped as `user.changes.name` and therefore included into `related.users`.

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-assume-role-json.log-expected.json
imays11: @efd6 A lot of our cloud based detection rules are anomaly based, so doing it this way allows us to track the overall behavior of a role across many different temporary sessions in the same way that we track a user's behavior over time. Here is an [example rule](https://github.com/elastic/detection-rules/blob/b9b8e245145af979fbe7296f5907eb4586a2d6a8/rules/integrations/aws/privilege_escalation_iam_update_assume_role_policy.toml#L13) that uses the user.name field to determine the first time a User or a Role updates an existing AssumeRole Policy, which could be a privilege escalation attempt. We want to examine the history of the Role's behavior in the same way we would an IAMUser identity.

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-create-db-instance-json.log-expected.json
```diff
@@ -174,6 +174,9 @@
                     "subnet-fee506df",
                     "subnet-bf6ab5b1",
                     "subnet-8bdf6bc6"
```
imays11, line 182: I would add the `"identityStoreArn\":\"arn:aws:identitystore::123456789012:identitystore/d-9067642ac7\"` value here under `related.entity`

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-create-db-instance-json.log-expected.json
```diff
@@ -174,6 +174,9 @@
...
+                ],
+                "user": [
+                    "544894e8-80c1-707f-60e3-3ba6510dfac1"
```
imays11, line 185: I would add the identity store id value `d-9067642ac7`, which is attached to the `identityStoreArn`. This value in conjunction with the userId value `544894e8-80c1-707f-60e3-3ba6510dfac1`, are both necessary for running a separate query against the Identity Center API to [determine the user name](https://docs.amazonaws.cn/en_us/singlesignon/latest/userguide/sso-cloudtrail-use-cases.html#user-session-iam-identity-center) behind the API calls. We should provide both values easily in our logs.

I'd also consider adding the `"credentialId\":\"EXAMPLEVHULjJdTUdPJfofVa1sufHDoj7aYcOYcxFVllWR_Whr1fEXAMPLE\"` value here as well. It can be used to track all behavior taking place within a [single authenticated session](https://docs.amazonaws.cn/en_us/singlesignon/latest/userguide/sso-cloudtrail-use-cases.html#correlating-users-same-session). Similar to the session-name for an AssumedRole identity type.

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-user-authentication.log-expected.json
```diff
@@ -58,6 +58,7 @@
             "related": {
                 "entity": [],
```
imays11: Again, I'd add the `"identityStoreArn\":\"arn:aws:identitystore::redacted:identitystore/redacted\"` value here

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -815,11 +818,30 @@ processors:
   - rename:
       field: json.userIdentity.sessionContext.sessionIssuer.userName
       target_field: user.name
-      ignore_failure: true
+      ignore_missing: true
+      override: true
+      if: ctx.aws?.cloudtrail?.user_identity?.type == 'AssumedRole'
```
imays11: This is going to prevent proper mapping for `FederatedUser` type, which also provides the `userIdentity.sessionContext.sessionIssuer.userName` field that needs to continue to be mapped to `user.name` field

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -815,11 +818,30 @@ processors:
...
+      tag: rename_user_name_assumed_role
   - rename:
       field: json.userIdentity.sessionContext.sessionIssuer.principalId
-      target_field: aws.cloudtrail.user_identity.session_context.session_issuer.principal_id
-      ignore_failure: true
+      target_field: user.id
+      ignore_missing: true
+      override: true
+      if: ctx.aws?.cloudtrail?.user_identity?.type == 'AssumedRole'
```
imays11: Same here, this change should apply to `FederateUser` type as well

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -815,11 +818,30 @@ processors:
...
+      tag: rename_user_id_assumed_role
+  - grok:
+      field: aws.cloudtrail.user_identity.arn
+      patterns:
+        - "arn:aws:sts:.*/%{GREEDYDATA:user.changes.name}$"
+      ignore_missing: true
+      if: ctx.aws?.cloudtrail?.user_identity?.type == 'AssumedRole' && ctx.aws?.cloudtrail?.user_identity?.arn != null
```
imays11: We should use this to also map the  `user.changes.name` value for `FederatedUser` identity type. Example: `aws.cloudtrail.user_identity.arn : "arn:aws:sts::123445667789:federated-user/fed_consoler"` . The arn is created a little differently from Assumed Role since it doesn't keep the calling user's name in the arn but `fed_consoler` is the name change that should be extracted and treated just like we are treated the session.name for AssumedRole identity type.

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-assume-role-json.log-expected.json
efd6: Maybe the rules could more closely follow the ECS definitions of the fields? Confounding the semantics confuses things for the future.

imays11: The rules have to use the fields that best allow us to detect the behavior, regardless of what the ECS definition of that field is. The problem really is that ECS wasn't created with cloud identities in mind so the real solution would be adding to ECS, right now we're just trying to get as close as we can to the current definitions.

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-create-db-instance-json.log-expected.json
chemamartinez: I have a question with adding the identity store ID into the `related.user` field. That could be confusing if we don't know exactly what kind of user field is, and if we don't map it first with that ECS user field.

If it also an entity, maybe it better fits into `actor.entity.id` or just `related.entity`?

chemamartinez: > Similar to the session-name for an AssumedRole identity type.

Does this mean you would also map this as `user.changes.id`?

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-assume-role-json.log-expected.json
chemamartinez: I agree ECS fields should be extended for these cloud use cases.

packages/aws/data_stream/cloudtrail/_dev/test/pipeline/test-create-db-instance-json.log-expected.json
imays11: The `store id value` doesn't change, it remains associated with the same `userId`. Which is why both values can be used to look up a user name from the Identity Store. So I would say adding it to `related.entity` or `related.user` is probably most appropriate here.  

The one that changes per session is the `credentialId`. Mapping it to `user.changes.id` makes the most sense given the fields we have to choose from for cloud-specific things like this.

chemamartinez:
@imays11 @Mikaayenson I have a question for you, the latest changes to the change request respond to Isai's suggestions to avoid breaking the existing rules, so role names are still mapped as `user.name` and the only change would be to assign the session name to `user.changes.name`, which I don't think should be a problem for the rules. Correct me if I'm wrong.

Do you still agree with these changes? I understand that the most painful change for you would be to separate roles and user names into different fields. If so, assigning `some-role` to `user.roles` would break the rules, right?

imays11:
@chemamartinez 
> @imays11 @Mikaayenson I have a question for you, the latest changes to the change request respond to Isai's suggestions to avoid breaking the existing rules, so role names are still mapped as `user.name` and the only change would be to assign the session name to `user.changes.name`, which I don't think should be a problem for the rules. Correct me if I'm wrong.

Yes I still agree with those changes

> Do you still agree with these changes? I understand that the most painful change for you would be to separate roles and user names into different fields. If so, assigning `some-role` to `user.roles` would break the rules, right?

Yes that is correct, if the `some-role` > `user.name` mapping is removed that would require us to duplicate a set of new terms rules to utilize the new field mapping for roles.

imays11:
@chemamartinez 
Have you considered using `cloud.entity.name` as a field that can be used for both IAM Users and Roles? 
https://www.elastic.co/docs/reference/ecs/ecs-cloud#ecs-cloud-nestings
https://www.elastic.co/docs/reference/ecs/ecs-entity#field-entity-name

chemamartinez:
> Yes I still agree with those changes

In that case, if no more changes are required the PR is ready for review I think.

> Yes that is correct, if the some-role > user.name mapping is removed that would require us to duplicate a set of new terms rules to utilize the new field mapping for roles.
> Have you considered using cloud.entity.name as a field that can be used for both IAM Users and Roles?

@imays11 aren't these two comments contradictory? If we now map the fields to `cloud.entity.name` instead of keeping using `user.name` all rules should be rewritten, isn't it?

imays11:
> > Yes that is correct, if the some-role > user.name mapping is removed that would require us to duplicate a set of new terms rules to utilize the new field mapping for roles.
> > Have you considered using cloud.entity.name as a field that can be used for both IAM Users and Roles?
> 
> @imays11 aren't these two comments contradictory? If we now map the fields to `cloud.entity.name` instead of keeping using `user.name` all rules should be rewritten, isn't it?

@chemamartinez 
No, as long as both user identity types are under the same field (`user.name` OR `cloud.entity.name`), then we can simply do a rule tuning of the existing rules and replace the `user.name` field to use `cloud.entity.name` field instead. 

The problem comes with keeping IAM Users under `user.name` but moving Roles to a different field like `user.roles`. That would require us to create a duplicate set of rules in order to utilize 2 different fields. 

I'm fine keeping both under `user.name`, I only gave the suggestion because I know there was some concern of "long-term semantic weakening of an ECS field definition" by keeping Roles under `user.name` field.

chemamartinez:
@imays11 thanks for the clarification!

efd6:
Am I correct in understanding that this change is still going to conflate roles with users? If that's the case, I still do not agree, but I am not going to block this.

kcreddy:
> @chemamartinez Have you considered using `cloud.entity.name` as a field that can be used for both IAM Users and Roles? https://www.elastic.co/docs/reference/ecs/ecs-cloud#ecs-cloud-nestings https://www.elastic.co/docs/reference/ecs/ecs-entity#field-entity-name

One problem is that the field [`entity.name`](https://www.elastic.co/docs/reference/ecs/ecs-entity#field-entity-name) is in _Beta_ and may not be ideal for GA integration. Also, the field is introduced in ECS version `9.1` while `aws` integration uses [`8.11`](https://github.com/elastic/integrations/blob/main/packages/aws/_dev/build/build.yml). Changing it to `9.1` might require approvals from all [owners](https://github.com/elastic/integrations/blob/main/.github/CODEOWNERS#L35) of aws integration.

chemamartinez:
> Am I correct in understanding that this change is still going to conflate roles with users? If that's the case, I still do not agree, but I am not going to block this.

@efd6 yes, but this is not something in this PR. It is something it was happening before so the SIEM rules work.

agithomas (approved):
LGTM, considering the design approach taken! Code owner approval.

PR #15650, opened 2025-10-14
[system] [elastic_agent] Add existing README as KB doc for system + elastic_agent integrations

kpollich:
## Proposed commit message

Make existing README available as a knowledge base document for Elastic AI assistants.

## Checklist

### DIFF
--- a/packages/elastic_agent/changelog.yml
+++ b/packages/elastic_agent/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.6.4"
+  changes:
+    - description: Make existing README available as AI assistant knowledge
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/15650
 - version: "2.6.3"
   changes:
     - description: Elastic Agent memory charts now prioritise RSS memory for more accurate usage reporting.
--- a/packages/elastic_agent/manifest.yml
+++ b/packages/elastic_agent/manifest.yml
@@ -1,6 +1,6 @@
 name: elastic_agent
 title: Elastic Agent
-version: 2.6.3
+version: 2.6.4
 description: Collect logs and metrics from Elastic Agents.
 type: integration
 format_version: 3.1.4
--- a/packages/system/changelog.yml
+++ b/packages/system/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.6.2"
+  changes:
+    - description: Make existing README available as AI assistant knowledge
+      type: enhancement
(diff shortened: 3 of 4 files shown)


jen-huang:
I did something similar for testing, but coping the exact READMEs does place additional burden on package authors to remember to look for and update them in two places. Not sure how folks feel about that...

It would be interesting if Fleet treated the base READMEs as additional KB content that is automatically indexed, I remember we discussed this during design of KB content but don't remember if we deferred or decided to not do that.

ishleenk17:
> I did something similar for testing, but coping the exact READMEs does place additional burden on package authors to remember to look for and update them in two places. Not sure how folks feel about that...
> 
> It would be interesting if Fleet treated the base READMEs as additional KB content that is automatically indexed, I remember we discussed this during design of KB content but don't remember if we deferred or decided to not do that.

I tend to agree with @jen-huang on this @kpollich, since the KB README's are manually duplicated and not generated as part of the build.
Could you please share what is the motive for this change and why can't the current README'e be referred to ?

jen-huang:
> I tend to agree with @jen-huang on this @kpollich, since the KB README's are manually duplicated and not generated as part of the build.
> 
> Could you please share what is the motive for this change and why can't the current README'e be referred to ?

Fleet currently looks specifically in this `/docs/knowledge` directory for content to index for consumption by AI assistants. It will be additional work (but not a ton) to also index the base README.

But actually I like the idea of automatically duplicating the base README to that directory during the package build process. That would require no additional Fleet changes or work for package developers. We would just have to come to the consensus that all base READMEs are useful for the AI assistants to consume.

nimarezainia:
@jen-huang I think at this point in time all the READMEs should be copied into the `knowledge` directory to give us an immediate content for AI to noodle on. We can always refine/remove that if better content becomes available.

ishleenk17:
> @jen-huang I think at this point in time all the READMEs should be copied into the `knowledge` directory to give us an immediate content for AI to noodle on. We can always refine/remove that if better content becomes available.

@nimarezainia : I understand we're planning to roll this out across all Integrations. 

Given that Jen mentioned [here](https://github.com/elastic/integrations/pull/15650#issuecomment-3404626884) that it's relatively straightforward, could we incorporate this README generation into the elastic-package build process? Since we update Integrations and their READMEs frequently, manually maintaining these could quickly become outdated without automated generation."

jen-huang:
@kpollich I'm going to close this in favor of https://github.com/elastic/ingest-dev/issues/6280 per discussion here and on the main meta issue.

PR #15651, opened 2025-10-14
NGINX OpenTelemetry Input Package

ishleenk17:
This is a draft version of NGINX OpenTelemetry Input Package. 
Since this is the very first OTEL Input package, this PR is being used to figure out the structure, template, what works what doesn't while creating packages. 

FYI, using [HTTPChecker](https://github.com/elastic/integrations/pull/14315/) example as reference 

Relates: https://github.com/elastic/obs-integration-team/issues/622

### DIFF
--- a/packages/nginx_input_otel/manifest.yml
+++ b/packages/nginx_input_otel/manifest.yml
@@ -24,7 +24,7 @@ policy_templates:
   - name: nginxreceiver
     type: metrics
     title: NGINX OpenTelemetry Input
-    description: Collect NGINX metrics using OpenTelemetry
+    description: Collect NGINX metrics using OpenTelemetry Collector
     input: otelcol
     template_path: input.yml.hbs
     vars:


ishleenk17:
@jsoriano @cmacknz : Could you please review this first OTEL input package 

### Point to discuss:

1. We need to have a check in data_stream.dataset. It should always be same as the receiver name (for metrics) for the content package to autoinstall. For logs we need to see how will we ensure about the datastream field.
2. Under compatibility section of documentation, what all should we mention or shall we just point to the upstream receiver ? I think pointing to upstream receiver might not be right as our EDOT collector might not be at that the upstream collector label ?
3. Ideally we don't need base-fileds.yml. But elastic package doesn't allow to build the package without that. We should remove this check for these fields in OTEL Input package. 
4. If we are updating this, we should also update the corresponding content package documentation. We need to discuss how. Currently it contains how a collecotr config looks like. But since now that info is there in the input package. Should we have that in content package ? We might need it if someone tries to install that content package via upstream collector and not EDOT. Thoughts ?

cc: @tommyers-elastic

packages/nginx_input_otel/_dev/build/docs/README.md
```diff
@@ -0,0 +1,63 @@
+# NGINX OpenTelemetry Input Package
```
tommyers-elastic, line 1: i wonder how much of this we should just delegate to the upstream receiver. i.e. i could see an argument for literally just something like 

> An Elastic Agent wrapper around the [nginxreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/nginxreceiver). Compatible with the [Nginx Content Pack](https://github.com/elastic/integrations/tree/main/packages/nginx_otel).

packages/nginx_input_otel/agent/input/input.yml.hbs
```diff
@@ -0,0 +1,12 @@
+receivers:
+  nginx:
+    endpoint: {{nginx_endpoint}}
+    collection_interval: {{period}}
+processors:
```
tommyers-elastic, line 4: it would be great if agent could provide more flexibility for resource detection processing. i.e it could be a setting at the agent policy level. `system` seems a sensible default, but `ec2` looks out of place to me here.

tommyers-elastic: additionally, do you know if agent injects any batching or memory limiter processor configs?

packages/nginx_input_otel/fields/base-fields.yml
```diff
@@ -0,0 +1,12 @@
+- name: data_stream.type
```
tommyers-elastic: do the ECS dynamic mappings not get applied to OTel data streams? 

(i did see your comment about needing this file to get a green validation, which i agree we should resolve, but also wondered about whether these fields get mapped by default for OTel data streams)

packages/nginx_input_otel/img/sample-logo.svg
tommyers-elastic, line 1: shouldn't ths be the nginx logo?

packages/nginx_input_otel/manifest.yml
```diff
@@ -0,0 +1,54 @@
...
+    input: otelcol
+    template_path: input.yml.hbs
+    vars:
+      - name: nginx_endpoint
+        type: text
+        title: NGINX Endpoint
+        description: The endpoint for NGINX status page to collect metrics
+        required: true
+        show_user: true
+        default: http://localhost:8080/status
```
tommyers-elastic: nginx default port is `80`, how come we default to `8080`?

packages/nginx_input_otel/img/sample-logo.svg
ishleenk17: I was thinking about generating NGINX+OTEL logo. The way we have been doing for content packs.
WDYT ?

packages/nginx_input_otel/agent/input/input.yml.hbs
ishleenk17: Yes, I would expect the RDP processor to be a default one by Fleet at policy level. 

> additionally, do you know if agent injects any batching or memory limiter processor configs?

Nothing that I see in the agent policy that is created. @jsoriano is there any ?

packages/nginx_input_otel/_dev/build/docs/README.md
ishleenk17: I am open to discussions on this one.

packages/nginx_input_otel/fields/base-fields.yml
ishleenk17: The otel@mappings component template used in the metrics-otel@template has these mappings.

Screenshot 2025-10-27 at 5 46 23 PM

packages/nginx_input_otel/agent/input/input.yml.hbs
jsoriano: > it would be great if agent could provide more flexibility for resource detection processing. i.e it could be a setting at the agent policy level. system seems a sensible default, but ec2 looks out of place to me here.

Do you think most packages are going to have this `resourcedetection` processor? If that's the case this should be probably managed by Fleet, and not included in all packages.

> > additionally, do you know if agent injects any batching or memory limiter processor configs?
> 
> Nothing that I see in the agent policy that is created. [@jsoriano](https://github.com/jsoriano) is there any ?

Not that I know.

packages/nginx_input_otel/fields/base-fields.yml
jsoriano, line 1: This file can be removed. OTel input packages rely on pre-built component templates.

If this is the only fields file, build will fail, till we release https://github.com/elastic/package-spec/pull/994 (probably this week).

packages/nginx_input_otel/manifest.yml
```diff
@@ -0,0 +1,54 @@
...
+        default: http://localhost:80/status
+      - name: period
+        type: text
+        title: Collection Interval
+        description: How often to collect metrics
+        default: 10s
+elasticsearch:
+  index_template:
+    mappings:
+      subobjects: false
```
jsoriano: This is not compatible with pre-build mappings for OTel. I am surprised to see tests passing 🤔
```suggestion
```

packages/nginx_input_otel/manifest.yml
```diff
@@ -0,0 +1,54 @@
...
+    title: Sample screenshot
+    size: 600x600
+    type: image/png
+icons:
+  - src: /img/sample-logo.svg
+    title: Sample logo
+    size: 32x32
+    type: image/svg+xml
+policy_templates:
+  - name: nginx_otel_input
```
jsoriano: This is used as the default dataset by Fleet, and should be the receiver name, so probably `nginxreceiver`. We plan to automatically check this in the hopefuly near future (see https://github.com/elastic/elastic-package/issues/2974).

```suggestion
  - name: nginxreceiver
```

packages/nginx_input_otel/manifest.yml
ishleenk17: We should probable remove it in the default creation of the OTEL input package type.

packages/nginx_input_otel/manifest.yml
ishleenk17: For now is the dataset name picked up from the policy template name ?
How would this work in case of generic log receivers, say filelogreceiver ?
Is the expectation we name the policy to the datastream where we would expect our logs to go ?

packages/nginx_input_otel/fields/base-fields.yml
ishleenk17: Yes, this is mostly going to be the only file under fields. I don't expect any mappings to go in here. 
Do we actually expect predefined mappings for OTEL Input packages. 
Or we can rely on the otel template and the dynamic mappings ?

packages/nginx_input_otel/manifest.yml
jsoriano: > For now is the dataset name picked up from the policy template name ?

Yes, that is what Fleet does for input packages in general, not only for `otelcol`.

> How would this work in case of generic log receivers, say filelogreceiver ?

It will work the same. This is only the default, users can still customize it.

> Is the expectation we name the policy to the datastream where we would expect our logs to go ?

Yes, this will affect to what data stream data will be ingested into. By using the default receiver name, we would do the same as vanilla receivers, and content packages can make the same assumptions (as they already do for example [here](https://github.com/elastic/integrations/blob/50dfc4f09fbd145d67cfc9941f866e308cce1858/packages/mysql_otel/kibana/dashboard/mysql_otel-73691849-6ca6-461b-b37a-b8b0d7d3d0a5.json#L49)).

packages/nginx_input_otel/fields/base-fields.yml
jsoriano: > Do we actually expect predefined mappings for OTEL Input packages.
> Or we can rely on the otel template and the dynamic mappings ?

We can rely on the pre-installed template and the dynamic mappings it includes.

We still have the option of defining fields if needed in some case, but defining fields will be completely optional for OTel input packages. It will be actually expected for OTel input packages to don't contain any field definition.

packages/nginx_input_otel/_dev/build/docs/README.md
ishleenk17: @daniela-elastic WDYT ?

packages/nginx_input_otel/fields/base-fields.yml
tommyers-elastic: based on jaime's comment below we can remove this file entirely. resolving this thread.

ishleenk17: @tommyers-elastic : 
FYI refer to below statement form Jaime 

> If this is the only fields file, build will fail, till we release https://github.com/elastic/package-spec/pull/994 (probably this week).

Hence if not this we will need to keep some file till the next package spec is released. So for now I think we can keep this file or maybe remove this and just create a dummy field in it.

packages/nginx_input_otel/_dev/build/docs/README.md
daniela-elastic: I like that, if we could delegate to the upstream receiver this means less work on our side. Do you foresee any cons to delegating to the upstream receiver? Wouldn't want to go into one way doors.

packages/nginx_input_otel/agent/input/input.yml.hbs
cmacknz: `resourcedetection` should be a default global processor, and the list of global processors needs to be editable (unlike the hard coded list in agent today). I always envisioned that global processors would be associated with exporters.

jsoriano: > `resourcedetection` should be a default global processor, and the list of global processors needs to be editable (unlike the hard coded list in agent today).

So in the case of Fleet-managed integrations they should be handled by Fleet, right?
Do they need to be included in packages then as is being done here? We risk having duplicates if this starts being managed by Fleet.

> I always envisioned that global processors would be associated with exporters.

There probably needs to be both, processors associated to exporters and to policies.

ishleenk17: For now are we going to go ahead with processor as part of the agent config ?
@jsoriano @cmacknz

packages/nginx_input_otel/_dev/build/docs/README.md
ishleenk17: @tommyers-elastic : Maybe we can write this in the Metric reference section. Other parts of the documentation should be there IMO. WDYT ?

packages/nginx_input_otel/agent/input/input.yml.hbs
cmacknz: > So in the case of Fleet-managed integrations they should be handled by Fleet, right?

Yes, the default processors would have to be defined in Fleet which we don't do yet and has some problems e.g. needing a Kibana release to change the config. Creating an [output package type](https://github.com/elastic/ingest-dev/issues/6361) was one idea for how to solve this.

Per policy global processors would also make sense but this could just be syntactic sugar that inserts them into each exporter pipeline.

> For now are we going to go ahead with processor as part of the agent config ?

Yes there is no other way for us to do this right now. I think you would want the detectors or even the whole processor configuration to be configurable though.

ishleenk17:
@jsoriano : COuld you pelase rereview this ?

packages/nginx_input_otel/fields/base-fields.yml
jsoriano: Did you consider removing this file? You can leave it empty or wait for next elastic-package release (coming soon) if you have problems with validation.

packages/nginx_input_otel/manifest.yml
```diff
@@ -0,0 +1,50 @@
+format_version: 3.5.0
+name: nginx_otel_input
+title: "NGINX OpenTelemetry Input Package"
+version: 0.0.1
```
jsoriano: Nit. I would start with 0.1.0, so there can be patch released for the first minor.

packages/nginx_input_otel/agent/input/input.yml.hbs
jsoriano: @ishleenk17 did you consider making this configurable as suggested by Craig?

packages/nginx_input_otel/fields/base-fields.yml
ishleenk17: @jsoriano  On removing the base-fields.yml

1. file "/Users/ishleenkaur/go/ish_12Oct_integ/ish_integrations/build/packages/nginx_otel_input-0.0.1.zip/fields/base-fields.yml" is invalid: file is empty, but media type is defined

2. On having this file empty, we get below message: 

Error: building package failed: invalid content found in built zip package: found 4 validation errors:
   1. expected field "data_stream.dataset" with type "constant_keyword" not found
   2. expected field "data_stream.namespace" with type "constant_keyword" not found
   3. expected field "@timestamp" with type "date" not found
   4. expected field "data_stream.type" with type "constant_keyword" not found
   
 So for now we need to have this file with the fields, or we wait for the next release of elastic package.

packages/nginx_input_otel/agent/input/input.yml.hbs
ishleenk17: @jsoriano As we discussed in our meeting. Shall we just give a yaml box for procesors, keeping the whole yaml configuration configurable.
I don't think parts of the rdp processor should be made configurable. 
WDYT  @tommyers-elastic

packages/nginx_input_otel/_dev/build/docs/README.md
ishleenk17: @tommyers-elastic : we would go ahead with the documentation as is for now, as we discussed.
I can maybe remove the metrics reference section here as for that they can refer to the upstream receiver.

packages/nginx_input_otel/agent/input/input.yml.hbs
jsoriano: > As we discussed in our meeting. Shall we just give a yaml box for procesors, keeping the whole yaml configuration configurable.

We discussed this solution in the context of the `hostmetrics` receiver, which has a pretty extensive set of options. I wouldn't generalize the use of yaml boxes for all cases.

Thinking on this case of the `resourcedetection` processor, in the mid-long term I would like to see this managed by Fleet, so in the meantime as you prefer.

ishleenk17: I would keep it as non configurable until this comes to Fleet.

packages/nginx_input_otel/_dev/build/docs/README.md
tommyers-elastic: i still think this is maybe too much to include here. if we want to auto generate these input packages then this type of readme would be really hard to make.

ishleenk17:
After discussing with @jsoriano , we will proceed with the merge after the new version of elastic package is released. It will help us removing the bas-fields.yml file.

packages/nginx_input_otel/fields/base-fields.yml
fred-maussion: This is working as expected with the latest elastic-package version where the build is successful without the `base-fields.yml` and the `fields` folder
`elastic-package v0.117.1 version-hash undefined (build time: unknown)`

packages/nginx_input_otel/_dev/build/docs/README.md
ishleenk17: Keeping the documentation minimal.

ishleenk17:
> After discussing with @jsoriano , we will proceed with the merge after the new version of elastic package is released. It will help us removing the bas-fields.yml file.

With new elastic package, packages pass CI even without any fields related files.

packages/nginx_input_otel/agent/input/input.yml.hbs
```diff
@@ -0,0 +1,11 @@
+receivers:
+  nginx:
+    {{receiver_config}}
+processors:
+  resourcedetection/system:
+    detectors: ["system"]
```
tommyers-elastic, line 6: this is required for host.name, right?

packages/nginx_input_otel/agent/input/input.yml.hbs
```diff
@@ -0,0 +1,11 @@
+receivers:
+  nginx:
+    {{receiver_config}}
+processors:
+  resourcedetection/system:
```
tommyers-elastic, line 5: i think we should add unique component identifiers for everything we have here to avoid potential conflicts

i.e. `nginx/elastic_nginx_input` and `resourcedetection/elastic_nginx_input` or something

packages/nginx_input_otel/img/nginx_otellogo.svg
tommyers-elastic, line 1: is this LLM generated? something funky about the wobbly lines...

tommyers-elastic (approved):
thanks ishleen. approving, but i think we should add the component identifiers before we merge.

packages/nginx_input_otel/agent/input/input.yml.hbs
cmacknz: Namespacing the components is something Fleet should take care of, since it can consistently add the integration name and package policy ID as a suffix. 

This would be similar to how we inject unique `id` keys into Beat inputs.

@jsoriano I thought we did this already? Do we need a lint rule to avoid collisions with component identifiers added manually or do we just append to them?

jsoriano: Correct, Fleet takes care of namespacing by appending randomized ids. If a package uses the same component multiple times in a single template it needs to be namespaced, and Fleet will still handle it. In the case here I don't think namespacing is neccesary as the resource is only used once.

packages/nginx_input_otel/agent/input/input.yml.hbs
ishleenk17: Correct

packages/nginx_input_otel/agent/input/input.yml.hbs
ishleenk17: Considering this, I think we are good to leave this configuration as it is, without component identifiers @tommyers-elastic

packages/nginx_input_otel/img/nginx_otellogo.svg
ishleenk17: This is the one we used for NGINX OTEL Content package

ishleenk17: But yeah its not that clear. We need a standardised way of generating these logos for OTEL packages. Currently developer is following their own ways of having these logos.

packages/nginx_input_otel/manifest.yml
```diff
@@ -0,0 +1,42 @@
...
+    vars:
+      - name: receiver_config
+        type: yaml
+        title: Receiver Configuration
+        required: true
+        show_user: true
+        description: Configuration for the NGINX receiver. This configuration will be inserted directly into the receiver section of the OpenTelemetry Collector configuration.
+        default: |
+          endpoint: http://localhost:80/status
+          collection_interval: 10s
```
jsoriano, line 39: Hey,

I would like to reopen the discussion about this approach of having yaml blocks for OTel settings. I think this makes sense in cases whose complexity and variety of configurations is big, such as with `resourcedetection`, but for simple receivers this is bringing complexity to users.

Specifically for this case, this receiver has only three settings: endpoint, collection interval and start delay. Defining these three variables provides a much better UX. The user can see what they can configure, what are the defaults and we perform some validations for them.
More generally speaking, the approach of defining all variables gives us more control on what we support and the UX that we provide to our users.

When thinking on composable packages, having a single yaml block in the input package forces us as package developers to do the same only for these inputs, or define the variables and templates in the composable integration, what can lead to inconsistent UXs depending on the packages and the inputs they reuse.

It also defeats the purpose of defining everything related to each input in input packages. Each composable integration that wants to offer higher level variables will have to define them, instead of reusing what is defined in the input packages it uses.

I have opened a PR that illustrates how the package would look like if we would define the variables: https://github.com/elastic/integrations/pull/16313

cmacknz: We plan to build a pipeline editor for people who want more direct control over the configuration, and integration packages will be the more curated and guided UX for people who do not want to do that and just want things to work with as few knobs as possible.

I don't think we should try to turn input packages into free form YAML editors for individual otel components, it will just be a worse pipeline editor.

ishleenk17: > When thinking on composable packages, having a single yaml block in the input package forces us as package developers to do the same only for these inputs, or define the variables and templates in the composable integration, what can lead to inconsistent UXs depending on the packages and the inputs they reuse.

My thought was have a single yaml block in input package. And when we develop say a composable package for nginx, it would pass the specific input variables. Will composable packages not work this way @jsoriano ?

jsoriano: > > When thinking on composable packages, having a single yaml block in the input package forces us as package developers to do the same only for these inputs, or define the variables and templates in the composable integration, what can lead to inconsistent UXs depending on the packages and the inputs they reuse.
> 
> My thought was have a single yaml block in input package. And when we develop say a composable package for nginx, it would pass the specific input variables. Will composable packages not work this way @jsoriano ?

The problem with this approach is that then all the composable packages that use a given input package need to define the variables, when one of the ideas of input packages is to make them the single point of truth for inputs. For the nginx input package it might be not such an issue because there are not going to be many integrations using it, but it might be in other cases.

Defining the variables in the input packages is like defining their interface, what helps when reusing them. The alternative is more error-prone, as will require to merge configurations with lower level yaml blocks instead of well defined variables.

I understand in some cases the complexity of the possible configurations might require to have some kind of yaml block, but I think this should be the exception. In most cases there will be a limited set of settings that we want to support, as is the case here.

PR #15657, opened 2025-10-15
[ess_billing] Mismatch due to partial periods being dropped

5hanker:
On the [Metrics ESS Billing] Billing dashboard, there is a mismatch in the numbers when you compare Total cost for period (ECUs) with "Cost per deployments/projects over time" or "Cost per features over time" or "ECH Capacity Costs per instance flavors over time" or "Serverless features per service types over time". This happens when you choose custom time range like "from:'2024-07-01T00:00:00.000Z',to:'2025-09-30T23:59:59.999Z". I see that its because this viz have "Drop partial intervals". We need to make the numbers consistent.

image


3kt:
@5hanker can you start the PR from upstream:main rather than from one of my former (stale) fork? I believe the diff is inaccurate because of this.

Issue #15605, opened 2025-10-08
Microsoft Defender for Endpoint: make OAuth2 endpoint params configurable


chemamartinez:
We need to expose the [`oauth_endpoint_params`](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-cel#_auth_oauth2_endpoint_params) option for the Microsoft Defender for Endpoint integration.

It provides users more flexibility to set or update parameters when using oauth2 token authentication.

The following Microsoft integrations could also apply for this request:

- Microsoft Defender XDR (M365 Defender)

- Microsoft Defender for Cloud

- Microsoft Exchange Online Message Trace

- Microsoft Sentinel

- Microsoft Office 365 Metrics

PR #15667, opened 2025-10-16
microsoft_defender_endpoint: add support for oauth endpoint params

chemamartinez:
## Proposed commit message

Add support for the `oauth_endpoint_params` configuration parameter for all available data streams.

`Log` data stream still works under httpjson so the option has been added under data stream level along with all the OAuth2 options for this data stream.

For the another data streams, as they work under the CEL input, it has been added at input level so adding any value to this option will affect all data streams that rely on CEL (`machine`, `machine_action`, and `vulnerability`).

Finally, the auth logic for the vulnerability data stream is implemented in the CEL program instead of delegate in the CEL auth options for the input. Therefore, the oauth endpoint params in this case are added manually in the program as well.

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/15605
- Closes https://github.com/elastic/enhancements/issues/26099

## Screenshots

image

### DIFF
--- a/packages/microsoft_defender_endpoint/manifest.yml
+++ b/packages/microsoft_defender_endpoint/manifest.yml
@@ -89,7 +89,7 @@ policy_templates:
           - name: oauth_endpoint_params
             type: yaml
             title: OAuth2 Endpoint Params
-            description: Endpoint Params used for OAuth2 authentication as YAML. See [documentation](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-cel#_auth_oauth2_endpoint_params) for details.
+            description: Set of values that will be sent on each resource to the OAuth Server URL. Each param key can have multiple values and they are appended to the URL as query parameters.
             show_user: false
             multi: false
             required: false


packages/microsoft_defender_endpoint/manifest.yml
```diff
@@ -86,6 +86,13 @@ policy_templates:
             required: true
             show_user: false
             secret: false
+          - name: oauth_endpoint_params
+            type: yaml
+            title: OAuth2 Endpoint Params
+            description: Endpoint Params used for OAuth2 authentication as YAML. See [documentation](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-cel#_auth_oauth2_endpoint_params) for details.
```
efd6: I don't believe this documentation is correct to link here. We are not using that field, we're constructing the headers ourselves in the CEL code.

packages/microsoft_defender_endpoint/data_stream/log/agent/stream/httpjson.yml.hbs
```diff
@@ -16,6 +16,9 @@ auth.oauth2.scopes:
 auth.oauth2.provider: azure
 auth.oauth2.azure.resource: {{azure_resource}}
 {{/if}}
+{{#if oauth_endpoint_params}}
+auth.oauth2.endpoint_params: {{oauth_endpoint_params}}
```
efd6, line 20: Can you confirm that the elements of the map are promoted to arrays when the config is deserialised? The final destination for the data is a [`map[string][]string`](https://github.com/elastic/beats/blob/784e0f0fe03f63abfb3775302307168772d0acdc/x-pack/filebeat/input/httpjson/config_auth.go#L86), while while the data here is a map of string to string.

packages/microsoft_defender_endpoint/manifest.yml
chemamartinez: Ok, I will modify the description to avoid linking it to the CEL option. This is true for one of the three data streams that use this parameter, but adding this distinction in the param description I think is too technical for users and can be confusing.

packages/microsoft_defender_endpoint/data_stream/log/agent/stream/httpjson.yml.hbs
chemamartinez: Trying the same example value as the documentation shows:

Screenshot 2025-10-21 at 18 12 35

I see elements in the array are deserialised as expected, here the content of different config files from diagnostics:

**computed-config.yaml**
```
streams:
    - auth:
        oauth2:
            client:
                id: clientid
                secret: <REDACTED>
            endpoint_params:
                Param1:
                    - ValueA
                    - ValueB
                Param2:
                    - Value
```

**components-actual.yaml**
```
endpoint_params:
    kind:
        structvalue:
            fields:
                Param1:
                    kind:
                        listvalue:
                            values:
                                - kind:
                                    stringvalue: ValueA
                                - kind:
                                    stringvalue: ValueB
                Param2:
                    kind:
                        listvalue:
                            values:
                                - kind:
                                    stringvalue: Value
```

And the params are well formatted in the query:
```
"http.request.body.content":"Param1=ValueA&Param1=ValueB&Param2=Value&grant_type=client_credentials&scope=https%3A%2F%2Fapi.securitycenter.windows.com%2F.defaults"
```

This is working for all data streams (httpjson and cel), but it shows an error when adding an array in the vulnerability data stream:
```
{"log.level":"error","@timestamp":"2025-10-21T16:13:40.502Z","message":"failed evaluation","component":{"binary":"filebeat","dataset":"elastic_agent.filebeat","id":"cel-default","type":"cel"},"log":{"source":"cel-default"},"id":"cel-microsoft_defender_endpoint.vulnerability-6be99673-02b3-4c3b-9224-3bdcb63ca863","input_url":"https://api.security.microsoft.com","error":{"message":"failed eval: ERROR: <input>:2:43: no such overload for format_query: type conversion error from list to 'string'\n |   state.?work_list.orValue([]).size() > 0 ?\n | ..........................................^"},"service.name":"filebeat","input_source":"https://api.security.microsoft.com","ecs.version":"1.6.0","log.logger":"input.cel","log.origin":{"file.line":301,"file.name":"cel/input.go","function":"github.com/elastic/beats/v7/x-pack/filebeat/input/cel.input.run.func1"},"ecs.version":"1.6.0"}
```

chemamartinez: Fixed this case: eac67d9525c016f43772e9c50b3e508fbfb84d25

Now the test case for the vulnerability data stream contemplates the array possibility, the resulting query for this test is:
```
microsoft-defender-endpoint-vulnerability-cel-1  | {"level":"debug","ts":"2025-10-21T16:42:29.106Z","caller":"httpserver/httpserver.go:191","msg":"Rule #0 matched: request #1 => Request path: POST /tenant_id/oauth2/v2.0/token?client_id=test-app-id&client_secret=test-secret&grant_type=refresh_token&refresh_token=refresh_token_1&refresh_token=refresh_token_2&scope=https%3A%2F%2Fsecuritycenter.onmicrosoft.com%2Fwindowsatpservice%2F.default, Request Headers: 'Accept-Encoding: [gzip]' 'Connection: [close]' 'User-Agent: [Elastic-Filebeat/8.19.3 (linux; arm64; 7a508104ea6b6ddae7456dff8b61423e59a73962; 2025-08-26 01:13:25 +0000 UTC)]' 'Content-Length: [214]' 'Content-Type: [application/x-www-form-urlencoded]' , Request Body: ","address":":8080"}
```

packages/microsoft_defender_endpoint/manifest.yml
efd6: Absolutely. I just didn't want users to get confused by the link to the filebeat field.

packages/microsoft_defender_endpoint/data_stream/log/agent/stream/httpjson.yml.hbs
efd6: Thanks

PR #15673, opened 2025-10-17
[checkpoint_firewall] Update type from integer to long to support higher values

haetamoudi:
## Proposed commit message
Update count types from integer to long

A customer reported that the pipeline fails when handling large numbers (e.g., 4294947622).
Updated the following fields to use the long type instead of integer:
- checkpoint.update_count
- checkpoint.connection_count
- checkpoint.aggregated_log_count

Related Issue
See [elastic/sdh-beats#6530](https://github.com/elastic/sdh-beats/issues/6530)

## Checklist

### DIFF
--- a/packages/checkpoint/docs/README.md
+++ b/packages/checkpoint/docs/README.md
@@ -147,7 +147,7 @@ An example event for `firewall` looks as following:
 | checkpoint.additional_rdata | List of additional resource records. | keyword |
 | checkpoint.administrator | Source administrator name. | keyword |
 | checkpoint.advanced_changes |  | keyword |
-| checkpoint.aggregated_log_count | Number of logs aggregated in the event. | integer |
+| checkpoint.aggregated_log_count | Number of logs aggregated in the event. | long |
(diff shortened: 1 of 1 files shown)


qcorporation:
@haetamoudi I believe that `update_count`, `connection_count` and `aggregated_log_count` are defined as `type: integer` within the `fields.yml` file.
Question 1: do we need to update `fields.yml` so that these fields are now updated to long?
Question 2: if we update the field definition to long, is it now a breaking change?

cc.ing @taylor-swanson

andrewkroh:
> Question 1: do we need to update fields.yml so that these fields are now updated to long?

Yes, we need to update the fields.yml files in order to store `long` values. Without the change indexing could fail when the number no longer fits in an the fields defined Elasticsearch data type.

> Question 2: if we update the field definition to long, is it now a breaking change?

Not a breaking change IMO. Data types within a field families are safe although the performance characteristics may differ:

- keyword family - `keyword`, `constant_keyword`, `wildcard`
- text family
- numeric - `byte`, `short`, `integer`, `long`, `float`, `scaled_float`, `double` (not an official field family type)

PR #13350, opened 2025-03-28
AWS S3 input bulk fix for number_of_workers

flexitrev:
- Bug

## Proposed commit message

Fixes number_of_workers setting so it is not dependent on S3 bucket collection being enabled

## Checklist

- [X] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [] I have verified that all data streams collect metrics or logs.
- [X] I have added an entry to my package's `changelog.yml` file.
- [X] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [N/A] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally
Modify number of worker settings on previous version of the integration without enabling S3 bucket selection. 
Download policy
Observe that default number of workers is still present or not present at all, relying on aws-s3 input defaults

Modify number of worker settings on this version of the integration without enabling S3 bucket selection. 
Download policy
Observe that default number of workers is the value of user input

## Related issues

- Closes #13179

Issue #13179, opened 2025-03-19
[Cloudflare Logpush] Setting number_of_workers is ignored if SQS is used.


leandrojmp:
This issue was created from a comment on another issue, the comment has the description of the problem.

> @strawgate I'm using the Cloudflare Logpush integration, the Firewall Events and HTTP Requests dataset.
> 
> I have multiple Inputs with the same configuration only the queue change, I'm sharing one of them here:
> 
> ```
>   - id: aws-s3-cloudflare-f56bb89c-97bd-441e-ae4b-071d5296cd1d
>     name: cloudflare-REDACT
>     revision: 6
>     type: aws-s3
>     use_output: c20b778c-0220-40f4-a8ca-03b974ada63b
>     meta:
>       package:
>         name: cloudflare_logpush
>         version: 1.35.0
>     data_stream:
>       namespace: cloudflare
>     package_policy_id: f56bb89c-97bd-441e-ae4b-071d5296cd1d
>     streams:
>       - id: >-
>           aws-s3-cloudflare_logpush.firewall_event-f56bb89c-97bd-441e-ae4b-071d5296cd1d
>         data_stream:
>           dataset: cloudflare_logpush.firewall_event
>           type: logs
>         queue_url: >-
>           https://sqs.us-east-1.amazonaws.com/REDACTED
>         visibility_timeout: 300s
>         api_timeout: 120s
>         max_number_of_messages: 70
>         file_selectors:
>           - regex: fw/
>         access_key_id: '${SECRET_3}'
>         secret_access_key: '${SECRET_4}'
>         tags:
>           - collect_sqs_logs
>           - cloudflare-firewall
>           - cloudflare-fw-REDACT
>         processors:
>           - drop_event:
>               when:
>                 or:
>                   - contains:
>                       message: REDACTED
>       - id: >-
>           aws-s3-cloudflare_logpush.http_request-f56bb89c-97bd-441e-ae4b-071d5296cd1d
>         data_stream:
>           dataset: cloudflare_logpush.http_request
>           type: logs
>         queue_url: >-
>           https://sqs.us-east-1.amazonaws.com/REDACTED
>         visibility_timeout: 300s
>         api_timeout: 120s
>         max_number_of_messages: 70
>         file_selectors:
>           - regex: http/
>         access_key_id: '${SECRET_3}'
>         secret_access_key: '${SECRET_4}'
>         tags:
>           - collect_sqs_logs
>           - cloudflare-http
[...]

PR #15697, opened 2025-10-20
[Netskope] Add alerts_events_v2 data stream to fetch the data for alerts_v2 and events_v2 from a single queue

### LINKED ISSUE #15459: [Netskope]: Using Log Streaming mode does not support to get both Events and Alerts
Hello,

Currently we use the netskope log shipper to get logs from netskope, but it seems that this will be deprecated and the new way to get logs is using the Log Streaming.

The integration supports getting Alerts and Events using the Log Streaming mode, but it seems that it cannot can get **both** at the same time.

On Netskope side, when using Log Streaming the Events and Alerts are streamed together. [[docs](https://docs.netskope.com/en/choosing-data-sets)]

> When admins select Alerts and Events, they are streamed together.

So, you would have logs from alerts and events on the same bucket, which would end up on the same SQS queue.

On the integration side, you have different configurations for Alerts and Events, the ingest pipeline for Alerts will drop Events logs and the ingest pipeline for Events will drop Alerts logs.

You cannot use the same SQS queue in both configurations, and using SNS to send notifications to 2 different queues is an unnecessary work and would lead to the data being processed twice.

And you also cannot have multiple streams of the same data type on Netskope side.

The integration needs to have a **routing** ingest pipeline that would direct the Alerts to the Alerts pipeline and the Events to the Events pipeline.

The way it currently works the user needs to choose to collect Alerts or to collect Events.

moxarth-rathod:
## Proposed commit message

```
netskope: add alerts_events_v2 data stream to fetch the data for alerts_v2 and events_v2 from a single queue
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/netskope directory.

> elastic-package test -v

Additionally, the following cloud credentials are required to setup:

AWS:
- AWS_DEFAULT_PROFILE
- AWS_ACCESS_KEY_ID
- AWS_SECRET_ACCESS_KEY
- AWS_SESSION_TOKEN.

## Related issues

- Closes #15459

### DIFF
--- a/packages/netskope/data_stream/alerts_events_v2/_dev/test/pipeline/test-combined.log-expected.json
+++ b/packages/netskope/data_stream/alerts_events_v2/_dev/test/pipeline/test-combined.log-expected.json
@@ -28,7 +28,7 @@
                 "port": 1143
             },
             "ecs": {
-                "version": "8.17."
+                "version": "8.17.0"
             },
             "email": {
                 "subject": "TESTING MALWARE"
@@ -324,7 +324,7 @@
                 "port": 1144
             },
             "ecs": {
-                "version": "8.17."
+                "version": "8.17.0"
             },
             "email": {
                 "subject": "ALERT: Suspicious Activity"
@@ -627,7 +627,7 @@
                 "port": 2022
             },
             "ecs": {
-                "version": "8.17."
+                "version": "8.17.0"
             },
             "event": {
                 "action": "bypass",
@@ -790,7 +790,7 @@
                 "port": 6253
             },
             "ecs": {
-                "version": "8.17."
+                "version": "8.17.0"
             },
             "event": {
                 "action": "restrictaccess",


packages/netskope/data_stream/alerts_events_v2/agent/stream/aws-s3.yml.hbs
```diff
@@ -0,0 +1,98 @@
+{{#if collect_s3_logs}}
+
+{{#if bucket_arn}}
+bucket_arn: {{bucket_arn}}
+{{/if}}
+{{#if number_of_workers}}
+number_of_workers: {{number_of_workers}}
```
leandrojmp: Hello @moxarth-rathod 

This needs to be outside `#if collect_s3_logs` and `#if queue_url` as the setting `number_of_workers` is applied to both ways of getting data from S3, polling and SQS, the setting `max_number_of_messages` is ignored on agents higher than 8.16+

For more context this was reported on `#13179` and fixed on multiple integrations on `#13350`

packages/netskope/_dev/build/docs/README.md
```diff
@@ -44,6 +44,10 @@ Please make sure to use the given response formats.
 
 Considering you already have an AWS S3 bucket setup, to configure it with Netskope, follow [these steps](https://docs.netskope.com/en/stream-logs-to-amazon-s3) to enable the log streaming.
 
+**Note**: It is recommended to use the combined alerts_events_v2 data stream rather than configuring the individual events_v2 or alerts_v2 data stream. The alerts_events_v2 stream automatically directs logs to the appropriate individual data streams.
+
+If the individual v2 data streams, events_v2 or alerts_v2, are used via SQS, it is necessary to implement event-based bucket segregation.
```
leandrojmp: This may change in the future, but according to current Netskope documentation, if you choose any event type and alerts, they will be streamed together.

The user may choose to stream only events or to stream only alerts, but when choosing alerts and any other event type, they will be streamed together.

packages/netskope/data_stream/alerts_events_v2/agent/stream/aws-s3.yml.hbs
moxarth-rathod: Thanks! I've made the changes accordingly.

packages/netskope/changelog.yml
```diff
@@ -1,4 +1,11 @@
 # newer versions go on top
+- version: "2.3.1"
+  changes:
+    - description: >-
+        Add alerts_events_v2 data stream to support fetching alert v2 and event v2 data from a single queue.
+        Users using SQS input should consider disabling alerts_v2 and events_v2 to avoid conflicts, and use the combined data stream instead.
+      type: breaking-change
```
efd6, line 7: If this is breaking, the version should be 3.0.0.

packages/netskope/changelog.yml
```diff
@@ -1,4 +1,11 @@
...
+      link: https://github.com/elastic/integrations/pull/1
```
efd6:
```suggestion
      link: https://github.com/elastic/integrations/pull/15697
```

packages/netskope/data_stream/alerts_events_v2/_dev/test/pipeline/test-combined.log-expected.json
```diff
@@ -0,0 +1,936 @@
...
+                "geo": {
+                    "city_name": "Ikebukuro",
+                    "country_iso_code": "AU",
+                    "location": {
+                        "lat": 37.7749,
+                        "lon": 151.2093
+                    },
+                    "postal_code": "2099",
+                    "region_name": "Western Australia",
+                    "timezone": "America/Los_Angeles"
```
efd6: :rofl:

packages/netskope/data_stream/alerts_events_v2/routing_rules.yml
```diff
@@ -0,0 +1,12 @@
+- source_dataset: netskope.alerts_events_v2
+  rules:
+    - target_dataset: netskope.alerts_v2
+      if: ctx.netskope?.alert_v2?.alert == "yes"
+      namespace:
+        - "{{data_stream.namespace}}"
+        - default
+    - target_dataset: netskope.events_v2
+      if: ctx.netskope?.events_v2?.alert == "no" || ctx.netskope?.events_v2?.alert == null
```
efd6:
```suggestion
      if: ctx.netskope?.events_v2?.alert == null || ctx.netskope.events_v2.alert == "no"
```
if the first branch is true, the second is not evaluated, so if it is evaluated, `ctx.netskope.events_v2` is not null.

packages/netskope/data_stream/alerts_v2/manifest.yml
```diff
@@ -150,7 +150,7 @@ streams:
   - input: aws-s3
     template_path: aws-s3.yml.hbs
     title: Collect Alert logs via AWS S3 or SQS
-    description: Collect Alert logs via AWS S3 or SQS input.
+    description: Collect Alert logs via AWS S3 or SQS input. Note- When using SQS, implementing event-based bucket segregation is required to guarantee that each queue receives only the events relevant to the respective data stream.
```
efd6: What does this look like in the UI? Does `description` respect new lines? If it does, I think that would be better. Also, `s/Note-/Note:/`.

packages/netskope/data_stream/events_v2/manifest.yml
```diff
@@ -298,7 +298,7 @@ streams:
   - input: aws-s3
     template_path: aws-s3.yml.hbs
     title: Collect Event logs via AWS S3 or SQS
-    description: Collect Event logs via AWS S3 or SQS input.
+    description: Collect Event logs via AWS S3 or SQS input. Note - When using SQS, implementing event-based bucket segregation is required to guarantee that each queue receives only the events relevant to the respective data stream.
```
efd6: Same here.

packages/netskope/data_stream/alerts_events_v2/manifest.yml
```diff
@@ -0,0 +1,395 @@
...
+        title: Polling interval
+        multi: false
+        required: false
+        show_user: true
+        default: 15s
+        description: Determines the time interval between polling operations.
+      - name: containers
+        type: yaml
+        title: Containers
+        description: "This attribute contains the details about a specific container like, name, number_of_workers, poll, poll_interval etc. \nThe attribute 'name' is specific to a container as it describes the container name, while the fields number_of_workers, poll, poll_interval can exist both at the container level and at the global level. \nIf you have already defined the attributes globally, then you can only specify the container name in this yaml config. \nIf you want to override any specific attribute for a container, then, you can define it here. \nAny attribute defined in the yaml will override the global definitions. Please see the relevant [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-azure-blob-storage.html#attrib-containers) for further information.\n"
```
efd6: Can we use string literal syntax here so that the `\n` are not needed?

packages/netskope/data_stream/alerts_events_v2/manifest.yml
```diff
@@ -0,0 +1,395 @@
...
+          #  poll_interval: 10s
+      - name: file_selectors
+        type: yaml
+        title: File Selectors
+        multi: false
+        required: false
+        show_user: false
+        default: |
+          # - regex: "event/"
+        description: "If the container will have events that correspond to files that this integration shouldn’t process, file_selectors can be used to limit the files that are downloaded. This is a list of selectors which is made up of regex patters. \nThe regex should match the container filepath. Regexes use [RE2 syntax](https://pkg.go.dev/regexp/syntax). Files that don’t match one of the regexes will not be processed.\n"
```
efd6: Same here.

packages/netskope/data_stream/alerts_events_v2/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,54 @@
...
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - pipeline:
+      name: '{{ IngestPipeline "pipeline_alerts_v2" }}'
+      if: ctx.netskope?.alerts_or_events?.alert == "yes"
+      tag: pipeline_alerts
+      ignore_missing_pipeline: true
+  - pipeline:
+      name: '{{ IngestPipeline "pipeline_events_v2" }}'
+      if: ctx.netskope?.alerts_or_events?.alert == "no" || ctx.netskope?.alerts_or_events?.alert == null
```
efd6:
```suggestion
      if: ctx.netskope?.alerts_or_events?.alert == null || ctx.netskope.alerts_or_events.alert == "no"
```

packages/netskope/changelog.yml
ShourieG: This definitely a breaking change

packages/netskope/data_stream/alerts_v2/manifest.yml
ShourieG: @moxarth-rathod, I'm curious, what's the purpose of having input manifests in these routed data streams if we are not actually consuming the data here? Rather these routed data streams should contain the field definitions, which are currently missing.

The manifests in the routed data streams should contain `title`, `datastream.dataset`, `type`, and you can keep any `elasticsearch` configs if necessary. This change is a completely breaking change so keeping input manifests seems pointless.

packages/netskope/data_stream/events_v2/manifest.yml
ShourieG: same as [this](https://github.com/elastic/integrations/pull/15697/files#r2472085274)

packages/netskope/data_stream/alerts_v2/manifest.yml
moxarth-rathod: I was considering that the user might manually update their configured input as suggested in the changelog, but I agree, we can remove the inputs.

packages/netskope/data_stream/alerts_events_v2/manifest.yml
moxarth-rathod: Yes, i've made the changes here,

packages/netskope/data_stream/alerts_events_v2/_dev/test/pipeline/test-combined.log-expected.json
kcreddy: 🌏 🌍 🌎

packages/netskope/data_stream/alerts_events_v2/elasticsearch/ingest_pipeline/pipeline_events_v2.yml
```diff
@@ -4,31 +4,18 @@ processors:
   - set:
       field: ecs.version
       tag: set_ecs_version
-      value: 8.17.0
+      value: 8.17.
```
kcreddy:
```suggestion
      value: 8.17.0
```

packages/netskope/data_stream/alerts_events_v2/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,54 @@
...
+  - remove:
+      field: message
+      tag: remove_message
+      ignore_missing: true
+      description: The `message` field is no longer required if the document has an `event.original` field.
+      if: ctx.event?.original != null
+  - json:
+      field: event.original
+      tag: json_event_original
+      target_field: netskope.alerts_or_events
```
kcreddy: Any reason why we are not using the data stream name i.e., `netskope.alerts_events_v2` but instead opting for `netskope.alerts_or_events` ?

packages/netskope/_dev/build/docs/README.md
```diff
@@ -44,6 +44,8 @@ Please make sure to use the given response formats.
 
 Considering you already have an AWS S3 bucket setup, to configure it with Netskope, follow [these steps](https://docs.netskope.com/en/stream-logs-to-amazon-s3) to enable the log streaming.
 
+**Note**: It is recommended to use the combined alerts_events_v2 data stream rather than configuring the individual events_v2 or alerts_v2 data stream. The alerts_events_v2 stream automatically directs logs to the appropriate individual data streams.
```
kcreddy: Unless users look at the integration code, they are not aware of `alerts_events_v2`, `events_v2`, or `alerts_v2`. It might be better to use titles instead. 
`Alerts V2 and Events V2`, `Events V2`, and `Alerts V2`.

packages/netskope/_dev/build/docs/README.md
```diff
@@ -176,24 +178,12 @@ Default port: _9021_
...
 {{fields "events"}}
 
 {{event "events"}}
 
-### Events V2
-
-{{fields "events_v2"}}
-
-{{event "events_v2"}}
-
```
kcreddy, line 196: Can we also add system tests from this routing data stream `alerts_events_v2`, and add sample event and field mappings to README? Looks like it is possible, [example](https://github.com/elastic/integrations/tree/main/packages/entityanalytics_entra_id/data_stream/entity)

efd6 (approved):
LGTM, but please wait for @kcreddy and @ShourieG.

ShourieG:
@moxarth-rathod, the field definitions are not present in each of the routed data streams, It will likely cause errors when the data is routed and published in the respective dataset. Have we tested this with live data ? I believe it's necessary to test this with the fields absent & present. Can we do this ?

ShourieG:
> @moxarth-rathod, the field definitions are not present in each of the routed data streams, It will likely cause errors when the data is routed and published in the respective dataset. Have we tested this with live data ? I believe it's necessary to test this with the fields absent & present. Can we do this ?

After discussing with @moxarth-rathod, the respective field definitions are present and also a live test has been performed.

packages/netskope/data_stream/alerts_events_v2/elasticsearch/ingest_pipeline/default.yml
moxarth-rathod: No, there was no specific reason. I've made changes as per your suggestion.

packages/netskope/_dev/build/docs/README.md
kcreddy: Ignoring for now and will be added later.

PR #15699, opened 2025-10-20
[o365] Fixed parsing and indexing errors

StacieClark-Elastic:
Fixes flattening error in Actions list when the list is encoded json string instead of json objects. Adds fields ActorInfoString, OperationCount, TokenObjectId, TokenTenantId. Added fields Messages and Folders as ExchangeAggregatedMessages and ExchangeAggregatedFolder for record type 50: ExchangeItemAggregated and explicitly convert the SizeInBytes values to long. There are other schemas for fields called Messages and Folders so explicit naming is defensive. 

Fixes error messages:

field "o365.audit.ActorInfoString" is undefined
field "o365.audit.Folders" is used as array of objects, expected explicit definition with type group or nested
field "o365.audit.OperationCount" is undefined
field "o365.audit.TokenObjectId" is undefined
field "o365.audit.TokenTenantId" is undefined
failed to parse field [o365.audit.Actions] of type [flattened] in document with id
[o365.audit.Folders.FolderItems.SizeInBytes] cannot be changed from type [long] to [float]
[o365.audit.Messages.MessageItems.SizeInBytes] cannot be changed from type [long] to [float]

## Checklist

Pipeline tests have examples in added test file that will produce errors before the change. 
Passing pipeline tests verify the fix.
Note that the missing fields errors do not occur until earlier errors in the pipeline are fixed.


packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -132,11 +132,12 @@ processors:
         if (!(ctx.o365audit.Actions instanceof List)) {
           ctx.o365audit.Actions = [ctx.o365audit.Actions];
         }
+        def regex = /,\"QueryTime\":\"[0-9\/]+\s[0-9]+:[0-9]+:[0-9]+\s[AP]M\"|\"QueryTime\":\"[0-9\/]+\s[0-9]+:[0-9]+:[0-9]+\s[AP]M\",/;
```
efd6: `/,"QueryTime\":"[0-9\/]+\s[0-9]+:[0-9]+:[0-9]+\s[AP]M"|"QueryTime":"[0-9\/]+\s[0-9]+:[0-9]+:[0-9]+\s[AP]M",/` since `/` has literal string quoting behaviour.

Suggest also `s/regex/queryTimePattern/g` and adding a comment about why we are doing this.

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -132,11 +132,12 @@ processors:
...
         for (def e: ctx.o365audit.Actions) {
           if (e instanceof Map) {
             actions.add(e);
           } else if (e instanceof String) {
-            ctx._tmp.action_strings.add(e);
+            ctx._tmp.action_strings.add(regex.matcher(e).replaceAll(''));
```
efd6: Do we ever see more than two? If not, this could be `replaceFirst` in order to not lose the information.

Also, maybe MS could fix their "bug".

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1770,6 +1792,36 @@ processors:
       copy_from: o365audit.ApplicationDisplayName
       tag: set_application_name
       ignore_empty_value: true
+
+  # ExchangeItemAggregated Schema
+  - append:
+      field: event.type
+      value: access
+      if: 'ctx.o365audit?.RecordType != null && ctx.o365audit?.RecordType == "50"'
```
efd6:
```suggestion
      if: ctx.o365audit?.RecordType != null && ctx.o365audit.RecordType == "50"
```
(similar below)

packages/o365/changelog.yml
```diff
@@ -1,4 +1,13 @@
 # newer versions go on top
+- version: "2.31.1"
+  changes:
+    - description: >-
+        Fix flattening errors in Action List items due to duplicate QueryTime fields. 
+        Added fields ActorInfoString, OperationCount, TokenObjectId, TokenTenantId.
+        Added fields Messages and Folders as ExchangeMessages and ExchangeFolders 
+        for record type 50: `ExchangeItemAggregated`.
```
efd6: These all get rendered into a single line in the change log; suggest breaking these out into separate changes. Also field names in ``` `` ```.

packages/o365/data_stream/audit/fields/fields.yml
```diff
@@ -275,6 +277,53 @@
...
+          type: nested
+          description: Items in the message
+          fields:
+            - name: SizeInBytes
+              type: long
+              description: Size of the message item in bytes
+            - name: Id
+              type: keyword
+              description: Message item ID
+
```
efd6:
```suggestion
```

packages/o365/changelog.yml
```diff
@@ -1,4 +1,13 @@
...
+      type: bugfix
```
efd6, line 16: I think this is both a bugfix (the missing field definitions) and an enhancement (tolerating (arguably<sup>1</sup>) invalid JSON).

<sup>1</sup>Repeated keys are not strictly invalid, with the [rfc](https://www.rfc-editor.org/rfc/rfc8259.txt) only saying SHOULD.

efd6:
Please also note the origin of the new test cases in the proposed commit message.

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: This is replication of logic that we already have for the Data field.  There are 2 QueryTime fields, with the same time data. The difference being one is more human readable than the other using AM/PM. We remove the the AM/PM one, If there were 2 with AM/PM, we would want it removed as well.

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: This is replication of logic that we already have for the Data field.

packages/o365/changelog.yml
StacieClark-Elastic: Repeated keys in a field being flattened produce errors

packages/o365/data_stream/audit/fields/fields.yml
StacieClark-Elastic: done

packages/o365/changelog.yml
StacieClark-Elastic: It's not really an enhancement. We did this already for another field and missed that this field can do the same thing. Since it's keeping the data from being indexed I'm going to keep it as a bugfix.

packages/o365/changelog.yml
StacieClark-Elastic: done

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: done

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: I added the comment but the kept the way I'm using the regex as this appears to be the documented way to do it.

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
StacieClark-Elastic: We want any QueryTime with AM/PM removed.

StacieClark-Elastic:
From a security standpoint, I don't think we should commenting on where the bug was found. The SDH points to this PR which should be sufficient. The test case is obviously completely fake data.

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
efd6: Documented where?

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
efd6: OK

packages/o365/changelog.yml
efd6: > We did **this** already for another field and missed that this field can do the same thing.

What "this" is this? If it's the addition of fields, sure. If it's the tolerance of invalid inputs, then the previous change should have been marked as an enhancement.

> Since it's keeping the data from being indexed I'm going to keep it as a bugfix.

I don't think this is the correct semantics. Making a thing work with buggy input is not a bugfix (bug's depend on mismatch between specified behaviour not matching actual behaviour, and I do not believe that we state that we will ingest invalid data), and we _can_ have both; we can have two descriptions, one for the bugfix (field definitions) and one for the enhancement (the tolerance of invalid input). However, there are other hills that are more important to me.

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -132,11 +132,16 @@ processors:
         if (!(ctx.o365audit.Actions instanceof List)) {
           ctx.o365audit.Actions = [ctx.o365audit.Actions];
         }
+        /*
```
efd6: Painless allows `//` comments. These are lighter on the page.

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -132,11 +132,16 @@ processors:
...
+        * Actions contains both a human readable `QueryTime` using AM/PM and an ISO8601 format `QueryTime`
+        * We remove the AM/PM containing `QueryTime` to avoid duplicate field errors on flattening.
+        */
+        def regex = /,"QueryTime":"[0-9\/]+\s[0-9]+:[0-9]+:[0-9]+\s[AP]M"|"QueryTime":"[0-9\/]+\s[0-9]+:[0-9]+:[0-9]+\s[AP]M",/;
```
efd6:
```suggestion
        def queryTimePattern = /,"QueryTime":"[0-9\/]+\s[0-9]+:[0-9]+:[0-9]+\s[AP]M"|"QueryTime":"[0-9\/]+\s[0-9]+:[0-9]+:[0-9]+\s[AP]M",/;
```
and below

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1770,6 +1796,75 @@ processors:
       copy_from: o365audit.ApplicationDisplayName
       tag: set_application_name
       ignore_empty_value: true
+
+  # ExchangeItemAggregated Schema
+  - append:
+      field: event.type
+      value: access
+      if: ctx.o365audit?.RecordType != null && ctx.o365audit?.RecordType == "50"
```
efd6:
```suggestion
      if: ctx.o365audit?.RecordType == "50"
```
If `ctx.o365audit?.RecordType` is "50" it cannot be `null`, so the first clause is redundant.

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1770,6 +1796,75 @@ processors:
...
+  - append:
+      field: event.category
+      value: email
+      if: ctx.o365audit?.RecordType != null && ctx.o365audit?.RecordType == "50"
```
efd6:
```suggestion
      if: ctx.o365audit?.RecordType == "50"
```

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1770,6 +1796,75 @@ processors:
...
+  - rename:
+      field: o365audit.Messages
+      target_field: o365audit.ExchangeAggregatedMessages
+      tag: rename_messages_exchange
+      description: 'Move generic Messages field to the ExchangeAggregatedMessages field type'
+      if: ctx.o365audit?.Messages != null && ctx.o365audit?.RecordType != null && ctx.o365audit?.RecordType == "50"
```
efd6:
```suggestion
      if: ctx.o365audit?.Messages != null && ctx.o365audit.RecordType == "50"
```
… same, but if `ctx.o365audit?.Messages` is not `null`, `ctx.o365audit` is not `null`.

packages/o365/changelog.yml
```diff
@@ -1,4 +1,20 @@
...
+    - description: >-
+        Fix flattening errors in `Action` List items due to duplicate `QueryTime` fields by removing duplicate field.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/15699
+    - description: >- 
+        Fixes undefined errors by adding fields `ActorInfoString`, `OperationCount`, `TokenObjectId`, `TokenTenantId`
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/15699
+    - description: >-
+        Fixes errors due to SizInBytes fields in `Messages` and `Folders` structures previously imported as long
```
efd6:
```suggestion
        Fixes errors due to SizeInBytes fields in `Messages` and `Folders` structures previously imported as long 
```

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1770,6 +1796,75 @@ processors:
...
+  - script:
+      tag: convert_exchange_message_size_to_long
+      if: ctx.o365audit?.ExchangeAggregatedMessages != null
+      lang: painless
+      source: |
+        for (def i = 0; i < ctx.o365audit.ExchangeAggregatedMessages.length; i++) {
+          if (ctx.o365audit.ExchangeAggregatedMessages[i].MessageItems != null) {
```
efd6: I would prefer this to be inverted with a conditional `continue`. This moves the for loop left. (also below)

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1770,6 +1796,75 @@ processors:
...
+      tag: rename_folders_exchange
+      description: 'Move generic Folders field to the O365 ExchangeAggregatedFolders field type'
+      if: ctx.o365audit?.Folders != null && ctx.o365audit?.RecordType != null && ctx.o365audit?.RecordType == "50"
+  - script:
+      tag: convert_exchange_folder_size_to_long
+      if: ctx.o365audit?.ExchangeAggregatedFolders != null
+      lang: painless
+      source: |
+        for (def i = 0; i < ctx.o365audit.ExchangeAggregatedFolders.length; i++) {
+          if (ctx.o365audit.ExchangeAggregatedFolders[i].FolderItems != null) {
```
efd6: invert and dedent for loop

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1770,6 +1796,75 @@ processors:
...
+      tag: remove_messages_field
+      if: ctx.o365audit?.Messages != null
+      description: 'remove o365audit.Messages if we have not explicitly renamed them based on record type'
+
+  - rename:
+      field: o365audit.Folders
+      target_field: o365audit.ExchangeAggregatedFolders
+      tag: rename_folders_exchange
+      description: 'Move generic Folders field to the O365 ExchangeAggregatedFolders field type'
+      if: ctx.o365audit?.Folders != null && ctx.o365audit?.RecordType != null && ctx.o365audit?.RecordType == "50"
```
efd6:
```suggestion
      if: ctx.o365audit?.Folders != null && ctx.o365audit.RecordType == "50"
```

packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1770,6 +1796,75 @@ processors:
...
+              }
+            }
+          }
+        }
+
+  - remove:
+      field: o365audit.Folders
+      tag: remove_folders_field
+      if: ctx.o365audit?.Folders != null
+      description: 'Remove o365audit.Folders if we have not explicitly renamed them based on record type'
```
efd6: I believe this is a breaking change, as is the case with the rename and the equivalent changes affecting `o365audit.Messages`.

StacieClark-Elastic: > I believe this is a breaking change, as is the case with the rename and the equivalent changes affecting `o365audit.Messages`.

That was a mistake. I removed the remove processors.

packages/o365/changelog.yml
StacieClark-Elastic: I understand what you are saying about ingesting invalid inputs not being expected behavior. The "bugfix" for the Actions is that it was missed when the Data field was updated to tolerate invalid input and this oversight keeps the event from being ingested. From a customers standpoint, I wanted them to understand that this fix stops a set of error messages. An enhancement they can shrug off. To stop the error messages they need to apply this version. I changed the labels.

packages/o365/changelog.yml
```diff
@@ -1,4 +1,20 @@
 # newer versions go on top
+- version: "2.31.1"
```
efd6:
```suggestion
- version: "2.32.0"
```
because enhancement

packages/o365/manifest.yml
```diff
@@ -1,6 +1,6 @@
 name: o365
 title: Microsoft Office 365
-version: "2.31.0"
+version: "2.31.1"
```
efd6:
```suggestion
version: "2.32.0"
```

PR #15700, opened 2025-10-20
Update unique keys for latest issues transform to catch all updates

clement-fouque:
## Proposed commit message

This pull request updates the configuration for the GitHub issues Elasticsearch transform to improve how the latest state of issues is tracked and stored. The main focus is on making the transform more robust by using more reliable unique keys and bumping the transform version.

The previous transform wasn't taking into account all updates (e.g. adding a label to an issue). Since the field `github.issues.updated_at` changes each time the issue is updated, this field covers more use cases that the initial `github.issues.created_at` field. I've also changed the unique issue identifier from url to `github.issues.id`.

### Key changes:

**Transform logic improvements:**
* Changed the `unique_key` fields in `latest_issues/transform.yml` to use `github.issues.id` and `github.issues.updated_at` instead of repository and URL fields. This ensures that the transform identifies the latest issue state more reliably and efficiently.

**Versioning:**
* Updated the `_meta.fleet_transform_version` from `1.0.0` to `2.0.0` to indicate a breaking change and trigger reinstallation and restart of the transform.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


clement-fouque:
I confirmed locally that the transform is working as expected.
image
image

packages/github/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.18.0"
+  changes:
+    - description: Update unique keys for latest issues transform
+      type: bugfix
```
efd6:
```suggestion
- version: "2.17.1"
  changes:
    - description: Update unique keys for latest issues transform.
      type: bugfix
```
Bugfixes bump patch, but you've noted in the description and in the transform version that this is breaking. ISTM that this should be reflected in the package version as well; if a subcomponent has a breaking change then the component has a breaking change, unless the component hides it, which I do not believe is the case here.

efd6 (commented):
This is not a refactor; it changes behaviour. Can we have a clearer title to explain what is being done?

clement-fouque:
I’ve updated the title and PR description, and changed the version from ⁠2.18.0 to ⁠2.17.1. Initially, I thought this was a breaking change. However, since the transform is deleted and recreated, it’s not a breaking change after all.

@efd6, could you please review? Thanks.

efd6 (approved):

Issue #11338, opened 2024-10-05
[Akamai] - Add option for a CEL based agent and integrate automatic error recovery.


ShourieG:
The current Akamai integration works with the httpjson input, this however puts certain limitations on the flexibility and recoverability of the integration in certain scenarios. For example, if the agent fails and restarts after sometime there is a possibility that the offset used internally to paginate might get expired, this would cause continuous failures unless the registry is cleaned and ingestion resumes. 

To handle such scenarios and make the integration have some auto-recoverability it would be in our best interest to add an option for the CEL input. The httpjson would continue to be supported with the current limitations.

ShourieG:
cc: @narph, @andrewkroh

singhvirohan:
@andrewkroh @narph I wanted to follow up on this ticket to see if there are any plans to get this addressed? 
The workarounds from [this KB](https://support.elastic.co/knowledge/3f5607e5) aren't sufficient because:

`While it initially restores continuous log flow, the issue reoccurs after some time, requiring a manual restart of the integration. Unfortunately, manual intervention is not a sustainable solution for our environment.`

narph:
@ShourieG, we should avoid the second option (_Add an option for the CEL input. The httpjson would continue to be supported with the current limitations._) as we don't have a deprecation path and we don't want to maintain 2 solutions for a long period of time.
@efd6 can you confirm that moving to CEL is the only option in this case?

efd6:
@narph In HTTPJSON, [status codes at or above 400 result in response errors and request failures](https://github.com/elastic/beats/blob/e1d92ebda55f492b0be4ced020abeb4347469685/x-pack/filebeat/input/httpjson/request.go#L279-L286) which are now no longer available to the templating logic, so yes, this would require porting to CEL to be able to handle the fallback.

efd6:
Looking further into this, I think that it may be possible to avoid the error in the first place (my answer above is for the smaller scope question of whether we can handle a 400 status code in HTTPJSON); if we know the API's constraints on the reach back, we can incorporate that into the [`url.params.from`](https://github.com/elastic/integrations/blob/51fe49d1d74eaf0e0d39035c155b9e774c379b5f/packages/akamai/data_stream/siem/agent/stream/httpjson.yml.hbs#L22) template so that we do not reach back past what the API allows. This should recover invalid cursors naturally.

[...]

PR #15713, opened 2025-10-22
[Akamai] - Migrated SIEM data stream from HTTPJSON to CEL input

ShourieG:
## Type of change
- Enhancement
- Breaking Change

## Proposed commit message
```
Migrated SIEM data stream from HTTPJSON to CEL input with 
with necessary reworks and working system tests. Updated 
minimum stack version to 8.18 to allow usage of required CEL functions. 
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/integrations/issues/11338
- Relates https://github.com/elastic/integrations/issues/11338

### DIFF
--- a/packages/akamai/changelog.yml
+++ b/packages/akamai/changelog.yml
@@ -3,11 +3,13 @@
   changes:
     - description: | 
         Migrated SIEM data stream from HTTPJSON to CEL.
-      type: breaking-change
+      type: enhancement
       link: https://github.com/elastic/integrations/pull/15713
     - description: |
         Credentials will likely need to be re-configured since the integration has been
         updated to use the new CEL input.
+      type: breaking-change
+      link: https://github.com/elastic/integrations/pull/15713
 - version: "2.28.2"
   changes:
     - description: Remove empty HTTP message headers placeholder.


packages/akamai/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.29.0"
+  changes:
+    - description: Migrated siem data stream from Httpjson to CEL.
```
efd6:
```suggestion
    - description: Migrated siem data stream from HTTP JSON to CEL.
```

packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,142 @@
...
+redact:
+  fields:
+    - client_secret
+    - access_token
+    - client_token  
+
+program: |
+  state.with(
+    (
+      state.?cursor.?recovery_mode.orValue(false) ?
```
efd6:
```suggestion
      state.?cursor.recovery_mode.orValue(false) ? 
```
similar throughout; after a `.?` operator, all the subsequent accesses are on optional types, so after the first the `?` is redundant.

packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,142 @@
...
+                  ?"offset": params.?offset.orValue("") != "" ? optional.of([string(params.offset)]) : optional.none(),
+                }.format_query()).parse_url().as(u,
+                  (
+                    "GET\t" +
+                    string(u.Scheme) + "\t" +
+                    string(u.Host) + "\t" +
+                    string(u.Path) + "?" +
+                    string(u.RawQuery) + 
+                    "\t\t\t" + sig_base
+                  ).as(to_sign,
```
efd6: use `sprintf`

packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
efd6, line 1: run `celfmt` on this

packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,142 @@
...
+            "Authorization": [auth_header]
+          }
+        }).do_request().as(resp,
+          resp.StatusCode == 200 ?
+            string(resp.Body).split("\n").filter(line, line != "").as(lines, {
+            "events": lines.map(line, { "message": line }),
+            "cursor": {
+              ?"last_offset": lines.size() > 0 ? 
+                lines[lines.size() - 1].decode_json().as(lastEvent,
+                  has(lastEvent.offset) ? optional.of(lastEvent.offset) : optional.none()
```
efd6:
```suggestion
                  lastEvent.?offset
```

ShourieG:
@efd6, addressed all the changes

packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,159 @@
...
+                  ?"offset": (params.?offset.orValue("") != "") ? optional.of([string(params.offset)]) : optional.none(),
+                }.format_query()
+              ).parse_url().as(u,
+                sprintf(
+                  "GET\t%s\t%s\t%s?%s\t\t\t%s",
+                  [
+                    string(u.Scheme),
+                    string(u.Host),
+                    string(u.Path),
+                    string(u.RawQuery),
```
efd6:
```suggestion
                    u.Scheme,
                    u.Host,
                    u.Path,
                    u.RawQuery,
```

packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,159 @@
...
+        uuid().as(nonce,
+          sprintf(
+            "EG1-HMAC-SHA256 client_token=%s;access_token=%s;timestamp=%s;nonce=%s;",
+            [state.client_token, state.access_token, timestamp, nonce]
+          ).as(sig_base,
+            base64(hmac(timestamp, "sha256", bytes(state.client_secret))).as(sig_key,
+              (
+                state.url.trim_right("/") + "?" + {
+                  "limit": [string(state.event_limit)],
+                  ?"from": (params.?from.orValue("") != "") ? optional.of([string(params.from)]) : optional.none(),
```
efd6: Why are we doing `params.?from.orValue("") != ""`? AFAICS `params.from` will never be ""; it will eigher be a unix time or absent. Same for `to`. `offset` is less clear, but that's under our control, coming from `cursor.last_offset` which is only ever set to the last event's `offset` (which I guess could conceivably be "", but then it should not be stored).

packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,159 @@
...
+        )
+      ).as(auth_header,
+        request(
+          "GET",
+          state.url.trim_right("/") + "?" + {
+            "limit": [string(state.event_limit)],
+            ?"from": (params.?from.orValue("") != "") ? optional.of([string(params.from)]) : optional.none(),
+            ?"to": (params.?to.orValue("") != "") ? optional.of([string(params.to)]) : optional.none(),
+            ?"offset": (params.?offset.orValue("") != "") ? optional.of([string(params.offset)]) : optional.none(),
+          }.format_query()
```
efd6: I think we should be using the value that we obtained above, purely for efficiency (the query will be canonicalised in order, so it's not a correctness issue).

ShourieG:
@efd6, addressed all the comments. Let's approve this once we have successfully tested with a live api, there seems to be some outstanding issues still. Let me know if the current changes are ok or not.

packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,170 @@
...
+                            :
+                              string(resp.Status) + " (" + string(resp.StatusCode) + ")"
+                          ),
+                        },
+                      },
+                      {
+                        "info": {
+                          "message": "Initiating recovery mode due to offset expiry.",
+                        },
+                      },
```
efd6: This event is not necessary.

packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,165 @@
...
+        }
+      :
+        {
+          "from": max(int(now - duration(state.initial_interval)), int(now - duration("12h"))),
+          "to": int(now - duration("1m")),
+        }
+    ).as(params,
+      (
+        state.url.trim_right("/") + "?" + {
+          "limit": [string(state.event_limit)],
```
chrisberkhout, line 55: There's no default `event_limit` value in the manifest so maybe this will end up as `"limit": ["null"]`?

efd6: Good catch, this would be an eval error since there is no overload for a null to string conversion.

ShourieG: @chrisberkhout, added a default limit value

PR #15722, opened 2025-10-22
Add filtering out of documents with error.message from latest indexes

alexreal1314:
## Proposed commit message

Purpose of this PR is to filter out documents who contain error.message in the source indexes for all supported 3p integration and our native integration. This is in order to improve UI experience by filtering them out from the cdr workflows.

## Checklist

## Author's Checklist

- [ ] 

1. install one of the changed integrations - for example with `wiz`
2. toggle `Collect Wiz logs via API`
3. type in the wrong client id or client secret
4. add an agent and let the integration ingest data
5. vulnerability source index `logs-wiz.vulnerability-default` should contain a document with error.message field but no documents should be in the dest index `security_solution-wiz.vulnerability_latest-v2`.

## Screenshots

source index - wiz:
image

dest index - wiz:
image

### DIFF
--- a/packages/aws/changelog.yml
+++ b/packages/aws/changelog.yml
@@ -9,9 +9,6 @@
     - description: Update the AWS CloudWatch documentation.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/15737
-    - description: Update transform to filter out document containing an error.message from AWS Config, AWS Inspector, and AWS Security Hub latest indexes.
-      type: bugfix
-      link: https://github.com/elastic/integrations/pull/15722
 - version: "4.3.0"
   changes:
     - description: Improve documentation to align with new guidelines.
--- a/packages/aws/elasticsearch/transform/latest_cdr_misconfigurations/transform.yml
+++ b/packages/aws/elasticsearch/transform/latest_cdr_misconfigurations/transform.yml
@@ -32,4 +32,4 @@ _meta:
   managed: true
   # Bump this version to delete, reinstall, and restart the transform during package.
   # Version bump is needed if there is any code change in transform.
-  fleet_transform_version: 0.2.0
+  fleet_transform_version: 0.2.1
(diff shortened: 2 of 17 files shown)


efd6:
Is there an issue for this explaining why it's necessary?

alexreal1314:
> Is there an issue for this explaining why it's necessary?

This PR closes this [issue](https://github.com/elastic/security-team/issues/11031), the intention it to maintain clean latest indexes and prevent documents with errors reaching them.

efd6:
I think this should either be conditional on user configuration or done dynamically rather than my mutating the index.

alexreal1314:
> I think this should either be conditional on user configuration or done dynamically rather than my mutating the index.

AFAIK we don't mutate any index, just filtering out findings containing errors messages.

@maxcold do we need to make the filtering configurable? didn't see such requirement.

maxcold:
@alexreal1314 @efd6 I don't see a reason to make it configurable tbh. The sole reason the transforms exist today for these integrations is that we want the data from the latest transform destination index to be used in the Findings page in Security Solution. This change only affects the destination index, not a source index which is used by the users for their needs (dashbaords, alerts, whatever usercase pre tranform existence). Destination index is only for our internal use really and we don't want error documents to show up on the Findings page

maxcold (approved):

efd6:
@maxcold That's reasonable.

packages/aws/changelog.yml
```diff
@@ -1,9 +1,17 @@
...
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/15722
 - version: "4.3.1"
   changes:
     - description: Update the AWS CloudWatch documentation.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/15737
+    - description: Update transform to filter out document containing an error.message from AWS Config, AWS Inspector, and AWS Security Hub latest indexes.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/15722
```
efd6: Why is this here?

alexreal1314: @efd6 could you elaborate please?

efd6: This text is an addition to a previous version that was added in #15737 and is a duplicate of the text above.

alexreal1314: @efd6 fixed, thanks.

PR #15725, opened 2025-10-22
[Neon Cyber] New Elastic integration neon_cyber for the Neon Cyber platform

codypierce:
## Proposed commit message

Changes made

- Initial package boilerplate
- Readme updated to spec
- Changelog updated to spec
- Event data stream for CEL processing from Neon events API
- Field definitions for events
- Pipeline processor definitions
- Sample log included for supported events
- Detections data stream for CEL processing from Neon detections API
- Field definitions for detections
- Sample log included for supported detections
- Pipeline, policy, and system tests for new data streams

## Checklist

## Author's Checklist

- [ ] 

1. Run pipeline and system tests

## Related issues

- N/A

### DIFF
--- a/packages/neon_cyber/data_stream/detections/_dev/test/pipeline/test-detections.log-expected.json
+++ b/packages/neon_cyber/data_stream/detections/_dev/test/pipeline/test-detections.log-expected.json
@@ -70,6 +70,11 @@
                     "url": "https://support.notanapp.com"
                 }
             },
+            "url": {
+                "domain": "support.notanapp.com",
+                "original": "https://support.notanapp.com",
+                "scheme": "https"
+            },
             "user_agent": {
                 "name": "chrome",
                 "original": "Mozilla/5.0 (X11; Linux aarch64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36",
@@ -140,6 +145,11 @@
                     "tab_id": 12345,
                     "url": "https://click.this.link.banckcorp.com"
                 }
+            },
+            "url": {
+                "domain": "click.this.link.banckcorp.com",
+                "original": "https://click.this.link.banckcorp.com",
+                "scheme": "https"
             },
             "user_agent": {
                 "name": "chrome",
@@ -215,6 +225,12 @@
                     "total_bytes": 31337,
                     "url": "https://example.com/docs/support"
                 }
+            },
+            "url": {
+                "domain": "example.com",
+                "original": "https://example.com/docs/support",
+                "path": "/docs/support",
+                "scheme": "https"
             },
             "user_agent": {
                 "name": "chrome",
--- a/packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
@@ -120,6 +120,10 @@ processors:
           target_field: neon_cyber.detections.url
           ignore_missing: true
           tag: rename_detection_url
+    - uri_parts:
+          field: neon_cyber.detections.url
+          ignore_missing: true
+          tag: uri_parts_detections_url
     - rename:
           field: json.filename
           target_field: neon_cyber.detections.filename
--- a/packages/neon_cyber/data_stream/events/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/neon_cyber/data_stream/events/elasticsearch/ingest_pipeline/default.yml
@@ -185,6 +185,10 @@ processors:
           target_field: neon_cyber.events.url
           ignore_missing: true
           tag: rename_events_url
+    - uri_parts:
+          field: neon_cyber.detections.url
+          ignore_missing: true
+          tag: uri_parts_detections_url
     - rename:
           field: json.agent
           target_field: neon_cyber.events.agent
--- a/packages/neon_cyber/manifest.yml
+++ b/packages/neon_cyber/manifest.yml
@@ -11,7 +11,7 @@ categories:
     - edr_xdr
 conditions:
     kibana:
-        version: '^8.18.0 || ^9.0.0'
+        version: '^8.17.0 || ^9.0.0'
     elastic:
         subscription: 'basic'
 screenshots:
--- a/packages/neon_cyber/data_stream/detections/_dev/test/policy/test-default.expected
+++ b/packages/neon_cyber/data_stream/detections/_dev/test/policy/test-default.expected
@@ -25,31 +25,34 @@ inputs:
                 ).do_request().as(resp, (resp.StatusCode == 200) ?
                   resp.Body.decode_json().as(body,
                     {
-                      "events": (has(body.data) && body.data.size() > 0) ?
-                        body.data.map(e,
-                          {
-                            "message": e.encode_json(),
-                          }
-                        )
-                      :
-                        []
+                      "events": dyn(
+                        (has(body.data) && body.data.size() > 0) ?
+                          body.data.map(e,
+                            {
+                              "message": e.encode_json(),
+                            }
+                          )
+                        :
+                          []
+                      )
                     }
                   )
                 :
                   {
-                    "events": {
-                      "error": {
-                        "code": string(resp.StatusCode),
-                        "id": string(resp.Status),
-                        "message": "GET " + state.url + (
-                          (size(resp.Body) != 0) ?
-                            string(resp.Body)
-                          :
-                            string(resp.Status) + " (" + string(resp.StatusCode) + ")"
-                        ),
-                      },
-                    },
-                    "want_more": false,
+                    "events": dyn(
+                      {
+                        "error": {
+                          "code": string(resp.StatusCode),
+                          "id": string(resp.Status),
+                          "message": "GET " + state.url.trim_right("/") + "/v1/detections: " + (
+                            (size(resp.Body) != 0) ?
+                              string(resp.Body)
+                            :
+                              string(resp.Status) + " (" + string(resp.StatusCode) + ")"
+                          ),
+                        },
+                      }
+                    )
                   }
                 )
               )
--- a/packages/neon_cyber/data_stream/detections/sample_event.json
+++ b/packages/neon_cyber/data_stream/detections/sample_event.json
@@ -1,11 +1,11 @@
 {
     "@timestamp": "2025-10-10T17:09:32.988Z",
     "agent": {
-        "ephemeral_id": "6c04cb50-36c5-43f1-94f0-77bb00ca4e5b",
-        "id": "1ba78911-4b81-4991-8df6-21fd231ccb6d",
-        "name": "elastic-agent-92421",
+        "ephemeral_id": "0ea008fb-25c6-4e96-a0cb-580156905c33",
+        "id": "b328b57b-9887-4a4a-a714-81cc1c317559",
+        "name": "elastic-agent-75127",
         "type": "filebeat",
-        "version": "9.1.3"
+        "version": "8.17.0"
     },
     "client": {
         "as": {
@@ -18,11 +18,8 @@
             "city_name": "Brighton",
             "country_name": "GB",
             "location": {
-                "coordinates": [
-                    -0.11100003495812416,
-                    11.11109996214509
-                ],
-                "type": "Point"
+                "lat": 11.1111,
+                "lon": -0.111
             },
             "postal_code": "BN3",
             "region_iso_code": "BNH",
@@ -34,16 +31,16 @@
     },
     "data_stream": {
         "dataset": "neon_cyber.detections",
-        "namespace": "32299",
+        "namespace": "11517",
         "type": "logs"
     },
     "ecs": {
         "version": "8.17.0"
     },
     "elastic_agent": {
-        "id": "1ba78911-4b81-4991-8df6-21fd231ccb6d",
+        "id": "b328b57b-9887-4a4a-a714-81cc1c317559",
         "snapshot": false,
-        "version": "9.1.3"
+        "version": "8.17.0"
     },
     "event": {
         "agent_id_status": "verified",
@@ -51,9 +48,8 @@
             "network"
         ],
         "dataset": "neon_cyber.detections",
-        "ingested": "2025-11-05T22:48:29Z",
+        "ingested": "2025-11-06T04:11:02Z",
         "kind": "alert",
-        "module": "neon_cyber",
         "original": "{\"agent\":\"1.2.4\",\"arch\":\"arm64\",\"asn\":9105,\"asn_isp\":\"TalkTalk\",\"city\":\"Brighton\",\"client_id\":\"d19143bd-3a34-4c7e-886b-87f643df4835\",\"country\":\"GB\",\"deployment_id\":\"ec080215-913c-490c-bb81-a1baa311ee45\",\"description\":null,\"detection_subtype\":null,\"detection_timestamp\":\"2025-10-10T17:09:32.988+00:00\",\"detection_type\":\"phishing\",\"display\":\"Chrome\",\"id\":\"c24b2526-11e4-48bf-af45-f6837232037f\",\"incognito\":false,\"inserted_at\":\"2025-10-13T18:13:15.88219+00:00\",\"ip\":\"192.0.2.1\",\"ip_latitude\":11.1111,\"ip_longitude\":-0.111,\"latitude\":11.1111,\"longitude\":-0.111,\"name\":\"chrome\",\"os\":\"linux\",\"postal_code\":\"BN3\",\"region_code\":\"BNH\",\"region_name\":\"Brighton and Hove\",\"registration_id\":\"a00db187-ad55-4074-8a84-d085555abe29\",\"source\":\"Phishing AI\",\"tab_id\":12345,\"ua\":\"Mozilla/5.0 (X11; Linux aarch64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36\",\"updated_at\":\"2025-10-13T18:13:15.88219+00:00\",\"url\":\"https://click.this.link.banckcorp.com\",\"version\":\"134.0.0.0\"}",
         "type": [
             "info"
@@ -63,11 +59,8 @@
         "architecture": "arm64",
         "geo": {
             "location": {
-                "coordinates": [
-                    -0.11100003495812416,
-                    11.11109996214509
-                ],
-                "type": "Point"
+                "lat": 11.1111,
+                "lon": -0.111
             }
         },
         "os": {
@@ -98,6 +91,11 @@
         "forwarded",
         "neon_cyber-detections"
     ],
+    "url": {
+        "domain": "click.this.link.banckcorp.com",
+        "original": "https://click.this.link.banckcorp.com",
+        "scheme": "https"
+    },
     "user_agent": {
         "name": "chrome",
         "original": "Mozilla/5.0 (X11; Linux aarch64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36",
--- a/packages/neon_cyber/data_stream/events/_dev/test/policy/test-default.expected
+++ b/packages/neon_cyber/data_stream/events/_dev/test/policy/test-default.expected
@@ -25,31 +25,34 @@ inputs:
                 ).do_request().as(resp, (resp.StatusCode == 200) ?
                   resp.Body.decode_json().as(body,
                     {
-                      "events": (has(body.data) && body.data.size() > 0) ?
-                        body.data.map(e,
-                          {
-                            "message": e.encode_json(),
-                          }
-                        )
-                      :
-                        []
+                      "events": dyn(
+                        (has(body.data) && body.data.size() > 0) ?
+                          body.data.map(e,
+                            {
+                              "message": e.encode_json(),
+                            }
+                          )
+                        :
+                          []
+                      )
                     }
                   )
                 :
                   {
-                    "events": {
-                      "error": {
-                        "code": string(resp.StatusCode),
-                        "id": string(resp.Status),
-                        "message": "GET " + state.url + (
-                          (size(resp.Body) != 0) ?
-                            string(resp.Body)
-                          :
-                            string(resp.Status) + " (" + string(resp.StatusCode) + ")"
-                        ),
-                      },
-                    },
-                    "want_more": false,
+                    "events": dyn(
+                      {
+                        "error": {
+                          "code": string(resp.StatusCode),
+                          "id": string(resp.Status),
+                          "message": "GET " + state.url.trim_right("/") + "/v1/events: " + (
+                            (size(resp.Body) != 0) ?
+                              string(resp.Body)
+                            :
+                              string(resp.Status) + " (" + string(resp.StatusCode) + ")"
+                          ),
+                        },
+                      }
+                    )
                   }
                 )
               )
--- a/packages/neon_cyber/data_stream/events/sample_event.json
+++ b/packages/neon_cyber/data_stream/events/sample_event.json
@@ -1,11 +1,11 @@
 {
     "@timestamp": "2025-10-12T21:39:13.241Z",
     "agent": {
-        "ephemeral_id": "8d5e2f33-8d2d-4686-af39-b75382a9cd8e",
-        "id": "d843e021-3d85-4d94-a94e-e25eb1b165f3",
-        "name": "elastic-agent-85201",
+        "ephemeral_id": "4557466d-fe9d-488d-a8e8-2598d68ac9e8",
+        "id": "9ccd0f38-920a-4eb6-ab60-4eb6163b537c",
+        "name": "elastic-agent-78481",
         "type": "filebeat",
-        "version": "9.1.3"
+        "version": "8.17.0"
     },
     "client": {
         "as": {
@@ -18,11 +18,8 @@
             "city_name": "Brighton",
             "country_name": "GB",
             "location": {
-                "coordinates": [
-                    -0.11100003495812416,
-                    11.11109996214509
-                ],
-                "type": "Point"
+                "lat": 11.1111,
+                "lon": -0.111
             },
             "postal_code": "BN3",
             "region_iso_code": "BNH",
@@ -34,16 +31,16 @@
     },
     "data_stream": {
         "dataset": "neon_cyber.events",
-        "namespace": "53021",
+        "namespace": "67591",
         "type": "logs"
     },
     "ecs": {
         "version": "8.17.0"
     },
     "elastic_agent": {
-        "id": "d843e021-3d85-4d94-a94e-e25eb1b165f3",
+        "id": "9ccd0f38-920a-4eb6-ab60-4eb6163b537c",
         "snapshot": false,
-        "version": "9.1.3"
+        "version": "8.17.0"
     },
     "event": {
         "agent_id_status": "verified",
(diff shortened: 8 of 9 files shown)


packages/neon_cyber/LICENSE.txt
efd6, line 1: We don't need to add this file, it will be automatically generated.

packages/neon_cyber/_dev/build/build.yml
```diff
@@ -0,0 +1,3 @@
+dependencies:
+  ecs:
+    reference: git@v8.11.0
```
efd6:
```suggestion
    reference: git@v8.17.0
```

packages/neon_cyber/_dev/build/docs/README.md
```diff
@@ -0,0 +1,45 @@
+{{- generatedHeader }}
```
efd6: What is this for?

packages/neon_cyber/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "1.0.0"
+  changes:
+    - description: Initial version of the package
```
efd6:
```suggestion
    - description: Initial version of the package.
```

packages/neon_cyber/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "1.0.0"
```
efd6: Happy with GA from the outset?

packages/neon_cyber/data_stream/detections/_dev/test/pipeline/test-detections.log-expected.json
```diff
@@ -0,0 +1,238 @@
...
+                    "info"
+                ]
+            },
+            "host": {
+                "architecture": "arm64",
+                "geo": {
+                    "location": [
+                        {
+                            "lat": "ctx.json.latitude",
+                            "lon": "ctx.json.longitude"
```
efd6: This does not look right.

packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,82 @@
...
+        resp.Body.decode_json().as(body,
+          {
+            "events": (has(body.data) && body.data.size() > 0) ?
+              body.data.map(e,
+                {
+                  "message": e.encode_json(),
+                }
+              )
+            :
+              [{"message": "retry"}]
```
efd6: What is this for? We are not retrying at all, so this is not needed. Same in the other CEL programs.

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,260 @@
+---
+description: Pipeline for processing Neon detections.
+processors:
+    - drop:
+          description: Ignore retry placeholder message.
+          if: ctx.message == "retry"
```
efd6: This is not needed.

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,260 @@
...
+          field: json.arch
+          target_field: host.architecture
+          ignore_missing: true
+    - rename:
+          field: json.os
+          target_field: host.os.platform
+          ignore_missing: true
+    - append:
+          field: host.geo.location
+          value: { 'lat': ctx.json.latitude, 'lon': ctx.json.longitude }
```
efd6: Variable interpolation is not a thing here. Also below. If you want to do this, I'd suggest using a painless script, but I'm not sure how an array of `geo.location` is going to work out with searches.

packages/neon_cyber/data_stream/detections/_dev/test/policy/test-default.yml
```diff
@@ -0,0 +1,13 @@
+vars:
+  url: http://host.tld
```
efd6: What is this for?

packages/neon_cyber/data_stream/detections/manifest.yml
```diff
@@ -0,0 +1,46 @@
...
+streams:
+    - input: cel
+      title: 'Collect Neon Cyber Detections'
+      description: |-
+          Collect workforce detections from the Neon Cyber Platform
+      template_path: cel.yml.hbs
+      vars:
+          - name: resource_url
+            type: text
+            title: 'Resource URL'
```
efd6: I'd suggest that the word "resource" be replaced with something more locally meaningful (throughout). The input uses this, but that is because it applies to a generic resource. That's not the case here.

packages/neon_cyber/data_stream/detections/_dev/test/policy/test-default.yml
```diff
@@ -0,0 +1,13 @@
...
+  api_token: test_api_token
```
efd6: It seems odd to have the API token defined at the package level, but the host be defined at the data stream level. What is the reason for that?

packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,82 @@
...
+              "id": string(resp.Status),
+              "message": "GET " + state.url + (
+                (size(resp.Body) != 0) ?
+                  string(resp.Body)
+                :
+                  string(resp.Status) + " (" + string(resp.StatusCode) + ")"
+              ),
+            },
+          },
+          "want_more": false,
```
efd6: This is not needed since `want_more` is never set to true. Same below.

packages/neon_cyber/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,327 @@
+---
+description: Pipeline for processing Neon events.
+processors:
+    - drop:
+          description: Ignore retry placeholder message.
+          if: ctx.message == "retry"
```
efd6: This is not needed.

packages/neon_cyber/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,327 @@
...
+    - set:
+          field: ecs.version
+          value: 8.11.0
```
efd6:
```suggestion
          value: 8.17.0
```

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
efd6, line 1: Processors should have tags to allow failures to be identified in the error message. Also below.

packages/neon_cyber/manifest.yml
```diff
@@ -0,0 +1,92 @@
...
+            description: Collecting Neon Cyber events and detections via API
+            vars:
+                - name: url
+                  type: url
+                  title: URL
+                  default: https://api.neoncyber.io/v1
+                  description: Base URL of the Neon Cyber API
+                  multi: false
+                  required: true
+                  show_user: true
```
efd6, line 50: This is not ever used AFAICS.

packages/neon_cyber/manifest.yml
```diff
@@ -0,0 +1,92 @@
...
+                - name: api_token
+                  type: password
+                  title: API Token
+                  description: Neon Cyber developer API token
+                  multi: false
+                  required: true
+                  show_user: true
+                  secret: true
```
efd6, line 59: Is this shared for all endpoints?

packages/neon_cyber/manifest.yml
```diff
@@ -0,0 +1,92 @@
...
+                - name: ssl
+                  type: yaml
+                  title: SSL Configuration
+                  description: SSL configuration options. See [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/configuration-ssl.html#ssl-common-config) for details.
+                  multi: false
+                  required: false
+                  show_user: false
+owner:
+    github: elastic/security-service-integrations
+    type: elastic
```
efd6: This likely should be either "community" or "partner" depending on the arrangement here.

packages/neon_cyber/_dev/build/docs/README.md
codypierce: It was in the template and adds the auto generation "Do not edit" in the rendered README.md

packages/neon_cyber/changelog.yml
codypierce: I'll be more conservative with 0.1.0

packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs
codypierce: removed retry message

packages/neon_cyber/data_stream/detections/_dev/test/pipeline/test-detections.log-expected.json
codypierce: Thanks, converted to float

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
codypierce: removed

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
codypierce: My mistake, fixed in both

packages/neon_cyber/data_stream/detections/_dev/test/policy/test-default.yml
codypierce: I thought this was a placeholder for the package vars for policy tests?

packages/neon_cyber/data_stream/detections/manifest.yml
codypierce: Updated

packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs
codypierce: done

packages/neon_cyber/data_stream/events/elasticsearch/ingest_pipeline/default.yml
codypierce: thanks

packages/neon_cyber/manifest.yml
codypierce: Yes. I understand now, removing package URL for data stream resource url.

packages/neon_cyber/data_stream/detections/_dev/test/policy/test-default.yml
codypierce: I'm moving package url to data stream resource_url

packages/neon_cyber/manifest.yml
codypierce: yes

packages/neon_cyber/manifest.yml
codypierce: Changing to partner

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
codypierce: done

packages/neon_cyber/_dev/build/docs/README.md
efd6: We don't need it.

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,315 @@
...
+              boolean dropEmptyFields(Object object) {
+                  if (object == null || object == '') {
+                  return true;
+                  } else if (object instanceof Map) {
+                  ((Map) object).values().removeIf(value -> dropEmptyFields(value));
+                  return (((Map) object).size() == 0);
+                  } else if (object instanceof List) {
+                  ((List) object).removeIf(value -> dropEmptyFields(value));
+                  return (((List) object).length == 0);
+                  }
```
efd6, line 310:
```suggestion
                  if (object == null || object == '') {
                      return true;
                  } else if (object instanceof Map) {
                      ((Map) object).values().removeIf(value -> dropEmptyFields(value));
                      return (((Map) object).size() == 0);
                  } else if (object instanceof List) {
                      ((List) object).removeIf(value -> dropEmptyFields(value));
                      return (((List) object).length == 0);
                  }
```
(same for the other data stream)

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,315 @@
...
+          target_field: neon_cyber.detections.detection_timestamp
+          tag: date_detection_timestamp
+          formats:
+              - ISO8601
+    - date:
+          field: json.detection_timestamp
+          target_field: '@timestamp'
+          tag: date_timestamp
+          formats:
+              - ISO8601
```
efd6, line 73: For these we might want to have `on_failure` handlers? How much can we tolerate the timestamps being absent?

packages/neon_cyber/manifest.yml
```diff
@@ -0,0 +1,84 @@
...
+                  show_user: true
+                  secret: true
+                - name: enable_request_tracer
+                  type: bool
+                  title: Enable request tracing
+                  default: false
+                  multi: false
+                  required: false
+                  show_user: false
+                  description: The request tracer logs requests and responses to the agent's local file-system for debugging configurations. Enabling this request tracing compromises security and should only be used for debugging. See [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-cel.html#_resource_tracer_filename) for details.
```
efd6: This should be at the data stream level.

packages/neon_cyber/data_stream/detections/manifest.yml
```diff
@@ -0,0 +1,46 @@
...
+      template_path: cel.yml.hbs
+      vars:
+          - name: resource_url
+            type: text
+            title: 'Detections Endpoint URL'
+            description: |-
+                i.e. scheme://host:port/path
+            required: true
+            show_user: true
+            default: https://api.neoncyber.io/v1/detections
```
efd6: For the sake of user ergonomics, is it the case that the detections endpoint is always at /v1/detections? (similarly for events at /v1/events). If it is, I would suggest hard coding the endpoint path into the CEL program and making the base URL a package level configuration next to the API token field. This would mean that that user only needs to configure it once (assuming that the base URL is almost constant for a user).

packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,82 @@
...
```
efd6: This is not necessary.

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
codypierce: Added on_failure just in case but these are guaranteed to be there

packages/neon_cyber/manifest.yml
codypierce: moved to DS

packages/neon_cyber/data_stream/detections/manifest.yml
codypierce: Moved URL to package and hardcoded and removed resource_url

packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs
codypierce: Any time I remove this "want_more" line "elastic_package test system" fails to get a hit on the mock server and I don't know why.

efd6: OK. I will take a look and see what is going on.

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
efd6: Sorry, I should be clearer (your interpretation is entirely reasonable, but not what I meant).

Let's say there is an invalid timestamp so the parse fails, is the document now useless (because it will not have the timestamp) or is it something that we can salvage some information from? If it's the former, we don't need to have an `on_failure` handler since we can just say definitively that the document had an error and we didn't try to do any further work, but if it's the latter, we can say, "there was a problem with this field, but I kept going and be aware of the error when you look at the document".

It's no doubt unlikely that there _would_ be a corrupted timestamp, but we live in the real world, so it is always possible.

codypierce: The latter would be correct. Especially for detections we wouldn't want to ignore any information. In that case is the on_failure I added correct, or is there another suggestion?

packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs
codypierce: Thank you for your help! I couldn't figure out how that works since on paper it seem extraneous but the system test definitely prevents it from completing if that single line is removed.

efd6: The issue is identified in the error message that unfortunately only gets shown in the logs
```
{
    "log.level": "error",
    "@timestamp": "2025-11-05T22:21:36.369Z",
    "log.origin": {
        "function": "github.com/elastic/elastic-agent/internal/pkg/agent/application/coordinator.logComponentStateChange",
        "file.name": "coordinator/coordinator.go",
        "file.line": 800
    },
    "message": "Unit state changed cel-default-cel-neon_cyber-50f5cedf-f9ea-481d-af64-5a5da5a63c74 (STARTING->FAILED): failed to check program: failed compilation: ERROR: <input>:12:54: found no matching overload for '_?_:_' applied to '(bool, map(string, list(map(string, string))), map(string, map(string, map(string, string))))'\n |     ).do_request().as(resp, (resp.StatusCode == 200) ?\n | .....................................................^ accessing config",
    "log": {
        "source": "elastic-agent"
    },
    "component": {
        "id": "cel-default",
        "state": "HEALTHY"
    },
    "unit": {
        "id": "cel-default-cel-neon_cyber-50f5cedf-f9ea-481d-af64-5a5da5a63c74",
        "type": "input",
        "state": "FAILED",
        "old_state": "STARTING"
    },
    "ecs.version": "1.6.0"
}
```
extracting the relevant part
```
found no matching overload for '_?_:_' applied to '(bool, map(string, list(map(string, string))), map(string, map(string, map(string, string))))'
 |     ).do_request().as(resp, (resp.StatusCode == 200) ?
 | .....................................................^ accessing config
```
What this is saying is that the ternary has branches that are not the same type, `map(string, list(map(string, string)))` and `map(string, map(string, map(string, string)))`. They need to agree, and we can achieve this by making them both `dyn` types so that the compile-time types are the same even though the runtime types are not identical.

This is done by (with some other cosmetic changes):
```diff
diff --git a/packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs b/packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs
index ea6c2524d4..efa226430e 100644
--- a/packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs
+++ b/packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs
@@ -34,31 +34,34 @@ program: |-
       ).do_request().as(resp, (resp.StatusCode == 200) ?
         resp.Body.decode_json().as(body,
           {
-            "events": (has(body.data) && body.data.size() > 0) ?
-              body.data.map(e,
-                {
-                  "message": e.encode_json(),
-                }
-              )
-            :
-              []
+            "events": dyn(
+              (has(body.data) && body.data.size() > 0) ?
+                body.data.map(e,
+                  {
+                    "message": e.encode_json(),
+                  }
+                )
+              :
+                []
+            )
           }
         )
       :
         {
-          "events": {
-            "error": {
-              "code": string(resp.StatusCode),
-              "id": string(resp.Status),
-              "message": "GET " + state.url + (
-                (size(resp.Body) != 0) ?
-                  string(resp.Body)
-                :
-                  string(resp.Status) + " (" + string(resp.StatusCode) + ")"
-              ),
-            },
-          },
-          "want_more": false,
+          "events": dyn(
+            {
+              "error": {
+                "code": string(resp.StatusCode),
+                "id": string(resp.Status),
+                "message": "GET " + state.url.trim_right("/") + "/v1/detections: " + (
+                  (size(resp.Body) != 0) ?
+                    string(resp.Body)
+                  :
+                    string(resp.Status) + " (" + string(resp.StatusCode) + ")"
+                ),
+              },
+            }
+          )
         }
       )
     )
diff --git a/packages/neon_cyber/data_stream/events/agent/stream/cel.yml.hbs b/packages/neon_cyber/data_stream/events/agent/stream/cel.yml.hbs
index 40341dcf6b..54cb70b0ac 100644
--- a/packages/neon_cyber/data_stream/events/agent/stream/cel.yml.hbs
+++ b/packages/neon_cyber/data_stream/events/agent/stream/cel.yml.hbs
@@ -34,31 +34,34 @@ program: |-
       ).do_request().as(resp, (resp.StatusCode == 200) ?
         resp.Body.decode_json().as(body,
           {
-            "events": (has(body.data) && body.data.size() > 0) ?
-              body.data.map(e,
-                {
-                  "message": e.encode_json(),
-                }
-              )
-            :
-              []
+            "events": dyn(
+              (has(body.data) && body.data.size() > 0) ?
+                body.data.map(e,
+                  {
+                    "message": e.encode_json(),
+                  }
+                )
+              :
+                []
+            )
           }
         )
       :
         {
-          "events": {
-            "error": {
-              "code": string(resp.StatusCode),
-              "id": string(resp.Status),
-              "message": "GET " + state.url + (
-                (size(resp.Body) != 0) ?
-                  string(resp.Body)
-                :
-                  string(resp.Status) + " (" + string(resp.StatusCode) + ")"
-              ),
-            },
-          },
-          "want_more": false,
+          "events": dyn(
+            {
+              "error": {
+                "code": string(resp.StatusCode),
+                "id": string(resp.Status),
+                "message": "GET " + state.url.trim_right("/") + "/v1/events: " + (
+                  (size(resp.Body) != 0) ?
+                    string(resp.Body)
+                  :
+                    string(resp.Status) + " (" + string(resp.StatusCode) + ")"
+                ),
+              },
+            }
+          )
         }
       )
     )
```

packages/neon_cyber/data_stream/detections/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,11 @@
+input: cel
+service: neon_cyber
+vars:
+    api_token: xxxx
+    enable_request_tracer: true
```
efd6: This needs to be moved to match the new manifest layout. Same in the other data stream.

packages/neon_cyber/data_stream/detections/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,11 @@
...
+data_stream:
+    vars:
+        resource_url: http://{{Hostname}}:{{Port}}/v1/detections
```
efd6: This will need to be moved to the package level and changed to `url` (at the moment the test is reaching out to api.neoncyber.io and failing (correctly) with a 401).
```
{
    "log.level": "debug",
    "@timestamp": "2025-11-05T22:40:13.138Z",
    "message": "HTTP request",
    "transaction.id": "S6QK57BN7LQHG-1",
    "url.original": "https://api.neoncyber.io/v1/detections",
    "url.scheme": "https",
    "url.path": "/v1/detections",
    "url.domain": "api.neoncyber.io",
    "url.port": "",
    "url.query": "",
    "http.request.method": "GET",
    "http.request.header": {
        "Authorization": [
            "xxxx"
        ],
        "User-Agent": [
            "Elastic-Filebeat/8.19.3 (linux; amd64; 7a508104ea6b6ddae7456dff8b61423e59a73962; 2025-08-26 01:12:39 +0000 UTC)"
        ]
    },
    "user_agent.original": "Elastic-Filebeat/8.19.3 (linux; amd64; 7a508104ea6b6ddae7456dff8b61423e59a73962; 2025-08-26 01:12:39 +0000 UTC)",
    "http.request.body.content": "",
    "http.request.body.truncated": false,
    "http.request.body.bytes": 0,
    "http.request.mime_type": "",
    "ecs.version": "1.6.0"
}
{
    "log.level": "debug",
    "@timestamp": "2025-11-05T22:40:14.281Z",
    "message": "HTTP response",
    "transaction.id": "S6QK57BN7LQHG-1",
    "http.response.status_code": 401,
    "http.response.body.content": "{\"error\":{\"code\":401,\"message\":\"Unauthorized\"},\"data\":null}",
    "http.response.body.truncated": false,
    "http.response.body.bytes": 59,
    "http.response.mime_type": "application/json; charset=utf-8",
    "http.response.header": {
        "Apigw-Requestid": [
            "Tl3MPgXKIAMEJbw="
        ],
        "Content-Length": [
            "59"
        ],
        "Content-Type": [
            "application/json; charset=utf-8"
        ],
        "Date": [
            "Wed, 05 Nov 2025 22:40:14 GMT"
        ],
        "Etag": [
            "W/\"3b-B6EKjFN6TdKYf7gDSWml69abDvE\""
        ],
        "Server": [
            "envoy"
        ],
        "X-Envoy-Upstream-Service-Time": [
            "159"
        ]
    },
    "ecs.version": "1.6.0"
}
```

efd6 (commented):
Minor changes only, otherwise looking good.

packages/neon_cyber/data_stream/detections/agent/stream/cel.yml.hbs
codypierce: Thanks! Updated to your suggestion and system tests are passing now.

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
efd6: No, if that's the semantics, what you've done is right.

I would say though that the `json` processor failure would be unrecoverable; if that fails there's no point in doing any further work since we can guarantee that the fields will be absent.

packages/neon_cyber/data_stream/detections/_dev/test/system/test-default-config.yml
codypierce: Good catch. My bad.

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
codypierce: Got it, committed a change to remove json on_failure

packages/neon_cyber/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,406 @@
...
+    - rename:
+          field: json.ua
+          target_field: user_agent.original
+          ignore_missing: true
+          tag: rename_user_agent_original
+    - rename:
+          field: json.url
+          target_field: neon_cyber.events.url
+          ignore_missing: true
+          tag: rename_events_url
```
efd6, line 187:
```
    - uri_parts:
          field: neon_cyber.events.url
          ignore_missing: true
          tag: uri_parts_events_url
```

packages/neon_cyber/data_stream/detections/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,323 @@
...
+    - rename:
+          field: json.description
+          target_field: neon_cyber.detections.description
+          ignore_missing: true
+          tag: rename_detection_description
+    - rename:
+          field: json.url
+          target_field: neon_cyber.detections.url
+          ignore_missing: true
+          tag: rename_detection_url
```
efd6, line 122:
```
    - uri_parts:
          field: neon_cyber.detections.url
          ignore_missing: true
          tag: uri_parts_detections_url
```

efd6 (commented):
You will need to run
```
elastic-package test system -g
elastic-package test policy -g
elastic-package build
```
using a v8.18.0 stack (the version that is the minimum specified in the manifest)

codypierce:
Switched manifest version to match testing config version 8.17.0. Regenerated expected and samples

efd6:
Quick query: where were the test sample obtained from? Live API with modification? Constructed from documentation?

codypierce:
> Quick query: where were the test sample obtained from? Live API with modification? Constructed from documentation?

Live API for a test account and modifications

efd6 (approved):
Thanks

PR #15726, opened 2025-10-22
aws_bedrock: tolerate non-object elements in invocation output.outputBodyJson lists

efd6:
## Proposed commit message
```
aws_bedrock: tolerate non-object elements in invocation output.outputBodyJson lists

Some models will include string elements in this list in chat streaming
responses, so promote non-object elements to an object with a single
"value" field holding the original element value. Without this change
those documents are otherwise rejected during ingest since flattened
fields may not contain non-object elements.

The test case is a copy of an existing test case, with the final element
in the list replaced with the string "[DONE]" to reflect the original
report.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


agithomas (approved):
LGTM!

ShourieG (approved):
LGTM

PR #15728, opened 2025-10-22
m365_defender: fix construction of file.path fields in the incident data stream

efd6:
## Proposed commit message
```
m365_defender: fix construction of file.path fields in the incident data stream

The ECS definition states that the file.path field should include the
file's basename as well as the directory.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/m365_defender/data_stream/incident/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/m365_defender/data_stream/incident/elasticsearch/ingest_pipeline/default.yml
@@ -1412,9 +1412,14 @@ processors:
       processor:
         append:
           field: file.path
-          value: '{{{_ingest._value.file_details.path}}}'
+          value: '{{{_ingest._value.file_details.path}}}\{{{_ingest._value.file_details.name}}}'
           allow_duplicates: false
           ignore_failure: true
+  - script:
+      tag: script_remove_backslash
+      if: ctx.file?.path instanceof List
+      source: |-
+        ctx.file.path.removeIf(v -> v == '\\');
   - foreach:
       field: json.alerts.evidence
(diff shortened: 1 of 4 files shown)


packages/m365_defender/data_stream/incident/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1412,9 +1412,14 @@ processors:
       processor:
         append:
           field: file.path
-          value: '{{{_ingest._value.file_details.path}}}'
+          value: '{{{_ingest._value.file_details.path}}}\{{{_ingest._value.file_details.name}}}'
```
andrewkroh, line 1415: Do we ever get events with fileDetails evidence from Microsoft Defender for Endpoint on {Linux,macOS,android,ios}? If so this might need a different path separator.

efd6: Yeah, I was wondering about that. I'm not sure how we can know this, or deal with it if we do.

ShourieG: if this issue is resolved, I can approve @efd6

efd6: I'll wait for @andrewkroh.

andrewkroh: Let's proceed with this for now. If/when we gather proof that it isn't working properly for some users, we can use additional logs from other Defender sources to develop a solution.

PR #15729, opened 2025-10-22
feat: add Agentless Hello World integration

nkvoll:
## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/agentless_hello_world/data_stream/generic/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/agentless_hello_world/data_stream/generic/elasticsearch/ingest_pipeline/default.yml
@@ -17,27 +17,6 @@ processors:
   - terminate:
       tag: data_collection_error
       if: ctx.error?.message != null && ctx.message == null && ctx.event?.original == null
-  - rename:
-      field: message
-      target_field: event.original
-      ignore_missing: true
-      if: ctx.event?.original == null
-  - remove:
-      field: message
-      tag: remove_message
-      ignore_missing: true
-      description: The `message` field is no longer required if the document has an `event.original` field.
-      if: ctx.event?.original != null
-  - json:
-      field: event.original
-      target_field: agentless_hello_world.generic
-      on_failure:
-        - set:
-            field: error.type
-            value: "json_parse_error"
-        - set:
-            field: error.message
-            value: "{{{ _ingest.on_failure_message }}}"
   - set:
       field: event.kind
       value: event
--- a/packages/agentless_hello_world/data_stream/generic/fields/fields.yml
+++ b/packages/agentless_hello_world/data_stream/generic/fields/fields.yml
@@ -1,6 +1,2 @@
-- name: agentless_hello_world.generic
-  type: group
-  fields:
-    - name: status_code
-      type: long
-      description: HTTP Status Code
+- name: http.response.status_code
+  external: ecs
--- a/packages/agentless_hello_world/data_stream/generic/_dev/test/pipeline/test-hello-world.json
+++ b/packages/agentless_hello_world/data_stream/generic/_dev/test/pipeline/test-hello-world.json
@@ -1,7 +1,11 @@
 {
   "events": [
     {
-      "message": "{\"status_code\":200}"
+      "http": {
+        "response": {
+          "status_code": 200
+        }
+      }
     }
   ]
 }
--- a/packages/agentless_hello_world/data_stream/generic/_dev/test/pipeline/test-hello-world.json-expected.json
+++ b/packages/agentless_hello_world/data_stream/generic/_dev/test/pipeline/test-hello-world.json-expected.json
@@ -1,8 +1,8 @@
 {
     "expected": [
         {
-            "agentless_hello_world": {
-                "generic": {
+            "http": {
+                "response": {
                     "status_code": 200
                 }
             },
--- a/packages/agentless_hello_world/data_stream/generic/agent/stream/cel.yml.hbs
+++ b/packages/agentless_hello_world/data_stream/generic/agent/stream/cel.yml.hbs
@@ -9,9 +9,11 @@ program: |
     .do_request()
     .as(resp, {
         "events": [{
-            "message": {
+            "http": {
+              "response": {
                 "status_code": resp.StatusCode
-            }.encode_json()
+            }
+          }
         }]
     })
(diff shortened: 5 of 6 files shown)


packages/agentless_hello_world/_dev/build/build.yml
```diff
@@ -0,0 +1,3 @@
+dependencies:
+  ecs:
+    reference: git@v8.11.0
```
andrewkroh:
```suggestion
    reference: git@v9.1.0
```

packages/agentless_hello_world/data_stream/generic/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,72 @@
+---
+description: Pipeline for processing Agentless Hello World generic logs.
+processors:
+  - set:
+      field: ecs.version
+      value: '8.11.0'
```
andrewkroh:
```suggestion
      value: '9.1.0'
```

packages/agentless_hello_world/data_stream/generic/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,72 @@
...
+  - rename:
```
andrewkroh: You'll want a `terminate` processor early in the pipeline (after the `ecs.version` processor) for when CEL produces an error.

https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#use-a-terminate-processor-with-inputs-that-produce-failure-events

packages/agentless_hello_world/data_stream/generic/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,72 @@
...
+              } else if (x instanceof List) {
+                  handleList(x);
+              }
+          }
+        }
+        handleMap(ctx);
+on_failure:
+  - set:
+      field: error.message
+      value: '{{ _ingest.on_failure_message }}'
```
andrewkroh:
```suggestion
on_failure:
  - set:
      field: event.kind
      value: pipeline_error
  - append:
      field: error.message
      value: >-
        Processor '{{{ _ingest.on_failure_processor_type }}}'
        {{#_ingest.on_failure_processor_tag}}with tag '{{{ _ingest.on_failure_processor_tag }}}'
        {{/_ingest.on_failure_processor_tag}}failed with message '{{{ _ingest.on_failure_message }}}'
```

https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#pipeline-on_failure-handler-must-set-errormessage

kpollich:
@jsoriano @mrodm Do you think there's a reasonable way for us to hide this integration in Kibana or will that require a code change that excludes this integration from all deployment types? Not sure if we have a flag for internal integrations like this already.

jsoriano:
> @jsoriano @mrodm Do you think there's a reasonable way for us to hide this integration in Kibana or will that require a code change that excludes this integration from all deployment types? Not sure if we have a flag for internal integrations like this already.

We don't have any way to hide integrations. While this is kept with a prerelease version at least it won't appear by default.

In any case, maybe it is not so bad to show this integration? There is an official hello-world Docker image :slightly_smiling_face: https://hub.docker.com/_/hello-world/

Maybe we could add a new "Sample" category, that Fleet could use to show this kind of integrations in a different way.

jsoriano:
> Maybe we could add a new "Sample" category, that Fleet could use to show this kind of integrations in a different way.

Thinking more about this I think this category can be actually useful, issue open to follow up on this https://github.com/elastic/package-spec/issues/1003.

kpollich:
> Maybe we could add a new "Sample" category, that Fleet could use to show this kind of integrations in a different way.

Sure having a dedicated category for this type of package is fine with me. Perhaps the risk of having this in the UI is not so high and it'd be easy enough to just include it. If someone wants to deploy an integration that pings https://epr.elastic.co/search and logs the status code of the response then I suppose we could just let them 😁 

There are lists in Kibana that allow us to exclude packages from certain deployment types, but I think these result in the packages being excluded from the Kibana APIs as well as the UIs entirely, meaning it will be slightly  more difficult to install them. If the package is just available as-is in the the UI and API as "Agentless hello world" under a "Sample" category with a clear description of its intended usage then we're probably fine to just ship it.

nkvoll:
@andrewkroh I re-requested a review from you before merging in case you have any further input or adjustments we should do before merging or if looks good from your pov now :)

packages/agentless_hello_world/data_stream/generic/fields/base-fields.yml
```diff
@@ -0,0 +1,23 @@
...
+  type: constant_keyword
+  description: Event module
+  value: agentless_hello_world
+- name: event.dataset
+  type: constant_keyword
+  description: Event dataset
+  value: agentless_hello_world.generic
+- name: '@timestamp'
+  type: date
+  description: Event timestamp.
```
andrewkroh: Since these fields exist in ECS, I recommend to depend the ECS definition for consistency. (More background in https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#defining-an-ecs-field-without-using-an-external-definition)

```suggestion
- name: data_stream.type
  external: ecs
- name: data_stream.dataset
  external: ecs
- name: data_stream.namespace
  external: ecs
- name: event.module
  type: constant_keyword
  external: ecs
  value: agentless_hello_world
- name: event.dataset
  type: constant_keyword
  external: ecs
  value: agentless_hello_world.generic
- name: "@timestamp"
  external: ecs
```

andrewkroh (approved):
LGTM. Added a minor suggestion to use `external: ecs` where possible.

packages/agentless_hello_world/data_stream/generic/fields/fields.yml
```diff
@@ -0,0 +1,6 @@
+- name: agentless_hello_world.generic
+  type: group
+  fields:
+    - name: status_code
```
andrewkroh: You could use ECS http.response.status_code (https://www.elastic.co/docs/reference/ecs/ecs-http#field-http-response-status-code) for this field instead.

PR #15664, opened 2025-10-16
add discovery.datasets to _otel packages owned by infraobs

mykola-elastic:
Adding `discovery.datasets` field (to support auto-install) for:
- apache_otel
- iis_otel
- mysql_otel
- nginx_otel
- nginx_ingress_controller_otel
- postgresql_otel

Increased `kibana.version` to `^9.2.0` as all those packages use `format_version: 3.5.0` which is only supported beginning Kibana 9.2.0.

## Proposed commit message

See title.

## Checklist

- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [x] add `discovery.datasets` to packages
- [x] test if the packages get auto-installed on the data ingestion

## How to test this PR locally

- build any of the packages listed (for example, `mysql_otel`)
- spin up local elastic stack 9.2
- set up OTel receiver + exporter (for example, `mysqlreceiver`), and the corresponding service (I spinned up `mysql` in docker)
- check that data is ingested (for mysql - `data_stream.dataset: mysqlreceiver.otel`)
- wait 5 mins
- check that the corresponding package was installed (dashboard added)

PR #14770, opened 2025-08-01
Add discovery datasets field to system_otel

juliaElastic:
## Proposed commit message

Added `discovery.datasets` field to enable auto-install of content package when data is ingested with dataset `hostmetricsreceiver`.

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## Related issues

- Relates https://github.com/elastic/package-spec/issues/919

PR #15733, opened 2025-10-23
add discovery.datasets to docker_otel and kubernetes_otel

mykola-elastic:
Follow-up PR for https://github.com/elastic/integrations/pull/15664, https://github.com/elastic/integrations/pull/14770

Adding `discovery.datasets` field (to support auto-install) for:
- docker_otel
- kubernetes_otel

Increased `kibana.version` to `^9.2.0` as all those packages use `format_version: 3.5.0` which is only supported beginning Kibana 9.2.0.

## Proposed commit message

See title.

### DIFF
--- a/packages/docker_otel/changelog.yml
+++ b/packages/docker_otel/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Add `discovery` field to support auto-install
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/1 # TODO
+      link: https://github.com/elastic/integrations/pull/15733
 - version: "0.1.2"
   changes:
     - description: Add `opentelemetry` category
--- a/packages/kubernetes_otel/changelog.yml
+++ b/packages/kubernetes_otel/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Add `discovery` field to support auto-install
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/1 # TODO
+      link: https://github.com/elastic/integrations/pull/15733
 - version: 1.2.2
   changes:
     - description: Update EDOT quick start url


mykola-elastic:
The dashboard filter for `kubernetes_otel` package is `data_stream.dataset: *.otel`. @elastic/obs-ds-hosted-services Please confirm that I picked the right datasets to detect for auto-install:

```
  datasets:
    - name: kubeletstatsreceiver.otel
    - name: k8sobjectsreceiver.otel
    - name: k8seventsreceiver.otel
    - name: k8sclusterreceiver.otel
```

packages/kubernetes_otel/manifest.yml
```diff
@@ -11,11 +11,16 @@ categories:
...
 conditions:
   kibana:
-    version: "^8.18.0 || ^9.2.0"
+    version: "^9.2.0"
   elastic:
     subscription: "basic"
 discovery:
   fields: []
+  datasets:
+    - name: kubeletstatsreceiver.otel
```
ishleenk17, line 20: These you figured out from the dashboard ? @mykola-elastic

mykola-elastic: @ishleenk17 no, in the dashboard there is only `*.otel` filter, none of them are specified there. I picked the datasets names from the [receiver list in the elastic-agent (EDOT)](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/otel/README.md#receivers). I also, asked the @elastic/obs-ds-hosted-services team for clarification/checking https://github.com/elastic/integrations/pull/15733#issuecomment-3435373064

gizas: Mykola you are correct. You can find in the list of receivers also [here](https://github.com/elastic/elastic-agent/blob/main/deploy/helm/edot-collector/kube-stack/values.yaml)

Also a comment about logs for kubernetes that go to logs-generic.otel . Do you want also to add this?

mykola-elastic: Thanks for confirming!
About generic.otel - I would rather keep only "specialized" names. I guess it is possible to some other data to get that "generic.otel" dataset name which would lead to auto-installation of kubernetes dashboard where there is no kubernetes data
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/elasticsearchexporter#elasticsearch-document-routing

PR #15739, opened 2025-10-23
[Draft] Add docker_otel_input package

strawgate:
This package implements a Docker Stats input using the OpenTelemetry
Collector's dockerstats receiver, following the pattern established
in PR #14315.

Key features:
- Type: integration with otelcol input (not content package)
- Configurable collection interval, endpoint, and filtering
- Comprehensive field definitions for container metrics
- Full documentation and test policy

Resolves: elastic/integrations#15731

### DIFF
--- a/packages/docker_input_otel/agent/input/input.yml.hbs
+++ b/packages/docker_input_otel/agent/input/input.yml.hbs
@@ -1,6 +1,9 @@
 receivers:
   docker_stats:
     collection_interval: {{period}}
+    metrics:
+      container.cpu.usage.percpu:
+        enabled: {{provide_per_core_cpu_metrics}}
 {{#if endpoint}}
     endpoint: {{endpoint}}
 {{/if}}
@@ -15,9 +18,6 @@ receivers:
 {{/if}}
 {{#if api_version}}
     api_version: "{{api_version}}"
-{{/if}}
-{{#if provide_per_core_cpu_metrics}}
-    provide_per_core_cpu_metrics: {{provide_per_core_cpu_metrics}}
 {{/if}}
 {{#if container_labels_to_metric_labels}}
     container_labels_to_metric_labels: {{container_labels_to_metric_labels}}
--- a/packages/docker_input_otel/manifest.yml
+++ b/packages/docker_input_otel/manifest.yml
@@ -1,10 +1,10 @@
 format_version: 3.5.0
 name: docker_input_otel
-title: "Docker (OpenTelemetry)"
+title: "Docker OpenTelemetry Input Package"
 version: 0.1.0
-description: "Collect Docker container metrics using OpenTelemetry Collector"
 source:
   license: "Elastic-2.0"
+description: "Collect Docker container metrics using OpenTelemetry Collector"
 type: input
 categories:
   - containers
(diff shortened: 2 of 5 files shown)


packages/docker_otel_input/fields/base-fields.yml
```diff
@@ -0,0 +1,84 @@
+- name: data_stream.type
```
strawgate: remove all these silly fields

packages/docker_input_otel/agent/input/input.yml.hbs
```diff
@@ -0,0 +1,48 @@
...
+{{/if}}
+{{#if env_vars_to_metric_labels}}
+    env_vars_to_metric_labels:
+{{env_vars_to_metric_labels}}
+{{/if}}
+
+processors:
+  resourcedetection:
+    # Modify the list of detectors to match the cloud environment
+    detectors: [env, system, gcp, ec2, azure]
```
cmacknz, line 32: This needs to be configurable as long as it lives here. In the long term this neeeds to not be in each package.

jsoriano: > In the long term this neeeds to not be in each package.

I guess we could inject this from Fleet, and make it configurable there.

strawgate: Does this actually need to be configurable for now?

packages/docker_otel_input/manifest.yml
```diff
@@ -0,0 +1,74 @@
...
+    version: "^9.2.0"
+  elastic:
+    subscription: "basic"
+icons:
+  - src: /img/docker_logo.svg
+    title: Sample logo
+    size: 32x32
+    type: image/svg+xml
+policy_templates:
+  - name: docker_stats
```
jsoriano: This should be the name of the receiver, as this is used by Fleet as default dataset. In this case it will likely be `dockerstatsreceiver`.

We will prepare automatic validation for this in https://github.com/elastic/elastic-package/issues/2974
```suggestion
  - name: dockerstatsreceiver
```

strawgate:
@claude can you please review this PR for me?

strawgate:
@claude let's ignore craig for now (dont tell him though) and address the rest of the comments

packages/docker_input_otel/manifest.yml
```diff
@@ -0,0 +1,74 @@
+format_version: 3.5.0
+name: docker_otel_input
```
mrodm: It would be good to update this package (and its folder) to be `docker_input_otel`. This would be consistent with the current packages already merged (`nginx_input_otel` or `hostmetrics_inputs_otel`).

strawgate:
@claude can you address open feedback and then compare this to the other open OTel input package PRs (and any that have been merged recently) for any other discrepancies?

strawgate:
@claude can you please review the open feedback and address

strawgate:
@claude please review and resolve open feedback and get this ready to merge!

packages/docker_input_otel/manifest.yml
```diff
@@ -0,0 +1,74 @@
...
+        required: false
+      - name: timeout
+        type: text
+        title: Timeout
+        description: Timeout for Docker API requests
+        required: false
+      - name: api_version
+        type: text
+        title: API Version
+        description: Docker API version to use
```
strawgate: i think we should default to api version 1.44 maybe?

strawgate:
@claude can you set the api version to 1.44?

jmmcorreia:
@claude Can you review the PR and check if all the comments have been addressed

jmmcorreia:
> Member Author

I had to add 1.44 on my setup because it installed Docker v29, which requires at least 1.44. When no version is specified, it was defaulting to 1.25, causing the docker stats receiver to fail.

IMO, I guess we could make 1.44 default since it has been present in Docker Engine since v25, which has been released in January 2024. ref: https://docs.docker.com/engine/release-notes/25.0/#2500

However, introducing this version will have drawbacks in the sense we will have to come back and update the repo once more at a later point either to remove the default and bump the version.

Since this will be a preview version, I guess we could add the default to align with what docker is mandating, but what are your thoughts?

packages/docker_input_otel/docs/README.md
```diff
@@ -0,0 +1,109 @@
...
+Docker Container Stats are collected via the `docker_stats` receiver. It collects container metrics from the Docker daemon using the Docker Stats API. It provides comprehensive metrics about container resource usage including CPU, memory, network, and block I/O statistics.
+
+## Requirements
+
+- Docker API version 1.25 or higher
+- Access to the Docker daemon socket (default: `unix:///var/run/docker.sock` on linux and `npipe:////./pipe/docker_engine` on Windows)
+
+## Compatibility
+
+This integration uses the OpenTelemetry dockerstats receiver which is currently in **alpha** stability. The configuration and metrics may change between versions.
```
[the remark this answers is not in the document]
jmmcorreia: @claude that link just points to the collector repo, not related to stability

packages/docker_input_otel/agent/input/input.yml.hbs
jmmcorreia: Should this just be a variable that can be configured by the user when enabling the integration?

Other integrations so far only seem to use system, so we could use system as a default, but allow it to be modified when enabling the integration. Would that be acceptable?

jmmcorreia: Screenshot 2026-02-10 at 16 51 24

@strawgate, @cmacknz , @jsoriano 

I have a change ready locally that allows the detectors to be configurable. (See attached image, but ignore the comma separated description, that will be fixed to reflect the current UX)


Is this what is expected here and am I ok to push it? Or is the expectation that this should be done in a different way ?

strawgate:
@claude how does this compare to the other feedback given to other input package PRs? as well as what has merged into main? any important discrepancies?

jmmcorreia:
@claude Can you review again the PR, compare against other existing packages and compare against all the feedback in this thread ?

strawgate:
@claude Can you review again the PR, compare against other existing packages and compare against all the feedback in this thread ?

strawgate:
moved to https://github.com/elastic/integrations/pull/17372

PR #15740, opened 2025-10-23
openai: Update dashboard to remove internal regex filters

shmsr:
## Proposed commit message

No visual changes. So internally, OpenAI changed the name of their image model to `dall-e` to `unclip-029-dapi` and it is reflecting in the older data as well. So specifically in image and audio charts we were using regex filters like `dall-e.*`, `whisper.*`, etc. and hence this PR removes those filters so that because of any change in model name or new models, charts render as expected.

This issue was noticed when one of the charts that depended on it did not render although data was there because it was using `dall-e.*` regex and OpenAI changed the name and that model name isn't even there.

## Checklist

### DIFF
--- a/packages/openai/kibana/dashboard/openai-651bb059-f606-44fc-b704-2078d0af26da.json
+++ b/packages/openai/kibana/dashboard/openai-651bb059-f606-44fc-b704-2078d0af26da.json
@@ -2573,17 +2573,13 @@
                                 "id": "logs-*",
                                 "name": "indexpattern-datasource-layer-0d76c878-9134-41d9-a09e-d2dab922806f",
                                 "type": "index-pattern"
-                            },
-                            {
-                                "id": "logs-*",
-                                "name": "27dea502-7bcb-466b-8d50-4dea78307b9b",
-                                "type": "index-pattern"
                             }
                         ],
                         "state": {
                             "adHocDataViews": {},
                             "datasourceStates": {
                                 "formBased": {
+                                    "currentIndexPatternId": "logs-*",
                                     "layers": {
                                         "0d76c878-9134-41d9-a09e-d2dab922806f": {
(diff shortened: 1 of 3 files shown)


packages/openai/kibana/dashboard/openai-651bb059-f606-44fc-b704-2078d0af26da.json
```diff
@@ -2600,10 +2596,8 @@
                                                     "params": {
                                                         "exclude": [],
                                                         "excludeIsRegex": false,
-                                                        "include": [
-                                                            "dall-e.*"
```
muthu-mps, line 2604: Once after removing these filter, Is this panel still rendering only the image model metrics?

ishleenk17:
> hence this PR removes those filters so that because of any change in model name or new models, charts render as expected.

Do we expect it to render other models too incorrectly , since there was a need to add regex initially ?

packages/openai/kibana/dashboard/openai-651bb059-f606-44fc-b704-2078d0af26da.json
shmsr: Yes!

muthu-mps: I see, this is taken care by dataset filter.

muthu-mps (approved):
LGTM!

shmsr:
> > hence this PR removes those filters so that because of any change in model name or new models, charts render as expected.
> 
> Do we expect it to render other models too incorrectly , since there was a need to add regex initially ?

> Do we expect it to render other models too incorrectly

I checked, it does not. I might have put them as that time there were only one model per image/audio models.

PR #15745, opened 2025-10-24
[checkpoint] Update kv split for checkpoint

ash-darin:
Enhancement

## Proposed commit message

This uses a more complex Regex to split the KV pairs in the Checkpoint message.

This is necessary if you use it to process email, as the headers might contain strings of the form:

`X-IronPort-AV: E=Sophos;i=\\\"8.18,219,1756234400\\\";   d=\\\"png`

`";       Sophos
`

The original regex would regeard both examples as a split, when in reality they are not. Use (negative) lookbehind to box the kv split parameter in.

`(?<!\\")(?<="); (?=\w)`

lookbehind: `(?<=")` react to `"; `
but
negative lookbehind: `(?<!\\")` do not react to `\";` 

lookahead: `(?=\w)` expect a character after 1 (one) space: `; abcd`

## Checklist

Check your logs/indices if any messages are rejected. Apply fix and check whether numbers go down..

### DIFF
--- a/packages/checkpoint/docs/README.md
+++ b/packages/checkpoint/docs/README.md
@@ -534,12 +534,14 @@ The `firewall` data stream provides events from Check Point devices, including f
 | checkpoint.rule_action | Action of the matched rule in the access policy. | keyword |
 | checkpoint.rulebase_id | Layer number. | integer |
 | checkpoint.scan_direction | Scan direction. | keyword |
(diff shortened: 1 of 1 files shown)


ash-darin:
@taylor-swanson I updated the changelog and added a test case in a hopefully sufficient way, as I shortened it considerably. The original message with all the headers would have been 24kb.

PR #15766, opened 2025-10-27
[Fortinet Fortiproxy] handle malformed quotes

haetamoudi:
## Proposed commit message

add support for malformed quotes and log version field

Fix for  https://github.com/elastic/sdh-beats/issues/6526
- Add missing log version field `logver`
- Remove timestamp field as the event already as a `eventime` which seems more precise
- handle malformed quotes

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/fortinet_fortiproxy/data_stream/log/fields/fields.yml
```diff
@@ -1201,6 +1201,10 @@
       type: keyword
       description: >-
         SSH login Name
+    - name: logver
+      type: integer
```
taylor-swanson:
```suggestion
      type: long
```

Long would probably be a better choice, unless the docs specifically call out that it's a 32-bit integer.

haetamoudi: I tried but I got the following error `field "fortinet.proxy.logver"'s Go type, string, does not match the expected field type: long (field value: 704080649)` do I need to add a convert processor?

taylor-swanson: You would need one, yes. The k/v script that extract fields by default extracts values to strings, which is why you see that message. There might be some implicit conversion happening in the background on the ES side, but it would be better to have an explicit convert.

Issue #15155, opened 2025-09-04
[qualys_gav]: Incorrect parsing of ip_v4


clement-fouque:
### Integration Name

Qualys Global AssetView [qualys_gav]

### Dataset Name

qualys_gav.asset

### Integration Version

0.1.0

### Agent Version

9.1.0

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.1.0

### OS Version and Architecture

elastic-package

### Software/API Version

2.0

### Error Message

`Processor convert with tag convert_asset_network_interface_list_data_network_interface_address_ip_v4 in pipeline logs-qualys_gav.asset-0.1.0 failed with message: 'XXX.XXX.XXX.XXX, XXX.XXX.XXX.XXX, XXX.XXX.XXX.XXX, XXX.XXX.XXX.XXX, XXX.XXX.XXX.XXX, XXX.XXX.XXX.XXX' is not an IP string literal.`

### Event Original

```json
{
  "activity": { "lastScannedDate": 1736429592000, "source": "DNSNAME" },
  "address": "XXX.XXX.XXX.XXX",
  "agent": {
    "activations": null,
    "configurationProfile": null,
    "connectedFrom": null,
    "errorStatus": null,
    "lastActivity": 0,
    "lastCheckedIn": 0,
    "lastInventory": 0,
    "udcManifestAssigned": null,
    "version": null
  },
  "agentId": null,
  "asn": null,
  "assetId": 760617201,
  "assetName": "XXX",
  "assetType": null,
  "assetUUID": "XXX",
  "assignedLocation": null,
  "biosAssetTag": "",
  "biosDescription": null,
  "biosSerialNumber": "",
  "businessAppListData": null,
  "businessInformation": null,
  "cloudProvider": null,
  "container": {
    "hasSensor": null,
    "noOfContainers": 0,
    "noOfImages": 0,
    "product": null,
    "version": null
  },
  "cpuCount": 0,
  "createdDate": "2023-06-01T15:08:06.000Z",
  "criticality": {
    "isDefault": false,
    "lastUpdated": "2023-06-01T15:08:05.000Z",
    "score": 3
  },
  "customAttributes": null,
  "dnsName": "XXX",
  "domain": null,
  "domainRole": null,
  "easmTags": null,
  "hardware": {
    "category": "Computers / Unidentified",
    "category1": "Computers",
    "category2": "Unidentified",
    "fullName": null,
    "lifecycle": null,
    "manufacturer": "Unidentified",
[...]

PR #15782, opened 2025-10-28
[Qualys GAV] - Added support for comma separated list of IPv4 addresses in network_interface list data

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
Current implementation fails to process logs where the ipv4
address in network_interface list data is a comma separated string
literal. This PR addresses that parsing issue.
```

## NOTE: 
The newly added test uses a modified existing log to reproduce our use-case. No new logs or customer sensitive data has been added/used in this test.
 

## Checklist

## Author's Checklist

- [ ] 

## Related issues
- Closes https://github.com/elastic/integrations/issues/15155

### DIFF
--- a/packages/qualys_gav/changelog.yml
+++ b/packages/qualys_gav/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Added support for comma separated list of IPv4 addresses in network_interface list data.
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/1111
+      link: https://github.com/elastic/integrations/pull/15782
 - version: "0.3.0"
   changes:
     - description: Added support for configuring cel max_executions parameter.


packages/qualys_gav/data_stream/asset/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -728,31 +728,69 @@ processors:
...
-      tag: foreach_asset_network_interface_list_data_network_interface_to_convert_address_ip_v4
+      tag: foreach_network_interface_list_data_network_interface_to_split_address_ip_v4
       if: ctx.qualys_gav?.asset?.network_interface_list_data?.network_interface instanceof List
       processor:
-        convert:
+        split:
           field: _ingest._value.address_ip_v4
-          tag: convert_asset_network_interface_list_data_network_interface_address_ip_v4
-          type: ip
+          separator: ','
```
efd6, line 736:
```suggestion
          separator: '\s*,\s*'
```
defensively. The better alternative would be to use `trim` on all the results, but that would require an additional `foreach`.

ShourieG: @efd6, I'm already using a `trim` on all results below in an additional foreach

efd6: I missed that, LGTM  then.

PR #15785, opened 2025-10-28
New httpcheck OTel input package

jsoriano:
## Proposed commit message

New OTel input package for the HTTP Check Receiver.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Supersedes https://github.com/elastic/integrations/pull/14315.

### DIFF
--- a/packages/httpcheck_otel/_dev/test/policy/test-eprcheck.expected
+++ b/packages/httpcheck_otel/_dev/test/policy/test-eprcheck.expected
@@ -25,7 +25,7 @@ processors:
             - context: datapoint
               statements:
                 - set(attributes["data_stream.type"], "metrics")
-                - set(attributes["data_stream.dataset"], "httpcheck.httpcheckreceiver")
+                - set(attributes["data_stream.dataset"], "httpcheck_otel.httpcheckreceiver")
                 - set(attributes["data_stream.namespace"], "ep")
 receivers:
     httpcheck/componentid-0:


packages/httpcheck_otel/manifest.yml
```diff
@@ -0,0 +1,67 @@
...
+title: "Simple HTTP Check"
+version: 0.1.0
+source:
+  license: "Elastic-2.0"
+description: "Perform HTTP checks using the HTTP Check receiver of the OTel Collector."
+type: input
+categories:
+  - monitoring
+  - network
+  - web
```
ishleenk17, line 13: Add opentelemtry category

packages/httpcheck/sample_event.json
```diff
@@ -0,0 +1,28 @@
+{
+    "@timestamp": 1761653610755,
+    "_metric_names_hash": "dbc9255cf6baf03b",
+    "attributes": {
+        "http.method": "GET",
+        "http.status_class": "2xx",
+        "http.url": "http://svc-web:80"
+    },
+    "data_stream": {
+        "dataset": "httpcheck.httpcheckreceiver.otel",
```
ishleenk17: Shouldn't it be httpcheckreceiver.otel ?

packages/httpcheck_otel/_dev/test/policy/test-eprcheck.yml
```diff
@@ -0,0 +1,3 @@
+vars:
+  endpoints:
+    - https://epr.elastic.co
```
ishleenk17, line 3: The test for input package is going to be if the policy is correctly formed ?
Any data related check we plan to add ?

packages/httpcheck_otel/_dev/test/policy/test-eprcheck.expected
```diff
@@ -0,0 +1,51 @@
+connectors:
+    forward: {}
+exporters:
+    elasticsearch/componentid-0:
```
ishleenk17, line 4: What are these componentid-0 ? Fetched from epr ?

jsoriano: Fleet generates this configuration with randomized component ids. Here elastic-package is replacing the random part with predictable strings so they can be compared in tests.

packages/httpcheck_otel/_dev/test/policy/test-eprcheck.yml
jsoriano: > The test for input package is going to be if the policy is correctly formed ?

This kind of test is available also for integration packages, is not specific to input packages.

> Any data related check we plan to add ?

System tests are the closest thing to a data related check that we have at the moment.

jsoriano: Just to be clear, policy tests are not intended to collect any data, they only create a policy with some given variables and check that the generated agent policy is the expected one.

packages/httpcheck_otel/manifest.yml
jsoriano: Added in 209e7cffa1d.

packages/httpcheck/sample_event.json
jsoriano: In a real case it will indeed be `httpcheckreceiver.otel`. 

Here this document is obtained with `elastic-package test system`, what is configuring the dataset as `<package name>.<policy template name>` for legacy reasons.

I have created a follow up issue about this, I think we should fix this discrepancy at some point: https://github.com/elastic/elastic-package/issues/3036

packages/httpcheck_otel/manifest.yml
```diff
@@ -0,0 +1,68 @@
+format_version: 3.5.0
```
mrodm, line 1: As this package does not define `fields` folder, `format_version` should be updated to `3.5.1` when available.

```suggestion
format_version: 3.5.1
```

jsoriano: This change was introduced for versions over 3.5.0 https://github.com/elastic/package-spec/pull/994/files#diff-53df6a0943d7cccefa15cfe92140696438e2d6f9425a94f0ea7d9d294f600da3R78

I would prefer to keep this on 3.5.0.

mrodm: You're right, it is allowed to be used starting with 3.5.0.
So as you mention, it can be kept using the same format_version. No need to apply the change proposed above. :+1:

packages/httpcheck_otel/_dev/test/policy/test-eprcheck.yml
ishleenk17: Should we consider them as a good to have for input packages.
We usually don't have these in regular integration packages. @jsoriano

ishleenk17:
Also, shall we make the package name as httpcheck_otel ? For us to easily relate it to OTEL

packages/httpcheck_otel/_dev/test/policy/test-eprcheck.yml
jsoriano: Yeah, they could be used more broadly.

They aren't more used now probably because they were introduced not so long ago and we haven't promoted them a lot.

jsoriano:
> Also, shall we make the package name as httpcheck_otel ? For us to easily relate it to OTEL

Not sure, I don't think there is going to be other `httpcheck` package.

packages/httpcheck_otel/_dev/build/docs/README.md
```diff
@@ -0,0 +1,102 @@
...
+{{/*
+This template can be used as a starting point for writing documentation for your new integration. For each section, fill in the details
+described in the comments.
+
+Find more detailed documentation guidelines in https://www.elastic.co/docs/extend/integrations/documentation-guidelines
+*/}}
+# HTTP Check Integration for Elastic
+
+## Overview
+{{/* Complete this section with a short summary of what data this integration collects and what use cases it enables */}}
```
teresaromero, line 11: Should the text under `{{}}`. be removed? or does the automation removes it when rendering the final doc?

jsoriano: This is removed when the template is rendered, you can find the result in the README below: https://github.com/elastic/integrations/pull/15785/files#diff-fa6b5192175adde94f4f96ad566b7715502f10e3b6ab28e149c3f8a712da262bR7

jsoriano:
Updated after updaging package-spec and elastic-package..

jsoriano:
Hey, sorry, last minute change, I am finally adding the `_otel` suffix as suggested by Ishleen. Even if it is only for consistency with what is being done with other packages.

packages/system/data_stream/cpu/fields/agent.yml
mrodm, line 1: It looks like that these files from `system` package were not supposed to be pushed, is that right?

jsoriano: Too much multitasking :grimacing:, removed.

PR #15786, opened 2025-10-28
[citrix_adc] Handled failed status in sslvpn pipeline

mjwolf:
## Proposed commit message

Correct the sslvpn ingest pipeline to properly handle events where the event contains a failed status.
Previously, the pipeline was expecting a single word, which is used for "Passed" status, but failed status has multiple words, to explain the problem. The pipeline will now handle multi-word status.

## Checklist


packages/citrix_adc/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.17.5"
+  changes:
+    - description: Properly parse failed status conditions in sslvpn pipeline
+      type: enhancement
```
taylor-swanson:
```suggestion
      type: bugfix
```

packages/citrix_adc/data_stream/log/_dev/test/pipeline/test-citrix-waf-native-14-1.log-expected.json
```diff
@@ -463,6 +463,89 @@
...
+                "extended": {
+                    "message": "CaseID: 847e8 - Client IP 192.0.2.0 - Vserver 198.51.100.2:443 - Client_security_expression \"CLIENT.SYSTEM('DOMAIN_SUFFIX_anyof_example.com.jp[COMMENT: Domain check]') EXISTS && CLIENT.APPLICATION('ANTIVIR_1346_3064_RTP_==_TRUE[COMMENT: Cortex XDR]') EXISTS\" - Client_security_check \"Failed - User not allowed to login\""
+                },
+                "host": "PRODSYST001",
+                "name": "CLISEC_CHECK"
+            },
+            "citrix_adc": {
+                "log": {
+                    "client_ip": "192.0.2.0",
+                    "client_security_check_status": "\"Failed - User not allowed to login\"",
```
taylor-swanson: Quotes should be stripped for `client_security_check_status`. `Client_security_expression` already seems to do the right thing, so you could base off of that. You may need to tolerate cases with and without quotes as well (basically just making them optional, but that could potentially interfere with GREEDYDATA). Should be able to test that easily, though.

packages/citrix_adc/changelog.yml
mjwolf: I've changed this

packages/citrix_adc/data_stream/log/_dev/test/pipeline/test-citrix-waf-native-14-1.log-expected.json
mjwolf: I've updated the pattern to work with quotes and GREEDYDATA or no quotes and WORD, which should match the real data

PR #15799, opened 2025-10-29
[eset_protect] Fix parsing of parsing HEARTBEAT messages, Enhance error handling for agent failures

brijesh-elastic:
## Proposed commit message
```
eset_protect: Fix parsing of parsing HEARTBEAT messages, Enhance error handling for agent failures

ESET sends heartbeat messages every 1–3 minutes to keep the connection open. We collect those events and
drop them from the ingest pipeline.
Improve error handling and display for CEL code, and use the terminate processor to handle agent errors
in the ingest pipeline.

Update format_version to 3.3.2.
Bump the minimum stack version to ^8.16.0 || ^9.0.0.
```

## Checklist

- Clone integrations repo.

> elastic-package test -v


packages/eset_protect/data_stream/detection/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -19,9 +30,6 @@ processors:
         - append:
             field: error.message
             value: "Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}"
-  - drop:
-      if: ctx.json?.message == "retry"
-      tag: drop_retry_events
```
efd6, line 24: This needs a matching change to the pipeline test; the retry case needs to be removed since the pipeline is never expected to see that kind of document.

packages/eset_protect/changelog.yml
```diff
@@ -1,4 +1,15 @@
 # newer versions go on top
+- version: "1.10.0"
+  changes:
+    - description: "Enhance error handling for CEL failures."
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/15799
+    - description: "Use `terminate` processor to handle agent errors."
```
efd6: I don't think users should care about this. The interesting thing for them is that agent errors will be more meaningful since they won't have additional subsequent failures, so the first item in the list should be enough, though I would replace "CEL" with agent, since again the mechanism should not be something that they care about.

efd6 (approved):

PR #15801, opened 2025-10-29
Align categories 

trisch-me:
This PR aligns categories from data_streams in templates with global categories of the integration.

## Proposed commit message

Aligning categories of sub datastreams with category of integration itself.

Issues:
was discussed in https://github.com/elastic/protections-cloud/issues/15713

@jsoriano @JDKurma FYI


trisch-me:
@elastic/security-service-integrations @elastic/obs-ds-hosted-services @elastic/obs-infraobs-integrations could you please look into it?

andrewkroh:
I think proposed commit message could use a deeper explanation. And we should refer to "policy template" instead of "datastreams". Suggestion:

```
This commit addresses inconsistencies in package categorization by adding
missing categories to various integration packages.

Previously, policy templates within certain integrations used categories
(or their parent categories) that were not explicitly defined at the
integration level. This was identified by ensuring all parent categories
of policy template categories are a subset of integration-level
categories.

References

https://github.com/elastic/package-registry/blob/1bec8bf2e7f67f04f38a6310f5b70e56e3b22bfd/categories/categories.yml
```

That doesn't answer the _why_ behind this change so you should add that.

muthu-mps (approved):

trisch-me:
@uri-weisman could you please check this PR as a maintainer of @elastic/contextual-security

uri-weisman (approved):
LGTM

PR #15819, opened 2025-10-30
[Contextual Security][CNVM] change vulnerability.published_date field type from keyword to date

alexreal1314:
This is a BREAKING CHANGE that updates the field mapping that aligns cnvm with other integrations - tenable_io, m365 etc.

## Summary

This PR updates the vulnerability.published_date field mapping from keyword to date type in the Cloud Native Vulnerability Management (CNVM) integration. The change includes:

1. modifying the field definition in `fields/vulnerability.yml` to use the date type
2. adding a date processor to the ingest pipeline (default.yml) that parses the published_date field using ISO8601 format with proper error handling.
3. bumping the package version from 3.1.1 to 4.0.0 following semantic versioning for breaking changes
4. documenting the breaking change in changelog.yml. The ingest pipeline processor includes conditional logic to only parse the field when it exists and is not empty, with on_failure handling to append error messages without breaking the entire pipeline. 

change was made to address this [issue](https://github.com/elastic/security-team/issues/12860).

## Checklist

### Breaking Change Impact
⚠️ Important: This is a breaking change that will result in mapping conflicts within the data stream during the upgrade process:
- Existing indices will retain the old keyword mapping for vulnerability.published_date
- New indices (created after rollover) will use the new date mapping
- Multiple backing indices with different mappings will coexist under the same logs-cloud_security_posture.vulnerabilities-* data stream
- 
**User Impact:**
- Queries spanning both old and new indices may produce inconsistent results or errors due to field type conflicts
- Sorting by vulnerability.published_date will work correctly only for data in new indices
- Data views and dashboards may show mapping conflicts

## Author's Checklist

- [ ] 

## Related issues

- 
https://github.com/elastic/sdh-security-team/issues/1477
## Screenshots

**Screenshots**
image

**Screen recording**
https://github.com/user-attachments/assets/75a01ab4-c48a-4e5e-8fdb-c85b260eb229

### DIFF
--- a/packages/cloud_security_posture/changelog.yml
+++ b/packages/cloud_security_posture/changelog.yml
@@ -1,6 +1,7 @@
 # newer versions go on top
 # version map:
 # IMPORTANT: this map doesn't apply to serverless where package availability depends on the spec version https://github.com/elastic/kibana/blob/main/config/serverless.yml#L14-L15
+# 4.0.x - 9.2.x
 # 3.1.x - 9.2.x
 # 3.0.x - 9.1.x
 # 2.0.x - 8.19.x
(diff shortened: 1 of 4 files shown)


maxcold:
From the discussion in Slack:
The main concern is that this breaking change forces users to manually recreate the latest index (and reindex data?). This might be too much to ask from our customers, and they may end up with mapping conflicts. If we are to do that, we need to agree on the strategy and make the trade-offs clear for the product folks.
An alternative is to implement first the approach we have for misconfigurations - have the destination index versioned and add an alias that won't change with breaking changes. Maybe doing that first would make the mapping change smoother

PR #15824, opened 2025-10-30
Citrix adc update documentation

jrmolin:
## Proposed commit message
[citrix_adc] Update Documentation via LLM

* ask an LLM to use the existing documentation to fill out the must-be-correct information
* ask an LLM to use the existing documentation, as well as web queries to update the README.md to the new format / template

## Checklist

- [ ] new documentation is at least not inferior to previous documentation

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/integration-experience/issues/420

## Screenshots

localhost_5601_app_integrations_detail_citrix_adc-1 17 6_overview

### DIFF
--- a/packages/citrix_adc/_dev/build/docs/README.md
+++ b/packages/citrix_adc/_dev/build/docs/README.md
@@ -7,7 +7,7 @@ The Citrix NetScaler (ADC) integration for Elastic enables you to collect and an
 
 ### Compatibility
 
-This integration is compatible with Citrix NetScaler (ADC) versions that produce syslog data in the standard format. It has been tested with NetScaler version 13.x.
+This integration is compatible with Citrix NetScaler (ADC) versions that produce syslog data in the standard format. It has been tested with NetScaler version `v13.0`, `v13.1`, and `v14.1`.
 
 ### How it works
 
@@ -23,14 +23,14 @@ To maximize performance and minimize scaling concerns, you can perform the follo
 
 ## What data does this integration collect?
 
-The Citrix NetScaler (ADC) integration collects logs and events from the following data streams:
+The Citrix NetScaler (ADC) integration collects logs, metrics, and events from the following data streams:
 
-*   **`system`**: System-level events, including hardware, configuration, and system status messages.
-*   **`vpn`**: Events related to VPN user activity, such as logins, logouts, and session details.
-*   **`lbvserver`**: Performance and health metrics for load balancing virtual servers.
-*   **`interface`**: Network interface statistics and status events.
-*   **`service`**: Backend service health, status, and performance metrics.
-*   **`log`**: General log messages and events that provide insight into the appliance's operations.
+*   **`interface`**: Collects metrics related to id, state, inbound packets, outbound packets and received packets for Citrix ADC interfaces, which are numbered in slot/port notation.
+*   **`lbvserver`**: Collects metrics related to name, state, client connections, requests and responses for the load balancing server, which is logically located between the client and the server farm and manages traffic flow.
+*   **`service`**: Collects metrics like throughput, client-server connections, and request bytes for Service resources, identified by name, IP address, and port.
+*   **`system`**: Collects metrics like memory in use, total system memory, and CPU count for system resources.
+*   **`vpn`**: Collects metrics like CPS, ICA license, client-server requests, file system and sockets for Citrix VPN, which provides full SSL VPN capabilities to Citrix Gateway.
+*   **`logs`**: Provides events from the configured syslog server.
 
 ## What do I need to use this integration?
 
@@ -47,6 +47,10 @@ Elastic Agent must be installed to collect data from your NetScaler appliance. Y
 For detailed installation instructions, refer to the Elastic Agent [installation guide](https://www.elastic.co/docs/reference/fleet/install-elastic-agents). Only one Elastic Agent is needed per host.
 
 ### Onboard and configure
+
+For step-by-step instructions on how to set up an integration, check the [quick start](integrations://docs/extend/quick-start.md).
+
+**NOTE:** It is recommended to configure the application firewall to enable CEF-formatted logs.
 
 #### Set up steps in Citrix NetScaler (ADC)
 
--- a/packages/citrix_adc/docs/README.md
+++ b/packages/citrix_adc/docs/README.md
@@ -8,7 +8,7 @@ The Citrix NetScaler (ADC) integration for Elastic enables you to collect and an
 
 ### Compatibility
 
-This integration is compatible with Citrix NetScaler (ADC) versions that produce syslog data in the standard format. It has been tested with NetScaler version 13.x.
+This integration is compatible with Citrix NetScaler (ADC) versions that produce syslog data in the standard format. It has been tested with NetScaler version `v13.0`, `v13.1`, and `v14.1`.
 
 ### How it works
 
@@ -24,14 +24,14 @@ To maximize performance and minimize scaling concerns, you can perform the follo
 
 ## What data does this integration collect?
 
-The Citrix NetScaler (ADC) integration collects logs and events from the following data streams:
+The Citrix NetScaler (ADC) integration collects logs, metrics, and events from the following data streams:
 
-*   **`system`**: System-level events, including hardware, configuration, and system status messages.
-*   **`vpn`**: Events related to VPN user activity, such as logins, logouts, and session details.
-*   **`lbvserver`**: Performance and health metrics for load balancing virtual servers.
-*   **`interface`**: Network interface statistics and status events.
-*   **`service`**: Backend service health, status, and performance metrics.
-*   **`log`**: General log messages and events that provide insight into the appliance's operations.
+*   **`interface`**: Collects metrics related to id, state, inbound packets, outbound packets and received packets for Citrix ADC interfaces, which are numbered in slot/port notation.
+*   **`lbvserver`**: Collects metrics related to name, state, client connections, requests and responses for the load balancing server, which is logically located between the client and the server farm and manages traffic flow.
+*   **`service`**: Collects metrics like throughput, client-server connections, and request bytes for Service resources, identified by name, IP address, and port.
+*   **`system`**: Collects metrics like memory in use, total system memory, and CPU count for system resources.
+*   **`vpn`**: Collects metrics like CPS, ICA license, client-server requests, file system and sockets for Citrix VPN, which provides full SSL VPN capabilities to Citrix Gateway.
+*   **`logs`**: Provides events from the configured syslog server.
 
 ## What do I need to use this integration?
 
@@ -48,6 +48,10 @@ Elastic Agent must be installed to collect data from your NetScaler appliance. Y
 For detailed installation instructions, refer to the Elastic Agent [installation guide](https://www.elastic.co/docs/reference/fleet/install-elastic-agents). Only one Elastic Agent is needed per host.
 
 ### Onboard and configure
+
+For step-by-step instructions on how to set up an integration, check the [quick start](integrations://docs/extend/quick-start.md).
+
+**NOTE:** It is recommended to configure the application firewall to enable CEF-formatted logs.
 
 #### Set up steps in Citrix NetScaler (ADC)
 
--- a/packages/citrix_adc/docs/knowledge_base/service_info.md
+++ b/packages/citrix_adc/docs/knowledge_base/service_info.md
@@ -12,6 +12,12 @@
 - Metrics: `interface`, `lbvserver`, `service`, `system`, `vpn` (collected via HTTP/JSON polling of Citrix APIs).
 - Logs: Citrix NetScaler syslog events. CEF format is recommended; RFC 5424-compliant syslog is recommended (supported in NetScaler 14.1+).
 - Example events and exported fields for each dataset are available in the integration README.
+  - **`interface`**: This is the `interface` data stream. The Citrix ADC interfaces are numbered in slot/port notation. In addition to modifying the characteristics of individual interfaces, you can configure virtual LANs to restrict traffic to specific groups of hosts. `interface` data stream collects metrics related to id, state, inbound packets, outbound packets and received packets.
+  - **`lbvserver`**: This is the `lbvserver` data stream. The load balancing server is logically located between the client and the server farm, and manages traffic flow to the servers in the server farm. `lbvserver` data stream collects metrics related to name, state, client connections, requests and responses.
+  - **`service`**: This is the `service` data stream. With the help of the service endpoint, metrics like throughput, client-server connections, request bytes can be collected along with other statistics for Service resources. `service` data stream collects metrics related to name, IP address, port, throughput and transactions.
+  - **`system`**: This is the `system` data stream. With the help of the system endpoint, metrics like memory in use, total system memory, CPU count can be collected along with other statistics for system resources.
+  - **`vpn`**: This is the `vpn` data stream. Citrix VPN is the add-on that provides full Secure Sockets Layer (SSL) virtual private network (VPN) capabilities to Citrix Gateway, allowing users to access remote applications on internal networks securely. `vpn` data stream collects metrics like CPS, ICA license, client-server requests, file system and sockets.
+  - **`logs`**: The `citrix_adc.log` dataset provides events from the configured syslog server.
 
 ## Compatibility
 - Tested against Citrix ADC versions `v13.0`, `v13.1`, and `v14.1`.
@@ -23,6 +29,10 @@
 - Use multiple Agent inputs or scale syslog receivers as ingestion volume increases. Vendor-specific performance guidance is not provided in the README.
 
 # Set Up Instructions
+
+For step-by-step instructions on how to set up an integration, check the [quick start](integrations://docs/extend/quick-start.md).
+
+**NOTE:** It is recommended to configure the application firewall to enable CEF-formatted logs.
 
 ## Vendor prerequisites
 - Host(s) and administrator credentials for the Citrix ADC instance.


taylor-swanson (commented):
I see `8.12.0` listed several times as the minimum stack version, but the manifest defines `8.13.0` as the minimum version. Where is `8.12.0` coming from?

jrmolin:
I used Cursor to generate the ground-truth-file, and that chose 8.12.0, which I missed. Thanks for catching that, @taylor-swanson !

packages/citrix_adc/_dev/build/docs/README.md
```diff
@@ -1,166 +1,180 @@
...
+    *   **File:** set **Paths**.
+    *   **TCP:** set **Listen Address** and **Listen Port**.
+    *   **UDP:** set **Listen Address** and **Listen Port**.
+5.  Save the integration.
+
+### Validation
+
+After configuration, open the Citrix ADC integration’s **Assets** tab to view the available dashboards. They should populate with data for the configured datasets. You can also verify metrics and logs in **Discover** under the `logs-*` data view.
+
+## Troubleshooting
```
qcorporation: @jrmolin I feel the troubleshoot section is actually worst than the original one.
what happened to this section? Is it not valuable in documentation?


-------
Dummy values
It is possible that for some fields, Citrix ADC sets dummy values. For example, a field cpuusagepcnt is represented by citrix_adc.system.cpu.utilization.pct. cpuusagepcnt is set to 4294967295 for some https://github.com/citrix/citrix-adc-metrics-exporter/issues/44. If you also encounter it for some fields, reach out to the [Citrix ADC support team](https://support.citrix.com/plp/products/citrix_adc/tabs/popular-solutions).

Type conflicts
If host.ip is shown conflicted under logs-* data view, this issue can be solved by [reindexing](https://www.elastic.co/guide/en/elasticsearch/reference/current/use-a-data-stream.html#reindex-with-a-data-stream) the Interface, LBVserver, Service, System, and VPN data stream's indices.

mjwolf:
This should also increase the integration version, so the new docs will be published when this is merged

packages/citrix_adc/_dev/build/docs/README.md
mjwolf: I think if you put more content from the original readme into service_info, you'll get something that's more similar to the original

packages/citrix_adc/docs/knowledge_base/service_info.md
```diff
@@ -0,0 +1,104 @@
...
+- NITRO API metrics references:
+  - Interface: [link](https://developer-docs.citrix.com/projects/netscaler-nitro-api/en/12.0/statistics/network/interface/)
+  - LBVserver: [link](https://developer-docs.citrix.com/projects/netscaler-nitro-api/en/12.0/statistics/lb/lbvserver/)
+  - Service: [link](https://developer-docs.citrix.com/projects/netscaler-nitro-api/en/12.0/statistics/basic/service/)
+  - System: [link](https://developer-docs.citrix.com/projects/netscaler-nitro-api/en/12.0/statistics/system/system/)
+  - VPN: [link](https://developer-docs.citrix.com/projects/netscaler-nitro-api/en/12.0/statistics/vpn/vpn/)
+
+# Documentation sites
+- Elastic ECS Field Reference: [link](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html)
+- Citrix ADC integration README in this package for examples, exported fields, and setup notes.
```
mjwolf: I don't think this is needed, the built-in LLM prompt has instructions to read the existing package files

packages/citrix_adc/docs/knowledge_base/service_info.md
```diff
@@ -0,0 +1,104 @@
...
+    - Listen Port
+- You can also verify metrics and logs in Discover under the `logs-*` data view.
+
+# Troubleshooting
+
+## Dummy values
+It is possible that for some fields, Citrix ADC sets dummy values. For example, a field cpuusagepcnt is represented by citrix_adc.system.cpu.utilization.pct. cpuusagepcnt is set to 4294967295 for some instances. If you also encounter it for some fields, reach out to the Citrix ADC support team.
+
+## Type conflicts
+If host.ip is shown conflicted under logs-* data view, this issue can be solved by reindexing the Interface, LBVserver, Service, System, and VPN data stream's indices.
```
mjwolf: Troubleshooting should also have the solutions to problems listed

packages/citrix_adc/docs/knowledge_base/service_info.md
```diff
@@ -0,0 +1,104 @@
...
+4. Choose how to collect logs and configure parameters:
+   - File: set Paths.
+   - TCP: set Listen Address and Listen Port.
+   - UDP: set Listen Address and Listen Port.
+5. Save the integration.
+6. Note: It is recommended to enable CEF logging on the application firewall.
+
+# Validation Steps
+- After configuration, open the Citrix ADC integration’s Assets tab to view the available dashboards; they should populate with data for the configured datasets.
+- While adding the integration, if you want to collect logs via logfile, keep Collect logs from Citrix ADC via file toggle on and then configure following parameters:
```
mjwolf: These `While adding the integration...` lines would work better in the Kibana set up steps sections. Validation should assume everything is configured already, and is just how to check data is flowing

packages/citrix_adc/docs/knowledge_base/service_info.md
```diff
@@ -0,0 +1,104 @@
...
+- Understand virtual server load, client/server connections, requests, and responses across Citrix ADC.
+
+## Data types collected
+- Metrics: `interface`, `lbvserver`, `service`, `system`, `vpn` (collected via HTTP/JSON polling of Citrix APIs).
+- Logs: Citrix NetScaler syslog events. CEF format is recommended; RFC 5424-compliant syslog is recommended (supported in NetScaler 14.1+).
+- Example events and exported fields for each dataset are available in the integration README.
+
+## Compatibility
+- Tested against Citrix ADC versions `v13.0`, `v13.1`, and `v14.1`.
+- Minimum Kibana version: `8.13.0`.
```
mjwolf: The minimum kibana version shouldn't be needed here, the info can be found in the manifest file already

packages/citrix_adc/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.17.6"
+  changes:
+    - description: Update documentation with LLM
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
```
mjwolf: the link should be updated

packages/citrix_adc/docs/knowledge_base/service_info.md
jrmolin: i don't know what you mean here. this whole file was auto-generated, and i'm not in love with what it did.

mjwolf: I guess it does have a solution already, I was expecting something different, more like:

```
Problem: Type conflict errors are reported in Kibana (for example, `host.ip` under `logs-*`)
Solution: Reindex the `Interface`, `LBVserver`, `Service`, `System`, and `VPN` data stream indices.
```

There's nothing for dashboards not populated in the final generated document, so that would be good to add. I don't know if it could be because the LLM had trouble interpretting the problem/solution statement here, or something else. This is also a pretty generic troubleshooting step, so maybe it doesn't really make sense to include in the final doc

mjwolf:
@jrmolin It looks like you need to rebuild the package and commit the updated `docs/README.md`, to pass the CI check

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+<!-- This file is automatically generated by Elastic Package -->
+# Citrix NetScaler (ADC) Integration for Elastic
+
+## Overview
+
+The Citrix NetScaler (ADC) integration for Elastic enables you to collect and analyze logs and events from your Citrix NetScaler appliances. It provides visibility into system health, performance metrics for load balancing virtual servers, VPN user activity, and more. This integration facilitates monitoring, troubleshooting, and securing your application delivery infrastructure by parsing and visualizing critical operational data.
+
+### Compatibility
+
+This integration is compatible with Citrix NetScaler (ADC) versions that produce syslog data in the standard format. It has been tested with NetScaler version 13.x.
```
ishleenk17: This compatibility section is incorrect. 
The versions its compatible with is mentioned in the actual documentation as` v13.0`, `v13.1` and `v14.1`
Pls see [here](https://github.com/elastic/integrations/pull/15824/files#diff-1eb9db55c17000b88f1831de49d97aede088119068f67c4b7a6057f012e11065L29). 

Was this incorrectly generated by LLM ?

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+
+### How it works
+
+The integration collects syslog data streamed from a configured Citrix NetScaler appliance. Elastic Agent listens for these logs on a specified port, processes them using the integration's ingest pipelines to parse and structure the data, and sends the enriched events to your Elastic cluster.
```
ishleenk17, line 15: Here, we are mentioning only about logs. No mention about metrics collection ?
Also, I don't see any Setup section in the README unlike before. See [here](https://github.com/elastic/integrations/pull/15824/files#diff-c44be15438591ff5e47b818d87eff9be03565326f8552bc156b71397cc0f179cL44). This is critical. Looks like it got missed out here ?

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+
+### Scaling and Performance
+
+To maximize performance and minimize scaling concerns, you can perform the following:
+
+* Prefer CEF logging for application firewall events; use RFC 5424-compliant syslog where supported (NetScaler 14.1+).
```
ishleenk17, line 21: How is this related to scaling ?

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+* Choose the appropriate log transport (file, TCP, or UDP) based on your volume and reliability needs.
+* Use multiple Agent inputs or scale syslog receivers as ingestion volume increases.
```
ishleenk17, line 23: The scaling section is very generic. Which mentions just usage of more elastic agents if required. 
IMO, this doesn't add any value to the users reading about this Integration. It needs to be more specific to the Integration. Also if we are mentioning about scaling we need to be sure that we have validated that the scaling procedure works fine.

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+
+## What data does this integration collect?
+
+The Citrix NetScaler (ADC) integration collects logs and events from the following data streams:
```
ishleenk17: This integration collects logs and metrics. 
No mention about metrics here.

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+
+*   **`system`**: System-level events, including hardware, configuration, and system status messages.
```
ishleenk17: System, vpn, interface are metric datastreams. The documentation mentions it as events which is incorrect.

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+*   **`vpn`**: Events related to VPN user activity, such as logins, logouts, and session details.
+*   **`lbvserver`**: Performance and health metrics for load balancing virtual servers.
+*   **`interface`**: Network interface statistics and status events.
+*   **`service`**: Backend service health, status, and performance metrics.
```
ishleenk17: Another observation. 
When we write the documentation, we prefer mentioning about the specific metric types that we collect in the datastreams. Refer [here](https://github.com/elastic/integrations/pull/15824/files#diff-1eb9db55c17000b88f1831de49d97aede088119068f67c4b7a6057f012e11065L120) for service datastream.
LLM makes it very generic. 
IMO, customer would gain by specific details than generic. WDYT ?

ishleenk17: The same holds true for other datastreams as well.

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+
+### Agent-based deployment
+
+Elastic Agent must be installed to collect data from your NetScaler appliance. You can install only one Elastic Agent per host. Elastic Agent streams the syslog data to your Elastic cluster, where the events are processed by the integration's ingest pipelines.
+
+For detailed installation instructions, refer to the Elastic Agent [installation guide](https://www.elastic.co/docs/reference/fleet/install-elastic-agents). Only one Elastic Agent is needed per host.
+
+### Onboard and configure
+
+#### Set up steps in Citrix NetScaler (ADC)
```
ishleenk17, line 56: This mentions steps only for logs. Also no CEF related details mentioned here as mentioned [previously](https://github.com/elastic/integrations/pull/15824/files#diff-1eb9db55c17000b88f1831de49d97aede088119068f67c4b7a6057f012e11065L50).

It would be good if the tool would have suggested adding configuration for metrics as thats the missing part here.

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+    *   **Log Levels**: Select the desired log levels (e.g., ALL).
+5.  Click **Create** to save the configuration.
+
+#### Set up steps in Kibana
+
+1.  From the Kibana main menu, go to **Management** > **Integrations**.
+2.  Search for "Citrix ADC" and select it.
+3.  Click **Add Citrix NetScaler (ADC)**.
+4.  Configure the integration policy:
+    *   **Syslog Host**: Set to `0.0.0.0` to listen on all network interfaces of the agent host.
```
ishleenk17, line 73: Is this value to be set always as 0.0.0.0. 
For such scenarios we need to validate first ?

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+1.  Log in to your Citrix NetScaler (ADC) management interface.
+2.  Navigate to **Configuration** > **System** > **Auditing** > **Syslog**.
+3.  In the **Servers** tab, click **Add** to create a new syslog server entry.
+4.  Configure the following settings:
+    *   **Server Name / IP Address**: Enter the IP address of the host where your Elastic Agent is running.
+    *   **Port**: Enter the port number that you will configure in the Elastic Agent integration policy (e.g., 9001).
+    *   **Log Levels**: Select the desired log levels (e.g., ALL).
+5.  Click **Create** to save the configuration.
+
+#### Set up steps in Kibana
```
ishleenk17, line 67: Here we are mentioning how to setup an integration. 

`Setup steps in Kibana` title seems confusing. Can we have it as` setup of Integration` ?
This section is not part of the [template](https://github.com/elastic/integrations/blob/b34c073e115758efb489cbe2d1cae734d837cd76/docs/extend/documentation-guidelines.md). Was this section manually added ?

packages/citrix_adc/docs/README.md
ishleenk17: This [type](https://github.com/elastic/integrations/pull/15824/files#diff-1eb9db55c17000b88f1831de49d97aede088119068f67c4b7a6057f012e11065L72) of mention in setting up an Integration looks important. It is getting missed here.

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+
+1.  From the Kibana main menu, go to **Management** > **Integrations**.
+2.  Search for "Citrix ADC" and select it.
+3.  Click **Add Citrix NetScaler (ADC)**.
+4.  Configure the integration policy:
+    *   **Syslog Host**: Set to `0.0.0.0` to listen on all network interfaces of the agent host.
+    *   **Syslog Port**: Enter the same port number you configured in the NetScaler syslog settings.
+5.  Save the integration policy and assign it to your Elastic Agent(s).
```
ishleenk17, line 75: These steps don't mention how to setup the metrics section from the Integration.

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+
+### Troubleshooting
```
ishleenk17, line 77: This is a very generic troubleshooting. 
The LLM's have removed the important part of troubleshooting from the EADME. See [here](https://github.com/elastic/integrations/pull/15824/files#diff-1eb9db55c17000b88f1831de49d97aede088119068f67c4b7a6057f012e11065L84)

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+
+If you are not seeing data, verify the following:
+*   Ensure the NetScaler appliance can reach the Elastic Agent host over the configured port. Check for firewalls or network security groups that might be blocking the traffic.
+*   Verify that the port configured in the NetScaler syslog settings matches the port in the Elastic Agent integration policy.
+*   Check the Elastic Agent logs for any errors related to receiving syslog data.
+
+## Reference
+
+### Vendor Resources
```
ishleenk17, line 86: Do we think these external references are important for a user of our Integration ?
As per our [template](https://github.com/elastic/integrations/blob/b34c073e115758efb489cbe2d1cae734d837cd76/docs/extend/documentation-guidelines.md#reference-idg-docs-reference), reference means something else.

packages/citrix_adc/docs/README.md
```diff
@@ -1450,4 +1304,214 @@ Check this [reference document](https://www.elastic.co/guide/en/ecs/current/ecs-
...
+        "extension": "html",
+        "original": "http://vpx247.example.net/FFC/login_post.html?abc\\=def",
+        "path": "/FFC/login_post.html",
+        "query": "abc\\=def",
+        "scheme": "http"
+    }
+}
+```
+
+### Inputs used
```
ishleenk17, line 1430: Do we need this information for the user ? 
Isn't it giving too much information in our README ?
And this doens't look to be part of our current agreed upon [template](https://github.com/elastic/integrations/blob/b34c073e115758efb489cbe2d1cae734d837cd76/docs/extend/documentation-guidelines.md#documentation-guidelines-documentation-guidelines) ?

packages/citrix_adc/docs/README.md
```diff
@@ -1,221 +1,201 @@
...
+- How to send WAF messages to a separate syslog server: [link](https://support.citrix.com/s/article/CTX138973-how-to-send-application-firewall-messages-to-a-separate-syslog-server)
+- How to send NetScaler Application Firewall logs to external syslog/NS.log: [link](https://support.citrix.com/s/article/CTX483235-send-logs-to-external-syslog-server?language=en_US)
+- Configuring audit log action (RFC 5424): [link](https://docs.netscaler.com/en-us/citrix-adc/current-release/system/audit-logging/configuring-audit-logging.html#configuring-audit-log-action)
+- NITRO API metrics references:
+  - Interface: [link](https://developer-docs.citrix.com/projects/netscaler-nitro-api/en/12.0/statistics/network/interface/)
+  - LBVserver: [link](https://developer-docs.citrix.com/projects/netscaler-nitro-api/en/12.0/statistics/lb/lbvserver/)
+  - Service: [link](https://developer-docs.citrix.com/projects/netscaler-nitro-api/en/12.0/statistics/basic/service/)
+  - System: [link](https://developer-docs.citrix.com/projects/netscaler-nitro-api/en/12.0/statistics/system/system/)
+  - VPN: [link](https://developer-docs.citrix.com/projects/netscaler-nitro-api/en/12.0/statistics/vpn/vpn/)
+
```
ishleenk17, line 97: Certain other important points missed out that was mentioned in previous documentation:

1. We are using HTTPJSON fo metrics collection . Hence mention of [this](https://github.com/elastic/integrations/pull/15824/files#diff-1eb9db55c17000b88f1831de49d97aede088119068f67c4b7a6057f012e11065L25) is important but was missed out.
2. [Kibana version](https://github.com/elastic/integrations/pull/15824/files#diff-1eb9db55c17000b88f1831de49d97aede088119068f67c4b7a6057f012e11065L35) mention missed out.
3. Metrics configuration like [here](https://github.com/elastic/integrations/pull/15824/files#diff-1eb9db55c17000b88f1831de49d97aede088119068f67c4b7a6057f012e11065L39).
4. Previous troubleshooting steps were much more specific to the integration.

ishleenk17 (commented):
Thanks for the PR. 
We have reviewed the documentation update.
IMO, while the LLM documentation provides helpful context, it is adding a lot of information which is very generic in nature, and not specific to the Integration.

It would be good if the LLM's could help us pointing out some of the missing config steps required for metrics collection.

Please see the detailed review inline.

packages/citrix_adc/docs/README.md
jrmolin: yes, this was incorrectly generated by the LLM. the correct information is in the knowledge base.

packages/citrix_adc/docs/README.md
jrmolin: i added those bits back in. if we re-generate, i'm worried that these will be removed again.

packages/citrix_adc/docs/README.md
jrmolin: i think CEF scales better than syslog. I'm not sure why this was put in here, but it's in the knowledge base.

packages/citrix_adc/docs/README.md
jrmolin: so you'd suggest removing this? i'm fine doing that. i don't know how to tell the LLM not to put generic information in the documentation.

packages/citrix_adc/docs/README.md
jrmolin: this is indicated in the knowledge base. i'll change `events` to `metrics` where appropriate

jrmolin:
closing for now

PR #13856, opened 2025-05-08
buildkite: use GCP OIDC

v1v:
## Proposed commit message

- Use GCP OIDC to copy the logs to the private storage.
- Use gcloud instead of gsutil (so it honours the Env variables)
- Tear-down is now managed by the BK plugin itself.

### Test

See the ci build

image

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [ ] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## Related issues

Relates to https://github.com/elastic/elastic-package/pull/2569

PR #15829, opened 2025-10-31
[Security Rules] Integrate security_detection_engine OOM testing pipeline

maximpn:
**Partially addresses:** https://github.com/elastic/kibana/issues/188090

## Summary

This PR integrates [Prebuilt Rules OOM testing Buildkite pipeline](https://buildkite.com/elastic/appex-qa-stateful-security-prebuilt-rules-ftr-oom-testing) into the Pull Request Buildkite pipeline.

## Details

Pull Request Builkite pipeline script have been extended in a generic way to support custom package checker scripts located under `<repo-root>/.buildkite/scripts/packages/<package-name>.sh`.  It allows to run any custom verification and testing logic specific to a package.

This PR adds `.buildkite/scripts/packages/security_detection_engine.sh` script file. This script runs only for **security_detection_engine** package and triggers the [Prebuilt Rules Out-Of-Memory testing pipeline](https://buildkite.com/elastic/appex-qa-stateful-security-prebuilt-rules-ftr-oom-testing). The triggered pipeline performs e2e testing to reveal potential blockers due to Kibana Out-Of-Memory instance failures when performing actions upon the package (installing the package, review prebuilt rules available in the package, installing prebuilt rules from the package etc.).

### Tested stack versions

For now `.buildkite/scripts/packages/security_detection_engine.sh` triggers [Prebuilt Rules OOM testing Buildkite pipeline](https://buildkite.com/elastic/appex-qa-stateful-security-prebuilt-rules-ftr-oom-testing) against compatible minor versions under development. The decision is made based on Kibana's [versions.json](https://raw.githubusercontent.com/elastic/kibana/main/versions.json). While compatibility is determined via `conditions.kibana.version` field in the package's `manifest.yml`.

For example `conditions.kibana.version` has `^9.2.0` restriction and we have `9.2.2` and `9.3.0` under development. It means the OOM tests will run against  `9.2.2-SNAPSHOT` and `9.3.0-SNAPSHOT`.

We consider extending the testing surface to the latest release patch versions after collecting more data in the CI runs.

## Affected teams

@elastic/threat-research-and-detection-engineering,

FYI this PR will affect **security_detection_engine** package release process. Every PR containing changes to the **security_detection_engine** package will trigger [Prebuilt Rules OOM testing ECH Buildkite pipeline](https://buildkite.com/elastic/appex-qa-stateful-security-prebuilt-rules-ftr-oom-testing).

## Further improvements

- Pushing commits to this repo in a quick succession may lead to leaving rouge resources in the cloud. It happens due to `cancel_intermediate_builds: true` configuration at the Integrations PR Buildkite build. Pushing a fresh commit cancels the currently running PR build leading to cancelling the triggered build. Eventually the clean up steps in the triggered build can't execute and clean up resources in the cloud. 
- We may speed up the build by using an **elastic-package** Docker container published to `docker.elastic.co`. **elastic-package** installation is a complex process requiring a chain on installations GVM -> Go -> elastic-package. And it takes in average **3 minutes** per each integration (integrations build in parallel). On top of that [Prebuilt Rules OOM testing Buildkite pipeline](https://buildkite.com/elastic/appex-qa-stateful-security-prebuilt-rules-ftr-oom-testing) has to install **elastic-package** as well. It sums up to **6 minutes** which could be reduced.

### DIFF
--- a/packages/azure_ai_foundry/changelog.yml
+++ b/packages/azure_ai_foundry/changelog.yml
@@ -1,3 +1,8 @@
+- version: "0.7.0"
+  changes:
+    - description: Add Alerting Rule Templates.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/15411
 - version: "0.6.1"
   changes:
     - description: Update overview and guardrails dashboard panel to use donut chart and fix protected_material_code filter.
--- a/packages/azure_ai_foundry/kibana/alerting_rule_template/azure_ai_foundry-latency-spike.json
+++ b/packages/azure_ai_foundry/kibana/alerting_rule_template/azure_ai_foundry-latency-spike.json
@@ -1,0 +1,28 @@
+{
+  "id": "azure_ai_foundry-latency-spike",
+  "type": "alerting_rule_template",
+  "attributes": {
+    "name": "[Azure AI Foundry] Latency high",
+    "tags": ["Azure AI Foundry"],
+    "ruleTypeId": ".es-query",
+    "schedule": {
+      "interval": "5m"
+    },
+    "params": {
+      "searchType": "esqlQuery",
+      "timeWindowSize": 10,
+      "timeWindowUnit": "m",
+      "esqlQuery": {
+        "esql": "// Alert triggers when the response latency exceeds the recommended threshold value {5000ms} within the look back time window.\n// The alert is grouped by Model Deployment Name.\n// You can adjust the threshold value by modifying the time_to_response in the WHERE clause, which is specified in milliseconds.\nFROM metrics-azure.ai_foundry-default\n| KEEP azure.ai_foundry.time_to_response.avg, azure.dimensions.model_deployment_name, @timestamp\n| WHERE azure.dimensions.model_deployment_name IS NOT NULL\n| STATS time_to_response = MAX(azure.ai_foundry.time_to_response.avg) BY azure.dimensions.model_deployment_name\n| WHERE time_to_response > 5000\n| EVAL time_to_response = ROUND(time_to_response, 2)\n| SORT time_to_response DESC"
+      },
+      "groupBy": "row",
+      "timeField": "@timestamp"
+    },
+    "alertDelay": {
+      "active": 2
+    }
+  },
+  "managed": true,
+  "coreMigrationVersion": "8.8.0",
+  "typeMigrationVersion": "10.1.0"
+}
--- a/packages/azure_ai_foundry/kibana/alerting_rule_template/azure_ai_foundry-model-availability.json
+++ b/packages/azure_ai_foundry/kibana/alerting_rule_template/azure_ai_foundry-model-availability.json
@@ -1,0 +1,28 @@
+{
+  "id": "azure_ai_foundry-model-availability",
+  "type": "alerting_rule_template",
+  "attributes": {
+    "name": "[Azure AI Foundry] Model Availability above threshold",
+    "tags": ["Azure AI Foundry"],
+    "ruleTypeId": ".es-query",
+    "schedule": {
+      "interval": "5m"
+    },
+    "params": {
+      "searchType": "esqlQuery",
+      "timeWindowSize": 10,
+      "timeWindowUnit": "m",
+      "esqlQuery": {
+        "esql": "// Alert triggers when the model availability is less than the recommended threshold value {99%} within the look back time window.\n// The alert is grouped by Model Deployment Name.\n// You can adjust the threshold value by modifying the model_availability in the WHERE clause, which is specified in percent.\nFROM metrics-azure.ai_foundry-default\n| KEEP azure.ai_foundry.model_availability_rate.avg, azure.dimensions.model_deployment_name, @timestamp\n| WHERE azure.dimensions.model_deployment_name IS NOT NULL\n| STATS model_availability = MIN(azure.ai_foundry.model_availability_rate.avg) * 100 BY azure.dimensions.model_deployment_name\n| WHERE model_availability < 99\n| EVAL model_availability = ROUND(model_availability, 2)\n| SORT model_availability ASC"
+      },
+      "groupBy": "row",
+      "timeField": "@timestamp"
+    },
+    "alertDelay": {
+      "active": 3
+    }
+  },
+  "managed": true,
+  "coreMigrationVersion": "8.8.0",
+  "typeMigrationVersion": "10.1.0"
+}
--- a/packages/azure_ai_foundry/kibana/alerting_rule_template/azure_ai_foundry-provisioned-utilization.json
+++ b/packages/azure_ai_foundry/kibana/alerting_rule_template/azure_ai_foundry-provisioned-utilization.json
@@ -1,0 +1,28 @@
+{
+  "id": "azure_ai_foundry-provisioned-utilization",
+  "type": "alerting_rule_template",
+  "attributes": {
+    "name": "[Azure AI Foundry] Provisioned Utilization above threshold",
+    "tags": ["Azure AI Foundry"],
+    "ruleTypeId": ".es-query",
+    "schedule": {
+      "interval": "5m"
+    },
+    "params": {
+      "searchType": "esqlQuery",
+      "timeWindowSize": 10,
+      "timeWindowUnit": "m",
+      "esqlQuery": {
+        "esql": "// Alert triggers when the provisioned utilization exceeds the recommended threshold value {85%} within the look back time window.\n// The alert is grouped by Model Deployment Name.\n// You can adjust the threshold value by modifying the provisioned_utilization in the WHERE clause, which is specified in percent.\nFROM metrics-azure.ai_foundry-default\n| KEEP azure.ai_foundry.provisioned_utilization.avg, azure.dimensions.model_deployment_name, @timestamp\n| WHERE azure.dimensions.model_deployment_name IS NOT NULL\n| STATS provisioned_utilization = MAX(azure.ai_foundry.provisioned_utilization.avg) * 100 BY azure.dimensions.model_deployment_name\n| WHERE provisioned_utilization > 85\n| EVAL provisioned_utilization = ROUND(provisioned_utilization, 2)\n| SORT provisioned_utilization DESC"
+      },
+      "groupBy": "row",
+      "timeField": "@timestamp"
+    },
+    "alertDelay": {
+      "active": 2
+    }
+  },
+  "managed": true,
+  "coreMigrationVersion": "8.8.0",
+  "typeMigrationVersion": "10.1.0"
+}
--- a/packages/azure_ai_foundry/manifest.yml
+++ b/packages/azure_ai_foundry/manifest.yml
@@ -1,7 +1,7 @@
-format_version: 3.3.2
+format_version: 3.4.0
 name: azure_ai_foundry
 title: "Azure AI Foundry"
-version: "0.6.1"
+version: "0.7.0"
 source:
   license: "Elastic-2.0"
 description: "Collects Azure AI Foundry logs and metrics"
@@ -14,7 +14,7 @@ categories:
   - security
 conditions:
   kibana:
-    version: "^9.0.0"
+    version: "^9.2.1"
   elastic:
     subscription: "basic"
 vars:
--- a/packages/azure_openai/changelog.yml
+++ b/packages/azure_openai/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.10.0"
+  changes:
+    - description: Add Alerting Rule Templates.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/15412
 - version: "1.9.0"
   changes:
     - description: Add a flag `fips_compatible` to control whether the package is allowed in the ECH FedRAMP High environment.
--- a/packages/azure_openai/kibana/alerting_rule_template/azure_openai-latency-spike.json
+++ b/packages/azure_openai/kibana/alerting_rule_template/azure_openai-latency-spike.json
@@ -1,0 +1,28 @@
+{
+  "id": "azure_openai-latency-spike",
+  "type": "alerting_rule_template",
+  "attributes": {
+    "name": "[Azure OpenAI] Latency high",
+    "tags": ["Azure OpenAI"],
+    "ruleTypeId": ".es-query",
+    "schedule": {
+      "interval": "1m"
+    },
+    "params": {
+      "searchType": "esqlQuery",
+      "timeWindowSize": 10,
+      "timeWindowUnit": "m",
+      "esqlQuery": {
+        "esql": "// Alert triggers when the response latency exceeds the recommended threshold value {5000ms} within the look back time window.\n// The alert is grouped by Model Deployment Name.\n// You can adjust the threshold value by modifying the time_to_response in the WHERE clause, which is specified in milliseconds.\nFROM metrics-azure.open_ai-default\n| KEEP azure.open_ai.time_to_response.avg, azure.dimensions.model_deployment_name, @timestamp\n| WHERE azure.dimensions.model_deployment_name IS NOT NULL\n| STATS time_to_response = MAX(azure.open_ai.time_to_response.avg) BY azure.dimensions.model_deployment_name\n| WHERE time_to_response > 5000\n| EVAL time_to_response = ROUND(time_to_response, 2)\n| SORT time_to_response DESC"
+      },
+      "groupBy": "row",
+      "timeField": "@timestamp"
+    },
+    "alertDelay": {
+      "active": 2
+    }
+  },
+  "managed": true,
+  "coreMigrationVersion": "8.8.0",
+  "typeMigrationVersion": "10.1.0"
+}
--- a/packages/azure_openai/kibana/alerting_rule_template/azure_openai-provisioned-utilization.json
+++ b/packages/azure_openai/kibana/alerting_rule_template/azure_openai-provisioned-utilization.json
@@ -1,0 +1,28 @@
+{
+  "id": "azure_openai-provisioned-utilization",
+  "type": "alerting_rule_template",
+  "attributes": {
+    "name": "[Azure OpenAI] Provisioned Utilization above threshold",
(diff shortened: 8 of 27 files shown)


.buildkite/scripts/test_one_package.sh
```diff
@@ -25,14 +25,27 @@ with_kubernetes
...
+elif [ -x "$custom_package_checker_script_path" ]; then
+  echo "--- [${package}] Run individual package checker"
+  "$custom_package_checker_script_path"
+else
+  echo "--- [${package}] Individual package checker $custom_package_checker_script_path is not found, continue..."
 fi
 popd > /dev/null
 
+pwd
+
```
mrodm: Is this a leftover? Could it be removed ?

.buildkite/scripts/test_one_package.sh
```diff
@@ -25,14 +25,27 @@ with_kubernetes
...
 if ! process_package "${package}" ; then
     # keep this message as a collapsed group in Buildkite, so it
     # is not hidden by the previous collapsed group.
     echo "--- [${package}] failed"
     exit_code=1
+elif [ -x "$custom_package_checker_script_path" ]; then
+  echo "--- [${package}] Run individual package checker"
+  "$custom_package_checker_script_path"
+else
+  echo "--- [${package}] Individual package checker $custom_package_checker_script_path is not found, continue..."
```
mrodm: Maybe this message could be skipped.
WDYT about just writing a message in the output if the script is found ?

.buildkite/scripts/packages/security_detection_engine.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -euo pipefail
+
```
mrodm, line 4: Could this be just run for now in Pull Requests ? 

In this repository there are also configured some Buildkite pipelines that run daily or weekly. If that runs for all those builds, there will be a lot of pipelines triggered every day.

This code would check that this script just runs with pull requests:
```suggestion
if [[ "${BUILDKITE_PULL_REQUEST}" == "false" ]]; then
  exit 0
fi
```

.buildkite/scripts/common.sh
```diff
@@ -763,10 +763,19 @@ is_pr_affected() {
...
     fi
+    echoerr "[${package}] git-diff: check custom package checker script file (${commit_merge}..${to})"
+    # Avoid using "-q" in grep in this pipe, it could cause that some files updated are not detected due to SIGPIPE errors when "set -o pipefail"
+    # Example:
+    # https://buildkite.com/elastic/integrations/builds/25606
+    # https://github.com/elastic/integrations/pull/13810
+    if git diff --name-only "${commit_merge}" "${to}" | grep -E "^\.buildkite/scripts/packages/${package}.sh" > /dev/null; then
+        echo "[${package}] PR is affected: found package checker script changes"
+        return 0
+    fi
```
mrodm, line 778: :+1:

.buildkite/scripts/packages/security_detection_engine.sh
mrodm, line 1: Currently, this script would just be executed to just those Pull Requests targeting the `main` branch.
This would not be executed in the existing backport branches. Is that ok for now?

Once merged, new backport branches created will include these steps too.

.buildkite/scripts/test_one_package.sh
```diff
@@ -25,14 +25,27 @@ with_kubernetes
 
 use_elastic_package
 
+# Detect an absolute path to the individual package checker script.
+# This avoids issues caused by changing the working directory.
+current_folder=$(dirname "$(realpath $0)")
+custom_package_checker_script_path="$current_folder/packages/$package.sh"
+
```
mrodm: As this script sources `.buildkite/script/common.sh`, there is a variable `SCRIPTS_BUILDKITE_PATH` that could be used to obtain that path too:

```suggestion
custom_package_checker_script_path="${SCRIPTS_BUILDKITE_PATH}/packages/${package}.sh"
```

Reference:
https://github.com/elastic/integrations/blob/6954c94dcbf18965fbfa92d5c40194499ce99cda/.buildkite/scripts/common.sh#L11

.buildkite/scripts/packages/security_detection_engine.sh
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/Masterminds/semver/v3"
+)
+
+func main() {
+	c, _ := semver.NewConstraint(os.Args[1])
+	for _, s := range os.Args[2:] {
+		if v, _ := semver.NewVersion(s); c.Check(v) {
+			fmt.Println(s + "-SNAPSHOT")
+		}
+	}
```
mrodm: Should this function take into account the errors that could be raised from parsing constraints or versions? 

```suggestion
	c, err := semver.NewConstraint(os.Args[1])
	if err != nil {
		panic(err)
	}
	for _, s := range os.Args[2:] {
		if v := semver.MustParse(s); c.Check(v) {
			fmt.Println(s + "-SNAPSHOT")
		}
	}
```

I guess that they should be valid, but it could be better to check those too.

.buildkite/scripts/common.sh
```diff
@@ -728,7 +728,7 @@ is_pr_affected() {
             return 1
         fi
         if ! is_supported_capability ; then
-            echo "[${package}] PR is not affected: capabilities not mached with the project (${SERVERLESS_PROJECT})"
+            echo "[${package}] PR is not affected: capabilities not matched with the project (${SERVERLESS_PROJECT})"
```
mrodm, line 731: Thanks!

.buildkite/scripts/test_one_package.sh
maximpn: Yes. Thanks for noticing.

However, it's important to note `process_package` changes the working directory to `<repo-root>/packages`. I'm not sure whether it's an expected behavior. For my script it manifested as failure to find a script under `./.buildkite/...`.

.buildkite/scripts/test_one_package.sh
maximpn: Makes sense.

Using the integration specific full script path makes it insensitive to changing directory by an upstream script. Meaning `cd <something>` command somewhere upstream was my main concern as the integration specific script file may not be found while it exists in the file system.

.buildkite/scripts/packages/security_detection_engine.sh
maximpn: > Currently, this script would just be executed to just those Pull Requests targeting the main branch.
This would not be executed in the existing backport branches. Is that ok for now?

We'd like this script to be executed in the supported backport branches. In particular

- `backport-security_detection_engine-9.1`
- `backport-security_detection_engine-8.19`

> Once merged, new backport branches created will include these steps too.

Yes. This is expected behavior. We expect the script to work in the new backport branches. It's expected `backport-security_detection_engine-9.2` will be created as the next backport branch.

---

@mrodm I planned to backport this PR to `backport-security_detection_engine-9.1` and `backport-security_detection_engine-8.19`. And considered doing that manually. What the recommended way to do backports?

maximpn:
@mrodm,

> > Pushing commits to this repo in a quick succession may lead to leaving rouge resources in the cloud. It happens due to cancel_intermediate_builds: true configuration at the Integrations PR Buildkite build. Pushing a fresh commit cancels the currently running PR build leading to cancelling the triggered build. Eventually the clean up steps in the triggered build can't execute and clean up resources in the cloud.
>
>  That setting is added to the Buildkite configuration to try to avoid "wasting" resources if there is a new commit pushed to the branch. I think it is interesting to keep that setting as it is.
An option that you could follow is adding some daily Buildkite job or similar to clean up those resources that could be left behind.

Ideally it should be possible to prevent the cleanup steps in the triggered pipeline from cancelling. I've explored the possibilities but it may time to figure out a proper solution. So it will be a follow up PR when there is time to address this task.

maximpn:
@mrodm,

I've addressed your comments. Could you have a look?

It seems the last build failed due to unrelated to this PR resons.

maximpn:
@mrodm,

> About the improvements mentioned in the description:
> 
> > We may speed up the build by using an elastic-package Docker container published to docker.elastic.co. elastic-package installation is a complex process requiring a chain on installations GVM -> Go -> elastic-package. And it takes in average 3 minutes per each integration (integrations build in parallel). On top of that [Prebuilt Rules OOM testing Buildkite pipeline](https://buildkite.com/elastic/appex-qa-stateful-security-prebuilt-rules-ftr-oom-testing) has to install elastic-package as well. It sums up to 6 minutes which could be reduced.
> 
> I'm not totally sure this could lead to unexpected failures, since elastic-package already is based on docker to run some commands (e.g. elastic-package test system or even elastic-package stack up). Creating a docker image for elastic-package would involve to ensure that everything works in a docker-in-a-docker scenario. I think I would try to avoid that, WDYT @elastic/ecosystem ?

I just wanted to clarify my suggestion.

AFAIK Buildkite jobs run in a Kibernetes environment. For example **elastic/integrations** pipelines already [use](https://github.com/elastic/integrations/blob/main/.buildkite/pipeline.yml#L13-L14) Go and Ubuntu images. 

Consequently it opens up a possibility to build a custom image with **elastic-package** from a standard Go image and use it for Buildklite jobs/steps requiring **elastic-package**.

.buildkite/scripts/packages/security_detection_engine.sh
mrodm: You could create manually those PRs in the backport branches.

Or, I saw that in some other PRs it was used mergify to create those PRs: https://github.com/elastic/integrations/pull/13856#issuecomment-2869617823
I don't know if there would be conflicts if they are created by mergify.

.buildkite/scripts/test_one_package.sh
mrodm: At least there is code to prevent that in `test_one_package`.
For instance here:
https://github.com/elastic/integrations/blob/6c676bcfa7a0425e11b93662388d15ccc5f0fab5/.buildkite/scripts/test_one_package.sh#L36

And also in the `process_package` function:
https://github.com/elastic/integrations/blob/6c676bcfa7a0425e11b93662388d15ccc5f0fab5/.buildkite/scripts/common.sh#L1025-L1082

mrodm: Ok, I see it now
You see that other path, because the first step it runs this script is changing the path via `pushd` and then return to the previous folder via `popd`.

This is the code in the current main:
```shell
pushd packages > /dev/null
exit_code=0
if ! process_package "${package}" ; then
    # keep this message as a collapsed group in Buildkite, so it
    # is not hidden by the previous collapsed group.
    echo "--- [${package}] failed"
    exit_code=1
fi
popd > /dev/null
```

In order to ensure that the path is the root of the repository, I think it would be needed a small refactor @maximpn : 


```shell
pushd packages > /dev/null
exit_code=0
if ! process_package "${package}" ; then
    # keep this message as a collapsed group in Buildkite, so it
    # is not hidden by the previous collapsed group.
    echo "--- [${package}] failed"
    exit_code=1
fi
popd > /dev/null

if [ "${exit_code}" -ne 0 ]] ; then
  exit "${exit_code}"
fi

if [ -x "$custom_package_checker_script_path" ]; then
  echo "--- [${package}] Run individual package checker"
  "$custom_package_checker_script_path"
else
  echo "--- [${package}] Individual package checker $custom_package_checker_script_path is not found, continue..."
fi
```

.buildkite/scripts/test_one_package.sh
mrodm: So, could this be changed to be just this else-if clause ?

```shell
elif [ -x "$custom_package_checker_script_path" ]; then
  echo "--- [${package}] Run individual package checker"
  "$custom_package_checker_script_path"
fi
```

.buildkite/scripts/common.sh
```diff
@@ -763,10 +763,19 @@ is_pr_affected() {
     # Example:
     # https://buildkite.com/elastic/integrations/builds/25606
     # https://github.com/elastic/integrations/pull/13810
-    if git diff --name-only "${commit_merge}" "${to}" | grep -E -v '^(packages/|\.github/(CODEOWNERS|ISSUE_TEMPLATE|PULL_REQUEST_TEMPLATE|workflows/)|CODE_OF_CONDUCT\.md|README\.md|docs/|catalog-info\.yaml|\.buildkite/(pull-requests\.json|pipeline\.schedule-daily\.yml|pipeline\.schedule-weekly\.yml|pipeline\.backport\.yml))' > /dev/null; then
+    if git diff --name-only "${commit_merge}" "${to}" | grep -E -v '^(packages/|\.github/(CODEOWNERS|ISSUE_TEMPLATE|PULL_REQUEST_TEMPLATE|workflows/)|CODE_OF_CONDUCT\.md|README\.md|docs/|catalog-info\.yaml|\.buildkite/(pull-requests\.json|pipeline\.schedule-daily\.yml|pipeline\.schedule-weekly\.yml|pipeline\.backport\.yml|scripts/packages.+\.sh))' > /dev/null; then
```
mrodm: Should it contain another slash here in `scripts/packages/`? 

```suggestion
    if git diff --name-only "${commit_merge}" "${to}" | grep -E -v '^(packages/|\.github/(CODEOWNERS|ISSUE_TEMPLATE|PULL_REQUEST_TEMPLATE|workflows/)|CODE_OF_CONDUCT\.md|README\.md|docs/|catalog-info\.yaml|\.buildkite/(pull-requests\.json|pipeline\.schedule-daily\.yml|pipeline\.schedule-weekly\.yml|pipeline\.backport\.yml|scripts/packages/.+\.sh))' > /dev/null; then
```

.buildkite/scripts/test_one_package.sh
mrodm: As mentioned in the other comment, this could be refactored as to ensure that it is running from the root of the repository those commands:

```shell
pushd packages > /dev/null
exit_code=0
if ! process_package "${package}" ; then
    # keep this message as a collapsed group in Buildkite, so it
    # is not hidden by the previous collapsed group.
    echo "--- [${package}] failed"
    exit_code=1
fi
popd > /dev/null

if [ "${exit_code}" -ne 0 ]] ; then
  exit "${exit_code}"
fi

if [ -x "$custom_package_checker_script_path" ]; then
  echo "--- [${package}] Run individual package checker"
  "$custom_package_checker_script_path"
else
  echo "--- [${package}] Individual package checker $custom_package_checker_script_path is not found, continue..."
fi
```

mrodm:
> > About the improvements mentioned in the description:
> > > We may speed up the build by using an elastic-package Docker container published to docker.elastic.co. elastic-package installation is a complex process requiring a chain on installations GVM -> Go -> elastic-package. And it takes in average 3 minutes per each integration (integrations build in parallel). On top of that [Prebuilt Rules OOM testing Buildkite pipeline](https://buildkite.com/elastic/appex-qa-stateful-security-prebuilt-rules-ftr-oom-testing) has to install elastic-package as well. It sums up to 6 minutes which could be reduced.
> > 
> > 
> > I'm not totally sure this could lead to unexpected failures, since elastic-package already is based on docker to run some commands (e.g. elastic-package test system or even elastic-package stack up). Creating a docker image for elastic-package would involve to ensure that everything works in a docker-in-a-docker scenario. I think I would try to avoid that, WDYT @elastic/ecosystem ?
> 
> I just wanted to clarify my suggestion.
> 
> AFAIK Buildkite jobs run in a Kibernetes environment. For example **elastic/integrations** pipelines already [use](https://github.com/elastic/integrations/blob/main/.buildkite/pipeline.yml#L13-L14) Go and Ubuntu images.
> 
> Consequently it opens up a possibility to build a custom image with **elastic-package** from a standard Go image and use it for Buildklite jobs/steps requiring **elastic-package**.


Ah! Now I know what you meant @maximpn 

Not all steps in CI run in a kubernetes environment. There are steps that run on VMs, for instance all the steps testing the packages.

Not sure about moving `elastic-package` to the base image itself, since old images I think are deleted after some time. And for backport branches it is required to be able to run old versions of `elastic-package` too, that IIUC there could be missing when they are needed. And I don't know if there could be other issues, for instance how easy would be reverting a elastic-package release in integrations.

But I see that there could be other advantages of using a custom base image for those steps, having some software installed.

maximpn:
@mrodm,

I've applied your [suggestion](https://github.com/elastic/integrations/pull/15829#discussion_r2545579926) and got rid of `elif` section to make the implementation cleaner. Could you have a look?

.buildkite/scripts/test_one_package.sh
```diff
@@ -35,4 +35,13 @@ if ! process_package "${package}" ; then
 fi
 popd > /dev/null
 
-exit "${exit_code}"
+if [ "${exit_code}" -ne 0 ]] ; then
```
mrodm: My bad there was a typo in my suggestion, sorry

```suggestion
if [ "${exit_code}" -ne 0 ] ; then
```

Example of the error raised: https://buildkite.com/elastic/integrations/builds/34426#019aa193-ce73-491c-b08d-f4836e450610/803-851

jsoriano:
> > We may speed up the build by using an elastic-package Docker container published to docker.elastic.co. elastic-package installation is a complex process requiring a chain on installations GVM -> Go -> elastic-package. And it takes in average 3 minutes per each integration (integrations build in parallel). On top of that [Prebuilt Rules OOM testing Buildkite pipeline](https://buildkite.com/elastic/appex-qa-stateful-security-prebuilt-rules-ftr-oom-testing) has to install elastic-package as well. It sums up to 6 minutes which could be reduced.
> 
> I'm not totally sure this could lead to unexpected failures, since `elastic-package` already is based on docker to run some commands (e.g. `elastic-package test system` or even `elastic-package stack up`). Creating a docker image for `elastic-package` would involve to ensure that everything works in a docker-in-a-docker scenario. I think I would try to avoid that, WDYT @elastic/ecosystem ?

I agree that could be interesting to remove the build of elastic-package, but maybe the easiest path is to download the binaries [from the release page](github.com/elastic/elastic-package/releases/).

I would keep the code to build from source too because this is useful to test elastic-package and package-spec branches, but by default I agree that it could be interesting to use the pre-built binaries.

> FYI @elastic/ecosystem this PR introduces a way to add custom steps or tests to each package in CI, that could be used in other packages too.

Nice, this is great.

Issue #15671, opened 2025-10-17
[Multiple Integrations]: System test failures due to missing required variables in data stream configs


andrewkroh:
These data streams fail system tests due to missing required variables in their configurations, as enforced by elastic/elastic-package#2980. The systems need updated to declare values for the variables.

- **abnormal_security.ai_security_mailbox_not_analyzed**
  - Package policy is invalid: inputs.cel.streams.abnormal_security.ai_security_mailbox_not_analyzed.vars.wait_interval: Recent Message Grace Interval is required

- **aws_bedrock.invocation**
  - Package policy is invalid: inputs.aws-s3.streams.aws_bedrock.invocation.vars.file_selectors: Invalid format for File Selectors: expected array

- **crowdstrike.fdr**
  - Package policy is invalid: inputs.aws-s3.streams.crowdstrike.fdr.vars.queue_url: Queue URL is required

- **nvidia_gpu.stats**
  - Package policy is invalid: inputs.prometheus/metrics.streams.nvidia_gpu.stats.vars.ssl: Invalid format for SSL Configuration: expected array

The elastic-package change ensures system tests are strict about required variables. By making missing variable errors explicit, contributors and maintainers can more easily identify misconfigured tests and ensure coverage of required configuration options.

### Related

- https://github.com/elastic/integrations/pull/15618
- https://github.com/elastic/elastic-package/pull/2980

PR #15832, opened 2025-10-31
abnormal_security,aws_bedrock,crowdstrike: add required variables in system tests

moxarth-rathod:
## Proposed commit message

```
Fix system tests for abnormal_security, aws_bedrock, crowdstrike integrations
```

## Checklist

- [s] I have added an entry to my package's `changelog.yml` file.

- Clone integrations repo.
- Move to integrations/packages/abnormal_security or integrations/packages/aws_bedrock or 
  integrations/packages/crowdstrike directory.

> elastic-package test -v

## Related issues

- Part Of #15671


andrewkroh (commented):
Why are there three different YAML approaches to defining the required variables? This change has

```yaml
wait_interval: ""
file_selectors: |
queue_url:
ssl: |
```

which equates to either empty string or `null`

```json
{
  "wait_interval": "",
  "file_selectors": "",
  "queue_url": null,
  "ssl": ""
}
```

Could we use the same approach in all cases? (To be clear, I'm not thrilled that we can bypass the required variable check by passing empty strings or nulls, but that's a different problem.)

moxarth-rathod:
> Why are there three different YAML approaches to defining the required variables? This change has
> 
> ```yaml
> wait_interval: ""
> file_selectors: |
> queue_url:
> ssl: |
> ```
> 
> which equates to either empty string or `null`
> 
> ```json
> {
>   "wait_interval": "",
>   "file_selectors": "",
>   "queue_url": null,
>   "ssl": ""
> }
> ```
> 
> Could we use the same approach in all cases?
> 
> (To be clear, I'm not thrilled that we can bypass the required variable check by passing empty strings or nulls, but that's a different problem.)

I initially tried using null everywhere, but it didn’t resolve issue in system test. So, I went with a different approach. For consistency, I’ll change it to use an empty string everywhere.

packages/abnormal_security/data_stream/ai_security_mailbox_not_analyzed/_dev/test/system/test-default-config.yml
efd6, line 1: Looking at the change that added this configuration value (#14235), I think that it would be entirely reasonable for the config to be made non-required. This is based on the agent config code that was added and the description in the commit.

packages/aws_bedrock/data_stream/invocation/_dev/test/system/test-default-config.yml
```diff
@@ -12,5 +12,6 @@ data_stream:
     queue_url: '{{TF_OUTPUT_queue_url}}'
     preserve_original_event: true
     preserve_duplicate_custom_fields: true
+    file_selectors: ""
```
efd6: The complaint from ep is that this should be an array. Should this not be `null`?

moxarth-rathod: It throws the same error when I use null. The other option was to use an empty array, like file_selectors: |
So, I’ve changed it to an empty string for consistency.

efd6: Why not use the empty array? `file_selectors: []` (`file_selectors: |` is an empty string, depending on the lines that follow).

It seems to me that the validation is missing something if it allows a string in an array slot.

moxarth-rathod: @efd6 I've made the changes as per your suggestion.

packages/nvidia_gpu/data_stream/stats/_dev/test/system/test-default-config.yml
```diff
@@ -4,5 +4,6 @@ data_stream:
     preserve_original_event: true
     hosts: 
       - http://{{Hostname}}:{{Port}}/metrics
+    ssl: []
```
andrewkroh: This satisfies the required var check, but I think there is a bug here because the `ssl` configuration should not have `multi: true`. Also the handlebar template is wonky because the `if` check is looking at `ssl.certificate_authorities`.

https://github.com/elastic/integrations/blob/a5139fed4b1dc79b151d64541859a3624527919d/packages/nvidia_gpu/data_stream/stats/agent/stream/stream.yml.hbs#L18-L20

I think the package needs a policy test to check the handlebar rendering. If my assumption is correct, then `multi: true` should be removed, and then an empty string can satisfy the required check.

It may be best to merge this as is, and open a separate ticket for the nvidia_gpu package owner to sort this out, because I'm not sure of their original intent with the certificate_authorities if check.

kcreddy: Public issue: https://github.com/elastic/integrations/issues/15928

@moxarth-rathod, please change the team label in the issue as per [`CODEOWNERS`](https://github.com/elastic/integrations/blob/main/.github/CODEOWNERS#L347)

moxarth-rathod: @kcreddy done, thanks.

PR #15836, opened 2025-10-31
[juniper_srx] Update Juniper SRX documentation

mjwolf:
## Proposed commit message

Update Juniper SRX documentation, with more details on use cases, compatibility, set up instructions, and troubleshooting steps.

This also adds a knowledge_base file to the package, which can be used by other tools to which use information about this integration.

## Checklist

## Screenshots
Update documentation as rendered in Kibana:
juniper_srx_docs

Closes: https://github.com/elastic/integration-experience/issues/412

### DIFF
--- a/packages/juniper_srx/_dev/build/docs/README.md
+++ b/packages/juniper_srx/_dev/build/docs/README.md
@@ -33,6 +33,35 @@ The Juniper SRX integration collects the following log types from SRX devices:
 *   **RT_SECINTEL**: Security intelligence events tracking connections to known malicious sources.
 *   **System logs**: General system-level events from the SRX device.
 
+The following processes and tags are supported:
+
+| JunOS processes | JunOS tags                                |
+|-----------------|-------------------------------------------|
+| RT_FLOW         | RT_FLOW_SESSION_CREATE                    |
+|                 | RT_FLOW_SESSION_CLOSE                     |
+|                 | RT_FLOW_SESSION_DENY                      |
+|                 | APPTRACK_SESSION_CREATE                   |
+|                 | APPTRACK_SESSION_CLOSE                    |
+|                 | APPTRACK_SESSION_VOL_UPDATE               |
+| RT_IDS          | RT_SCREEN_TCP                             |
+|                 | RT_SCREEN_UDP                             |
+|                 | RT_SCREEN_ICMP                            |
+|                 | RT_SCREEN_IP                              |
+|                 | RT_SCREEN_TCP_DST_IP                      |
+|                 | RT_SCREEN_TCP_SRC_IP                      |
+| RT_UTM          | WEBFILTER_URL_PERMITTED                   |
+|                 | WEBFILTER_URL_BLOCKED                     |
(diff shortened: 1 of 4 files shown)


packages/juniper_srx/docs/README.md
```diff
@@ -1,43 +1,203 @@
...
+
+**Issue: No logs are appearing in Elasticsearch.**
+*   **Verify Connectivity**: Ensure the SRX device can reach the Elastic Agent host and port. Use `ping` or `telnet <ELASTIC_AGENT_IP> <PORT>` from the SRX CLI.
+*   **Check Firewalls**: Confirm that any firewalls between the SRX device and the agent host allow traffic on the configured syslog port.
+*   **Confirm SRX Configuration**: Use `show configuration system syslog` on the SRX device to verify the settings are correct.
+*   **Check Agent Logs**: Review the Elastic Agent logs for any connection errors or parsing issues.
+
+**Issue: Events are not parsed correctly.**
+*   **Verify Syslog Format**: Ensure the log format on the SRX device is set to `structured-data + brief`. Custom log formats may not parse correctly.
+*   **Check Junos OS Version**: Confirm you are running Junos OS 19.x or later.
```
qcorporation, line 225: @mjwolf how do we know that it's been tested for 19.x or later?
From https://www.juniper.net/documentation/product/us/en/junos-os/#cat=release_notes JunOS is at 25.2 - is that what we want to say? Have we actually tested it against 19.x version or later?

packages/juniper_srx/_dev/build/docs/README.md
```diff
@@ -1,42 +1,215 @@
...
+
+### How it works
+
+The integration collects logs sent from Juniper SRX devices via the syslog protocol. Elastic Agent listens for these syslog messages on a configured TCP or UDP port, or reads them from a log file. The agent processes and parses the incoming logs, enriching them with additional metadata before sending them to Elasticsearch for indexing and analysis.
+
+## What data does this integration collect?
+
+The Juniper SRX integration collects the following log types from SRX devices:
+
+*   **RT_FLOW**: Session creation, closure, and denial events, including NAT translations.
```
qcorporation, line 28: do you think it's valuable to list the subsections of the different message types. The old version of the documentation had a full table, i.e. RT_FLOW -> RT_FLOW_SESSION_CREATE, RT_FLOW_SESSION_CLOSE. Do you think we are loosing information here that the user might want

mjwolf: I've added back the table

packages/juniper_srx/docs/README.md
mjwolf: I couldn't find any information on when this was tested, or what it supports. I've taken out the exact supported version for Juniper.
The only information I could find is that the structured data form was added in 8.x, which is very old.

Oddly:
Can we still add this? Would be nice to have this information regardless of whether it's fully validated or not.

cpascale43:
Hey @mjwolf per the feedback above - can we consider reopening this?

PR #15840, opened 2025-11-02
qualys_vmdr: simplify and improve robustness of host queries in asset_host_detection

### LINKED ISSUE #15741: qualys_vmdr: improve handling of long ID lists in the asset_host_detection data stream
There is a [TODO in the code](https://github.com/elastic/integrations/blob/e65e5d717463a9b13a42a77b8fe23af642bdb92d/packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs#L73-L76) to improve the handling of long lists of IDs.

The `kibana.version` for the package (proxy for agent version) is [now v8.19.0||v9.1.0](https://github.com/elastic/integrations/blob/e65e5d717463a9b13a42a77b8fe23af642bdb92d/packages/qualys_vmdr/manifest.yml#L12-L13); v8.19.0||v9.1.0 use v1.22.0 of the mito/lib, which has the mentioned [diadic `tail` function](https://github.com/elastic/mito/commit/7b80f66411222b948f53e2520cc52589020e1802), so we can make use of `tail` in the CEL program to more carefully define work lists of IDs.

efd6:
## Proposed commit message
```
qualys_vmdr: simplify and improve robustness of host queries in asset_host_detection

Replace odd/even approach with queue consumption.

Limit is parameterised, but with a constant value; this should not be user
configurable, but allow it to be changed easily.

Also fix rendering of the CEL program in the fleet configuration by removing
trailing whitespace on source lines.
```

> [!NOTE]
> The API also allows a POST request mechanism which does not have the limits, but it is poorly documented and would require constructing an XML body. This could be done purely stringly, but given the absence of good docs and the poor ergonomics of using XML for … anything, I've retained the GET request.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #15741

### DIFF
--- a/packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
+++ b/packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
@@ -177,7 +177,7 @@ program: |
       )
     ).as(state, state.with(
       !has(state.worklist) ? state :
-        has(state.worklist.HOST_QID_LIST) && size(state.worklist.HOST_QID_LIST[0]) > 0 ?
+        has(state.worklist.HOST_QID_LIST) && size(state.worklist.HOST_QID_LIST) > 0 ?
           request("GET", state.url.trim_right("/") + "/api/3.0/fo/knowledge_base/vuln/?" +
             {
               "ids": [front(state.worklist.HOST_QID_LIST, int(state.query_limit)).join(",")],


packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs
```diff
@@ -190,11 +176,11 @@ program: |
         ))
       )
     ).as(state, state.with(
-      !has(state.worklist) ? state : 
+      !has(state.worklist) ? state :
         has(state.worklist.HOST_QID_LIST) && size(state.worklist.HOST_QID_LIST[0]) > 0 ?
```
kcreddy:
```suggestion
        has(state.worklist.HOST_QID_LIST) && size(state.worklist.HOST_QID_LIST) > 0 ?
```

efd6: Good catch.

PR #15849, opened 2025-11-03
qualys_gav: revise date processing for asset inventory list to use foreach processor

clement-fouque:
## Proposed commit message

The qualys_gav.asset.inventory_list_data.inventory field obtaind from the API comes as an array, so previously the ingest pipeline would fail with an 'Illegal list shortcut value' message.

* Replaced individual date processors for created and last_updated fields with foreach processors that iterate over each item in the qualys_gav.asset.inventory_list_data.inventory array, applying the date processor to each element. This change ensures correct processing when the inventory contains multiple items.
* Updated error handling within the new foreach processors to remove only the problematic field if date parsing fails, and set ignore_missing to true for robustness.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/qualys_gav/data_stream/asset/sample_event.json
+++ b/packages/qualys_gav/data_stream/asset/sample_event.json
@@ -1,11 +1,11 @@
 {
-    "@timestamp": "2025-11-04T13:41:51.662Z",
+    "@timestamp": "2025-11-05T07:38:28.689Z",
     "agent": {
-        "ephemeral_id": "6b78f4f1-2fe7-41bc-ba21-b2a95737736d",
-        "id": "cd4b416e-a47a-42d6-865d-a01b7fe3d45d",
-        "name": "elastic-agent-63287",
+        "ephemeral_id": "4068ddb1-7171-462f-b59a-1d0918801b8a",
+        "id": "a8e4fdfe-886d-4528-aa7d-b3977f1b26fd",
+        "name": "elastic-agent-30002",
         "type": "filebeat",
-        "version": "9.2.0"
+        "version": "8.19.3"
     },
     "cloud": {
         "account": {
@@ -27,7 +27,7 @@
     },
     "data_stream": {
         "dataset": "qualys_gav.asset",
-        "namespace": "22762",
+        "namespace": "96771",
         "type": "logs"
     },
     "device": {
@@ -40,9 +40,9 @@
         "version": "8.17.0"
     },
     "elastic_agent": {
-        "id": "cd4b416e-a47a-42d6-865d-a01b7fe3d45d",
+        "id": "a8e4fdfe-886d-4528-aa7d-b3977f1b26fd",
         "snapshot": false,
-        "version": "9.2.0"
+        "version": "8.19.3"
     },
     "event": {
         "agent_id_status": "verified",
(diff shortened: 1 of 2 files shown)


packages/qualys_gav/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.4.1"
+  changes:
+    - description: Refactor date processing for asset inventory list to use foreach processor.
+      type: bugfix
```
efd6, line 5: What bug does this fix?

efd6 (commented):
Please add a proposed commit message to the PR description and add a test case that exercises the change here.

packages/qualys_gav/changelog.yml
clement-fouque: While ingesting, I had the following error: 
```
Processor 'conditional'
with tag 'date_qualys_gav_asset_inventory_list_data_inventory_created'
failed with message 'Illegal list shortcut value [created].'
```
After analysis it's due to the fact that field `qualys_gav.asset.inventory_list_data.inventory` is an array and we should process it as an array.

clement-fouque: FYI, this is not linked to my initial use case but the integration wasn't able to pass tests so I fixed it.

clement-fouque:
It's ready for review. 

cc @efd6

packages/qualys_gav/data_stream/asset/sample_event.json
```diff
@@ -268,6 +268,13 @@
                 "last_updated": "2025-07-11T14:21:10.000Z",
                 "source": "EASM"
             },
+            "inventory_list_data": {
+                "inventory": {
```
efd6: Why is this an object?

efd6 (commented):
Thanks.

Suggest the following for the commit message (commit messages should not have markdown and should follow [these guidelines](https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#commit-messages).

commit message title (already applied; the action here is a change rather than a refactor since refactors do not change behaviour):
```
qualys_gav: revise date processing for asset inventory list to use foreach processor
```
commit message body:
```
The qualys_gav.asset.inventory_list_data.inventory field obtaind from
the API comes as an array, so previously the ingest pipeline would fail
with an 'Illegal list shortcut value' message.

* Replaced individual date processors for created and last_updated
  fields with foreach processors that iterate over each item in the
  qualys_gav.asset.inventory_list_data.inventory array, applying the
  date processor to each element. This change ensures correct
  processing when the inventory contains multiple items.
* Updated error handling within the new foreach processors to remove
  only the problematic field if date parsing fails, and set
  ignore_missing to true for robustness.
```

packages/qualys_gav/data_stream/asset/sample_event.json
clement-fouque: This file is generated automatically from command `elastic-package test system --generate`. I agree that it's not the right format but I don't know why it is set like this. 

I did some checks of adding another element (to have 2 elements in the array and not 1) but I have a similar behaviour.

I checked other elements that have the same issue. For example tag_list has the same behaviour: 
https://github.com/elastic/integrations/blob/a8fcb8e64ac9d70dc4bf7c14fe4d549912177a86/packages/qualys_gav/data_stream/asset/sample_event.json#L451-L460

But `user_account_list_data` doesn't behave the same: 
https://github.com/elastic/integrations/blob/a8fcb8e64ac9d70dc4bf7c14fe4d549912177a86/packages/qualys_gav/data_stream/asset/sample_event.json#L463-L475

It seems related to the presence of the foreach processor since we have one for tag_list: 
https://github.com/elastic/integrations/blob/a8fcb8e64ac9d70dc4bf7c14fe4d549912177a86/packages/qualys_gav/data_stream/asset/elasticsearch/ingest_pipeline/default.yml#L1746-L1762
But we don't have it for user_account_data_list. 

To sum up, yes there is an issue but it's not related nor introduced by this PR.

clement-fouque:
@efd6 I believe I've implemented the changes you requested. Please let me know what you want to do on the `sample_event.json` issue.

packages/qualys_gav/data_stream/asset/sample_event.json
efd6: The data that the mock sends has this as an object. The following changes would fix that
```diff
diff --git a/packages/qualys_gav/_dev/deploy/docker/files/config.yml b/packages/qualys_gav/_dev/deploy/docker/files/config.yml
index f4fc20ff7c..1c0c1ba742 100644
--- a/packages/qualys_gav/_dev/deploy/docker/files/config.yml
+++ b/packages/qualys_gav/_dev/deploy/docker/files/config.yml
@@ -321,11 +321,13 @@ rules:
                               "noOfImages": 3,
                               "hasSensor": "temp_value"
                           },
-                          "inventory": {
-                              "source": "EASM",
-                              "created": 1752070872000,
-                              "lastUpdated": 1752243670000
-                          },
+                          "inventory": [
+                              {
+                                  "source": "EASM",
+                                  "created": 1752070872000,
+                                  "lastUpdated": 1752243670000
+                              }
+                          ],
                           "activity": {
                               "source": "EASM",
                               "lastScannedDate": 1752243670000
@@ -736,11 +738,13 @@ rules:
                               "noOfImages": 5,
                               "hasSensor": "temp_values"
                           },
-                          "inventory": {
-                              "source": "EASM",
-                              "created": 1752070874000,
-                              "lastUpdated": 1752416469000
-                          },
+                          "inventory": [
+                              {
+                                  "source": "EASM",
+                                  "created": 1752070874000,
+                                  "lastUpdated": 1752416469000
+                              }
+                          ],
                           "activity": {
                               "source": "EASM",
                               "lastScannedDate": 1752416469000
@@ -1145,11 +1149,13 @@ rules:
                               "noOfImages": 0,
                               "hasSensor": "temp_value"
                           },
-                          "inventory": {
-                              "source": "QAGENT",
-                              "created": 1751889561000,
-                              "lastUpdated": 1752520814000
-                          },
+                          "inventory": [
+                              {
+                                  "source": "QAGENT",
+                                  "created": 1751889561000,
+                                  "lastUpdated": 1752520814000
+                              }
+                          ],
                           "activity": {
                               "source": "QAGENT",
                               "lastScannedDate": 1752520814000
@@ -1531,11 +1537,13 @@ rules:
                               "noOfImages": 0,
                               "hasSensor": "temp_value"
                           },
-                          "inventory": {
-                              "source": "QAGENT",
-                              "created": 1751889561000,
-                              "lastUpdated": 1752520814000
-                          },
+                          "inventory": [
+                              {
+                                  "source": "QAGENT",
+                                  "created": 1751889561000,
+                                  "lastUpdated": 1752520814000
+                              }
+                          ],
                           "activity": {
                               "source": "QAGENT",
                               "lastScannedDate": 1752520814000
@@ -1938,11 +1946,13 @@ rules:
                               "noOfImages": 5,
                               "hasSensor": "temp_values"
                           },
-                          "inventory": {                                                                                                                                                 
-                              "source": "EASM",                                                                                                                                          
-                              "created": 1752070874000,                                                                                                                                  
-                              "lastUpdated": 1752416469000                                                                                                                               
-                          },                                                                                                                                                             
+                          "inventory": [                                                                                                                                                 
+                              {                                                                                                                                                          
+                                  "source": "EASM",                                                                                                                                      
+                                  "created": 1752070874000,                                                                                                                              
+                                  "lastUpdated": 1752416469000                                                                                                                           
+                              }                                                                                                                                                          
+                          ],                                                                                                                                                             
                           "activity": {                                                                                                                                                  
                               "source": "EASM",                                                                                                                                          
                               "lastScannedDate": 1752416469000                                                                                                                           
```

clement-fouque: There are 2 differents object: `inventory` and `inventory_list_data`. The two objects are correctly formatted and the issue persists on `inventory_list_data` only. 
https://github.com/elastic/integrations/blob/901e063f8ebdbacb89490ea0a0d06144c4166fd4/packages/qualys_gav/_dev/deploy/docker/files/config.yml#L324-L337

clement-fouque: Here is an example of raw data from Qualys API:
image

efd6: I just ran `elastic-package test system -g` and I get what I would expect. I'll push that.

efd6 (approved):

PR #15865, opened 2025-11-04
[ER] Enhancement for Cisco IOS XR Integration

qcorporation:
## Proposed commit message
Because the grok_header was too finicky decided to move parsing down to grok_message_at_top_level

- able to isolate message and a dedicated grok pattern for pulling out hostname
- generated pipeline tests for the new message formats
- handled the space/no space separation for the dissect_login scenario
- fixed missing ecs fields, `related.hosts` and `source.domain`

## Checklist

## Related issues
- Closes [#24697](https://github.com/elastic/enhancements/issues/24697)


packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -77,8 +77,11 @@ processors:
       tag: grok_message_at_top_level
       patterns:
         - '^%%{GREEDYDATA:message}$'
+        - '^(%{IP}|%{CISCO_HOSTNAME:log.syslog.hostname}) %{GREEDYDATA}%%{GREEDYDATA:message}$' # message could still have hostname, continue to parse if available
```
qcorporation: I was having problems modifying `grok_header`. It's super flaky and prone to error.
```
'^%{CISCO_PRIORITY_MSGCOUNT}?(?:(?:%{IP}|%{CISCO_HOSTNAME:log.syslog.hostname})(?:: \*%{DATA}:|:?)? )?(?:%{NUMBER:cisco.ios.sequence}: )?(?:%{CISCO_UPTIME:cisco.ios.uptime}|%{CISCO_TIMESTAMP:_temp_.timestamp}): %{GREEDYDATA:_temp_.message}$'
```

Decided to move the logic one step down, but that does mean I have to duplicate the `pattern_definitions` unfortunately

packages/cisco_ios/data_stream/log/_dev/test/pipeline/test-syslog.log-expected.json
```diff
@@ -1017,6 +1017,190 @@
...
+            "tags": [
+                "preserve_original_event"
+            ]
+        },
+        {
+            "cisco": {
+                "ios": {
+                    "facility": "SECURITY-SSHD",
+                    "message_count": 2300,
+                    "uptime": ""
```
taylor-swanson: Empty string here is odd. Could be legitimate, but remains because this pipeline doesn't have script to remove null/empty fields.

packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: If I recall correctly, that grok handles about 20+ different header variations. Super flaky is an understatement.

I also learned a while ago that grok patterns are NOT run in sequential order, rather the order is chosen by "best match" (whatever that means), and GREEDYDATA really screws with that. There's really no way to know what order these will be applied.

I'm concerned about the empty uptime value, but I don't know enough of what patterns are being used to say what is really happening here.

I've long groaned about how we need to rethink header parsing here, but I don't have any good ideas. Last year, I was able to reduce grok_header down to 3 patterns... well we're back up to 7 and counting.

packages/cisco_ios/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.32.2"
```
taylor-swanson:
```suggestion
- version: "1.33.0"
```

packages/cisco_ios/changelog.yml
```diff
@@ -1,4 +1,9 @@
...
+  changes:
+    - description: Fix parsing hostname for 4 different log formats.
+      type: bugfix
```
taylor-swanson:
```suggestion
      type: enhancement
```

packages/cisco_ios/data_stream/log/_dev/test/pipeline/test-syslog.log-expected.json
qcorporation: @taylor-swanson, i agree, let me look at the uptime value to see what's going on there...

qcorporation: From my understanding It looks like the grok_header
```
'^%{CISCO_PRIORITY_MSGCOUNT}?(?:(?:%{IP}|%{CISCO_HOSTNAME:log.syslog.hostname})(?:: \*%{DATA}:|:?)? )?(?:%{NUMBER:cisco.ios.sequence}: )?(?:%{CISCO_UPTIME:cisco.ios.uptime}|%{CISCO_TIMESTAMP:_temp_.timestamp}): %{GREEDYDATA:_temp_.message}$'
```

Is still getting triggered and hence the empty uptime field, even though there is no uptime within the log line.
If I move the grok pattern up, with some modifications, I don't see the uptime, I see other failures but the field doesn't exist when i generate the expected
I'll keep working on an more optimal solution

packages/cisco_ios/data_stream/log/elasticsearch/ingest_pipeline/default.yml
qcorporation: @taylor-swanson so I refactored to be at `grok_header` and there's no more uptime dangling value. Unfortunately this is another entry in `grok_header`

Marking as resolved

Issue #15703, opened 2025-10-21
[azure logs] make azure-eventhub processor v2 the default option in all logs integrations


zmoog:
## Context

The Azure Logs integration (and related logs integrations) has used azure-eventhub processor v1 for years. This processor is based on the legacy Event Hubs SDK.

We later introduced azure-eventhub processor v2 as an option, but kept processor v1 as the default.

Since we've decided to retire processor v1, we plan to set processor v2 as the default option in the next release, and completely remove v1 in the following release.

## Goal

Checklist

- [ ] Make azure-eventhub processor v2 the default processor in all Azure logs integrations.
- [ ] Policy template: add processor v2 option only when `processor_version` is set to `v2`.
- [ ] Replace `type: text` with `type: select` for `processor_version` and `processor_start_position` config options.

## Impact

We want to switch all users to processor v2 while we still have the option to revert back to processor v1, if needed.

kaiyan-sheng:
Reopening this issue since the [PR](https://github.com/elastic/beats/pull/47292) I merged is only for the Beats part of the change.

PR #15900, opened 2025-11-07
[azure logs] Add processor version options to Raw Events integration (v1)

zmoog:
## Proposed commit message

Add the processor version option to the Raw Events integration v1. With these new options, users can switch from processor v1 (current default) to the processor v2.

This change anticipate the switch to v2 as default processor, see https://github.com/elastic/integrations/issues/15703 for context and more details.

## Checklist

- Install the Raw Events integration (v1)
- Open "Advanced settings", and set Processor version to `v2`

## Screenshots

CleanShot 2025-11-07 at 12 54 05@2x

### DIFF
--- a/packages/azure_logs/changelog.yml
+++ b/packages/azure_logs/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.4.2"
+  changes:
+    - description: Switch `processor_version` and `processor_start_position` configuration options to `select` type (UI change).
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/15900
 - version: "0.4.1"
   changes:
     - description: Add support to override `Endpoint Suffix` based on deployment environment.
--- a/packages/azure_logs/manifest.yml
+++ b/packages/azure_logs/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.3.0
 name: azure_logs
 title: "Custom Azure Logs"
-version: "0.4.1"
+version: "0.4.2"
 source:
   license: Elastic-2.0
 description: "Collect log events from Azure Event Hubs with Elastic Agent"


packages/azure/data_stream/activitylogs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -24,9 +30,23 @@ processors:
       ignore_missing: true
       if: 'ctx.event?.original != null'
       description: 'The `message` field is no longer required if the document has an `event.original` field.'
+
+  # Encoding extension  
+  - rename:
+      field: body.structured
```
muthu-mps: Why does this field renamed only for activity_logs? Should the `body.structured` needs to be renamed in other data streams like `sign-in logs, graph_activity logs` as well?

zmoog: My bad. I included a couple of things that don't belong here 🤦 I'll clean up the commits.

zmoog: @muthu-mps, PR content fixed!

packages/azure/data_stream/eventhub/manifest.yml
```diff
@@ -74,6 +74,94 @@ streams:
...
+      # Processor v2 only settings
+      #
+      - name: processor_version
+        type: text
+        title: Processor version
+        multi: false
+        required: false
+        show_user: false
+        default: v1
+        description: "The processor version that the integration should use. Possible values are `v1` and `v2` (preview). \nThe v2 event hub processor is in preview, so using the v1 processor is recommended for typical use cases.\nDefault is `v1`."
```
kaiyan-sheng, line 93:
```suggestion
      - name: processor_version
        type: text
        title: Processor version
        multi: false
        required: false
        show_user: false
        default: v1
        description: "The processor version that the integration should use. Possible values are `v1` and `v2` (preview). \nThe v2 event hub processor is in preview, so using the v1 processor is recommended for typical use cases.\nDefault is `v1`."
      #
      # Processor v2 only settings
      #
```
```suggestion
      #
      # Processor v2 only settings
      #
      - name: processor_version
        type: text
        title: Processor version
        multi: false
        required: false
        show_user: false
        default: v1
        description: "The processor version that the integration should use. Possible values are `v1` and `v2` (preview). \nThe v2 event hub processor is in preview, so using the v1 processor is recommended for typical use cases.\nDefault is `v1`."
```

kaiyan-sheng (commented):
It looks good to me! Lets add a changelog and it's good to go!

zmoog:
> It looks good to me! Lets add a changelog and it's good to go!

Adding changelog and running a final test before merging.

zmoog:
@elastic/security-service-integrations, could you take a look at this PR when you have time?

packages/azure/data_stream/eventhub/manifest.yml
```diff
@@ -74,6 +74,94 @@ streams:
...
+      #
+      # Processor v2 only settings
+      #
+      - name: processor_version
+        type: text
+        title: Processor version
+        multi: false
+        required: false
+        show_user: false
+        default: v1
```
efd6, line 87: Since this is an enumerated set (presumably with a low cardinality for the foreseeable future), would a `type: select` be appropriate rather than a `type: text`?

zmoog: Oh, this is new to me—my bad I suppose. Checking `type: select` right away! 

Update: Oh my god, some uses are 2-year old! 🤦 I totally missed this, shame on me.

zmoog: Added! Thanks for the suggestion.

CleanShot 2025-11-17 at 22 32 27@2x

efd6 (approved):
This is waiting for obs-ds-hosted-services, but LGTM.

packages/azure/data_stream/eventhub/agent/stream/stream.yml.hbs
```diff
@@ -19,10 +19,31 @@ storage_account: {{storage_account}}
...
+{{#if processor_version}}
+processor_version: {{processor_version}}
+{{/if}}
+{{#if migrate_checkpoint}}
+migrate_checkpoint: {{migrate_checkpoint}}
+{{/if}}
+{{#if processor_update_interval}}
+processor_update_interval: {{processor_update_interval}}
+{{/if}}
+{{#if processor_start_position}}
```
gizas, line 37: Is it possible to add sth like if processor_version}==v2 then processor_start_position? 

Same also for below fields that are only realted to v2

packages/azure/data_stream/eventhub/manifest.yml
```diff
@@ -74,6 +74,99 @@ streams:
...
+          (Processor v2 only) How often the processor should attempt to claim partitions.
+
+          Default is `10` seconds.
+      - name: processor_start_position
+        type: text
+        title: Processor start position
+        multi: false
+        required: false
+        show_user: false
+        default: earliest
```
gizas, line 111: should also be options with latest/earliest?

zmoog: Yep! Thanks, I missed this opportunity.

zmoog: Done

packages/azure/data_stream/eventhub/agent/stream/stream.yml.hbs
zmoog: I also added a couple of checklist items for this in the next issue https://github.com/elastic/integrations/issues/15703, so I can do the same across all the integrations using this input.

zmoog: It seems that the built-in helper `#if` only supports checking if a value is empty to skip rendering a block:

From https://handlebarsjs.com/guide/builtin-helpers.html#if:

> You can use the if helper to conditionally render a block. If its argument returns false, undefined, null, "", 0, or [], Handlebars will not render the block.

This problem should auto-resolve soon as we are switching to processor v2 as default and dropping v1 after that.

zmoog (commented):
Switched to `type: select` UI element in Azure Logs and Custom Azure Logs integrations.

The built-in helper `#if` doesn't seem to support expressions though.

PR #15911, opened 2025-11-10
[eset_protect] Fix 404 Not Found responses caused by a stale cached response-id header by correctly clearing cursor on API errors

brijesh-elastic:
## Proposed commit message
```
eset_protect: Fix 404 errors caused by stale Response-Id

The ESET API sometimes replies with 202 Accepted and a Response-Id header, 
which must be sent with the next poll request. Once the task completes and the API returns 
200 OK, that Response-Id becomes invalid.

Currently, the integration does not correctly update the cursor in the case of an error scenario.
Once the API returns an error, we attempt to clear the cursor, but it's not actually updating.
This is because whenever we return an error, cursor updating doesn't work. Therefore, we
have to wrap the error and publish it as a single element in an array
so that the cursor clears correctly.

NOTE: Since we're publishing the error in a single element array,
the health status update (from healthy to degraded)
won't work in the case of errors.

This patch clears the cached Response-Id correctly, ensuring the header is only 
included when valid.
```

## Checklist

- Clone integrations repo.

> elastic-package test -v

### DIFF
--- a/packages/eset_protect/changelog.yml
+++ b/packages/eset_protect/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "1.11.2"
   changes:
-    - description: "Fix 404 Not Found responses caused by a stale cached `response-id` header by correctly clearing `cursor` on API errors."
+    - description: "Fix `404 Not Found` responses caused by a stale cached `response-id` header by correctly clearing `cursor` on API errors."
       type: bugfix
       link: https://github.com/elastic/integrations/pull/15911
 - version: "1.11.1"


packages/eset_protect/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.11.2"
+  changes:
+    - description: "Fix 404 Not Found responses caused by a stale cached `response-id` header by correctly clearing `cursor` on API errors."
```
ShourieG:
```suggestion
    - description: "Fix `404 Not Found` responses caused by a stale cached `response-id` header by correctly clearing `cursor` on API errors."
```

ShourieG (approved):
NIT. Otherwise LGTM

PR #15917, opened 2025-11-10
Add Elastic Streams support to TCP and UDP integrations

jamesagarside:
This change adds the "Use the logs data stream" toggle to both the Custom TCP Logs and Custom UDP Logs integrations, enabling Elastic Streams functionality that was previously only available in the Custom Logs (Filestream) integration.

Changes:
- Added `use_logs_stream` boolean configuration variable to TCP and UDP manifests
- Updated TCP and UDP input templates to conditionally use `index: logs` when enabled
- Bumped TCP version from 2.0.1 to 2.1.0
- Bumped UDP version from 2.2.1 to 2.3.0
- Updated changelogs for both integrations

When enabled, this feature sends all ingested data to the "logs" data stream instead of the configured dataset. The "Write to logs streams" option must be enabled in the output settings for this to work.

## Checklist

### DIFF
--- a/packages/tcp/manifest.yml
+++ b/packages/tcp/manifest.yml
@@ -6,7 +6,7 @@ type: input
 version: "2.1.0"
 conditions:
   kibana:
-    version: "^8.13.0 || ^9.0.0"
+    version: "^9.2.0"
 categories:
   - custom
   - custom_logs
@@ -38,7 +38,7 @@ policy_templates:
         type: bool
         title: Use the "logs" data stream
         description: |
-          Enabling this will send all the ingested data to the "logs" data stream. This feature is disabled by default. If enabled the Dataset name option is ignored. "Write to logs streams" option must be enabled in the output settings for this to work.
+          Enabling this will send all the ingested data to the "logs" data stream. This feature requires Elasticsearch 9.2.0 or later and is disabled by default. If enabled the Dataset name option is ignored. "Write to logs streams" option must be enabled in the output settings for this to work.
         required: false
         show_user: true
         default: false
--- a/packages/udp/manifest.yml
+++ b/packages/udp/manifest.yml
@@ -6,7 +6,7 @@ type: input
 version: "2.3.0"
 conditions:
   kibana:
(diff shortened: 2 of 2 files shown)


jamesagarside:
@andrewkroh looking like we're G2G if I could get a review please.

packages/tcp/manifest.yml
```diff
@@ -3,7 +3,7 @@ name: tcp
 title: Custom TCP Logs
 description: Collect raw TCP data from listening TCP port with Elastic Agent.
 type: input
-version: "2.0.1"
+version: "2.1.0"
 conditions:
   kibana:
     version: "^8.13.0 || ^9.0.0"
```
taylor-swanson: Do we need to adjust the stack version at all? (same applies to UDP package).

For 9.x, probably want to raise to the minimum supported version for streams. For 8.x, I'm pretty sure streams is 9.x only? I dunno if we want to drop 8.x support (since a lot of users are still on it), but I also know we can't conditionally display configurations based on stack version. Perhaps a warning in the description saying the feature is only supported on 9.x?

jamesagarside: A very good point, will get on that

packages/tcp/manifest.yml
```diff
@@ -3,10 +3,10 @@ name: tcp
 title: Custom TCP Logs
 description: Collect raw TCP data from listening TCP port with Elastic Agent.
 type: input
-version: "2.0.1"
+version: "2.1.0"
 conditions:
   kibana:
-    version: "^8.13.0 || ^9.0.0"
+    version: "^9.2.0"
```
taylor-swanson, line 9: @qcorporation, what are your thoughts about dropping support for 8.x here?

Alternative is to keep it around, but then there is this new config option, which if enabled on 8.x will cause the input to break. Unfortunately, we can't conditionally show config options.

jamesagarside: I did this to match the Filestream integration which supports Streams https://github.com/elastic/integrations/blob/main/packages/filestream/manifest.yml

PR #15926, opened 2025-11-11
tenable_io: allow max_executions to be configured for vulnerability ds

### LINKED ISSUE #15899: [tenable_io]: Integration sometimes fails to collect vulnerability logs
### Integration Name

Tenable Vulnerability Management [tenable_io]

### Dataset Name

tenable_io.vulnerability

### Integration Version

4.4.1

### Agent Version

8.19.4

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.19.5

Ubuntu 24.04 LTS (x86_64)

### Software/API Version

_No response_

### Error Message

exceeding maximum number of CEL executions

### Event Original

_No response_

### What did you do?

We have the integration configured to collect asset, plugin and vulnerability logs. All settings are the default values apart from removing "info" from the severity levels to collect.

### What did you see?

The integration usually works, but sometimes it fails to ingest vulnerability logs and we see the error message mentioned above. I have checked the Tenable API and the export requests are there, but the integration doesn't collect the results. I think it is sending lots of requests to check whether the export is ready to be downloaded and sometimes it reaches the maximum number of retries before the export is ready. When it fails, it takes about 2-3 minutes to go from starting the request to getting the error message.

It should be possible to collect logs even if the export request takes a few minutes to be processed. I think it might help if there was an option to add a delay between retries so it doesn't reach the maximum number so quickly.

### Anything else?

_No response_

efd6:
## Proposed commit message

See title.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #15899

### DIFF
--- a/packages/tenable_io/data_stream/vulnerability/manifest.yml
+++ b/packages/tenable_io/data_stream/vulnerability/manifest.yml
@@ -66,7 +66,7 @@ streams:
         multi: false
         required: false
         show_user: false
-        default: 1000
+        default: 2000
       - name: tags
         type: text
         title: Tags


packages/tenable_io/data_stream/vulnerability/manifest.yml
```diff
@@ -59,6 +59,14 @@ streams:
...
         show_user: false
         default: 30s
+      - name: max_executions
+        type: integer
+        title: Maximum Pages Per Interval
+        description: Maximum Pages Per Interval is the maximum number of pages that can be collected at each interval.
+        multi: false
+        required: false
+        show_user: false
+        default: 1000
```
chrisberkhout: If this integration is known to hit the limit, maybe its default should double the input's default value?

efd6: I think this is up to the size of the data set, but I guess it won't hurt.

Issue #14135, opened 2025-06-03
Enhance CrowdStrike Integrations with Comprehensive Testing and Coverage Expansion


jamiehynds:
Current Situation:
We support multiple CrowdStrike data ingestion methods, including the Falcon SIEM Connector, Event Streaming API, Falcon Data Replicator, and REST API. Our integration covers various CrowdStrike modules such as endpoint, vulnerabilities, threat intelligence, host data, and more.

Challenge:
Until recently, testing our CrowdStrike integrations was limited by lack of access to the Falcon platform. Thanks to our new technology partnership with CrowdStrike, we now have direct access to a CrowdStrike instance.

**Goals:**

**Populate the CrowdStrike Instance:**

Fully populate our CrowdStrike instance to closely replicate a typical customer deployment.

Align the data in this instance with the types of data supported by our existing integrations.

**Continuous Testing and Validation:**

Use our integrations to ingest data from the populated CrowdStrike instance into an Elastic instance.

Perform ongoing testing across dashboards, pipelines, performance, and scalability.

Identify and resolve any bugs, performance bottlenecks, or data normalization issues to improve reliability and user experience.

**Expand Integration Coverage:**

Review and compare our current CrowdStrike integrations against all available CrowdStrike modules.

Prioritize building support for modules we do not currently cover, such as Identity Protection, to broaden our offering and meet evolving customer needs.

navnit-elastic:
Here's are some findings from the ongoing testing of CrowdStrike integration. This include findings  for alert, falcon, host and vulnerability data streams and not for FDR.

**Alert**

Findings:

- [x] Alert data stream making multiple requests with the same alert ids. The current implementation fetches a list of alert IDs and then fetches the full details for each alert. However the ids list never emptied, resulting in the same ids being passed in subsequent intervals. This behavior only happens when no new alerts are found in the first API call.
- [x] Instead of querying the "/alerts/queries/alerts/v2” endpoint to get a list of 1000 IDs, formatting them properly, and then sending them to "/alerts/entities/alerts/v2", you can just send a request to the "/alerts/combined/alerts/v1" endpoint and receive up to 1000 full records at a time (supports FQL and pagination).


**Falcon**

Supported event types:

- [x] DetectionSummaryEvent
- [x] EppDetectionSummaryEvent
- [x] AuthActivityAuditEvent
- [x] UserActivityAuditEvent
- [ ] APIActivityAuditEvent
- [ ] HashSpreadingEvent
[...]

PR #15927, opened 2025-11-11
crowdstrike: update the integration layout and documentation

### LINKED ISSUE #15972: CrowdStrike: Update the integration documentation
Update the CrowdStrike integration's README.md as per the Elastic [documentation guidelines](https://www.elastic.co/docs/extend/integrations/documentation-guidelines).

navnit-elastic:
## Proposed commit message

```
crowdstrike: update the integration layout and documentation

update the crowdstrike integration's README.md
as per the Elastic documentation guidelines[1].

update the policy templates and manifest documentation
to ensure consistency with the README. 

[1] https://www.elastic.co/docs/extend/integrations/documentation-guidelines
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Relates #14135 
- Closes #15972 

## Screenshots

**README:**

image

image
Screenshot 2026-02-26 at 5 09 37 PM
Screenshot 2026-02-26 at 5 10 14 PM
image
Screenshot 2026-02-26 at 5 10 45 PM
Screenshot 2026-02-26 at 5 10 59 PM

crowdstrike-elastic-data-flow drawio


kcreddy:
@navnit-elastic, please let us know when the PR is ready for review, so we can take a look.

@chrisberkhout, it will be nice to get your feedback on this once its ready.

packages/crowdstrike/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.8.1"
```
kcreddy:
```suggestion
- version: "2.9.0"
```
Lets go with minor increment as technically its still an enhancement.

packages/crowdstrike/data_stream/falcon/manifest.yml
```diff
@@ -80,7 +80,7 @@ streams:
       - name: app_id
         type: text
         title: App ID
-        description: App ID for the CrowdStrike.
+        description: App ID is an alphanumeric string to identify an event stream. App IDs can have a maximum of 32 characters. App IDs must be unique to each active event stream.
```
kcreddy: I don't think this updated description maybe appropriate here.
We ask users to fetch App ID from a Crowdstrike instance and users may not care about the specification of App ID. The specification should be part of Crowdstrike documentation, which you can link here if you need.

packages/crowdstrike/img/crowdstrike-flowchart.png
kcreddy, line 1: fdr data stream also has file based ingestion.

packages/crowdstrike/_dev/build/docs/README.md
```diff
@@ -1,107 +1,79 @@
...
+## Overview
 
-1. **Falcon SIEM Connector**: This is a pre-built integration designed to connect CrowdStrike Falcon with Security Information and Event Management (SIEM) systems. It streamlines the flow of security data from CrowdStrike Falcon to the SIEM, providing a standardized and structured way of feeding information into the SIEM platform. It includes the following datasets for receiving logs:
+The [CrowdStrike](https://www.crowdstrike.com/) integration allows you to easily connect your CrowdStrike Falcon platform to Elastic for seamless onboarding of alerts and telemetry from CrowdStrike Falcon and Falcon Data Replicator. Elastic Security can leverage this data for security analytics including correlation, visualization and incident response.
 
-- `falcon` dataset: consists of endpoint data and Falcon platform audit data forwarded from Falcon SIEM Connector.
+### Compatibility
 
-    **Log File Format and Location**
+This integration is compatible with CrowdStrike Falcon SIEM-Connector-v2.0, REST API, and CrowdStrike Event Streams API.
```
kcreddy:
```suggestion
This integration is compatible with CrowdStrike Falcon SIEM Connector v2.0, REST API, and CrowdStrike Event Streams API.
```
As per documentation, there are no hyphens.

packages/crowdstrike/_dev/build/docs/README.md
```diff
@@ -1,107 +1,79 @@
...
+For Rest API support, this module has been tested against the **CrowdStrike API Version v1/v2**.
```
kcreddy:
```suggestion
This integration has been tested against the **CrowdStrike REST API Version v1/v2**.
```

kcreddy: Also I don't see `v2` in our data streams? Or did I miss it

packages/crowdstrike/manifest.yml
```diff
@@ -84,8 +84,8 @@ policy_templates:
             show_user: false
             description: This specifies the headers to be sent to the proxy server.
       - type: cel
-        title: Collect CrowdStrike logs via API
-        description: Collecting CrowdStrike logs via API.
+        title: Collect CrowdStrike Falcon Alerts, Hosts and Vulnerabilities
+        description: Collect CrowdStrike Falcon Alerts, Hosts and Vulnerabilities.
```
kcreddy: Since we have been consistent in using ingestion method for all above inputs, I think it makes sense to add `through API` at the end.

packages/crowdstrike/data_stream/falcon/manifest.yml
navnit-elastic: The App ID is a value users define themselves to distinguish between different event streams they set up. I agree that the specification should not be mentioned here, but when I first saw this configuration option, I was a bit confused about where the App ID should come from. After checking CrowdStrike’s documentation, I realized that this value must be chosen by the user. It would be helpful to include a short explanation clarifying the purpose of this parameter and how it should be used.

packages/crowdstrike/img/crowdstrike-flowchart.png
navnit-elastic: The file-based Falcon FDR data ingestion option has not gained much adoption, as it requires a relatively complex setup. This method is also not mentioned in CrowdStrike's documentation.

The flow using file-based ingestion look like this:
**Falcon FDR** → **AWS S3 Bucket** → **AWS SQS** → **FDR** → **File System** → **Elasticsearch**

here FDR in the middle is a stand-alone solution to replicate Falcon FDR data to a stand-alone target (https://github.com/CrowdStrike/FDR)

For now I'll mention this method to the diagram since we support it. I believe that we should consider removing this option if is not actively used by any users.

packages/crowdstrike/_dev/build/docs/README.md
navnit-elastic: You are correct. we do not use any `v2` endpoints, but for the hosts dataset we will be reintroducing a `v2` endpoint (ref: https://github.com/elastic/integrations/pull/16007). Since we are specifying the exact endpoints used, it is not necessary to mention API versioning here, so I think we can remove this line.

packages/crowdstrike/data_stream/falcon/manifest.yml
navnit-elastic: The Elastic Beats documentation for streaming input describe this as following:
_"This field specifies the appId parameter sent to the Crowdstrike API. See the Crowdstrike documentation for details."_

packages/crowdstrike/img/crowdstrike-flowchart.png
kcreddy: > The file-based Falcon FDR data ingestion option has not gained much adoption
> not actively used by any users.

@navnit-elastic, are you basing this conclusion based on telemetry? Just curious if we already have telemetry by input 🤔

navnit-elastic: @kcreddy, Sorry for confusion here, What was trying to convey is that I haven't seen any cases where customer is using logfile based FDR ingestion.

kcreddy: Lets avoid the logfile addition. Thanks!

packages/crowdstrike/data_stream/falcon/manifest.yml
kcreddy: > The Elastic Beats documentation for streaming input describe this as following: _"This field specifies the appId parameter sent to the Crowdstrike API. See the Crowdstrike documentation for details."_

Lets go with that. This is more concise.

packages/crowdstrike/img/crowdstrike-flowchart.png
navnit-elastic: Thanks for confirming @kcreddy, I'll add a commit to resolve comments and conflicts.

chrisberkhout (commented):
Thanks. I think this is a step in the right direction.

It's great to have a diagram in the README. I'd like to see that expanded upon. In particular, rather than having the middle be "ingestion methods" as a concept, I think it would be better to have boxes representing the integration policy, its inputs and streams, and match the labels to exactly what we have in the integration policy editor UI.

Something like this:

image

Having the specific endpoints used is good, but I would move those labels to the left, to clarify that those are CrowdStrike endpoints that the integration consumes from.

For the rest API, I think each data set involves a pair of endpoints. For example, for alerts it uses `/alerts/queries/alerts/v2` and `/alerts/entities/alerts/v2`. I had a spreadsheet of some of these details that I think it linked in another issue, but you can also find it in the integration.

It would be handy to have a file for the diagram in some kind of vector/source format, so that we can more easily update it in the future.

navnit-elastic:
>I'd like to see that expanded upon. In particular, rather than having the middle be "ingestion methods" as a concept, I think it would be better to have boxes representing the integration policy, its inputs and streams, and match the labels to exactly what we have in the integration policy editor UI.

This is a good idea. I’ll try adding those details and see how the updated diagram looks.

>Having the specific endpoints used is good, but I would move those labels to the left, to clarify that those are CrowdStrike endpoints that the integration consumes from.

Agreed, that should be updated.

>For the rest API, I think each data set involves a pair of endpoints. For example, for alerts it uses /alerts/queries/alerts/v2 and /alerts/entities/alerts/v2. I had a spreadsheet of some of these details that I think it linked in another issue, but you can also find it in the integration.

We have recently made changes to the APIs used by the integration, so I'll go through that again.

>It would be handy to have a file for the diagram in some kind of vector/source format, so that we can more easily update it in the future.

I have a Mermaid chart for diagram, but I'm not sure where the best place to would be to keep it.

I think I may need help from the team to prepare this diagram.

chrisberkhout:
> > It would be handy to have a file for the diagram in some kind of vector/source format, so that we can more easily update it in the future.
> 
> I have a Mermaid chart for diagram, but I'm not sure where the best place to would be to keep it.
> 
> I think I may need help from the team to prepare this diagram.

If it gets too complicated in Mermaid, a good option may be to use [Inkscape](https://inkscape.org/).

By default it will save as an "Inkscape SVG", which is a standard SVG with some extra information embedded that makes it easier to edit again in Inkscape. We should be able to use the SVG in the `README.md` without converting to JPEG. If the file is named `*.inkscape.svg` hopefully future editors will get the idea.

navnit-elastic:
Putting this in draft since we are changing the structure of the FDR data stream, which may lead to changes in the diagram.

navnit-elastic:
Hello @chrisberkhout,

Thank you for your suggestions. As an extension of the previous diagram, here's a clearer version built in Inkscape:

image

I'm using an Inkscape SVG format for the file in the integration. Please share your thoughts on this.

chrisberkhout:
Hi @navnit-elastic,

I like the updated diagram. Having the middle column map directly to the integration's policy editor UI is very helpful.
I see it's not currently used in the PR. Do you still plan to use it?

When you're ready for another review, please click the "Re-request review" button in the reviewer's list and it'll come up in my list of things to review.

navnit-elastic:
Thanks @chrisberkhout, While the current diagram is functional, it lacks visual appeal and the typography could be improved. I am finding it difficult to achieve the desired results in Inkscape, so I am considering switching to a different tool for these enhancements.

packages/crowdstrike/manifest.yml
```diff
@@ -60,14 +60,14 @@ policy_templates:
...
       - type: aws-s3
-        title: "Collect CrowdStrike Falcon Data Replicator logs (input: aws-s3)"
-        description: "Collecting logs from CrowdStrike Falcon Data Replicator (input: aws-s3)"
+        title: Collect Falcon Data Replicator logs using AWS S3 and SQS
+        description: Collect Falcon Data Replicator (FDR) logs using AWS S3 and SQS notifications.
       - type: streaming
-        title: Collect CrowdStrike Falcon Logs via Event Stream
-        description: Collecting CrowdStrike Falcon Logs via Event Stream.
+        title: Collect Falcon events using Event Streams API
+        description: Collect CrowdStrike Falcon events using the Event Streams API.
```
kcreddy, line 70:
```suggestion
        description: Collect Falcon events using the Event Streams API.
```
To be consistent with other titles

packages/crowdstrike/manifest.yml
```diff
@@ -84,20 +84,20 @@ policy_templates:
             show_user: false
             description: This specifies the headers to be sent to the proxy server.
       - type: cel
-        title: Collect CrowdStrike logs via API
-        description: Collecting CrowdStrike logs via API.
+        title: Collect data using the CrowdStrike REST API
```
kcreddy, line 87:
```suggestion
        title: Collect events using the CrowdStrike REST API
```
For consistency with other integrations.

packages/crowdstrike/manifest.yml
```diff
@@ -84,20 +84,20 @@ policy_templates:
...
+        description: Collect CrowdStrike Falcon data (alerts, hosts, vulnerabilities) using the REST API.
```
kcreddy, line 88:
```suggestion
        description: Collect CrowdStrike Falcon events (alerts, hosts, and vulnerabilities) using the REST API.
```

packages/crowdstrike/data_stream/falcon/manifest.yml
```diff
@@ -58,29 +58,29 @@ streams:
...
+        description: Client Secret for the CrowdStrike API.
         multi: false
         required: true
         show_user: true
         secret: true
       - name: app_id
         type: text
         title: App ID
-        description: App ID for the CrowdStrike.
+        description: This field specifies the appId parameter sent to the CrowdStrike API. See the CrowdStrike documentation for details.
```
kcreddy:
```suggestion
        description: This field specifies the `appId` parameter sent to the CrowdStrike API. See the CrowdStrike documentation for details.
```

packages/crowdstrike/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.7.0"
+  changes:
+    - description: Update the CrowdStrike Integration documentation.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/15927
```
kcreddy: Add another entry for updating layout titles and descriptions.

packages/crowdstrike/_dev/build/docs/README.md
```diff
@@ -284,23 +302,49 @@ The values used in `event.severity` are consistent with Elastic Detection Rules.
 
 If the severity name is not available from the original document, it is determined from the numeric severity value according to the following table.
 
-| Crowdstrike `severity` | Severity Name |
+| CrowdStrike Severity | Severity Name |
```
kcreddy, line 305: It would be nice to have another column showing Elastic `event.severity` mapping

kcreddy (commented):
nits.

packages/crowdstrike/manifest.yml
navnit-elastic: The policy template description isn’t shown in the integration UI; only the title is what users see when they choose an input. So changing the description wording doesn’t affect the user-facing experience. I’ve kept “CrowdStrike” in the description for consistency with the rest of the integration docs and to make it clear which product the text refers to when someone reads the manifest. If you’d still prefer to drop “CrowdStrike” from this description for consistency with other non-user-facing copy, I can do that, but I’d leave it as-is since it’s not user-visible.

packages/crowdstrike/manifest.yml
navnit-elastic: I used “data” on purpose. The REST API input collects alerts, host inventory, and vulnerability data and not only “events” in the event-streaming sense. In this codebase “events” is often used for the streaming context, so “data” is the more accurate and general term here.

packages/crowdstrike/manifest.yml
navnit-elastic: I used “data” on purpose. The REST API input collects alerts, host inventory, and vulnerability data and not only “events” in the event-streaming sense. In this codebase “events” is often used for the streaming context, so “data” is the more accurate and general term here.

packages/crowdstrike/changelog.yml
navnit-elastic: I kept a single changelog entry because “documentation” here is intended to cover all doc-facing changes in this PR: README restructure, manifest/policy template titles and descriptions, and layout. If the team prefers a separate entry per area (e.g. “Update README and layout” and “Update policy template titles and descriptions”), I can split it.

packages/crowdstrike/_dev/build/docs/README.md
navnit-elastic: The Elastic `event.severity` mapping is already documented in the table directly above. The second table explains how we derive the severity name from CrowdStrike’s numeric value when it’s missing in the source. Adding another column there would duplicate that mapping and make the troubleshooting section heavier without adding new information. Happy to add a short line that references the first table if you think it would help.

navnit-elastic (commented):
Thanks for the review. The doc and copy changes in this PR were reviewed for consistency with the rest of the integration (README, manifests, policy template) and with Elastic/CrowdStrike terminology. I’ve addressed the points above; happy to adjust further if the team has a strong preference on any of them.

packages/crowdstrike/changelog.yml
kcreddy: yes, we should split it up.

packages/crowdstrike/_dev/build/docs/README.md
kcreddy: > Happy to add a short line that references the first table if you think it would help.

Yes we should add that. Right now, it only says how `severity name` determined and doesn't mention if `event.severity` is also mapped accordingly. This is not duplication, but lack of clear information.

packages/crowdstrike/manifest.yml
kcreddy: I got the wrong line number based on your screenshots in the PR description (which seems to be old)

Screenshot 2026-03-03 at 3 40 41 PM

Update them accordingly.

navnit-elastic:
> I think we could highlight our integration in the workflow a bit with different color

Using a different color is a fair idea. I kept a natural flow from source to destination; highlighting our integration could read as promotional, so I’d leave it as-is.

packages/crowdstrike/manifest.yml
navnit-elastic: Thanks for the clarification; I’ve updated the PR description/screenshots accordingly.

packages/crowdstrike/_dev/build/docs/README.md
```diff
@@ -127,11 +100,23 @@ Current supported event types are:
...
+    | Data Stream   | Scope         |
+    | ------------- | ------------- |
+    | Alert         | read:alert    |
+    | Host          | read:host     |
+    | Vulnerability | read:vulnerability |
+
+### Collect data using CrowdStrike Falcon Data Replicator (FDR)
 
 The CrowdStrike Falcon Data Replicator allows CrowdStrike users to replicate data from CrowdStrike
 managed S3 buckets. CrowdStrike writes notification events to a CrowdStrike managed SQS queue when new data is available in S3.
```
chrisberkhout: > CrowdStrike writes notification events to a CrowdStrike managed SQS queue when new data is available in S3.

I'm not sure this is true. Doesn't CrowdStrike just write to S3 and it's set up in AWS so S3 changes get SQS messages?

navnit-elastic: I’ve updated the description so it’s clear CrowdStrike writes data to S3 and notifications are sent to the SQS queue via S3 event notifications configured in AWS.

packages/crowdstrike/_dev/build/docs/README.md
navnit-elastic: Added in https://github.com/elastic/integrations/pull/15927/commits/2b02a13c739673a6fcfc244522db35412489816b.

navnit-elastic:
> We list specific types of events that are supported for event stream data, and maybe that's fine as is. It would be good to refer to documentation if it exists that lists all the possible types.

Yes, that's a good idea. However, we don't currently have such documentation.

> For FDR, what I've seen by looking at the tool is that it'll ship any kind of data and there are many possible event types. That could perhaps be explain more clearly.

Agreed.


For the diagram color (Krishna's idea), we can revisit adding color to distinguish the integration from the source and destination in a future change, if the team wants to.

Issue #15723, opened 2025-10-22
Sentinel One: ECS compatibility


kcreddy:
_This is a broad issue without scope._

Identify ECS improvements and apply them to Sentinel One integration for all data streams.

mohitjha-elastic:
[PR ](https://github.com/elastic/integrations/pull/15931)is merged hence closing the issue.

PR #15771, opened 2025-10-27
[sentinel_one] Add Support for Threat Event Data Stream

mohitjha-elastic:
## Proposed commit message
```
sentinel_one: Add support for threat event data stream.

Added support for ingesting data through the SentinelOne Threat Event data stream, 
enabling the collection and parsing of threat-related events for accurate ingestion
and processing of security insights.

Tested on the live samples collected through the SentinelOne API.
```

## Checklist

- [X] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [X] I have verified that all data streams collect metrics or logs.
- [X] I have added an entry to my package's `changelog.yml` file.
- [X] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [X] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally

- Clone integrations repo.
- Install elastic package locally.
- Start elastic stack using elastic-package.
- Move to integrations/packages/sentinel_one directory.
- Run the following command to run tests.

> elastic-package test -v

## Related Issue

- Closes https://github.com/elastic/integrations/issues/15633
- Closes https://github.com/elastic/enhancements/issues/25418

PR #15931, opened 2025-11-11
[sentinel_one] Enhance ECS Mappings and Unify Field Structures across all Data Streams

mohitjha-elastic:
## Proposed commit message
```
sentinel_one: Enhance ECS mappings and unify fields across all data streams.

Refined and expanded ECS field mappings to ensure consistency across all data streams. 
Aligned field names and structures with the latest ECS standards to improve interoperability, 
data quality, and search normalization in Elastic SIEM.

```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/sentinel_one directory.

> elastic-package test -v

## Related Issue

- Closing https://github.com/elastic/integrations/issues/15723


packages/sentinel_one/data_stream/activity/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -481,10 +505,18 @@ processors:
...
+      copy_from: sentinel_one.activity.data.threat.classification.name
+      ignore_empty_value: true
   - rename:
       field: json.data.threatClassificationSource
       target_field: sentinel_one.activity.data.threat.classification.source
       ignore_missing: true
+  - set:
+      field: sentinel_one.threat_classification.source
+      copy_from: sentinel_one.activity.data.threat.classification.source
+      ignore_empty_value: true
```
kcreddy: I don't like that we are copying same value into 2 different fields with this change:
`sentinel_one.activity.data.threat.classification.source` and `sentinel_one.threat_classification.source`
Same with: `sentinel_one.activity.site.name`, and `sentinel_one.site.name` and several others.
Is this to avoid `breaking-change`? 

It is mentioned [here](https://github.com/elastic/enhancements/issues/20980#issuecomment-3387017714) that this is worthy to be considered as a breaking-change, with updates to dashboards. @cpascale43 can you confirm if its still the case?

If its going to be a breaking-change, then we should ideally remove the previous fields `sentinel_one.activity.data.threat.classification.source`, `sentinel_one.activity.site.name`, etc. unless `preserve_duplicate_custom_fields` is `true`.

I checked the pre-built detection rules for [`sentinel_one`](https://github.com/search?q=repo%3Aelastic%2Fdetection-rules+%22sentinel_one.%22&type=code) and there is no effect there if its a breaking-change.

packages/sentinel_one/data_stream/activity/agent/stream/httpjson.yml.hbs
```diff
@@ -42,6 +42,7 @@ response.pagination:
 cursor:
   last_create_at:
     value: '[[.last_event.createdAt]]'
+    ignore_empty_value: true
```
kcreddy, line 46: Can you add this to changelog and commit message?

packages/sentinel_one/data_stream/application/fields/unified-fields.yml
```diff
@@ -0,0 +1,8 @@
+- name: sentinel_one
+  type: group
+  fields:
+    - name: account
+      type: group
+      fields:
+        - name: name
+          type: keyword
```
kcreddy, line 8: Can you confirm if `site.id` and `site.name` is not available in all data streams as per API reference, or if we just ignored adding to the pipeline and hence ignored here as well ?

kcreddy (commented):
@cpascale43 can you confirm if this requires a dashboard changes as well?

packages/sentinel_one/data_stream/activity/agent/stream/httpjson.yml.hbs
mohitjha-elastic: It was added here to avoid CI failures since the related [PR ](https://github.com/elastic/integrations/pull/15771/files#diff-bb09f2c71a05ae2c898349ba02d5adecc707a50b02d3608489ce1648d92261ed) hadn’t been merged yet. As that PR has been merged today, this change will disappear once I sync the latest updates.

packages/sentinel_one/data_stream/application/fields/unified-fields.yml
mohitjha-elastic: We had ignored this since it’s not present in the pipeline neither in the test samples. I’ll review the API documentation again to verify whether these fields exist in the data streams and will update both the pipeline and this if needed.

packages/sentinel_one/data_stream/activity/elasticsearch/ingest_pipeline/default.yml
mohitjha-elastic: Yes, only a few fields are currently being copied into the `sentinel_one.*` namespace. We decided to keep the existing fields to avoid introducing breaking changes that could impact existing users.

I’ll wait for @cpascale43's suggestion on this. If we can treat it as a breaking change, we can proceed with removing the old fields and updating the dashboards if needed.

cpascale43: Thanks @mohitjha-elastic @kcreddy I think since the prebuilt rules are not affected, we can remove the old fields and update the dashboards

kcreddy: @mohitjha-elastic, can you update this as a `breaking-change` in this case?

packages/sentinel_one/data_stream/application/fields/unified-fields.yml
mohitjha-elastic: Verified the fields against the API documentation and updated them wherever applicable. These fields do not exist in some datasets.

packages/sentinel_one/data_stream/activity/elasticsearch/ingest_pipeline/default.yml
kcreddy: @narph , in our last discussion you mentioned whether we wanted to do this as a breaking-change or not. Do you think this is fine, or should be hold off on the approval?

kcreddy: Confirmed, we will be going forward with the breaking-change.

PR #15942, opened 2025-11-11
[MIMECAST] Allow setting of batch size for all datastreams

StacieClark-Elastic:
## Proposed commit message

Added setting batch_size in manifest.yml in the mimecast siem_logs, cloud_integrated_logs, threat_intel_malware_customer and threat_intel_malware_grid datastreams. batch_size is used to control page size for pagination. Without pagination close to 100K events have been seen being sent. This can cause memory pressure. Setting the batch_size is an attempt to reduce the memory pressure. The trade-off is mimecast V2 authentication is rate limited.

## Checklist

Install package using elastic-package.
Configure the integration using the V2 API. Verify that the batch size parameter exists for all datastreams under Mimecast v2 API in the advance options section.

### DIFF
--- a/packages/mimecast/changelog.yml
+++ b/packages/mimecast/changelog.yml
@@ -1,8 +1,8 @@
 # newer versions go on top
-- version: "3.3.0"
+- version: "3.2.2"
   changes:
-    - description: Allow setting of `batch size` in the siem_logs datastream
-      type: enhancement
+    - description: Allow setting of `batch size` in all datastreams.
+      type: bugfix
       link: https://github.com/elastic/integrations/pull/15942
 - version: "3.2.1"
   changes:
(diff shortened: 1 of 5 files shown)


packages/mimecast/data_stream/siem_logs/manifest.yml
```diff
@@ -75,6 +75,14 @@ streams:
         show_user: false
         description: Initial interval for the first API call. Defaults to 24 hours. Supported units for this parameter are h/m/s.
         default: 24h
+      - name: batch_size
```
efd6, line 78: Would you mind doing the same for the other data streams where it is also missing? (this is cloud_integrated_logs, threat_intel_malware_customer and threat_intel_malware_grid)

packages/mimecast/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.3.0"
+  changes:
+    - description: Allow setting of `batch size` in the siem_logs datastream
+      type: enhancement
```
efd6: The agent template refers to the var, but it is not defined in the manifests, so this is a bug.

packages/mimecast/data_stream/siem_logs/manifest.yml
StacieClark-Elastic: done

packages/mimecast/changelog.yml
StacieClark-Elastic: changed

PR #15948, opened 2025-11-12
[google_workspace] drop events that have missing `items[]` in the raw logs from the API

brijesh-elastic:
```
google_workspace: Discard events that are missing the `items[]` field during the split operation
and returned as the root object

The Google Workspace Reports API sometimes does not return the `items[]` array, resulting
in the absence of the target field in the `response.split` operation. This leads to the
root level object being returned, which causes failures in the ingest pipeline.

An issue[1] has been created to resolve the problem with the split[].ignore_empty_value operation.

To address this issue as of now, a `drop` processor has been added at the start of the pipeline to ensure
that we discard events that are not required.

Here is the list of affected data streams:

- access_transparency
- admin
- context_aware_access
- device
- drive
- gcp
- group_enterprise
- groups
- login
- rules
- saml
- token
- user_accounts

[1] https://github.com/elastic/beats/issues/47699
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/google_workspace directory.

> elastic-package test -v

### DIFF
--- a/packages/google_workspace/changelog.yml
+++ b/packages/google_workspace/changelog.yml
@@ -1,9 +1,9 @@
 # newer versions go on top
 - version: "2.47.2"
   changes:
-    - description: Discard events that are missing the target during the split operation and are subsequently returned as the root object.
+    - description: Discard events that are missing the `items[]` field during the split operation and are returned as the root object.
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/1
+      link: https://github.com/elastic/integrations/pull/15948
 - version: "2.47.1"
   changes:
     - description: Fix handling of `google_workspace.gmail.message_info.post_delivery_info.interaction.attachment` records.


brijesh-elastic:
> I think the problem, at least as described, should be avoided by using the [`ignore_empty_value` option](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-httpjson#_response_split_ignore_empty_value).
  If that's not the case, can you please explain why not?

We already have `ignore_empty_value: true` set ([see here](https://github.com/elastic/integrations/blob/main/packages/google_workspace/data_stream/drive/agent/stream/httpjson.yml.hbs#L41-L46)), but events are still being published.
If we have one level of `response.split`, then the event will be discarded when `ignore_empty_value: true` is configured.

chrisberkhout:
> If we have one level of `response.split`, then the event will be discarded when `ignore_empty_value: true` is configured.

Okay, so it's not working because it's multi-level. That sounds like a beats bug.

I'll take a look there and probably file an issue. If so, we can merge this workaround with a comment that refers to the issue and we should probably also look for the same issue in other integrations.

brijesh-elastic:
@chrisberkhout, Sounds good. Please share the issue link once you confirm and file it so that I can add a comment with the link to the ingest pipeline. Meanwhile, I'll look for other integrations that may have the same issue.

brijesh-elastic:
The following data streams may have a similar issue. They have multiple `response.split` operations and are missing a `drop` processor in the ingest pipeline to remove those root element events:

- ceph.osd_performance
- ceph.psd_pool_stats
- ceph.osd_tree
- ceph.pool_disk
- github.dependabot
- panw_cortex_xdr.alerts
- panw_cortex_xdr.incidents

PR #15950, opened 2025-11-12
Billing integration add tags

3kt:
## Proposed commit message

This adds a toggle button on the billing integration allowing the collection of deployment tags, which are then added to each billing object related to the corresponding deployment. In case the toggle isn't activated, the old behavior is kept identical

## Considerations

There are some caveats:

1. As the commercial cloud URL is hardcoded, this is only functional for commercial cloud and not gov cloud. This also prevents the override to admin.
2. The tags are stored as a list of `key:value` strings, rather than nested objects to prevent mapping explosion.
3. This doesn't consider historical tags. In other words, the deployment receive the tags that they have at the moment of the data pull, rather than at the time the billing occurred.
4. This doesn't work for serverless projects.

## Checklist

## Author's Checklist

- [x] Validate that the caveats are accepted
-> moved to configurable parameter for deployment search

- [ ] Evaluate whether we want to extend dashboards 
- [x] Evaluate whether we want to make the tag pull behavior default, and remove the control of the user over the setting altogether.
-> Kept default behavior

1. Spin a local environment with `elastic-package stack up -v -d`
2. Set some relevant tags on the deployment of your choice
3. Collect the organization ID and billing API key for your organization.
4. Deploy the integration while activating the `Add deployment tags` toggle.

## Related issues

None

## Screenshots

image


asazallesmilner:
We definitely have people using this for gov cloud.  Can we expose that as a configurable parameter (separate from the main base URL if needed ) ?

JohannesMahne:
The functionality works on ECH deployments, but, for some reason we cannot add tags on Serverless projects:

```
Something went wrong

patch request validation failed: field is not allowed in the current configuration: project tagging is not enabled
```

asazallesmilner:
I tried patching the code to be configurable, but doesn't seem to work with govcloud still.

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
```diff
@@ -18,113 +18,222 @@ state:
...
+          "size": 100
+        }.encode_json()
       }
-    ).as(req, (req.to > now) ?
-      // We would fetch data in the future, back off
+    ).do_request().as(deployments_resp, (deployments_resp.StatusCode == 200) ?
+      bytes(deployments_resp.Body).decode_json().as(deployments_body,
+        zip(
+          deployments_body.deployments.map(d,
+            d.resources.elasticsearch[0].id
```
ghost, line 63: Is it possible that this can have no items and break the code?

JohannesMahne: Realistically: No, this should always have a value. Even during plan changes the Elasticsearch resource configuration remains present in the API response - only the underlying infrastructure changes, not the logical resource definition.

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
```diff
@@ -18,113 +18,222 @@ state:
...
+                "from": [req.from.format(time_layout.RFC3339)],
+                "to": [req.to.format(time_layout.RFC3339)],
+              }.format_query()
+            ).with(
+              { "Header": { "Authorization": ["ApiKey " + state.api_key] } }
+            ).do_request().as(resp, (resp.StatusCode == 200) ?
+              bytes(resp.Body).decode_json().as(body, (has(body.instances) && size(body.instances) > 0) ?
+                {
+                  "events": body.instances.map(instance,
+                    instance.product_line_items.map(line_item,
```
JohannesMahne, line 98: Would `instance.product_line_items` be empty or missing for serverless projects where there wasn't any usage for 24 hours?

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
```diff
@@ -18,113 +18,222 @@ state:
...
+                              "organization_id": state.organization_id,
+                              "from": req.from.format(time_layout.RFC3339),
+                              "to": req.to.format(time_layout.RFC3339),
+                              "quantities": null
+                            }
+                          ),
+                        },
+                      }
+                    )
+                  ).flatten(),
```
JohannesMahne, line 118: I saw in the code you are replacing there is a comment `// overwrite useless quantities to avoid flatten failure` where you set `"quantities": null`. Is it not possible that we can have a similar issue with `product_line_items` itself being empty or missing?

If a serverless project has no usage in a 24h window, I guess the billing API would return an instance with empty `product_line_items`, right? If the field is missing entirely, I think line 98's `.map()` would fail with a runtime error. If all instances have empty `product_line_items` arrays, If I understand it correctly, it will cause `[[], [], []].flatten() → []`. 

Maybe we can add a filter to make sure that it is not empty before the `.map()`:
`.filter(instance, has(instance.product_line_items) && size(instance.product_line_items) > 0)`

Of course, if all instances are filtered out, we'd still need to ensure a fake event like you have on line 123.

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
3kt: Can an active serverless project ever be billed 0? I don't think that's the case, since there is a "flat" cost just to maintain it.

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
3kt: > If a serverless project has no usage in a 24h window
Can you ever have this? My understanding is that there's blanket usage baseline pushed to all projects.

JohannesMahne: > My understanding is that there's blanket usage baseline pushed to all projects.

OK, if that is the case, then it is not needed. 👍🏼

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
JohannesMahne: Resolving this item.

JohannesMahne:
@asazallesmilner - would you mind reviewing this PR again?

packages/ess_billing/data_stream/billing/fields/fields.yml
```diff
@@ -21,6 +21,9 @@
     - name: deployment_type
       type: keyword
       description: Type of the Elasticsearch Service deployment.
+    - name: deployment_tags
+      type: keyword
+      description: Tags associated with the Elasticsearch Service deployment.
```
agithomas, line 26:
```suggestion
    - name: deployment_tags
      type: keyword
      description: >-
        Tags associated with the Elasticsearch Service deployment, stored as 
        "key:value" strings. Reflects current tags at collection time.
```

packages/ess_billing/data_stream/billing/agent/stream/cel.yml.hbs
agithomas, line 63: Could you double-check if this must be replaced as follows
```
zip(
  deployments_body.deployments.map(d, d.id),                    // ← Use deployment ID
  deployments_body.deployments.map(d, d.metadata.collate("tags"))
)
```

gpop63 (approved):

Issue #15928, opened 2025-11-11
NVIDIA GPU Monitoring : Wrong configuration for SSL parameter


moxarth-rathod:
In the input configuration, the SSL parameter is set to multi: true and the conditional check uses ssl.certificate_authorities, which is incorrect.

This should be corrected by setting multi: false for the SSL parameter and updating the conditional check to use ssl instead of ssl.certificate_authorities.

agithomas:
++ @Linu-Elias

jsoriano:
PR open to fix this: https://github.com/elastic/integrations/pull/16001

PR #16001, opened 2025-11-17
[nvidia_gpu] Fix variable that was expected to be an array

jsoriano:
Fix https://github.com/elastic/integrations/issues/15928.

Variable is defined as array, but expected to be used as single value.


moxarth-rathod:
@jsoriano can we please add the necessary labels to the PR?

PR #16011, opened 2025-11-18
Salesforce wrap HTTP links in < >

alaudazzi:
FIxes https://github.com/elastic/integration-docs/issues/881


alaudazzi:
@shmsr 
This fix didn't work. We try an alternative fix in https://github.com/elastic/integrations/pull/16023

PR #16015, opened 2025-11-18
[checkpoint] Update docs checkpoint

vinit-chauhan:
## Proposed commit message
Update Docs for checkpoint integration. 

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/integration-experience/issues/406

## Screenshots

image

### DIFF
--- a/packages/checkpoint/_dev/build/docs/README.md
+++ b/packages/checkpoint/_dev/build/docs/README.md
@@ -5,6 +5,7 @@
 The Check Point integration for Elastic collects logs from Check Point Security Gateways and Management Servers. This enables comprehensive security monitoring, threat detection, and network traffic analysis within the Elastic Stack. By ingesting Check Point logs, you can gain centralized visibility into firewall traffic, security policies, VPN activity, and system health to enhance your security posture.
 
 This integration facilitates:
+
 - Centralized visibility into security policies, blocked connections, accepted connections, and VPN activity.
 - Compliance auditing and reporting by centralizing Check Point logs.
 - Enhanced threat detection and incident response capabilities through real-time log analysis.
@@ -12,9 +13,9 @@ This integration facilitates:
 
 ### Compatibility
 
-This integration is compatible with Check Point Security Gateways and Management Servers running R80.x, R81, and R81.x versions.
+This integration is compatible with Check Point Security Gateways and Management Servers running R81.x versions.
 
-This integration is compatible with Elastic Stack version 8.11.0 or higher.
+This integration is compatible with Elastic Stack versions 8.11.0 or higher.
 
 ### How it works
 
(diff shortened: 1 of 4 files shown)


packages/checkpoint/docs/knowledge_base/service_info.md
```diff
@@ -0,0 +1,156 @@
...
+- Check Point Security Gateway and Management Server logs in Syslog format (firewall connections, VPN, audit, system events).
+- Check Point firewall connection logs (accept, drop, reject, etc.).
+- VPN logs.
+- SmartConsole audit logs (administrator actions on the Management Server).
+- Gaia OS system-level logs (e.g., `/var/log/messages`, `/var/log/secure`, `/var/log/dmesg`) for appliance health and activity.
+
+## Compatibility
+
+- **Check Point**: This integration is compatible with Check Point Security Gateways and Management Servers running R80.x, R81, and R81.x versions.
+- **Elastic Stack**: Requires Kibana version 8.11.0 or later, or 9.0.0 or later.
```
mjwolf: The kibana version is also in the manifest, so it doesn't need to be here. You can either remove it entirely, or say something like "The stack version and agentless support is determined by the manifest."

packages/checkpoint/docs/knowledge_base/service_info.md
```diff
@@ -0,0 +1,156 @@
...
+- If any API-based collection were part of this integration, verify API keys, access tokens, and their associated permissions within the Check Point system.
+
+## Vendor Resources
+
+- [Check Point R81 Logging and Monitoring Administration Guide - Log Exporter Configuration](https://sc1.checkpoint.com/documents/R81/WebAdminGuides/EN/CP_R81_LoggingAndMonitoring_AdminGuide/Topics-LMG/Log-Exporter-Configuration-in-SmartConsole.htm?tocpath=Log%20Exporter%7C%5F%5F%5F%5F%5F2)
+- [Check Point R81 Logging and Monitoring Administration Guide - Log Exporter Overview](https://sc1.checkpoint.com/documents/R81/WebAdminGuides/EN/CP_R81_LoggingAndMonitoring_AdminGuide/Topics-LMG/Log-Exporter.htm)
+- [Check Point sk122323: R80.x / R81.x Log Exporter command line utility](https://support.checkpoint.com/results/sk/sk122323)
+- [Check Point sk115015: How to collect logs from Check Point Security Management Server and Security Gateway](https://support.checkpoint.com/results/sk/sk115015)
+
+# Documentation sites
```
mjwolf, line 149: I don't think you need to have these links duplicated in both sections

packages/checkpoint/_dev/build/docs/README.md
```diff
@@ -1,71 +1,171 @@
...
+- Centralized visibility into security policies, blocked connections, accepted connections, and VPN activity.
+- Compliance auditing and reporting by centralizing Check Point logs.
+- Enhanced threat detection and incident response capabilities through real-time log analysis.
+- Monitoring of system health and administrator actions on Check Point appliances.
 
-## Data streams
+### Compatibility
 
-The Check Point integration collects one type of data: logs.
+This integration is compatible with Check Point Security Gateways and Management Servers running R80.x, R81, and R81.x versions.
```
mjwolf: > running R80.x, R81, and R81.x versions

Is this true for all these versions? The previous docs only said "This integration has been tested against Check Point Log Exporter on R81.X."

packages/checkpoint/_dev/build/docs/README.md
```diff
@@ -1,71 +1,171 @@
...
+### Set up steps in Kibana
+
+1.  In Kibana, navigate to **Management > Integrations**.
+2.  Search for "Check Point" and select the integration.
+3.  Click **Add Check Point**.
+4.  Follow the prompts to add the integration to an existing Elastic Agent policy or create a new one.
+5.  Configure the input types based on your vendor setup:
+    - **For UDP/TCP (Syslog)**:
+      - Select **Collect Check Point firewall logs (input: tcp)** or **(input: udp)**.
+      - Specify the `Syslog Host` (IP address of the Elastic Agent where logs will be received).
```
mjwolf: > Specify the `Syslog Host` (IP address of the Elastic Agent where logs will be received).

I think this could be confusing (although the existing variable name of "Syslog Host" is the most confusing part). This is actually the address to listen on. I think we should make the description more clear, given the name is already not the best.

 Maybe something like "Specify the `Syslog Host` (The IP address or hostname for the Elastic Agent to listen on. Use `0.0.0.0` to listen on all available network interfaces.)"

vinit-chauhan: Completely agree.

vinit-chauhan: I think it'd also be useful to add the same description in `manifest.yml`, WDYT?

PR #16018, opened 2025-11-19
[system_otel] Use `metrics-*` DataViewId 

girodav:
This PR changes the dataViewId and IndexPattern to `metrics-*` in the Opentelemetry Systems Dashboard.


girodav:
@rogercoll / @ChrsMark could you give it a try? I don't see anymore any error like "Could not find data view..." with these changes, but I'd appreciate a 2nd pair of eyes.

mykola-elastic:
1. Noticed the following:
```
"esql": "FROM metrics-hostmetricsreceiver.otel-*\n| WHERE metrics.system.cpu.utilization IS NOT NULL\n| STATS cpu_util_per_state = AVG(metrics.system.cpu.utilization) BY state = attributes.state \n| STATS `CPU Utilization` = 1 - SUM(cpu_util_per_state) WHERE state IN (\"idle\", \"wait\")"
```

So, why not `FROM metrics-*` ?

2. Also panels have `data_stream.dataset` filters. Why not having one filter per dashboard on the top?

ChrsMark:
> 1. Noticed the following:
> 
> ```
> "esql": "FROM metrics-hostmetricsreceiver.otel-*\n| WHERE metrics.system.cpu.utilization IS NOT NULL\n| STATS cpu_util_per_state = AVG(metrics.system.cpu.utilization) BY state = attributes.state \n| STATS `CPU Utilization` = 1 - SUM(cpu_util_per_state) WHERE state IN (\"idle\", \"wait\")"
> ```
> 
> So, why not `FROM metrics-*` ?
> 
> 2. Also panels have `data_stream.dataset` filters. Why not having one filter per dashboard on the top?

The current PR fixes a specific issue afaik and the patch looks specific to this need. I would avoid changing the queries or filters in the same PR, but feel free to file a follow-up if needed.


> @rogercoll / @ChrsMark could you give it a try? I don't see anymore any error like "Could not find data view..." with these changes, but I'd appreciate a 2nd pair of eyes.

I could install the updated dashboard on a fresh serverless project and verified that the DataView issue does not occur any more:
Screenshot 2025-11-26 at 3 14 28 PM
Screenshot 2025-11-26 at 3 14 56 PM

Screenshot 2025-11-26 at 3 15 10 PM
Screenshot 2025-11-26 at 3 15 27 PM


@rogercoll please take a look and I think we can merge this.

Note: cloud.* attributes are missing because I'm testing on a local k8s cluster.

PR #15445, opened 2025-09-24
o365: fix error propagation within cel program

efd6:
## Proposed commit message
```
o365: fix error propagation within cel program

The logic for handling error propagation from failed subscriptions
requests was incorrect. The downstream logic was expecting to find the
error in the root of state, but the subscription request block was
leaving it in the work object.

Later error handling was incorrectly leaving any error state in the root
of state, when it should have been placed under events.

In the case that a subscription has failed, there may not be an entry in
the last_for list for the subscription, resulting in a look-up failure
during assessment for completion of the hour-window steps.

Finally, the postamble logic was brittle to failed requests, assuming
that the todo_type and todo_content list would be present, which is not
necessarily the case when a subscription has failed.

This has been manually tested against the mock committed in the deploy
directory (with removal of authentication for convenience) with the
following outcomes:

only one, invalid, subscription:
"""
-- data.json --
{
	"url": "http://localhost:9001",
	"want_more": false,
	"base": {
		"tenant_id": "test-cel-tenant-id",
		"list_contents_start_time": "12h",
		"batch_interval": "1h",
		"maximum_age": "167h55m",
		"content_types": "Audit.TypeRequiringAdditionalPermissions"
	}
}
-- out.json --
{
	"base": {
		"batch_interval": "1h",
		"content_types": "Audit.TypeRequiringAdditionalPermissions",
		"list_contents_start_time": "12h",
		"maximum_age": "167h55m",
		"tenant_id": "test-cel-tenant-id"
	},
	"events": {
		"error": {
			"code": "401",
			"id": "401 Unauthorized",
			"message": "POST /activity/feed/subscriptions/start?contentType=Audit.TypeRequiringAdditionalPermissions: {\"error\":{\"code\":\"AF10001\",\"message\":\"The permission set (...) sent in the request does not include the expected permission.\"}}"
		}
	},
	"url": "http://localhost:9001",
	"want_more": false,
	"work": {
		"curr_type": "Audit.TypeRequiringAdditionalPermissions",
		"todo_type": []
	}
}
"""

start with an invalid subscription:
"""
-- data.json --
{
	"url": "http://localhost:9001",
	"want_more": false,
	"base": {
		"tenant_id": "test-cel-tenant-id",
		"list_contents_start_time": "12h",
		"batch_interval": "1h",
		"maximum_age": "167h55m",
		"content_types": "Audit.TypeRequiringAdditionalPermissions, Audit.SharePoint"
	}
}
-- out.json --
{
	"base": {
[...]

PR #15325, opened 2025-09-15
o365: tolerate changed API next page URI behaviour

efd6:
## Proposed commit message
```
o365: tolerate changed API next page URI behaviour

The API now returns the next page URI with the parameters lowercased.
This is not documented[1], so retain support for the document behaviour.

[1]https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-reference
```

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [ ] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## Related issues

-

PR #16024, opened 2025-11-19
[o365] Simplification of data fetching logic

### LINKED ISSUE #15066: o365: flatten structure, one request per evaluation
Following https://github.com/elastic/integrations/pull/14962, we can simplify the structure of the cel program as described in https://github.com/elastic/integrations/pull/14962#discussion_r2287871651.

Benefits:
* Less work/requests per evaulation
* Mapping from existing to new stored state is more direct
* Can do less frequent subscription requests (e.g. once per input startup, or time-based periodic)

chrisberkhout:
## Proposed commit message

```
[o365] Simplification of data fetching logic

Flattens the structure of the CEL program.

Simplifications:
- Make only one request per evaluation (or none).
- Expired items are skipped during fetching rather than filtering them
  out in multiple places.
- The non-canonical header name `NextPageUri` is no longer considered,
  as it's always normalized by the HTTP client.
- Assume that items can be fetched in the order they are listed.
- Assume that content items will not be empty.
- Update the `last_for` times once based on the listing range, rather
  than repeatedly (with the same value) for each followed listing link.
- Unify handling of generated listing links (for initial requests) and
  received listing links (for later pages).
- Subscribe once per input start (an alternative to once for the life of
  cursor data, as introduced in #15476).

Other changes:
- Moves some state into `state.cursor`, so that it persists across
  restarts: `state.work.todo_content` → `state.cursor.todo_content`,
  `state.work.next_list` (string) → `state.cursor.todo_links` (array).
- Renames `state.work.todo_type` (array) → `state.todo_types` (plural
  name, array). This stays out of the cursor because it can be
  reconstructed.
- Do all subscriptions first, then rotate types so everything is roughly
  chronological rather than type-by-type.
- Adds `state.subscribed` (map). It's not in the cursor data because we
  want to resubscribe if restarted.
- Keep querying until the time 3 seconds before the start has been
  reached (exclusive). The 3 second buffer avoids requesting times that
  may have unstable results.
- Log an error if no type is configured.
- The `max_executions` limit is raised. Getting up to date means
  hour-long listings for 168 hours of data, possibly over multiple pages
  each, likely for multiple content types, and fetching everything that
  was listed.
```

## Notes for the reviewer

I started this before the last 4 PRs, and I checked that there are no conflicts with those changes: 
* #15325
  We no longer parse the next page URLs, since we have the `endTime` from the initial request.
* #15380
  #15445
  This error handling logic isn't relevant in the new flattened structure.
* #15476
  The new code still limits subscription requests, but to a lesser degree, which I think is better overall.

The new CEL code passes with the original system tests.

~The system tests have been updated to use a mock o365 server rather than the `stream` tool. The mock server has configurability, assertions and logging beyond what is used in the system test, which may be useful for future debugging. The amount of code is not insignificant, and the quality is just okay, so if this is a maintenance concern, it can be moved to a separate PR or removed entirely. Let me know what you think.~
Introduction of a new mock server has been moved to a separate PR: https://github.com/elastic/integrations/pull/16190

## Checklist

You can manually run against the new mock server.

Get the `o365mock.go` file from https://github.com/elastic/integrations/pull/16190, then start the mock server like this:

```
go run o365mock.go chunks_with_gaps_and_1_expired
```

In another terminal, run the CEL code in `mito` like this:
```bash
mito \
  -cfg <(echo '
auth:
  oauth2:
    client.id: test-cel-client-id
    client.secret: test-cel-client-secret
    provider: azure
    scopes:
      - "https://manage.office.com/.default"
    endpoint_params:
      grant_type: ["client_credentials"]
    token_url: http://localhost:9999/test-cel-tenant-id/oauth2/v2.0/token
') \
   -data <(echo '
{
	"url": "http://localhost:9999",
	"want_more": false,
	"base": {
		"tenant_id": "test-cel-tenant-id",
		"list_contents_start_time": "15h",
		"batch_interval": "1h",
		"maximum_age": "167h55m",
		"content_types": "Audit.AzureActiveDirectory, Audit.Exchange"
	}
}
') \
  -log_requests \
  <(awk '/^program:/{iscel=1; next} /^\{\{/{iscel=0} iscel' ./data_stream/audit/agent/stream/cel.yml.hbs)
```

Stop the mock server with Ctrl+C to trigger it's shutdown report.

You can remove the `mito` auth configuration if you disable `CheckAccessToken` in the mock server's inline configuration.

This version of `mito` will run faster than the [following one](https://github.com/elastic/mito/pull/112), which rate limits to 1 rps:
```
go install github.com/elastic/mito/cmd/mito@835128
```

## Related issues

- Closes #15066
- Relates #16190

### DIFF
--- a/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
+++ b/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
@@ -402,7 +402,8 @@ tags:
 {{#contains "forwarded" tags}}
 publisher_pipeline.disable_host: true
 {{/contains}}
-{{#if processors}}
 processors:
+- drop_event.when.equals.retry: true
+{{#if processors}}
 {{processors}}
 {{/if}}
--- a/packages/o365/data_stream/audit/_dev/test/policy/test-cel-all.expected
+++ b/packages/o365/data_stream/audit/_dev/test/policy/test-cel-all.expected
@@ -21,6 +21,7 @@ inputs:
           interval: 3m
           max_executions: 10000
           processors:
+            - drop_event.when.equals.retry: true
             - add_fields:
                 fields:
                     id: "574734885120952459"
--- a/packages/o365/data_stream/audit/_dev/test/policy/test-cel-default.expected
+++ b/packages/o365/data_stream/audit/_dev/test/policy/test-cel-default.expected
@@ -20,6 +20,8 @@ inputs:
             type: logs
           interval: 3m
           max_executions: 10000
+          processors:
+            - drop_event.when.equals.retry: true
           program: |-
             (
               // This exists purely to rewrite the cursor from the original list format
--- a/packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
@@ -1,9 +1,6 @@
 ---
 description: Pipeline for Office 365 Audit logs
 processors:
-  - drop:
-      if: ctx.retry == true
-      tag: drop_retry_events
   - remove:
       field:
         - organization


packages/o365/_dev/deploy/docker/o365mock.go
```diff
@@ -0,0 +1,1083 @@
...
+}
+
+func MakeListItem(cfg *Config, run *Run, contentType, timeOffset string, expirationOverride string) ListItem {
+	timeOffsetDuration, err := time.ParseDuration(timeOffset)
+	if err != nil {
+		panic(err)
+	}
+
+	contentCreated := run.StartTime.Add(timeOffsetDuration)
+	contentExpiration := contentCreated.Add(time.Hour * 25 * 7)
```
efd6: Is 25 intentional?

packages/o365/_dev/deploy/docker/o365mock.go
```diff
@@ -0,0 +1,1083 @@
...
+	}
+	return result
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Fetch item pool - Real data that can be cycled through for responses.
+//                   (CreationTime and Id will be overridden with random values)
+////////////////////////////////////////////////////////////////////////////////
+
+var fetchItemPool = []FetchItem{
```
efd6: Could this be an embed JSON file with a must unmarshal?

packages/o365/_dev/deploy/docker/o365mock.go
```diff
@@ -0,0 +1,1083 @@
...
+	SubscriptionsByType             map[string]int
+	FetchItemPoolIndex              int
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Main - Initialize the run state and config, start the server.
+//        Scenarios are defined here. The requested one is added to the config.
+////////////////////////////////////////////////////////////////////////////////
+
+func main() {
```
efd6: My preference would be to have the entry point at the top to give context for the reader.

packages/o365/_dev/deploy/docker/o365mock.go
```diff
@@ -0,0 +1,1083 @@
...
+	if len(os.Args) == 3 {
+		var err error
+		seed, err = strconv.ParseInt(os.Args[2], 10, 64)
+		if err != nil {
+			log.Printf("ERROR: Couldn't parse seed value '%s'!", os.Args[2])
+			os.Exit(1)
+		}
+	}
+	if seed == 0 {
+		seed = run.StartTime.UnixNano()
```
efd6: If there is an option to use a seed based on some entropy as is the case here, the seed should be emitted in the log to allow a replay.

packages/o365/_dev/deploy/docker/o365mock.go
```diff
@@ -0,0 +1,1083 @@
...
+			PageLimit:                20,
+			MinFetchItems:            1,
+			MaxFetchItems:            5,
+		},
+	}
+
+	if len(os.Args) < 2 {
+		log.Printf("Usage: go run o365mock.go SCENARIONAME [SEED]")
+		os.Exit(1)
+	}
```
efd6: Move this up to just under `func main()`.

packages/o365/_dev/deploy/docker/o365mock.go
```diff
@@ -0,0 +1,1083 @@
...
+
+	s := NewServer(&cfg, &run, log)
+	s.Run()
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Miscellaneous helpers
+////////////////////////////////////////////////////////////////////////////////
+
+func SortListItems(slice []ListItem) []ListItem {
```
efd6: The helpers don't need to be exported, also for the types in the preamble to the file.

packages/o365/_dev/deploy/docker/o365mock.go
```diff
@@ -0,0 +1,1083 @@
...
+	slices.SortFunc(slice, func(a, b ListItem) int {
+		switch {
+		case a.ContentCreated.Before(b.ContentCreated):
+			return -1
+		case a.ContentCreated.After(b.ContentCreated):
+			return 1
+		default:
+			return 0
+		}
```
efd6:
```suggestion
		return int(a.ContentCreated.Sub(b.ContentCreated))
```

packages/o365/_dev/deploy/docker/o365mock.go
```diff
@@ -0,0 +1,1083 @@
...
+		s.log.Printf("ERROR received tenantId %s does not match expected %s", tenantId, s.cfg.TenantId)
+	}
+	if s.cfg.CheckCredentials {
+		decoded, err := base64.StdEncoding.DecodeString(strings.TrimPrefix(authorization, "Basic "))
+		if err != nil {
+			s.log.Printf("ERROR could not decode Authorization header value '%s'", authorization)
+			http.Error(w, "bad authorization header value", http.StatusBadRequest)
+			return
+		}
+		if string(decoded) != strings.Join([]string{s.cfg.ClientId, s.cfg.ClientSecret}, ":") {
```
efd6:
```suggestion
		if string(decoded) != s.cfg.ClientId + ":" + s.cfg.ClientSecret {
```

packages/o365/_dev/deploy/docker/o365mock.go
```diff
@@ -0,0 +1,1083 @@
...
+		} else if requestGap < 0*time.Second {
+			s.log.Printf("WARNING overlap between listing ranges of %s", (-requestGap).String())
+		}
+	}
+	// update requested/served range
+	if _, ok := s.run.MinStartByType[contentType]; !ok {
+		s.run.MinStartByType[contentType] = startTime
+	} else if startTime.Before(s.run.MinStartByType[contentType]) {
+		s.run.MinStartByType[contentType] = startTime
+	}
```
efd6: These can be simplified by making use of the fact that the zero time is before any times we would be using.

packages/o365/_dev/deploy/docker/o365mock.go
```diff
@@ -0,0 +1,1083 @@
...
+	contentId := r.PathValue("contentId")
+	reqTime := time.Now()
+
+	if _, ok := s.run.ServedListItemsFetchCount[contentId]; !ok {
+		s.log.Printf("ERROR fetching unknown item contentId=%s", contentId)
+	} else {
+		s.run.ServedListItemsFetchCount[contentId] += 1
+	}
+
+	if expiry, ok := s.run.ServedItemsExpiry[contentId]; !ok || expiry.Before(reqTime) || expiry.Equal(reqTime) {
```
efd6: `!ok || expiry.Before(reqTime) || expiry.Equal(reqTime)` can be expressed as `!reqTime.Before(expiry)`

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,437 +76,291 @@ redact:
...
+        "GET",
+        state.cursor.todo_links[0]
+      ).do_request().as(list_resp, (list_resp.StatusCode == 200) ?
+        // Listing successful - Enqueue any additional link, enqueue content to fetch
+        list_resp.Body.decode_json().as(content_list,
+          state.with(
+            {
+              "cursor": state.cursor.with(
+                {
+                  // The non-canonical header name NextPageUri is always normalized to Nextpageuri
```
efd6, line 286: This is based on [this](https://cs.opensource.google/go/go/+/refs/tags/go1.25.4:src/net/http/response.go;l=50), correct? I know it should be true, but I can't find where it happens in net/http for responses.

packages/o365/_dev/deploy/docker/o365mock.go
chrisberkhout: No, should have been 24. Updated.

packages/o365/_dev/deploy/docker/o365mock.go
chrisberkhout: Yep. The log starts like this (minus wrapping):

```
2025/11/20 13:11:33 RunStart StartTime=2025-11-20T13:11:33.110076134Z,
  scenarioName=chunks_with_gaps_and_1_expired,
  seed=1763644293110076134,
  Addr=0.0.0.0:9999,
  rerun: 'PORT=9999 go run o365mock.go chunks_with_gaps_and_1_expired 1763644293110076134'
```

packages/o365/_dev/deploy/docker/o365mock.go
chrisberkhout: :+1:

packages/o365/_dev/deploy/docker/o365mock.go
chrisberkhout: :+1:

packages/o365/_dev/deploy/docker/o365mock.go
chrisberkhout: :+1:

packages/o365/_dev/deploy/docker/o365mock.go
chrisberkhout: Simplified that and following to:
```go
	// update requested/served range
	minStart := s.run.MinStartByType[contentType]
	if minStart.IsZero() || startTime.Before(minStart) {
		s.run.MinStartByType[contentType] = startTime
	}
	if endTime.After(s.run.MaxEndByType[contentType]) {
		s.run.MaxEndByType[contentType] = endTime
	}
```

packages/o365/_dev/deploy/docker/o365mock.go
chrisberkhout: :+1:

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: It goes like this:

`(*Transport).RoundTrip`
[`(*Transport).roundTrip`](https://cs.opensource.google/go/go/+/refs/tags/go1.24.10:src/net/http/transport.go;l=581-582)
...
`(*persistConn).readResponse`
`ReadResponse`
`(*Reader).ReadMIMEHeader`
`readMIMEHeader`
[`canonicalMIMEHeaderKey`](https://cs.opensource.google/go/go/+/refs/tags/go1.24.10:src/net/textproto/reader.go;l=737-794)

packages/o365/_dev/deploy/docker/o365mock.go
chrisberkhout: :+1:

packages/o365/_dev/deploy/docker/o365mock.go
chrisberkhout: Reordered to:

- Main
- Config types
- Run state type
- Miscellaneous Helpers
- API response data types and helpers
- Fetch item pool
- Server

packages/o365/_dev/deploy/docker/o365mock.go
chrisberkhout: Could be. It means there's two files to mount into the container.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6: Thanks

packages/o365/_dev/deploy/docker/o365mock.go
```diff
@@ -0,0 +1,1076 @@
+package main
```
efd6: Apparently this needs
```
// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
// or more contributor license agreements. Licensed under the Elastic License;
// you may not use this file except in compliance with the Elastic License.

```

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -75,437 +76,291 @@ redact:
...
-        )
-      :
-        {"todo_content": []}
-      ).as(w, !has(w.curr_content) ?
-        {
-          "work": state.work.drop("curr_content").with(w),
-          "events": [],
-        }
-      :
+  (
```
efd6, line 107: I think it would be nice to have an overview of the flow here for future maintainers.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
efd6, line 1: Because we lean on repeated input loops more now, I think it may be worth doing the dropping of the retry events in the agent with a beat processor, rather than sending the event to be dropped by the ingest pipeline. An example of this is [here](https://github.com/elastic/integrations/blob/75ecd7b5f6118a64ad1fe77874c24b2bfae35cc8/packages/beyondinsight_password_safe/data_stream/managedsystem/agent/stream/cel.yml.hbs#L177-L178) corresponding to this [alternative way of saying `{"retry": true}`](https://github.com/elastic/integrations/blob/75ecd7b5f6118a64ad1fe77874c24b2bfae35cc8/packages/beyondinsight_password_safe/data_stream/managedsystem/agent/stream/cel.yml.hbs#L136).

packages/o365/_dev/deploy/docker/o365mock.go
chrisberkhout: Thanks.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: Yeah, the comments and sections cover it pretty well, but it's useful to have something that fits on one screen.

I added this outside of the CEL code because otherwise `celfmt` was killing it:

```
STATE OVERVIEW
                                                                                                     
These variables are set to direct the work of later evaluations:
                                                                                                     
state.stopping_point:      The time just before the start of the current sequence of evaluations.
state.todo_types:          The current list of content types to process.
state.subscribed:          A map tracking which content types have been subscribed to since restart.
state.cursor.todo_content: A list of content items to fetch.
state.cursor.todo_links:   A list of listing links, either generated or received.
state.cursor.last_for:     A map tracking latest time considered in listings for each content type.
                                                                                                     
LOGIC OVERVIEW
                                                                                                     
Initializations that rewrite state and continue the evaluation:
- Rewrite legacy cursor data if present.
- Set a stopping point if absent.
- Generate the list of content types to process if empty.
                                                                                                     
A check done on every evaluation:
- Ensure there is an active content type.
                                                                                                     
Processing steps, only one of which will be performed per evaluation:
- If not subscribed to the current type, subscribe and rotate to the next type.
- If there is a content item ready to fetch, fetch it.
- If there is a prepared listing link, use it to list content items.
- If the current type is not up-to-date, generate a listing link for it, or
  if the current type is up-to-date, move to the next type, or
  if there are no more content types to process, stop.
```

efd6: Thanks, that perfect. Sorry about the celfmt behaviour. That's likely elastic/celfmt#58.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -1,4 +1,5 @@
 interval: {{interval}}
+max_executions: 10000
```
efd6: On the basis of a support case, we may want to make this configurable, but defaulting to a higher value like this.

chrisberkhout: :+1: Done.

chrisberkhout:
@efd6 I split the mock server part into a separate draft PR.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: Sure. Done.

PR #16083, opened 2025-11-21
add log.* mapping for elastic-agent

jrmolin:
## Checklist

- ~[ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices)~

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/integration-experience/issues/425


.buildkite/scripts/common.sh
```diff
@@ -47,7 +47,7 @@ retry() {
 
 cleanup() {
   echo "Deleting temporary files..."
-  rm -rf ${WORKSPACE}/${TMP_FOLDER_TEMPLATE_BASE}.*
+  rm -rf ${WORKSPACE}/${TMP_FOLDER_TEMPLATE_BASE:-tmp}.*
```
mrodm: I see that this was added because of this error: 
https://buildkite.com/elastic/integrations/builds/34885#019adb5a-c571-4e28-86c2-10296b72dde4/112-114

This happens because `post-checkout` hook failed (due to merge conflicts) and the `.buildkite/hooks/pre-command` hook was not executed. Once these merge conflicts with the target branch are solved, this failure should disappear.

That variable is set in the `pre-command` hook and I was supposed to have value in the `pre-exit` hook too, but this is not the case if `post-checkout` fails:
https://github.com/elastic/integrations/blob/bb632b7322719307295706416a67c51ed4b39eb3/.buildkite/hooks/pre-command#L25-L28


Looking at the current usages of these variables, I think those variables can be removed completely @jrmolin
You can revert this change, and I'll create a PR to remove those variables. Thereby, this PR just modifies `elastic_agent` package (so no need to test all the other packages in the repo).

mrodm: Created PR for that @jrmolin , I think you can revert that change in favor of the changes in https://github.com/elastic/integrations/pull/16233

mrodm: Hi @jrmolin 
#16233 has been merged, so those variables have been removed.
You would need to update your branch and solve the conflicts (sorry :sweat_smile: )

PR #16129, opened 2025-11-26
ti_opencti: fix pipeline errors due to access method from a null reference

chemamartinez:
## Proposed commit message

```
Pipeline tests were failing with the error:
"cannot access method/field [add] from a null def reference".

This failure was caused by script processors calling .add()
on fields whose parent object (threat.indicator) could be null.

The fix updates the ingest pipeline scripts to safely initialize
ctx.threat, ctx.threat.indicator, and the target array field before adding elements.
```

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/16049
- Closes https://github.com/elastic/integrations/issues/16052

### DIFF
--- a/packages/ti_opencti/data_stream/indicator/elasticsearch/ingest_pipeline/ecs_from_artifact.yml
+++ b/packages/ti_opencti/data_stream/indicator/elasticsearch/ingest_pipeline/ecs_from_artifact.yml
@@ -52,15 +52,9 @@ processors:
       description: Append to the destination
       lang: painless
       source: |
-        if (ctx.threat == null) {
-          ctx.threat = [:];
-        }
-        if (ctx.threat.indicator == null) {
-          ctx.threat.indicator = [:];
-        }
-        if (ctx.threat.indicator.file == null) {
-          ctx.threat.indicator.file = [];
-        }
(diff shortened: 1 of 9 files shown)


packages/ti_opencti/data_stream/indicator/elasticsearch/ingest_pipeline/ecs_from_artifact.yml
```diff
@@ -48,13 +48,19 @@ processors:
...
       source: |
+        if (ctx.threat == null) {
+          ctx.threat = [:];
+        }
+        if (ctx.threat.indicator == null) {
+          ctx.threat.indicator = [:];
+        }
+        if (ctx.threat.indicator.file == null) {
+          ctx.threat.indicator.file = [];
+        }
```
efd6: This can be less verbose:

```suggestion
        ctx.threat = ctx.threat ?: [:];
        ctx.threat.indicator = ctx.threat.indicator ?: [:];
        ctx.threat.indicator.file = ctx.threat.indicator.file ?: [];
```

Also below.

efd6 (approved):

PR #14995, opened 2025-08-20
Remove unused httpjson agent file from {zeek, apache, windows, aws.cloudtrail, nginx}

vinit-chauhan:
## Proposed commit message

httpjson input was removed from these integrations in [this](https://github.com/elastic/integrations/pull/13246) PR. However, the agent files were still there with the package. 

This PR removes those unused httpjson files. 

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## Related issues

- Closes #14200 

## Screenshots

image

image

image

image

image

Issue #11767, opened 2024-11-19
Remove Deprecated "Collect Logs from third-party REST API" for 9.0


qcorporation:
Parent Issue:
https://github.com/elastic/dev/issues/2802

Issue:
Based upon the work done from the parent issue, for the 9.0 release, Integrations would like to remove deprecated options entirely.
[Collect logs from third-party REST API](https://github.com/search?q=repo%3Aelastic%2Fintegrations%20%2FCollect%20logs%20from%20third-party%20REST%20API%2F&type=code)
For:

- Zeek
- Nginx
- Windows
- Apache
- System
- AWS CloudTrail

andrewkroh:
The release of these changes should coincide with the timing of the 9.0 Elastic stack release. We should prepare the changes before then so that they are ready, but we hold off on the merge until 9.0 ships.

muthu-mps:
@qcorporation - There is one more integration which is not included in this issue is `AWS cloudtrail` that's currently using the deprecated input. Is this integration already tracked somewhere?

- Also noticed that this feature is still in experimental for this integration. 
-     `Collect CloudTrail logs from third-party REST API (experimental)`

qcorporation:
> [@qcorporation](https://github.com/qcorporation) - There is one more integration which is not included in this issue is `AWS cloudtrail` that's currently using the deprecated input. Is this integration already tracked somewhere?
> 
> * Also noticed that this feature is still in experimental for this integration.
>   `Collect CloudTrail logs from third-party REST API (experimental)`

@muthu-mps I didn't know about this. We had originally marked the targeted integrations with a `(deprecation)` warning before we actually pulled the trigger on removing it. I'm glad to do it, I just need to make sure that product is aligned with that decision.

@jamiehynds @norrietaylor are you OK to remove CloudTrail as well - it's marked as `experimental` but there's forehand description that these will be deprecated

qcorporation:
> Yep, no issues with removal and prefer not to have one integration as an outlier. It's been in experimental for years and hasn't been used/maintained so very low risk in removing it.

Thanks @jamiehynds for weighing in. It'll be removed

norrietaylor:
Can we mark this as done, or is there work still to be done?

PR #16133, opened 2025-11-26
Remove unused pipelines {apache, aws.cloudtrail, nginx, zeek}

### LINKED ISSUE #15881: Consider Removing third-party pipeline invokation and definition from Nginx, Apache, AWS CloudTrail, Zeek integration
In 2021, the third-party pipeline was introduced in [elastic/integrations#545](https://github.com/elastic/integrations/pull/545).
From my understanding, it was developed to integrate or collect logs from third-party sources (e.g., Splunk) into Elastic.
In 2025, we removed the third-party stream input from the same integration in [elastic/integrations#14995](https://github.com/elastic/integrations/pull/14995).
However, the corresponding third-party ingest pipelines (and their references in the main pipeline) were not removed.
If these pipelines are no longer needed, the main pipeline is currently performing unnecessary checks.

I am opening this issue to ask whether it is expected to have these third-party ingest pipelines and invokations:

- [ ] Nginx
- [ ] Apache
- [ ] AWS CloudTrail
- [ ] Zeek

**EXAMPLE**

The NGINX Access pipeline is invoking the third-party pipeline here:

https://github.com/elastic/integrations/blob/685d79a58c5665422453dc57aeb4616f7317121d/packages/nginx/data_stream/access/elasticsearch/ingest_pipeline/default.yml#L5C3-L7C49

```yaml
  - pipeline:
      if: ctx.message.startsWith('{')
      name: '{{ IngestPipeline "third-party" }}'
```

If the third-party pipeline should not exist in the first place, we can spare the first processor in the pipeline (and remove the invoked third-party pipeline)

vinit-chauhan:
## Proposed commit message
In the PR [elastic/integrations#14995](https://github.com/elastic/integrations/pull/14995) unused `httpjson` file for third-party ingest pipelines was removed. However, the pipeline and the its reference in default pipeline still existed. 

This PR removes those references. Including,
- The pipeline yaml file ( third-party.yml )
- Redirection to third-party pipeline in default pipeline.
- Pipeline tests associated with it. 

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #15881


packages/aws/changelog.yml
efd6, line 1: This is a removal of functionality that we do not know that people are not using (apologies for the mass of negatives). I'm not sure that this is safe to do, but if it is, it would be a breaking change.

vinit-chauhan: The ingestion path for third-party Rest APIs has already been remove quite a while ago in [#11767](https://github.com/elastic/integrations/issues/11767). However, the reference to the ingest pipeline still exists. 
I'm note sure about the telemetry around that. 

@qcorporation - do you have any telemetry just to make sure?

vinit-chauhan: Updated the PR to highlight it as a breaking change Just in case someone's still using the Pipeline.

packages/aws/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 3.3.2
 name: aws
 title: AWS
-version: "4.5.0"
+version: "4.5.1"
```
efd6: This will need to match the version in the version changelog. Also please check that obs are OK with ignoring semver and making a breaking change only bump minor.

packages/aws/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "4.7.0"
```
efd6: Can @elastic/obs-infraobs-integrations and @elastic/obs-ds-hosted-services confirm that they are happy to use non-semver semantics for breaking changes here?

ishleenk17: Any reason why we are not using the 5.0.0 version(like we have done for other integrations) if we are considering this a breaking change ?

vinit-chauhan: I know that AWS is a quite big integration and in past I saw breaking-change as a minor update. [here](https://github.com/elastic/integrations/blob/5a5d30c6ebf4fef2c4226a67c31d772ceb93230d/packages/aws/changelog.yml#L250-L254). 

So my question would be, for a breaking-change in one data-stream, are we strictly following the semver for AWS or do we have any other strategy? 

I'm okay either way, let me know what are your thoughts on this and I'll update the PR.

efd6: I would prefer that we use semver semantics, but I'm not a code owner for the package as a whole. The question above is just deferring to convention in this package.

vinit-chauhan: > I would prefer that we use semver semantics, but I'm not a code owner for the package as a whole. The question above is just deferring to convention in this package.

++

ishleenk17: @tommyers-elastic : WDYT ?
Also @elastic/obs-ds-hosted-services : Since they are the codeowners as well here.

tommyers-elastic: are there any downsides to bumping the major version? i would also bump to v5

ishleenk17: I don't see any issues, since the kibana version remains the same.

vinit-chauhan: Alright, I see it a unanimous vote for bumping up the major version. Let me update the PR.

PR #16145, opened 2025-11-27
ti_google_threat_intelligence: update default initial interval

chemamartinez:
## Proposed commit message

[API specifications for Google Threat Lists](https://gtidocs.virustotal.com/reference/get-hourly-threat-list) changed the availability time range of generated lists from one hour to two hours back.

> Threat Lists are hourly generated as IoCs packages, with 2 hours difference from the current time. This means that if the current time in UTC is T you can get T-2h Threat List but any more recent than that.

Default initial intervals have been updated to align with that requirement.

## Checklist

### DIFF
--- a/packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
+++ b/packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
@@ -33,7 +33,7 @@ program: |
         ?"query": has(state.query) ? optional.of([state.query]) : optional.none(),
         "limit": ["4000"],
         "x-tool": ["Elastic"],
-        "User-Agent": ["v0.7.0"], // Keep this in sync with 'version' in package level manifest.yml.
+        "User-Agent": ["v0.8.0"], // Keep this in sync with 'version' in package level manifest.yml.
       }.format_query()
     ).with({
       "Header": {


kcreddy (commented):
LGTM. 
For `initial_interval` and `interval`, we should consider the `duration` type with [`min_duration`](https://github.com/elastic/package-spec/blob/main/spec/integration/data_stream/manifest.spec.yml#L73-L78) option to help with such validations.

chemamartinez:
Thanks, @kcreddy

> For `initial_interval` and `interval`, we should consider the `duration` type with [`min_duration`](https://github.com/elastic/package-spec/blob/main/spec/integration/data_stream/manifest.spec.yml#L73-L78) option to help with such validations.

It is a good point, the problem I see right now is that would force the integration to bump its minimum Kibana version to 9.2.0, and this is related to an SDH so I'd wait to add this change for now.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
```diff
@@ -33,7 +33,7 @@ program: |
         ?"query": has(state.query) ? optional.of([state.query]) : optional.none(),
         "limit": ["4000"],
         "x-tool": ["Elastic"],
-        "User-Agent": ["v0.6.1"], // Keep this in sync with 'version' in package level manifest.yml.
+        "User-Agent": ["v0.7.0"], // Keep this in sync with 'version' in package level manifest.yml.
```
kcreddy:
```suggestion
        "User-Agent": ["v0.8.0"], // Keep this in sync with 'version' in package level manifest.yml.
```

Nit.
@chemamartinez, please update them accordingly.

chemamartinez: Thank you for the reminder!

PR #16159, opened 2025-11-28
Fix indentation in filestream.yml.hbs

lucabelluccini:
The indentation prevents the correct usage of processors.

It triggers
```
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": """end of the stream or a document separator is expected (9:1)

  6 |   "##scanner##"
  7 | processors:
  8 |   - add_locale: ~
  9 | - drop_event:
------^
 10 |     when:
 11 |       not:"""
}
```

When saving

- Bug

## Proposed commit message

The indentation prevents the correct usage of custom processors.

## Checklist

### DIFF
--- a/packages/microsoft_exchange_server/changelog.yml
+++ b/packages/microsoft_exchange_server/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.4.1"
+  changes:
+    - description: Fix the input templates to correctly include custom processors.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/16159
 - version: "1.4.0"
   changes:
     - description: Allow @custom pipeline access to event.original without setting preserve_original_event.
--- a/packages/microsoft_exchange_server/manifest.yml
+++ b/packages/microsoft_exchange_server/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.0.3
 name: microsoft_exchange_server
 title: "Microsoft Exchange Server"
-version: "1.4.0"
+version: "1.4.1"
 source:
   license: "Elastic-2.0"
 description: Collect logs from Microsoft Exchange Server with Elastic Agent.


packages/microsoft_exchange_server/data_stream/httpproxy/agent/stream/filestream.yml.hbs
```diff
@@ -15,7 +15,7 @@ prospector.scanner:
   {{scanner}}
 {{/if}}
 processors:
-  - add_locale: ~
+- add_locale: ~
```
gogochan, line 18: seems like it would be more consistent to indent line 20. But if it works.

intxgo: I was thinking the same. I've checked yaml validator, it's allowed to have:
```
some_indented_list:
  - first
  - second

some_list:
- first
- second
```
but indeed looks like `{{processors}}` could be indented to match the rest of the document, see `{{scanner}}` above

andrewkroh:
Adding a policy test that includes custom `processors` in its configuration would help to prove the fix is effective (and does not regress in the future).

https://github.com/elastic/elastic-package/blob/f6341f05f699e34683b765cce8c1bad923b3edc3/docs/howto/policy_testing.md

andrewkroh:
The other agent yml.hbs templates in the same packages should also be fixed if they suffer the same problem.

packages/microsoft_exchange_server/data_stream/httpproxy/agent/stream/filestream.yml.hbs
nfritts: ironically it can't.  The whitespace gets stripped out due to the way it gets processed.   Indenting things in the template ends up not working.

marc-gr:
I added the tests and the same changes to the remaining data streams

PR #16181, opened 2025-12-01
Initial filelog input package for OTel collector

jsoriano:
Initial filelog input package for OTel collector. I am mainly adding it because I need a generic log collector for "Composable" packages based on OTel technologies.

## Proposed commit message

Add an initial package for the filelog receiver of the OTel collector.

## Checklist

Install and configure the integration with an agent that has access to some logs. It should collect the logs when configured for it.

### DIFF
--- a/packages/filelog_otel/manifest.yml
+++ b/packages/filelog_otel/manifest.yml
@@ -1,6 +1,6 @@
 format_version: 3.5.0
 name: filelog_otel
-title: "Collect logs from files"
+title: "File Log OpenTelemetry input"
 version: 0.1.0
 source:
   license: "Elastic-2.0"


packages/filelog_otel/manifest.yml
```diff
@@ -0,0 +1,73 @@
...
+        show_user: false
+        title: Whether to add the file path as the attribute log.file.path.
+        type: bool
+        default: false
+      - name: poll_interval
+        required: false
+        show_user: false
+        title: The duration between filesystem polls.
+        type: duration
+        default: 200ms
```
jsoriano, line 69: Many more options are available (see [docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/filelogreceiver/README.md)), I am adding by now the ones I think I am going to need for the most basic use cases.

packages/filelog_otel/docs/README.md
```diff
@@ -0,0 +1,73 @@
...
+### How it works
+
+This integration tails and parses logs from files. Observed files are the ones
+matching the configured patterns
+
+## What data does this integration collect?
+
+
+
+Key fileds are stored folowing [Semantic Conventions for logs](https://opentelemetry.io/docs/specs/semconv/general/logs/), among them:
```
mrodm:
```suggestion
Key fields are stored following [Semantic Conventions for logs](https://opentelemetry.io/docs/specs/semconv/general/logs/), among them:
```

packages/filelog_otel/docs/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+Elastic Agent must be installed. For more details, check the Elastic Agent [installation instructions](docs-content://reference/fleet/install-elastic-agents.md). You can install only one Elastic Agent per host.
+
+Elastic Agent is required to have permissions to read the monitored files.
+
+
+
+### Onboard / configure
+
+This integration needs to be configured with the glob patters of the files to
```
mrodm:
```suggestion
This integration needs to be configured with the glob patterns of the files to
```

packages/filelog_otel/_dev/build/docs/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+### How it works
+{{/* Add a high level overview on how this integration works. For example, does it collect data from API calls or recieving data from a network or file.*/}}
+This integration tails and parses logs from files. Observed files are the ones
+matching the configured patterns
+
+## What data does this integration collect?
+{{/* Complete this section with information on what types of data the integration collects, and link to reference documentation if available */}}
+
+{{/* TODO */}}
```
mrodm: Should this TODO be removed? 
```suggestion
```

packages/filelog_otel/docs/README.md
```diff
@@ -0,0 +1,73 @@
...
+logs from files using the [File Log
+Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/filelogreceiver/README.md)
+of the OTel Collector.
+
+This integration can be used to collect logs from files.
+
+### How it works
+
+This integration tails and parses logs from files. Observed files are the ones
+matching the configured patterns
```
mrodm:
```suggestion
matching the configured patterns.
```

packages/filelog_otel/_dev/test/system/test-filelog-custom-dataset-config.yml
```diff
@@ -0,0 +1,8 @@
+service: logfile
+vars:
+  data_stream.dataset: elastic_agent.test
```
mrodm, line 3: Should this `data_stream.dataset` variable be added to the manifest too?
I think it should be used just for testing and it should not be added there, but double checking.

mrodm (approved):
LGTM
Just a minor question and suggestions for the readme files.

packages/filelog_otel/agent/input/input.yml.hbs
```diff
@@ -0,0 +1,21 @@
+receivers:
+  filelog:
+    include:
```
ishleenk17, line 3: @jsoriano : Since there are many options here and same is true for other input packages.
We were thinking that it might be a good idea to have the whole receiver configuration open for users to enter. 
that means a plain yaml box. WDYT ? 
Since this is an input package.

packages/filelog_otel/docs/README.md
```diff
@@ -0,0 +1,73 @@
+<!-- NOTICE: Do not edit this file manually.-->
+<!-- This file is automatically generated by Elastic Package -->
+
+# Custom Logs Integration for Elastic based on the OTel collector
```
ishleenk17: We should not use the word integration for input package. 
This would confuse the users. 
We can use the word integration once composable integration packages comes into picture.

packages/filelog_otel/docs/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Overview
+
+The Custom Logs integration for Elastic based on the OTel collector collects
```
ishleenk17: This is incorrect

packages/filelog_otel/docs/README.md
```diff
@@ -0,0 +1,73 @@
...
+|-------------|-------------|
+| message | Collected log line. |
+| log.file.name | File name. |
+| log.file.path | File path. |
+
+### Supported use cases
+
+This integration can be used to collect custom log files.
+
+## How do I deploy this integration?
```
ishleenk17, line 33: I think we can skip this section in the documentation and keep the documentation very basic

packages/filelog_otel/docs/README.md
```diff
@@ -0,0 +1,73 @@
...
+### Validation
+
+Once configured, you can find documents with the content of the collected log files.
+
+## Troubleshooting
+
+For help with Elastic ingest tools, check [Common problems](https://www.elastic.co/docs/troubleshoot/ingest/fleet/common-problems).
+
+
+## Scaling
```
ishleenk17, line 57: Scaling, troubleshooting. These sections look irrelevant to the input package documentation

jsoriano: These sections come from the consolidation efforts driven by the integrations experience team cc @haetamoudi.

packages/filelog_otel/_dev/test/system/test-filelog-custom-dataset-config.yml
jsoriano: This variable is provided by Fleet for input packages, it doesn't need to be defined in manifest unless one wants to modify the default value.

We don't really need to have this test here, I only copied it from the filestream input, we added it there to ensure that dataset customization kept working after migrating it from integration package.

packages/filelog_otel/agent/input/input.yml.hbs
jsoriano: I don't know. I would prefer to have the settings defined, validated and documented. This also allows us to limit what we support, control defaults and so on. Though I understand how this may be cumbersome for inputs with many options.

packages/filelog_otel/docs/README.md
ishleenk17: Thats right. But something which applies to regular integration package would not apply to an input package or content package.

packages/filelog_otel/docs/README.md
jsoriano: Do you refer to the use of the word "integration"?

packages/filelog_otel/docs/README.md
jsoriano: I think scaling and troubleshooting sections can be also useful in the context of input packages, as users can still use them individually.
There is no much content here yet, but I would still keep the sections for consistency and to link to the general documentation.

packages/filelog_otel/docs/README.md
ishleenk17: No, custom logs.
The package name says filelog_otel. And here we use custom logs. These should be same. 
Also I think we should have a better package name, like FileLog Input Package. 
Any Input Package should have the name Input in it, IMO. WDYT ?

jsoriano: Ah yes, this is precisely why I am using the long form of "... based on the OTel Collector". Do you have other suggestion?

ishleenk17: FileLog OpenTelemetry Input Package, WDYT ?

jsoriano: Ok, changed to "File Log OpenTelemetry input"

packages/filelog_otel/docs/README.md
haetamoudi: > These sections come from the consolidation efforts driven by the integrations experience team

Thanks for that 🙌 
FYi if some sections do not seem relevant to a specific integration, you can ignore them in the validation.yaml (if you are using the parameter `docs_structure_enforced `)

packages/filelog_otel/manifest.yml
```diff
@@ -0,0 +1,73 @@
+format_version: 3.5.0
+name: filelog_otel
+title: "Collect logs from files"
```
mrodm: As this title is going to be shown in the integrations UI, should the title be the same as in the title of the README?

```suggestion
title: "File Log OpenTelemetry input"
```

image

jsoriano: Thanks

packages/filelog_otel/agent/input/input.yml.hbs
ishleenk17: Also, since many of the receivers are in alpha currently, these options could change. If they do it would become a breaking change for us. Keeping the flexibility of having everything entered by the user in input package, keeps us agnostic to any changes happening to receivers upstream.

teresaromero (approved):

packages/filelog_otel/agent/input/input.yml.hbs
jsoriano: > Also, since many of the receivers are in alpha currently, these options could change. If they do it would become a breaking change for us. Keeping the flexibility of having everything entered by the user in input package, keeps us agnostic to any changes happening to receivers upstream.

Sorry, missed this comment.

In any case I would say this argument is valid for both sides.

If options in the receivers change and we define them as package variables, we can handle the breaking change, for example if the setting is renamed, we could keep the variable name, and rename it only in the template.

On the other hand, if options in the receivers change, and users need to use these options through plain yaml, they will need to handle the breaking change.

So regarding this argument, it is a matter of who is handling these potential breaking changes: developers or users. And I think I prefer if developers handle this in packages.

PR #16182, opened 2025-12-01
SKU based chargeback

sholzhauer-es:
## Proposed commit message
Adding SKU's to the `billing_cluster_cost_lookup` using the transform and ingest pipeline to parse into a `cost_type`.
This is preparation for [supporting all the cost allocation](https://github.com/elastic/elasticsearch-chargeback/issues/8)


sholzhauer-es:
doing this in anticipation of needing it for the node type cost insights. That way transforms can already start adding the info.
Problem

sholzhauer-es:
@JohannesMahne do you agree to add this in already or wait for the full thing to be ready?

JohannesMahne:
I agree. Let me merge it, and then redo my PR. I did not see this before.

PR #16184, opened 2025-12-02
snyk: add rate limit handling to audit_logs and issues data streams

### LINKED ISSUE #15608: snyk: enhance snyk integration to properly handle rate limits
The snyk audit_logs integration does not currently implement any rate limit handling. This should be added to the integration.

depends: elastic/beats#46997

efd6:
## Proposed commit message

```
snyk: add rate limit handling to audit_logs and issues data streams

This enhancement implements comprehensive rate limiting for Snyk API calls
to prevent quota exhaustion and improve collection reliability.
```

## Checklist

## Author's Checklist

- [ ] 

The system tests exercise the rate limit handling path and you will see this in the pattern of event publications in the issues data stream when tests are run with `-v`.

To see the rate limit values and the extracted headers, make the following changes to the CEL programs

<details>

```diff
diff --git a/packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs b/packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
index 2a3e949a55..c3270bfd5b 100644
--- a/packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
+++ b/packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
@@ -124,14 +124,14 @@ program: |-
                 // Threading rate-limit results obtained here through to the final
                 // result is not tenable, so we do not do any work to allow them
                 // to be interpreted by the input and just drop them.
-                rate_limit(
+                debug("RATE_LIMIT_1", rate_limit(
                   headers,
                   "X-Ratelimit",
                   false,
                   true,
                   duration(string(headers[?"X-Ratelimit-Reset"][0].orValue("1")) + "s"),
                   0
-                )
+                ))
               ).drop(["rate", "next"])
             )
           ).as(resp, (resp.StatusCode == 200) ?
@@ -198,14 +198,14 @@ program: |-
           resp.with(
             resp.Header.as(headers,
               // Calculate and apply rate limits.
-              rate_limit(
+              debug("RATE_LIMIT_2", rate_limit(
                 headers,
                 "X-Ratelimit",
                 false,
                 true,
                 duration(string(headers[?"X-Ratelimit-Reset"][0].orValue("1")) + "s"),
                 0
-              )
+              ))
             ).as(rate_headers,
               {
                // Rate limit side-effects have already been applied to the
diff --git a/packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs b/packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
index dfeccaa246..3f1d0ff8c3 100644
--- a/packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
+++ b/packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
@@ -134,14 +134,14 @@ program: |-
         resp.with(
           resp.Header.as(headers,
             // Calculate and apply rate limits.
-            rate_limit(
+            debug("RATE_LIMIT_1", rate_limit(
               headers,
               "X-Ratelimit",
               false,
               true,
               duration(string(headers[?"X-Ratelimit-Reset"][0].orValue("1")) + "s"),
               0
-            )
+            ))
           ).as(rate_headers,
               {
                // Rate limit side-effects have already been applied to the
@@ -195,14 +195,14 @@ program: |-
                                     // Threading rate-limit results obtained here through to the final
                                     // result is not tenable, so we do not do any work to allow them
                                     // to be interpreted by the input and just drop them.
-                                    rate_limit(
+                                    debug("RATE_LIMIT_2", rate_limit(
                                       headers,
                                       "X-Ratelimit",
                                       false,
                                       true,
                                       duration(string(headers[?"X-Ratelimit-Reset"][0].orValue("1")) + "s"),
                                       0
-                                    )
+                                    ))
                                   ).drop(["rate", "next"])
                                 )
                               ).as(resp, (resp.StatusCode == 200) ?
```

</details>

And examine the agent logs (either via docker logs or by getting a diagnostic).

The `RATE_LIMIT_1` and `RATE_LIMIT_2` tags will appear in the agent debug logs with the computed `rate`, `burst`, `next`, and `reset` fields from the rate limiter. Check that the values of these fields are sane relative to the HTTP headers which will also be present in the debug output in the message.

## Related issues

- Closes #15608

### DIFF
--- a/packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
+++ b/packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
@@ -117,7 +117,7 @@ program: |-
               "limit": ["100"],
             }.format_query()
           ).with(auth_header).do_request().as(resp,
-            // We are doing this work for the side-effects of the rate_limit call.
+            // We are doing this work purely for the side-effects of the rate_limit call.
             resp.with(
               // Calculate and apply rate limits.
               // Threading rate-limit results obtained here through to the final
@@ -192,7 +192,7 @@ program: |-
             }.format_query()
           )
         ).with(auth_header).do_request().as(resp,
-          // We are doing this work for the side-effects of the rate_limit call.
+          // We are doing this work mainly for the side-effects of the rate_limit call.
           resp.with(
             // Calculate and apply rate limits.
             rate_limit(
--- a/packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
+++ b/packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
@@ -130,7 +130,7 @@ program: |-
           }.format_query()
         )
       ).with(auth_header).do_request().as(resp,
-        // We are doing this work for the side-effects of the rate_limit call.
+        // We are doing this work mainly for the side-effects of the rate_limit call.
         resp.with(
           // Calculate and apply rate limits.
           rate_limit(
@@ -186,7 +186,7 @@ program: |-
                               get_request(
                                 state.url.trim_right("/") + "/rest" + issue.relationships.scan_item.links.related + "?" + {"version": [state.version]}.format_query()
                               ).with(auth_header).do_request().as(resp,
-                                // We are doing this work for the side-effects of the rate_limit call.
+                                // We are doing this work purely for the side-effects of the rate_limit call.
                                 resp.with(
                                   // Calculate and apply rate limits.
                                   // Threading rate-limit results obtained here through to the final


packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
```diff
@@ -47,156 +47,243 @@ redact:
...
+              // we'll initially always limit to 100. In future we can paginate
+              // over multiple requests if there is demand, possibly with a
+              // configurable limit.
+              "limit": ["100"],
+            }.format_query()
+          ).with(auth_header).do_request().as(resp,
+            // We are doing this work for its side-effects.
+            resp.with(
+              // Canonicalise header keys to match rate_limit conventions.
+              resp.Header.transformMapEntry(k, v, {k.canonical_mime_header_key(): v}).as(headers,
```
chrisberkhout: Is this definitely necessary? I thought `net/http` would canonicalize by default.

packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
```diff
@@ -47,156 +47,243 @@ redact:
...
+                "last_created": body.data.items.map(item,
+                  has(item.created),
+                  timestamp(item.created)
+                ).as(times, (size(times) == 0) ? item.?last_created.orValue(now) : times.max()),
+              },
+              "want_more": has(body.?links.next),
+            }
+          )
+        : (resp.StatusCode == 429) ?
+          // We were not able to avoid a rate limit excession.
```
chrisberkhout, line 260: excession - I guess I need to read more sci fi

packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
```diff
@@ -47,156 +47,243 @@ redact:
...
+              "version": [state.version],
+              // The /rest/orgs endpoint returns between 10 and 100 results per
+              // request (multiples of 10 only), defaulting to 10. For simplicity
+              // we'll initially always limit to 100. In future we can paginate
+              // over multiple requests if there is demand, possibly with a
+              // configurable limit.
+              "limit": ["100"],
+            }.format_query()
+          ).with(auth_header).do_request().as(resp,
+            // We are doing this work for its side-effects.
```
chrisberkhout:
```suggestion
            // We are doing this work for the side-effects of the rate_limit call.
```
Correct, right?

packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
```diff
@@ -47,156 +47,243 @@ redact:
...
+              ).as(rate_headers,
+                rate_headers.with(
+                  {
+                    // Work around inf detection in input.
+                    // If the headers are missing or rate_limit failed, rate and
+                    // next may be missing. So use optional types.
+                    ?"rate": (rate_headers.?rate == optional.of(double("Infinity"))) ? optional.of("inf") : optional.none(),
+                    ?"next": (rate_headers.?next == optional.of(double("Infinity"))) ? optional.of("inf") : optional.none(),
+                  }
+                )
```
chrisberkhout: This part will also only affect the debugging value, right?

I think it'd be better to combine it the the following by adding the `{"rate_limit": ...}` wrapper, so we have one block for side effects and one block for a debugging value.

chrisberkhout (commented):
What do you think about testing this? I think the system test could be modified to exercise the new logic, but that's probably not worthwhile.

I think it would be nice if the commit message or at least the PR's "How to test this PR locally" section had some advice about how to manually exercise it.

I'm thinking I'd run the system test's `stream` config outside of a container, with extra response headers, then modify the CEL program to run in miko/mito and add some debug() calls to inspect the debugging value.

packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
efd6: Yes

packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
efd6: I can't say that I have not seen it in sci-fi novels, but it did exist before February 1954.

packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
efd6: I agree. The reason I have it here is because I thought I had seen that it was necessary in the o365 case. I've looked again to confirm and I'll remove this.

efd6:
@chrisberkhout there is significant work to do to get this to merge due to conflicts. I'll ping you when it's ready.

packages/snyk/data_stream/audit_logs/agent/stream/cel.yml.hbs
efd6: Not quite, but you have identified a nest of bugs that are addressed in 94b456a.

packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
```diff
@@ -144,26 +143,29 @@ program: |-
               0
             )
           ).as(rate_headers,
```
chrisberkhout: Could use `resp.Header` directly rather than as `headers`.

packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
```diff
@@ -144,26 +143,29 @@ program: |-
...
+               // limiter. This value is propagated to the final output for
+               // handleResponse to extract a waitUntil time when the
+               // quota is zero.
+               "rate_limit": rate_headers.with(
+                  {
+                    // Work around inf detection in input.
+                    // If the headers are missing or rate_limit failed, rate and
+                    // next may be missing. So use optional types.
+                    ?"rate": (rate_headers.?rate == optional.of(double("Infinity"))) ? optional.of("inf") : optional.none(),
+                    ?"next": (rate_headers.?next == optional.of(double("Infinity"))) ? optional.of("inf") : optional.none(),
```
chrisberkhout: Just thinking...

Looks like this checks for an Infinity valued double and switches it to the string "inf", which in the input will be switched back to `rate.Inf`, which is the finite `math.MaxFloat64` rather than infinite.

Seems like it could have been made simpler, especially in `x/time/rate`.

But given that's the way it is, maybe at some point `rate_limit` should handle avoiding infinite rates?

packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
```diff
@@ -188,9 +190,11 @@ program: |-
                               ).with(auth_header).do_request().as(resp,
                                 // We are doing this work for the side-effects of the rate_limit call.
                                 resp.with(
-                                  // Canonicalise header keys to match rate_limit conventions.
-                                  resp.Header.transformMapEntry(k, v, {k.canonical_mime_header_key(): v}).as(headers,
+                                  resp.Header.as(headers,
                                     // Calculate and apply rate limits.
+                                    // Threading rate-limit results obtained here through to the final
+                                    // result is not tenable, so we do not do any work to allow them
+                                    // to be interpreted by the input and just drop them.
```
chrisberkhout: Just thinking about the existing structure of the CEL program...

Doing multiple requests per eval means you have to reduce the results with complicated logic and accept compromises regarding the feedback that can go back to the input. This is a case of that.

I think the work list approach like we have now in o365 is easier to reason about and can provide better feedback and error handling. The downside is the overhead of extra evals, but I think it's usually worth it. Or maybe I'm missing something?

chrisberkhout (changes requested):
Looks good.

I put a few comments that don't need any response.

One thing left is that the "We are doing this work for the side-effects of the rate_limit call" comments are a bit misleading in the cases where the values are added to the response and returned in the final result.

Maybe some could be changed to:

> We are doing this work only for the side-effects of the rate_limit call.

and the others could be removed or changed to something like:

> The rate_limit call has side effects as well as generating values for the final result.

packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
efd6: Much clearer. Done

packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
efd6: Yeah, this is unfortunate history. It's a conflict between between encoding/json, mito/lib, x/time/rate and cel-go. The origin of the "inf" is from JSON serialisation of `math.Inf(1)` that results from divisions. We could conceivably condition the return values from the rate limit extensions so that if either of these end up being infinite, it gets replaced with `rate.Inf`. I think this would be backwards compatible; the relevant text in the documentation is 'The map returned by the policy functions should have "rate" and "next" fields with type **rate.Limit or string with the value "inf"**, a "burst" field with type int and a "reset" field with type time.Time in the UTC location. The semantics of "rate" and "burst" are described in the documentation for the golang.org/x/time/rate package.' This would remain true, but the second option would never happen.

Do you want to file an issue in mito?

packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
efd6: I think this would be worthwhile. I also think it's work for another PR.

packages/snyk/data_stream/issues/agent/stream/cel.yml.hbs
chrisberkhout: :+1: https://github.com/elastic/mito/issues/120

PR #16185, opened 2025-12-02
Chargeback Integration: Extract deployment group from Billing tags

JohannesMahne:
## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/chargeback/elasticsearch/transform/billing_cluster_cost/transform.yml
+++ b/packages/chargeback/elasticsearch/transform/billing_cluster_cost/transform.yml
@@ -22,7 +22,7 @@ source:
           emit('');
 dest:
   index: billing_cluster_cost_lookup
-  pipeline: 0.2.4-billing
+  pipeline: 0.2.5-billing
 frequency: 60m
 sync:
   time:
@@ -59,4 +59,4 @@ _meta:
   run_as_kibana_system: false
   # Bump this version to delete, reinstall, and restart the transform during package.
   # Version bump is needed if there is any code change in transform.
-  fleet_transform_version: 0.2.4
+  fleet_transform_version: 0.2.5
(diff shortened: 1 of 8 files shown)


JohannesMahne:
The second commit is to add the deployment group expandable section with a table and chart, and also add explanations to the config section.

packages/chargeback/elasticsearch/transform/billing_cluster_cost/transform.yml
```diff
@@ -6,9 +6,23 @@ source:
     range:
       ess.billing.total_ecu:
         gt: 0
+  runtime_mappings:
```
sholzhauer-es, line 9: Why did you opt to use runtime field instead of parsing it using the pipeline? I am thinking performance can be improved when parsing during ingest.

packages/chargeback/pre-setup.sh
```diff
@@ -4,13 +4,12 @@
 PUT chargeback_conf_lookup
 {
   "settings": { 
-    "index.mode": "lookup", 
-    "index.hidden": true
```
sholzhauer-es, line 8: removing the hidden so it can be used in the filter I am assuming?

packages/chargeback/elasticsearch/transform/billing_cluster_cost/transform.yml
JohannesMahne: I simply could not get the pipeline route to work. I tested the pipeline on the docs from the billing metrics records, and it managed to extract the tags correctly and create deployment_group, but the transform would simply not give any output. That is when I started thinking that, maybe, the transform happens, and then the pipeline applies afterwards. I have not confirmed this in the documentation, though...

packages/chargeback/pre-setup.sh
JohannesMahne: All of the other lookup indices are not hidden anymore, and it is just troublesome to work with it. Also, we tell the user to update the config lookup, so, why keep it hidden?

PR #16175, opened 2025-12-01
ti_google_threat_intelligence: Set destination pipeline name automatically in transforms

kcreddy:
## Proposed commit message
```
ti_google_threat_intelligence: Set destination pipeline name automatically in transforms

Due to an earlier limitation of elastic-package to unable to extract
integration version into the transform, the destination pipeline 
needed to be set/updated manually.

Currently the integration requires users to manually add destination 
pipelines into the transform during the package installation and 
upgrade. This presents significant inconvenience to users and is 
often error-prone.

elastic-package#2973 [1] removes the dest pipeline limitation 
which now allows the pipeline names to be templated. This PR
updates the transform definition to set the pipeline names using 
a template which renders integration version.
Change involves updating all transforms with:
dest:
  pipeline: '{{ ingestPipelineName "<pipeline-name>"}}'

Also:
- With this the transform's does not require any manual 
intervention, hence they are now auto-enabled.
- README is updated accordingly.

[1] https://github.com/elastic/elastic-package/pull/2973
```

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally

- Integration successfully installed.
- System tests pass. 
- Transforms definition in Kibana show the correct destination pipeline (see screenshots).

**System tests (sample data stream)**

```
--- Test results for package: ti_google_threat_intelligence - START ---
╭───────────────────────────────┬─────────────┬───────────┬───────────┬────────┬──────────────╮
│ PACKAGE                       │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT │ TIME ELAPSED │
├───────────────────────────────┼─────────────┼───────────┼───────────┼────────┼──────────────┤
│ ti_google_threat_intelligence │ cryptominer │ system    │ default   │ PASS   │  45.2311125s │
╰───────────────────────────────┴─────────────┴───────────┴───────────┴────────┴──────────────╯
[...]

PR #16187, opened 2025-12-02
[ti_threatconnect] Fix the ingest pipeline failure

brijesh-elastic:
## Proposed commit message

```
ti_threatconnect: fix ingest pipeline failure when attributes field is missing and set
destination pipeline name automatically in transforms

The foreach processor in the ingest pipeline now uses ignore_missing: true
to handle cases where threat_connect.indicator.associated_groups.data but
lacks the attributes field, preventing pipeline failures during data
processing.

elastic-package#2973 [1] removes the dest pipeline limitation 
which now allows the pipeline names to be templated. This PR
updates the transform definition to set the pipeline names using 
a template which renders integration version.

[1] https://github.com/elastic/elastic-package/pull/2973
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/ti_threatconnect directory.
> elastic-package test

## Related issues

- Closes https://github.com/elastic/integrations/issues/16056
- Closes https://github.com/elastic/integrations/issues/16044

### DIFF
--- a/packages/ti_threatconnect/elasticsearch/transform/latest/transform.yml
+++ b/packages/ti_threatconnect/elasticsearch/transform/latest/transform.yml
@@ -9,7 +9,7 @@ source:
 # us that ability in order to prevent having duplicate IoC data and prevent query
 # time field type conflicts.
 dest:
-  index: "logs-ti_threatconnect_latest.dest_indicator-9"
+  index: "logs-ti_threatconnect_latest.dest_indicator-8"
   pipeline: '{{ ingestPipelineName "tactics_compatibility" }}'
   aliases:
     - alias: "logs-ti_threatconnect_latest.indicator"


efd6 (commented):
I think that we should add a test that has `threat_connect.indicator.associated_groups.data` with an element that lacks `attributes`.

Rémy suggests the following commit message:
```
ti_threatconnect: fix ingest pipeline failure when attributes field is missing

The foreach processor in the ingest pipeline now uses ignore_missing: true
to handle cases where threat_connect.indicator.associated_groups.data but
lacks the attributes field, preventing pipeline failures during data
processing.
```
(with modification)

brijesh-elastic:
> I think that we should add a test that has `threat_connect.indicator.associated_groups.data` with an element that lacks `attributes`.

The test case is already in `test-indicator.log`, and I am still trying to understand why it's not failing locally, even if I remove the `ignore_missing: true`.

brijesh-elastic:
> I am still trying to understand why it's not failing locally, even if I remove the `ignore_missing: true`.

It was failing in serverless. I tried setting the serverless environment, and its failing when we remove `ignore_missing: true`

packages/ti_threatconnect/elasticsearch/transform/latest/transform.yml
```diff
@@ -10,7 +10,7 @@ source:
 # time field type conflicts.
 dest:
   index: "logs-ti_threatconnect_latest.dest_indicator-8"
-  pipeline: "1.10.2-tactics_compatibility"
+  pipeline: "1.10.3-tactics_compatibility"
```
kcreddy:
```suggestion
  pipeline: "{{ ingestPipelineName "tactics_compatibility" }}"
```

See https://github.com/elastic/integrations/pull/16175

kcreddy:
> > I am still trying to understand why it's not failing locally, even if I remove the `ignore_missing: true`.
> 
> It was failing in serverless. I tried setting the serverless environment, and its failing when we remove `ignore_missing: true`

Failing on `9.3.0-SNAPSHOT` locally. Isn't this problem similar to https://github.com/elastic/elasticsearch/pull/126417?

packages/ti_threatconnect/elasticsearch/transform/latest/transform.yml
brijesh-elastic: Update in [commit](https://github.com/elastic/integrations/pull/16187/commits/e37be207b46a0ebd4e3518238190d49717e93b51).

brijesh-elastic:
> Failing on `9.3.0-SNAPSHOT` locally. Isn't this problem similar to [elastic/elasticsearch#126417](https://github.com/elastic/elasticsearch/pull/126417)?

Yes, Looks like it.

packages/ti_threatconnect/elasticsearch/transform/latest/transform.yml
```diff
@@ -9,8 +9,8 @@ source:
 # us that ability in order to prevent having duplicate IoC data and prevent query
 # time field type conflicts.
 dest:
-  index: "logs-ti_threatconnect_latest.dest_indicator-8"
-  pipeline: "1.10.2-tactics_compatibility"
+  index: "logs-ti_threatconnect_latest.dest_indicator-9"
```
kcreddy: Why do we need to increment index number?

brijesh-elastic: Well, its not required. I did it to keep up to date with transform version.

kcreddy: If users need data from previous index, it would require them to copy. I think we should avoid unnecessarily incrementing DEST suffixes. Suffixes was originally added for supporting breaking changes.

kcreddy (approved):
LGTM. Please update the commit message to reflect the transform change. Thanks!

PR #16222, opened 2025-12-03
[Tenable_IO] - Enriched the scan data stream with WAS API scan details data.

ShourieG:
## Type of change
- Enhancement

## Proposed commit message
```
tenable_io: enrich scan data stream with detailed scan information

The previous implementation only collected basic scan metadata. This change
adds a worklist pattern to fetch detailed scan information for each scan,
providing richer data for security analysis.

API documentation:
- List Scans: https://developer.tenable.com/reference/scans-list
- Scan details: https://developer.tenable.com/reference/was-v2-scans-details
   
The test for enrichment was created using the public example of scan-details available 
on https://developer.tenable.com/reference/scans-list and integrating this with an existing
list scans test.
```

## Checklist

## Author's Checklist
The tests used have been derived from the existing tests and examples listed on https://developer.tenable.com/reference/scans-list

- [ ] 

## Related issues

- Closes https://github.com/elastic/enhancements/issues/25737


packages/tenable_io/data_stream/scan/agent/stream/cel.yml.hbs
```diff
@@ -22,36 +22,82 @@ redact:
...
+            },
+            "want_more": false
+          }
+      )
+  ).as(state,
+    // Process first scan in worklist
+    (has(state.worklist) && size(state.worklist) > 0) ?
+      state.worklist[0].as(scan,
+        state.with(
+          request("GET", state.url.trim_right("/") + "/was/v2/scans/" + string(scan.id)).with({
```
efd6, line 64: Is there a documentation link for the /was/v2/scans endpoint?

ShourieG: Yes, I've updated the description to include the [link](https://developer.tenable.com/reference/was-v2-scans-details). I think there is one more enhancement that I need to do, which is ignore archived details for now. Will update the documentation accordingly.

efd6: I think you should include both since there are both being used.

ShourieG: @efd6, you can ignore the comment on the scan archival, I mixed it up with the [vul management scan details api](https://developer.tenable.com/reference/scans-details). This has the archival limit. The `was-v2-scans-details` api used here, does not have any such limit. So not changing anything else here. If everything else is ok, we should be good to merge.

Issue #16186, opened 2025-12-02
[akamai.siem]: Pagination logic causes excessive API requests


andrewkroh:
### Integration Name

Akamai [akamai]

### Dataset Name

akamai.siem

### Integration Version

3.0.0

### Agent Output Type

elasticsearch

### Software/API Version

Akamai SIEM API v1

### What did you do?

Configured Akamai SIEM integration to collect security events from the API.

### What did you see?

The integration uses incorrect logic to determine when to continue paginating through API responses. The CEL program at [packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs:112-117](https://github.com/elastic/integrations/blob/6c676bcfa7a0425e11b93662388d15ccc5f0fab5/packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs#L112-L115) checks whether the last event has a non-empty `offset` field to set `want_more`:

```cel
"want_more": (lines.size() > 0) ?
  lines[lines.size() - 1].decode_json().as(lastEvent,
    has(lastEvent.offset) && lastEvent.offset != ""
  )
:
  false,
```

This causes an issue where pagination may continue unnecessarily when no more data is available, causing wasted API calls.

### What did you expect to see?

According to the Akamai SIEM API documentation, the response includes a metadata object on the last line with a `limit` field that "appears if the size limit was reached during data fetch." This is the correct indicator for whether more data is available.

The pagination logic should check for the presence of the `limit` field instead:

```cel
"want_more": (lines.size() > 0) ?
  lines[lines.size() - 1].decode_json().as(lastEvent,
    has(lastEvent.limit)
  )
:
  false,
```

### Anything else?

The `offset` field is present on every response metadata line regardless of whether more data is available, making it an unreliable indicator for pagination. Only the `limit` field's presence definitively signals that the API has more data to fetch.

https://techdocs.akamai.com/siem-integration/reference/get-configid

PR #16225, opened 2025-12-03
[Akamai SIEM] - Fixed pagination logic to use the number_of_events_returned >= event_limit  conditional check instead of offset presence

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
akamai: fix pagination logic to use  number_of_events_returned >= event_limit 
conditional check instead of offset presence.

The integration was incorrectly using the offset field for pagination,
which caused issues with data retrieval. The Akamai SIEM API requires
checking the condition number_of_events_returned >= event_limit for 
proper pagination control.

Documentation: https://techdocs.akamai.com/siem-integration/reference/offsetcontext
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues
- Closes https://github.com/elastic/integrations/issues/16186

### DIFF
--- a/packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
+++ b/packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
@@ -109,12 +109,7 @@ program: |-
                           optional.none(),
                         "recovery_mode": false,
                       },
-                      "want_more": (lines.size() > 0) ?
-                        lines[lines.size() - 1].decode_json().as(lastEvent,
-                          has(lastEvent.total) && int(lastEvent.total) > 0
-                        )
-                      :
-                        false,
+                      "want_more": lines.size() >= int(state.event_limit),
                     }
                   )
                 : (resp.StatusCode == 416) ?
--- a/packages/akamai/changelog.yml
+++ b/packages/akamai/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "3.0.1"
   changes:
-    - description: Fixed pagination logic to use the conditional check `total > 0` instead of relying on the offset field's presence.
+    - description: Fixed pagination logic to use the conditional check `number of events returned >= event_limit` instead of relying on the offset field's presence.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/16225
 - version: "3.0.0"


packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
```diff
@@ -111,7 +111,7 @@ program: |-
                       },
                       "want_more": (lines.size() > 0) ?
                         lines[lines.size() - 1].decode_json().as(lastEvent,
-                          has(lastEvent.offset) && lastEvent.offset != ""
+                          has(lastEvent.limit)
```
efd6: The docs say "Appears if the size limit was reached during data fetch", but also say that it's required. These seem inconsistent to me. Do you have any clarity on that?

andrewkroh: I had not noticed the OpenAPI spec marking the field as [required](https://github.com/akamai/akamai-apis/blob/85ee42aa8cd3985b116800a7a8830f6d3d79ba78/apis/siem/v1/openapi.json#L1028). I personally do not have any saved request trace logs from the real server to validate that [statement](https://github.com/akamai/akamai-apis/blob/85ee42aa8cd3985b116800a7a8830f6d3d79ba78/apis/siem/v1/openapi.json#L1013) of "Appears if the size limit was reached during data fetch."

Does anyone else have access to such information?

An alternative approach to stopping pagination would be to look at the item count in the response to see if it is less than the requested `limit`.

ShourieG: Another approach  would be to use the `lastEvent.total > 0` condition. The httpjson implementation used a similar logic ` value: '[[ if not (eq (toInt .last_event.total) 0) ]][[ .last_event.offset ]][[ end ]]'`, since total is always number of records fetched, if it becomes 0, then we can effectively say pagination has finished.

ShourieG: I updated the code to check for  `total > 0`, I feel this should work as per what's written in the documentation.

efd6: I think that's fair. Thanks

andrewkroh: When the API returns a batch that is not full (less than the limit), I would like it to stop making more requests. This gives the service time to buffer events so that we make the most efficient use of our API requests.

If it does `want_more: total > 0`, then it either "wastes" an additional API request to fetch an empty response before stopping. Or if you have a steady stream of events, then it never rests (just a few events per second could be enough to make the input never stop).

ShourieG: Yea, this is definitely more efficient in the long run, will make this change. Thanks.

ShourieG:
@andrewkroh, made the suggested change, let me know if it's good to merge or not atm.

PR #16235, opened 2025-12-03
[Check Point] Process the packets field in SecureXL format

ilyannn:
## Proposed commit message

The specification at https://support.checkpoint.com/results/sk/sk144192 describes the `packets` field in the _Security Gateway - SecureXL Fields_ section as the tuple of 

- Source IP address
- Source Port
- Destination IP address
- Destination Port
- Protocol Number

The current pipeline, however, parses the `packets` field as an integer, handling only the format described in the _Security Gateway - Firewall Fields_ section. When this field comes as defined in the SecureXL section – a string describing the dropped packets – the current pipeline fails.

In practice the tuple can also contain an additional member denoting the network interface.

This PR adds the parsing to the pipeline. The result is an array of json objects, each structured according to the ECS, that is:

```
"packets_dropped" => [{
    "source": {
        "ip": ...
        "port": ...
    },
    "destination": {
        "ip": ...
        "port": ...
    },
    "network": {
        "iana_number": ...
    },
    "interface": {
        "name": ...
    }
}, ...]
```

We handle both the case when the interface name is present (as in practice) and when it's not (as per specification). The added group is defined as `nested` to preserve the relationships within each tuple. 

The parsing is implemented as a Painless script; alternative approaches were tried, but were not ultimately successful.

## Screenshot

From the specification:

CleanShot 2025-12-03 at 18 27 12@2x

## Checklist

## Author's Checklist

- [x] Add test cases for the new functionality
- [x] `elastic-package check`

## GenAI Note

Cursor with the Claude Opus 4.5 model was used in creating this PR's contents, especially the Painless script.

### DIFF
--- a/packages/checkpoint/manifest.yml
+++ b/packages/checkpoint/manifest.yml
@@ -1,6 +1,6 @@
 name: checkpoint
 title: Check Point
-version: "1.44.0"
+version: "1.45.0"
 description: Collect logs from Check Point with Elastic Agent.
 type: integration
 format_version: "3.0.3"


packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1018,6 +1018,29 @@ processors:
                   tag: remove_checkpoint_subs_exp_19a22c63
                   field: checkpoint.subs_exp
                   ignore_missing: true
+  - script:
+      tag: script_parse_checkpoint_packets_dropped
```
P1llus, line 1022: It would be preferred if we could use a common processor to parse this out, as script regex is quite heavy, was this the only way? looking at the data I guess it does seem like that, but its quite troublesome for performance reasons

packages/checkpoint/data_stream/firewall/fields/fields.yml
```diff
@@ -1187,6 +1187,47 @@
       type: integer
       description: |
         Amount of packets dropped.
+    - name: packets_dropped
+      type: nested
```
P1llus, line 1191: Just leaving it as a note rather than something that needs to be resolved, nested types are something we do try to avoid as much as possible, but at this point I am unsure about any other better alternative

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1018,6 +1018,29 @@ processors:
...
+      description: Parse packets field containing connection tuples into structured packets_dropped array.
+      if: ctx.checkpoint?.packets instanceof String && ctx.checkpoint.packets.contains('<')
```
P1llus: I am a bit unsure but it could be startsWith is better performance wise, though if the only times this is a string is when it is the SecureXL format it might be okay.
If we had sample data we could see if there was an already parsed field that showcases that it's SecureXL type data so we wouldn't need to use either

packages/checkpoint/data_stream/firewall/fields/fields.yml
ilyannn: Yeah it's a tradeoff. With nested types we accurately show the underlying structure, so that they can do "find packets where source is X and destination is Y", but it does come at a cost. The alternative could be to use a group but then we'll only have a list of all sources and a list of all destinations in the event and the information about which source goes to which destination would be lost.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
ilyannn: Good point. I don't know if their tuple format is well-documented, for example if it's guaranteed that it does not start with a whitespace. The cleanest way is probably to try a conversion to number first and only run the script if it fails – I'll do that.

packages/checkpoint/data_stream/firewall/fields/fields.yml
ilyannn: I don't know the details, but my assumption was that there is no performance penalty as long as nobody is searching on that field.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
ilyannn: I welcome any ideas 😄. I tried `split` + `grok` but couldn't make it work.

packages/checkpoint/data_stream/firewall/fields/fields.yml
P1llus: Yeah nested fields don't induce a performance penalty but you cannot use aggregations on them, so any dashboard visualization etc wont work, the only things you can do is querying for example on the interface name of any of the objects

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
P1llus: Yeah once I noticed it was meant to produce an array then any other processor mostly goes out the window with the exception of maybe foreach, but that just introduces many other issues, so script was the way to go, I was just hoping we could have managed to skip the regex part, as in theory the data is structured in some ways.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
ilyannn: It actually starts with a whitespace – ` <` – so I went with `trim().startsWith('<')`

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
ilyannn: I see. Yeah a review of https://www.elastic.co/docs/reference/scripting-languages/painless/painless-ingest-processor-context has shown to me how this is possible, so I've implemented a non-regexp version.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1018,6 +1018,39 @@ processors:
                   tag: remove_checkpoint_subs_exp_19a22c63
                   field: checkpoint.subs_exp
                   ignore_missing: true
+  - script:
+      tag: script_parse_checkpoint_packets_dropped
+      description: Parse packets field containing connection tuples into structured packets_dropped array.
+      if: ctx.checkpoint?.packets instanceof String && ctx.checkpoint.packets.trim().startsWith('<')
```
taylor-swanson, line 1024: As a note, if statements are also painless scripts, so could move this into the source of this processor and it should work just the same, basically one script invocation instead of two. I've never benchmarked it (and not sure if it matters, I don't think conditionals are included in benchmarks), but it's probably very minimal overhead, especially once the JIT has warmed up.

taylor-swanson (approved):
LGTM

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
ilyannn: Sure but this way we have a clear visual on when the script is run, I think it's moire readable this way.

PR #16240, opened 2025-12-04
crowdstrike: remove all statically set constant keyword fields in fdr at beginning of ingest

efd6:
## Proposed commit message
```
crowdstrike: remove all statically set constant keyword fields in fdr at beginning of ingest

In cases where an unrecoverable error occurs in the ingest pipeline, the
previous code would skip removal of ecs.version, resulting in a mapping
failure since beats sets this value to 8.0.0, which does not match the
static value. So remove this value at the beginning of the ingest
pipeline to avoid this. Also remove all other statically defined
constant keyword field values in case they have been set.

Also fix a rename processor failure that is uncovered by this change
where a condition attempts to access the ctx.crowdstrike?.User?.Name
field through a string ctx.crowdstrike?.User field, resulting in a
painless dynamic getter error.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/crowdstrike/data_stream/fdr/_dev/test/system/test-keep-metadata-config.yml
+++ b/packages/crowdstrike/data_stream/fdr/_dev/test/system/test-keep-metadata-config.yml
@@ -14,4 +14,4 @@ data_stream:
     keep_metadata: true
     enable_deduplication: true
 assert:
-  hit_count: 131
+  hit_count: 132


packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -2618,12 +2627,17 @@ processors:
       field: crowdstrike.User.Name
       target_field: user.name
       ignore_missing: true
-      if: ctx.crowdstrike?.User?.Name != null && ctx.user?.name == null
+      if: '!(ctx.crowdstrike?.User instanceof String) && ctx.crowdstrike?.User?.Name != null && ctx.user?.name == null'
```
andrewkroh:
```suggestion
      if: 'ctx.crowdstrike?.User instanceof Map && ctx.crowdstrike?.User?.Name != null && ctx.user?.name == null'
```

If ctx.crowdstrike.User were an array or another type without a Name field, attempting to read .Name could fail.

efd6: In practice, I don't think this could ever happen (at least, the documents that we see from CrowdStrike that have this field not be an object have it as a string, because they are userinfo objects and so are only talking about a single user). However, the defensive implementation is done.

Issue #16022, opened 2025-11-19
[mimecast.siem_logs] Refactor to reduce memory pressure


andrewkroh:
The Mimecast integration's CEL-based siem_log data stream can cause significant memory pressure, potentially leading to Elastic Agent OOM restarts. This is because the current implementation downloads and processes all available event batches from the Mimecast API within a single execution cycle. When a large number of events are available, this can lead to a spike in memory usage.

Version: Mimecast 3.2.1

### Problem

The CEL program for the `siem_logs` data stream does the following in a single execution:
1.  Fetches a list of batch URLs from the Mimecast API (`/siem/v1/batch/events/cg`).
2.  Downloads the gzipped NDJSON files from *all* of these URLs.
3.  Processes all the events from these files.

If the API returns many URLs, or the files are large, holding all the events in memory for processing can lead to memory pressure related issues (stalls, OOM kills, unresponsiveness to heartbeats.)

### Proposed Solution

To mitigate this, the CEL program should be refactored to process only one batch file per execution. This can be achieved by introducing a new work list in the cursor (e.g., `cursor.blobs`).

The refactored logic would be:
1.  If the `cursor.blobs` work list is not empty, take the first URL from the list, download and process that single batch file.
2.  If `cursor.blobs` is empty, call the Mimecast API to get a new list of batch URLs and populate the `cursor.blobs` list with them. Then, process the first URL in the list.

This change will significantly reduce the memory footprint of the integration by ensuring that only one batch of events is processed at a time.

### Proof of Concept

A proof-of-concept implementation of this approach is available in this branch: [https://github.com/andrewkroh/integrations/commits/mimecast-siem-batch-per-execution/](https://github.com/andrewkroh/integrations/commits/mimecast-siem-batch-per-execution/)

### Additional Context

A temporary workaround of making the `page_size` configurable was implemented in PR [elastic/integrations#15942](https://github.com/elastic/integrations/pull/15942). However, this is not an ideal long-term solution as it can lead to inefficient use of API tokens. This issue is for tracking the implementation of the more robust, long-term solution.

PR #16308, opened 2025-12-04
mimecast: refactor CEL code to prevent memory issues and OOM failures

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
mimecast: refactor CEL code to prevent memory issues and OOM failures

The existing CEL implementation in siem_logs data stream was causing
out-of-memory failures during processing. The refactored approach uses
batch processing techniques to reduce memory consumption.

The cloud_integrated_logs data stream received similar optimizations
due to code sharing with siem_logs.

Based on Andrew's POC implementation:
https://github.com/andrewkroh/integrations/commits/mimecast-siem-batch-per-execution/
```

## Checklist

## Note:

Skipped the httpjson test in 'siem_logs' as a fix for the health status degraded issue did not seem
fit for this current PR.

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/integrations/issues/16022

### DIFF
--- a/packages/mimecast/data_stream/siem_logs/_dev/test/system/test-v1-config.yml
+++ b/packages/mimecast/data_stream/siem_logs/_dev/test/system/test-v1-config.yml
@@ -11,6 +11,3 @@ data_stream:
     enable_request_tracer: true
 assert:
   hit_count: 1
-skip:
-  reason: "The fleet health status changes to degraded when the HTTPJSON template's value evaluation comes up empty or logs a failure, which leads to system test failures but does not interrupt the data flow."
-  link: https://github.com/elastic/beats/issues/45664


packages/mimecast/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "3.2.3"
+  changes:
+    - description: Refactored the CEL program for siem_logs data stream to improve memory usage.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/16308
+    - description: Refactored the CEL program for cloud_integrated_logs data stream to improve memory usage.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/16308
```
efd6:
```suggestion
    - description: Refactored the CEL program for cloud_integrated_logs and siem_logs data streams to improve memory usage.
      type: bugfix
      link: https://github.com/elastic/integrations/pull/16308
```

packages/mimecast/data_stream/cloud_integrated_logs/agent/stream/cel.yml.hbs
```diff
@@ -73,87 +77,98 @@ program: |
...
+                    "id": string(batch.Status),
+                    "message": "GET " + blobs[0] + ": " + (
+                      (size(batch.Body) != 0) ?
+                        string(batch.Body)
+                      :
+                        string(batch.Status) + " (" + string(batch.StatusCode) + ")"
+                    ),
+                  },
+                },
+                "want_more": false,
```
andrewkroh, line 130: I am wondering if we are going to get into a non-recoverable state if one of these blobs ages out (Mimecast claims to retain data for 7 days). Assuming the API returns a 404, maybe it should move to the next blob (send an error event, remove the item from cursor.blobs, continue to the next). WDYT?

packages/mimecast/data_stream/cloud_integrated_logs/agent/stream/cel.yml.hbs
```diff
@@ -73,87 +77,98 @@ program: |
...
-          }
+                "cursor": {
+                  "blobs": tail(blobs),
+                  "work_list": work_list,
+                },
+                "token": {
+                  "access_token": token.access_token,
+                  "expires": token.expires,
+                },
+                "want_more": size(tail(blobs)) != 0 || size(work_list) != 0,
```
efd6:
```suggestion
                "want_more": size(blobs) > 1 || size(work_list) != 0,
```
in both files

efd6 (commented):
Rémy suggests the following commit message (with my modification)
```
mimecast: refactor CEL code to prevent memory issues and OOM failures

The existing CEL implementation in siem_logs data stream was causing
out-of-memory failures during processing. The refactored approach uses
batch processing techniques to reduce memory consumption.

The cloud_integrated_logs data stream received similar optimizations
due to code sharing with siem_logs.

Based on Andrew's POC implementation:
[1] https://github.com/andrewkroh/integrations/commits/mimecast-siem-batch-per-execution/
```
I've replaced code "coupling" with "sharing" since there is no coupling.

packages/mimecast/data_stream/cloud_integrated_logs/agent/stream/cel.yml.hbs
ShourieG: I'm just concerned if a 404 is strictly enforced from mimecast's end in such a scenario or not. If a temporary issue causes a 404 then we will be acting upon a false positive. 

CEL already retries 5 times by default, what if we increment the `waitMin` and `waitMax` to a greater value and then on a **non 200**, we deterministically remove the faulty blob ? This way we keep a retry fall back and if it does not succeed, we effectively remove the blob from the list permanently and emit a error event, but continue to process the rest of the list.

andrewkroh: > on a non 200, we deterministically remove the faulty blob

I disagree with the **non-200**. I think that is too wide of a condition. I would limit it to HTTP 404 to begin with because this is the conventional status code for this state, but if we learn more, we can expand the status codes as required (e.g. [HTTP 409](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/410)).

ShourieG:
@andrewkroh, @efd6, I've made the suggested changes and updated the 404 error handling

PR #16312, opened 2025-12-04
[cisco_ise] Extract av-pairs into attributes mapping

taylor-swanson:
## Proposed commit message

- Improved script for extracting nested sub-keys from cisco-av-pair attribute/value pairs.
- Add pipeline test for passed authentications.
- Added mappings for known sub-keys to satisfy tests. Dynamic mappings is enabled for cisco_av_pair, so any unknown keys will be created on-demand.

## Checklist

```
cd packages/cisco_ise
elastic-package test
```


qcorporation (approved):
This looks good.
The only way that I can see this fail is if there's a third level nesting which I don't think is the case.

taylor-swanson:
> This looks good. The only way that I can see this fail is if there's a third level nesting which I don't think is the case.

```
cisco-av-pair=mdm-tlv=this=can=go=quite=deep=no-real-limit=device-platform-version=10.0.26100
```

yields (cisco-av-pair, mdm-tlv omitted because it wasn't shown in the test diff and I don't feel like adding it:

```
+                            "device-uid-global": "2897F9C5734F41A03F0E89064463F6FAB5EEDBF1",
+                            "this": {
+                                "can": {
+                                    "go": {
+                                        "quite": {
+                                            "deep": {
+                                                "no-real-limit": {
+                                                    "device-platform-version": "10.0.26100"
+                                                }
+                                            }
+                                        }
+                                    }
``` 
There is no imposed limit, it can handle any depth. I did try to test with escaped quotes in the key (since we talked about that), but it struggled with that (given how I'm detecting when the last key happens). I'm not really going to address that unless I see real evidence that there's a key with an escaped quote in it. It handles values with escaped quotes just fine.

Issue #16396, opened 2025-12-08
[Netskope] - Add options to override content-type for ABS input.


ShourieG:
We need to allow explicit content-type overrides in advanced configurations to address scenarios where content type is not set by default at source.

This is trivial in 9.1 stack and the required changes are in place to allow this to occur. 

For 8.19 stack we need to back-port the following PR's in beats:
 - https://github.com/elastic/beats/pull/44583
 - https://github.com/elastic/beats/pull/44873

The back-port was initially skipped as this re-work was a big overhaul to the code structure and code-flow resulting in more of an enhancement. Since 8.19 now has a very long support plan, we feel this back-port is necessary.

Post the back-port process we will add the configuration options to the integration

Until then if any-one is using the 9.1 stack wants to override the the content type in an integration using ABS input
please use a similar structure to the following under container settings directly:

```
- name: container_1
  override_content_type: true
  content_type: text/csv
``` 

If you want to use csv decoding in conjuction with content-overrides, then use:

```
- name: container_1
  override_content_type: true
  content_type: text/csv
  decoding.codec.csv.enabled: true,
  decoding.codec.csv.comma:   " ",
```

PR #16399, opened 2025-12-08
[Netskope] - Added advanced configuration options for setting and overriding content-type of log files.

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
netskope: add advanced configuration options for content-type handling

This allows users to set and override the content-type of log files when
the default detection is insufficient or incorrect, providing more
flexibility for processing various log formats.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/integrations/issues/16396
- Relates https://github.com/elastic/beats/pull/47964, https://github.com/elastic/beats/pull/47976

### DIFF
--- a/packages/netskope/data_stream/alerts_events_v2/manifest.yml
+++ b/packages/netskope/data_stream/alerts_events_v2/manifest.yml
@@ -135,7 +135,7 @@ streams:
         multi: false
         required: false
         show_user: false
-        description: Sets the content-type of the logs if there is no content-type specified in the file properties. If override_content_type is set to true, any existing content-type in the file properties will be overridden with the value of this attribute.
+        description: Sets the MIME content-type of the logs if there is no content-type specified in the file properties. If override_content_type is set to true, any existing content-type in the file properties will be overridden with the value of this attribute.
       - name: override_content_type
         type: bool
         title: Override Content Type
@@ -143,7 +143,7 @@ streams:
         required: false
         show_user: false
         default: false
-        description: Overrides existing content type of the logs with the value of the content_type attribute.
+        description: Overrides existing content-type of the logs with the value of the content_type attribute.
       - name: preserve_duplicate_custom_fields
         required: true
         show_user: false


packages/netskope/data_stream/alerts_events_v2/manifest.yml
```diff
@@ -129,6 +129,21 @@ streams:
...
+        show_user: false
+        description: Sets the content-type of the logs if there is no content-type specified in the file properties. If override_content_type is set to true, any existing content-type in the file properties will be overridden with the value of this attribute.
+      - name: override_content_type
+        type: bool
+        title: Override Content Type
+        multi: false
+        required: false
+        show_user: false
+        default: false
+        description: Overrides existing content type of the logs with the value of the content_type attribute.
```
efd6: Is it ever the case that the content type would be not set and the override content type would be true or vice versa? If not, can we make it so that a non-falsy content type automatically sets `override_content_type` to true?

packages/netskope/data_stream/alerts_events_v2/manifest.yml
```diff
@@ -129,6 +129,21 @@ streams:
         show_user: false
         default: " "
         description: The field separator character used by the CSV format.
+      - name: content_type
+        type: text
+        title: Content Type
+        multi: false
+        required: false
+        show_user: false
+        description: Sets the content-type of the logs if there is no content-type specified in the file properties. If override_content_type is set to true, any existing content-type in the file properties will be overridden with the value of this attribute.
```
efd6:
```suggestion
        description: Sets the MIME content-type of the logs if there is no content-type specified in the file properties. If override_content_type is set to true, any existing content-type in the file properties will be overridden with the value of this attribute.
```

packages/netskope/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.1.1"
+  changes:
+    - description: Added advanced configuration options for setting and overriding content-type of log files.
+      type: bugfix
```
efd6, line 5: Is this a bugfix or an enhancement; the issue says it's a bug, but the PR is labelled as an enhancement.

ShourieG: Yes this is a bugfix, I'll update the labels

packages/netskope/data_stream/alerts_events_v2/manifest.yml
ShourieG: So the override will never trigger if the content-type is empty. But there are scenarios where you can set the content-type without overriding, i.e (set content-type if there is no existing content-type set in blob properties). 

So `content-type` is technically mutually exclusive, but `override_content_type` only works if something is set inside `content-type`

efd6: OK. Thanks

ShourieG:
@efd6, made the suggested changes

efd6 (approved):

Issue #14069, opened 2025-05-29
[New Integration] Forescout


cpascale43:
# Description
Forescout provides continuous, agentless device visibility and network access control across connected devices including IT, IoT and OT environments. It discovers and classifies all network connected devices in real time and provides contextual information about device security posture, compliance status, user associations and network behavior patterns. 
Forescout exposes device intelligence via its Web API and eyeExtend integration framework. The Elastic <> Forescout integration should consist of:
- Forescout eyeExtend Connect module (pushes data from Forescout to Elastic)
- Elastic integration package (receives data, normalizes to ECS, provides visualizations)
- Elastic response actions (triggers policies in Forescout such as isolating or quarantining potentially compromised endpoints)

# Architecture
This integration package creates an HTTP-based data ingestion pipeline to receive device intelligence data from Forescout systems and normalize it into ECS format. 

### Scope of this issue
- Elastic integration package (HTTP endpoint input, data transformation and dashboards)

### Related/dependent issues
- Forescout eyeExtend Connect module development (external)
- Elastic response actions (future)

# User stories
As a SOC analyst I want to: 
- Correlate Forescout device context with security events from other sources in Elastic Security
- Detect anomalies and threats using Forescout's device intelligence 
- View device compliance status, security posture gaps and policy violations in Elastic dashboards

As a network security engineer, I want to:
- Monitor real-time inventory of all connected devices across IT, IoT and OT envs
- Track device classification, authentication, connection patterns and location over time
- Maintain accurate asset inventory including device type, OS, applications, users and peripherals

# Acceptance criteria 
Data ingestion & transformation 
- [ ] Integration package uses `http_endpoint` input type for receiving Forescout webhook data
  - [ ] Support for comprehensive device properties:
      - Device classification
      - OS, applications and peripheral information
      - Network connection details and location context
      - Vuln scan results and IOCs
      - Policy compliance status 
- [ ] Ingest pipeline transforms Forescout device properties to ECS format

Visualizations
- [ ] OOTB dashboards show:
  - Real-time device inventory across all device types
  - Device compliance status and security posture gaps
  - User types (corporate, guest) and access patterns
  - Network connection details and location mapping
  - Policy trends and violation tracking

Scale & performance
- [ ] Support customer requirement of ~2 million endpoints through proper architecture

cpascale43:
Notes:
[...]

PR #16426, opened 2025-12-09
[forescout] Initial release of the forescout

sharadcrest:
## Proposed commit message
```
The initial release includes event data stream, associated dashboard
and visualizations.

forescout fields are mapped to their corresponding ECS fields where possible.

sanitized.
```
## Checklist

### To test the forescout package:

- Clone integrations repo.
- Move to integrations/packages/forescout directory.

> elastic-package test
```
2025/12/09 15:08:54  INFO License text found in "/home/devuser/github/integrations/LICENSE.txt" will be included in package
--- Test results for package: forescout - START ---
╭───────────┬─────────────┬───────────┬────────────────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE   │ DATA STREAM │ TEST TYPE │ TEST NAME                                                          │ RESULT │ TIME ELAPSED │
├───────────┼─────────────┼───────────┼────────────────────────────────────────────────────────────────────┼────────┼──────────────┤
│ forescout │             │ asset     │ dashboard forescout-52ac29ab-2ce2-4d68-937d-cac1cb92ab47 is loaded │ PASS   │        921ns │
│ forescout │             │ asset     │ search forescout-09d5d60b-1995-4851-b518-9337a4761cae is loaded    │ PASS   │        136ns │
│ forescout │ event       │ asset     │ index_template logs-forescout.event is loaded                      │ PASS   │        261ns │
│ forescout │ event       │ asset     │ ingest_pipeline logs-forescout.event-0.1.0 is loaded               │ PASS   │        203ns │
╰───────────┴─────────────┴───────────┴────────────────────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: forescout - END   ---
Done
--- Test results for package: forescout - START ---
╭───────────┬─────────────┬───────────┬────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE   │ DATA STREAM │ TEST TYPE │ TEST NAME                                          │ RESULT │ TIME ELAPSED │
├───────────┼─────────────┼───────────┼────────────────────────────────────────────────────┼────────┼──────────────┤
│ forescout │ event       │ pipeline  │ (ingest pipeline warnings test-pipeline-event.log) │ PASS   │ 353.515412ms │
│ forescout │ event       │ pipeline  │ test-pipeline-event.log                            │ PASS   │ 183.035682ms │
╰───────────┴─────────────┴───────────┴────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: forescout - END   ---
Done
--- Test results for package: forescout - START ---
No test results
--- Test results for package: forescout - END   ---
Done
Run script tests for the package
--- Test results for package: forescout - START ---
PKG forescout
[no test files]
--- Test results for package: forescout - END ---
Done
--- Test results for package: forescout - START ---
╭───────────┬─────────────┬───────────┬──────────────────────────┬────────┬──────────────╮
│ PACKAGE   │ DATA STREAM │ TEST TYPE │ TEST NAME                │ RESULT │ TIME ELAPSED │
├───────────┼─────────────┼───────────┼──────────────────────────┼────────┼──────────────┤
│ forescout │ event       │ static    │ Verify sample_event.json │ PASS   │ 125.779254ms │
╰───────────┴─────────────┴───────────┴──────────────────────────┴────────┴──────────────╯
--- Test results for package: forescout - END   ---
Done
2025/12/09 15:08:59  INFO Installing package...
2025/12/09 15:08:59  INFO License text found in "/home/devuser/github/integrations/LICENSE.txt" will be included in package
2025/12/09 15:09:11  INFO Running test for data_stream "event" with configuration 'udp'
2025/12/09 15:09:20  INFO Setting up independent Elastic Agent...
2025/12/09 15:09:31  INFO Setting up service...
2025/12/09 15:09:53  INFO Validating test case...
2025/12/09 15:09:54  INFO Tearing down service...
2025/12/09 15:09:55  INFO Write container logs to file: /home/devuser/github/integrations/build/container-logs/forescout-event-udp-1765273195104041584.log
2025/12/09 15:09:58  INFO Tearing down agent...
2025/12/09 15:09:58  INFO Write container logs to file: /home/devuser/github/integrations/build/container-logs/elastic-agent-1765273198432716438.log
2025/12/09 15:10:11  INFO Running test for data_stream "event" with configuration 'tcp'
2025/12/09 15:10:20  INFO Setting up independent Elastic Agent...
2025/12/09 15:10:31  INFO Setting up service...
2025/12/09 15:10:51  INFO Validating test case...
2025/12/09 15:10:52  INFO Tearing down service...
2025/12/09 15:10:53  INFO Write container logs to file: /home/devuser/github/integrations/build/container-logs/forescout-event-tcp-1765273253274622964.log
2025/12/09 15:10:56  INFO Tearing down agent...
2025/12/09 15:10:56  INFO Write container logs to file: /home/devuser/github/integrations/build/container-logs/elastic-agent-1765273256451904211.log
2025/12/09 15:11:09  INFO Uninstalling package...
--- Test results for package: forescout - START ---
╭───────────┬─────────────┬───────────┬───────────┬────────┬───────────────╮
│ PACKAGE   │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT │  TIME ELAPSED │
├───────────┼─────────────┼───────────┼───────────┼────────┼───────────────┤
│ forescout │ event       │ system    │ tcp       │ PASS   │ 41.396462621s │
│ forescout │ event       │ system    │ udp       │ PASS   │ 43.344074374s │
╰───────────┴─────────────┴───────────┴───────────┴────────┴───────────────╯
--- Test results for package: forescout - END   ---
Done
```

## Related issues
- Relates #14069 

## Screenshots
image
image

## Go Code for Ingest Pipeline Generation
The event data stream pipeline is generated using Go code with the help of [Dispear](https://github.com/efd6/dispear) library.
```
package main

import (
	"fmt"
	"strings"

	. "github.com/efd6/dispear"
)

const (
	ECSVersion = "9.2.0"
	PkgRoot    = "json"
)
const errorFormat = "Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}"

// safeNavigateAndCheck converts a dot-separated field path to a safe navigation string.
//
// Example:
// "forescout.event.message" -> "ctx.forescout?.event?.message"
func safeNavigateAndCheck(field string) string {
	parts := strings.Split(field, ".")
	condition := "ctx"
	for i, part := range parts {
		if i > 0 { // Skip the first part which is already included in the condition
			condition += fmt.Sprintf("?.%s", part)
		} else {
			condition += fmt.Sprintf(".%s", part)
		}
	}
	return condition
}

func main() {

	// Initial processors of pipeline

	DESCRIPTION("Pipeline for processing event logs.")

	SET("ecs.version").VALUE(ECSVersion).TAG("set ecs.version to 9.2.0")

	// Setting event.* fields
	BLANK()
	BLANK().COMMENT("Set event.* fields")

	SET("event.kind").
		TAG(fmt.Sprintf("set %s to %s", "event.kind", "event")).
		VALUE("event")

	// extract fields from message
	BLANK()
	BLANK().COMMENT("extract fields from message")

	DISSECT("message", "%{forescout.event.service} : TTY=%{forescout.event.tty} ; PWD=%{forescout.event.pwd} ; USER=%{forescout.event.user} ; COMMAND=%{forescout.event.command}").
		IF(safeNavigateAndCheck("message") + " != null").
		DESCRIPTION("Extract fields from the `message` field using the Dissect processor.").
		IGNORE_FAILURE(true)

	// Map custom fields to corresponding ECS and related fields.
	BLANK()
	BLANK().COMMENT("Map custom fields to corresponding ECS and related fields.")

	SET("forescout.event.message").
		TAG(fmt.Sprintf("set %s to %s", "forescout.event.message", "message")).
		COPY_FROM("message").
		IGNORE_EMPTY(true)

	for _, mapping := range []struct {
		field, targetField string
	}{
		{field: "forescout.event.pwd", targetField: "process.working_directory"},
		{field: "forescout.event.user", targetField: "user.name"},
		{field: "forescout.event.user", targetField: "process.user.name"},
		{field: "forescout.event.command", targetField: "process.command_line"},
	} {
		SET(mapping.targetField).
			COPY_FROM(mapping.field).
			TAG(fmt.Sprintf("set %s from %s", mapping.targetField, strings.ReplaceAll(mapping.field, ".", "_"))).
			IGNORE_EMPTY(true)
	}

	// Map related fields
	BLANK()
	BLANK().COMMENT("Append related.* fields")

	APPEND("related.user", "{{{process.user.name}}}").
		ALLOW_DUPLICATES(false).IF(safeNavigateAndCheck("forescout.event.user") + " != null")
	APPEND("related.hosts", "{{{log.syslog.hostname}}}").
		ALLOW_DUPLICATES(false).IF(safeNavigateAndCheck("log.syslog.hostname") + " != null")

	// Remove duplicate custom fields
	BLANK()
	BLANK().COMMENT("Remove duplicate custom fields")
	REMOVE(
		"forescout.event.message",
		"forescout.event.pwd",
		"forescout.event.user",
		"forescout.event.command",
	).
		IF("ctx.tags == null || !ctx.tags.contains('preserve_duplicate_custom_fields')").
		TAG("remove_custom_duplicate_fields").
		IGNORE_MISSING(true)

	// Clean up script

	BLANK()
	BLANK().COMMENT("Cleanup")

	SCRIPT().
		TAG("script_to_drop_null_values").
		DESCRIPTION("This script processor iterates over the whole document to remove fields with null values.").
		LANG("painless").
		SOURCE(`
		void handleMap(Map map) {
			map.values().removeIf(v -> {
				if (v instanceof Map) {
					handleMap(v);
				} else if (v instanceof List) {
					handleList(v);
				}
				return v == null || v == '' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0)
			});
		}
		void handleList(List list) {
			list.removeIf(v -> {
				if (v instanceof Map) {
					handleMap(v);
				} else if (v instanceof List) {
					handleList(v);
				}
				return v == null || v == '' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0)
			});
		}
		handleMap(ctx);
		`)

	// Set and Append processor on last

	SET("event.kind").VALUE("pipeline_error").
		TAG("set event.kind to pipeline_error").
		IF("ctx.error?.message != null")
	APPEND("tags", "preserve_original_event").
		IF("ctx.error?.message != null").
		ALLOW_DUPLICATES(false)

	// Global on failure processor

	ON_FAILURE(
		APPEND("error.message", errorFormat),
		SET("event.kind").VALUE("pipeline_error").
			TAG("set event.kind to pipeline_error"),
		APPEND("tags", "preserve_original_event").
			ALLOW_DUPLICATES(false),
	)

	// Generate the pipeline

	Generate()
}

```


.github/CODEOWNERS
```diff
@@ -231,6 +231,7 @@
 /packages/first_epss @elastic/security-service-integrations
 /packages/fleet_server @elastic/fleet
 /packages/forcepoint_web @elastic/security-service-integrations
+/packages/forescout @elastic/security-service-integrations
```
narph: @cpascale43 , @qcorporation from a quick glance this integration belongs to the integration experience team, using tcp/udp input

qcorporation: @narph this makes a lot of sense.
@sharadcrest, can you update the owner to @elastic/integration-experience and assign it to us for review

packages/forescout/manifest.yml
```diff
@@ -0,0 +1,37 @@
...
+    description: Collect Forescout logs.
+    inputs:
+      - type: tcp
+        title: Collect logs from Forescout via TCP
+        description: Collecting logs from Forescout via TCP.
+      - type: udp
+        title: Collect logs from Forescout via UDP
+        description: Collecting logs from Forescout via UDP.
+owner:
+  github: elastic/security-service-integrations
```
narph: @sharadcrest , a team update is also necessary here

packages/forescout/validation.yml
```diff
@@ -0,0 +1,3 @@
+errors:
+  exclude_checks:
+    - SVR00004 # References in dashboards.
```
taylor-swanson: Please embed items into the dashboard rather than referencing them (usually this is just checkbox when adding/editing the element). Once all references are removed, this file can be deleted.

packages/forescout/data_stream/event/fields/ecs.yml
```diff
@@ -0,0 +1,9 @@
+# Define ECS constant fields as constant_keyword
+- name: log.syslog.version
+  external: ecs
+  type: constant_keyword
+  value: RFC3164
```
taylor-swanson: https://www.elastic.co/docs/reference/ecs/ecs-log#field-log-syslog-version

>The version of the Syslog protocol specification. Only applicable for RFC 5424 >messages.
>
>type: keyword
>
>example: 1

This refers to the syslog protocol specification (only applicable for RFC 5424 messages). I would remove this entry.

packages/forescout/data_stream/event/fields/fields.yml
```diff
@@ -0,0 +1,18 @@
+- name: forescout
+  type: group
+  fields:
+    - name: event
+      type: group
+      fields:
+        - name: command
+          type: keyword
+        - name: message
+          type: keyword
```
taylor-swanson:
```suggestion
          type: match_only_text
```

I think `match_only_text` might be a better fit here (`message` from ECS uses `match_only_text`)

packages/forescout/data_stream/event/fields/ecs.yml
taylor-swanson, line 1: While I know that ECS fields can be dynamically mapped, I think moving forward, we'll still define them in the this file, mainly for two reasons:

- Documentation will be generated for the fields
- The proper ECS definitions (that is, the version referenced in `_dev/build/build.yml`) will be written to the package at build time, rather than trying to rely on runtime resolution. The latter of which may not work in situations where the stack is older (say we have an ECS 9.3.0 field, but the stack only references ECS 8.17.0).

I believe these should be all of the used fields as of right now:
```yaml
- name: event.category
  external: ecs
- name: event.kind
  external: ecs
- name: event.type
  external: ecs
- name: error.message
  external: ecs
- name: log.syslog.appname
  external: ecs
- name: log.syslog.facility.code
  external: ecs
- name: log.syslog.facility.name
  external: ecs
- name: log.syslog.hostname
  external: ecs
- name: log.syslog.priority
  external: ecs
- name: log.syslog.procid
  external: ecs
- name: log.syslog.severity.code
  external: ecs
- name: log.syslog.severity.name
  external: ecs
- name: message
  external: ecs
- name: process.command_line
  external: ecs
- name: process.user.name
  external: ecs
- name: process.working_directory
  external: ecs
- name: user.name
  external: ecs
- name: related.hosts
  external: ecs
- name: related.user
  external: ecs
```

packages/forescout/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,127 @@
...
+  - set:
+      tag: set_ecs_version_to_9_2_0_3273339c
+      field: ecs.version
+      value: 9.2.0
+
+  # Set event.* fields
+  - set:
+      tag: set_event_kind_to_event_de80643c
+      field: event.kind
+      value: event
```
taylor-swanson, line 13: Looks like we are missing the following ECS categorization fields:

- `event.category`
- `event.type`

packages/forescout/_dev/build/docs/README.md
```diff
@@ -0,0 +1,95 @@
+# Forescout Integration for Elastic
+
+## Overview
+[Forescout](https://www.forescout.com) is a leading device visibility and control platform that enables organizations to continuously identify, classify, and enforce security policies across all connected devices. It provides real-time visibility into IT, IoT, OT, and unmanaged devices across enterprise networks.
+
+The Forescout integration for Elastic allows you to collect event data using the TCP and UDP, then visualize the data in Kibana.
```
taylor-swanson:
```suggestion
The Forescout integration for Elastic allows you to collect event data using the TCP and UDP input, then visualize the data in Kibana.
```

or

```suggestion
The Forescout integration for Elastic allows you to collect event data using TCP and UDP, then visualize the data in Kibana.
```

packages/forescout/data_stream/event/fields/ecs.yml
```diff
@@ -0,0 +1,9 @@
...
+- name: observer.vendor
+  external: ecs
+  type: constant_keyword
+  value: forescout
```
taylor-swanson:
```suggestion
  value: Forescout
```

ECS typically capitalizes the vendor names.

taylor-swanson (commented):
~~Overall this looks great, just a few comments.~~

Edit: I missed the bigger issues with unparsed logs that @ilyannn pointed out. See my updated response below.

packages/forescout/data_stream/event/_dev/test/pipeline/test-pipeline-event.log-expected.json
```diff
@@ -0,0 +1,382 @@
...
+            },
+            "event": {
+                "kind": "event"
+            },
+            "forescout": {
+                "event": {
+                    "message": "Accepted publickey for root from 172.20.10.101 port 46018 ssh2: RSA SHA256:WokXPUll0YJnJwbAFK1xYfYR+DaVN2RVFEyK6lMW78c"
+                }
+            },
+            "message": "Accepted publickey for root from 172.20.10.101 port 46018 ssh2: RSA SHA256:WokXPUll0YJnJwbAFK1xYfYR+DaVN2RVFEyK6lMW78c",
```
ilyannn, line 89: This event is unparsed – is that expected?

packages/forescout/data_stream/event/_dev/test/pipeline/test-pipeline-event.log-expected.json
```diff
@@ -0,0 +1,382 @@
...
+        {
+            "ecs": {
+                "version": "9.2.0"
+            },
+            "event": {
+                "kind": "event"
+            },
+            "forescout": {
+                "event": {
+                    "message": "New session 14906 of user root."
```
ilyannn, line 120: This is also unparsed; in fact there is only one DISSECT pattern and most of the test logs are not processed by it. Is this a conscious decision or an oversight?

packages/forescout/data_stream/event/fields/ecs.yml
```diff
@@ -0,0 +1,9 @@
+# Define ECS constant fields as constant_keyword
+- name: log.syslog.version
+  external: ecs
+  type: constant_keyword
+  value: RFC3164
+- name: observer.vendor
```
ilyannn, line 24: Should we also add the `observer.product` etc.?

taylor-swanson (commented):
So looking at the example logs (particularly the system logs as they are the full syslog messages), it looks like we are receiving all of the system messages (see logs from `sshd` and `systemd`), not just the Forescout service (`_fsservice`).

To echo @ilyannn's question, is this intended? Looking at the [syslog docs for Forescout](https://docs.forescout.com/bundle/syslog-3-6-1-h/page/syslog-3-6-1-h.Syslog-Triggers.html), I suppose it is intentional, since you can enable system logs. It appears it just forwards regular systemd and other service logs.

This is problematic since in order to properly parse these, not only do we have to write parsers for Forescout-specific logs, we also have to write parsers for regular linux-type logs. However, if we just want to generically ingest them, then the current solution is "okay", although I would prefer if the dissect keys off of the syslog app name, rather than blindly running the dissect and relying on `ignore_failure` to recover it. This is bad because a legitimate failure will be silenced by ignore_failure.

Since it sounds like there are other types of events to handle (of which I don't think the current dissect can handle?), a better solution needs to be found to handle other log types.

packages/forescout/data_stream/event/fields/ecs.yml
sharadcrest: The syslog data we’re collecting is forwarded through the syslog plugin, which is a core module of Forescout. So we can either set `observer.product` to `Forescout` which will be same as `observer.vendor` or leave it unmapped as it is in the current implementation. I’m happy to go with whichever approach you think is more appropriate.

packages/forescout/data_stream/event/elasticsearch/ingest_pipeline/default.yml
sharadcrest: `event.type` can be a `info` but for event.category I couldn't identify suitable value. I’d really appreciate it if you could take a look and share your suggestions or guidance.

sharadcrest:
Regarding the comments https://github.com/elastic/integrations/pull/16426/#discussion_r2829925271 and https://github.com/elastic/integrations/pull/16426/#discussion_r2829930934, I agree with @taylor-swanson(https://github.com/elastic/integrations/pull/16426/#pullrequestreview-3828600423). It was intentionally left unparsed.

According to [RFC 3164 log format](https://datatracker.ietf.org/doc/html/rfc3164#section-4.1.3), the MSG portion of a syslog message contains the tag and message text, and it doesn’t enforce a structured format. Because of that, even if we extract important values like IP or port, it’s not always possible to determine their semantic role (e.g., source.ip vs destination.ip vs host.ip) from the free-form content.

packages/forescout/manifest.yml
taylor-swanson:
```suggestion
  github: elastic/integration-experience
```

packages/forescout/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: '0.1.0'
+  changes:
+    - description: Initial release.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1 # FIXME Replace with the real PR link
```
taylor-swanson:
```suggestion
      link: https://github.com/elastic/integrations/pull/16426
```

taylor-swanson (commented):
> Regarding the comments #16426 and #16426, I agree with @taylor-swanson(#16426). It was intentionally left unparsed.
> 
> According to [RFC 3164 log format](https://datatracker.ietf.org/doc/html/rfc3164#section-4.1.3), the MSG portion of a syslog message contains the tag and message text, and it doesn’t enforce a structured format. Because of that, even if we extract important values like IP or port, it’s not always possible to determine their semantic role (e.g., source.ip vs destination.ip vs host.ip) from the free-form content.

I would think that exporting system logs (or really anything non-Forescout) should be turned off, but it's still a good idea to gracefully handle these logs coming through, even if it's just indexing them as-is.

I was going to mention looking at `log.syslog.appname` again, but then I looked closer at the example logs. It seems like all of these were run through sudo, which isn't a bad thing, it's just that the appname is going to be `sudo` instead of `_fsservice`. I was going to suggest something like: "If the appname is NOT `_fsservice`, then maybe think about dropping the log," but that may not be as easy as I think. It's probably safer right now to just index everything. If a user doesn't want those logs, then the best course of action would be to turn off the export of those logs.

packages/forescout/data_stream/event/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: Yeah this highly depends on the content of the log. To give you an idea of how varied this can be, this is a [large script from cisco_asa](https://github.com/elastic/integrations/blob/main/packages/cisco_asa/data_stream/log/elasticsearch/ingest_pipeline/default.yml#L2498) that sets these fields. It's a bit "easier" in that integration as there are message IDs we can key off of to reliably set the fields. It doesn't look like we have that luxury with this integration.

I'm okay with this being deferred to a follow-up task, but is something that should be addressed before the integration is GA-ed.

sharadcrest:
Hey @taylor-swanson, @ilyannn,

Could you please let me know if anything is still pending from my side? If everything looks good, can we proceed with the merge?

ilyannn (approved):
Since it seems to be fine that we just pass through and not parse the other logs, LGTM

ilyannn:
> I would think that exporting system logs (or really anything non-Forescout) should be turned off, but it's still a good idea to gracefully handle these logs coming through, even if it's just indexing them as-is.

I think it would be good to add this information to the README.

taylor-swanson (approved):
LGTM as well

@elastic/integrations-triaging, could someone take a look (@qcorporation perhaps)?

qcorporation (commented):
@sharadcrest
can you please update `./.github/integrations_bug.yml` and `integration_feature_request.yml` please

qcorporation:
> @sharadcrest can you please update `./.github/integrations_bug.yml` and `integration_feature_request.yml` please

@sharadcrest it would be also nice if you used the LLM docs generation - to generate the docs. @narph have we introduced this tooling to @sharadcrest ?

sharadcrest:
@qcorporation in the `integrations_bug.yml` and `integration_feature_request.yml` files, apart from adding the integration name to the dropdown, is there anything else you would like me to update?

qcorporation:
> @qcorporation in the `integrations_bug.yml` and `integration_feature_request.yml` files, apart from adding the integration name to the dropdown, is there anything else you would like me to update?

@sharadcrest 
If you can add the lines in both files that would be great, that will satisfy my review of this PR as the triaging portion.

Have you talked to @narph about docs generation? We have an LLM based docs generation workflow that I feel could be applied to this new integration - are you aware of it and do you think you should be using it to update the docs instead of writing it manually?

sharadcrest:
> Have you talked to @narph about docs generation?

I haven’t connected with @narph yet regarding the docs generation workflow. I’ll reach out and get guidance on the LLM-based docs generation process. If it fits well for this integration, I’ll use that approach instead of writing the documentation manually.

sharadcrest:
@narph Does LLM based doc generation tool is ready to use?

piyush-elastic:
@qcorporation  — I noticed that we are currently waiting for the LLM-based documentation generation tool. I connected with @ShourieG , and it appears that it is still in progress. Would it be possible to proceed with this PR for now? If there are any changes required later, we can address them as enhancements. This would help us avoid blocking other [PR](https://github.com/elastic/integrations/pull/18097) that are currently in draft cause on this.
CC- @narph , @cpascale43

qcorporation:
> @qcorporation — I noticed that we are currently waiting for the LLM-based documentation generation tool. I connected with @ShourieG , and it appears that it is still in progress. Would it be possible to proceed with this PR for now? If there are any changes required later, we can address them as enhancements. This would help us avoid blocking other [PR](https://github.com/elastic/integrations/pull/18097) that are currently in draft cause on this. CC- @narph , @cpascale43

sound good

qcorporation (approved):

PR #16428, opened 2025-12-09
Remove Windows from Docker integration compatibility section

alaudazzi:
This PR removes the mention of Windows from the Docker integration doc, as it has not been tested and might create confusion for customers.


bturquet (approved):
LGTM, thank you so much @alaudazzi  !

Issue #15439, opened 2025-09-23
ti_misp: fixed expiration date isn't accurate enough for MISP attributes


chemamartinez:
The MISP attributes data stream supports expiration of indicators of compromise (IOC).

This mechanism allows users to control when a valid indicator/attribute will be considered expired/decayed thanks to a configuration option called `IOC Expiration Duration`. The expiration date for a particular attribute is calculated as the `last seen/timestamp + IOC expiration duration`, which is 90 days by default.

For the MISP integration this approach could be considered not aligned with the [MISP criteria for dacaying attributes](https://www.misp-project.org/2019/09/12/Decaying-Of-Indicators.html/), each ingested attribute contains a `decay_score` section like this:
```
"decay_score": [
    {
        "base_score": 80,
        "decayed": false,
        "decaying_model": {
            "id": "12345",
            "name": "Phishing model"
        },
        "score": 30.5
    }
],
```

which gives us a clue of when an attribute is going to be decayed (the score decreases over time and is marked as decayed when reaches certain threshold).

Therefore, in this particular case we can compare when MISP marks attributes as decayed versus when they are decayed within Elastic, and most of cases they don't match (constantly decreasing of the score vs static timeframe).

Trying to adopt MISP criteria and calculate expiration date dynamically gives us some complexes challenges that makes it not a good option, so an alternative would be to be able to create a new mode of working for this data stream where:
- We are able to set an interval where every valid attribute in destination indexes are flushed, let's say daily for example.
- Followed by a reingestion for all the attributes from the API that makes decayed attributes being removed from mentioned indexes.

That would help us to keep valid attributes up-to-date according to MISP. 

We need to take into account transforms limitations explained [here](https://www.elastic.co/docs/explore-analyze/transforms/transform-limitations).

Finally, some changes in the documentation would be good to add:

1. Add documentation about this new operation mode.
2. Add a clarification in the current docs, under the section [`Handling Orphaned IOCs`](https://www.elastic.co/docs/reference/integrations/ti_misp#handling-orphaned-iocs), it suggests only orphaned attributes are affected by the `IOC Expiration duration` when at the end, every ingested attribute that is not already decayed is controlled by that setting.

PR #16491, opened 2025-12-10
ti_misp: add daily refetch option to threat attributes

chemamartinez:
## Proposed commit message

Added a new configuration option `daily_refetch` that, when enabled:
- Performs a daily full refetch: Every 24 hours, the cursor is reset and all attributes from the configured Initial Interval are re-fetched from MISP.
- Ensures decay state synchronization: The next daily refresh cycle will re-ingest attributes with their current decay scores from MISP and removed decayed attributes from destination indices.

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/15439

## Screenshots

image

image


chrisberkhout (commented):
The daily refetch seems good, but setting a 24-hour expiration is confusing.

Since we already have a latest transform, can't we just make sure the transform picks up the latest version of each item (this may require adding an ingest time or something) and leave the decay score and decayed at the same?

Things still get updated as their decay info changes in MISP, and we have the same rules as usual for orphans.

chemamartinez:
> The daily refetch seems good, but setting a 24-hour expiration is confusing.
> 
> Since we already have a latest transform, can't we just make sure the transform picks up the latest version of each item (this may require adding an ingest time or something) and leave the decay score and decayed at the same?
> 
> Things still get updated as their decay info changes in MISP, and we have the same rules as usual for orphans.

I have a doubt with that approach, the idea for forcing a 24-hour expiration for ingested indicators with this new daily refresh mode, is that we make sure that indicators that have been decayed by MISP are dropped from the destination indices every day.

If we leave the `decayed_at` the same, it would mean every indicator that is ingested and not decayed it would be indexed with a `decayed_at` defined by `ioc_expiration_duration` which is 90 days by default. How could we remove that indicator before that date if MISP marked it as decayed in later iterations?

chrisberkhout (approved):

PR #16497, opened 2025-12-10
[kubernetes] Fix conditionals in fields rename in audit logs

zmoog:
## Proposed commit message

Fix conditionals to use the correct field before attempting a field rename operation.

The conditionals are checking the existence of the wrong field `audit.stageTimestamp`, so the `rename` processor fails with the "field [kubernetes.audit.stageTimestamp] already exists" error.

## Checklist


zmoog:
I ran a quick test on Kind and the integration seems to work correctly, I see no errors so far:

image

PR #16516, opened 2025-12-11
kafka_log: clarify default protocol version in description

mauri870:
## Proposed commit message

The default Kafka protocol version depends on the filebeat version. For 8.x it defaults to 1.0.0, and for 9.0+ it defaults to 2.1.0. Update the description to clarify this.

## Checklist

### DIFF
--- a/packages/kafka_log/changelog.yml
+++ b/packages/kafka_log/changelog.yml
@@ -1,3 +1,8 @@
+- version: "1.9.1"
+  changes:
+    - description: Updated the version field description to clarify default protocol versions.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/16516
 - version: "1.9.0"
   changes:
     - description: Add SASL mechanism configuration.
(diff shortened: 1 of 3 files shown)


lalit-satapathy:
@agithomas / @stefans-elastic for review, given recent kafka upgrades.

agithomas:
The change looks good to me. My suggestion would be to add a troubleshooting section in the README document to help those users migrating from the 8.x+ stack to the 9.x+ stack. In that process, the users would fail to notice this dependency.

@stefans-elastic , as you have worked on the sarama library dependency issue pertaining to the metrics, kindly check from your end as well.  Thanks.

stefans-elastic (approved):
I agree: the change looks good. Also it matches the version we use in [metricbeat](https://github.com/elastic/beats/blob/72314a046f861de2e6b1bc3ae1beb8d979e72cd2/metricbeat/module/kafka/consumergroup/consumergroup.go#L57).
Overall according to sarama code [comment](https://github.com/elastic/sarama/blob/7672917f26b6112627457d6bd1736a8636449c5b/config.go#L490) it is better to use lower version in sarama config than actual Kafka version it is running against because otherwise it may lead to random breakages.

agithomas (approved):
Code owner approval based on the above-mentioned validations.

mykola-elastic (approved):

PR #16542, opened 2025-12-12
refactor(system.auth): consolidate auth message parsing

andrewkroh:
## Proposed commit message

```
Extract duplicated message parsing logic from log.yml and journald.yml
into a shared message.yml pipeline. This consolidates processing for
SSH events, sudo commands, PAM messages, user/group management, and
event categorization into a single location.

The shared pipeline operates on an input contract where callers
set the 'message' field before invocation. This eliminates the
_temp.message intermediate field.

The log pipeline parses the syslog header and extracts the message body
directly into the message field (when the beats syslog processor hasn't 
already done so).

The journald pipeline does some minor cleanup to fields produced by the
journald input to align with ECS and eliminate noise.

The message field is now preserved for all events regardless
of input type, unifying behavior between log and journald inputs.

Add event.kind: pipeline_error to the default.yml on_failure handler
per best practices. Fix mustache template escaping in message.yml
error messages to use triple braces for proper value interpolation.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

This will make it easier to incorporate the enhanced PAM log processing logic from https://github.com/elastic/integrations/pull/14456.


Tacklebox (approved):
The auth log pipelines have a single source of truth and can't get out of sync now 👍

PR #16565, opened 2025-12-15
filestream: Promote GZIP decompression to GA

### LINKED ISSUE #16219: [filestream] Adjust Filestream Integration for GZIP being enabled by default
# ONLY AFTER 9.3.0 IS RELEASED

Update the filestream integration to align with the removal of the `gzip_experimental` flag and the new default behaviour.

### Acceptance Criteria

- Remove `gzip_experimental` toggle from the integration UI/manifests.
- Ensure default configuration templates do not explicitly set `gzip_experimental: true` (it is now redundant).
- keep the default `exclude_files` excluding gzip files, no change in behaviour.
- add a note that GZIP files can be correctly ingested
- Keep mentions to `gzip_experimental` as a 9.2.x only flag

AndersonQ:
## Proposed commit message

```
    filestream: Promote GZIP decompression to GA
    
    Update the filestream integration to use `compression` instead of
    `gzip_experimental`.
    
    Enabling GZIP now sets `compression: auto` and explicitly sets
    `gzip_experimental: true` for backwards compatibility with 9.2.x.
```

## Checklist

~~## Author's Checklist~~

1. Generate 3 plain text log files (a.ndjson, b.ndjson, c.ndjson)
   Note: We generate 100 lines per file to ensure size > 1024 bytes.

```sh
mkdir -p /tmp/test-logs && cd /tmp/test-logs
for f in a b c; do
  for i in {1..100}; do
    echo "file '$f', line $i" >> "$f.ndjson"
  done
done
```

2. **Gzip one of them (a.ndjson.gz) and keep the plain file**

```sh
gzip -k a.ndjson
```

3. **Gzip another and delete the original file (b.ndjson)**

```sh
gzip b.ndjson
```

4. **Verify file state**
   * You should have: a.ndjson, a.ndjson.gz, b.ndjson.gz, c.ndjson
   * Verify sizes are > 1KB (uncompressed):

```sh
ls -lh *.ndjson*
```

5. **Configure the Filestream integration (Fleet UI)**
   * Add the Filestream integration to your policy.
   * Set **Paths** to `/tmp/test-logs/*.ndjson*`.
   * **Enable** "Decompress GZIP files".
   * **Remove** `\.gz$` from the `exclude_files` setting (if present by default).

6. Test Elastic Agent 9.2.x
  - grab the agent `curl -O -L https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-9.2.2-linux-x86_64.tar.gz`
  - extract `tar -xf elastic-agent-9.2.2-linux-x86_64.tar.gz`
  - install the agent `sudo ./elastic-agent-9.2.2-linux-x86_64/elastic-agent install --namespace tmp -nf --url=<FLEET_SERVER_URL> --enrollment-token=<TOKEN>`
  - **Verify Ingestion**
    * filter by `log.file.path:/tmp/test-logs/*`
    * breakdown by `log.file.path` to see the ingested files and the amount of events per file
    * Ensure the contents of a.ndjson/a.ndjson.gz were ingested **once** (deduplication check).
    * Ensure b.ndjson.gz and c.ndjson were ingested.

7. Uninstall the agent (to ensure the state is erased)
 - `sudo /opt/Elastic/Agent-tmp/elastic-agent uninstall`

8. Test Elastic Agent 9.3.x
  - grab the agent `curl -O -L https://snapshots.elastic.co/9.3.0-40d73df7/downloads/beats/elastic-agent/elastic-agent-9.3.0-SNAPSHOT-linux-x86_64.tar.gz`
  - extract `tar -xf elastic-agent-9.3.0-SNAPSHOT-linux-x86_64.tar.gz`
  - install the agent `sudo ./elastic-agent-9.3.0-SNAPSHOT-linux-x86_64/elastic-agent install --namespace tmp -nf --url=<FLEET_SERVER_URL> --enrollment-token=<TOKEN>`
  - **Verify Ingestion**
     * filter by `log.file.path:/tmp/test-logs/*`
     * breakdown by `log.file.path` to see the ingested files and the amount of events per file
     * Ensure the contents of a.ndjson/a.ndjson.gz were ingested **once** (deduplication check).
     * Ensure b.ndjson.gz and c.ndjson were ingested.

## Related issues

- Closes #16219 

## Screenshots

Screenshot from 2025-12-16 08-12-18
Screenshot from 2025-12-16 08-13-15
Screenshot from 2025-12-16 08-13-39

### DIFF
--- a/packages/filestream/manifest.yml
+++ b/packages/filestream/manifest.yml
@@ -37,7 +37,7 @@ policy_templates:
         show_user: true
         default: false
         description: |
-          When enabled, filestream will decompress GZIP-compressed files (.gz) as they are read. For full details, see the [documentation](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-filestream#reading-gzip-files). If enabled, you **must** also remove the '\.gz$' pattern from the "Exclude Files" setting to ensure `.gz` files are ingested. Available for Elastic Agent 9.2.0 in beta and for and Elastic Agent 9.3.0 or newer in GA.
+          When enabled, filestream will decompress GZIP-compressed files (.gz) as they are read. For full details, see the [documentation](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-filestream#reading-gzip-files). If enabled, you **must** also remove the '\.gz$' pattern from the "Exclude Files" setting to ensure `.gz` files are ingested. Available for Elastic Agent 9.2.0 in beta and for Elastic Agent 9.3.0 or newer in GA.
       - name: use_logs_stream
         type: bool
         title: Use the "logs" data stream


AndersonQ:
@nimarezainia, @cmacknz I let the integration using the toggle, making it compatible with 9.2 and easier for the users. Thus, it sets `compression: auto` if enabled and nothing if disabled. That means the users cannot set `compression: gzip`. Honestly, I don't know why one would set it anyway. I think it's fine and aligned with making it easier and hiding complexity.

The only annoyance is the backward compatibility with 9.2. Because of that I also set `gzip_experimental: true` and on 9.3+ it logs a warning saying to use `compression` instead. I'll opened a [quick PR](https://github.com/elastic/beats/pull/48103) on the beats to improve the message so it does not look like the configuration is wrong. But we need it because we can't differentiate from filestream running from the filestream integration or running another way :/
For example, the k8s integration uses filestream and on 9.2 we say users to set `gzip_experimental: true`. So we need to alert them about it.

packages/filestream/manifest.yml
```diff
@@ -29,15 +29,15 @@ policy_templates:
         show_user: true
         default:
           - /var/log/*.log
-      - name: gzip_experimental
+      - name: compression_gzip
```
cmacknz, line 32: Can we get this to be a selection box instead with each of the none, gzip, and auto options in the compression case?

This is to prevent a feature request in the future to expose the gzip option instead of having none and auto be the only choices.

cmacknz: If there is a way to provide arbitrary configuration already then maybe this is fine as the simplest configuration choice.

AndersonQ: we don't have drop box on handlebars. However I cannot say it's impossible because there is a radio button on this integration, but [it's built on kibana](https://github.com/elastic/kibana/commit/4337d10c215f0e87b2da746026f455e62ab60e22#diff-d41c377d0c6c43682936560102144406bd8528ac0530defcb93fc0f03dba1bd2R344-R412). I'll ask Jaime about it. But I'd expect it to be bound to a stack release, but let's see.

Some time ago I've been through that with Tiago, I being the one asking for a drop down. At the end we concluded it wasn't really an option because it could not be done in the integration itself.

belimawr: Yeah, we need to add some features in Fleet UI, those kinds of ask keep poping up on reviews, but, as Anderson said, we're limited by handlebars/Kibana. Having a drop down is one of them.

cmacknz: Thanks, fine as is then

AndersonQ: @cmacknz how can we work with the other teams to try to add custom helpers to our handlebars. There is a way to add, at least, some extra functions to it. I believe we could benefit from that.

packages/filestream/manifest.yml
```diff
@@ -29,15 +29,15 @@ policy_templates:
...
         type: bool
         title: Decompress GZIP files
         multi: false
         required: false
         show_user: true
         default: false
         description: |
-          When enabled, filestream will decompress GZIP-compressed files (.gz) as they are read. This is beta feature. For full details, see the [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream#reading-gzip-files). If enabled, you **must** also remove the '\.gz$' pattern from the "Exclude Files" setting to ensure `.gz` files are ingested. Available for Elastic Agent 9.2.0 and newer.
+          When enabled, filestream will decompress GZIP-compressed files (.gz) as they are read. For full details, see the [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream#reading-gzip-files). If enabled, you **must** also remove the '\.gz$' pattern from the "Exclude Files" setting to ensure `.gz` files are ingested. Available for Elastic Agent 9.2.0 in beta and for and Elastic Agent 9.3.0 or newer in GA.
```
belimawr: The link for the documentation is not working, I believe it should be https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-filestream#reading-gzip-files:
```suggestion
          When enabled, filestream will decompress GZIP-compressed files (.gz) as they are read. For full details, see the [documentation](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-filestream#reading-gzip-files). If enabled, you **must** also remove the '\.gz$' pattern from the "Exclude Files" setting to ensure `.gz` files are ingested. Available for Elastic Agent 9.2.0 in beta and for and Elastic Agent 9.3.0 or newer in GA.
```

packages/filestream/manifest.yml
cmacknz: Create an issue in ingest-dev as the spec for what you want is the starting point, the features you want to add are owned by the UI and Integrations team so pinging them with Team:Fleet on the issue would be a starting point, or asking about this in #fleet.

packages/filestream/manifest.yml
```diff
@@ -29,15 +29,15 @@ policy_templates:
...
+      - name: compression_gzip
         type: bool
         title: Decompress GZIP files
         multi: false
         required: false
         show_user: true
         default: false
         description: |
-          When enabled, filestream will decompress GZIP-compressed files (.gz) as they are read. This is beta feature. For full details, see the [documentation](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream#reading-gzip-files). If enabled, you **must** also remove the '\.gz$' pattern from the "Exclude Files" setting to ensure `.gz` files are ingested. Available for Elastic Agent 9.2.0 and newer.
+          When enabled, filestream will decompress GZIP-compressed files (.gz) as they are read. For full details, see the [documentation](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-filestream#reading-gzip-files). If enabled, you **must** also remove the '\.gz$' pattern from the "Exclude Files" setting to ensure `.gz` files are ingested. Available for Elastic Agent 9.2.0 in beta and for and Elastic Agent 9.3.0 or newer in GA.
```
rdner:
```suggestion
          When enabled, filestream will decompress GZIP-compressed files (.gz) as they are read. For full details, see the [documentation](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-filestream#reading-gzip-files). If enabled, you **must** also remove the '\.gz$' pattern from the "Exclude Files" setting to ensure `.gz` files are ingested. Available for Elastic Agent 9.2.0 in beta and for Elastic Agent 9.3.0 or newer in GA.
```

Issue #14912, opened 2025-05-20
[Issue]: Audit Log Parsing


jameswiggins:
### Type of issue

None

### What documentation page is affected

https://www.elastic.co/docs/reference/integrations/panw/

### What happened?

I have a multitude of firewalls sending me logs across different companies and versions of PA. For some reason all of their audit type logs match this grok pattern: 

`^%{DATA} - - - - %{FIELD:observer.serial_number},%{TIMESTAMP:_temp_.generated_time},%{FIELD:panw.panos.type},%{GREEDYDATA:message}$`

instead of the two patterns listed in the grok processor of the `logs-panw.panos-5.2.1` pipeline.

Can this pattern be added to the patterns in the grok processor of `logs-panw.panos-5.2.1` pipeline?

I had to modify a "managed" pipeline (which I received a warning about) in order to get my audit logs to parse correctly

alaudazzi:
It is an enhancement request for the integration pipeline, rather than a documentation request. Reassigning it as appropriate.

kgeller:
Hello all! I've moved this issue to the integrations repo, where we typically track integration enhancement requests / bugs and I've added it to the team backlog.

jameswiggins:
Has the correct team been notified of this feature request? EverytimeI update the PA integration resources my audit logs start failing to parse and I have to add the above grok pattern to the managed pipeline. I've provided the Grok pattern to use... seems like an easy win to just add it to the integration.

jameswiggins:
Pinging this conversation. Can someone please re-open and approve my merge request?

PR #16566, opened 2025-12-15
Add Grok pattern for PANW Audit Logs 

jameswiggins:
I added an additional Grok pattern to the list available in the default ingest pipeline. I did this because the PANW audit logs were failing to parse. I have tested these changes in production and verified their efficacy in parsing the panw audit logs. 

To test, the reviewer could install the PANW integration, try to ingest some PANW audit logs and verify this pattern is required to parse audit logs and that adding this pattern does in fact parse them correctly

## Related issues
https://github.com/elastic/integrations/issues/14912


jameswiggins:
I have signed the CLA

taylor-swanson (commented):
Hey @jameswiggins, sorry we didn't get to this before.

Do you have a sanitized log sample we can use as pipeline test? We'll also need the version bumped in the manifest file (a patch version change should suffice since this seems like a bugfix) and an entry in the changelog.

PR #16568, opened 2025-12-15
Akamai: add recovery_interval parameter and handle 400 invalid timestamp

chemamartinez:
## Proposed commit message

```
akamai: add configurable recovery interval for historical data re-fetch

The integration now supports a recovery_interval parameter (default: 12h)
that controls the lookback period when re-fetching historical data after
encountering API errors (416, 400 with "Invalid timestamp").

Additionally, enhanced error detection to properly identify and handle
400 Bad Request responses containing "Invalid timestamp" errors by
triggering recovery mode.
```

## Checklist

### DIFF
--- a/packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
+++ b/packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
@@ -113,7 +113,7 @@ program: |-
                       "want_more": lines.size() >= int(state.event_limit),
                     }
                   )
-                : (resp.StatusCode == 416 || (resp.StatusCode == 400 && size(resp.Body) != 0 && bytes(resp.Body).decode_json().as(errorBody, has(errorBody.detail) && errorBody.detail.contains("Invalid timestamp")))) ?
+                : (resp.StatusCode == 416 || (resp.StatusCode == 400 && size(resp.Body) != 0 && bytes(resp.Body).decode_json().as(errorBody, has(errorBody.detail) && errorBody.detail.to_lower().contains("invalid timestamp")))) ?
                   {
                     "events": [
                       {


packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
```diff
@@ -135,6 +136,29 @@ program: |-
...
+                        },
+                      }
+                    ],
+                    "cursor": state.cursor.drop("last_offset").with(
+                      {
+                        "recovery_mode": true,
+                      }
+                    ),
+                    "want_more": true,
+                  }
```
ShourieG: can we combine these 2 into one as the operation is the same, 2 cases seem redundant here

packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
```diff
@@ -112,7 +113,7 @@ program: |-
                       "want_more": lines.size() >= int(state.event_limit),
                     }
                   )
-                : (resp.StatusCode == 416) ?
+                : (resp.StatusCode == 416 || (resp.StatusCode == 400 && size(resp.Body) != 0 && bytes(resp.Body).decode_json().as(errorBody, has(errorBody.detail) && errorBody.detail.contains("Invalid timestamp")))) ?
```
ShourieG:
```suggestion
                : (resp.StatusCode == 416 || (resp.StatusCode == 400 && size(resp.Body) != 0 && bytes(resp.Body).decode_json().as(errorBody, has(errorBody.detail) && errorBody.detail.to_lower().contains("invalid timestamp")))) ?
```

I think we should use to_lower() just to be safe.

ShourieG: We just need to check if to_lower() can be chained with contains() correctly once.

ShourieG:
I also updated the commit message a bit

packages/akamai/data_stream/siem/agent/stream/cel.yml.hbs
chemamartinez: I've verified that they can be chained.

packages/akamai/manifest.yml
```diff
@@ -1,6 +1,6 @@
 name: akamai
 title: Akamai
-version: "3.0.2"
+version: "3.1.0"
```
ShourieG, line 3: @chemamartinez ,Just noticed this, I saw it was tagged as a bugfix, and the custom package shared was `3.0.3`. Should we keep this the same or do we want to label it as an enhancement ? I feel an enhancement is appropriate as we are introducing a new config variable. If so we can add the enhancement label.

ShourieG: I've added the enhancement label

chemamartinez: Yes, it includes two changes and for me one is an enhancement and the other one a bugfix, so I changed the minor version and marked it as both in the changelog but forgot to add the label. Thank you!

Issue #16517, opened 2025-12-11
[qualys_gav]: rate limit missing rate


clement-fouque:
### Integration Name

Qualys Global AssetView [qualys_gav]

### Dataset Name

qualys_gav.asset

### Integration Version

0.6.1

### Agent Version

9.2.2

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.2.2

### OS Version and Architecture

N/A

### Software/API Version

Qualys GAV 2.0

### Error Message

```
{"log.level":"error","@timestamp":"2025-12-11T01:42:47.375Z","message":"rate limit missing rate","component":{"binary":"filebeat","dataset":"elastic_agent.filebeat","id":"cel-zocalo_logs","type":"cel"},"log":{"source":"cel-zocalo_logs"},"service.name":"filebeat","ecs.version":"1.6.0","log.logger":"input.cel","log.origin":{"file.line":786,"file.name":"cel/input.go","function":"github.com/elastic/beats/v7/x-pack/filebeat/input/cel.getLimit"},"id":"cel-qualys_gav.asset-066eee3b-3015-4ffd-acc4-715ffa63b1d9","input_source":"https://gateway.qg2.apps.qualys.com/","input_url":"https://gateway.qg2.apps.qualys.com/","rate_limit":{"headers":"x-ratelimit-Limit=\"\" x-ratelimit-Remaining=\"\" x-ratelimit-Reset=\"3600\""},"ecs.version":"1.6.0"}
```

### Event Original

We don't have the original event since the API call is not working. 

### What did you do?

Launch a full ingest from Qualys GAV

### What did you see?

What's surprising is that we were able to ingest data (2 intervals) and then it stopped. 

<img width="5082" height="1624" alt="Image" src="https://github.com/user-attachments/assets/ce1e761e-8bf5-4108-b848-79bf20b644b5" />

### What did you expect to see?

No rate limit error 😅

clement-fouque:
Thanks folks ! It's working as expected. 

<img width="1714" height="375" alt="Image" src="https://github.com/user-attachments/assets/27d1b0db-99ca-4429-9436-5684effc8a45" />

Note for others: make sure the agent is running at least on `9.3.0`.

PR #16594, opened 2025-12-17
[qualys_gav] Update the data collection to get unique identifiers for each interval

### LINKED ISSUE #16215: [qualys_gav]: Generate unique identifiers for each interval ingestion
### Integration Name

Qualys Global AssetView [qualys_gav]

### Dataset Name

qualys_gav.asset

### Integration Version

0.6.0

### Agent Version

latest

Docker

### User Goal

When using the Qualys Global Asset View integration, the full data ingestion process can take several hours to complete. Tracking when ingestion begins and ends can be challenging. Additionally, if the process spans on 2 days, grouping data by day results in an inaccurate representation of daily breakdowns, as it spreads records across two days.

To address this, adding a unique field for each ingestion interval will allow for precise comparisons of document counts per interval. Ideally, two fields would suffice (names may vary):
- **interval_start**: Example format: `2025-03-19T08:13:31.953Z`
- **interval_id**: Example value: `582fc58b-f983-4c35-abb1-65c507c1dc0c`

> [!TIP]
> It has been already implemented in the Qualys VMDR integration: https://github.com/elastic/integrations/pull/13840 

### Existing Features

There are no existing feature for this use case.

### What did you see?

cf User Goal

### Anything else?

_No response_

brijesh-elastic:
## Proposed commit message

```
qualys_gav: update data collection to get unique identifiers for each interval

This will help to debug if there is any issue in data ingestion.
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/qualys_gav directory.
> elastic-package test

## Related issues

- Closes #16215


packages/qualys_gav/data_stream/asset/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -256,6 +277,19 @@ processors:
...
+      target_field: qualys_gav.asset.interval_start
+      formats:
+        - ISO8601
+      if: ctx.qualys_gav?.asset?.interval_start != null && ctx.qualys_gav.asset.interval_start != ''
+      on_failure:
+        - remove:
+            field: qualys_gav.asset.interval_start
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```
chrisberkhout: Isn't this processing already done in lines 13-28?

clement-fouque (commented):
PR looks great but I could only validated this PR partially as I can't perform a full ingest due to this bug: https://github.com/elastic/integrations/issues/16517

Image

packages/qualys_gav/data_stream/asset/elasticsearch/ingest_pipeline/default.yml
brijesh-elastic: Earlier processing was for fields that aren't inside the `message` field. (i.e., they will be populated when there are errors; see [code](https://github.com/elastic/integrations/pull/16594/files#diff-0eb867a4ee87e9a4ef8d6858526b9e37147eee971c4b4da96a7186dacd3d3cf6R56-R72))

So, processing of both fields is different, as they are in different places.

chrisberkhout: Oh, okay.

I know it expands the scope of the change a bit, but would you mind removing the `"message"` and just returning the final event? It'll be less confusing because then we won't have different event layouts for errors and non-errors and double processing for these new fields.

The part of the CEL where the events are returned would become:
```
            resp.Body.decode_json().as(body, {
              "events": body.assetListData.asset.map(e, e.with({
                "interval_start": state.interval.start,
                "interval_id": state.interval.id,
              })),
            ...
```
then in the ingest pipeline we don't need to handle `"message"` but can add the serialized `event.original` like [this](https://github.com/elastic/integrations/blob/9d9269897460f68637d604f2dcb7ba7e7130ffeb/packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml#L25-L30).

To have the ingest pipeline really not consider `"message"`, i think the pipeline tests may need to be provided in a different file format. I think `test-asset.log` in nsjson format means the testing will add the message wrapper, but you can provide unwrapped events by having a `.json` file as the test input, like [here](https://github.com/elastic/integrations/tree/main/packages/ti_opencti/data_stream/indicator/_dev/test/pipeline).

brijesh-elastic: I kept the `message` field and was also able to keep `interval_*` fields at a single position, and simplified the pipeline logic as well. See [commit](https://github.com/elastic/integrations/pull/16594/commits/cf4123605be6fbf5a39e67652a57d101eb01720f).

chrisberkhout: Okay. So the interval fields are outside of `"message"`, and only the `"message"` value is saved in `event.original`. I think this is okay.

brijesh-elastic:
> PR looks great but I could only validated this PR partially as I can't perform a full ingest due to this bug: #16517
> 
> Image

Thanks @clement-fouque, We will look into the bug and resolve it separately.

Issue #15034, opened 2025-08-25
[New Integration] Axonius


cpascale43:
# Description
Axonius transforms asset data into actionable insights for cybersecurity, IT, and GRC teams. It covers devices, identities, applications, SaaS, exposures, and more across hybrid cloud, on-prem, and OT environments.

The Elastic integration feeds activity logs and asset intelligence into Elastic for advanced analytics and threat hunting.

Sample data is available on request.

# Architecture
Data can be retrieved via Axonius' [REST API](https://docs.axonius.com/docs/axonius-rest-api). Transforms in Elastic should be used to maintain latest device data/prevent duplicates.

The integration syncs normalized asset data from Axonius into Elastic for use in asset inventory, leveraging Axonius' correlation and deduplication capabilities across multiple data sources. Key asset properties include device identification (hostname, IP, MAC address, serial number), software inventory, security tool coverage and network discovery data. 

It should power downstream use cases like: managed/unmanaged device identification, forbidden software detection and network-based asset discovery for gap analysis.

cpascale43:
Heya @caitlinbetz and @jaredburgettelastic - tagging you folks in here for assistance on the asset inventory requirements. 

- What special considerations are there since this integration ingests assets/entities, not just events? 
- Are there specific Axonius fields that must be mapped to populate Elastic's asset inventory correctly?

jaredburgettelastic:
👋 Hey there, @cpascale43 ! Great to hear from you.

I haven't taken a look at Axonius yet, but will put it on my todo list to provide feedback soon!

In the meantime, can you tell me if you all are currently looking to provide both an event-based and an entity analytics-based version of an Axonius integration (similar to other integrations such as Okta)? At a glance and based on your comment, Axonius seems to be mostly asset/entity-centric, so just curious.

Regarding use cases, another downstream use case is threat hunting through entity risk analytics: Axonius seems to provide vulnerability and business criticality information, so I'll be additionally interested in parameters that we may be able to include in our entity risk computations.

Thanks!

cpascale43:
Hey @jaredburgettelastic - the primary use case is entity analytics. Thanks for the info on vulnerability/business criticality.

chrisberkhout:
Some things that came up in https://github.com/elastic/integrations/pull/16603 that should be done for other data streams before this integration is merged to `main`:
- Remove redundant field nesting in the output document. See [discussion](https://github.com/elastic/integrations/pull/16603#discussion_r2774981482).
[...]

PR #16603, opened 2025-12-17
[axonius][alert_and_incident] Add Axonius Alert and Incident datastream

muskan-agarwal26:
## Proposed commit message

This new data stream collects alert findings and incidents data from the Axonius platform via the 
/api/v2/assets/ endpoint. Axonius fields are mapped to their 
corresponding ECS fields where possible.

sanitized.

API documentation: https://docs.axonius.com/docs/axonius-rest-api

## Checklist

To test the axonius package:

- Clone integrations repo.
- Move to integrations/packages/axonius directory.

> elastic-package test
```
2025/12/17 18:21:56  INFO New version is available - v0.117.1. Download from: https://github.com/elastic/elastic-package/releases/tag/v0.117.1
2025/12/17 18:21:57  INFO License text found in "/root/GITHUB/integrations/LICENSE.txt" will be included in package
--- Test results for package: axonius - START ---
╭─────────┬────────────────────┬───────────┬──────────────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE │ DATA STREAM        │ TEST TYPE │ TEST NAME                                                        │ RESULT │ TIME ELAPSED │
├─────────┼────────────────────┼───────────┼──────────────────────────────────────────────────────────────────┼────────┼──────────────┤
│ axonius │                    │ asset     │ dashboard axonius-50841034-1314-4e26-8be3-47d38241c359 is loaded │ PASS   │      2.703µs │
│ axonius │                    │ asset     │ search axonius-852dd7b3-5b99-456b-8f00-9db6b8d6f4bd is loaded    │ PASS   │        481ns │
│ axonius │                    │ asset     │ search axonius-d3b2048c-016a-4ebb-9406-89687aa90aee is loaded    │ PASS   │        546ns │
│ axonius │ alert_and_incident │ asset     │ index_template logs-axonius.alert_and_incident is loaded         │ PASS   │        696ns │
│ axonius │ alert_and_incident │ asset     │ ingest_pipeline logs-axonius.alert_and_incident-0.1.0 is loaded  │ PASS   │        345ns │
╰─────────┴────────────────────┴───────────┴──────────────────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: axonius - END   ---
Done
--- Test results for package: axonius - START ---
╭─────────┬────────────────────┬───────────┬───────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE │ DATA STREAM        │ TEST TYPE │ TEST NAME                                         │ RESULT │ TIME ELAPSED │
├─────────┼────────────────────┼───────────┼───────────────────────────────────────────────────┼────────┼──────────────┤
│ axonius │ alert_and_incident │ pipeline  │ (ingest pipeline warnings test-alert-finding.log) │ PASS   │ 726.121672ms │
│ axonius │ alert_and_incident │ pipeline  │ (ingest pipeline warnings test-incident.log)      │ PASS   │ 712.669141ms │
│ axonius │ alert_and_incident │ pipeline  │ test-alert-finding.log                            │ PASS   │ 194.767219ms │
│ axonius │ alert_and_incident │ pipeline  │ test-incident.log                                 │ PASS   │ 209.348823ms │
╰─────────┴────────────────────┴───────────┴───────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: axonius - END   ---
Done
--- Test results for package: axonius - START ---
No test results
--- Test results for package: axonius - END   ---
Done
--- Test results for package: axonius - START ---
╭─────────┬────────────────────┬───────────┬──────────────────────────┬────────┬──────────────╮
│ PACKAGE │ DATA STREAM        │ TEST TYPE │ TEST NAME                │ RESULT │ TIME ELAPSED │
├─────────┼────────────────────┼───────────┼──────────────────────────┼────────┼──────────────┤
│ axonius │ alert_and_incident │ static    │ Verify sample_event.json │ PASS   │ 220.387313ms │
╰─────────┴────────────────────┴───────────┴──────────────────────────┴────────┴──────────────╯
--- Test results for package: axonius - END   ---
Done
2025/12/17 18:22:05  INFO Installing package...
2025/12/17 18:22:05  INFO License text found in "/root/GITHUB/integrations/LICENSE.txt" will be included in package
2025/12/17 18:22:16  INFO Running test for data_stream "alert_and_incident" with configuration 'default'
2025/12/17 18:22:25  INFO Setting up independent Elastic Agent...
2025/12/17 18:22:39  INFO Setting up service...
2025/12/17 18:23:03  INFO Validating test case...
2025/12/17 18:23:04  INFO Tearing down service...
2025/12/17 18:23:05  INFO Write container logs to file: /root/GITHUB/integrations/build/container-logs/axonius-1765975985552405152.log
2025/12/17 18:23:08  INFO Tearing down agent...
2025/12/17 18:23:08  INFO Write container logs to file: /root/GITHUB/integrations/build/container-logs/elastic-agent-1765975988989458774.log
2025/12/17 18:23:36  INFO Uninstalling package...
--- Test results for package: axonius - START ---
╭─────────┬────────────────────┬───────────┬───────────┬────────┬───────────────╮
│ PACKAGE │ DATA STREAM        │ TEST TYPE │ TEST NAME │ RESULT │  TIME ELAPSED │
├─────────┼────────────────────┼───────────┼───────────┼────────┼───────────────┤
│ axonius │ alert_and_incident │ system    │ default   │ PASS   │ 47.688153488s │
╰─────────┴────────────────────┴───────────┴───────────┴────────┴───────────────╯
--- Test results for package: axonius - END   ---
Done
``` 

## Related issues

- Relates https://github.com/elastic/integrations/issues/15034
- Relates https://github.com/elastic/elasticsearch/pull/139512

## Screenshots
image (5)
image (6)


packages/axonius/changelog.yml
```diff
@@ -3,4 +3,4 @@
   changes:
     - description: Initial release.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/16599
+      link: https://github.com/elastic/integrations/pull/16603
```
chrisberkhout: This can be left out as #16603 is a PR to `feature/axonius-0.1.0` rather than `main`.
The user-facing changelog entry should have the PR that actually does the initial release.

packages/axonius/docs/README.md
chrisberkhout: Mostly not in this PR but I think this generic Filebeat/input content can be removed.

packages/axonius/data_stream/alert_and_incident/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,180 @@
...
+  - pipeline:
+      name: '{{ IngestPipeline "pipeline-alert-finding" }}'
+      tag: pipeline_alert_finding
+      if: >-
+        ctx.axonius?.alert_and_incident?.asset_type.contains('alert_findings')
+  - pipeline:
+      name: '{{ IngestPipeline "pipeline-incident" }}'
+      tag: pipeline-incident
+      if: >-
+        ctx.axonius?.alert_and_incident?.asset_type.contains('incidents')
```
chrisberkhout: I think it would be better to have separate data streams for `alert` an `incident`. I see in the CEL that they use similar collection logic, but I think it would be better to have either 
1) The same CEL code in each data stream, with the only difference being variable values that are set different for each data stream in its manifest. That way the whole `cel.yml.hbs` file is identical across data streams. Data for each type goes into its data stream.
2) Have one CEL program that collects the different types of data, but then it get routed to the appropriate data stream alter. An example of that is in the Amazon Security Lake integration, where the event data stream has the agent config to fetch events, and a [routing_rules.yml](https://github.com/elastic/integrations/blob/main/packages/amazon_security_lake/data_stream/event/routing_rules.yml) file that sends them to the appropriate destination. It's also worth looking at those destination data streams to see how they're set up.

Do you agree one of those would be better?

packages/axonius/data_stream/alert_and_incident/elasticsearch/ingest_pipeline/pipeline-alert-finding.yml
```diff
@@ -0,0 +1,96 @@
+---
+description: Pipeline for processing alert_and_incident logs.
+processors:
+  - rename:
+      field: json.event.data.alert_config_id
+      tag: rename_event_data_alert_config_id
+      target_field: axonius.alert_and_incident.event.data.alert_config_id
+      ignore_missing: true
```
chrisberkhout: There are lot of processors that do renames like this:

`json.event.data.XXX` -> `axonius.alert_and_incident.event.data.XXX`

If the data had been deserialized from JSON directly to `axonius.alert_and_incident` then all that renaming would not be necessary.

Also, I think the `.event` and `.data` prefixes are redundant. For this field, for example, `axonius.alert_and_incident.alert_config_id` is better (or `axonius.alert.alert_config_id` if we have an alert-specific data stream as discussed elsewhere).

packages/axonius/elasticsearch/transform/latest_alert_and_incident/transform.yml
```diff
@@ -0,0 +1,37 @@
+# Use of "*" to use all namespaces defined.
```
chrisberkhout: It would be good to explain the transform in the README. I see it's mentioned that transforms are installed.

As a user I'd like to know that all data is periodically refetched and that's why there are latest transforms, and also the names of the aliases to access the latest data rather than the potentially duplicated data in the source data streams.

packages/axonius/img/axonius-alert-and-incident-dashboard.png
chrisberkhout, line 1: Usually we add the filters on each panel rather then having top-level filters. That is a bit more work in the setup but makes it easier for users to avoid problems. Is there any reason to not do that in this case?

chrisberkhout (commented):
Had a high-level look and I have a few questions about the approach.

I see that this is a PR to a feature branch rather than to main. That's fine, but it would have been helpful to mention that in the PR description and maybe one sentence about the scope of this PR and where it fits in the project.

packages/axonius/img/axonius-alert-and-incident-dashboard.png
muskan-agarwal26: Recently, we had got a comment on PR to apply filters on dashboard level, hence followed same here.

packages/axonius/data_stream/alert_and_incident/elasticsearch/ingest_pipeline/default.yml
muskan-agarwal26: Hi @chrisberkhout, there is one key difference across the data streams: the asset types passed within the endpoint vary for each stream. We can’t send all asset types in a single request because there are around 60 of them, each carrying a large volume of data, which could negatively impact API performance and cause latency.

One possible solution is to provide toggles that allow users to select which asset types they want to fetch. For example, the Exposure data stream may include 3 asset types, while the Compute data stream may include 10. If a user enables the Exposure toggle, only those 3 asset types would be fetched, and similarly for the other data streams.

chrisberkhout: Okay. With many asset types that can be organized into related groups, maybe having one data stream do all the fetching (and re-route later, as in point 2. above) would be too slow (there would be no parallelism).

But having multiple data streams and making the list of asset types a variable that gets passed into identical CEL files could still be a good strategy.

An example of this is in the `checkpoint_harmony_endpoint` integration, where all `cel.yml.hbs` files are identical, but each data stream has a different value for the `filter` variable.

muskan-agarwal26: We don't have to make it user-configurable, it must be hard-coded.

packages/axonius/data_stream/alert_and_incident/elasticsearch/ingest_pipeline/pipeline-alert-finding.yml
muskan-agarwal26: `Also, I think the .event and .data prefixes are redundant. For this field, for example, axonius.alert_and_incident.alert_config_id is better (or axonius.alert.alert_config_id if we have an alert-specific data stream as discussed elsewhere).`

@chrisberkhout 
We follow this same mapping approach across the other data streams as well. Making a change here would require updating the others too—one is already merged and seven more are in review state. This would result in significant rework, so I suggest we skip this change for now. What do you think?

packages/axonius/elasticsearch/transform/latest_alert_and_incident/transform.yml
muskan-agarwal26: @chrisberkhout 
So far, we have been following the current approach across all integrations, where the README template is generated automatically. Is there any reference or documentation for the approach you’re suggesting so that I can implement the changes accordingly?

packages/axonius/docs/README.md
chrisberkhout: Not removed in this PR but that's okay

packages/axonius/data_stream/alert_and_incident/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Thanks for reducing the diff in the cel files to
```diff
--- ./data_stream/alert_finding/agent/stream/cel.yml.hbs	2026-02-26 11:33:52.801393341 +0100
+++ ./data_stream/incident/agent/stream/cel.yml.hbs	2026-02-26 11:33:52.803736523 +0100
@@ -18,11 +18,11 @@
 state:
   api_key: {{api_key}}
   secret_key: {{secret_key}}
   batch_size: {{batch_size}}
   asset_type_list:
-    - alert_findings
+    - incidents
 
 redact:
   fields:
     - api_key
     - secret_key
```

I still think it's better to make the files fully identical and pass this value in as variable.

It's true that users could modify it, but if we make an an advanced option and note that it shouldn't be modified, then I think it's not going to be a problem. That's what's done [here](https://github.com/elastic/integrations/blob/3557f7f9249e12facb3631ae00724e13e43d4977/packages/checkpoint_harmony_endpoint/data_stream/antibot/manifest.yml#L17-L23) (use [here](https://github.com/elastic/integrations/blob/3557f7f9249e12facb3631ae00724e13e43d4977/packages/checkpoint_harmony_endpoint/data_stream/antibot/agent/stream/cel.yml.hbs#L16)).

Having the files identical will make the integration more maintainable, especially once we start using the new `elastic-package links -h` functionality (the current version needs to be fixed - coming soon).

packages/axonius/data_stream/alert_and_incident/elasticsearch/ingest_pipeline/pipeline-alert-finding.yml
chrisberkhout: I think it is worth doing this for all data streams before the `elastic:feature/axonious-0.1.0` branch is merged into `main`. After it's in `main` and released it won't be possible without a breaking change.

packages/axonius/elasticsearch/transform/latest_alert_and_incident/transform.yml
chrisberkhout: A good example is the `ti_abusech` README. I think it's also mentioned in the READMEs of most integrations with transforms.

packages/axonius/img/axonius-alert-and-incident-dashboard.png
chrisberkhout: As I mentioned, it would be good to know what that recommendation was, so we can make sure we're aligned.

Unless there's a compelling argument against it, let's follow the advice in the SEI Integration Dashboard styling guide, as usual.

chrisberkhout (commented):
Thanks for splitting the `alert_and_incident` data stream into `alert_finding` and `incident` data streams!

Regarding the other points from the last review, I checked each and made some comments. I think most of those still need some attention.

packages/axonius/elasticsearch/transform/latest_alert_and_incident/transform.yml
muskan-agarwal26: Still didn't find any idea about it.
In `ti_abusech`, it is mentioned for the ioc indicators.

chrisberkhout: Yes. In that ["Expiration of Indicators of Compromise (IOCs)"](https://github.com/elastic/integrations/blob/main/packages/ti_abusech/docs/README.md#expiration-of-indicators-of-compromise-iocs) section it explains the purpose of the transforms and lists the destination aliases that users should use.

That information is not provided for the transforms in this PR.

packages/axonius/_dev/build/docs/README.md
```diff
@@ -107,6 +107,18 @@ For more information, refer to [Agentless integrations](https://www.elastic.co/g
 3. In the search bar, type **Axonius**.
 4. All transforms from the search results should indicate **Healthy** under the **Health** column.
 
+An [Elastic Transform](https://www.elastic.co/guide/en/elasticsearch/reference/current/transforms.html) is created for every source index to make sure only updated axonius logs are available to the end users. Each transform creates a destination index named `logs-axonius_latest.dest_*` which only contains active and updated axonius logs. The logs match rules and dashboards are updated to list only updated logs.
```
chrisberkhout:
```suggestion
An [Elastic Transform](https://www.elastic.co/guide/en/elasticsearch/reference/current/transforms.html) is created for each data stream, to provide a view of the most recent, active Axonius data. Use the relevant destination alias from the table below to access the latest data, whether for use in dashboards, rules, or elsewhere.
```

packages/axonius/data_stream/alert_and_incident/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: I see the two new `cel.yml.hbs` files now match. If it's possible to do this for others it would be good, but that can be done separately.

packages/axonius/data_stream/alert_and_incident/elasticsearch/ingest_pipeline/pipeline-alert-finding.yml
chrisberkhout: I see this is done for `alert_finding` and `incident`. I think it's important to match the others but it can be done in a separate change.

packages/axonius/changelog.yml
chrisberkhout: This change isn't necessary. A changelog entry is necessary when it's going to be merged to main.

packages/axonius/img/axonius-alert-and-incident-dashboard.png
chrisberkhout: I see the `labels.is_transform_source: false` is still at the dashboard level. Any reason why?

chrisberkhout (changes requested):
Please review the unresolved items from earlier reviews.

chrisberkhout (approved):

Issue #12822, opened 2025-02-18
Crowdstrike FDR: Scaling host metadata enrichment


chemamartinez:
As part of https://github.com/elastic/integrations/issues/2816, a [cache processor](https://www.elastic.co/guide/en/beats/filebeat/current/add-cached-metadata.html) was added to enrich FDR events with host and user metadata at ingest-time.

![Image](https://github.com/user-attachments/assets/271339bd-48a7-41ac-a198-f7b0f2818086)

That means that right now, the cached metadata is stored locally in the agent and the enrichment doesn't work when agents are scaled horizontally.

This issue is intended to track our progress on researching possible solutions/workarounds.

### Ideas

- Support storing data in memcached or redis (something that is available as a service on CSPs). Make the existing cache processor "multi-layer" with read-through to the distributed cache when the local memory cache doesn't contain the key.

- ES|QL is adding a new `lookup join` feature that could be used to perform the metadata join at query time. That would simplify the architecture as it doesn't require any changes on the agent side. See https://github.com/elastic/package-spec/issues/873.

andrewkroh:
With [package-spec v3.4.2](https://github.com/elastic/package-spec/blob/e205165f04340a7e607a114f4638362c96722090/spec/changelog.yml#L46-L65), which introduces support for `index.mode: lookup` indices, I think we should create a proof-of-concept that moves host and user metadata into their own dedicated lookup mode indices, one index for user metadata and another for host metadata.

Next, we need to validate that there are no issues enriching events with ES|QL queries and demonstrate that dashboards can display events enriched with host and user information.

One challenge I foresee is the lack of a curation mechanism in Fleet for managing lookup indices. As a result, these indices could grow indefinitely without any cleanup process. However, this should be manageable in most environments as long as we avoid duplicating users or hosts in the indices.

navnit-elastic:
There may be a challenge with the user metadata enrichment, which is also the case with the existing approach. The UserInfo directory in the FDR feed does not contain the details of all users that appear in the actual data events. This may result in the majority of events remaining unenriched with user metadata.

We have reports saying that other events in the FDR data could improve the user metadata enrichment.

These event includes:
- UserIdentity
- UserLogon
- UserLogoff
- UserInformationEtw

A possible approach I think, could be to route these events as well to the user metadata index. We can modify the events to conform to a common schema to make enrichment easier.

@andrewkroh, could you please share your thoughts on this?

andrewkroh:
[...]

PR #16617, opened 2025-12-18
crowdstrike: add option to route FDR aidmaster and userinfo events to separate data streams

### LINKED ISSUE #16211: crowdstrike: split fdr data stream into three
The plan for scaling CrowdStrike metadata enrichment is to use LOOKUP JOIN, first splitting the current fdr data stream into three: fdr, fdr_aidmaster and fdr_userinfo.

The plan for this is to use routing rules to perform the split based on the `log.file.path` field. The fdr data would be processed essentially as the data stream currently is, and the other two data streams (fdr_aidmaster and fdr_userinfo) would be passed through largely unaltered (exact changes to be determined).

Initially I think rerouting should be configurable, default: "off" so that this is not a breaking change. This would allow us to roll out the code changes without impacting users in any way. The rerouting configuration would need to be considered in the agent cache-based enrichment logic; if rerouting is turned on, we would be expecting that the user would be doing enrichment downstream with LOOKUP JOIN, so enrichment in the agent seems like pointless work, though this may not be the case and this should be discussed further.

navnit-elastic:
## Proposed commit message

```
crowdstrike: add option to route FDR aidmaster and userinfo events to separate data streams

This change adds an option to route FDR host metadata (aidmaster) and user metadata
(userinfo) events to dedicated data streams. This is required to enable lookup
join-based data enrichment capabilities.

This change also adds user.id mapping for userinfo events, which may be used as a
key for user data enrichment.

Adds system test to validate routing behavior with route_metadata_to_datastreams enabled.
```

> [!NOTE]
> This change alone does not enable metadata enrichment. However, this is
required to build transforms on top of the separated data streams, and the
destination index for transform can be used to perform lookup joins.

## Checklist

## Author's Checklist

- [ ] 

Make the following change before running system tests:
```diff
diff --git a/packages/crowdstrike/data_stream/fdr/_dev/deploy/tf/env.yml b/packages/crowdstrike/data_stream/fdr/_dev/deploy/tf/env.yml
index b795fcdeb..6e1f17f7a 100644
--- a/packages/crowdstrike/data_stream/fdr/_dev/deploy/tf/env.yml
+++ b/packages/crowdstrike/data_stream/fdr/_dev/deploy/tf/env.yml
@@ -7,3 +7,4 @@ services:
       - AWS_SESSION_TOKEN=${AWS_SESSION_TOKEN}
       - AWS_PROFILE=${AWS_PROFILE}
       - AWS_REGION=${AWS_REGION:-us-east-1}
+      - TF_VAR_eventbridge_role_arn=arn:aws:iam::144492464627:role/eb-scheduler-role-20231101165501426500000001
```

Then:
```
$ export AWS_ACCESS_KEY_ID="your-access-key-id"
$ export AWS_SECRET_ACCESS_KEY="your-secret-access-key"
$ export AWS_SESSION_TOKEN="your-session-token"
```

Then test with `elastic-package` as normal.

## Related issues

- Closes #16211 
- Relates #12822 

## Screenshots

screencapture-localhost-59556-app-fleet-integrations-crowdstrike-3-2-0-add-integration-2025-12-18-14_11_39

screencapture-localhost-59556-app-discover-2025-12-18-18_25_52

### DIFF
--- a/packages/crowdstrike/changelog.yml
+++ b/packages/crowdstrike/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Add option to route FDR aidmaster (host metadata) and userinfo (user metadata) events to separate data streams (fdr_aidmaster and fdr_userinfo).
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/1
+      link: https://github.com/elastic/integrations/pull/16617
 - version: "3.1.0"
   changes:
     - description: Improve ingest pipeline maintainability.
--- a/packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
@@ -45,10 +45,10 @@ processors:
             value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
   # Conditionally append routing tag to route metadata to separate data streams.
   - append:
-      tag: append_route_metadata_tag
+      tag: append_route_metadata_tag_1adc66f4
+      if: ctx._conf?.route_metadata_to_datastreams == true
       field: tags
       value: route_metadata_to_datastreams
-      if: ctx._conf?.route_metadata_to_datastreams == true
       allow_duplicates: false
   - remove:
       tag: remove_metadata_host_aid_and_user_sid_a4bf7be9
@@ -1539,8 +1539,8 @@ processors:
       target_field: user.id
       ignore_missing: true
   - set:
-      tag: set_user_id_from_crowdstrike_UserSid_readable
-      if: ctx.user?.id == null || ctx.user.id == ""
+      tag: set_user_id_from_crowdstrike_UserSid_readable_53b81ff5
+      if: ctx.user?.id == null || ctx.user.id == ''
       field: user.id
       copy_from: crowdstrike.UserSid_readable
       ignore_empty_value: true


packages/crowdstrike/data_stream/fdr/_dev/test/pipeline/test-fdrv2-userinfo.log
```diff
@@ -0,0 +1 @@
+{"AccountType":"Domain User","LastLoggedOnHost":"COMPUTER1","LocalAdminAccess":"No","LogonInfo":"Domain User Logon","LogonTime":"1702546155.197","LogonType":"Interactive","PasswordLastSet":"1699971198.062","User":"DOMAIN\\BRADLEYA","UserIsAdmin":"0","UserLogonFlags_decimal":"0","UserName":"Alan-One","UserSid_readable":"S-1-12-1-3697283754-1083485977-2164330645-2516515886","_time":"1702546168.576","cid":"ffffffff15754bcfb5f9152ec7ac90ac","event_platform":"Win","monthsincereset":"1.0"}
```
navnit-elastic, line 1: Test case taken from the service deployer files for system tests.

packages/crowdstrike/data_stream/fdr_aidmaster/fields/fields.yml
navnit-elastic, line 1: Field definitions are taken from existing FDR data stream.

packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -43,6 +43,13 @@ processors:
             tag: append_error_message_4ef54c75
             field: error.message
             value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  # Conditionally append routing tag to route metadata to separate data streams.
```
navnit-elastic, line 46: Require to add this tag as we drop `_conf` field in ingest pipeline.

packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1531,6 +1538,12 @@ processors:
       field: crowdstrike.UID
       target_field: user.id
       ignore_missing: true
+  - set:
```
navnit-elastic, line 1541: The `user.id` mapping is required for userinfo events to perform join on this field.

packages/crowdstrike/data_stream/fdr/routing_rules.yml
```diff
@@ -0,0 +1,12 @@
+- source_dataset: crowdstrike.fdr
+  rules:
+    - target_dataset: crowdstrike.fdr_aidmaster
+      if: ctx.tags != null && ctx.tags.contains('route_metadata_to_datastreams') && ((ctx.log?.file?.path != null && ctx.log.file.path.contains('aidmaster')) || (ctx.aws?.s3?.object?.key != null && ctx.aws.s3.object.key.contains('aidmaster')))
```
navnit-elastic, line 4: This condition consider both **logfile** and **aws-s3** inputs.

packages/crowdstrike/data_stream/fdr/_dev/test/system/test-route-metadata-config.yml
```diff
@@ -0,0 +1,18 @@
...
+  - process.command_line
+data_stream:
+  vars:
+    access_key_id: "{{AWS_ACCESS_KEY_ID}}"
+    secret_access_key: "{{AWS_SECRET_ACCESS_KEY}}"
+    session_token: "{{AWS_SESSION_TOKEN}}"
+    queue_url: "{{TF_OUTPUT_queue_url}}"
+    preserve_original_event: true
+    enable_deduplication: true
+    enrich_metadata: false
```
navnit-elastic, line 15: `enrich_metadata: false` tells the Agent not to enrich metadata, so aidmaster and 
userinfo events eventually reach Elasticsearch. Then `route_metadata_to_datastreams: true` 
routes them to separate data streams.

navnit-elastic:
Pinging @elastic/security-service-integrations (Team:Security-Service Integrations)

packages/crowdstrike/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.2.0"
+  changes:
+    - description: Add option to route FDR aidmaster (host metadata) and userinfo (user metadata) events to separate data streams (fdr_aidmaster and fdr_userinfo).
```
kcreddy, line 4:
```suggestion
    - description: Add option to route FDR aidmaster (host metadata) and userinfo (user metadata) events to separate data streams (`fdr_aidmaster` and `fdr_userinfo` respectively).
```

packages/crowdstrike/data_stream/fdr/_dev/test/system/test-route-metadata-config.yml
kcreddy: Irrespective of whether `enrich_metadata` is true or false, `route_metadata_to_datastreams: true` always routes them to separate data streams, correct?

packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -43,6 +43,13 @@ processors:
             tag: append_error_message_4ef54c75
             field: error.message
             value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  # Conditionally append routing tag to route metadata to separate data streams.
+  - append:
+      tag: append_route_metadata_tag_1adc66f4
+      if: ctx._conf?.route_metadata_to_datastreams == true
+      field: tags
+      value: route_metadata_to_datastreams
+      allow_duplicates: false
```
kcreddy, line 52: Also add a test for this

packages/crowdstrike/data_stream/fdr/_dev/test/system/test-route-metadata-config.yml
navnit-elastic: No, that is not correct. However, if `keep_metadata: true` then your statement is correct.

packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
navnit-elastic: Are you referring to the pipeline test? I believe this should be covered by the system test added in this PR.

packages/crowdstrike/data_stream/fdr/_dev/test/system/test-route-metadata-config.yml
kcreddy: Thats what I meant when I wrote it, possibly missed `keep_metadata: true` part.
`enrich_metadata: false` isn't the only case that aidmaster and userinfo events eventually reach Elasticsearch. 
With `enrich_metadata: true` + `keep_metadata: true`, they can also get indexed.

packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
kcreddy: Right

packages/crowdstrike/data_stream/fdr/_dev/test/system/test-route-metadata-config.yml
navnit-elastic: The test scenario represents the ideal configuration we would expect for routing metadata to separate data streams.

kcreddy: Right, thats confirmed.

navnit-elastic:
> I didn't realize that we could not directly write to an index.mode: lookup index. I thought that we were going to be able to split the data and route metadata directly to a lookup index and keep events to a data stream.

Yes, because the `LOOKUP JOIN` has left-join behavior, and If many rows in the lookup index match, it adds one row per match. We would need to keep one event per host in the metadata index.

> So if we still have to use a transform to populate/create a lookup mode index, then do we even need to create intermediary data streams for host and user metadata? Wouldn't it be simpler to set up the transform with a DSL query that only processes the metadata docs?

I hadn't thought of this possibility. I agree, adding a transform with a DSL query on the existing `fdr` data stream to populate a lookup mode index would be simpler than creating separate data streams.

navnit-elastic:
We are discussing a transform based approach mentioned in Andrew's comment above. so we might not need this change in the integration.

navnit-elastic:
The change in https://github.com/elastic/integrations/pull/17877 allows users to enrich host metadata at query-time using transform and ES|QL LOOKUP JOINS. Hence, closing this request.

PR #16618, opened 2025-12-18
[Azure Logs / Custom Azure Logs] Switch default processor to v2

zmoog:
## Proposed commit message

Makes the processor v2 the default processor for all integrations in the Azure Logs and Custom Azure Logs packages. The processor v1 is still available as option, but it's listed as "legacy". 

The processor v1 uses deprecated libraries and it will be removed in the next release (targeting 9.4).

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/15703

### DIFF
--- a/packages/azure/_dev/build/docs/events.md
+++ b/packages/azure/_dev/build/docs/events.md
@@ -609,7 +609,11 @@ Possible values are `earliest` and `latest`.
 
 `migrate_checkpoint` :
 _boolean_
-(processor v2 only) Controls processor behavior upon the initial transition from v1 to v2. Defaults to `true`, which means the processor will attempt to migrate checkpoint information from the v1 format to v2. If set to `false`, the processor will bypass existing v1 checkpoints and replay all available events in the Event Hub from the start of the retention window. For example, if set to `false` on an Event Hub with a 1-hour retention period, the processor will replay the last hour's worth of data.
+(processor v2 only) Controls processor behavior upon the initial transition from v1 to v2. If you are starting directly with v2, you can disregard this setting.
+
+Defaults to `true`, which means the processor will automatically migrate checkpoint information from the v1 format to v2.
+
+If set to `false`, the processor will bypass existing v1 checkpoints and replay all events in the Event Hub from the start of the retention window. For example, if set to `false` on an Event Hub with a 1-hour retention period, the processor will replay the last hour's worth of data.
 
 `endpoint_suffix` :
 _string_
--- a/packages/azure/docs/events.md
+++ b/packages/azure/docs/events.md
@@ -609,7 +609,11 @@ Possible values are `earliest` and `latest`.
 
 `migrate_checkpoint` :
 _boolean_
-(processor v2 only) Controls processor behavior upon the initial transition from v1 to v2. Defaults to `true`, which means the processor will attempt to migrate checkpoint information from the v1 format to v2. If set to `false`, the processor will bypass existing v1 checkpoints and replay all available events in the Event Hub from the start of the retention window. For example, if set to `false` on an Event Hub with a 1-hour retention period, the processor will replay the last hour's worth of data.
+(processor v2 only) Controls processor behavior upon the initial transition from v1 to v2. If you are starting directly with v2, you can disregard this setting.
+
+Defaults to `true`, which means the processor will automatically migrate checkpoint information from the v1 format to v2.
+
+If set to `false`, the processor will bypass existing v1 checkpoints and replay all events in the Event Hub from the start of the retention window. For example, if set to `false` on an Event Hub with a 1-hour retention period, the processor will replay the last hour's worth of data.
 
 `endpoint_suffix` :
 _string_


packages/azure/_dev/build/docs/events.md
```diff
@@ -47,16 +47,16 @@ The integration v2 preview avoids contention and inefficiencies from using multi
 
 ### Event Hub Processor v2 ✨
 
-The integration v2 preview offers a new processor v2 starting with integration version 1.23.0.
+As of version 1.23.0, the integration preview includes the new Processor v2.
```
tommyers-elastic, line 50: what does "the integration preview" mean?

packages/azure/_dev/build/docs/events.md
```diff
@@ -47,16 +47,16 @@ The integration v2 preview avoids contention and inefficiencies from using multi
...
-The processor v2 introduces several changes:
+Processor v2 introduces several key improvements:
 
-* It uses the latest Event Hubs SDK from Azure.
-* It uses a more efficient checkpoint store based on Azure Blob Storage metadata.
+* Built on the latest Azure Event Hubs SDK.
+* Features a more efficient checkpoint store leveraging Azure Blob Storage metadata.
 
-The processor v2 is in preview. Processor v1 is still the default and is recommended for typical use cases.
+Processor v2 is now the default and is recommended for most use cases. Processor v1 remains available as a legacy option for backward compatibility until the next major release.
```
tommyers-elastic: > _most_ use cases

is there somewhere we can link that details when it's not recommended?

packages/azure/_dev/build/docs/events.md
```diff
@@ -575,7 +575,7 @@ Possible values are `earliest` and `latest`.
 _boolean_
 (processor v2 only) Flag to control whether the processor should perform the checkpoint information migration from v1 to v2 at startup. The checkpoint migration converts the checkpoint information from the v1 format to the v2 format.
 
-Default is `false`, which means the processor will not perform the checkpoint migration.
+Default is `true`, which means the processor will attempt to migrate the checkpoint information from v1 to v2 the first time you v1 to v2.
```
tommyers-elastic: what happens if you have `processor_version: v2` and `processor_update_interval: false`?

packages/azure/data_stream/activitylogs/agent/stream/azure-eventhub.yml.hbs
```diff
@@ -19,10 +19,30 @@ storage_account: {{storage_account}}
 {{/if}}
 {{#if storage_account_key}}
 storage_account_key: {{storage_account_key}}
+storage_account_connection_string: DefaultEndpointsProtocol=https;AccountName={{storage_account}};AccountKey={{storage_account_key}};EndpointSuffix={{endpoint_suffix}}
```
tommyers-elastic, line 22: nit: i'm guessing `storage_account` shouldn't be empty if `storage_account_key` is set, but since it's not in the `{{#if storage_account}}` there is a chance it's unset. 

same with `endpoint_suffix`

packages/azure/data_stream/activitylogs/manifest.yml
```diff
@@ -57,6 +57,105 @@ streams:
         type: bool
         multi: false
         default: false
+      #
+      # Processor v2 only settings
+      #
+      - name: processor_version
```
tommyers-elastic, line 63: are we ok that integration _upgrades_ are going to automatically enable v2 without the user explicitly choosing?

packages/azure/_dev/build/docs/events.md
zmoog: We introduced the new `events` data stream as the entry point and router for the Azure Logs integration package.

I plan to make the `events` data stream (aka Azure Logs v2) as the only data stream running an input.

Unfortunately, years ago we updated the Azure Logs packages to allow multiple inputs to read from the same event hub, but this is a fatal flaw. We need to get rid of this.

The other data streams (activitylogs, firewall_logs, etc) are going to stay without an input.

zmoog: We marked this `events` data stream named Azure Logs v2 as "preview" because we were not 100% sure this approach would work. But it seems working pretty well.

packages/azure/_dev/build/docs/events.md
zmoog: Um, no, there aren't known use cases. Maybe change this to only inform users v1 is still available for some time?

packages/azure/_dev/build/docs/events.md
zmoog: > and processor_update_interval: false?

I assume you mean `migrate_checkpoint: false`.

In that case, the input will start over from the beginning of the stream. If the retention is set to 1 hour, the input will re-process the last hour's worth of events.

I think we should make this explicit in the description rather than leaving it implied, right?

packages/azure/data_stream/activitylogs/manifest.yml
zmoog: We need to push users to the processor v2 while v1 is still available as fallback option. So if v2 falls short for any reason, users can still fallback to v1 while we address the problem.

We need to drop v1[^1] as soon as possible, since the v1's libraries are so old that sticking with them is becoming a risk.


[^1]: drop v1 as an option and remove it from the Beats codebase.

packages/azure/_dev/build/docs/events.md
zmoog: I clarified the description. Let me know if it makes sense.

packages/azure/_dev/build/docs/events.md
zmoog: @tommyers-elastic, I’ve fleshed out the `migrate_checkpoint` description. I think this version is much more comprehensive.

tommyers-elastic: thanks!

packages/azure/data_stream/activitylogs/agent/stream/azure-eventhub.yml.hbs
zmoog: - The `storage_account` contains the storage account name; this option must be set for the input to work.
- The `storage_account_key` contains the auth for processor v1.
- The `storage_account_connection_string` contains the auth for processor v2

We'll have the opportunity to clean things up a little when we'll drop v1 in a couple of iterations.

zmoog:
@tommyers-elastic, let me know if there's anything else you'd like me to tweak!

tommyers-elastic (approved):

PR #13410, opened 2025-04-03
[CI] Update CI scripts

mrodm:
## Proposed commit message

Update CI scripts among the changes
- Some refactors to place all the code related to filtering out packages in the same function.
- Avoid creating kind clusters in CI build testing with Serverless.
- Remove some output:
   - List of files updated in the PR that affected to test packages. As example:
     ```
     [auth0] git-diff: check non-package files
     .buildkite/pipeline.yml
     dev/testsreporter/_static/summary.tmpl
     dev/testsreporter/builderror.go
     dev/testsreporter/dataerror.go
     dev/testsreporter/errorlinks.go
     dev/testsreporter/format_test.go
     dev/testsreporter/packageerror.go
     dev/testsreporter/reporter.go
     dev/testsreporter/reporter_test.go
     dev/testsreporter/testsreporter.go
     magefile.go
     ```

Example of the output shown before and after to check whether or not a package should be tested:
- Before
```
[DEBUG] Checking with commits: from: 'origin/main' to: '32e72850687167562d5f93b39f4c4b98e60aa333'
[bitwarden] Original commits: from 'origin/main' - to: '32e72850687167562d5f93b39f4c4b98e60aa333'
[bitwarden]: commits: from: 'origin/main' - to: '32e72850687167562d5f93b39f4c4b98e60aa333'
[bitwarden] git-diff: check non-package files
.buildkite/pipeline.yml
.buildkite/scripts/common.sh
dev/testsreporter/_static/summary.tmpl
dev/testsreporter/builderror.go
dev/testsreporter/dataerror.go
dev/testsreporter/errorlinks.go
dev/testsreporter/format_test.go
dev/testsreporter/packageerror.go
dev/testsreporter/reporter.go
dev/testsreporter/reporter_test.go
dev/testsreporter/testsreporter.go
magefile.go
[bitwarden] PR is affected: found non-package files
[fleet_server] Original commits: from 'origin/main' - to: '32e72850687167562d5f93b39f4c4b98e60aa333'
[fleet_server]: commits: from: 'origin/main' - to: '32e72850687167562d5f93b39f4c4b98e60aa333'
[fleet_server] git-diff: check non-package files
.buildkite/pipeline.yml
.buildkite/scripts/common.sh
dev/testsreporter/_static/summary.tmpl
dev/testsreporter/builderror.go
dev/testsreporter/dataerror.go
dev/testsreporter/errorlinks.go
dev/testsreporter/format_test.go
dev/testsreporter/packageerror.go
[...]

PR #16622, opened 2025-12-18
[CI] Avoid reading secret from vault in steps testing packages - local stacks

mrodm:
## Proposed commit message

After the refactor in Buildkite scripts done in #13410 , `BUILDKITE_API_TOKEN` is not used in Buildkite stesps with id `test-integrations-*`.


mrodm:
Merging since the CI failure is unrelated to this change.

PR #16657, opened 2025-12-21
[axonius][network] Add Axonius Network datastream

muskan-agarwal26:
## Proposed commit message

The release includes network data stream and associated dashboard.

Axonius fields are mapped to their corresponding ECS fields where possible.

sanitized.

## Checklist

To test the axonius package:

- Clone integrations repo.
- Move to integrations/packages/axonius directory.

> elastic-package test
```
2025/12/21 16:36:24  INFO New version is available - v0.117.1. Download from: https://github.com/elastic/elastic-package/releases/tag/v0.117.1
2025/12/21 16:36:25  INFO License text found in "/root/GITHUB/integrations/LICENSE.txt" will be included in package
--- Test results for package: axonius - START ---
╭─────────┬─────────────┬───────────┬──────────────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME                                                        │ RESULT │ TIME ELAPSED │
├─────────┼─────────────┼───────────┼──────────────────────────────────────────────────────────────────┼────────┼──────────────┤
│ axonius │             │ asset     │ dashboard axonius-a47bc47d-5f69-473f-92a1-07a79cc71cf8 is loaded │ PASS   │      1.847µs │
│ axonius │ network     │ asset     │ index_template logs-axonius.network is loaded                    │ PASS   │        256ns │
│ axonius │ network     │ asset     │ ingest_pipeline logs-axonius.network-0.1.0 is loaded             │ PASS   │        275ns │
╰─────────┴─────────────┴───────────┴──────────────────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: axonius - END   ---
Done
--- Test results for package: axonius - START ---
╭─────────┬─────────────┬───────────┬────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME                                          │ RESULT │ TIME ELAPSED │
├─────────┼─────────────┼───────────┼────────────────────────────────────────────────────┼────────┼──────────────┤
│ axonius │ network     │ pipeline  │ (ingest pipeline warnings test-load-balancer.log)  │ PASS   │ 762.895972ms │
│ axonius │ network     │ pipeline  │ (ingest pipeline warnings test-network-device.log) │ PASS   │  693.13953ms │
│ axonius │ network     │ pipeline  │ test-load-balancer.log                             │ PASS   │ 233.874266ms │
│ axonius │ network     │ pipeline  │ test-network-device.log                            │ PASS   │ 347.384569ms │
╰─────────┴─────────────┴───────────┴────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: axonius - END   ---
Done
--- Test results for package: axonius - START ---
No test results
--- Test results for package: axonius - END   ---
Done
--- Test results for package: axonius - START ---
╭─────────┬─────────────┬───────────┬──────────────────────────┬────────┬──────────────╮
│ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME                │ RESULT │ TIME ELAPSED │
├─────────┼─────────────┼───────────┼──────────────────────────┼────────┼──────────────┤
│ axonius │ network     │ static    │ Verify sample_event.json │ PASS   │  249.74697ms │
╰─────────┴─────────────┴───────────┴──────────────────────────┴────────┴──────────────╯
--- Test results for package: axonius - END   ---
Done
2025/12/21 16:36:34  INFO Installing package...
2025/12/21 16:36:35  INFO License text found in "/root/GITHUB/integrations/LICENSE.txt" will be included in package
2025/12/21 16:36:49  INFO Running test for data_stream "network" with configuration 'default'
2025/12/21 16:36:57  INFO Setting up independent Elastic Agent...
2025/12/21 16:37:08  INFO Setting up service...
2025/12/21 16:37:33  INFO Validating test case...
2025/12/21 16:37:35  INFO Tearing down service...
2025/12/21 16:37:36  INFO Write container logs to file: /root/GITHUB/integrations/build/container-logs/axonius-1766315256251921752.log
2025/12/21 16:37:38  INFO Tearing down agent...
2025/12/21 16:37:39  INFO Write container logs to file: /root/GITHUB/integrations/build/container-logs/elastic-agent-1766315259135394865.log
2025/12/21 16:39:21  INFO Uninstalling package...
--- Test results for package: axonius - START ---
╭─────────┬─────────────┬───────────┬───────────┬────────┬───────────────╮
│ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT │  TIME ELAPSED │
├─────────┼─────────────┼───────────┼───────────┼────────┼───────────────┤
│ axonius │ network     │ system    │ default   │ PASS   │ 46.195212588s │
╰─────────┴─────────────┴───────────┴───────────┴────────┴───────────────╯
--- Test results for package: axonius - END   ---
Done
``` 

## Related issues

- Relates https://github.com/elastic/integrations/issues/15034
- Relates https://github.com/elastic/elasticsearch/pull/139512

## Screenshots
image (5)
image (6)


packages/axonius/data_stream/network/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,142 @@
...
+                  "asset_type": string(state.worklist.asset_type_list[0]),
+                },
+              }.encode_json(),
+            },
+          ],
+          "worklist": {
+            "asset_type_list": tail(state.worklist.asset_type_list),
+            ?"next_page": optional.none(),
+          },
+          "want_more": size(tail(state.worklist.asset_type_list)) > 0,
```
efd6: This is an expensive way to say `size(state.worklist.asset_type_list) > 1`; the `tail` call allocates the list, and then we throw it away.

packages/axonius/data_stream/network/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,142 @@
...
+redact:
+  fields:
+    - api_key
+    - secret_key
+program: |
+  (
+    state.?worklist.asset_type_list[0].hasValue() ?
+      state
+    :
+      state.drop("worklist").with(
```
efd6: What is the `drop` for? We are replacing the `worklist` field always with the `with` call?

packages/axonius/data_stream/network/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,142 @@
...
+                      string(resp.Status) + " (" + string(resp.StatusCode) + ")"
+                  ),
+                  "asset_type": string(state.worklist.asset_type_list[0]),
+                },
+              }.encode_json(),
+            },
+          ],
+          "worklist": {
+            "asset_type_list": tail(state.worklist.asset_type_list),
+            ?"next_page": optional.none(),
```
efd6: This is a no-op; are you sure this is what you mean? I think you just want to have this line deleted.

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,370 @@
...
+      type: long
+      ignore_missing: true
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - script:
+      tag: script_rename_event_data_fields
+      lang: painless
+      description: Renames all event.data.* fields to root level (removes event.data prefix). Prefixes conflicting keys with 'data_' to avoid overwriting root-level fields.
```
efd6:
```suggestion
      description: Renames all axonius.network.event.data.* fields to root level (removes event.data prefix). Prefixes conflicting keys with 'data_' to avoid overwriting custom data-stream-root-level fields.
```

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/pipeline-nat-rule.yml
```diff
@@ -0,0 +1,224 @@
...
+      ignore_empty_value: true
+  - convert:
+      field: axonius.network.is_enabled
+      tag: convert_is_enabled_to_boolean
+      type: boolean
+      ignore_missing: true
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```
efd6, line 59: Same as https://github.com/elastic/integrations/pull/16657/changes#r3000751036 here.

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/pipeline-network-device.yml
```diff
@@ -0,0 +1,2272 @@
...
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - convert:
+      field: axonius.network._keep_hostname_empty
+      tag: convert__keep_hostname_empty_to_boolean
+      type: boolean
+      ignore_missing: true
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```
efd6, line 52: Also here and below.

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/pipeline-network.yml
```diff
@@ -0,0 +1,204 @@
...
+      ignore_empty_value: true
+  - convert:
+      field: axonius.network.mtu
+      tag: convert_mtu_to_long
+      type: long
+      ignore_missing: true
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```
efd6, line 47: Also here and below.

packages/axonius/kibana/dashboard/axonius-a47bc47d-5f69-473f-92a1-07a79cc71cf8.json
efd6, line 1: Do we want a link panel?

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/pipeline-network-device.yml
```diff
@@ -0,0 +1,2272 @@
...
+      field: axonius.network.plugin_and_severities
+      tag: foreach_plugin_and_severities_severity
+      if: ctx.axonius?.network?.plugin_and_severities instanceof List
+      processor:
+        append:
+          field: vulnerability.severity
+          tag: append_plugin_and_severities_severity_into_vulnerability_severity
+          value: '{{{_ingest._value.severity}}}'
+          allow_duplicates: false
+  - foreach:
```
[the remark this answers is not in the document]
efd6: It is not possible to include an `if` condition in a processor child of a `foreach` processor, elastic/elasticsearch#60470.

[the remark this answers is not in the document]
efd6: Sure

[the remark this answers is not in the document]
efd6: We'll leave this for the author.

packages/axonius/data_stream/network/manifest.yml
```diff
@@ -0,0 +1,87 @@
...
+        type: text
+        title: Asset Types
+        description: List of asset types to collect from the Axonius API. Do not modify.
+        multi: true
+        required: true
+        show_user: false
+        default: 
+          - networks
+          - load_balancers
+          - network_services
```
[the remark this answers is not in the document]
muskan-agarwal26: This asset type mapping is handled in default.yml, so we do not require any separate pipeline for this.

packages/axonius/data_stream/network/fields/fields.yml
```diff
@@ -0,0 +1,1289 @@
...
+          type: group
+          fields:
+            - name: creation_timestamp
+              type: date
+            - name: gateway_address
+              type: ip
+            - name: id
+              type: keyword
+            - name: ip_cidr_range
+              type: ip
```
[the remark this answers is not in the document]
muskan-agarwal26: In live logs, we get values like : `44.177.201.219`, `57.227.181.44`, `92.185.11.38`, hence kept `ip` as type.

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/pipeline-nat-rule.yml
muskan-agarwal26: We haven't mapped this with any ecs field, do you think still we have to write `remove` proc for this?

packages/axonius/kibana/dashboard/axonius-a47bc47d-5f69-473f-92a1-07a79cc71cf8.json
muskan-agarwal26: Yes, Link panel will be created once all data stream PRs are merged.
There are more 3 in the queue.

packages/axonius/data_stream/network/agent/stream/cel.yml.hbs
muskan-agarwal26: This is to ensure there is no past data in worklist while going for the first call after interval, hence creating a new worklist inside state.

packages/axonius/data_stream/network/agent/stream/cel.yml.hbs
muskan-agarwal26: This line was added to handle cases where the API fails for an item in the worklist during pagination. In such situations, the process continues with the next item, and we should not retain any page value in next_page; otherwise, it could be incorrectly used to fetch data for next item in list, which is not intended.

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/pipeline-network-device.yml
```diff
@@ -0,0 +1,2272 @@
...
+      field: axonius.network.associated_device_users
+      tag: foreach_associated_device_users_is_latest_used_user
+      if: ctx.axonius?.network?.associated_device_users instanceof List
+      processor:
+        append:
+          field: related.user
+          tag: append_associated_device_users_is_latest_used_user_into_related_user
+          value: '{{{_ingest._value.is_latest_used_user}}}'
+          allow_duplicates: false
+  - foreach:
```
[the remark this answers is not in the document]
muskan-agarwal26: Removed the mapping, as not required.

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/pipeline-network-device.yml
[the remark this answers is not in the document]
muskan-agarwal26: Added line of code in script processor to handle this scenario.

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/pipeline-nat-rule.yml
kcreddy: I think we should in order to avoid mapping conflict on the field

packages/axonius/data_stream/network/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,142 @@
...
+            "api-key": [state.api_key],
+            "api-secret": [state.secret_key],
+          },
+          "Body": {
+            "include_metadata": true,
+            "page": {
+              "limit": state.batch_size,
+            },
+            ?"next_page": state.?worklist.?next_page,
+            "fields": ["specific_data"],
```
kcreddy, line 61: What other options exist here other than `specific_data`, does it makes sense to make it configurable for users?

packages/axonius/elasticsearch/transform/latest_network/transform.yml
```diff
@@ -0,0 +1,37 @@
...
+  aliases:
+    - alias: 'logs-axonius_latest.network'
+      move_on_creation: true
+latest:
+  unique_key:
+    - event.dataset
+    - axonius.network.transform_unique_id
+  sort: '@timestamp'
+description: >-
+  Latest networks from Axonius. As networks get updated, this transform stores only the latest state of each network asset inside the destination index. Thus the transform's destination index contains only the latest state of the network asset.
```
kcreddy: If we are trying to get latest network assets into destination index, don't we need to use something like an `assetID` as a unique key instead of fingerprint on `event.original` (ie., `transform_unique_id`)?

Have we tested a scenario: update `asset` -> 2 documents in source, and 1 document in destination?

packages/axonius/manifest.yml
kcreddy, line 12: I think we need to add more categories based on the data streams. Please check [this enum list](https://github.com/elastic/package-spec/blob/main/spec/integration/manifest.spec.yml#L21) and add accordingly for all data streams that are added until now.

packages/axonius/data_stream/network/lifecycle.yml
```diff
@@ -0,0 +1 @@
+data_retention: '90d'
```
kcreddy: The data retention of `90d` on source may be on the high side considering we already have a `90d` retention on the destination side. Is this already discussed somewhere and finalised?

muskan-agarwal26: Nope, changing it to 30d.

packages/axonius/elasticsearch/transform/latest_network/transform.yml
muskan-agarwal26: **There is no unique ID or other unique field available in the document that could be used for the transform, so `event.original` was used instead. This approach was also discussed earlier with you on Slack.**

kcreddy: With this approach, I am curious how the dashboards will look like.
For example `Assets by Status` visualisation: say an asset changes its state from `Available` to `Deployed`. It now has 2 documents in the destination index (1 with `Available` state and 1 with `Deployed` state). 
How do you show only the latest state?

muskan-agarwal26: This scenario can be managed using the transform retention period, as there is no field available that indicates when a log has been updated. If you have any other approach in mind to handle this case, please let me know.

packages/axonius/data_stream/network/agent/stream/cel.yml.hbs
muskan-agarwal26: `specific_data` returns the complete set of asset details. There are no other configurable fields available for users to fetch this data. While filters such as `specific_data.date` exist, they only return data for that specific field or object, whereas `specific_data` itself provides the full set of related details.

packages/axonius/elasticsearch/transform/latest_network/transform.yml
kcreddy: The retention approach will eventually work, but remain stale until retention gets triggered.
I am looking at the `config.yml` used for system tests. Since we are splitting on `assets[].specific_data[]`, have you tried a unique id that is a combination:
- `assets[].internal_axon_id` + `assets[].specific_data[].id` OR
- `assets[].internal_axon_id` + `assets[].specific_data[].id_raw`
That may seem to give us a unique identifier for each network asset.

muskan-agarwal26: Any of both fields below do not guarantee to populate in each log of all assets., there is no schema provided for the assets' `specific_data` details.
`assets[].specific_data[].id`, `assets[].specific_data[].id_raw`

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,390 @@
...
+  - foreach:
+      field: axonius.network.source_addresses
+      tag: foreach_event_data_source_addresses
+      if: ctx.axonius?.network?.source_addresses instanceof List
+      processor:
+        append:
+          field: related.ip
+          tag: append_event_data_source_addresses_into_related_ip
+          value: '{{{_ingest._value}}}'
+          allow_duplicates: false
```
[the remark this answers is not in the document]
muskan-agarwal26: This has been tested: when an invalid value appears in `source_addresses`, the field is removed, and neither `related.ip` nor `source.address` are created.
Hence, no need for any change here.

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/pipeline-network.yml
```diff
@@ -0,0 +1,210 @@
...
+          tag: convert_subnetworks_ip_cidr_range_to_ip
+          type: ip
+          ignore_missing: true
+          on_failure:
+            - remove:
+                field: _ingest._value.ip_cidr_range
+                ignore_missing: true
+            - append:
+                field: error.message
+                value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```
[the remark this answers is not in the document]
muskan-agarwal26: In live logs, we get values like : `44.177.201.219`, `57.227.181.44`, `92.185.11.38`, hence kept `ip` as type.

packages/axonius/elasticsearch/transform/latest_network/transform.yml
kcreddy: Without a proper unique `id` of the entity, I am not sure what we are even finding the latest of.
For example, your previously created [`latest_user`](https://github.com/elastic/integrations/blob/feature/axonius-0.1.0/packages/axonius/elasticsearch/transform/latest_user/transform.yml#L10-L13) is correct. It will correctly collapse an entity's old state when its state changes. This PR's network transforn won't.

When the unique key is a hash of the full document content, there is a clear problem here:
- Same asset, any field changed (status, IP, name, anything) -- different hash -- transform treats old and new as separate entries and keeps both

Only case it works correctly:
- Same asset, unchanged data -- same hash -- transform deduplicates correctly

If truly no stable identifier exists for network assets, the honest thing would be to either:
1. Reach out to vendor for clarifications in schema.
2. Acknowledge the transform only deduplicates identical documents (and adjust the description and dashboard expectations accordingly), 
3. Is `latest` transform even required when it can only solve the 1 case -- Same asset, unchanged data

cc: @ShourieG @piyush-elastic

muskan-agarwal26: I confirmed with the client that users can only delete assets, not update them. In that case, I believe the `latest` transform approach would work here. Am I correct? The deleted asset should also be removed from the transform after the retention period.

kcreddy: > I confirmed with the client that users can only delete assets, not update them

If thats the case, it changes things. The transform works in the sense that it correctly collapses repeated fetches into one document per entity inside the destination index. But it is fragile to API response stability. If the API response changes its structure - add a new field, change its date format, etc. the fingerprint changes and leads to duplicate documents for the same entity. 

I am curious on the `90d` retention approach. If the entity gets deleted, are we deleting it `90 days` after in our destination index?

packages/axonius/data_stream/network/agent/stream/cel.yml.hbs
efd6: This is not a correct assessment; in the general case, `x.drop(f).with({f: v})` is identical to `x.with({f: v})`.
```
-- src.cel --
[
	state.drop("field").with({"field": "newdata"}),
	state.with({"field": "newdata"}),
]
-- data.json --
{
	"field": "data"
}
-- out.json --
[
	{
		"field": "newdata"
	},
	{
		"field": "newdata"
	}
]
```

packages/axonius/data_stream/network/agent/stream/cel.yml.hbs
efd6: This is not correct; an assignment of `optional.none()` to an optional field leads to an absent field, that is `{?f: optional.none()}` is identical to `{}`.

```
-- src.cel --
{?"field": optional.none()}
-- out.json --
{}
```

packages/axonius/elasticsearch/transform/latest_network/transform.yml
muskan-agarwal26: Yes, this is because, as confirmed by the client, deleted assets can be rediscovered and reappear in Axonius if they still exist in connected data sources and are detected in a future discovery cycle. Since we do not know the exact duration for which assets may be retained, we have configured a retention period of 90 days to be on the safer side.

kcreddy: Okay. Can you document this behaviour in the dashboard and also inside the README so that users are aware?

muskan-agarwal26: In Readme I can, but where shall I document in dashboard?

kcreddy: It should be in the `Overview`

kcreddy (commented):
Update README and Dashboard overview to reflect limitations discussed in https://github.com/elastic/integrations/pull/16657/changes#r3056389784

muskan-agarwal26:
I have added a note in both as discussed. @kcreddy

packages/axonius/_dev/build/docs/README.md
```diff
@@ -129,6 +138,9 @@ Destinations indices are aliased to `logs-axonius_latest.<data_stream_name>`.
 | `logs-axonius.user-*`              | `logs-axonius_latest.dest_user-*`                | `logs-axonius_latest.user`              |
 | `logs-axonius.storage-*`              | `logs-axonius_latest.dest_storage-*`                | `logs-axonius_latest.storage`              |
 | `logs-axonius.ticket-*`              | `logs-axonius_latest.dest_ticket-*`                | `logs-axonius_latest.ticket`
+| `logs-axonius.network-*`              | `logs-axonius_latest.dest_network-*`                | `logs-axonius_latest.network`
+
+**Note:** Assets deleted from Axonius may reappear in a future discovery cycle if they are still present in connected data sources and get re-detected. Because the exact duration for which a deleted asset may remain dormant before being rediscovered is unknown, the transform retention period is set to **90 days** to reduce the risk of data loss for such assets.
```
kcreddy: I think we are not capturing the real user-facing consequence. You should also add:
`This means deleted assets will continue to appear in dashboards for up to 90 days after deletion.`

kcreddy: The users should also understand the difference between this data stream and other datastreams such as `user`, `gateway`, etc. (which has unique entity id which actually contains latest state of the entity).

Add:
`The network destination index is a content-based deduplicated view, not an entity-level latest-state view like the other data streams.`

packages/axonius/elasticsearch/transform/latest_network/transform.yml
kcreddy: The description is misleading: `As networks get updated, this transform stores only the latest state of each network asset`
But you said as per the vendor, network assets are never updated, but only created and deleted.
The description should accurately reflect what the transform does now: deduplicates identical documents from repeated API fetches.

packages/axonius/elasticsearch/transform/latest_network/transform.yml
```diff
@@ -0,0 +1,37 @@
...
+    - 'logs-axonius.network-*'
+dest:
+  index: 'logs-axonius_latest.dest_network-1'
+  aliases:
+    - alias: 'logs-axonius_latest.network'
+      move_on_creation: true
+latest:
+  unique_key:
+    - event.dataset
+    - axonius.network.transform_unique_id
```
kcreddy, line 18: Add code comments explaining the choice of unique key.

- The Axonius API provides no stable unique identifier for network assets
- `transform_unique_id` is a fingerprint of `event.original` (the full document)
- Per vendor: network assets cannot be updated, only created and deleted
- Known fragility: if the API response format changes (field ordering, new fields, etc.), the fingerprint will differ, creating duplicate entries in the destination index

packages/axonius/data_stream/network/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,390 @@
...
+  - json:
+      field: event.original
+      tag: json_event_original
+      target_field: axonius.network
+  - fingerprint:
+      fields:
+        - event.original
+      tag: fingerprint_event_original
+      target_field: axonius.network.transform_unique_id
+      ignore_missing: true
```
kcreddy, line 49: Here, there should be explanation to why we fingerprint full document.
Add `description` or code comments noting that it exists because no stable entity ID is available from the API.

packages/axonius/_dev/build/docs/README.md
kcreddy: Also add the same to the dashboard overview.

kcreddy (approved):
LGTM, please wait for @efd6 approval.

efd6:
My concerns have not been addressed.

muskan-agarwal26:
@efd6 , I have addressed your comments, please have a look.

efd6 (approved):

PR #16681, opened 2025-12-23
aws_bedrock: fix handling of strings with braces in content fields

chemamartinez:
## Proposed commit message

```
aws_bedrock: fixing the parsing of field invocation.messages.content.content when it
contains a string surrounded by braces.

Error looks like:
{\"type\":\"document_parsing_exception\",\"reason\":\"[1:44051] failed to parse field 
[aws_bedrock.invocation.messages.content.content] of type [keyword] in document ...
,\"caused_by\":{\"type\":\"illegal_argument_exception\",\"reason\":\"Expected text at
1:29749 but found START_OBJECT\"}}, dropping event!"

The error is caused by a string starting with `{` which makes the parser expect a JSON object.

The fix consists on removing the trailing braces and create a JSON object with the expected structure.

Added a test case to show how the issue is replicated.
```

## Checklist

### DIFF
--- a/packages/aws_bedrock/changelog.yml
+++ b/packages/aws_bedrock/changelog.yml
@@ -2,7 +2,7 @@
   changes:
     - description: Fix handling of string elements with braces in invocation `messages.content.content` fields.
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/99999
+      link: https://github.com/elastic/integrations/pull/16681
 - version: "1.5.0"
   changes:
     - description: Tolerate non-object elements in invocation `messages.content` and `system` fields.


chemamartinez:
@chrisberkhout I have tried in version 8.18.4 which is the one where the error originally happened and the example case in Dev Tools work as well. It also works with a string starting with a brace, and with the same string that originally caused the index error.

So I decided to apply this fix based on the error message as I think something should be missing with the field format related to the error.

chrisberkhout:
Thinking about this some more... I think the incoming document must have had an object where a JSON string was expected.

Here's an example of that:
```
PUT my2ndexampleindex
{
  "mappings": {
    "properties": {
      "myfield": { "type": "keyword" }
    }
  }
}
POST my2ndexampleindex/_doc
{
  "myfield": {"k": "v"}
}
```

It produces this error:
```
{
  "error": {
    "root_cause": [
      {
        "type": "document_parsing_exception",
        "reason": "[1:24] failed to parse field [myfield] of type [keyword] in document with id 'sbj2T5sBs1ZNOKVFoK5c'. Preview of field's value: '{k=v}'"
      }
    ],
    "type": "document_parsing_exception",
    "reason": "[1:24] failed to parse field [myfield] of type [keyword] in document with id 'sbj2T5sBs1ZNOKVFoK5c'. Preview of field's value: '{k=v}'",
    "caused_by": {
      "reason": "Expected text at 1:15 but found START_OBJECT"
    }
  },
  "status": 400
}
```

This might have something to do with having array values for messages and for content, which those two processors before the change in the pipeline don't seem to handle.

PR #16718, opened 2025-12-29
Asset discovery enhancement: added entity group type as a field for users and hosts.

chennn1990:
## Checklist

## Related issues

- [Closes #123](https://github.com/elastic/kibana/issues/233603)
- [Relates #123](https://github.com/elastic/cloudbeat/pull/3791)

### DIFF
--- a/packages/cloud_asset_inventory/changelog.yml
+++ b/packages/cloud_asset_inventory/changelog.yml
@@ -8,7 +8,7 @@
   changes:
     - description: Added nested entity fields for user and host
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/15847
+      link: https://github.com/elastic/integrations/pull/16718
 - version: "1.2.3"
   changes:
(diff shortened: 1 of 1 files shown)


packages/cloud_asset_inventory/data_stream/asset_inventory/fields/ecs.yml
```diff
@@ -44,11 +44,41 @@
   external: ecs
 - name: host.mac
   external: ecs
+- name: host.entity
+  type: group
```
kubasobon, line 48: Is there a reason we're going with `type: group` for those two properties instead of full breakdown, like we do for `cloud.project.name` etc.?

chennn1990: i just took the same declaration as it appears under the `entity.yaml` file

PR #16742, opened 2025-12-31
[checkpoint_harmony_endpoint] Fix cursor data not updating

chrisberkhout:
## Proposed commit message

```
[checkpoint_harmony_endpoint] Fix cursor data not updating

When a single event is returned in order to indicate an error that should be
logged and update fleet health status, cursor data updates will be ignored[1].

Without successful cursor updates, this integration was never able to clear an
expired task ID (returning 404) from the cursor data.

The fix is to always return an array of events when cursor data needs to be
updated. The error event will continue to be ingested. It will not be
logged or update the fleet status.

All `cel.yml.hbs` files in this integration are identical.

[1]: https://github.com/elastic/beats/blob/11da96ab9ad2c3ca110ed99d16bbab0b89ddf582/x-pack/filebeat/input/cel/input.go#L492-L493
```

## Checklist

### DIFF
--- a/packages/checkpoint_harmony_endpoint/changelog.yml
+++ b/packages/checkpoint_harmony_endpoint/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.0.1"
+  changes:
+    - description: Fix cursor data not updating.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/16742
 - version: "1.0.0"
   changes:
     - description: Release package as GA.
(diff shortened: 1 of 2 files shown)


ShourieG (approved):
LGTM, only one question, should we ingest or drop the error events if we are just recovering ?

chrisberkhout:
> LGTM, only one question, should we ingest or drop the error events if we are just recovering ?

Recovery isn't guaranteed. If there was logic here for having limited retries and then skipping a time range if there are persistent errors there, then I'd probably only want to keep the final failure event. For now there's no change in what will be ingested, it'll just actually have a chance to recover.

Regarding how to do the failure events, the only thing I see in Fleet Package Code Review Comments is [Use a terminate processor with inputs that produce failure events](https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#use-a-terminate-processor-with-inputs-that-produce-failure-events) (but this integration allows ES before the terminate processor). Perhaps we could expand on that.

Issue #16514, opened 2025-12-11
[ServiceNow]: @timestamp field isn't being formatted and causing a "date_time_parse_exception"


TheRiffRafi:
### Integration Name

ServiceNow [servicenow]

### Integration Version

1.2.0

### Agent Version

8.19.4

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.19.4

### OS Version and Architecture

N/A

### Error Message

```json
{
  "error": {
    "root_cause": [
      {
        "type": "document_parsing_exception",
        "reason": "[1:1755] failed to parse field [@timestamp] of type [date] in document with id '57RRDpsBaLhj_JQCHW_x'. Preview of field's value: '2021-09-09 21:32:29'"
      }
    ],
    "type": "document_parsing_exception",
    "reason": "[1:1755] failed to parse field [@timestamp] of type [date] in document with id '57RRDpsBaLhj_JQCHW_x'. Preview of field's value: '2021-09-09 21:32:29'",
    "caused_by": {
      "type": "illegal_argument_exception",
      "reason": "failed to parse date field [2021-09-09 21:32:29] with format [strict_date_optional_time||epoch_millis]",
      "caused_by": {
        "type": "date_time_parse_exception",
        "reason": "Failed to parse with all enclosed parsers"
      }
    },
    "failure_store": "not_enabled"
  },
  "status": 400
}
```

### Event Original

```json
{
  "@timestamp": "2025-12-08T21:59:41.838Z",
  "@metadata": {
    "beat": "filebeat",
    "type": "_doc",
    "version": "8.19.4",
    "_id": "1765230525000000000-7ae1b06345-000000520865",
    "raw_index": "logs-servicenow.event-prod",
    "input_id": "aws-s3-servicenow-db272482-e001-48b5-916a-4635a82afd8f",
    "stream_id": "aws-s3-servicenow.event-db272482-e001-48b5-916a-4635a82afd8f"
  },
  "input": {
    "type": "aws-s3"
  },
  "_conf": {
    "tz_offset": "America/New_York",
    "table_name": "table1",
    "timestamp_field": "sys_updated_on"
  },
  "event": {
    "dataset": "servicenow.event"
  },
  "elastic_agent": {
    "snapshot": false,
    "version": "8.19.4",
    "id": "5b2d3828-6e37-46e5-b4a6-0399717dffe2"
  },
  "ecs": {
    "version": "8.0.0"
  },
[...]

PR #16884, opened 2026-01-06
[servicenow] Add Parsing for ECS Timestamp Field

mohitjha-elastic:
## Proposed commit message

```
servicenow: Add explicit date parsing for ECS @timestamp field.

Added explicit date parsing logic for the ECS @timestamp field to ensure consistent and accurate
timestamp handling during ingestion. Additionally, this ensures that the timestamp is explicitly parsed 
even when other date fields are not parsed, helping to avoid potential ingestion errors.
```

## Checklist

- Clone integrations repo.

> elastic-package test -v

## Related Issues

- Closes https://github.com/elastic/integrations/issues/16514

### DIFF
--- a/packages/servicenow/changelog.yml
+++ b/packages/servicenow/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Add explicit date parsing for the ECS '@timestamp' field.
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/1
+      link: https://github.com/elastic/integrations/pull/16884
 - version: "1.3.1"
   changes:
     - description: Fixed description for URL on the ServiceNow tables input.


packages/servicenow/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -3339,6 +3339,41 @@ processors:
...
+        - dd/MM/yyyy HH:mm:ss
+        - dd/MM/yyyy
+        - MM/dd/yy H:mm:ss
+        - MM/dd/yy HH:mm:ss
+        - MM/dd/yy
+        - dd.MM.yyyy H:mm:ss
+        - dd.MM.yyyy HH:mm:ss
+        - dd.MM.yyyy
+        - yyyy-MM-dd hh:mm:ss a
+        - ISO8601
```
ShourieG, line 3369: My concern here is with formats like 

- MM-dd-yyyy
- dd-MM-yyyy 

as these are ambiguous from a global timestamp perspective. In india it's common trend to use dd-MM-yyyy or dd/MM/yyyy or dd.MM.yyyy. But in many places we use mm-dd-yyyy and etc. 

How do we distinguish if the date 01-02-2026 or 05-07-2026 is in MM-dd-yyyy format or dd-MM-yyyy format ? Do we expect '@timestamp' to be normalised in any way ? If not how can we decide upon the right order of execution here ? If '@timestamp' is normalised as month-day-year, why have support for both patterns ?

efd6: Agree. This is a LOCALE issue and should be configurable if it is necessary to handle different formats. Given that the servicenow data sources are essentially a free-for-all, we cannot safely determine which to use without being told by the user during configuration.

> But in many places we use mm-dd-yyyy

Depending on the scope of "place"; this ordering is a US localisation.

mohitjha-elastic: The issue exists across the entire integration. All `date` processors currently support both `MM-dd-yyyy` and `dd-MM-yyyy` formats, while the preference is consistently set to `MM-dd-yyyy` across all processors.

I believe this is a separate concern, and we could open a new ticket to make the date format user-configurable instead of supporting multiple formats by default. Please let me know your thoughts, or if you think this should be addressed as part of the current PR.

efd6: I think it's reasonable to use this given that the pattern is already widespread. We should fix this in a follow-up though since currently dd-MM-yyyy dates in the source data are being incorrectly parsed when dd is not greater than 12. The problem is that currently the effective default LOCALE is for the US format, so we would need to maintain that, and non-US formats would be required to turn on their LOCALE preference to ensure that timestamps are parsed when they are not in dd ≤ 12. This would be a breaking change, unless we make it relaxed rather than strict (we keep both orders, but the configuration changes the order of appearance in the formats list).

mohitjha-elastic:
Here is the ticket to add the configurable date format handling date processors - 
https://github.com/elastic/integrations/issues/17286

PR #16886, opened 2026-01-06
[IIS] Application pool - Fix cpu_usage_perc unit_type

muthu-mps:
- Bug

## Proposed commit message

This fixes the incorrect unit type for **cpu_usage_perc**. Removing the unit type to make it consistent with other data stream CPU fields.

## Checklist

## Related issues

- Closes https://github.com/elastic/sdh-beats/issues/6791

### DIFF
--- a/packages/iis/data_stream/application_pool/fields/fields.yml
+++ b/packages/iis/data_stream/application_pool/fields/fields.yml
@@ -33,6 +33,7 @@
             Memory virtual bytes.
         - name: cpu_usage_perc
           type: float
+          unit: percent
           metric_type: gauge
           description: |
             The CPU usage percentage.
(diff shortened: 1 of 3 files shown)


packages/iis/data_stream/application_pool/fields/fields.yml
```diff
@@ -33,7 +33,6 @@
             Memory virtual bytes.
         - name: cpu_usage_perc
           type: float
-          unit: s
```
tommyers-elastic, line 36: `percent` is alllowed here, why not use that?

muthu-mps: This change focus on removing the unit type here. Yes, we can update to the percent type for `cpu_usage_perc` field in application_pol and for the Webserver datastream.

PR #16883, opened 2026-01-06
fix pipeline flaky test by adding regex to match dynamic date values

brijesh-elastic:
## Proposed commit message

```
fix pipeline flaky tests by adding regex to match dynamic date values.

Timestamps that are parsed without a year (such as those on BSD-style syslog messages) will have
their expected values inherit the year the expected files are generated in. This means that
tests will only pass in the year that the expected files are generated.

The relevant timestamp field (@timestamp, for example) has been added to the pipeline test config
as a dynamic field, and a regex pattern is used to match the expected format of the timestamp.

Fixes the tests for the following packages:
- haproxy
- pps
- sentinel_one_cloud_funnel
- trendmicro
- vsphere
- watchguard_firebox
- zscaler_zia
```

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally

- Clone integrations repo.
- Install elastic package locally.
- Start elastic stack using elastic-package.
- Move to integrations/packages/<integration_name> directory.
- Run the following command to run tests.
> elastic-package test

PR #16895, opened 2026-01-07
[statsd_input] [airflow] [envoyproxy] [vsphere] flaky test: increase initial delay before sending statsd (UDP) to elastic-agent

mykola-elastic:
Increasing delay to send UDP to elastic-agent from `5s` to `10s` as sometimes it takes longer than 5s for the elastic-agent to start listening (I was able to reproduce when it took 6 seconds, and the "service" container which emits statsd(UDP) experienced `connection refused` when trying to reach `elastic-agent:8125`)

Probably related to some of the receivers running in the otel collector: https://github.com/elastic/elastic-agent/blob/f684ab33d7f09ea685910dbafe4b814ee0a92db4/pkg/component/component.go#L76

## Proposed commit message

See title.

## Checklist

- [ ] ~I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.~
- [ ] ~I have added an entry to my package's `changelog.yml` file.~
- [ ] ~I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).~

## Related issues

- Closes https://github.com/elastic/integrations/issues/16890
- Closes https://github.com/elastic/integrations/issues/16736
- Closes https://github.com/elastic/integrations/issues/16703
- Closes https://github.com/elastic/integrations/issues/16654
- Closes https://github.com/elastic/integrations/issues/16613

### DIFF
--- a/packages/envoyproxy/_dev/deploy/docker/docker-compose.yml
+++ b/packages/envoyproxy/_dev/deploy/docker/docker-compose.yml
@@ -10,4 +10,4 @@ services:
     image: docker.elastic.co/observability/stream:v0.18.0
     volumes:
       - ./sample_logs:/sample_logs:ro
-    command: log --start-signal=SIGHUP --delay=5s --addr elastic-agent:8125 -p=udp /sample_logs/envoy-stats.log
+    command: log --start-signal=SIGHUP --delay=10s --addr elastic-agent:8125 -p=udp /sample_logs/envoy-stats.log
--- a/packages/vsphere/_dev/deploy/docker/docker-compose.yml
+++ b/packages/vsphere/_dev/deploy/docker/docker-compose.yml
@@ -11,7 +11,7 @@ services:
     image: docker.elastic.co/observability/stream:v0.18.0
     volumes:
       - ./sample_logs:/sample_logs:ro
-    command: log --start-signal=SIGHUP --delay=5s --addr elastic-agent:9514 -p=udp /sample_logs/syslog.log
+    command: log --start-signal=SIGHUP --delay=10s --addr elastic-agent:9514 -p=udp /sample_logs/syslog.log
   vsphere-log-tcp:
     image: docker.elastic.co/observability/stream:v0.18.0
     volumes:


packages/statsd_input/_dev/deploy/docker/docker-compose.yml
```diff
@@ -3,4 +3,4 @@ services:
     image: docker.elastic.co/observability/stream:v0.18.0
     volumes:
       - ./sample_logs:/sample_logs:ro
-    command: log --start-signal=SIGHUP --delay=5s --addr elastic-agent:8125 -p=udp /sample_logs/test-statsd.log
+    command: log --start-signal=SIGHUP --delay=15s --addr elastic-agent:8125 -p=udp /sample_logs/test-statsd.log
```
mykola-elastic:
```suggestion
    command: log --start-signal=SIGHUP --delay=10s --addr elastic-agent:8125 -p=udp /sample_logs/test-statsd.log
```

packages/airflow/_dev/deploy/docker/docker-compose.yml
```diff
@@ -4,4 +4,4 @@ services:
     image: docker.elastic.co/observability/stream:v0.18.0
     volumes:
       - ./sample_logs:/sample_logs:ro
-    command: log --start-signal=SIGHUP --delay=5s --addr elastic-agent:8125 -p=udp /sample_logs/test-airflow.log
+    command: log --start-signal=SIGHUP --delay=15s --addr elastic-agent:8125 -p=udp /sample_logs/test-airflow.log
```
mykola-elastic:
```suggestion
    command: log --start-signal=SIGHUP --delay=10s --addr elastic-agent:8125 -p=udp /sample_logs/test-airflow.log
```

packages/airflow/_dev/deploy/docker/docker-compose.yml
```diff
@@ -4,4 +4,4 @@ services:
     image: docker.elastic.co/observability/stream:v0.18.0
     volumes:
       - ./sample_logs:/sample_logs:ro
-    command: log --start-signal=SIGHUP --delay=5s --addr elastic-agent:8125 -p=udp /sample_logs/test-airflow.log
+    command: log --start-signal=SIGHUP --delay=10s --addr elastic-agent:8125 -p=udp /sample_logs/test-airflow.log
```
shmsr, line 7: Can we add retries too? Just in case 10s delay also fails?

https://github.com/elastic/integrations/blob/7fa4389c0e0b63e94bf45c83f9dbbfe70572641f/packages/azure_network_watcher_vnet/_dev/deploy/docker/docker-compose.yml#L14

^ like here?

I mean, let's keep the delay `5s` and add retries >= 5/10

So that even if agent start after 5-6s (<10s), we do not have to wait for 10s everytime?

packages/airflow/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.0.1"
```
shmsr: We do not need a changelog right?

packages/airflow/manifest.yml
```diff
@@ -1,6 +1,6 @@
 name: airflow
 title: Airflow
-version: "1.0.0"
+version: "1.0.1"
```
shmsr: No upgrades if this just solving flaky tests.

packages/airflow/_dev/deploy/docker/docker-compose.yml
shmsr: I see there's one more helpful flag: https://github.com/elastic/stream/blob/5e820a107e6a43372e26b57da07e75e2151b09ea/internal/command/root.go#L67

mykola-elastic: > rootCmd.PersistentFlags().IntVar(&opts.Retries, "retry", 10, "connection retry attempts for tcp based protocols")

- says for TCP based protocols (we have `-p=udp`). Looks like we have to increase the delay.

> I see there's one more helpful flag:

- it uses `--start-signal=SIGHUP`, but the server listen happens a few seconds after this signal

packages/airflow/changelog.yml
mykola-elastic: yeah, I just didn't find a corresponding label  (`skip-changelog` or similar). Won't CI break?

shmsr: See: https://github.com/elastic/integrations/pull/16883

No CL

packages/airflow/_dev/deploy/docker/docker-compose.yml
shmsr: Ah, true. But weird that it is only from TCP; from GH the source seems to be protocol agnostic but did not check properly. But leave it. I will approve; but yes remove the CL and version update.

packages/airflow/changelog.yml
mykola-elastic: Thanks!

Issue #16640, opened 2025-12-19
[qualys_vmdr]: Deprecation of API 2.0/3.0 for Knowledge Base


clement-fouque:
### Integration Name

Qualys VMDR [qualys_vmdr]

### Dataset Name

qualys_vmdr.asset_host_detection and qualys_vmdr.knowledge_base

### Integration Version

6.14.1

### Agent Version

9.2.2

### OS Version and Architecture

Docker

### User Goal

Knowledge Base API 2.0 and 3.0 are end of support in December 2025 and End Of Life in June 2026. We need to upgrade Knowledge Base API endpoint to 4.0

### Existing Features

Qualys Knowledge Base API endpoint is currently using version 3.0. 

### What did you see?

Here is the Qualys announcement of the Qualys API endpoints: https://notifications.qualys.com/api/2025/08/17/updates-on-api-versioning-standards-deprecation-timelines

<img width="1022" height="318" alt="Image" src="https://github.com/user-attachments/assets/d6acb4fc-d1b9-42d3-8d18-8a10adbdd326" />

### Anything else?

Similar previous request: https://github.com/elastic/integrations/issues/13178

clement-fouque:
Definition 
```bash
QUALYS_USERNAME=XXX
QUALYS_PASSWORD="XXX"

curl -u '$QUALYS_USERNAME:$QUALYS_PASSWORD' -H "X-Requested-With: curl"  "https://qualysapi.qg2.apps.qualys.com/api/3.0/fo/asset/host/vm/detection/dtd/output.dtd" -o kb_3_0_output.dtd

curl -u '$QUALYS_USERNAME:$QUALYS_PASSWORD' -H "X-Requested-With: curl"  "https://qualysapi.qg2.apps.qualys.com/api/4.0/fo/asset/host/vm/detection/dtd/output.dtd" -o kb_4_0_output.dtd
```

```diff
--- kb_3_0_output.dtd	2025-12-19 14:20:23
+++ kb_4_0_output.dtd	2025-12-19 14:08:40
@@ -70,7 +70,8 @@
 <!ELEMENT DETECTION_LIST (DETECTION+)>
 <!ELEMENT DETECTION (UNIQUE_VULN_ID, QID, TYPE, SEVERITY?, PORT?, PROTOCOL?, FQDN?, SSL?, INSTANCE?, RESULT_INSTANCE?,
         RESULTS?, STATUS?,
[...]

PR #16903, opened 2026-01-08
[Qualys VMDR] Fix duplicate fields in transform

moxarth-rathod:
## Proposed commit message

```
qualys_vmdr: remove duplicate fields in transform as per the comment [1] in this PR [2]. 

Also, remove unsupported field `source` from the fields definition. The pipeline currently does not 
parse this field. This was incorrectly added into the transform fields inside the PR [2].

- [1] https://github.com/elastic/integrations/pull/16727#issuecomment-3719541508
- [2] https://github.com/elastic/integrations/pull/16727
```

## Checklist

## Related issues

- Relates #16640

### DIFF
--- a/packages/qualys_vmdr/elasticsearch/transform/latest_cdr_vulnerabilities/fields/fields.yml
+++ b/packages/qualys_vmdr/elasticsearch/transform/latest_cdr_vulnerabilities/fields/fields.yml
@@ -139,8 +139,6 @@
           type: keyword
         - name: severity
           type: long
-        - name: source
-          type: keyword
         - name: ssl
           type: keyword
         - name: status
@@ -152,8 +150,6 @@
         - name: trurisk_elimination_status
           type: keyword
         - name: type
-          type: keyword
-        - name: trurisk_elimination_status
           type: keyword
         - name: unique_vuln_id
           type: keyword


packages/qualys_vmdr/elasticsearch/transform/latest_cdr_vulnerabilities/fields/fields.yml
```diff
@@ -139,8 +139,6 @@
           type: keyword
         - name: severity
           type: long
-        - name: source
-          type: keyword
```
mrodm, line 143: Just in case, this field was not reported as duplicated in https://buildkite.com/elastic/integrations/builds/36182#019b9848-4c98-49a8-8af1-fac140f6e491/L479

moxarth-rathod: This field is currently not supported. To avoid any potential confusion or unexpected behaviour, it would be good to remove it from the field definition.

kcreddy: Can you elaborate on why we are removing this field? It seems to be present even in the DTD of new API: https://github.com/elastic/integrations/issues/16640#issuecomment-3675076048
Any field removal even if unsupported by the API should be a breaking-change.

moxarth-rathod: The pipeline currently does not parse this field. I initially planned to add it to the pipeline, but later removed it from the changes. However, I forgot to remove the field from the transform, so I now want to remove it for the time being.

kcreddy: @moxarth-rathod, can you please add this context to the commit message?

moxarth-rathod: @kcreddy done, updated.

kcreddy: @moxarth-rathod, suggested commit:

```
qualys_vmdr: remove duplicate fields in transform as per the comment [1] in this PR [2]. 

Also, remove unsupported field `source` from the fields definition. The pipeline currently does not 
parse this field. This was incorrectly added into the transform fields inside the PR [2].

- [1] https://github.com/elastic/integrations/pull/16727#issuecomment-3719541508
- [2] https://github.com/elastic/integrations/pull/16727
```

moxarth-rathod: Done

PR #14754, opened 2025-07-30
[azure_functions] Rename `properties` field to `properties.raw` when it contains invalid JSON

devamanv:
## Proposed commit message

If the azure.properties field in a log event is a string (instead of a structured object), the pipeline renames it to azure.properties.raw. This prevents Elasticsearch from rejecting the document due to a type conflict, allowing the event to be successfully indexed while preserving the original data for further processing or inspection.

Something similar was done for Azure Platform Logs, PR [here](https://github.com/elastic/integrations/pull/11732)

Note: We do not want to decide how to manage the invalid JSON that comes from Azure services.
With this approach, users have an option to decide how they want to process data in the `.raw` field using a custom pipeline.

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/obs-integration-team/issues/417
- Relates https://github.com/elastic/obs-integration-team/issues/415
- Relates https://github.com/elastic/integrations/issues/11729

PR #16904, opened 2026-01-08
[Context Security Apps] add  performance optimization enhancements for transform

Omolola-Akinleye:
We experienced an [SDH where Elasticsearch became overloaded following an agent upgrade](https://github.com/elastic/sdh-security-team/issues/1476#issuecomment-3517363704). The CSPM misconfiguration transform does not enforce a maximum document age, which contributed to data overload.

To reduce the impact of resource over-consumption, we need to optimize the Cloud Security Posture Management (CSPM) misconfiguration transform. Proposed improvements include adding query filters and maximum constraints to limit the data processed by the transform, thereby improving transform efficiency and overall resource utilization.

**Tickets Addressed:**
- [[CSPM] Add date filter to the transform to fetch only day fresher than 26 hours #14757](https://github.com/elastic/security-team/issues/14757)

- [[CSPM] Skip Cold and Frozen data in Transform #14754](https://github.com/elastic/security-team/issues/14754)

**Changes:**
** Add 26-hour time window** 
Only query findings from the last 26 hours, matching the retention policy
```yaml
must:
  range:
    "@timestamp":
      gte: "now-26h"
```

** Add Data Tier Exclusion Filter** Prevents the transform from querying cold/frozen storage tiers, which significantly improves query performance. Since the transform already uses a 26-hour retention window, there's no need to access archived data on slow storage tiers.
```yaml
must_not:
  terms:
    _tier:
      - data_frozen
      - data_cold
```
 **Bump Transform Version** Triggers Fleet to delete, reinstall, and restart the transform during package upgrade, ensuring all users receive the performance improvements.
```
fleet_transform_version: 0.3.0
```


packages/cloud_security_posture/elasticsearch/transform/misconfiguration/transform.yml
```diff
@@ -28,8 +36,10 @@ retention_policy:
     max_age: 26h
 settings:
   unattended: true
+  docs_per_second: 100
+  max_page_search_size: 100
```
opauloh: We need to be careful setting `docs_per_second` and `max_page_search_size` to a low number, this might lead to a point where the transform will start to lag behind and not being able to process everything between the transform executions, and this can reduce performance instead of improving it.

There was a great discussion in this Entity Store Performance PR about those two settings [here](https://github.com/elastic/kibana/pull/219825).

My suggestion is that we copy Entity Store strategy and have only with `max_page_search_size`, setting it to `500`. The Entity Store handles a heavy load of data and these settings are already being battle tested.

cc @romulets please correct me if I'm wrong here, but it seems that the higher performance improvement on https://github.com/elastic/sdh-security-team/issues/1476 came from removing the frozen and cold tiers and adding the time range that was added later on the investigation and not from setting the docs per second and max page to 100.

packages/cloud_security_posture/elasticsearch/transform/misconfiguration/transform.yml
```diff
@@ -3,9 +3,17 @@ source:
     - "logs-cloud_security_posture.findings-*"
   query:
     bool:
+      must:
+        range:
+          "@timestamp":
+            gte: "now-26h"
```
opauloh: There's a lint error on the package, from examples from [other integrations](https://github.com/elastic/integrations/blob/main/packages/beaconing/elasticsearch/transform/pivot_transform/transform.yml), it seems the syntax would be: 

```suggestion
      filter:
        range:
          '@timestamp':
            gte: "now-26h"
```

packages/cloud_security_posture/elasticsearch/transform/misconfiguration/transform.yml
```diff
@@ -3,9 +3,17 @@ source:
...
       must_not:
         exists:
           field: error.message
+        terms:
+            _tier:
+              - data_frozen
+              - data_cold
```
opauloh: it seems this is the correct syntax (took an example from [here](https://github.com/elastic/integrations/blob/3126431d7064365c6da5e7a2be3c53ddcdb11147/packages/beaconing/elasticsearch/transform/pivot_transform/transform.yml#L369)): 

```suggestion
        terms:
            _tier: ["data_frozen", "data_cold"]
```

opauloh (changes requested):
There are some adjustments needed, also can we test the new settings before we merge it?

You can do it by building the package locally, spinning up a new environment in cloud and upload the integration from the zip file. Then install a new agent  and verify if the findings are being loaded properly. Ideally, with these settings we should be seeing findings being generated incrementally in a few seconds

packages/cloud_security_posture/elasticsearch/transform/misconfiguration/transform.yml
Omolola-Akinleye: Thanks Paulo for the thorough review. Okay I can remove `docs_per_second` and   `max_page_search_size` to test query filters. I think  `max_page_search_size` default to   500.   We have the [separate ticket]([CSPM] Consider having docs per second and page size by default in transform
[#14755](https://github.com/elastic/security-team/issues/14755)) that can be utilized investigation and exploration of  settings `docs_per_second` and   `max_page_search_size`

packages/cloud_security_posture/changelog.yml
```diff
@@ -16,6 +16,11 @@
 # 1.4.x - 8.9.x
 # 1.3.x - 8.8.x
 # 1.2.x - 8.7.x
+- version: "3.3.0-preview01"
```
maxcold, line 21: why 3.2.0 is skipped?

packages/cloud_security_posture/elasticsearch/transform/misconfiguration/transform.yml
```diff
@@ -32,4 +38,4 @@ _meta:
   managed: true
   # Bump this version to delete, reinstall, and restart the transform during package.
   # Version bump is needed if there is any code change in transform.
-  fleet_transform_version: 0.2.1
+  fleet_transform_version: 0.2.2
```
maxcold: If we bump the minor version of the package, it would be nice to bump the minor here as well. Don't think it influences anything from a functionality standpoint, but it's good to keep consistency

maxcold (commented):
couple of questions regarding the version, otherwise from the logic perspective the change looks fine, but I agree with @opauloh that we need to verify it. Ideally on the env with data to make sure the expected number of findings is being present in the findings page after a cycle

Other points not directly related to this PR but to the task in general:
- we also need to update the vulnerability transform, which is still being installed from the Kibana code. With vulnerabilities it's even more important as the volume of vulnerabiltiies is usually bigger then the volume of misconfigurations
- we have the old transform definition in Kibana code, there is logic to install it for older versions of the package. Can you check if we need to backport this change to old transform? probably not, but I don't remember exactly starting from which version customers get the transform from the integration, the change was quite recent. Ido did it for the namespace epic

packages/cloud_security_posture/changelog.yml
Omolola-Akinleye: So I was confusing on versioning comment. I thought each package  version should match the release version and I didn't see the `3.2.0` but maybe we bump to `3.2.0` for `9.3.0` and `9.4.0` releases. Wdyt? @maxcold 
image

Omolola-Akinleye:
@maxcold  and @opauloh  Yes the original syntax was correct. I realized that I'm getting validation errors, but I didn't install the latest  elastic package version.

Omolola-Akinleye:
@opauloh @maxcold  Something off with query filter range timestamp  but I don't think it's the array structure. There issue with  query filter syntax
```
  must:
        - range:
            '@timestamp':
            gte: "now-26h"
```

>we also need to update the vulnerability transform, which is still being installed from the Kibana code. With vulnerabilities it's even more important as the volume of vulnerabiltiies is usually bigger then the volume of misconfigurations

Okay I can create a separate ticket and pr for vulnerabilities transform

>we have the old transform definition in Kibana code, there is logic to install it for older versions of the package. Can you check if we need to backport this change to old transform? probably not, but I don't remember exactly starting from which version customers get the transform from the integration, the change was quite recent. Ido did it for the namespace epic

So you are saying we should apply these to old transform? I believe [old transform](https://github.com/elastic/kibana/blob/b83feba219f73a9a2a3d43b4b6557d4f63d9948b/x-pack/solutions/security/plugins/cloud_security_posture/server/create_transforms/latest_findings_transform.ts#L15) was last updated `8.15.0`

packages/cloud_security_posture/changelog.yml
maxcold: Historically, during the release cycle, we add changes to the integration package in `-previewX` version and then between the first and second QA cycle, we release the version of the package under preview. During 9.3.0 work, there were only two changes, neither of them were not released under preview; there was no 3.2.0 release for 9.3.0. But that doesn't mean we need to skip this version. I will post on our QA channel for clarification if the package release process changed. If not, and we just skipped it for 9.1.1 and 9.1.2, you need to add your change under `9.2.0-preview01`. If, from now on, we decide not to release under preview, then you can add it under just `9.2.0`.

maxcold:
> So you are saying we should apply these to old transform?

@Omolola-Akinleye no, I'm just saying that you need to check and decide if the filter needs to be applied. If you think it's not needed, then no change to the old transform is needed

packages/cloud_security_posture/changelog.yml
```diff
@@ -1,6 +1,7 @@
 # newer versions go on top
 # version map:
 # IMPORTANT: this map doesn't apply to serverless where package availability depends on the spec version https://github.com/elastic/kibana/blob/main/config/serverless.yml#L14-L15
+# 3.2.x - 9.3.x, 9.4.x
```
maxcold: as there were no change to kibana condition https://github.com/elastic/integrations/blob/main/packages/cloud_security_posture/manifest.yml#L26 , 3.2.x will work for 9.2.x as well. Let's figure out the approch to package versioning first https://elastic.slack.com/archives/C03E5KGNWT1/p1768212825336389

packages/cloud_security_posture/elasticsearch/transform/misconfiguration/transform.yml
romulets: I agree with your assessment @opauloh, we can hold the docs per sec and max page until we have test it further.

It's important to say tho that in previous SDHs such a setting did yield big impact. So I'd not drop understanding what default config is good enough for us. I do think some lag in our use case might be acceptable, the cycle is once per day.

romulets (approved):
The changes to the transform itself look good to me!

packages/cloud_security_posture/changelog.yml
Omolola-Akinleye: Ahh okay thank you for providing clarity!

packages/cloud_security_posture/changelog.yml
Omolola-Akinleye: I can now bump to from `3.2.0` to `3.3.0-preview01` according to [docs](https://docs.elastic.dev/security-solution/cloud-security/release/promote-integration) so I just followed that approach since these changes are for `9.4.0` release

Omolola-Akinleye:
Transform continues to generate findings with the latest queries @opauloh 

image
image
image

packages/cloud_security_posture/elasticsearch/transform/misconfiguration/transform.yml
opauloh: Thanks for the clarification @romulets! I think it will be great to work on it in the separate ticket then, so we can isolate the performance gains benefits and battle test a good default for misconfigurations and vulnerabilities

opauloh (approved):
LGTM - thanks for performing the tests with real data!

packages/cloud_asset_inventory/manifest.yml
```diff
@@ -14,7 +14,7 @@ categories:
   - cloud
 conditions:
   kibana:
-    version: ">=9.3.0"
+    version: ">=9.4.0"
```
maxcold, line 17: @romulets I noticed that `cloud_asset_inventory` is the only integration using `>=` notation. It is supported and works, but was this intentional to use this instead `^9.4.0` which is used in all the integrations? One slight difference is that the versions of cloud_asset_inventory will be supported by 10.x of the stack if we ever get it. Which we probably don't want

maxcold (approved):
changes LGMT, the versioning now is correct

packages/cloud_asset_inventory/manifest.yml
romulets: Let's align with the rest of integrations. @Omolola-Akinleye is this something you could do in this PR since it's small pls?

PR #16915, opened 2026-01-09
[gitlab] Remove the forwarded tag from filestream input by default

### LINKED ISSUE #13564: [gitlab]: Do not include the "forwarded" tag in the defaults
The GitLab integration (version 2.2.0) includes the "forwarded" tag for all input streams by default. This disables the `add_host_metadata` processor which causes the `host.*` fields not to be filled. This is inconsistent with other defaults of the integration, especially the log file paths, since those indicate processing logs of a local instance (as they point to the default log file locations). So I think the "forwarded" tag should not be added by default for this integration. 

(BTW this might be a more widespread issue, as running a simple `grep` shows that this tag is added at a lot of places. But I didn't go very deep with this investigation.)

brijesh-elastic:
## Proposed commit message

```
gitlab: remove the `forwarded` tag from filestream input by default
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/gitlab directory.
> elastic-package test

## Related issues

- Closes #13564

### DIFF
--- a/packages/gitlab/changelog.yml
+++ b/packages/gitlab/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Remove the `forwarded` tag from filestream input by default.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/1
+      link: https://github.com/elastic/integrations/pull/16915
 - version: "2.3.3"
   changes:
     - description: Treat all variables fields as keyword.


ShourieG:
@brijesh-elastic, before we go ahead with this, we need to update the issue by saying, having forwarded in defaults is actually the consistent behaviour.

PR #16947, opened 2026-01-13
mimecast: fix errors with invalid timestamps and expired URLs

chemamartinez:
## Proposed commit message

```
mimecast: adding fixes for the following two errors.

- Some indicators for the threat_intel_malware_grid data stream contains timestamps
  in the format yyyy-MM-dd'T'HH:mmZ. That caused the following error when the
  timestamp() function is called over one of these fields.

  failed eval: ERROR: <input>:42:40: type conversion error from 'string' to 'google.protobuf.Timestamp'

  The solution given to avoid this error is to fill the field with the seconds part so
  the field format gets valid for calling timestamp().

- For the siem_logs data stream, when getting a "403 Forbidden: Request has expired"
  response from the API due to the expiration of the blob that is being processed.
  The code handled this as a generic error, waiting for the next interval but retrying with the
  same expired blob.
  
  Now, the error is handled as the 404 so expired blob is removed from the queue and keep
  processing the remaining ones.
```

## Checklist

### DIFF
--- a/packages/mimecast/changelog.yml
+++ b/packages/mimecast/changelog.yml
@@ -3,10 +3,10 @@
   changes:
     - description: Fix handling of timestamp values with missing seconds.
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/99999
+      link: https://github.com/elastic/integrations/pull/16947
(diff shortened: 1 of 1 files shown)


ShourieG (approved):
LTGM

Issue #16679, opened 2025-12-23
[Prometheus Remote Write] Configurable size limits on incoming requests for remote_write metricset (max_compressed_body_bytes, max_decoded_body_bytes)


gizas:
The prometheus remote-write integration should allow configurable size limits on incoming requests for the fields: 
 
- max_compressed_body_bytes
- max_decoded_body_bytes

Above fields have been introduced after https://github.com/elastic/beats/pull/48218

NOTE: The specific issue should be implemented once  https://github.com/elastic/beats/pull/48218 beats version have been released.

PR #16962, opened 2026-01-14
[Prometheus Remote Write] Update request size limits

gizas:
**Dont merge until ^9.2.4 || ^9.3.1 || ^9.4.0 are released**

- Enhancement

## Proposed commit message

- WHAT: Adds request size limits configuration variables in the elastic agent policy for the remote_write datastream.
- WHY:  In order to provide configuration options for max_compressed_body_bytes and max_decoded_body_bytes, after  https://github.com/elastic/beats/pull/48218

## Checklist

1. Clone this pr
1. Navigate to `integrations/packages/prometheus` and run `elastic-package build`
2. elastic-package stack up -d -v --version=9.2.4 (or later)

## Related issues

- Closes #https://github.com/elastic/integrations/issues/16679
 
## Screenshots

prometheus


packages/prometheus/data_stream/remote_write/agent/stream/stream.yml.hbs
```diff
@@ -14,6 +14,8 @@ period: {{period}}
 metrics_count: {{metrics_count}}
 rate_counters: {{rate_counters}}
 use_types: {{use_types}}
+max_compressed_body_bytes: {{max_compressed_body_bytes}}
```
gizas, line 17: See beats relevant config options: https://github.com/elastic/beats/blob/main/metricbeat/module/prometheus/remote_write/config.go#L35C58-L35C80

packages/prometheus/manifest.yml
```diff
@@ -10,7 +10,7 @@ categories:
   - containers
 conditions:
   kibana:
-    version: "^8.16.0 || ^9.0.0"
+    version: "^9.2.4 || ^9.3.1 || ^9.4.0"
```
gizas: According to https://github.com/elastic/beats/pull/48218/files#diff-3bca358ec7bb955db5fa88789cf4e1155ee015b30b28abd2b1892725b52190afR63

gizas:
@elastic/obs-infraobs-integrations can I have a review here please?

packages/prometheus/data_stream/remote_write/manifest.yml
```diff
@@ -75,6 +75,22 @@ streams:
         required: true
         show_user: true
         default: true
+      - name: max_compressed_body_bytes
+        type: text
```
MichaelKatsoulis: Should the mapping be text? This is a int64 field

gizas: Changed it to integer. 

FYI when trying to test:

```bash
   2. file "/Users/andreasgkizas/elastic/integrations/build/packages/prometheus-1.25.0.zip/data_stream/remote_write/manifest.yml" is invalid: field streams.0.vars.10.type: streams.0.vars.10.type must be one of the following: "bool", "email", "integer", "password", "select", "text", "textarea", "time_zone", "url", "yaml"

```
@jsoriano do you think we can have problems here with the size of the field if I add them as integers? What are the min and max values? I dont see a problem with validation with the default values of max_compressed_body_bytes and max_decoded_body_bytes

jsoriano: > do you think we can have problems here with the size of the field if I add them as integers? What are the min and max values?

Min and max values are the ones provided by Elasticsearch, `long` allows to store bigger values if needed https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/number

If you need even bigger numbers and you don't mind about losing precision, you can also use `double`.

jsoriano: Btw, as we are in the context of Prometheus, it will probably make sense to use `double`, as is the data type used there.

gizas: With double is that I get the validation error:

```bash
Error: building package failed: invalid content found in built zip package: found 2 validation errors:
   1. file "/Users/andreasgkizas/elastic/integrations/build/packages/prometheus-1.25.0.zip/data_stream/remote_write/manifest.yml" is invalid: field streams.0.vars.9.type: streams.0.vars.9.type must be one of the following: "bool", "email", "integer", "password", "select", "text", "textarea", "time_zone", "url", "yaml"

   2. file "/Users/andreasgkizas/elastic/integrations/build/packages/prometheus-1.25.0.zip/data_stream/remote_write/manifest.yml" is invalid: field streams.0.vars.10.type: streams.0.vars.10.type must be one of the following: "bool", "email", "integer", "password", "select", "text", "textarea", "time_zone", "url", "yaml"

```

gizas: I think integer is ok

jsoriano: Sorry, I was thinking this was a field definition :facepalm: Integer should be fine for this variable, yes.

gizas:
@elastic/obs-infraobs-integrations  kind reminder please

packages/prometheus/manifest.yml
ishleenk17: I don't think changing kibana version to this is a good idea. 
Any future bugfixes will be skipped for 18.9, 9.0, 9.1, 9.3.0.
Can we instead the agent version requirment for this configuration int he documentation of the conig parameters ?

jsoriano: `^9.2.4` already includes `^9.3.1` and `^9.4.0`, were you intending to set something like `~9.2.4 || ~9.3.1 || ^9.4.0`?

gizas: Let me clarify with an example: If someone tries 9.2.3 should not match.
And also if someone tries 9.3.0 should also not match. So I would like to include 9.2.4 but not 9.3.0

Is not this the correct approach?

jsoriano: `^` matches all the versions in the same major over the indicated version. `^9.2.4` is equal to ">= 9.2.4 AND < 10.0.0", so `^9.2.4 || ^9.3.1 || ^9.4.0` includes 9.3.0.

You need to use `~` to match only versions of the same minor, so `~9.2.4 || ~9.3.1 || ^9.4.0`.

gizas: Thank you ! done in [fix versions](https://github.com/elastic/integrations/pull/16962/commits/87d9441c99e81cd56bcc40f18a29595f9db790e8)

ishleenk17: @gizas : I still think this should not be done. As doing this we are actually removing any future support to versions like 8.19, 9.0, 9.1, 9.2 (lesser than 9.2.4), 9.3.0.
Can we instead update about this agent version dependency in the readme.

tommyers-elastic: what's the reason we are removing support for <9.2.4?

tommyers-elastic: for core integrations like prometheus we _really_ want to retain support for 8.x customers.

gizas: The reason for that was that the beats fix https://github.com/elastic/beats/pull/48218 was not imported to 8.19 versions. 

See above PR for relevant backports.
This relates to previous comment from Ishleen. 

If you agree I can do:

```
^8.16.0 || ^9.2.4 || ^9.3.1 || ^9.4.0
```

and just document the limitations in README?

jsoriano: `^8.16.0 || ^9.2.4 || ^9.3.1 || ^9.4.0` is equivalent to `^8.16.0 || ^9.2.4`, so the package will be available in 9.3.0.

You probably want `^8.16.0 || ~9.2.4 || ^9.3.1`.

gizas: Thank you @jsoriano ! @tommyers-elastic , @ishleenk17 do u agree to proceed with above versions and also document the fact that the existing variables will not work in 8.x ?

gizas: @tommyers-elastic , @ishleenk17 I did the backport to 8.19. See https://github.com/elastic/beats/pull/48918

So now "^8.19.3 || ~9.2.5 || ^9.3.1" is the final one. 
Can I have another review please?

gizas: See doc: [update doc](https://github.com/elastic/integrations/pull/16962/commits/dd4ccac2249efa662f4c4237ff53c33f051397d8)

PR #16973, opened 2026-01-15
[Cisco_Secure_Email_Gateway] Fix pattern in amp logs

bhapas:
## Proposed commit message

Fix parsing numeric values containing `,` in amp logs.

Currently if there is a `,` in numeric values in amp logs , then the parsing fails. This PR fixes it.

## Checklist


packages/cisco_secure_email_gateway/data_stream/log/elasticsearch/ingest_pipeline/pipeline_amp.yml
```diff
@@ -25,13 +25,13 @@ processors:
...
       if: ctx._tmp.new_message != null && ctx._tmp.grok_match_index != "1"
-      field_split: ",\\s*(?=[^',]*[=:][^',]*)"
+      field_split: ",\\s*(?=[A-Za-z_][^',]*[=:][^',]*)"
       value_split: '\s*=\s*|:\s*'
   - kv:
       tag: kv__tmp_new_message_grok_1_0bb63133
       field: _tmp.new_message
       if: ctx._tmp.new_message != null && ctx._tmp.grok_match_index == "1"
-      field_split: ",\\s*(?=[^',]*[=][^',]*)"
+      field_split: ",\\s*(?=[A-Za-z_][^',]*[=][^',]*)"
```
qcorporation, line 34: @bhapas could there be situations where other characters follow the comma which will not recognize the delimiter? Should we handle these?
Example: 
`File Size = 2634 bytes, File Type =` works
`File Size = 2634 bytes, -File Type =` doesn't work
`File Size = 2634 bytes, /File Type =` doesn't work

bhapas: Not sure if the messages have fields that start with a special character? The values can contain though and it should work.

qcorporation: @bhapas I'm happy to give a thumbs up, I'm just worried that we might run into an "Identifier" field that has a number at the beginning of it. Maybe that's very unlikely otherwise we'd see more error reports.
e.g.
`File Size = 2634 bytes, 0-Tag =` doesn't work

Just wanted to highlight a small edge case.

qcorporation: I'll leave it up to you to make the decision to fix the edge case or not.

bhapas: Looking at examples and other logs there is no evidence of such messages. So I leave this as is for now

PR #16986, opened 2026-01-16
Update CODEOWNERS to include kibana-management

susan-shu-c:
Added @elastic/kibana-management as a code owner for multiple packages.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -200,7 +200,7 @@
 /packages/cylance @elastic/security-service-integrations
 /packages/darktrace @elastic/security-service-integrations
 /packages/ded @elastic/sec-applied-ml @elastic/kibana-management
-/packages/dga @elastic/sec-applied-ml @elastic/kibana-management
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -199,8 +199,8 @@
 /packages/cyera @elastic/security-service-integrations
 /packages/cylance @elastic/security-service-integrations
 /packages/darktrace @elastic/security-service-integrations
-/packages/ded @elastic/sec-applied-ml
-/packages/dga @elastic/sec-applied-ml
+/packages/ded @elastic/sec-applied-ml @elastic/kibana-management
+/packages/dga @elastic/sec-applied-ml @elastic/kibana-management
```
sodhikirti07: We probably don't need them in packages without transforms that are hta, dga, problemchild?

susan-shu-c: Ah good point, anyone else we could add? I can remove for now

Issue #16333, opened 2025-12-05
[AWS] lastSync option of start_position parameter not available in stack version before 8.19.0


MichaelKatsoulis:
Coming from [SDH](https://github.com/elastic/sdh-beats/issues/6706). 
Details for the problem and solution can be found in https://github.com/elastic/sdh-beats/issues/6706#issuecomment-3616286282 and https://github.com/elastic/sdh-beats/issues/6706#issuecomment-3616705678

Useful link: 
https://www.elastic.co/docs/extend/integrations/developer-workflow-support-old-package

Kavindu-Dodan:
@MichaelKatsoulis unfortunately, backport is too complex and time consuming for this matter.

First, there are build CI issues that are preventing us from opening backport fixes for AWS `3.13.x` track (see my attempt at https://github.com/elastic/integrations/pull/16975). 

This is due to CI dependency check issues, which were fixed by https://github.com/elastic/integrations/pull/15353

This means we have to go through the following,

- Crete back port branch
- Backport CI fix at https://github.com/elastic/integrations/pull/15353 to the backport branch
- Patch the track (`3.13.x`)
- Update main

And per changelog, the Kibana version got updated with version `4.0.x`, so we have to repeat the above for `3.13.x`, `3.14.x`, `3.15.x`, `3.16.x` & `3.17.x`

My view is this is too much effort. Let me know your thoughts.

MichaelKatsoulis:
My understanding is that we need justo create version 3.17.1 just after this commit https://github.com/elastic/integrations/commit/5b593f66817afe5d576f9d5c67a3231254bc6ad0#diff-6ea379ea4e907be5cc25a65c404d5bb98841ab15439b2c43836118150c90c9af  and remove the last sync option.
That way users of 8.18 and below won't see it

Kavindu-Dodan:
@MichaelKatsoulis thank you very much for giving this perspective.

I have opened https://github.com/elastic/integrations/pull/16998 to fix this for AWS integration.

Kavindu-Dodan:
We have fix for AWS integration with version 3.17.1 (refer - https://github.com/elastic/integrations/pull/16998#issuecomment-3791022900)

PR #14733, opened 2025-07-29
[AWS] [CloudWatch] Improve description to highlight lastSync

Kavindu-Dodan:
## Proposed commit message

Improve the description of `start_position` field with `lastSync` configuration option.

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [ ] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Related issues

Related https://github.com/elastic/beats/pull/43251

PR #16998, opened 2026-01-19
chore: remove CloudWatch lastSync option from older tracks as it is incompatible with Kibana versions below 9.x

Kavindu-Dodan:
## Proposed commit message

Backport for AWS integration 3.17.x track to disable `lastSync` start_position configuration option for CloudWatch

This is incompatible with Kibana versions lower than 9.x

## Related issues

Addressed AWS integration concern of https://github.com/elastic/integrations/issues/16333 
Original PR - https://github.com/elastic/integrations/pull/14733

### DIFF
--- a/packages/aws/changelog.yml
+++ b/packages/aws/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.17.1"
+  changes:
+    - description: Disable lastSync start_position configuration for CloudWatch as it's incompatible with versions below 9.x.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/16998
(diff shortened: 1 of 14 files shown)


kcreddy (approved):

agithomas:
@Kavindu-Dodan , could you verify if similar backport changes are to be extended for the following packages as well?

- aws_bedrock
- aws_mq
- aws_logs

Kavindu-Dodan:
> @Kavindu-Dodan , could you verify if similar backport changes are to be extended for the following packages as well?

Yes these packages are also affected. But I did not focus on them given the effort we need on them. I will work on them if there are concerns raised on them :)

Issue #10333, opened 2024-07-02
Fix kv parsing in auditd integration to properly support quoted values.


Tacklebox:
The auditd integration uses regex patterns with the kv ingest pipeline processor to split key=value substrings out of the logs from auditd. Since the values can be arbitrary quoted data including whitespace, to fully support parsing the kv pairs, parsing should be done from the painless script instead to account for multiple spaces, escaped quotes/etc.

zez3:
error.message

Processor "kv" with tag "" in pipeline "logs-auditd.log-3.20.1" failed with message "field [auditd.log.kv] does not contain value_split [(?<!\\)=]"

zez3:
@andrewkroh 

are you part of Team:Security-Linux Platform?

kurtiam:
Hi @Tacklebox!   Checking in to see if there's a status update on this issue?

ono-sendaiOG:
👍

PR #17011, opened 2026-01-20
Update kv regex parser

nicholasberlin:
## Proposed commit message

The first regex is not “quote-aware”; it’s trying to avoid splitting based on a simplistic token model (`[\w"]+`). As soon as a word is followed by punctuation (like `profile,`) inside quotes, the lookahead no longer recognizes it as a protected token, and the delimiter splits in the middle of the quoted value.

The new approach is to key off the true structural marker (`=`), not the incidental character makeup of the next token.

## Checklist

1. clone the repo
2. `$ cd packages/auditd`
3. `$ elastic-package stack up -d --services=elasticsearch`
4. `$ elastic-package test pipeline`
5. verify tests pass


stanek-michal:
link to https://github.com/elastic/integrations/issues/10333

stanek-michal:
This doesn't solve the non-printable characters/control characters case right? Like line 3 in packages/auditd/data_stream/log/_dev/test/pipeline/test-auditd-enriched.log

I think it also doesn't solve this case:
https://github.com/elastic/sdh-beats/issues/6811

PR #16907, opened 2026-01-09
[Cloudflare Logpush] Fix uri parsing for http_request data stream

moxarth-rathod:
## Proposed commit message

```
cloudflare_logpush: fix uri parsing for http_request data stream
```

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally

- Clone integrations repo.
- Install elastic package locally.
- Start elastic stack using elastic-package.
- Move to integrations/packages/cloudflare_logpush directory.
- Run the following command to run tests.
> elastic-package test

## Related issues

- Closes #16753

Issue #16753, opened 2026-01-02
[Cloudflare Logpush]: Ingest Pipeline bug when URI starts with "//"


avpxyyf:
### Integration Name

Cloudflare Logpush [cloudflare_logpush]

### Dataset Name

http_request

### Integration Version

1.43.0

### Agent Version

9.2.2

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.2.2

### OS Version and Architecture

Ubuntu Server 24.04 LTS

### Event Original

{
    "BotDetectionIDs": [],
    "BotDetectionTags": [],
    "BotScore": 80,
    "BotScoreSrc": "Machine Learning",
    "BotTags": [],
    "ClientASN": 1241,
    "ClientCity": "Athens",
    "ClientCountry": "gr",
    "ClientDeviceType": "mobile",
    "ClientIP": "[redacted]",
    "ClientIPClass": "noRecord",
    "ClientLatitude": "[redacted]",
    "ClientLongitude": "[redacted]",
    "ClientMTLSAuthCertFingerprint": "",
    "ClientMTLSAuthStatus": "unknown",
    "ClientRegionCode": "I",
    "ClientRequestBytes": 2716,
    "ClientRequestHost": "[redacted].[redacted].gr",
    "ClientRequestMethod": "GET",
    "ClientRequestPath": "//android-chrome-192x192.png",
    "ClientRequestProtocol": "HTTP/1.1",
    "ClientRequestReferer": "",
    "ClientRequestScheme": "https",
    "ClientRequestSource": "eyeball",
    "ClientRequestURI": "//android-chrome-192x192.png",
    "ClientRequestUserAgent": "Mozilla/5.0 (Linux; Android 15; 23127PN0CG Build/AQ3A.240627.003) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.7499.34 Mobile Safari/537.36 OPX/3.0",
    "ClientSSLCipher": "AEAD-AES128-GCM-SHA256",
    "ClientSSLProtocol": "TLSv1.3",
    "ClientSrcPort": 48176,
    "ClientXRequestedWith": "",
    "ContentScanObjResults": [],
    "ContentScanObjSizes": [],
    "ContentScanObjTypes": [],
    "Cookies": {},
    "EdgeEndTimestamp": "2025-12-19T09:59:23Z",
    "EdgePathingOp": "wl",
    "EdgePathingSrc": "undef",
    "EdgePathingStatus": "nr",
[...]

PR #16999, opened 2026-01-20
cloudflare_logpush,netskope,symantec_endpoint_security: fix azurite version skew

efd6:
## Proposed commit message
```
cloudflare_logpush,netskope,symantec_endpoint_security: fix azurite version skew

Since 2026-01-05, the version of the Azure Storage service is
2026-02-06[1]. This version is not compatible with the CLI since Azurite
only recognises version up to 2025-11-05[2], the previous version. Until
Azurite is updated[3], this prevents blocking the unrecognised version.

[1]https://learn.microsoft.com/en-us/rest/api/storageservices/versioning-for-the-azure-storage-services
[2]https://github.com/Azure/Azurite
[3]https://github.com/Azure/Azurite/issues/2623
```

> [!NOTE]
> No changelog entry since this is not user-facing.

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [ ] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## Related issues

- Closes #16964

PR #17014, opened 2026-01-21
[Cloudflare Logpush] Fix URI parsing by ignoring invalid URI string for http_request data stream

moxarth-rathod:
## Proposed commit message

```
cloudflare_logpush: fix URI parsing for URIs starting with '/'

The previous implementation failed to handle URIs that start with '/'
(absolute path reference: '/' or protocol-relative URIs: '//'). This
constructs complete URIs by combining scheme, domain, and path
components in the http_request data stream.
```

## Checklist

- Clone integrations repo.
> elastic-package test

## Related issues

- Relates #16753
- Resolves the comment: https://github.com/elastic/integrations/pull/16907#discussion_r2706230216

### DIFF
--- a/packages/cloudflare_logpush/changelog.yml
+++ b/packages/cloudflare_logpush/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "1.43.2"
   changes:
-    - description: Fix URI parsing for invalid URIs starting with '//' by constructing complete URIs from scheme, domain, and path components in http_request data stream.
+    - description: Fix URI parsing for URIs starting with '/' by constructing complete URIs from scheme, domain, and path components in http_request data stream.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/17014
 - version: "1.43.1"


moxarth-rathod:
Build is failing due to the following PR, need to wait for that to merge:
https://github.com/elastic/integrations/pull/16999

packages/cloudflare_logpush/data_stream/http_request/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -414,6 +398,21 @@ processors:
       copy_from: json.ClientRequestQuery
       ignore_empty_value: true
       if: ctx.url?.query == null
+  - set:
+      field: tmp.constructed_uri
+      tag: set_tmp_constructed_uri
+      value: '{{url.scheme}}://{{url.domain}}{{json.ClientRequestURI}}'
+      if: ctx.json?.ClientRequestURI != null 
+          && ctx.url?.scheme != null 
+          && ctx.url?.domain != null
```
efd6, line 407:
```suggestion
      if: ctx.json?.ClientRequestURI != null && ctx.json.ClientRequestURI.startsWith('//') && ctx.url?.scheme != null && ctx.url.domain != null
```
Do we definitively know that we only get protocol-relative URIs? If not we should also handle the cases that `ClientRequestURI` is scheme-ful.

In that case we would also have something like
```
  - set:
      field: tmp.constructed_uri
      tag: set_tmp_constructed_uri_from_ClientRequestURI
      copy_from: json.ClientRequestURI
      if: ctx.json?.ClientRequestURI != null && !ctx.json.ClientRequestURI.startsWith('//')
```
immediately following the existing set.

moxarth-rathod: @efd6 Should we construct the URI when it contains only the path and query, for example
/s/example/api/telemetry/v2/clusters/_stats?

efd6: Do we see those as well?

moxarth-rathod: Yes, there are examples in the pipeline tests where the URI field contains only the path.

efd6: OK. Do we know that they are real? I guess we can be lenient and allow them, this would just require that the `startsWith` calls above only have a single slash.

packages/cloudflare_logpush/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.43.2"
+  changes:
+    - description: Fix URI parsing for invalid URIs starting with '//' by constructing complete URIs from scheme, domain, and path components in http_request data stream.
```
efd6:
```suggestion
    - description: Fix URI parsing for URIs starting with '/' by constructing complete URIs from scheme, domain, and path components in http_request data stream.
```

Issue #16567, opened 2025-12-15
[elastic_agent]: agent.name and component.id cannot be used for filtering


cmacknz:
### Integration Name

Elastic Agent [elastic_agent]

### Integration Version

2.6.10

### Agent Version

9.2.0+

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.3.0-SNAPSHOT

### OS Version and Architecture

Ubuntu 24.04

### Error Message

<img width="2546" height="796" alt="Image" src="https://github.com/user-attachments/assets/23fda258-3d0d-409d-8f84-471a91911a87" />

The default agent metrics dashboard appears to be broken, with CPU metrics showing "component.id can not be used for filtering" and the top level agent hostname display showing "agent.name can not be used for filtering"

### What did you do?

Enroll various 9.2.x versions of Elastic Agent.

### What did you see?

Broken Elastic Agent metrics overview dashboards.

### What did you expect to see?

Visualizations that worked.

belimawr:
There is a similar, older issue: https://github.com/elastic/integrations/issues/6545

AndersonQ:
I was just testing it and it works on my stack:

<img width="2546" height="1210" alt="Image" src="https://github.com/user-attachments/assets/d783e170-9194-4b89-80af-d7d8fd3c4c81" />

Just to confirm, the screenshot came from a new stack, no changes to the dashboard or indexes template/mapping, right?

AndersonQ:
Nevertheless, it's good to look at our mapping to ensure we have the important fields mapped correctly.

AndersonQ:
I've just seen a similar error on a 8.15.5 deployment. The stack it worked was a 9.4.0-SNAPSHOT. Perhaps something on ES changed and it works on 9.4

<img width="3835" height="2001" alt="Image" src="https://github.com/user-attachments/assets/c63147e6-f677-43dc-9468-2e9f7788f24f" />

khushijain21:
I tried reproducing this with 9.2.3, 9.3.0-SNAPSHOT and even 9.4.0-SNAPSHOT. But it is working.
I think this error shows up intermittently or only with certain combination of integrations are installed.

I installed `elastic-agent`  then `system-integration` followed by `elastic-defend` and could not reproduce this.
@cmacknz and @AndersonQ were any other integrations installed when you observed this error?

I will try some more options and report it here.

AndersonQ:
[...]

Issue #16982, opened 2026-01-16
[Elastic-Agent] Elastic-Agent integration fails to show cpu/memory usage if system-integration is not installed


khushijain21:
### Integration Name

Elastic Agent [elastic_agent]

### Integration Version

2.6.11 

### Agent Version

9.3.0

### Agent Output Type

elasticsearch

### What did you do?

I created elastic stack version 9.2.3 using elastic-package. I added `Elastic-Agent` integration to `docker-fleet-agent` policy (Note system integration is not enabled).

I looked up the `Agent Metrics Dashboard` and selected `agent.name: docker-fleet-agent` and you can see that cpu and memory usage is not available and the error is `system.process.cpu.total.pct` was not found.

<img width="483" height="298" alt="Image" src="https://github.com/user-attachments/assets/4376b26b-6c95-448d-8d58-c64d534c3f7f" />

and similarly for memory usage.

### What did you see?

If you see the visualization formula - it looks for both `process.cpu.total.time` as well as `system.process.cpu.total.pct`. 
Hence the visualization fails when only monitoring is enabled and `system.process.cpu.total.pct` is not present
```
pick_max(
  pick_max(
    differences(
      defaults(max(system.process.cpu.total.time.ms),0)
      )/interval(),
    0
  ),
  pick_max(
    differences(
      defaults(max(system.process.cpu.total.pct),0)),
    0
  )
)
```

### What did you expect to see?

I expected to see CPU usage even if defend is not installed

### Anything else?

And since agent monitoring metrics are collected and stored in `metrics-elastic_agent.elastic_agent.*` index instead of all `metrics-*` index - we should only querying the former index pattern. For example, if system integration is enabled and monitoring is disabled we will still visualization on `Agent Metrics Dashboard` which may or may not be true values

pierrehilbert:
cc @AndersonQ as you worked on this.

AndersonQ:
The [mapping](https://github.com/elastic/integrations/blob/f28e5ecc72bd177051b1ce86eea90a6d0eb8dee1/packages/elastic_agent/data_stream/endpoint_security_metrics/fields/fields.yml#L24-L45) is present for Elastic Defend (a.k.a `endpoint`). Now looking at it, I think what is happening is when neither Elastic Defend or the System integration is installed, no data stream on `metrics-*` has the mapping, thus the error.

[...]

PR #17015, opened 2026-01-21
[elastic_agent] Add mapping for `component.*` and `system.process.*` fields

khushijain21:
## Proposed commit message
This PR makes `component.id` consistent across metric data stream in elastic-agent integration. 
It also adds mappings for `system.process.cpu.total.pct` and `system.process.memory.rss.bytes`

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/16567

### DIFF
--- a/packages/elastic_agent/changelog.yml
+++ b/packages/elastic_agent/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "2.6.12"
   changes:
-    - description: Add mappings for `component.*` and `system.process.cpu.total.pct` fields
+    - description: Add mappings for `component.*`, `system.process.cpu.total.pct`, `system.process.memory.rss.bytes` fields
       type: bugfix
       link: "https://github.com/elastic/integrations/pull/16649"
 - version: "2.6.11"


khushijain21:
@elastic/elastic-agent-data-plane this change does not have user facing consequences. Should this still have a changelog.?

swiatekm (approved):

AndersonQ:
> @elastic/elastic-agent-data-plane this change does not have user facing consequences. Should this still have a changelog.?

It does have "consequences" because it changes the mappings.

I was thinking about adding all `component.*` mappings. I started to do it on another branch, using 
```
- name: component
  type: group
  fields:
    - name: id
      type: wildcard
      ignore_above: 1024
      description: Component id
    - name: type
      type: keyword
      ignore_above: 1024
      description: The type of the component
    - name: binary
      type: keyword
      ignore_above: 1024
      description: The binary that exeuctes the component
      example: filebeat
    - name: dataset
      type: keyword
      ignore_above: 1024
```

which is the most mappings for `component.*` I could find across all components data streams.

I think we can do a single PR, with this and fixing https://github.com/elastic/integrations/issues/16982 as well. that way we minimise the amount of versions/upgrades for the users.

khushijain21:
Currently `component.id` is defined like this
```
- name: component.id
  type: keyword
  dimension: true
```

with dimension set to true. Which I believe is required - hence I am not changing this. 

I'lll map the  rest of `component.dataset` `component.type` and `component.binary` like you defined above

AndersonQ:
hey @khushijain21 I think we can for my PR (https://github.com/elastic/integrations/pull/17016) instead? I'll get it ready for review tomorrow. 
I made more changes, including unifying the `component.*` mapping on the same file. Which I believe helps to make it easier to maintain

PR #17016, opened 2026-01-21
[elastic_agent]: single `system.process.cpu.*` and `component.*` mapping

AndersonQ:
## Proposed commit message

```
Standardise `system.process.cpu.*`, `system.process.memory.*` and `component.*` mapping on all metrics data streams and for filebeat_input_logs.

As the data streams are only created if data is sent to them, it's possible a data stream with more mappings is never created, thus the filed does not exist. In that case, if the one of the "non-existing" field is used on a dashboard, an error is shown to the user claiming the field is missing.
In order to prevent that, all data streams need all the mappings used on dashboards.
```

## Checklist

- ~~[ ] I have reviewed [tips for building integrations]~~(https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.~~

### missing `system.process.cpu.total.pct`

 - spin up a 8.15 stack - the lowest compatible version
 - create an agent policy WITHOUT System or any other integration. Only logs and metrics monitoring
 - install the agent
 - wait a few minutes, until there are a few data points.
 - confirm the error `Field system.process.cpu.total.pct was not found.` appears on the CPU Usage visualisations (the agent level and components level)
 - install the integration from this PR
 - confirm the new version is installed: navigate to `http://[your-stack]app/integrations/detail/elastic_agent/overview` and verify the version
 - go back to the dashboard - the error should have disappeared 
 - select "explore in discover" on the 3-dots menu of the visualisation and add the following filter to the query bar `system.process.cpu.total.pct : *`
 - no metric with `system.process.cpu.total.pct` should appear

### missing `component.id` 

 - the missing `component.id` seems to appear only if the dashboard is opened when not enough data has been ingested. See the issue https://github.com/elastic/integrations/issues/16567 for details.
 - so just confirm the error does not appear

## Related issues

- Closes https://github.com/elastic/integrations/issues/16982
- Closes https://github.com/elastic/integrations/issues/16567

### With issue
Screenshot from 2026-01-22 08-55-43

### Fixed
Screenshot from 2026-01-22 09-02-26


packages/elastic_agent/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.6.12"
+  changes:
+    - description: Standardize `system.process.cup.*` and `component.*` mapping on all metrics data streams.
```
khushijain21:
```suggestion
    - description: Standardize `system.process.cpu.*` and `component.*` mapping on all metrics data streams.
```

khushijain21:
LGTM otherwise!

packages/elastic_agent/changelog.yml
AndersonQ: fixed. thanks for catching it!

cmacknz:
> Standardise `system.process.cpu.*`, `system.process.memory.*` and `component.*` mapping on all metrics data streams and for filebeat_input_logs.

The Beat based inputs report `system.process.cpu` and `system.process.memory.size`, not `system.process.total.pct` or `system.process.rss.bytes` which are reported only for endpoint-security because we have to use the actual `system/metrics` Metricbeat module to collect it's data since it is a separate, non-Beat service and can't use the /stats implementation from the elastic-agent-system-metrics packages.

https://github.com/elastic/elastic-agent/blob/8caa0725230f1b9e176fdfeb68ec65d245d12803/internal/pkg/agent/application/monitoring/component/v1_monitor.go#L1383-L1395

```go
		// I should be able to see the CPU Usage on the running machine. Am using too much CPU?
		map[string]interface{}{
			"from": "http.agent.beat.cpu",
			"to":   "system.process.cpu",
		},

		// I should be able to see the Memory usage of Elastic Agent. Is the Elastic Agent using too much memory?
		map[string]interface{}{
			"from": "http.agent.beat.memstats.memory_sys",
			"to":   "system.process.memory.size",
		},
```

I think including mappings to data streams for metrics that are not sent or reported makes this more confusing to maintain.

Can we look at making the data collected from `endpoint-security` match with the existing metrics we collect from beats so that the data is consistent allowing this integration and the Fleet UI handle things in a consistent way? Clearly the existing way of doing this is causing us to miss things and have to account for special cases.

AndersonQ:
> I think including mappings to data streams for metrics that are not sent or reported makes this more confusing to maintain.

I agree it's not ideal. But in order to use it on visualisations, we need to always have it mapped in at least one data stream/index the dashboard queries. Otherwise we get the errors we see here.

> Can we look at making the data collected from `endpoint-security` match with the existing metrics we collect from beats so that the data is consistent allowing this integration and the Fleet UI handle things in a consistent way?

We can look at it. The thing to keep in mind, specially with the dashboards is that it needs to work with the older agents. So, mappings on the hard coded config won't fix the issue because the dashboards need to work with the older agents. Remapping the fields on the integration level fixes the issue to the new data, not the existing data. What might be a limitation we're willing to take.

I confess I keep thinking why we hard-code the monitoring config, why doesn't it come from the integration so we could change it as we upgrade the integration. That would give us much more flexibility.

cmacknz:
> I confess I keep thinking why we hard-code the monitoring config, why doesn't it come from the integration so we could change it as we upgrade the integration. That would give us much more flexibility.

This is an interesting idea but you can solve the field name problem for old agents with an ingest pipelines as a much simpler alternative.

Then you don't need mappings in data streams for data they'll never contain, the data from endpoint would match the existing mappings for `system.process.cpu"`.

AndersonQ:
> you can solve the field name problem for old agents with an ingest pipelines as a much simpler alternative.

I thought about it, however, unless I'm missing something, it'd only apply to new data, the old data would not work on the dashboard anymore. 

If we want to solve it just for the new data, if the dashboard might not work for old data, then it's a simpler problem.

Is it the idea? Is it fine if a solution works only for the newly ingested data?

AndersonQ:
> > you can solve the field name problem for old agents with an ingest pipelines as a much simpler alternative.
> 
> I thought about it, however, unless I'm missing something, it'd only apply to new data, the old data would not work on the dashboard anymore.
> 
> If we want to solve it just for the new data, if the dashboard might not work for old data, then it's a simpler problem.
> 
> Is it the idea? Is it fine if a solution works only for the newly ingested data?

@cmacknz the ingest pipeline can fix it for new data, but then the dashboard would miss the metrics for endpoint for the data ingested before the ingest pipeline is in place. Is it acceptable?

cmacknz:
So the problem here is mostly that we picked the wrong name for the metrics.

If we change agent to use the correct name, could we avoid breaking the dashboard by setting up an alias mapping so the queries written against the new name in the dashboard work against both names?

https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/field-alias

PR #17018, opened 2026-01-21
Add configuration snippets to System OTel integration docs

theletterf:
Contributes to https://github.com/elastic/docs-content/issues/4297

## Proposed commit message

Add configuration snippets to System OTel integration docs.

+CC @mlunadia @AlexanderWert

### DIFF
--- a/packages/system_otel/docs/README.md
+++ b/packages/system_otel/docs/README.md
@@ -38,7 +38,6 @@ Make sure that your receivers section in the Collector configuration file is as 
         metrics:
           system.memory.utilization:
             enabled: true
-      # process scraper is disabled for now: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39423
       #process:
       #  mute_process_exe_error: true
       #  mute_process_io_error: true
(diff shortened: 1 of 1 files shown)


packages/system_otel/docs/README.md
```diff
@@ -1,15 +1,156 @@
...
+  # Receiver for CPU, Disk, Memory, and Filesystem metrics
+  hostmetrics/system:
+    collection_interval: 60s
+    scrapers:
+      filesystem:
+      memory:
+        metrics:
+          system.memory.utilization:
+            enabled: true
+      # process scraper is disabled for now: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39423
```
AlexanderWert: nit: should we keep that commented-out section here? maybe let's remove it and we could add it back once we would actually add it to the default config?

packages/system_otel/docs/README.md
```diff
@@ -1,15 +1,156 @@
...
+        metrics:
+          system.cpu.utilization:
+            enabled: true
+          system.cpu.logical.count:
+            enabled: true
+      memory:
+        metrics:
+          system.memory.utilization:
+            enabled: true
+      # process scraper is disabled for now: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39423
```
AlexanderWert: nit: should we keep that commented-out section here? maybe let's remove it and we could add it back once we would actually add it to the default config?

AlexanderWert (approved):

packages/system_otel/docs/README.md
theletterf:
```suggestion
```

packages/system_otel/docs/README.md
theletterf:
```suggestion
```

PR #16610, opened 2025-12-17
[azure logs] Make routing.category field available to pipelines

zmoog:
## Proposed commit message

Support for custom log event routing in the Azure Logs v2 integration.

The Azure Logs v2 default pipeline now extracts the `routing.category` field from the log event and makes it available to custom pipelines, allowing users to apply custom routing logic.

For example, `SQLSecurityAuditEvents` would normally be routed to the `azure.platformlogs` dataset because Synapse log events are not yet supported natively.

By creating the `logs-azure.events@custom` pipeline with the following content:

```
PUT _ingest/pipeline/logs-azure.events@custom
{
  "processors": [
    {
      "reroute": {
        "if": "ctx?.routing?.category == \"SQLSecurityAuditEvents\"",
        "dataset": "azure.synapse"
      }
    }
  ]
}
```

Users can now route these events to a dedicated `azure.synapse` dataset, or an existing dataset, like `azure.activitylogs`.

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [ ] I have added an entry to my package's `changelog.yml` file.
- [ ] ~~I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).~~
- [ ] ~~I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices)~~

## How to test this PR locally

Define the `logs-azure.events@custom` with custom routing logic:

```text
PUT _ingest/pipeline/logs-azure.events@custom
{
  "processors": [
    {
      "reroute": {
        "if": "ctx?.routing?.category == \"SQLSecurityAuditEvents\"",
        "dataset": "azure.synapse"
      }
    }
  ]
}
```

Send a sample log event:

```text
POST logs-azure.events-default/_doc
{
  "@timestamp": "2025-12-17T07:15:26+01:00",
  "message": """{"category":"SQLSecurityAuditEvents"}"""
}
```

Search for the results:

```text
POST logs-azure.synapse-default/_search
```

<details>
  <summary>Results</summary>

```json
{
  "took": 1,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
[...]

PR #17027, opened 2026-01-21
[azure logs] Fix sign-in logs category check

### LINKED ISSUE #16994: [Azure Logs]: Null def reference on set processor
### Integration Name

Azure Logs [azure]

### Dataset Name

events

### Integration Version

1.34.1

### Agent Version

8.19.7

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.19.7

N/A

### Software/API Version

_No response_

### Error Message

cannot access method/field [endsWith] from a null def reference

### Event Original

_No response_

### What did you do?

Default integration config

### What did you see?

Received cannot access method/field [endsWith] from a null def reference error message on Azure Logs.

There should be a check on the set processor in the event that the field doesn't exist.

### Anything else?

Suggestions to check field exists on the set processor:

ctx.routing?.category.endsWith("SignInLogs")
ctx?.routing?.category != null && ctx.routing?.category.endsWith("SignInLogs")

The set processor affected is:

```
 - set:
      field: event.dataset
      value: azure.signinlogs
      # Use same logic as the `signinlogs` stream that drops any document that doesn't end with `SignInLogs`.
      if: 'ctx.routing?.category.endsWith("SignInLogs")'
```

zmoog:
## Proposed commit message

Add an explicit check on `ctx.routing?.category` to make sure it's not null before calling the `endsWith("SignInLogs")` method.

We can't assume category is always set.

## Checklist

## Related issues

- Closes #16994

### DIFF
--- a/packages/azure/changelog.yml
+++ b/packages/azure/changelog.yml
@@ -1,3 +1,8 @@
+- version: "1.35.1"
+  changes:
+    - description: Fix pipeline error when the `category` field is missing.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/17027
 - version: "1.35.0"
   changes:
     - description: Add support for AMQP-over-WebSockets transport protocol and proxy configuration.
--- a/packages/azure/manifest.yml
+++ b/packages/azure/manifest.yml
@@ -1,6 +1,6 @@
 name: azure
 title: Azure Logs
-version: "1.35.0"
+version: "1.35.1"
 description: This Elastic integration collects logs from Azure
 type: integration
 icons:


Kavindu-Dodan (approved):

packages/azure/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -116,7 +116,7 @@ processors:
       field: event.dataset
       value: azure.signinlogs
       # Use same logic as the `signinlogs` stream that drops any document that doesn't end with `SignInLogs`.
-      if: 'ctx.routing?.category.endsWith("SignInLogs")'
+      if: 'ctx.routing?.category != null && ctx.routing?.category.endsWith("SignInLogs")'
```
efd6, line 119:
```suggestion
      if: 'ctx.routing?.category != null && ctx.routing.category.endsWith("SignInLogs")'
```

muthu-mps, line 119: If we are not processing the logs with no category then should we **drop the event which doesn't have category**?

zmoog: Um, I think we keep the log events. 

Two reasons:

- Azure has been inconsistent in naming the field that contains the log category. So far we know they mostly usage `category`, but also found `Category` and `CategoryValue` in the wild. So I  guess an suboptimal indexing is better than losing the log event.
- custom routing: we recently [added](https://github.com/elastic/integrations/pull/16610) `routing.category` as candidate field for custom routing, but users may want to customize the routing based on other criteria and fields. If we drop the log event in the main pipeline, it will never reach the custom pipeline.

zmoog: Does this pipeline guarantee the routing field? It looks like it might be optional here.

mykola-elastic: if we remove the question mark then we would need to add `ctx.routing != null` in the beginning and remove all question marks.

I think we should keep it as it is in the PR now

PR #17049, opened 2026-01-25
Add Launch Stack button for ECF to AWS OTel content packs

strawgate:
Add a button to the Readmes
Screenshot 2026-01-24 at 8 51 01 PM

Pre-populates a stack name and type
pre-populate


mykola-elastic (approved):
That's a nice looking and handy button.

Just one question, would you actually use it? (putting oneself in the shoes of customer's SRE/Dev/Ops)

I think I would check the button link contents and then cloudformation yaml file it addresses, and then I would go to the EDOT deployment docs anyway to explore the nuances. So, the button would be more cosmetic for me, at least for the first usage.

cc @ishleenk17

stefans-elastic (approved):

strawgate:
@mykola-elastic i think it's more about drawing attention to the availability of Cloud Forwarder when looking for a specific integration. Right now the onboarding we have assumes you already know you need Cloud Forwarder

PR #17056, opened 2026-01-26
[hostmetrics_input_otel] Initial draft of System Input Package using hostmetricsreceiver

mykola-elastic:
## Proposed commit message

See title.

## Checklist

### The dashboards were auto-installed from `system_otel` package upon receiving some `hostmetricsreceiver.otel` data:

Screenshot 2026-01-27 at 10 29 02

### Example: dashboard got filled up with data:

Screenshot 2026-01-27 at 10 33 36

### Configuration section

Screenshot 2026-01-29 at 13 38 08

### DIFF
--- a/packages/hostmetrics_input_otel/manifest.yml
+++ b/packages/hostmetrics_input_otel/manifest.yml
@@ -96,7 +96,7 @@ policy_templates:
         title: Resource Detection Processor / System configuration
         description: Configuration for [resourcedetectionprocessor's system component](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/internal/system/documentation.md)
         show_user: true
-        required: true
+        required: false
         default: |
           hostname_sources: ["os"]
           resource_attributes:


packages/system_input_otel/manifest.yml
```diff
@@ -0,0 +1,37 @@
...
+title: "System OpenTelemetry Input Package"
+version: 0.1.0
+source:
+  license: "Elastic-2.0"
+description: "System OpenTelemetry Input Package"
+type: input
+categories:
+  - monitoring
+  - os_system
+  - observability
```
ishleenk17: Add opentelemtry category

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
```diff
@@ -0,0 +1,67 @@
...
+            enabled: true
+          process.memory.utilization:
+            enabled: true
+          process.disk.operations:
+            enabled: true
+      network:
+      processes:
+      load:
+      disk:
+      filesystem:
```
ishleenk17, line 44: Are we using metrics from all of these scarpers fir the dashboards ?

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
```diff
@@ -0,0 +1,67 @@
...
+processors:
+  resourcedetection:
+    detectors: ["system"]
+    system:
+      hostname_sources: ["os"]
+      resource_attributes:
```
ishleenk17: For processors, we were discussing to use minimal in the input package.
@jsoriano: We wanted to discuss that for something like a system input package. 
Where we need these for the dashboards. 
WDYT, if this should come through composable package or here ?
cc: @tommyers-elastic

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
```diff
@@ -0,0 +1,67 @@
...
+        metrics:
+          process.threads:
+            enabled: true
+          process.open_file_descriptors:
+            enabled: true
+          process.memory.utilization:
+            enabled: true
+          process.disk.operations:
+            enabled: true
+      network:
```
ishleenk17, line 32: Also, I was thinking if we enable all the scrapers, tomorrow when we use system composable package and we want to say have only data for cpu and not for others, doing it this way the flexibility will not be there. 
@jsoriano @tommyers-elastic : Should these be configurable and come through composable package int he future ?

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
mykola-elastic: yes, these scrapers are referenced in the [system_otel's receivers configuration section](https://www.elastic.co/docs/reference/integrations/system_otel#receivers-configuration)

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
mykola-elastic: FYI, used processors from https://www.elastic.co/docs/reference/integrations/system_otel#processors-configuration

dashboards also rely on these resource attributes

packages/system_input_otel/_dev/build/docs/README.md
```diff
@@ -0,0 +1,13 @@
+# System OpenTelemetry Input Package
+
+## Overview
+The System OpenTelemetry Input Package for Elastic enables collection of telemetry data about the host system through OpenTelemetry protocols using the [hostmetricsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver#host-metrics-receiver).
```
jsoriano: In principle OTel input packages should be basically wrappers over receivers. I think this package should be called `hostmetrics_input_otel`.

I guess in the future we will have something like a system package for OTel, this one can be called `system_otel`.

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
jsoriano: Yes, I'd say this needs to be configurable with variables, and the defaults could comply with what we need for the dashboards.

Though not sure what would be the best approach here, some ideas:
* Enable scraper by scraper with a multi-selector that then can be used in templates as something like this (but we may need more helpers):
  ```yaml
    scrapers:
      {{#if contains "cpu" scrappers }}
      cpu:
        metrics:
          system.cpu.utilization:
            enabled: true
          system.cpu.logical.count:
            enabled: true
      {{/if}} 
      {{#if contains "memory" scrappers }}
      memory:
        metrics:
          system.memory.utilization:
            enabled: true
      {{/if}}
  ```
* Have one variable for each scrapper, but can be too much.
* Allow to configure a yaml block.
* A combination of any of the above.

packages/system_input_otel/agent/input/input.yml.hbs
```diff
@@ -0,0 +1,67 @@
+receivers:
+  hostmetrics/system:
```
jsoriano: No need to add identifiers here, unless there were multiple entries for the same receiver.
```suggestion
  hostmetrics:
```
Fleet will add unique identifiers in the final configuration.

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
jsoriano: Maybe in this case we don't need to make it configurable, as this will be pretty common when using the `hostmetrics` receiver?

Though I would prefer to leave this package generic, and then be the system composable package what defines it.

For the case of generic settings that will be used in many packages we don't have a good solution yet, apart from duplicating it on every package. But in the mid term we would like to have something like base packages that other packages can "require" and reuse (https://github.com/elastic/package-spec/issues/1029).

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
mykola-elastic: @ishleenk17 @jsoriano there is a draft PR having that https://github.com/elastic/integrations/pull/17060

ishleenk17: @mykola-elastic : Lets make the scrapers configurable.

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
ishleenk17: @mykola-elastic : For now lets have the processors as is. Not making it configurable

packages/system_input_otel/_dev/build/docs/README.md
ishleenk17: Yes, lets rename this to hostmetrics_input_otel @mykola-elastic

mykola-elastic:
@ishleenk17
addressed the comments:
- renamed the package
- made all scrapers configurable (possible to disable with toggles)
- tests added
- additionally tested it manually

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
ishleenk17: Can you please share the screenshot of how this integration UI looks now. The toggles, for reference

mykola-elastic: @ishleenk17 

---

Screenshot 2026-01-29 at 13 37 58

---

Screenshot 2026-01-29 at 13 38 08

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
mykola-elastic: @jsoriano @ishleenk17 how about something like this?

also, for names `enable_system_resourcedetection` vs `enable_host_metadata`

```yaml
processors:
  {{#if enable_system_resourcedetection}}
  resourcedetection:
    detectors: ["system"]
    system:
      hostname_sources: ["os"]
      resource_attributes:
        # ...
  {{/if}}

service:
  pipelines:
    metrics:
      receivers: [hostmetrics]
      processors: 
        {{#if enable_system_resourcedetection}}
        - resourcedetection
        {{/if}}
        {{#if enable_batch}} # Example: later we will be able to put stuff here
        - batch
        {{/if}}
```

ishleenk17: resourcedetection:
    detectors: ["system"]
        system: {{system_config}}
        
        @mykola-elastic : Lets do this

packages/hostmetrics_input_otel/_dev/build/docs/README.md
```diff
@@ -0,0 +1,13 @@
+# Host Metrics OpenTelemetry Input Package
+
+## Overview
+The Host Metrics OpenTelemetry Input Package for Elastic enables collection of telemetry data about the host system through OpenTelemetry protocols using the [hostmetricsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver#host-metrics-receiver).
+
+
+### How it works
+This package receives telemetry data from host system by configuring the Input Package, which uses the hostmetrics receiver present in the EDOT collector, which then forwards the data to Elastic Agent. The Elastic Agent processes and enriches the data before sending it to Elasticsearch for indexing and analysis. Once the data arrives into Elasticsearch, its corresponding [System OpenTelemetry Assets Package](https://www.elastic.co/docs/reference/integrations/system_otel) gets auto installed and the dashboards light up.
+
+
```
ishleenk17, line 11: Lets add a configuration section and point to the upstream config section for the same. 
Assuming we are using the same param names as upstream

packages/hostmetrics_input_otel/agent/input/input.yml.hbs
mykola-elastic: @ishleenk17 That's how it looks like with the default being the current config:

---

Screenshot 2026-02-03 at 10 49 42

---

Screenshot 2026-02-03 at 10 50 37

packages/hostmetrics_input_otel/_dev/build/docs/README.md
mykola-elastic: done

Issue #15730, opened 2025-10-22
[OTel Host Metrics Receiver] Input Package


strawgate:
Similar to https://github.com/elastic/integrations/pull/14315/files we should add an input package for the hostmetrics receiver https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver

andrewkroh:
Closed by https://github.com/elastic/integrations/pull/17056

PR #17060, opened 2026-01-27
[Draft] OTel Input Package for Host Metrics Receiver

strawgate:
https://github.com/elastic/integrations/issues/15730


packages/hostmetrics_otel/img/hostmetrics-logo.svg
graphaelli, line 1: what is that supposed to be?

PR #17094, opened 2026-01-28
[Citrix_adc] Add headerless log patterns

robester0403:
## Proposed commit message

Added grok processing for headerless logs


packages/citrix_adc/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -188,6 +188,8 @@ processors:
         - yyyy/MM/dd:HH:mm:ss z
         - MM/dd/yyyy:HH:mm:ss
         - MM/dd/yyyy:HH:mm:ss z
+        - dd/MM/yyyy:HH:mm:ss
+        - dd/MM/yyyy:HH:mm:ss z
```
taylor-swanson: This isn't going to work as expected, you can't have both MM/dd and dd/MM in the same format list as it is ambiguous.

For instance, what is 04/10/2025?

If they are expecting `dd/MM`, it's going to select `MM/dd` since it is first in the list. So basically, it only works as expected when the date is > 12.

We had this come up in a previous PR for a different integration... Let me see if I can find it. If I recall, we had to have a configurable for the date format.

taylor-swanson: @jrmolin, was this something you added at one point to an integration? A configurable date format?

robester0403: The customer might have to set this as a config since we can't do this reliably:

>         fields:
>           _conf:
>             tz_offset: UTC
>             # custom_date_format: <CUSTOM_DATE_FORMAT>
>             #  # Custom Date Format: Format to use to parse the date/time fields in the data.
>             #  #This is used to build a Java DateTimeFormatter in [Painless](https://www.elastic.co/guide/en/elasticsearch/painless/current/painless-datetime.html) to parse the incoming date. For Day/Month/Year... parsing, this could be "dd/MM/yyyy:HH:mm:ss".
>             #  #If not populated, this matches any of "yyyy/MM/dd:HH:mm:ss", "MM/dd/yyyy:HH:mm:ss", or the ISO8601 standard.

jrmolin: yeah, we added something this custom date formatter, but i don't know if any customer is going to use it. the citrix devices send back local time, which can either be `MM/dd` or `dd/MM`, depending on the locale. so they'd have to be really sure their devices are all one locale or another

jrmolin: the custom date parsing happens immediately above this

robester0403: alright I am going to revert the date changes and notify support

packages/citrix_adc/data_stream/log/_dev/test/pipeline/test-citrix-sslvpn-message.log
```diff
@@ -1,5 +1,10 @@
...
-<135> 07/09/2025:19:48:23 GMT test-ns-internal-b 0-PPE-0 : default AAATM HTTPREQUEST 318039900 0 : Context 48x0529@hs.mydomain.com@175.16.199.1 - SessionId: 39526570 - test.customdomain.com User 48x0529@hs.mydomain.com : Group(s) N/A : Vserver 67.43.156.1:8085 - 07/09/2025:19:48:23 GMT : SSO is OFF : CONNECT test.customdomain.com:443 - -
+<135> 07/09/2025:19:04:21 GMT test-ns-internal-b 0-PPE-0 : default AAATM HTTPREQUEST 318039900 0 : Context 48x0529@hs.mydomain.com@175.16.199.1 - SessionId: 39526570 - test.customdomain.com User 48x0529@hs.mydomain.com : Group(s) N/A : Vserver 67.43.156.1:8085 - 07/09/2025:19:48:23 GMT : SSO is OFF : CONNECT test.customdomain.com:443 - -
```
taylor-swanson, line 5: Was the first timestamp changed to make sure that we aren't grabbing the second timestamp by accident?

taylor-swanson (approved):

packages/citrix_adc/data_stream/log/_dev/test/pipeline/test-citrix-sslvpn-message.log
robester0403: Yes the first change was to test the timestamp that the customer had issue with, but I reverted back to one that is similar to the original since we are solving it now through the config.

PR #17098, opened 2026-01-28
[Beaconing] Update readme for detection rules

susan-shu-c:
## Proposed commit message

Update readme with prebuilt rules supported version.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/beaconing/changelog.yml
+++ b/packages/beaconing/changelog.yml
@@ -2,7 +2,7 @@
   changes:
     - description: Clarify prebuilt rules available from version 8.11.3 and above
       type: enhancement
-      link: TODO
+      link: https://github.com/elastic/integrations/pull/17098
 - version: "1.5.1"
   changes:
     - description: Update package docs with customization steps for ML jobs and transforms


packages/beaconing/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 3.0.0
 name: beaconing
 title: "Network Beaconing Identification"
-version: 1.5.1
+version: 1.5.2
 source:
```
sodhikirti07, line 5: Could you also update the min kibana version in the manifest?

sodhikirti07: @susan-shu-c Actually, you don't need to change this as the version is only applicable to detection-alerts.

susan-shu-c: Hi, thanks for the comment, glad that not changing it makes sense!

susan-shu-c: Just to check, you're saying that the PR as-is is good?

sodhikirti07: @susan-shu-c Yes, it lgtm!

PR #17099, opened 2026-01-28
IIS input package

nimarezainia:
The integration:
image

Agent Policy:
image

Config on the Agent (with variables modified:
image

Ingested Data and the dashboard:
image


strawgate:
@claude can you switch this to making a time series datastream right now its just making a normal one

strawgate:
@claude do the other otel input packages define fields at all? i think the input fields file is causing an index template to be created which is making this a non time series data stream?

nimarezainia:
@claude remove the fields/ directory as suggested

jlind23:
@claude please add an entry in the codeowner file for this integration. The team owning this is @elastic/ecosystem

jlind23:
@claude fix the review comments and also make sure the owner in the manifest is the ecosystem team.

jlind23:
@claude please skip system test for windows only integration please.

jlind23:
@teresaromero @mrodm I believe this is now ready to be reviewed / merged.
cc @kpollich

packages/iis_input_otel/manifest.yml
```diff
@@ -0,0 +1,46 @@
...
+version: "0.1.0"
+source:
+  license: "Elastic-2.0"
+description: "IIS OpenTelemetry Input Package"
+type: input
+categories:
+  - web
+  - observability
+  - opentelemetry
+  - os_system
```
mrodm: This `os_system` category is not part of the current categories in the integration package:

https://github.com/elastic/integrations/blob/ff185e9f4c3453670350604036d04788860b4d73/packages/iis/manifest.yml#L12-L14

I think I would keep the same categories as in the integration package for now.

packages/iis_input_otel/img/iis_otellogo.svg
mrodm, line 1: Following this thread https://github.com/elastic/integrations/pull/17117#discussion_r2741353690

It would be interesting if the logo for this input package could include a small OpenTelemetry logo on the side too.

packages/iis_input_otel/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "0.1.0"
+  changes:
+    - description: Initial draft of the package with time series support
```
mrodm: According to the description there is support for time series, but there are no settings about it in the manifest.

Is that support related to this PR https://github.com/elastic/package-spec/pull/1066?
 
Should it be removed that reference from the description ?

jlind23: @nimarezainia can you speak to this?

jlind23:
@claude please make sure that integration categories are the same as the standard IIS package :
- https://github.com/elastic/integrations/blob/ff185e9f4c3453670350604036d04788860b4d73/packages/iis/manifest.yml#L12-L14
-  input package could include a small OpenTelemetry logo on the side like other input packages are doing

jlind23:
@claude could you please update the logo of this package in order to include a small OpenTelemetry logo on the side like we have done for other otel packages.

packages/iis_input_otel/img/iis_otellogo.svg
mrodm, line 1: Looking into other OTEL packages, these logos do not have a white background.
These logos are usually created with transparent background.

Examples of logos with this transparent background:
- https://github.com/elastic/integrations/blob/main/packages/gcp_audit_otel/img/gcp_otel_logo.svg
- https://github.com/elastic/integrations/blob/main/packages/httpcheck_otel/img/http-logo.svg


And actually, there is already a logo that maybe it could be used here too in the `iis_otel` package (available already in main branch):
- https://github.com/elastic/integrations/blob/main/packages/iis_otel/img/iis_otel.svg

It already contains the IIS plus the OpenTelemetry logo on the right side.

packages/iis_input_otel/_dev/test/policy/test-default.yml
mrodm, line 1: I would add another policy test to ensure that the default values in the manifest for the variables are taken into account too.

That policy test would be like:

```yaml
vars: ~
```

The policy test expected file should be the same as packages/iis_input_otel/_dev/test/policy/test-default.expected but changing these lines:

```yaml
receivers:
    iis/componentid-0:
        collection_interval: 10s
        initial_delay: 1s
```

jlind23:
@claude could you please address @mrodm comments above please.

packages/iis_input_otel/_dev/test/system/test-default-config.yml
teresaromero, line 1: have you been able to run this tests locally? i was trying to run them on windows but it does not run

teresaromero: tried to run the test at windows machine, but with docker, and still did not work, so system test will not be able to run. 👍🏻 okay with leaving this as skipped

kpollich (approved):

packages/iis_input_otel/changelog.yml
nimarezainia: it is related but I am not sure what change needs to be made.

@mrodm what change does the package author need to make to ensure it is a time series package?

mrodm: > it is related but I am not sure what change needs to be made.
> 
> @mrodm what change does the package author need to make to ensure it is a time series package?

@nima I think it is needed this spec change https://github.com/elastic/package-spec/pull/1066 to allow defining the time series mode, or waiting for this other PR in kibana: https://github.com/elastic/kibana/pull/251205#discussion_r2755543623

Related to https://github.com/elastic/integrations/issues/17183

PR #17100, opened 2026-01-29
cyberark_pta: retain observer IP address or hostname if available from syslog message

efd6:
## Proposed commit message

See title.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/cyberark_pta/data_stream/events/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/cyberark_pta/data_stream/events/elasticsearch/ingest_pipeline/default.yml
@@ -36,11 +36,10 @@ processors:
         - '^%{SYSLOGTIMESTAMP} (?:%{IP:observer.ip}|%{HOSTNAME:observer.hostname}) CEF'
       ignore_failure: true
       ignore_missing: true
-  - set: 
-     tag: set_host_ip_1d72afcb 
-     field: host.ip 
-     value: ['{{{host.ip}}}'] 
-     if: ctx.host?.ip instanceof String 
+  - script:
+      if: ctx.observer?.ip instanceof String
+      source: |-
+        ctx.observer.ip = Collections.singletonList(ctx.observer.ip);
   - rename:
       field: message
       target_field: event.reason


packages/cyberark_pta/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -30,6 +30,16 @@ processors:
...
+  - grok:
+      field: event.original
+      patterns:
+        - '^%{SYSLOGTIMESTAMP} (?:%{IP:observer.ip}|%{HOSTNAME:observer.hostname}) CEF'
+      ignore_failure: true
+      ignore_missing: true
+  - script:
+      if: ctx.observer?.ip instanceof String
+      source: |-
+        ctx.observer.ip = Collections.singletonList(ctx.observer.ip);
```
kcreddy: Just curious, do we know if this is faster than a `set` processor, say like [this](https://github.com/elastic/integrations/blob/main/packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/default.yml#L455-L459)?

https://github.com/elastic/integrations/blob/a6223f4e9fb317405acf5fdc8e04b88941ce5594/packages/panw/data_stream/panos/elasticsearch/ingest_pipeline/default.yml#L455-L459

efd6: I'll look into it.

efd6: If a benchmark set only includes the syslog form of the data (so the only case where this will be exercised), then the script shows up as costing 200ns and the set does not show up at all, presumably costing less than 100ns. There's not much in it, but I've taken your suggestion.

efd6: Actually, no. This does not work as expected.
```
--- Test results for package: cyberark_pta - START ---
FAILURE DETAILS:
cyberark_pta/events test-active-dormant-user-syslog.json:
--- want
+++ got
@@ -59,9 +59,7 @@
                 "url": "None."
             },
             "observer": {
-                "ip": [
-                    "81.2.69.144"
-                ],
+                "ip": "81.2.69.144",
                 "product": "PTA",
                 "vendor": "CyberArk",
                 "version": "11.4"
```

kcreddy: Thanks for taking a look

kcreddy (approved):

PR #17174, opened 2026-01-30
Add AGENTS.md

kpollich:
## Proposed commit message

Add AGENTS.md

## Notes

I generated this `AGENTS.md` file w/ Claude Opus + Cursor by reviewing relevant documentation, introspecting the repository structure, and tweaking the doc to include a comparison of package types based on https://github.com/elastic/elastic-package/blob/main/docs/howto/package_types.md

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,546 @@
+# AGENTS.md - AI Agent Guidance for Elastic Integrations
+
+This document provides essential context for AI agents working in the Elastic Integrations repository.
+
+## Repository Overview
+
+This repository contains sources for **Elastic Integrations** (Elastic Packages). Each package defines how to observe a specific product or service with the Elastic Stack, including:
+
+- Configuration for the Elastic Agent
+- Assets such as Kibana dashboards and Elasticsearch index templates
+- Documentation about the package
+- Tests to ensure functionality
+
+Built packages are published to the [Elastic Package Registry](https://epr.elastic.co) and served to Fleet in Kibana.
+
+## Key External Resources
+
+- **Integrations Developer Guide**: https://www.elastic.co/guide/en/integrations-developer/current/index.html
+- **Package Specification**: https://github.com/elastic/package-spec
+- **elastic-package CLI**: https://github.com/elastic/elastic-package
+- **Package Registry**: https://github.com/elastic/package-registry
+
+## Repository Structure
+
+```
+integrations/
+├── packages/                 # All integration packages (300+ packages)
+│   └── <package>/
+│       ├── manifest.yml      # Package metadata, owner, version, Kibana constraints
+│       ├── changelog.yml     # Version history (newest entries on top)
+│       ├── data_stream/      # Data stream definitions
+│       │   └── <stream>/
+│       │       ├── manifest.yml
+│       │       ├── fields/           # Field definitions (fields.yml, ecs.yml, base-fields.yml)
+│       │       ├── elasticsearch/    # Ingest pipelines
+│       │       │   └── ingest_pipeline/
+│       │       └── agent/            # Agent stream configs (.yml.hbs templates)
+│       │           └── stream/
+│       ├── _dev/             # Development assets
+│       │   ├── build/        # Build configuration (build.yml, docs templates)
+│       │   ├── deploy/       # Docker compose for local testing
+│       │   └── test/         # Test configurations
+│       ├── docs/             # Package documentation
+│       ├── kibana/           # Kibana assets (dashboards, visualizations)
+│       └── img/              # Icons and screenshots
+├── docs/                     # Repository-wide documentation
+│   └── extend/               # Extended developer documentation
+├── .buildkite/               # CI pipeline configurations
+├── dev/                      # Internal dev tools (Go utilities)
+├── .github/CODEOWNERS        # Package ownership assignments
+└── magefile.go               # Mage build targets
+```
+
+## Core Tool: elastic-package
+
+The `elastic-package` CLI is essential for all development tasks. Key commands:
+
+| Command | Description |
+|---------|-------------|
+| `elastic-package create package` | Scaffold a new integration package |
+| `elastic-package create data-stream` | Add a new data stream to a package |
+| `elastic-package check` | Run format, lint, and build together |
+| `elastic-package build` | Build the package (outputs to `build/`) |
+| `elastic-package lint` | Validate against package-spec |
+| `elastic-package format` | Format package files (JSON, YAML) |
+| `elastic-package stack up -d -v` | Start local Elastic Stack for testing |
+| `elastic-package test` | Run tests (asset, pipeline, system, static, policy) |
+| `elastic-package install` | Install package to Kibana |
+| `elastic-package changelog` | Manage changelog entries |
+| `elastic-package clean` | Clean build artifacts |
+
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,546 @@
...
+# Navigate to the package
+cd packages/<package-name>
+
+# Run all tests
+elastic-package test
+
+# Run specific test type
+elastic-package test pipeline
+elastic-package test system
+elastic-package test static
```
mrodm, line 156: Should it be added the other two test types ? 


```suggestion
elastic-package test static
elastic-package test asset
elastic-package test policy
```

AGENTS.md
```diff
@@ -0,0 +1,546 @@
...
+- Define field mappings using ECS (Elastic Common Schema)
+- Include ingest pipelines for data processing
+- Provide comprehensive documentation for the monitored service
+
+**Structure:**
+```
+packages/<name>/
+├── manifest.yml              # type: integration
+├── changelog.yml
+├── data_stream/              # One or more data streams
```
mrodm, line 346: Maybe it is not needed to be so explicit here.
Just in case, there could be an elasticsearch folder at the root of the package, containing for instance transforms and other ingest pipelines among other things.

Related:
https://github.com/elastic/package-spec/blob/main/spec/integration/elasticsearch/spec.yml

AGENTS.md
```diff
@@ -0,0 +1,546 @@
...
+
+Useful environment variables for `elastic-package`:
+
+| Variable | Description |
+|----------|-------------|
+| `ELASTIC_PACKAGE_KIBANA_HOST` | Kibana host URL |
+| `ELASTIC_PACKAGE_ELASTICSEARCH_HOST` | Elasticsearch host URL |
+| `ELASTIC_PACKAGE_ELASTICSEARCH_USERNAME` | ES username |
+| `ELASTIC_PACKAGE_ELASTICSEARCH_PASSWORD` | ES password |
+| `STACK_VERSION` | Force specific stack version for testing |
```
mrodm, line 541: This environment variable is just taken into account in the CI pipelines of this repository (there is some logic for that in the CI scripts). But it is not honored by elastic-package directly.

AGENTS.md
```diff
@@ -0,0 +1,546 @@
...
+
+Load stack environment after starting:
+```bash
+eval "$(elastic-package stack shellinit)"
+```
```
mrodm, line 546: This command should not be required in the latest `elastic-package` versions.

AGENTS.md
```diff
@@ -0,0 +1,546 @@
...
+|---------|-------------|
+| `elastic-package create package` | Scaffold a new integration package |
+| `elastic-package create data-stream` | Add a new data stream to a package |
+| `elastic-package check` | Run format, lint, and build together |
+| `elastic-package build` | Build the package (outputs to `build/`) |
+| `elastic-package lint` | Validate against package-spec |
+| `elastic-package format` | Format package files (JSON, YAML) |
+| `elastic-package stack up -d -v` | Start local Elastic Stack for testing |
+| `elastic-package test` | Run tests (asset, pipeline, system, static, policy) |
+| `elastic-package install` | Install package to Kibana |
```
mrodm, line 68:
```suggestion
| `elastic-package install` | Install package to Kibana |
| `elastic-package uninstall` | Uninstall the package from Kibana |
```

AGENTS.md
```diff
@@ -0,0 +1,546 @@
...
+
+| Command | Description |
+|---------|-------------|
+| `elastic-package create package` | Scaffold a new integration package |
+| `elastic-package create data-stream` | Add a new data stream to a package |
+| `elastic-package check` | Run format, lint, and build together |
+| `elastic-package build` | Build the package (outputs to `build/`) |
+| `elastic-package lint` | Validate against package-spec |
+| `elastic-package format` | Format package files (JSON, YAML) |
+| `elastic-package stack up -d -v` | Start local Elastic Stack for testing |
```
mrodm, line 66:
```suggestion
| `elastic-package stack up -d -v` | Start local Elastic Stack for testing |
| `elastic-package stack down -v` | Stop the local Elastic Stack |
```

andrewkroh:
I've been using a similar file (CLAUDE.md) for guidance: https://gist.github.com/andrewkroh/970cef93ea62cf3e5c2257113c6f623f

Based on that experience, I'd suggest including two items in AGENTS.md:

1. Documentation IP address ranges: Guidance to use IP addresses from ranges reserved for documentation in our samples. This helps us avoid referencing real IP addresses.

2. CEL validation: The [celfmt](https://github.com/elastic/celfmt) validation instruction for CEL programs has proven helpful in steering the LLM toward generating syntactically valid CEL input programs.

AGENTS.md
```diff
@@ -0,0 +1,546 @@
...
+
+### Test Types
+
+| Test Type | Description | Location |
+|-----------|-------------|----------|
+| **Static** | Validates fields, sample events, documentation | Automatic |
+| **Pipeline** | Tests ingest pipelines with sample data | `_dev/test/pipeline/` |
+| **System** | End-to-end tests with real data ingestion | `_dev/test/system/` |
+| **Asset** | Validates ES/Kibana assets load correctly | Automatic |
+| **Policy** | Tests policy configurations | `_dev/test/policy/` |
```
mrodm, line 134: A few weeks ago it was added a new test type `script`:
- location: `_dev/test/scripts`
- Command: `elastic-package test script`

https://github.com/elastic/elastic-package/issues/2944

At the time of writing this comment, there are no packages containing these type of tests yet..

AGENTS.md
mrodm, line 1: I think that updating this file should not trigger that all packages are tested. WDYT?

For that , it should be added a new exception in this line in the `grep -E -v`:

https://github.com/elastic/integrations/blob/a6223f4e9fb317405acf5fdc8e04b88941ce5594/.buildkite/scripts/common.sh#L760

As a note, even updating that line in this PR it will be triggered all the packages since the `.buildkite/scripts/common.sh` file is updated, and this would be expected. But it will take effect in the following PRs.

strawgate:
Claude supports `@path/to/file` syntax and so ideally we'd embed existing information in the repo before writing net-new in the agents.md file

AGENTS.md
```diff
@@ -0,0 +1,546 @@
...
+- Documentation about the package
+- Tests to ensure functionality
+
+Built packages are published to the [Elastic Package Registry](https://epr.elastic.co) and served to Fleet in Kibana.
+
+## Key External Resources
+
+- **Integrations Developer Guide**: https://www.elastic.co/guide/en/integrations-developer/current/index.html
+- **Package Specification**: https://github.com/elastic/package-spec
+- **elastic-package CLI**: https://github.com/elastic/elastic-package
```
jsoriano, line 20: Does it make any difference if we write that more documentation can be found specifically under https://github.com/elastic/elastic-package/tree/main/docs/howto?

strawgate:
Perhaps we should just add the agents_md generator mcp server in an mcp.json in the repo and just put in the agents.md file to use the generator for a detailed agents.md file? and then we dont have to maintain it?

kpollich:
Chatted w/ @strawgate and I'm going to replace this with a simple doc that walks through the process of creating an OTel-based input package that can be reused across agents rather than adding something about broad integration development.

For reference, the Claude workflows in GitHub generate an `agents.md` file when they run so they don't need one to exist in the repo necessarily.

PR #17199, opened 2026-02-02
[otel_internal_telemetry] Replace png icon with svg for OTel internal telemetry package

kpollich:
## Proposed commit message

Replace PNG icon w/ SVG icon for OTel internal telemetry package


kpollich:
Ref https://github.com/elastic/kibana/issues/251306

Confirmed in my local stack provided via `elastic-package stack up` that having this new version of the package fixes the errors in the integrations UI. 

image

PR #17203, opened 2026-02-02
[mysql_otel] Revamp dashboards for comprehensive MySQL monitoring

shmsr:
## Proposed commit message

This PR revamps the MySQL OpenTelemetry integration dashboards to provide a comprehensive and user-friendly monitoring experience. The previous 2 dashboards have been replaced with 4 purpose-built dashboards that better organize metrics based on database observability best practices.

## Checklist

## Screenshots

Please refer to files changed and look under `img/`

### DIFF
--- a/packages/mysql_otel/manifest.yml
+++ b/packages/mysql_otel/manifest.yml
@@ -19,19 +19,19 @@ discovery:
   datasets:
     - name: mysqlreceiver.otel
 screenshots:
-  - src: /img/Overview.png
+  - src: /img/overview.png
     title: Overview
     size: 1715x1920
     type: image/png
-  - src: /img/Performance.png
+  - src: /img/performance.png
     title: Performance
     size: 583x1920
     type: image/png
-  - src: /img/Queries.png
+  - src: /img/queries.png
     title: Queries
     size: 1887x1920
     type: image/png
-  - src: /img/Availability.png
+  - src: /img/availability.png
     title: Availability
     size: 1672x1920
     type: image/png


ishleenk17:
@shmsr : Looked at the live dashboards from serverless for the feedback. 

- DB namespace and schema. How are these different ?
- I see that the control filters are same across all the dashboards. Then I think we can enable the carrying forward of the filter values across dashboards. WDYT ?
- On selecting value in endpoint instance, the db namespace becomes empty. Please check
- Buffer pool used metric tile is the max value and below is the spread. Right ? (In Overview)
- Carrying the same time window forward might make sense. Think from debugging point. A particular time frame is being debugged across dashboards. And when they move thay loose that time window


- How are total and max connection errors different (Availability) ?
- Buffer Pool Pages by Type doesn't have dirty pages ?
- Storage capacity: Total rows, Total size is for what, this is not clear. And why would those panels be empty (checked last 24 hours)

- Queries (the top metrics panel shoes unable to load)
- Longest Wait of what (active queries), already run queries ? It shows NaN..

packages/mysql_otel/docs/README.md
```diff
@@ -1,8 +1,8 @@
-# MYSQL metrics for OpenTelemetry Collector
+# MySQL integration for OpenTelemetry Collector
```
ishleenk17:
```suggestion
# MySQL OpenTelemetry Assets
```

packages/mysql_otel/docs/README.md
```diff
@@ -1,8 +1,8 @@
...
 
-The MySQL metrics from MySQL OpenTelemetry receiver allow you to monitor [MySQL](https://www.mysql.com), an open-source Relational Database Management System (RDBMS) that enables users to store, manage, and retrieve structured data efficiently.
+This integration allows you to monitor [MySQL](https://www.mysql.com), an open-source Relational Database Management System (RDBMS) that enables users to store, manage, and retrieve structured data efficiently.
```
ishleenk17: We should not use the word integration in this. Instead use package. 
since this is a content package and not a full fledged integration, we refrain from using the word integration here.

packages/mysql_otel/docs/README.md
```diff
@@ -19,86 +19,141 @@ You can use our hosted Elasticsearch Service on Elastic Cloud, which is recommen
...
+```
+
+**MariaDB:**
+```sql
+GRANT REPLICA MONITOR ON *.* TO '<MYSQL_USER>'@'%';
+```
+
+### Configuration
+
+Install and configure the upstream OpenTelemetry Collector to export metrics to Elasticsearch. The configuration below uses separate receivers for primary and replica instances since replication metrics are only available on replicas.
```
ishleenk17: upstream OTEL collector or EDOT collector

packages/mysql_otel/docs/README.md
```diff
@@ -19,86 +19,141 @@ You can use our hosted Elasticsearch Service on Elastic Cloud, which is recommen
...
 
 ```yaml
 receivers:
-  mysql:
-    endpoint: localhost:3306
+  mysql/primary:
+    endpoint: <MYSQL_PRIMARY_ENDPOINT>
     username: <MYSQL_USER>
     password: <MYSQL_PASSWORD>
-    database: <your database name>
```
ishleenk17, line 30: This cnfig parameter is still supported, right ?

packages/mysql_otel/docs/README.md
```diff
@@ -19,86 +19,141 @@ You can use our hosted Elasticsearch Service on Elastic Cloud, which is recommen
...
     collection_interval: 10s
-    initial_delay: 1s
```
ishleenk17, line 32: Same with this

packages/mysql_otel/docs/README.md
```diff
@@ -19,86 +19,141 @@ You can use our hosted Elasticsearch Service on Elastic Cloud, which is recommen
...
+      mysql.query.count:
+        enabled: true
       mysql.query.client.count:
         enabled: true
-      mysql.client.network.io:
+      mysql.query.slow.count:
         enabled: true
       mysql.commands:
         enabled: true
-      mysql.max_used_connections:
```
ishleenk17, line 44: We are not using this metric in dashboard ?

packages/mysql_otel/docs/README.md
```diff
@@ -19,86 +19,141 @@ You can use our hosted Elasticsearch Service on Elastic Cloud, which is recommen
...
         enabled: true
       mysql.table_open_cache:
         enabled: true
-      mysql.replica.sql_delay:
+      mysql.joins:
+        enabled: true
+      mysql.page_size:
         enabled: true
+      mysql.mysqlx_worker_threads:
+        enabled: true
```
ishleenk17, line 113: Hoping all these are required in the dashboard. This example config should be the bare minimum required for the dashboard

packages/mysql_otel/docs/README.md
```diff
@@ -19,86 +19,141 @@ You can use our hosted Elasticsearch Service on Elastic Cloud, which is recommen
...
-```yaml
-mysql.replica.sql_delay:
-  enabled: true
-mysql.replica.time_behind_source:
-  enabled: true
-```
+### Metrics
 
-## Metrics reference
+Please refer to the [metadata.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.129.0/receiver/mysqlreceiver/metadata.yaml) of the OpenTelemetry MySQL receiver for details on available metrics.
```
ishleenk17: Please make the changes as per Vale linter suggestions

packages/mysql_otel/docs/README.md
```diff
@@ -19,86 +19,141 @@ You can use our hosted Elasticsearch Service on Elastic Cloud, which is recommen
...
+**MySQL:**
+```sql
+GRANT REPLICATION CLIENT ON *.* TO '<MYSQL_USER>'@'%';
+```
+
+**MariaDB:**
+```sql
+GRANT REPLICA MONITOR ON *.* TO '<MYSQL_USER>'@'%';
+```
+
```
ishleenk17, line 56: We don't have a compatibility section ? We should mention that we have tested this CP with which OTEL collector version.

packages/mysql_otel/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 3.5.0
 name: mysql_otel
 title: MySQL OpenTelemetry assets
```
ishleenk17:
```suggestion
title: MySQL OpenTelemetry Assets
```

packages/mysql_otel/docs/README.md
shmsr: We do have one; check the section just before this (doesn't show up in diff as we did not change it).

shmsr: Now it should as I bumped the version.

packages/mysql_otel/docs/README.md
shmsr: Yes. Without this, all DB metrics will be collected by default. So ideally it should not be there.

packages/mysql_otel/docs/README.md
shmsr: All extra configs; users can still do it if they want but not needed here.

packages/mysql_otel/docs/README.md
shmsr: Now it is absolute minimal.

packages/mysql_otel/docs/README.md
```diff
@@ -19,86 +19,135 @@ You can use our hosted Elasticsearch Service on Elastic Cloud, which is recommen
...
+**Replica instance (for replication metrics):**
+
+```sql
+-- mysql.replica.time_behind_source, mysql.replica.sql_delay
+-- Works on both MySQL and MariaDB (MariaDB 10.5+ remaps this to BINLOG MONITOR)
+GRANT REPLICATION CLIENT ON *.* TO '<MYSQL_USER>'@'%';
+```
+
+**Performance Schema consumers:**
```
shmsr, line 47: @ishleenk17 updated this list and tested; even this [README](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/mysqlreceiver/README.md#prerequisites) does not do a good job at listing the pre-reqs properly.

ishleenk17: Ideally any config related to the service itself should be updated in the receiver documentation rather than in the content pack. 
Please raise a ticket upstream to add these config options in the README as well.

ishleenk17 (approved):
Looks good!

PR #17208, opened 2026-02-03
[Enhancement][Synthetics] Add base64 encoding option for inline browser monitors

fkanout:
## Summary

Add support for `source.inline.encoding` option in browser monitors, enabling base64-encoded inline scripts. This allows inline monitor scripts to contain special characters that would otherwise break YAML format.

Related to: https://github.com/elastic/beats/pull/45100

## Changes

- Add `source.inline.encoding` variable to browser data stream manifest
- Add encoding field to browser handlebar template
- Bump package version to `1.5.0`
- Update Kibana version constraint to `^8.19.0 || ^9.1.0`

## Usage

When creating a browser monitor, the inline script can now be base64-encoded:
```json
{
  "source": {
    "inline": {
      "script": "c3RlcCgnR28gdG8gaG9tZXBhZ2UnLCBhc3luYyAoKSA9PiB7IC4uLiB9KQ==",
      "encoding": "base64"
    }
  }
}
```

### DIFF
--- a/packages/synthetics/changelog.yml
+++ b/packages/synthetics/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.5.0"
+  changes:
+    - description: Add base64 encoding option for inline browser monitors
+      type: enhancement
+      link: https://github.com/elastic/beats/pull/45100
 - version: "1.4.2"
   changes:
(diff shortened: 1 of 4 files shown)


shahzad31:
@fkanout i don't think heartbeat change exists in base 8.19.0

PR #17229, opened 2026-02-04
[Keycloak] Prevent data_stream fields from being overwritten by the log

SimonKoetting:
The ECS logs of Keycloak contain the data_stream.type field, which is set to logs.
As the pipeline is parsing the message as JSON and overwriting the root object, this leads to the data_stream object being overwritten in the ingest-pipeline so only data_stream.type is stored in the document in ES while dataset & namespace get lost.

This PR removes the data_stream object out of the json generated from the log before it overwrites the data_stream object already present in the event.

### DIFF
--- a/packages/keycloak/changelog.yml
+++ b/packages/keycloak/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Prevent data_stream fields from being overwritten by the log.
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/16685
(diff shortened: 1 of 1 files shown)


chrisberkhout:
@SimonKoetting Here's the more general fix I think we should do: https://github.com/elastic/integrations/pull/17430

SimonKoetting:
@chrisberkhout all right, fine for me :) I'll go ahead and close this PR. Thanks!

PR #17237, opened 2026-02-04
Fix excessive restarts rule to only monitor elastic-agent process

MichelLosier:
## Proposed commit message

Fixes the excessive restarts rule for elastic_agent package to only monitor the elastic-agent process, and hardcoding of default namespace in alerting rules.

This rescopes the excessive restarts rule to only focus on elastic-agent process restarts. Monitoring other elastic-agent processes could produce false signals since these can occur on config changes.

This also switches the rule to querying logs from the elastic-agent to avoid the existing cross-dependency with the `system` package:

```
FROM logs-elastic_agent*
| WHERE message like "Starting*elastic-agent*"
| STATS restart_count = count(*) BY host.name, bucket(@timestamp,5 minute)
| WHERE restart_count > 10
| STATS MAX(restart_count) BY host.name
```

Possible risk / questions here:
* The log message could change, or other new log messages could later be introduced with a similar pattern. For elastic agent maintainers: do we expect much risk with this?

An alternative solution using the `system` package:

```
FROM metrics-system*
| WHERE TO_LOWER(process.name) == "elastic-agent" AND agent.name NOT LIKE "*agentless*"
| STATS restart_count = COUNT_DISTINCT(process.cpu.start_time) BY host.name, bucket(@timestamp,5 minute)
| WHERE restart_count > 10
| STATS MAX(restart_count) BY host.name
```

This also addresses an issue where the `default` namespace was hardcoded into the index pattern of some alerting rules in this package.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes: https://github.com/elastic/ingest-dev/issues/6696
- Relates: https://github.com/elastic/integrations/issues/17311


packages/elastic_agent/kibana/alerting_rule_template/elastic-agent-excessive-restarts.json
```diff
@@ -16,7 +16,7 @@
       "thresholdComparator": ">",
       "size": 100,
       "esqlQuery": {
-        "esql": "FROM metrics-system*\n| WHERE TO_LOWER(process.executable) LIKE \"*elastic*agent*\" AND agent.name NOT LIKE \"*agentless*\"\n| STATS restart_count = COUNT_DISTINCT(process.cpu.start_time) BY host.name, process.name, bucket(@timestamp,5 minute) \n| WHERE restart_count > 10\n| STATS MAX(restart_count) BY host.name, process.name"
+        "esql": "FROM logs-elastic_agent*\n| WHERE message like \"Starting*elastic-agent*\"\n| STATS restart_count = count(*) BY host.name, bucket(@timestamp,5 minute)\n| WHERE restart_count > 10\n| STATS MAX(restart_count) BY host.name"
```
belimawr: What about restricting a bit more the index, so we don't include logs from the other components/beats?
```suggestion
        "esql": "FROM logs-elastic_agent-*\n| WHERE message like \"Starting*elastic-agent*\"\n| STATS restart_count = count(*) BY host.name, bucket(@timestamp,5 minute)\n| WHERE restart_count > 10\n| STATS MAX(restart_count) BY host.name"
```

This will prevent data streams like `logs-elastic_agent.filebeat-default`,`logs-elastic_agent.metricbeat-default`, `logs-elastic_agent.status_change-default`, etc to be searched.

AndersonQ:
> Regarding the possibility of the log message changing, I don't see it as an issue, but I'd like the @elastic/elastic-agent-data-plane opinion on it.

I think you meant @elastic/elastic-agent-control-plane.

While it seems fragile to rely on log messages, they don't change often. But perhaps we can add a comment on the agent to say this message is used here? Worst case we'd need to add a few OR if the log changes, so it should be fine.

I don't recall, but could we get this information from the `fleet-agents` index? I believe it'd be more reliable getting it from there.

swiatekm:
I'm not against relying on the log file for now, but we should definitely add a test on the agent side which verifies we emit exactly this line and links back to the package definition.

In the long run, it shouldn't be too difficult to have agent expose an uptime metric which would be much more reliable.

MichelLosier:
> I don't recall, but could we get this information from the fleet-agents index? I believe it'd be more reliable getting it from there.

We do have a `.fleet-agents` index which has docs for each agent reflecting their current state. Since it is a system index, it has restricted access constraints, and since its just a doc per agent we couldn't do a timeseries calculation of it.

> In the long run, it shouldn't be too difficult to have agent expose an uptime metric which would be much more reliable.

As discussed a little bit here: https://github.com/elastic/ingest-dev/issues/6696
I'd like to move some of the agent cpu and memory usage rule queries away from system package data streams to metrics that the elastic-agent can provide. I can create issues for those and this suggestion here.

packages/elastic_agent/kibana/alerting_rule_template/elastic-agent-excessive-restarts.json
MichelLosier: I've updated the query in https://github.com/elastic/integrations/pull/17237/commits/9d5c2a9addad7977714c0dda532d0f38486536ed

Sounds like using the log pattern is a fine for now

MichelLosier:
Just added in a separate small fix here: https://github.com/elastic/integrations/pull/17237/commits/12b0bb8b3385eb4ee33a6b995b2070c46d68c0fe where the `default` namespace was hardcoded in the index pattern for some rules

PR #17239, opened 2026-02-05
enable profilingmetrics dataset based auto installation

rogercoll:
## Proposed commit message

Enable profilingmetrics-otel dataset based auto-installation

## Checklist


packages/profilingmetrics_otel/manifest.yml
```diff
@@ -17,6 +17,8 @@ conditions:
     subscription: "basic"
 discovery:
   fields: []
+  datasets:
+    - name: profilingmetricsconnector.otel
```
rogercoll, line 21: This will be the default dataset name starting on `9.3.1`

florianl: kibana/dashboard/profilingmetrics_otel*json still uses `"metrics-*"` for the referenced `index`. Shouldn't this be aligned?

rogercoll: That's true, but I would suggest narrowing down the index when the writes to the new dataset are more adopted, `9.3.1` was just recently released and this would break most deployments. I will merge this for now and open a follow-up issue for the new filtering.

Issue #17030, opened 2026-01-21
[Azure] Add client secret (Oauth2) support


kaiyan-sheng:
We need add the new config parameters introduced in https://github.com/elastic/beats/pull/47256 into the related azure integrations.

This enhances Azure Event Hub input plugin for Elastic Agent with RBAC authorization (OAuth2) due to security requirements. Previously we only support shared access key (with connection string) for authentication.

The implementation added a new config parameter called `auth_type` for users to specify authentication method:
When `auth_type` is set to `connection_string`, or leave it blank: `connection_string` is required.  
When `auth_type` is set to `client_secret`, `oauth2` is used.

Note: We do expect users to use the same auth type for both eventhub and storage account.

OAuth2 specific Configuration Parameters (auth_type=client_secret):

- eventhub_namespace: Fully qualified namespace (e.g., namespace.servicebus.windows.net)
- tenant_id: Azure AD tenant ID
- client_id: Azure AD application (client) ID
- client_secret: Azure AD application client secret
- authority_host: Azure AD authority host (optional, defaults to Azure Public Cloud). `https://login.microsoftonline.com` is the default.

a03nikki:
@kaiyan-sheng : How does this ticket differ from #16586?

zmoog:
@kaiyan-sheng @bturquet, check out the growth for azure-eventhub.

<img width="784" height="762" alt="Image" src="https://github.com/user-attachments/assets/0ce77302-00ff-45e2-87c7-eee883524a90" />

What if we group the required PRs by team? It’ll help us get the Azure Logs update out the door faster and streamline how we coordinate the work. We could use this as the tracking issue for the credentials rollout.

kaiyan-sheng:
@zmoog I should have grouped the required PRs by team 😂 Now the current [PR](https://github.com/elastic/integrations/pull/17245) is so big and Im missing changes here and there.

I will move this task to the next iteration(it-125) to keep addressing all the comments and get reviews.

PR #17245, opened 2026-02-05
Add Client Secret support into azure event hub integrations

kaiyan-sheng:
## Proposed commit message

Add OAuth2 client secret authentication support for Azure Event Hub integrations
This enhancement adds OAuth2 (client_secret) authentication support to all Azure integrations that use Event Hub, following the implementation in https://github.com/elastic/beats/pull/47256.

Changes include:
- Added auth_type parameter to select between connection_string (default) and client_secret authentication
- Added OAuth2 parameters: eventhub_namespace, tenant_id, client_id, client_secret, authority_host
- Made connection_string and storage_account_key optional (conditionally required based on auth_type)
- Updated all affected template files to output OAuth2 parameters
- Version bumped 12 packages with changelog entries

## Checklist

## Author's Checklist

Affected 12 packages:

- [ ] azure
- [ ] azure_logs
- [ ] azure_openai
- [ ] azure_functions
- [ ] azure_frontdoor
- [ ] azure_app_service
- [ ] azure_ai_foundry
- [ ] microsoft_sentinel
- [ ] microsoft_defender_cloud
- [ ] m365_defender
- [ ] kubernetes
- [ ] github

## Related issues

- Closes https://github.com/elastic/integrations/issues/17030

## Screenshots

Screenshot 2026-02-06 at 1 04 34 AM

Screenshot 2026-02-10 at 9 57 45 AM

### DIFF
--- a/packages/azure/manifest.yml
+++ b/packages/azure/manifest.yml
@@ -45,7 +45,7 @@ vars:
     type: select
     title: Authentication Type
     multi: false
-    required: false
+    required: true
     show_user: true
     default: connection_string
     options:
--- a/packages/azure_frontdoor/manifest.yml
+++ b/packages/azure_frontdoor/manifest.yml
@@ -51,7 +51,7 @@ policy_templates:
             type: select
             title: Authentication Type
             multi: false
-            required: false
+            required: true
             show_user: true
             default: connection_string
             options:
--- a/packages/github/data_stream/audit/manifest.yml
+++ b/packages/github/data_stream/audit/manifest.yml
@@ -128,7 +128,7 @@ streams:
         type: select
         title: Authentication Type
         multi: false
-        required: false
+        required: true
         show_user: true
         default: connection_string
         options:
--- a/packages/azure_ai_foundry/data_stream/logs/manifest.yml
+++ b/packages/azure_ai_foundry/data_stream/logs/manifest.yml
@@ -25,7 +25,7 @@ streams:
         type: select
         title: Authentication Type
         multi: false
-        required: false
+        required: true
         show_user: true
         default: connection_string
         options:
--- a/packages/azure_app_service/manifest.yml
+++ b/packages/azure_app_service/manifest.yml
@@ -37,7 +37,7 @@ vars:
     type: select
     title: Authentication Type
     multi: false
-    required: false
+    required: true
     show_user: true
     default: connection_string
     options:
--- a/packages/azure_functions/data_stream/functionapplogs/manifest.yml
+++ b/packages/azure_functions/data_stream/functionapplogs/manifest.yml
@@ -26,7 +26,7 @@ streams:
         type: select
         title: Authentication Type
         multi: false
-        required: false
+        required: true
         show_user: true
         default: connection_string
         options:
--- a/packages/azure_logs/manifest.yml
+++ b/packages/azure_logs/manifest.yml
@@ -46,7 +46,7 @@ policy_templates:
         type: select
         title: Authentication Type
         multi: false
-        required: false
+        required: true
         show_user: true
         default: connection_string
         options:
--- a/packages/azure_openai/data_stream/logs/manifest.yml
+++ b/packages/azure_openai/data_stream/logs/manifest.yml
@@ -25,7 +25,7 @@ streams:
         type: select
         title: Authentication Type
         multi: false
-        required: false
+        required: true
         show_user: true
         default: connection_string
         options:
--- a/packages/kubernetes/data_stream/audit_logs/manifest.yml
+++ b/packages/kubernetes/data_stream/audit_logs/manifest.yml
@@ -216,7 +216,7 @@ streams:
         type: select
         title: Authentication Type
         multi: false
-        required: false
+        required: true
         show_user: true
         default: connection_string
         options:
--- a/packages/m365_defender/data_stream/event/manifest.yml
+++ b/packages/m365_defender/data_stream/event/manifest.yml
@@ -28,7 +28,7 @@ streams:
         type: select
         title: Authentication Type
         multi: false
-        required: false
+        required: true
         show_user: true
         default: connection_string
         options:
--- a/packages/microsoft_defender_cloud/data_stream/event/manifest.yml
+++ b/packages/microsoft_defender_cloud/data_stream/event/manifest.yml
@@ -27,7 +27,7 @@ streams:
         type: select
         title: Authentication Type
         multi: false
-        required: false
+        required: true
         show_user: true
         default: connection_string
         options:
--- a/packages/microsoft_sentinel/data_stream/event/manifest.yml
+++ b/packages/microsoft_sentinel/data_stream/event/manifest.yml
@@ -28,7 +28,7 @@ streams:
         type: select
         title: Authentication Type
         multi: false
-        required: false
+        required: true
         show_user: true
         default: connection_string
         options:


packages/azure/manifest.yml
```diff
@@ -41,15 +41,38 @@ vars:
...
+    required: false
+    show_user: true
+    default: connection_string
+    options:
+      - text: Connection String
+        value: connection_string
+      - text: Client Secret
+        value: client_secret
+    description: >-
+      Authentication method to use for Event Hub and Storage Account. When set to 'connection_string' or left blank: connection_string and storage_account_key are required. When set to 'client_secret': OAuth2 authentication with Azure AD is used, requiring tenant_id, client_id, client_secret, and eventhub_namespace. Note: The same authentication type applies to both Event Hub and Storage Account for security consistency.
```
zmoog: nit: since we're using `>-`, we can break the description in multiple lines fore readability (for editors w/o line wraps, like GitHub), and the result will still be treated as a single-line string.

packages/azure/manifest.yml
```diff
@@ -58,15 +81,48 @@ vars:
...
+    description: >-
+      (Required when auth_type is 'client_secret') Fully qualified Event Hub namespace (e.g., namespace.servicebus.windows.net). Do not use the short namespace name; use the complete FQDN.
+  - name: tenant_id
+    type: text
+    title: Tenant ID
+    multi: false
+    required: false
+    show_user: true
+    description: >-
+      (Required when auth_type is 'client_secret') Azure AD (Microsoft Entra ID) tenant ID. This is the directory/tenant where your Azure AD application is registered.
```
zmoog: Is Azure still using Azure AD somewhere? I guess they moved almost everything to Entra ID now.

nit: in a multi-tenant scenario, the tenant ID must be the Azure tenant where the resorces (event hub, storage, etc) exist. This is correct in the main use, so let's keep this and I'll revisit when I'll update the docs to share how to set up a multi-tenant scenario.

packages/azure/manifest.yml
```diff
@@ -58,15 +81,48 @@ vars:
...
+  - name: client_id
+    type: text
+    title: Client ID
+    multi: false
+    required: false
+    show_user: true
+    description: >-
+      (Required when auth_type is 'client_secret') Azure AD application (client) ID. The service principal must have 'Azure Event Hubs Data Receiver' role on the Event Hub and 'Storage Blob Data Contributor' role on the Storage Account.
```
zmoog: Should we also update the docs with the new authentication types and roles required for client secret?

packages/azure/manifest.yml
```diff
@@ -58,15 +81,48 @@ vars:
...
     show_user: true
     description: >-
-      The storage account key, this key will be used to authorize access to data in your storage account.
+      (Required when auth_type is 'client_secret') Azure AD application client secret. Generate this secret in your Azure AD app registration.
+  - name: authority_host
+    type: text
+    title: Authority Host
+    multi: false
+    required: false
+    show_user: false
```
zmoog, line 131: I agree it makes senst to keep this in the 'advanced' section.

kaiyan-sheng:
> I wonder if we should rebrand this as "Client Secret support" instead. This would align user expectations with what the feature actually provides.

@zmoog Good point!! I will change to just `client secret` instead!! Thanks for the review!!

packages/azure/manifest.yml
kaiyan-sheng: OK sounds good! I will change it to Microsoft Entra ID instead. Thanks!

packages/azure/manifest.yml
kaiyan-sheng: Completely forgot about the docs 😂 Thanks!

muthu-mps:
@kaiyan-sheng - What would be the behaviour of the configurations when a user upgrades to this version of the integration with the default options? Will the upgrade happen seamlessly without making any updates to the configurations?

packages/azure/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: "1.36.0"
+  changes:
+    - description: Add OAuth2 client secret authentication support for Azure Event Hub with RBAC.
```
zmoog:
```suggestion
    - description: Add client secret authentication support for Azure Event Hub with RBAC.
```

packages/azure/manifest.yml
```diff
@@ -41,15 +41,43 @@ vars:
...
+    default: connection_string
+    options:
+      - text: Connection String
+        value: connection_string
+      - text: Client Secret
+        value: client_secret
+    description: >-
+      Authentication method to use for Event Hub and Storage Account.
+      When set to 'connection_string' or left blank: connection_string and storage_account_key are required.
+      When set to 'client_secret': OAuth2 authentication with Microsoft Entra ID is used, requiring tenant_id, client_id, client_secret, and eventhub_namespace.
```
zmoog:
```suggestion
      When set to 'client_secret': the system authenticates with Microsoft Entra ID using the provided credentials, requiring tenant_id, client_id, client_secret, and eventhub_namespace.
```

packages/azure/manifest.yml
```diff
@@ -41,15 +41,43 @@ vars:
...
+    type: password
+    secret: true
+    title: Storage Account Key
+    multi: false
+    required: false
     show_user: true
     description: >-
-      The connection string required to communicate with Event Hubs. See [Get an Event Hubs connection string](https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-get-connection-string) to learn more.
+      (Required when auth_type is 'connection_string' or left blank) The storage account key used to authorize access to data in your storage account.
+      Not used when auth_type is 'client_secret'; OAuth2 is used for storage authentication instead.
```
zmoog:
```suggestion
      Not used when auth_type is 'client_secret'; client secret is used for storage authentication instead.
```

packages/azure/manifest.yml
```diff
@@ -58,15 +86,53 @@ vars:
...
+      Generate this secret in your Microsoft Entra ID app registration.
+  - name: authority_host
+    type: text
+    title: Authority Host
+    multi: false
+    required: false
+    show_user: false
+    default: https://login.microsoftonline.com
+    description: >-
+      (Optional, for OAuth2) Microsoft Entra ID authority endpoint. Defaults to https://login.microsoftonline.com (Azure Public Cloud).
```
zmoog:
```suggestion
      (Optional, for Client Secret) Microsoft Entra ID authority endpoint. Defaults to https://login.microsoftonline.com (Azure Public Cloud).
```

packages/azure_frontdoor/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.3.0"
+  changes:
+    - description: Add OAuth2 client secret authentication support for Azure Event Hub with RBAC.
```
zmoog:
```suggestion
    - description: Add client secret authentication support for Azure Event Hub with RBAC.
```

packages/azure_frontdoor/manifest.yml
```diff
@@ -47,26 +47,98 @@ policy_templates:
...
+            default: connection_string
+            options:
+              - text: Connection String
+                value: connection_string
+              - text: Client Secret
+                value: client_secret
+            description: >-
+              Authentication method to use for Event Hub and Storage Account.
+              When set to 'connection_string' or left blank: connection_string and storage_account_key are required.
+              When set to 'client_secret': OAuth2 authentication with Microsoft Entra ID is used, requiring tenant_id, client_id, client_secret, and eventhub_namespace.
```
zmoog:
```suggestion
              When set to 'client_secret': the system authenticates with Microsoft Entra ID using the provided credentials, requiring tenant_id, client_id, client_secret, and eventhub_namespace.
```

packages/azure_frontdoor/manifest.yml
```diff
@@ -47,26 +47,98 @@ policy_templates:
...
+          - name: storage_account_key
+            type: password
             secret: true
+            title: Storage Account Key
+            multi: false
+            required: false
+            show_user: true
+            description: >-
+              (Required when auth_type is 'connection_string' or left blank) The storage account key used to authorize access to data in your storage account.
+              Not used when auth_type is 'client_secret'; OAuth2 is used for storage authentication instead.
```
zmoog:
```suggestion
              Not used when auth_type is 'client_secret'; client secret is used for storage authentication instead.
```

packages/azure_frontdoor/manifest.yml
```diff
@@ -47,26 +47,98 @@ policy_templates:
...
+              Generate this secret in your Microsoft Entra ID app registration.
+          - name: authority_host
+            type: text
+            title: Authority Host
+            multi: false
+            required: false
+            show_user: false
+            default: https://login.microsoftonline.com
+            description: >-
+              (Optional, for OAuth2) Microsoft Entra ID authority endpoint. Defaults to https://login.microsoftonline.com (Azure Public Cloud).
```
zmoog:
```suggestion
              (Optional, for client secret) Microsoft Entra ID authority endpoint. Defaults to https://login.microsoftonline.com (Azure Public Cloud).
```

packages/github/data_stream/audit/manifest.yml
```diff
@@ -442,29 +508,29 @@ streams:
       - name: oauth2
         required: true
         show_user: true
-        title: Collect logs using OAuth2 authentication
-        description: To collect logs using OAuth2 authentication enable the toggle switch. By default, it will collect logs using service account key or URI.
+        title: Collect logs using client secret authentication
```
zmoog: @kaiyan-sheng, I think this is a legit generic OAuth2 client in the blob storage input.

packages/github/changelog.yml
```diff
@@ -51,7 +56,7 @@
       link: https://github.com/elastic/integrations/pull/15498
 - version: "2.15.1"
   changes:
-    - description: Added missing oauth2 toggle in audit data stream.
+    - description: Added missing client secret authentication toggle in audit data stream.
```
zmoog: @kaiyan-sheng, this also seems a legic oauth2 thing, isn't it?

zmoog:
> @kaiyan-sheng - What would be the behaviour of the configurations when a user upgrades to this version of the integration with the default options? Will the upgrade happen seamlessly without making any updates to the configurations?

@kaiyan-sheng, correct me if I’m wrong, but all existing deployments should continue to work as-is.

Since auth_type defaults to "connection string," existing variables should retain their values even if they aren't required anymore.

Users will need to be careful with new deployments, though, as we can't rely on UI checks for now. Unfortunately, the new variables group won't land until 9.4.

@muthu-mps, is there anything we can do here to signal which field is required scenarios like this one? What's the pattern and best practices we're applying in the integration ecosystem?

packages/azure_frontdoor/manifest.yml
efd6, line 1: Is there a reason not to add similar README documentation to this package? Similar below.

packages/github/data_stream/audit/manifest.yml
kaiyan-sheng: Wrong ctrl+f 😂 Thanks! Will fix it!

packages/azure_frontdoor/manifest.yml
kaiyan-sheng: Ahh I missed it, just added the change. Thanks!

kaiyan-sheng:
> What would be the behaviour of the configurations when a user upgrades to this version of the integration with the default options? Will the upgrade happen seamlessly without making any updates to the configurations?

@muthu-mps Yes. The upgrade is backward compatible and should work without any config changes. `auth_type` is optional and defaults to `connection_string` so if it's not set, it's treated as `connection_string`, which matches the previous behavior.

packages/azure_frontdoor/manifest.yml
efd6: Are the packages also a target for this? (the github and microsoft* packages).

kaiyan-sheng: Good point! Let me add a part in the readme to mention the authentication types and then refer back to the main azure logs integration and filebeat doc.

zmoog (approved):
Thanks for working on this @kaiyan-sheng, it LGTM overall, no blockers.

IMO there's one aspect to consider. 

In variable descriptions we use the internal field names, like `auth_type` or `connection_string`.

For example:

Image

I have no problem in matching the UI elements to the internal name, but I see my brain performing a lookup somewhere for a split second. 

I and think our users wouldn't find this complicated as well, but maybe we can remove this tiny friction without losing clarity.

What about (1) using actual field names that match the UI and (2) use bold/italics to field highlight dependencies?

Image

Is a subtle difference, but IMO it can smooth over this friction until we get proper var group dependencies in 9.4.

packages/azure/manifest.yml
```diff
@@ -41,15 +41,43 @@ vars:
     required: true
     show_user: true
     default: $Default
+  - name: auth_type
+    type: select
+    title: Authentication Type
+    multi: false
+    required: false
```
zmoog: Would making auto_type required cause any issues?

I know it’s a required field in Filebeat, but we can ensure it always gets a value via the policy template, right?

ishleenk17:
@claude pls review the PR

packages/azure/manifest.yml
kaiyan-sheng: Yeah it's set to `default: connection_string` by default in the manifest file. But I was trying to avoid any potential issue from upgrade. Let me test it with making this config parameter required!

kaiyan-sheng:
> What about (1) using actual field names that match the UI and (2) use bold/italics to field highlight dependencies?

@zmoog Good point! I think this is definitely worth improving. I'm changing them to use the actual parameter names in the UI instead of names in Filebeat!

But for the [README](https://www.elastic.co/docs/reference/integrations/azure) files I see we are using param names from Filebeat:
Screenshot 2026-02-22 at 1 12 44 AM

Do you think it's worth changing too?

packages/azure/manifest.yml
kaiyan-sheng: Screenshot 2026-02-22 at 10 53 38 PM

I changed `auty_type` to required parameter and tried the upgrade path, it's automatically set to `Connection String` since that's the default.

zmoog:
> Do you think it's worth changing too?

Yeah, makes sense. 

We should probably align this section with what we have at https://www.elastic.co/docs/reference/integrations/azure/events, adding the new options in the "Event Hub Processor v2 only" seciton.

kaiyan-sheng:
> adding the new options in the "Event Hub Processor v2 only" seciton.

@zmoog and I talked about this in slack, the decision is to keep the new options with authentication section so all the params around auth are together. 

Also I see _some_ benefits of keeping the config parameters in documentation instead of the UI names when we are trying to debug using config files instead of UI. So I think we can keep it as is for now (tbh maybe we should document both the UI name and the param name).

PR #17263, opened 2026-02-05
[Kafka Connect] New Integration

philippkahr:
## Proposed commit message

Customer approached me with the following handcrafted 
[metricbeat-kafkaconnect-module.yml](https://github.com/user-attachments/files/25102235/metricbeat-kafkaconnect-module.yml) that uses jolokia to expose Kafka Connect metrics. Since we didn't have anything apart from then generic Jolokia Input.

I decided to turn this metricbeat.yml into an elastic package.

The package contains tests with docker-compose like jolokia, spinning up a full kafka, creating a sink/source, sending data in. The docker part was created using LLM.

The Readme also was created by LLM.

Rest, mapping, pipelines, example docs, running it locally, creating the dashboard by hand.

## Checklist

It is all part of the `elastic-package test` suite and should bring up all the containers.

## Screenshots
Screenshot_5-2-2026_17125_localhost

### DIFF
--- a/packages/kafka_connect/data_stream/client/fields/ecs.yml
+++ b/packages/kafka_connect/data_stream/client/fields/ecs.yml
@@ -4,3 +4,12 @@
 - external: ecs
   name: agent.id
   dimension: true
+- external: ecs
+  name: container.id
+  dimension: true
+- external: ecs
+  name: service.address
+  dimension: true
+- external: ecs
+  name: cloud.instance.id
+  dimension: true
--- a/packages/kafka_connect/data_stream/connector/fields/ecs.yml
+++ b/packages/kafka_connect/data_stream/connector/fields/ecs.yml
@@ -4,3 +4,12 @@
 - external: ecs
   name: agent.id
   dimension: true
+- external: ecs
+  name: container.id
+  dimension: true
+- external: ecs
+  name: service.address
+  dimension: true
+- external: ecs
+  name: cloud.instance.id
+  dimension: true
--- a/packages/kafka_connect/data_stream/task/fields/ecs.yml
+++ b/packages/kafka_connect/data_stream/task/fields/ecs.yml
@@ -4,3 +4,12 @@
 - external: ecs
   name: agent.id
   dimension: true
+- external: ecs
+  name: container.id
+  dimension: true
+- external: ecs
+  name: service.address
+  dimension: true
+- external: ecs
+  name: cloud.instance.id
+  dimension: true
--- a/packages/kafka_connect/data_stream/worker/fields/ecs.yml
+++ b/packages/kafka_connect/data_stream/worker/fields/ecs.yml
@@ -4,3 +4,12 @@
 - external: ecs
   name: agent.id
   dimension: true
+- external: ecs
+  name: container.id
+  dimension: true
+- external: ecs
+  name: service.address
+  dimension: true
+- external: ecs
+  name: cloud.instance.id
+  dimension: true


philippkahr:
hey @elastic/obs-infraobs-integrations since you own the Kafka integration as well, I think this one fits nicely to it.

agithomas:
Thank you, @philippkahr, for the contribution. 

Kindly validate if certain fields are missing to be tagged as dimension fields. We have three cases [here](https://www.elastic.co/docs/extend/integrations/developer-tsds-guidelines#_ecs_fields) depending on the nature of the deployment. Inadequate tagging may lead to loss of data.

agithomas:
You may consider following the [the naming conventions](https://www.elastic.co/docs/extend/beats/event-conventions#field-names), especially [unit suffix](https://www.elastic.co/docs/extend/beats/event-conventions#units) and [standardised names](https://www.elastic.co/docs/extend/beats/event-conventions#units)

philippkahr:
>Kindly validate if certain fields are missing to be tagged as dimension fields. We have three cases [here](https://www.elastic.co/docs/extend/integrations/developer-tsds-guidelines#_ecs_fields) depending on the nature of the deployment. Inadequate tagging may lead to loss of data.

Did that, added the `agent.id`.

>You may consider following the [the naming conventions](https://www.elastic.co/docs/extend/beats/event-conventions#field-names), especially [unit suffix](https://www.elastic.co/docs/extend/beats/event-conventions#units) and [standardised names](https://www.elastic.co/docs/extend/beats/event-conventions#units)

Followed them as good as I can.

> GitHub Copilot comments.

Fixed them, when I find them necessary.

packages/kafka_connect/_dev/deploy/docker/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM confluentinc/cp-kafka-connect:7.5.0
```
stefans-elastic, line 1: In the README it is claimed that the integration supports Kafka Connect 2.8.x and 3.x versions. It would be better to introduce test variants and test against each supported version.
Also I'm not sure about using community images. Could they be different compared to official ones?)

stefans-elastic (commented):
Just 1 comment and one note regarding thee dashboard: do we want to include "description" section? Like in this example: 

Image

packages/kafka_connect/_dev/deploy/docker/Dockerfile
philippkahr: hmm, I could add variants, that seems like quite a bit of work though.

The benefit of the Confluent images is that it is all baked in and I need way less docker setup, like zookeeper, kafka itself and so on.

I've looked around but I can only find the JMX description here https://docs.confluent.io/platform/current/connect/monitoring.html

philippkahr:
> Just 1 comment and one note regarding thee dashboard: do we want to include "description" section? Like in this example:

@stefans-elastic , I have titles for all the visualisations, I don't think a description is needed.

packages/kafka_connect/_dev/deploy/docker/Dockerfile
stefans-elastic: > hmm, I could add variants, that seems like quite a bit of work though.

I'm not saying that it has to be done in this PR but I would prefer this work be at least planned for the future (maybe an issue opened for it). This will come in handy when there's a new version of `Kafka Connect` - we would simply add a new variant for it and make sure everything works as expected with the new version.

> The benefit of the Confluent images is that it is all baked in and I need way less docker setup, like zookeeper, kafka itself and so on.

I guess I'm just too paranoid about possibility of some compatibility quirks  between community version of the images and "official" Kafka/Kafka Connect

philippkahr:
@agithomas what do I need to do to get it over the finish line?

PR #17270, opened 2026-02-05
[HAproxy] Expose condition in UI

ckauf:
## Proposed commit message
Expose `Condition` in UI to allow dynamically enabling/disabling data collection.

## Checklist

Add a valid condition to the integration and check if data collection is/is not enabled, e.g. `${host.platform} == "windows"` on a linux host should not enable the input.

## Screenshots
image

### DIFF
--- a/packages/memcached/changelog.yml
+++ b/packages/memcached/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.8.0"
+  changes:
+    - description: Expose condition for metrics
+      type: enhancement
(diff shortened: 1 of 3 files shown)


packages/haproxy/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.18.0"
+  changes:
+    - description: Expose condition option in UI for metrics
+      type: bugfix
```
ishleenk17:
```suggestion
      type: enhancement
```

PR #17307, opened 2026-02-09
[nginx_otel] add slo for connection handling availability

jakubgalecki0:
## Proposed commit message

Add SLO template for connection handling availability

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- 

## Screenshots

image

image

image


ishleenk17:
@jakubgalecki0 : Pls see below review comments

1. The name of the SLO should be [ NGINX OTEL] as discussed in last meeting
2. The SLI name need not have OTEL mentioned: NGINX Connection Handling Availability 99.9% Rolling 30 Days
3. The calculation of drops in SLO and dashboard coincides ? To me it looks different. 
         **Dashboard approach**: max(accepted) - max(handled) gives the total cumulative gap between the two counters — drops          since NGINX started
         **SLI approach**: Delta_Accepted- Delta_Handled gives drops within each window

4. Grouping by host.name is not the default one here ?

jakubgalecki0:
@ishleenk17 there is a difference between calculation of the values but its due to the limitations of SLO's. The slo type used here is timeslice so it calculates how many times the constrains were violated in each timeslice window. We can't use the same exact calculations in SLO that are used in dashboard as they are not available. Thus we went for this delta approach.

Issue #15856, opened 2025-11-04
[CrowdStrike] Handle deprecating Threat fields


navnit-elastic:
The crowdstrike.falcon integration is using `TacticId`, `TacticIds`, `Tactic`, `Tactics`, `TechniqueId`, `TechniqueIds` `Technique` and `Techniques` fields to populate `threat.*` ECS fields for various detection events.

According to CrowdStirke documentation these fields are deprecated and these fields are moved under `MitreAttack` object.

Though the deprecating fields are still show up in the live logs, It would be good to switch to the recommended fields.

Deprecating note and schema for `MitreAttack` object:
<img width="1031" height="717" alt="Image" src="https://github.com/user-attachments/assets/7101788f-82d6-4ddc-a15f-1f8182149a69" />

Issue #10196, opened 2024-06-19
crowdstrike.alert: Integration does not map CrowdStrike tactics & techniques properly


NateUT99:
While reviewing the crowdstrike.alert dataset, I noticed that there seems to be an issue with the threat.tactic.id and threat.technique.id values when the alert references what CrowdStrike calls the "MITRE-Based Falcon Detections Framework". I've highlighted the issue in the attached screenshot; as you can see the tactic_id and technique_id from the original alert is "CSTA0004" and "CST0007/CST0008" respectively. The value for threat.tactic.name and threat.technique.name are correct; however it appears that the threat.tactic.id and threat.technique.id have dropped the "CS" prefixes completely.

My recommnedation would be either 1) do not write values to threat.tactic.* and threat.technique.* when the alert ids are prefixed with "CS", or 2) write an appropriate name to threat.framework for the "Falcon Detections Framework", and ensure that the prefixes remain as expected.

I do have some additional documentation on the Falcon Detections Framework if that would be useful. As it requires authentication to download, I would prefer to not attach it to a public GitHub issue. Please feel free to contact me if this would be beneficial.

<img width="2020" alt="crowdstrike_alert_threats" src="https://github.com/elastic/integrations/assets/48816468/c50450fb-d883-4ed5-8aa4-dd951216eb6c">

jamiehynds:
Thanks for reporting @NateUT99 

@piyush-elastic could someone from the team look into this one please, as it related to the alert pipeline added recently. https://github.com/elastic/integrations/blob/main/packages/crowdstrike/data_stream/alert/elasticsearch/ingest_pipeline/default.yml

piyushw-crest:
sure @jamiehynds , will check and keep posted on issue in case if any need help.

efd6:
This appears to be intentional:
- [threat.tactic.id](https://github.com/elastic/integrations/blob/ddea2d0936319bd79c48a30e030b37c43658c0d6/packages/crowdstrike/data_stream/alert/elasticsearch/ingest_pipeline/default.yml#L1182-L1196)
- [threat.technique.id](https://github.com/elastic/integrations/blob/ddea2d0936319bd79c48a30e030b37c43658c0d6/packages/crowdstrike/data_stream/alert/elasticsearch/ingest_pipeline/default.yml#L1213-L1227)

piyushw-crest:
[...]

PR #17322, opened 2026-02-10
crowdstrike: update threat.* field mappings in falcon data stream

navnit-elastic:
## Proposed commit message

```
crowdstrike: update threat.* field mappings in falcon data stream

Retain deprecated threat field mappings for backward compatibility.
Map fields under MitreAttack to standard ECS threat fields.

The fields under MitreAttack object are verified against live instance.
```

Here's the PR for a similar change in alert data stream: https://github.com/elastic/integrations/pull/17327
## Checklist

## Author's Checklist

- [ ] 

Pipeline tests:

```
--- Test results for package: crowdstrike - START ---
╭─────────────┬─────────────┬───────────┬──────────────────────────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE     │ DATA STREAM │ TEST TYPE │ TEST NAME                                                                    │ RESULT │ TIME ELAPSED │
├─────────────┼─────────────┼───────────┼──────────────────────────────────────────────────────────────────────────────┼────────┼──────────────┤
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-event-stream.log)                             │ PASS   │ 375.481833ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-audit-events.log)                      │ PASS   │  417.81225ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-auth-activity.log)                     │ PASS   │ 510.137209ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-cspmioa-streaming.log)                 │ PASS   │   383.5795ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-cspmsearch-streaming.log)              │ PASS   │ 433.183458ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-data-protection-detection-summary.log) │ PASS   │   410.0305ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-detection-summary.log)                 │ PASS   │ 458.835833ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-epp-detection-summary.log)             │ PASS   │ 477.506458ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-events.log)                            │ PASS   │ 400.699666ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-firewall.log)                          │ PASS   │ 672.830833ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-identity-protection-incident.log)      │ PASS   │ 410.312959ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-incident-summary.log)                  │ PASS   │ 444.751208ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-ipd-summary.log)                       │ PASS   │ 585.571542ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-mobile-detection-summary.log)          │ PASS   │  402.79525ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-recon-notification.log)                │ PASS   │ 404.135167ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-remote-response.log)                   │ PASS   │ 587.100292ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-sample.log)                            │ PASS   │ 399.907209ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-tags-list.log)                         │ PASS   │ 473.417708ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-tags.log)                              │ PASS   │ 491.358083ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-user-activity.log)                     │ PASS   │ 528.164709ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-xdr-detection-summary.log)             │ PASS   │ 402.458916ms │
│ crowdstrike │ falcon      │ pipeline  │ test-event-stream.log                                                        │ PASS   │ 112.466708ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-audit-events.log                                                 │ PASS   │ 118.848833ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-auth-activity.log                                                │ PASS   │  98.534625ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-cspmioa-streaming.log                                            │ PASS   │  80.522625ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-cspmsearch-streaming.log                                         │ PASS   │  76.391708ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-data-protection-detection-summary.log                            │ PASS   │     92.186ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-detection-summary.log                                            │ PASS   │  85.769292ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-epp-detection-summary.log                                        │ PASS   │ 223.977083ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-events.log                                                       │ PASS   │     76.438ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-firewall.log                                                     │ PASS   │  95.732791ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-identity-protection-incident.log                                 │ PASS   │    74.7775ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-incident-summary.log                                             │ PASS   │  78.907209ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-ipd-summary.log                                                  │ PASS   │  85.259833ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-mobile-detection-summary.log                                     │ PASS   │  72.691666ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-recon-notification.log                                           │ PASS   │  70.804333ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-remote-response.log                                              │ PASS   │  96.199209ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-sample.log                                                       │ PASS   │ 106.817458ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-tags-list.log                                                    │ PASS   │  67.728834ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-tags.log                                                         │ PASS   │  87.778542ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-user-activity.log                                                │ PASS   │  76.372041ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-xdr-detection-summary.log                                        │ PASS   │  74.539917ms │
╰─────────────┴─────────────┴───────────┴──────────────────────────────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: crowdstrike - END   ---
Done
```

## Related issues

- Relates #15856


packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-falcon-detection-summary.log-expected.json
```diff
@@ -169,19 +169,6 @@
...
-                    "name": [
-                        "Malware"
-                    ]
-                },
-                "technique": {
-                    "name": [
-                        "PUP"
-                    ]
-                }
-            },
```
kcreddy, line 184: Can you update each of these sample logs with new field, so that these ECS fields remain populated?

packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/detection_summary.yml
```diff
@@ -174,42 +174,6 @@ processors:
...
       if: ctx.host?.mac != null
-  - set: 
-      field: threat.framework
-      value: CrowdStrike Falcon Detections Framework
-      tag: set_threat_framework_cs
-      description: For Crowdstrike Falcon Detection tactics and/or that are prefixed with "CS"
-      if: >-
-        (ctx.crowdstrike?.event?.TacticId != null && ctx.crowdstrike.event.TacticId.startsWith("CS"))
-        || (ctx.crowdstrike?.event?.TechniqueId != null && ctx.crowdstrike.event.TechniqueId.startsWith("CS"))
-        || (ctx.crowdstrike?.event?.Tactic != null && ['malware','exploit','post-exploit','machine learning','custom intelligence','falcon overwatch','falcon intel','ai powered ioa','insecure security posture'].contains(ctx.crowdstrike.event.Tactic.toLowerCase()))
```
kcreddy, line 186: This handling is missed in `default.yml`. So in updated code there is no `CrowdStrike Falcon Detections Framework`

navnit-elastic: The updated code sets `threat.framework: MITRE ATT&CK`, regardless of the values in `TacticId` and `TechniqueId` fields. Shouldn't we stick with it, as the data is returned under the "MitreAttack" field that means data is related to Mitre Attack framework?

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-falcon-detection-summary.log-expected.json
navnit-elastic: I think we shouldn't alter the existing sample data, as we can not validate it with the current API behavior for these fields. Probably it isn't just a rename of fields to `MitreAttack`, it could be change the data also. Let me know what do you think?

kcreddy: I still think that we need sample logs (if not updating existing ones, add new logs) to confirm the pipeline tests working as expected.

packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/detection_summary.yml
kcreddy: This was a requirement actually: https://github.com/elastic/integrations/issues/10196. So we would still need it.

navnit-elastic: Got it. Now that detection events may contain multiple MITRE ATT&CK tactics and techniques, how do you think we should handle this case?

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-falcon-detection-summary.log-expected.json
navnit-elastic: OK. As I don't have sufficient sample data, I'll update the existing ones.

efd6: It's OK to duplicate with modification. This retains the existing test sample to avoid regression and adds the new case.

packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/detection_summary.yml
kcreddy: Can you confirm from live data if `MitreAttack` is multi-element array? If so, it might be worth to add that sample log into pipeline tests.

navnit-elastic: @kcreddy, It is confirmed from live events and documentation that the `MitreAttack` may contain multiple tactics and techniques. We can add sample test for it.

We can use [this](https://github.com/elastic/integrations/blob/a186039859c4e4a81b4c2fac4389b281569305be/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/xdr_detection_summary.yml#L227) script to extend `threat.framework` to contain a list of threat frameworks. WDYT?

kcreddy: Sometime it so happens that even though it was defined as an array, the live data may only contain a single object. But, thanks for checking on this. Please add this sample log.

> We can use [this](https://github.com/elastic/integrations/blob/a186039859c4e4a81b4c2fac4389b281569305be/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/xdr_detection_summary.yml#L227) script to extend threat.framework to contain a list of threat frameworks. WDYT?

Sounds good.

navnit-elastic:
In the recent commits I have renamed field `MitreAttack.PatternId` to `MitreAttack.PatternID` (capitalized 'D').

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-event-stream.log-expected.json
```diff
@@ -801,6 +777,190 @@
...
+                "ip": "81.2.69.144"
+            },
+            "tags": [
+                "preserve_original_event",
+                "SensorGroupingTags/USB_Storage_Exception"
+            ],
+            "threat": {
+                "tactic": {
+                    "id": [
+                        ""
```
kcreddy: These should be removed. Can you add a empty string check here? But also verify why null script processor not removing empty strings?

kcreddy:
> In the recent commits I have renamed field `MitreAttack.PatternId` to `MitreAttack.PatternID` (capitalized 'D').

@navnit-elastic, is this how it appears in the live data as well: `MitreAttack.PatternID`? If not, we should stick to the source naming convention. [Example](https://github.com/elastic/integrations/blob/main/packages/crowdstrike/data_stream/falcon/fields/fields.yml#L511)

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-event-stream.log-expected.json
navnit-elastic: Can you please explain how to add empty string check when using foreach-append processor? I'm not sure how we can access array elements, and that `ignore_empty_value: true` option is available in stack version 9.2+.

navnit-elastic (commented):
> > In the recent commits I have renamed field `MitreAttack.PatternId` to `MitreAttack.PatternID` (capitalized 'D').
> 
> @navnit-elastic, is this how it appears in the live data as well: `MitreAttack.PatternID`? If not, we should stick to the source naming convention. [Example](https://github.com/elastic/integrations/blob/main/packages/crowdstrike/data_stream/falcon/fields/fields.yml#L511)

@kcreddy, Yes, this is correct from live data.

I should have mentioned "corrected field name from `MitreAttack.PatternId` to `MitreAttack.PatternID`".

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-event-stream.log-expected.json
```diff
@@ -195,18 +195,6 @@
             "tags": [
                 "preserve_original_event"
             ],
-            "threat": {
```
efd6, line 198: The issue says that the deprecated fields are still found in live logs, and I imagine that there will be users who have retained older logs that use the deprecated fields. So we should continue to work with those. This means this test event should not change.

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-falcon-data-protection-detection-summary.log-expected.json
```diff
@@ -204,7 +222,9 @@
                 "preserve_original_event"
             ],
             "threat": {
-                "framework": "MITRE ATT&CK",
+                "framework": [
+                    "MITRE ATT&CK"
+                ],
```
efd6: I think this should not be an array; the [description](https://www.elastic.co/docs/reference/ecs/ecs-threat#field-threat-framework) refers to "name" in the singular.

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-event-stream.log-expected.json
efd6: The recursive empty value removal script should handle this. If it doesn't it should be straightforward to make it so that it does.

Checked; it does not currently have one. This fixes it (taken from the fdr ingest pipeline with modification):
```diff
diff --git a/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml b/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
index 8b8786f900..ada609b462 100644
--- a/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
@@ -592,6 +592,31 @@ processors:
         - crowdstrike.event.UserId
         - crowdstrike.event.UserName
       ignore_missing: true
+  - script:
+      description: This script processor iterates over the whole document to remove fields with null/empty values.
+      tag: remove_nulls
+      source: |-
+        void handleMap(Map map) {
+          map.values().removeIf(v -> {
+            if (v instanceof Map) {
+                handleMap(v);
+            } else if (v instanceof List) {
+                handleList(v);
+            }
+            return v == null || v == '' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0)
+          });
+        }
+        void handleList(List list) {
+          list.removeIf(v -> {
+            if (v instanceof Map) {
+                handleMap(v);
+            } else if (v instanceof List) {
+                handleList(v);
+            }
+            return v == null || v == '' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0)
+          });
+        }
+        handleMap(ctx);
 
 on_failure:
   - append:
```

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-event-stream.log-expected.json
navnit-elastic: I was thinking we should align with CrowdStrike’s deprecation notice and support the API exactly as it currently behaves. However, the argument regarding older logs is considerable. Should we explicitly call out the deprecation of the old threat fields and proceed with the changes in this PR, or would it be better to support both for backward compatibility?

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-falcon-data-protection-detection-summary.log-expected.json
navnit-elastic: Using an array allows us to capture both `MITRE ATT&CK` and `CrowdStrike Falcon Detections Framework` in a single event.

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-event-stream.log-expected.json
efd6: My preference would be to retain backwards compatibility. We can mark as deprecated with a view to removal in the future.

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-falcon-data-protection-detection-summary.log-expected.json
efd6: Do we see those together in a single document?

navnit-elastic: I couldn't find evidence for those to be in a single event.

navnit-elastic:
It is still unclear to me how we should determine the value for `threat.framework` when an event contains multiple Mitre Attack data. While Dan's argument that we likely won't see both Mitre Attack and Falcon Framework data in a single event is impactful, but the possibility still exists.

If we assume an event will only contain one framework, how should we validate that? Should we simply check the first value in the `MitreAttack` array?

efd6:
> If we assume an event will only contain one framework, how should we validate that? Should we simply check the first value in the `MitreAttack` array?

I don't think we can assume anything that is not documented, so we can't check the first one; if there is more than one, they must agree. If we don't have that as a documented invariant, can we enforce that as an invariant in the input? I don't think we can since we have both logs and streaming data; logs can't split docs and while streaming can, I would not want to include all that logic in the program. So we should try to confirm that there is only ever one type per document, and if we can't then concede that we need to have an array, or use a heuristic to choose one of the set we have (and we should think very hard before we take this path).

efd6:
We've discussed this now. We cannot make this an array; `threat.framework` should be a scalar keyword, consistent with ECS. The approach should be:

1. Pick one framework deterministically: If all tactic/technique IDs in the MitreAttack array share the same prefix, use that framework. This is almost certainly always the case.
2. If mixed frameworks ever appear: the pipeline should pick one, probably MITRE ATT&CK since it's the more widely recognised standard, and preserve the raw detail in the CrowdStrike-specific fields for anyone who needs the full picture. We can provide a preference-ordered list as a param to use in selecting which of the frameworks we've founds, e.g. `["MITRE ATT&CK", "CrowdStrike Falcon Detections Framework"]` (currently I can find only two, but implementing it this way provides a simple path for extension).

navnit-elastic:
> We've discussed this now. We cannot make this an array; `threat.framework` should be a scalar keyword, consistent with ECS. The approach should be:
> 
> 1. Pick one framework deterministically: If all tactic/technique IDs in the MitreAttack array share the same prefix, use that framework. This is almost certainly always the case.
> 2. If mixed frameworks ever appear: the pipeline should pick one, probably MITRE ATT&CK since it's the more widely recognised standard, and preserve the raw detail in the CrowdStrike-specific fields for anyone who needs the full picture. We can provide a preference-ordered list as a param to use in selecting which of the frameworks we've founds, e.g. `["MITRE ATT&CK", "CrowdStrike Falcon Detections Framework"]` (currently I can find only two, but implementing it this way provides a simple path for extension).

Thank you for the clarification, Dan.

packages/crowdstrike/data_stream/falcon/_dev/test/pipeline/test-falcon-events.log-expected.json
```diff
@@ -222,6 +222,157 @@
...
+                "name": "Process Terminated"
+            },
+            "source": {
+                "ip": "192.168.12.51"
+            },
+            "tags": [
+                "preserve_original_event"
+            ],
+            "threat": {
+                "framework": "MITRE ATT&CK",
```
navnit-elastic, line 349: This test case was modified to test multiple frameworks in a single event.

packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -419,6 +419,133 @@ processors:
...
+          - ai powered ioa
+          - insecure security posture
+      source: |
+        def tid = ctx.threat.tactic?.id;
+        def nid = ctx.threat.technique?.id;
+        def tname = ctx.threat.tactic?.name;
+        if ((tid == null || tid.isEmpty()) && (nid == null || nid.isEmpty()) && (tname == null || tname.isEmpty())) {
+          return;
+        }
+        Set list = new HashSet();
```
efd6: I think we can name this better; it's a set not a list, but what's more important is what its semantics are, rather than its structure. I would suggest `frameworks`.

packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -419,6 +419,133 @@ processors:
...
+        }
+        // Handling falcon specific tactics.
+        if (tname != null && !tname.isEmpty()) {
+          for (String t: tname) {
+            if (params.falcon_tactic_names.contains(t.toLowerCase())) {
+              list.add(params.framework_cs);
+            }
+          }
+        }
+
```
efd6, line 552: If the set of frameworks has only one entry, we can set `ctx.threat.framework` here and return early, or if it's empty, return without setting.

packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -419,6 +419,133 @@ processors:
...
+        def chosen = null;
+        for (def preferred : params.framework_preference) {
+          if (list.contains(preferred)) { chosen = preferred; break; }
+        }
+        if (chosen != null) {
+          ctx.threat.framework = chosen;
+        }
```
efd6: This is not quite right. Since we are being defensive here, we should make sure we handle cases where the we don't yet know about the framework (that is, it's not in the preference list). If we don't see any of the elements in the set of frameworks that we have in the preferences here, we end up leaving this empty. I don't think we should do that. If we haven't matched anything, we should just pick one randomly. This can be done by
```suggestion
        for (def preferred : params.framework_preference) {
          if (frameworks.contains(preferred)) {
            ctx.threat.framework = preferred;
            return;
          }
        }
        for (def framework : frameworks) {
          ctx.threat.framework = framework;
          if (framework != null) {
            return;
          }
        }
```
Note that I've used `frameworks` instead of `list`. The apparently pointless condition in the second loop is to evade the painless check that prevents always breaking out of the first iteration of the loop so that you can get the side-effect of getting a randomish element. The alternative is to assign the set to a list and then get the first element which just seem massively pointless.

packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
navnit-elastic: Thanks.

packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
navnit-elastic: Currently, we only ever add `framework_cs` and `framework_ma`, and both are in framework_preference, so that second loop is currently unreachable. We can keep it as a defensive fallback for the future.

efd6: Ack

packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -506,41 +506,56 @@ processors:
...
-        if (chosen != null) {
-          ctx.threat.framework = chosen;
+
+        // fallback when new frameworks are added and not yet in preference list
+        for (def framework : frameworks) {
+          ctx.threat.framework = framework;
+          if (framework != null) {
+            return;
+          }
         }
```
efd6: Your `ctx.threat.framework = frameworks.iterator().next();` is way better than what I had.

navnit-elastic:
@efd6, following up on this. Do these changes look good to you now?

navnit-elastic:
Thank you for the update.

Issue #17264, opened 2026-02-05
[Pfsense]: Add support for dnsmasq and kea dhcp logs


daviddreher2:
### Integration Name

pfSense [pfsense]

### Dataset Name

pfsense.log

### Integration Version

1.25.0

### Agent Version

9.3.0

### OS Version and Architecture

Debian 13 (LXC)

### User Goal

Add support for Parsing Kea and dnsmasq-dhcp logs from PFSense & OPNsense

ISC DHCP (dhcpd) is deprecated and replaced by Kea and dnsmasq
existing Grok patterns for parsing dhcp logs do not match and logs are dropped by the pipeline

### Existing Features

Existing Grok Patterns for dhcpd do not match kea or dnsmasq logs

### What did you see?

Logs with the event provider dnsmasq-dhcp are dropped by the ingest pipeline, no logs are ingested

### Anything else?

Current fix
added dnsmasq-dhcp to the exclusion list for the drop action and to the inclusions for the dhcp pipeline

added a grok pattern to match dnsmasq logs
%{DATA:_tmp.action}\(%{DATA:observer.ingress.interface.name}\)(?: %{IP:client.ip})? %{MAC:client.mac}(?: %{HOSTNAME:pfsense.dhcp.hostname})?

also added a gsub to substitute dnsmasq-dhcp to dhcpd in the event.provider field to fix dashboards

after changes the logs ingest and the dashboards populate

qcorporation:
Thank you @daviddreher2 for the filing of the issue and the fix.
Please look out for release 1.25.2 - marking this issue as fixed

PR #17375, opened 2026-02-11
[Enhancement] Add support for dnsmasq-dhcp for PFSense integration

daviddreher2:
Updated ingest pipelines for PFSense DHCP component to support logs from dnsmasq DHCP server, currently only supports deprecated ISC DHCP server logs

**Changes:**

**Default Pipeline**
 - added dnsmasq-dhcp to the list for the drop processor
 - added dnsmasq-dhcp to the list for the dhcp pipeline

**DHCP Pipeline**
 - added Grok pattern to extract dnsmasq log fields to the correct field names
 - added a gsub operator to replace dnsmasq-dhcp with  dhcpd in the event.provider field to ensure dashboards work correctly

**Checks:**
 - Confirmed Grok pattern extracts correctly
 - Confirmed dashboard loads data correctly
 - Checked pipeline for errors

**Validations:**
 - configure PFSense/OPNsense to use dnsmasq for DHCP
 - Send dnsmasq logs to Elastic agent integration via syslog
 - confirm data 

Partially resolves issue #17264


qcorporation:
Thanks you @daviddreher2 for the submissions, some comments:

1. Would it be possible to add logs with removed PII into `packages/pfsense/data_stream/log/_dev/test/pipeline/test-pfsense-dhcp.log` and then using the `--g` command within `elastic-package test pipeline` to generate the expected json within `packages/pfsense/data_stream/log/_dev/test/pipeline/test-pfsense-dhcp.log-expected.json`? This will allow us to test the grok pattern that was added to tag `grok_message_efb70afb`
2. Is it also possible to modify `changelog.yml` file with a description of this change
3. Also, when we bump the version can we also change the `manifest.yml` file to the new version?
4. One last thing, are there logs for kea, `kea-dhcp4` or `kea-dhcp6` that we have to support as well

daviddreher2:
@qcorporation i have run the pipeline tests and updated the files requested.

i have also not tested any kea dhcp or dhcpv6 logs and this change applies only to dnsmasq-dhcp

daviddreher2:
This change is approved but the build appears to be stuck in some way
please confirm, thanks

daviddreher2:
please re-run build, Thanks

PR #17305, opened 2026-02-09
[Cloud Asset Inventory] Support GCP Cloud Connectors

amirbenun:
### Summary
Adds GCP Cloud Connectors support to the Cloud Asset Inventory package, aligned with PR #17185 (Cloud Security Posture). This allows cloudbeat to use Cloud Connectors for GCP asset discovery with service account impersonation.

PR #17386, opened 2026-02-12
Implement verifier integration

jeniawhite:
## Proposed commit message

Adds the Permission Verifier integration package (verifier_otel) so Cloud Connectors can run permission checks for their integrations and send results to Elasticsearch via the OTEL Collector’s Verifier receiver.

Cloud Connector deployments need to confirm that IAM/API credentials have the permissions required by each attached integration. This integration configures the Elastic distribution’s Verifier receiver so that:
- Verification runs (on-demand or scheduled) with the connector’s credentials.
- Results are sent as OTEL logs to Elasticsearch with policy/integration/permission context.
- Users can see which permissions are granted, denied, or errored and fix configuration.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/verifier_otel/docs/README.md
```diff
@@ -0,0 +1,156 @@
...
+| `verification.type` | `on_demand` or `scheduled` |
+| `service.name` | Always `permission-verifier` |
+
+### Log Record Attributes
+
+| Attribute | Description |
+|-----------|-------------|
+| `policy.id` | Policy identifier |
+| `policy.name` | Policy name |
+| `integration.id` | Integration instance identifier |
```
Omolola-Akinleye: Thinking about the `Integration Type` field — would it make sense to map this directly to `policy_template` from `manifest.yml` instead? Since Fleet's package policy API already models around `policy_template`, we'd get consistency end-to-end and avoid any mapping logic on the Kibana side.

Would also propose embedding package metadata alongside the integration

```json
{
  "policy_template": "guardduty",
  "package_policy_id": "",
  "package": {
    "name": "aws",
    "title": "AWS",
    "version": "6.2.0"
  }
  "namespace": "default"
}
```

This makes the Agentless API request fully self-describing — no lookups needed, clean audit trail, and the vocabulary stays consistent with Fleet. One thing worth noting in the spec: `policy_template` alone isn't globally unique across packages, so the safe unique key is the composite of `package.name` + `policy_template` (which this shape already captures).

packages/verifier_otel/manifest.yml
```diff
@@ -0,0 +1,133 @@
...
+        multi: false
+        required: false
+        show_user: false
+        default: on_demand
+        options:
+          - value: on_demand
+            text: On Demand
+          - value: scheduled
+            text: Scheduled
+      - name: providers.aws.credentials.role_arn
```
Omolola-Akinleye: Prefixing will add maintenance overhead on the Kibana side. We've already had to deal with mapping prefixes for AWS package fields that are named differently across CSPM, Asset Inventory, and AWS packages. A lot of work went into normalizing these down to `role_arn`, `external_id`, and `default_region` across both the AWS package and Cloud Connector saved object — prefixing would undo that.

Suggestion: keep the normalized snake_case fields as-is and add a `provider` meta field to carry the context that would otherwise be encoded in prefixes:

```json
{
  "provider": "aws",
  "role_arn": "",
  "external_id": "",
  "default_region": ""
}
```
OR 

If some namespacing is preferred, a `credentials_` snake_case prefix is a reasonable middle ground — it's consistent, flat, and readable without introducing a new translation layer:

```json
{
  "provider": "aws",
  "credentials_role_arn": "",
  "credentials_external_id": "",
  "default_region": ""
}
```

`default_region` sits outside the prefix since it's config, not auth. Hopefully, this keeps the shape flat while still giving credential fields a clear grouping and keeps  fields stable and reusable across integrations without Kibana needing to own a translation layer for every package's naming conventions.

packages/verifier_otel/_dev/test/policy/test-eprcheck.yml
```diff
@@ -0,0 +1,17 @@
+vars:
+  cloud_connector_id: test-cc-123
+  cloud_connector_name: Test Connector
+  verification_id: test-verify-456
+  verification_type: on_demand
+  # AWS Provider Credentials
+  providers.aws.credentials.role_arn: arn:aws:iam::123456789012:role/TestRole
+  providers.aws.credentials.external_id: test-external-id
```
Omolola-Akinleye: Assuming a role behaves differently depending on whether you're targeting a single account vs an organization (management) account in cloudbeat. How about when verifying permissions? 

Do you need additional `account_type` as metadata to help verify permissions? 
```
// Cloud connector SO Model
{
        mappings: {
        dynamic: false,
        properties: {
          name: { type: 'keyword' },
          namespace: { type: 'keyword' },
          cloudProvider: { type: 'keyword' },
          accountType: { type: 'keyword' },
          vars: { type: 'flattened' },
          created_at: { type: 'date' },
          updated_at: { type: 'date' },
        },
   }
}
```

packages/verifier_otel/_dev/build/docs/README.md
```diff
@@ -0,0 +1,156 @@
...
+| Integration Name | No | Human-readable name of the integration |
+| Integration Version | No | Semantic version of the integration package (for example, `2.17.0`). Different versions can require different permissions. When empty, the latest permission set is used. |
+
+## Supported Integration Types
+
+### AWS Integrations
+
+| Integration Type | Permissions Verified |
+|-----------------|---------------------|
+| `aws_cloudtrail` | cloudtrail:LookupEvents, DescribeTrails, s3:GetObject, ListBucket, sqs:ReceiveMessage |
```
Omolola-Akinleye: So is the permission matrix  base on  least-privilege principle  so  each permission will be scope to integration type as opposed global permissions shared by integration package?

jeniawhite:
@claude

packages/verifier_otel/_dev/build/docs/README.md
```diff
@@ -0,0 +1,253 @@
...
+| `provider.type` | Provider type (`aws`, `azure`, `gcp`, `okta`) |
+| `provider.account` | Provider account identifier |
+| `provider.region` | Provider region |
+| `provider.subscription_id` | Azure subscription ID (when applicable) |
+| `provider.project_id` | GCP project ID (when applicable) |
+| `account_type` | `single_account` or `organization` |
+| `permission.action` | Permission being checked (for example, `cloudtrail:LookupEvents`) |
+| `permission.category` | Permission category (for example, `data_access`) |
+| `permission.status` | Result: `granted`, `denied`, `error`, or `skipped` |
+| `permission.required` | Whether this permission is required |
```
Omolola-Akinleye, line 170: @jeniawhite just one more thought when I sync'd with Oleg. He mentioned there can be IAM Policy with multiple permissions which I didn't think about.  What is the relationship between verification and handling IAM Policy/permissions.

For instance,if  user could update IAM policy include multiple permissions.  Will each verification doc check multiple permissions? 

IAM Policy  `AmazonS3ReadOnlyAccess` for Amazon S3.
```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "s3:Get*",
        "s3:List*",
        "s3:Describe*",
        "s3-object-lambda:Get*",
        "s3-object-lambda:List*"
      ],
      "Resource": "*"
    }
  ]
}
```

jeniawhite: Current state of the verifier maps the required actions per integration and these are the actions that are being verified and send as events. If we'd like we can verify the IAM roles as well by attempting to assume them (without triggering actions). This would require changes to the verifier by adding an additional type of verification (non API...). We'll need to understand what we want to verify for the integration and have it as part of the registry. If we decide to only assume then there are cons like SCP rules that might block requests and misaligned action configurations. Should I create discussion around this topic? @Omolola-Akinleye @olegsu

olegsu: When we ask customers in documentation page that in order for an integration to work properly we need an access to builtit, support by aws role, this is how I expect it to be reflected in case the trust was revoked.

packages/verifier_otel/manifest.yml
```diff
@@ -28,6 +28,11 @@ policy_templates:
     description: Verify cloud connector based integration permissions and report status.
     input: otelcol
     template_path: input.yml.hbs
+    deployment_modes:
+      default:
```
olegsu, line 32: Should agent-based deployment be supported as well? What is the use case?

jeniawhite: There were discussions about potentially supporting verifier policies for non-agentless deployments as well (in future iterations).
I thought of opening this option as well to avoid making changes in the future, but removed it for now (it is not expected for the user to create such policies and they will be created by Kibana background task, so this shouldn't have affected anything).

jeniawhite: It looks like the EPR tests will not pass since there is no support in EPR for packages that cannot be deployed on default and can be deployed on agentless.
The only package with the same case scenario is elastic_connectors and it doesn't have any EPR tests.
Since we'd like to have EPR tests and we don't expect users to create these policies (they will be created by Kibana in an automated manner), I think that we can continue with that being enabled and cover this gap later on by introducing changes to the EPR. I rolled back the changes.

packages/verifier_otel/_dev/build/docs/README.md
```diff
@@ -0,0 +1,256 @@
...
+cloud_connector_id: "cc-gcp-01"
+cloud_connector_name: "GCP Connector"
+verification_id: "verify-ghi789"
+verification_type: "on_demand"
+
+provider: "gcp"
+account_type: "single_account"
+
+credentials_project_id: "my-gcp-project-123"
+credentials_workload_identity_provider: "//iam.googleapis.com/projects/123/locations/global/workloadIdentityPools/pool/providers/provider"
```
Omolola-Akinleye, line 241: `credentials_workload_identity_provider` makes sense but I notice our integrations use term [audience](https://github.com/elastic/integrations/pull/17305/changes) can we keep it consistent here as well? I think [GCP Docs](https://docs.cloud.google.com/iam/docs/best-practices-for-using-workload-identity-federation#use-immutable-attributes) treats the url as audience as well.

packages/verifier_otel/_dev/build/docs/README.md
```diff
@@ -0,0 +1,256 @@
...
+| Field | Required | Description |
+|-------|----------|-------------|
+| `credentials_tenant_id` | Yes | Azure AD tenant ID |
+| `credentials_client_id` | Yes | Azure application (client) ID |
+
+#### GCP Credentials
+
+| Field | Required | Description |
+|-------|----------|-------------|
+| `credentials_project_id` | Yes | GCP project ID to scope verification to |
```
Omolola-Akinleye, line 59: When I was working with Amir we didn't include GCP Project ID in the CSPM and Asset Inventory Integrations Manifest. GCP Cloud Connector UX implementation  doesn't store Project ID. If we want to store GCP Project ID and verify permissions; there might changes need on backend.

packages/verifier_otel/docs/README.md
```diff
@@ -0,0 +1,256 @@
...
+account_type: "single_account"
+
+credentials_role_arn: "arn:aws:iam::123456789012:role/ElasticAgentRole"
+credentials_external_id: "elastic-external-id-from-setup"
+default_region: "us-east-1"
+
+policy_id: "policy-aws-security"
+policy_name: "AWS Security Monitoring"
+
+policy_template: "cloudtrail"
```
Omolola-Akinleye, line 198: Thinking about scale  how are we going  verify a cloud connector that have mutilple  packages such as CSPM and Asset Inventory) or single package with multiple policy templates (guard duty and cloud trail)

- If we want verify permission for multiple integration types  for AWS which  Guard Duty and Cloud trail. Should policy templates be an array in this case?

- We have edge case where two packages that share the same IAM Policy Security Audit; hence, can use the same cloud connector.  How should we verifying  multiple packages that share a cloud connector?

Omolola-Akinleye (approved):
LGTM! Thanks for the changes!

PR #17400, opened 2026-02-12
 New Integration Kiteworks Totemomail

WildDogOne:
## New Integration Kiteworks Totemomail

Integration for the Kiteworks Totemomail Service.
Totemomail is a Securemail service used for encrypted and or signed mail traffic between organisations and users.
The goal of the integration is that Mailflow through the service can be traced correctly.
Additionally I want to make sure that Debug logs are extracted to help troubleshooting the service if needed.

## Checklist

## Author's Checklist

- [ ] Evaluate if a variable to choose which logtypes are ingested makes sense (debug/info)
- [ ] Fix Build Error: ``ingest_pipeline/default.yml" is invalid: field processors.16.remove: if is required``
- [ ] Fix Build Error: ``ingest_pipeline/default.yml" is invalid: field processors.16.remove.field: rename "message" to "event.original" processor requires remove "message" processor (JSE00001)``

Logs are included in the pull requests


WildDogOne:
I want to ask if it would make sense to have a flag for choosing which logs the organisation wants to ingest, just informational logs or debug logs as well.

Of course an organisation could always make a custom pipeline to drop debug logs, but maybe this is a feature people would enjoy?

Issue #2859, opened 2022-03-22
modsecurity integration not working correctly with modsecurity v2 Apache and IIS


stefanpinter:
Hi!

There is a modsecurity integration available which I am very thankful about. Sadly it doesn't give me all the fields... I see rule.id when trying to create a vizualisation but it seems to be empty. Picking out some messages from "Logs" shows that they DO have some json-fields populated but the other expected fields arent.

What can I contribute to troubleshoot this?

example of modsec_audit.json (not complete though)
...link expired...

thank you!

stefanpinter:
same with modsecurity 2.9.5 on IIS... it doesnt event get the response code (json.response.status) right (says 0 instead of 403)

jamiehynds:
Hi @stefanpinter - the link for the sample event seemed to have expired. Can you post again please? We can take a look and confirm if our Modsecurity pipeline requires an adjustment. @sawwn23 contributed the integration and may have some familiarity with your sample event.

stefanpinter:
Hi!
oh I'd love that :)
https://ourvolaris-my.sharepoint.com/:u:/g/personal/stefan_pinter_motiondata-vector_com/EaskfNUQ8BZEqvIuhaHEcMkBzi_Kk4kHJTwAtQcsZNeueg?e=4wgGXq

stefanpinter:
@sawwn23  hi! sorry, I am unsure if you get a notification when I don't tag you, so I tag you now, sorry to annoy you

sawwn23:
Hi @stefanpinter I will work with your sample on this holiday

emnp:
Hi @stefanpinter We're reviewing Modsecurity issues. Can you pls send the previous sample data again?

stefanpinter:
hi @emnp !
This is good to see! 
of course I can supply the sample data again:
https://ourvolaris-my.sharepoint.com/:u:/g/personal/stefan_pinter_motiondata-vector_com/EaskfNUQ8BZEqvIuhaHEcMkBzi_Kk4kHJTwAtQcsZNeueg?e=F96m7s

on a site note, if anyone in a similar situation wants to work with something different, for now: I created a badly written bash-script that filters out relevant data of concurrent modsecurity v2 logs
https://github.com/stefanpinter/modsecurityhelpers
this has nothing to do with elastic/kibana

PR #17410, opened 2026-02-13
[modsecurity] Update documentation

ilyannn:
## Summary
- Add `service_info.md` knowledge base with reviewed, fact-checked integration details
- Rewrite documentation using `elastic-package update documentation` to follow the new template structure
- Add `docs_structure_enforced` validation
- Bump version to 1.23.0

## Breaking Changes in Docs

Compared to the previous docs, we:

- fix the error where the docs referred to ModSecurity v3 with Apache, even though [the original issue referred to v2](https://github.com/elastic/integrations/issues/2859)
- use `modsec-audit.json` file name in examples rather than `modsec_audit.json`
- do not describe the importance of `SecAuditLogType Serial` as critical
- only suggest `ABFHJZ` parts as an example (was `ABDEFHIJZ` previously)

## Kibana rendering

Note the code in numbered list looks like this but this apparently how our rendering works (perhaps we should reduce the indent of code in the lists in this case?). GitHub does not like this formatting either.

image

## Test plan
- [x] `elastic-package check` passes (tested with both system and dev builds)
- [x] Verify README renders correctly in Kibana integration UI
- [x] Review generated documentation for accuracy against `service_info.md` ground truth

Closes https://github.com/elastic/integration-experience/issues/456

🤖 These documentation changes have been largely GenAI-generated as per our protocol. Additionally, Claude Code was used for factual review and as a harness for editing/commit/PR generation under our editorial supervision.

### DIFF
--- a/packages/modsecurity/docs/README.md
+++ b/packages/modsecurity/docs/README.md
@@ -6,12 +6,6 @@
 
 The ModSecurity Audit integration for Elastic enables you to collect and analyze logs from ModSecurity, the open-source Web Application Firewall (WAF). This integration allows you to ingest detailed transaction logs into the Elastic Stack, providing visibility into HTTP requests and responses for real-time threat detection and forensic investigation.
 
-You can use this integration to support several security operations:
-- Monitor logs for common web attacks such as SQL injection (SQLi), Cross-Site Scripting (XSS), and Local File Inclusion (LFI)
-- Maintain a comprehensive audit trail of all HTTP transactions and security interventions to satisfy compliance requirements
-- Review detailed audit logs to identify legitimate traffic being blocked by restrictive rules, allowing you to create exceptions and reduce false positives
-- Correlate WAF events with other system and network logs to trace attacker activity across your infrastructure
-
 ### Compatibility
 
 This integration is compatible with the following third-party components:


packages/modsecurity/_dev/build/docs/README.md
```diff
@@ -1,22 +1,214 @@
...
-SecAuditLogParts ABDEFHIJZ
-SecAuditLogType Serial
-SecAuditLog /var/log/modsec_audit.json
-SecAuditLogFormat JSON
-```
+You can use this integration to support several security operations:
+- Monitor logs for common web attacks such as SQL injection (SQLi), Cross-Site Scripting (XSS), and Local File Inclusion (LFI)
+- Maintain a comprehensive audit trail of all HTTP transactions and security interventions to satisfy compliance requirements
+- Review detailed audit logs to identify legitimate traffic being blocked by restrictive rules, allowing you to create exceptions and reduce false positives
+- Correlate WAF events with other system and network logs to trace attacker activity across your infrastructure
```
mjwolf:
```suggestion
```

This list can be removed, it's duplicated by the Support Use case section below. Sorry, it was a mistake in the generator that this was added

ilyannn:
> please fix the reasonable vale recommendations

Which ones do you consider reasonable? I think I already fixed everything I could.

theletterf:
Most of those are false positives we need to fix in the Vale rules. We can disregard.

theletterf:
Alas, we don't accept `apache` as a valid syntax highlighting value. Opening a PR to fix this.

PR #17411, opened 2026-02-13
[beyondinsight_password_safe] Handle null password ion authentication

StacieClark-Elastic:
## Proposed commit message

beyondinsight_password_safe: handle optional password in authentication

The BeyondInsight API does not always require a password for
authentication. The password will be null when it is not supplied.
Whether one is needed depends on the "User Password
Required" setting on the API registration in BeyondInsight. When no
password was configured, the integration failed because it assumed
the password field was always present in state.

ref: https://docs.beyondtrust.com/bips/docs/bi-cloud-configure-api

## Checklist

- [x ] I have verified that all data streams collect metrics or logs.
- [ x] I have added an entry to my package's `changelog.yml` file.


packages/beyondinsight_password_safe/data_stream/asset/_dev/test/system/test-no-password-config.yml
```diff
@@ -0,0 +1,15 @@
+input: cel
+service: beyondinsight-asset-cel
+vars:
+  url: http://{{Hostname}}:{{Port}}/BeyondTrust/api/public/v3
+  apikey: test_api_key
+  username: testuser2
+  password: null
```
efd6: This has no default in the manifest, so it can be omitted here.
```suggestion
```

packages/beyondinsight_password_safe/data_stream/asset/_dev/test/policy/test-all.expected
efd6, line 1: In the `state` construction I think the password field should be conditionally included
```
{{#if password}}
  password: {{escape_string password}}
{{/if}}
```

packages/beyondinsight_password_safe/data_stream/asset/agent/stream/cel.yml.hbs
```diff
@@ -242,7 +242,7 @@ program: |-
           "Header": {
             "Authorization": [
               sprintf("PS-Auth key=%s; runas=%s;", [state.apikey, state.username]) +
-              ((state.password != "") ? (sprintf(" pwd=[%s];", [state.password])) : ""),
+              ((state.?password.orValue("") != "") ? (sprintf(" pwd=[%s];", [state.password])) : ""),
```
efd6: With conditional rendering into the config, this becomes `has(state.password) ? …`

packages/beyondinsight_password_safe/data_stream/asset/_dev/test/system/test-no-password-config.yml
StacieClark-Elastic: The test is specifically for when the password is null. Should we ever add a default in the manifest then this test will no longer test the condition for which it is written.

packages/beyondinsight_password_safe/data_stream/asset/_dev/test/policy/test-all.expected
StacieClark-Elastic: The password always exists. Sometimes it is null.

packages/beyondinsight_password_safe/data_stream/asset/agent/stream/cel.yml.hbs
StacieClark-Elastic: The password always exists. sometimes it is null.

packages/beyondinsight_password_safe/data_stream/asset/_dev/test/policy/test-all.expected
StacieClark-Elastic: I discussed this with @andrewkroh before I did the fix as I was also confused about the password existing and being null. This fix will allow users to not have to update their policies when the integration is updated.

packages/beyondinsight_password_safe/data_stream/asset/agent/stream/cel.yml.hbs
efd6: Can you confirm that `password: {{escape_string password}}` gives `{"password": ""}` if the `password` var is null?

StacieClark-Elastic: password: {{escape_string password}} gives null if the password is null. I added in the 'if' block and added some policy tests for null and empty string to verify that no password is set if the variable is null or empty and changed the system tests to test for a missing password.

packages/beyondinsight_password_safe/data_stream/asset/_dev/deploy/docker/files/config.yml
```diff
@@ -48,7 +68,6 @@ rules:
               }
           ]
           `}}
-
```
efd6, line 51: These deleted blank lines (also below) are stanza breaks for clarity. They should not be removed.

packages/beyondinsight_password_safe/data_stream/managedaccount/_dev/deploy/docker/files/config.yml
```diff
@@ -36,7 +36,27 @@ rules:
             "Name": "Test User"
           }
           `}}
-
```
efd6, line 39: Same here and below.

packages/beyondinsight_password_safe/data_stream/managedsystem/_dev/deploy/docker/files/config.yml
```diff
@@ -36,7 +36,26 @@ rules:
             "Name": "Test User"
           }
           `}}
-
```
efd6, line 39: And here and below.

packages/beyondinsight_password_safe/data_stream/session/_dev/deploy/docker/files/config.yml
```diff
@@ -33,7 +33,26 @@ rules:
             "Name": "Test User"
           }
           `}}
-
```
efd6, line 36: And here and below.

packages/beyondinsight_password_safe/data_stream/useraudit/_dev/deploy/docker/files/config.yml
```diff
@@ -41,7 +41,26 @@ rules:
             "Name": "Test User"
           }
           `}}
-
```
efd6, line 44: And here and below.

packages/beyondinsight_password_safe/data_stream/asset/agent/stream/cel.yml.hbs
efd6: Can you confirm that a password state update from "something" to falsey will remove the password from the profile's configuration state?

packages/beyondinsight_password_safe/data_stream/managedaccount/_dev/deploy/docker/files/config.yml
StacieClark-Elastic: elastic-package format made those changes. That needs to be fixed.

packages/beyondinsight_password_safe/data_stream/asset/agent/stream/cel.yml.hbs
StacieClark-Elastic: If I update a policy that had a password, then leave the password field blank when I update the policy, it generates a policy with no password:
original:
```
        state:
          limit: 1000
          apikey: ${SECRET_0}
          password: ${SECRET_1}
          username: amiauser
```

updated:
```
        state:
          limit: 1000
          apikey: ${SECRET_0}
          username: amiauser
```

efd6 (approved):

PR #17455, opened 2026-02-18
ti_threatconnect: update pipeline to lowercase all hashes

GShepherdTC:
## Proposed commit message

Updated pipeline to lowercase all hashes

## Checklist

## Author's Checklist

- [ ]  Check code changes. Should be pretty straight forward. 

Run the pipeline test. I updated the response data to match. 

## Related issues

- Relates: https://github.com/elastic/ecs/issues/2586

## Screenshots

No changes to UI.

### DIFF
--- a/.github/workflows/pr-buildkite-detective.yml
+++ b/.github/workflows/pr-buildkite-detective.yml
@@ -16,7 +16,7 @@ jobs:
     if: >-
       github.event.workflow_run.conclusion == 'failure' &&
       toJSON(github.event.workflow_run.pull_requests) != '[]' &&
-      contains(fromJSON('["v1v", "mrodm", "fr4nc1sc0-r4m0n"]'), github.actor)
+      contains(fromJSON('["v1v", "mrodm", "fr4nc1sc0-r4m0n"]'), github.event.workflow_run.pull_requests[0].user.login)
     uses: elastic/ai-github-actions/.github/workflows/gh-aw-pr-buildkite-detective.lock.yml@v0
     with:
       buildkite-org: "elastic"
--- a/packages/modsecurity/_dev/build/docs/README.md
+++ b/packages/modsecurity/_dev/build/docs/README.md
@@ -68,23 +68,23 @@ To configure ModSecurity for log collection, follow these steps to enable JSON-f
 
 1.  **Locate the configuration file**: Identify your main `modsecurity.conf` file. Common paths include `/etc/nginx/modsec/modsecurity.conf` for Nginx or `/etc/modsecurity/modsecurity.conf` for Apache.
 2.  **Enable the audit engine**: Set the engine to log only relevant events to reduce volume.
-    ```apache
+    ```
     SecAuditEngine RelevantOnly
     ```
 3.  **Configure JSON format**: Ensure the logs are structured for the integration parser.
-    ```apache
+    ```
     SecAuditLogFormat JSON
     ```
 4.  **Set logging type**: Use serial logging to write all events to a single file.
-    ```apache
+    ```
     SecAuditLogType Serial
     ```
 5.  **Define log parts**: Specify which transaction parts to include. It's recommended to exclude part `K` to prevent ingestion issues. You can also include `C` for the full request body.
-    ```apache
+    ```
     SecAuditLogParts ABFHJZ
     ```
 6.  **Specify log path**: Set the destination file for the audit logs (for example, `/var/log/modsec-audit.json`).
-    ```apache
+    ```
     SecAuditLog /var/log/modsec-audit.json
     ```
 7.  **Integrate with the web server**: Ensure your web server configuration loads these rules.
--- a/packages/modsecurity/changelog.yml
+++ b/packages/modsecurity/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.23.1"
+  changes:
+    - description: Update the documentation.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/17578
 - version: "1.23.0"
   changes:
     - description: Update the documentation.
--- a/packages/modsecurity/docs/README.md
+++ b/packages/modsecurity/docs/README.md
@@ -68,23 +68,23 @@ To configure ModSecurity for log collection, follow these steps to enable JSON-f
 
 1.  **Locate the configuration file**: Identify your main `modsecurity.conf` file. Common paths include `/etc/nginx/modsec/modsecurity.conf` for Nginx or `/etc/modsecurity/modsecurity.conf` for Apache.
 2.  **Enable the audit engine**: Set the engine to log only relevant events to reduce volume.
-    ```apache
+    ```
     SecAuditEngine RelevantOnly
     ```
 3.  **Configure JSON format**: Ensure the logs are structured for the integration parser.
-    ```apache
+    ```
     SecAuditLogFormat JSON
     ```
 4.  **Set logging type**: Use serial logging to write all events to a single file.
-    ```apache
+    ```
     SecAuditLogType Serial
     ```
 5.  **Define log parts**: Specify which transaction parts to include. It's recommended to exclude part `K` to prevent ingestion issues. You can also include `C` for the full request body.
-    ```apache
+    ```
     SecAuditLogParts ABFHJZ
     ```
 6.  **Specify log path**: Set the destination file for the audit logs (for example, `/var/log/modsec-audit.json`).
-    ```apache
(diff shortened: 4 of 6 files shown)


packages/ti_threatconnect/data_stream/indicator/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1815,6 +1815,10 @@ processors:
       copy_from: threat_connect.indicator.md5
       ignore_empty_value: true
       if: ctx.threat_connect?.indicator?.type instanceof String && ctx.threat_connect.indicator.type.contains('File')
+  - lowercase:
```
andrewkroh: This needs to occur _before_ the value is copied into the ECS `threat.indicator.file.hash.md5` field.

Please check the sha1 and sha256 get the same treatment if they need it.

packages/ti_threatconnect/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.0.0"
+  changes:
+    - description: Update all hash types to be lowercase
```
efd6:
```suggestion
    - description: Normalize all hash fields to be lowercase.
```

GShepherdTC: Sounds good.  Will do.

packages/ti_threatconnect/data_stream/indicator/elasticsearch/ingest_pipeline/default.yml
GShepherdTC: Thanks.  Will look at incorporating file size as that makes sense and its not specific to hash type so I believe I just have to add the mapping once. Seems they didn't map it originally. 

But geo location is another story.  Thats not a first party attribute in our system but keyed to attributes and I see they didn't map any attributes over from what I can tell.  So I'm not going to touch that one.

GShepherdTC:
Anything else required, just drop me a line.  Thanks.

packages/ti_threatconnect/data_stream/indicator/_dev/test/pipeline/test-indicator.log
```diff
@@ -1 +1 @@
...
...ge","value": "PhishingEmailSender","createdBy": {"id": 69,"userName": "johnxys@abc.co","firstName": "John","lastName": "Smith","pseudonym": "JohnS","owner": "Elastic"},"lastModified": "2023-08-25T13:16:12Z","pinned": false,"default": false}]},"address": "hohnabc@xyz.com","legacyLink": "https://partnerstage-intel.threatconnect.com/auth/indicators/details/emailaddress.xhtml?emailaddress=misoyil388%40poverts.com&owner=Elastic","associatedArtifacts": {"data": [{"id": 12345}]},"associatedCases": {"data": [{"id": 123457}]},"fileActions": {"data": [{"id": 123456}]},"fileOccurrences": {"data": [{"fileName": "win999301.dll","path": "C:\\Windows\\System","date": "2022-06-14T10:00:00Z"}]},"customAssociations": {"data": [{"id": 123458}]},"dnsResolution": {"data": [{"id": 123459}]},"enrichment": {"data": [{"id": 123455}]},"falsePositives": 1,"lastFalsePositive": "2023-10-04T12:34:56Z","falsePositiveReportedByUser": false,"genericCustomIndicatorValues": {"data": [{"id": 1234551}]},"geoLocation": {"data": [{"id": 1234552}]},"investigationLinks": {"data": [{"id": 1234553}]},"observations": {"data": [{"id": 1234556}]},"trackedUsers": {"data": [{"id": 1234557}]},"whoIs": {"data": [{"id": 1234558}]}}
```
efd6, line 1: What is the change here?

packages/ti_threatconnect/data_stream/indicator/_dev/test/pipeline/test-with-associated-groups-and-attributes.log-expected.json
```diff
@@ -561,4 +564,4 @@
             }
         }
     ]
-}
+}
```
efd6: I suspect that you are using an older version of `elastic-packag` on the basis of this whitespace change. Can you update an regenerate?

packages/ti_threatconnect/data_stream/indicator/elasticsearch/ingest_pipeline/default.yml
efd6: It doesn't look like the ordering issue has been addressed.

packages/ti_threatconnect/data_stream/indicator/_dev/test/pipeline/test-indicator.log
GShepherdTC: File size type.

packages/ti_threatconnect/data_stream/indicator/_dev/test/pipeline/test-with-associated-groups-and-attributes.log-expected.json
GShepherdTC: Ok.   I see I am running:
elastic-package version-hash c40a1c2 (build time: 2024-03-21T12:48:10-04:00)

GShepherdTC:
@efd6 I don't understand what the issue is here: https://github.com/elastic/integrations/pull/17455#discussion_r2850634318
My testing (in platform and pipeline) passed as I saw in ES the correct file hashes and sizes.  Can you elaborate on what the requirement is and I'll take a look?  Since I only deal with ES about once a year (which is great given its solid), my familiarity is lacking.

packages/ti_threatconnect/data_stream/indicator/_dev/test/pipeline/test-indicator.log
efd6: Do you mean this, `"samplesha1","size": "124"` → `"samplesha1","size": 124`?

Why was that necessary?

packages/ti_threatconnect/data_stream/indicator/elasticsearch/ingest_pipeline/default.yml
efd6: Re https://github.com/elastic/integrations/pull/17455#issuecomment-3958766034, the issue is that the [`lowercase` processor](https://github.com/GShepherdTC/es_integrations_forked/blob/5b3611048cb0e0d593f2f80f057977df7cf4c71a/packages/ti_threatconnect/data_stream/indicator/elasticsearch/ingest_pipeline/default.yml#L1828-L1831) which acts on `threat_connect.indicator.md5` happens **after** the value of that field is [used to populate `threat.indicator.file.hash.md5`](https://github.com/GShepherdTC/es_integrations_forked/blob/5b3611048cb0e0d593f2f80f057977df7cf4c71a/packages/ti_threatconnect/data_stream/indicator/elasticsearch/ingest_pipeline/default.yml#L1822-L1827). This means that `threat.indicator.file.hash.md5` will not see the canonicalised value. The order of those two processors needs to be reversed. Andre is also asking that you check whether the same issue applies to the sha1 and sha256 hash fields.

packages/ti_threatconnect/data_stream/indicator/_dev/test/pipeline/test-indicator.log
GShepherdTC: An earlier review comment in this pull was to add in file size so I aimed to do that. And during testing of that, turned out the original test data was wrong.  Since it was never mapped, not sure it ever showed up for the original developers of this integration.   The value is numeric from Threatconnect. 

image

packages/ti_threatconnect/data_stream/indicator/elasticsearch/ingest_pipeline/default.yml
GShepherdTC: Ok. Will aim to change the order. I must be looking at the wrong data in the ES ui as I know I have seen the lowercase values.

GShepherdTC: Ok... submitted a change.   I was looking at the threat_connect data in UI as compared to threat.indicator version.  Thanks for the catch.

packages/ti_threatconnect/data_stream/indicator/_dev/test/pipeline/test-indicator.log
efd6: Great. Thanks for clarifying.

PR #17492, opened 2026-02-20
[sql_input] add cursor-based incremental data fetching support

stefans-elastic:
## Proposed commit message

Add cursor-based incremental data fetching to the `sql_input` package, backed by the SQL module cursor feature in elastic/beats#48722 (merged into Elastic Stack 9.4+).

**What**
- New Fleet UI fields: `sql_query` (single query for cursor mode), `sql_response_format` (now a `variables`/`table` select, replacing free text), and `cursor` (YAML block for `enabled`/`column`/`type`/`default`/`direction`).
- `agent/input/input.yml.hbs` renders `sql_query` through the `escape_string` Handlebars helper (matching the convention used by `cel`/`http_endpoint`/`websocket`/`elastic_connectors`) so queries containing quotes render as valid YAML.
- `docs/README.md` documents cursor mode: required fields, cursor type inference, scan direction, state persistence, error-handling semantics, and exactly which query requirements are enforced by the backend at startup vs. left to the user (`ORDER BY` is not validated — see below).
- New system tests exercise cursor mode end to end against real MySQL and PostgreSQL (seeded `audit_log` tables via mounted init scripts), verified manually against Elastic Stack 9.4.0 and 9.4.3. They are marked `skip` for CI: this package's `kibana.version` constraint is `^9.1.0`, and CI tests packages against their *oldest* declared version (see `.buildkite/scripts/common.sh`), where the cursor backend does not exist yet. `elastic-package`'s system-test `skip` config has no version condition, so an unconditional skip (with a reason and link to elastic/beats#48722) is the correct mechanism until the package's minimum version is raised or elastic-package adds version-gated skipping.

## Checklist

- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition). — **Needs a maintainer decision**: `kibana.version` stays `^9.1.0` so existing non-cursor users aren't forced onto a newer stack, but this means Fleet won't stop a user on <9.4 from enabling cursor mode (it will silently do nothing). Flagging for discussion rather than deciding unilaterally.
- [x] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) — N/A, no dashboard changes.

## Author's Checklist

- [x] Rendered the Handlebars input template locally (Node + `escape_string` stand-in) for cursor and classic modes, including a query with embedded double quotes, and confirmed valid YAML output.
- [x] Ran `elastic-package format/lint/build/check` — all pass.
- [x] Verified cursor mode end-to-end against a real MySQL and a real PostgreSQL instance on Elastic Stack 9.4.0 and 9.4.3: rendered agent config correctly nests `cursor.{enabled,column,type,default}`, the `:cursor` placeholder translates and executes correctly for both drivers, and ingested documents matched expectations.
- [x] Verified the backend's actual startup-validation behavior (single query, response format, placeholder count enforced; `ORDER BY` not enforced) against the elastic/beats#48722 source, and corrected the docs where they previously overstated what's validated.
- [x] Confirmed the new system tests skip cleanly (before touching Docker) on a stack below 9.4, so they won't hang/fail CI.

1. `elastic-package stack up -d --version 9.4.3` (cursor mode requires Stack 9.4+; the package's own CI runs at 9.1.0 where cursor mode is a no-op).
2. `elastic-package build && elastic-package test system -v` from `packages/sql_input` — the `mysql`, `postgresql`, and `mssql` (non-cursor) configs run against real service containers; `mysql-cursor` and `postgresql-cursor` are `skip`ped by default (see reason in each config) but can be run manually by temporarily removing the `skip:` block once the stack is 9.4+.
3. To inspect the rendered agent config directly: enable cursor mode in a policy via Fleet, then `docker exec <agent-container> elastic-agent inspect` and check the `cursor:` block is correctly nested under the `sql` input.

## Related issues

- Requires elastic/beats#48722


stefans-elastic:
I've run a mall local test (with LIMIT 3):

Initial seed (ids 1–10) — the classic LIMIT 3 batching, one batch per 10s poll:
12:45:15  ids 1,2,3      (3)
12:45:25  ids 4,5,6      (3)
12:45:35  ids 7,8,9      (3)
12:45:45  id 10          (1)   ← then it stopped; no re-reading

After adding 5 rows at 12:50 (ids 11–15):
12:50:05  ids 11,12,13   (3)
12:50:15  ids 14,15      (2)   ← then stopped again

shmsr:
@stefans-elastic I added a few more changes. PTAL. Also just one more round of manual testing can you by bring up the stack? Like how the new experience looks like you of configuring the cursor, etc

As this is probably the biggest ever change that goes into SQL and that too of this hard complexity, I think we should test it more.

stefans-elastic:
Hey @shmsr — did the extra round of manual testing you asked for, on a live elastic-package 9.4.0-SNAPSHOT stack with a native 9.4.3 agent, driving the actual Fleet integration (not standalone metricbeat). Tested both MySQL and PostgreSQL end to end.

✅ Manual test results

| Check | MySQL | PostgreSQL |
|---|---|---|
| `escape_string` renders a query with embedded `"quotes"` as valid config | ✅ | ✅ |
| Initial drain 1–10 (`LIMIT 3` → 3→3→3→1, then 0) | ✅ | ✅ |
| Incremental: add rows → fetch **only** the new ones, no re-read, no dupes | ✅ | ✅ |
| `:cursor` → `$1` placeholder translation | n/a | ✅ |
| Cursor state survives package upgrade (resumes, doesn't reset) | ✅ | ✅ |

Your escape_string fix works exactly as intended — the note = has "quotes" here literal round-trips fine, and the old "{{sql_query}}" form would have broken on it.

I also live-reproduced the ORDER BY/LIMIT skip hazard your docs warn about: with a CAST(id AS CHAR) AS id alias and an unqualified ORDER BY id, MySQL sorts the alias lexicographically(1, 10, 2), the cursor jumps to 10, and rows 3–9 are skipped permanently. So the warning is very much warranted — the system tests correctly dodge it by qualifying as <table>.id.

🎨 UX cleanup (pushed to the branch)

While configuring it repeatedly, two things felt confusing, both introduced by this PR:
1. Two near-identically-named fields: "SQL Query (Cursor Mode)" vs "SQL Queries.
2. response_format lived in two places — a per-query key inside SQL Queries, and a separate top-level "SQL Response Format" field.

Since the single-query field is cursor-only, and cursor always requires table, that standalone field was pure redundancy. I made these changes:

- Removed the sql_response_format var entirely. The template now hardcodes sql_response_format: table on the cursor path. This actually supersedes your select-field change with something safer - table becomes impossible to get wrong instead of user-selected-and-validated, and multi-query mode no longer emits a stray sql_response_format: variables. It was added in this same unreleased PR, so removing it breaks nothing.
- Renamed "SQL Query (Cursor Mode)" -> "Cursor Query", reordered so it sits right after Cursor Configuration (the two cursor fields now read as a pair), and rewrote both descriptions to state the mutual exclusivity.
- Updated the two cursor system-test configs (dropped the now-nonexistent var) and the README accordingly.

New config page:

008 - sql-1 - SQL Input - Integrations - Elastic -  127 0 0 1

Re-tested the full flow after upgrading the integration to this build — incremrsor continuity preserved, zero error docs.

shmsr (approved):

PR #17495, opened 2026-02-20
Reorganize okta variables to simplify UI and make agentless the default deployment model

nchaulet:
## Description 

Reorganize okta variables to simplify UI and make agentless the default deployment model

Part of https://github.com/elastic/ingest-dev/issues/6979
Will close https://github.com/elastic/ingest-dev/issues/7534

How it's now rendered in Kibana

image


packages/okta/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.14.2"
+  changes:
+    - description: Reorganize variables, and make agentless the default deployment mode.
+      type: bugfix
```
efd6, line 5: I think this is an enhancement since there is no broken behaviour that is being fixed.
```suggestion
- version: "3.15.0"
  changes:
    - description: Reorganize variables, and make agentless the default deployment mode.
      type: enhancement
```

efd6 (commented):
This is safe; the concern that I do have though is that users who upgrade will no longer see their OAuth2.0 settings because they are now hidden. This may cause them some temporary confusion, though will not be an actual configuration state problem.

nchaulet:
> This is safe; the concern that I do have though is that users who upgrade will no longer see their OAuth2.0 settings because they are now hidden. This may cause them some temporary confusion, though will not be an actual configuration state problem.

@efd6 I am clearly not an expert of the okta integration, the idea was to provide a quickway for user to start and hide as many as possible options to advanced, but if you think oauth2 variables should stay very visible, in may make sense to show them to user, and reorder the variables so they are more visible.

efd6:
@nchaulet Thanks. I think maybe we want a PM's view. @cpascale43, do you have a view on this?

nchaulet:
> @nchaulet Thanks. I think maybe we want a PM's view. @cpascale43, do you have a view on this?

cc @nimarezainia

nimarezainia:
> > @nchaulet Thanks. I think maybe we want a PM's view. @cpascale43, do you have a view on this?
> 
> cc @nimarezainia

@efd6 are there any fields that are compulsory? `Okta System Log API URL` seems to be only one.
We are looking to minimise users having to click around too much and get lost.  Goal is to get them to their data as fast as possible. My assumption is that majority of the users would just provide what is compulsory (such as the  `Okta System Log API URL`) and click save and move on. So we want to optimize for the happy path.

efd6:
@nimarezainia I've taken another look and I have concerns about four of the fields being set to advanced, `okta_domain_url`, `client_id`, `interval` and `initial_interval`.

There is a trade-off here, default options that are not shown are unlikely to be revisited. When users hit problems caused by the defaults (in the case of `interval` and `initial_interval`), they're unlikely to realise there's a setting they could change, and for the `initial_interval` they may end up having missed ingesting data that they wanted because the look-back was not far enough. For `okta_domain_url` and `client_id`, by hiding them, we're making it harder for OAuth2.0 users to discover that we support OAuth2.0.

My concern in https://github.com/elastic/integrations/pull/17495#pullrequestreview-3857306386 still stands, but this is less of an issue than the four fields above.

nimarezainia:
> @nimarezainia I've taken another look and I have concerns about four of the fields being set to advanced, `okta_domain_url`, `client_id`, `interval` and `initial_interval`.
> 
> There is a trade-off here, default options that are not shown are unlikely to be revisited. When users hit problems caused by the defaults (in the case of `interval` and `initial_interval`), they're unlikely to realise there's a setting they could change, and for the `initial_interval` they may end up having missed ingesting data that they wanted because the look-back was not far enough. For `okta_domain_url` and `client_id`, by hiding them, we're making it harder for OAuth2.0 users to discover that we support OAuth2.0.
> 
> My concern in [#17495 (review)](https://github.com/elastic/integrations/pull/17495#pullrequestreview-3857306386) still stands, but this is less of an issue than the four fields above.

thanks @efd6 so let's show these fields as well and not hide them.

I think the order should be to have the compulsory fields first and then the rest. Please correct me

- Okta System Log API URL (compulsory)
- API Key
- interval
- initial interval
- Okta Domain URL
- Client ID

If you believe that all the options should be shown then perhaps we can re-arrange where the "Save & Continue" button could go so the average user doesn't have to sit through all the options.

@nchaulet  that "Change defaults" collapsable needs to be removed also.

efd6:
I think that order looks good. For the others, it was a flag more than anything. If we see that the concern actually comes about, we can revisit.

nimarezainia:
@efd6 My concern is now how do we roll this out to other *simple* integrations in a programmatic fashion. Picking and choosing what fields to expose is very manual and will be different for each integration. Perhaps we are already doing that.

is there any merit to do this:

> If you believe that all the options should be shown then perhaps we can re-arrange where the "Save & Continue" button could go so the average user doesn't have to sit through all the options.

efd6:
If "we can re-arrange where the "Save & Continue" button could go" is an option, sure.

nchaulet:
> @nchaulet that "Change defaults" collapsable needs to be removed also.

I will do a follow up PR with that, so it will not be collapsable (and remove the label) for single input/datastream integration

nchaulet:
@efd6 what it's looks like now 

Screen Shot 2026-03-04 at 14 35 47

nimarezainia:
@efd6 had a discussion with @jamiehynds about this. He had a good suggestion about modifying this further so that the user chooses how they want to authenticate. Just for illustrative purposes I am showing that below. So After they give us the URL, they click on the Authentication method, we show them the relevant fields for that Authentication method and every other config is hidden in the advanced section. For the most part the user won't change the default settings - at least we want to create efficiency in the happy path here for the user:

image


Which fields should be shown for each authentication method?

Supplementing:
Just a heads up, with Nicolas out for the next few weeks on PTO, I will take over as the main point of contact and try to get this one across the line as needed. I've added myself as an additional assignee so I (hopefully) dont miss any notifications, but please tag me if needed! Thanks all.

efd6:
> Which fields should be shown for each authentication method?

- API Key: `api_key` ("API Key")
- OAuth2.0: `okta_domain_url` ("Okta Domain URL"), `client_id` ("Client ID"), `jwk_json`/`key_pem`/`jwk_file` (one only) ("JWK JSON"/"PEM Encoded Key"/"JWK File"), optionally `dpop_key_pem` ("PEM DPoP Key") if using DPoP, and `okta_scopes` ("Okta Scopes")
- OIN App: `okta_domain_url` ("Okta Domain URL"), `client_id` ("Client ID"), and `api_key` ("API Key"), optionally `dpop_key_pem` ("PEM DPoP Key") if using DPoP

(from the [documentation](https://github.com/elastic/integrations/blob/main/packages/okta/_dev/build/docs/README.md#types-of-authentication))

Supplementing:
~~Blocked by https://github.com/elastic/ingest-dev/issues/7185~~

Supplementing:
Here's how it looks now that the Kibana and spec changes have landed/will land to support `var_groups` at the input level. We can now group by chosen auth method: 
image
cc: @nimarezainia 

If youd like to test this, the changes are in Kibana main, and you can use this package: 
[okta-3.14.2.zip](https://github.com/user-attachments/files/26252021/okta-3.14.2.zip)

Let me know if youd like me to spin up a deployment for you!

nimarezainia:
@Supplementing I got a new deployment and added the integration but don't see your changes there. Is there something else I should be doing to enable it?

nimarezainia:
Thanks @Supplementing please see the small comments below:

We need to remove this empty state all together. Users who need to add an agent have he chance to do so later on:
image

I believe that the API url is mandatory for all options. If so can it be moved up to the top before the Authentication method. So that the user enters all the mandatory fields and can leave the optionals to their default value:

image

Remove these extra lines:
image

Granted I put in dummy creds I get this error which I don;t think is related to the creds:
image

Polling interval details are common for all authentication types, can these be in their own section with a header. Just like we have a header for Authentication (see [this comment](https://github.com/elastic/integrations/pull/17495#issuecomment-4018080487)):
image

The other thing missing that was in the original changes by Nicolas, are the step (2) where the user has a back door to change it to Agent-based installation if needed. See the picture in [this comment](https://github.com/elastic/integrations/pull/17495#issuecomment-4005843159).

Supplementing:
> Thanks @Supplementing please see the small comments below:
> 
> We need to remove this empty state all together. Users who need to add an agent have he chance to do so later on: image
> 
> I believe that the API url is mandatory for all options. If so can it be moved up to the top before the Authentication method. So that the user enters all the mandatory fields and can leave the optionals to their default value:
> 
> image
> Remove these extra lines: image
> 
> Granted I put in dummy creds I get this error which I don;t think is related to the creds: image
> 
> Polling interval details are common for all authentication types, can these be in their own section with a header. Just like we have a header for Authentication (see [this comment](https://github.com/elastic/integrations/pull/17495#issuecomment-4018080487)): image
> 
> The other thing missing that was in the original changes by Nicolas, are the step (2) where the user has a back door to change it to Agent-based installation if needed. See the picture in [this comment](https://github.com/elastic/integrations/pull/17495#issuecomment-4005843159).

Alright, I spent a little time working on the manifest seeing what was possible with manifest-only changes vs more kibana changes. 

With the way var_groups work, we cant move the API url above the selector. With the way Kibana renders var_groups, it renders the entire block first, then all the other vars after. We will need Kibana changes for this. 

As for removing the horizontal lines, same story. Its not controlled by the manifest but a side-effect of how Kibana renders sections for inputs. 

Finally, as for adding the header above the polling settings, we will also need Kibana changes. This isnt currently supported via the manifest. I explored using the existing header functionality that is exposed by using var_groups, but var_groups enforces a dropdown selector, which we dont want. We cant use just the title header without adding some dropdown. 

All this being said, I think we can make these changes in one swing in Kibana as just a "visual improvement" PR. I know this is very high priority, so Im happy to carve out a few hours and see if I can get it knocked out if @kpollich agrees. 

As for the missing section and error, you need to use the `enableSimplifiedAgentlessUX: true` feature flag so that it shows up. I think you also need to have custom integrations enabled for agentless if you actually want to submit it, NIcolas talks a bit about it here https://github.com/elastic/kibana/pull/254721#issuecomment-3959647714

Supplementing:
@nimarezainia I spoke with @kpollich a bit on the changes you mentioned and the findings I added above. I have filed a follow up issue [here](https://github.com/elastic/ingest-dev/issues/7320) for polishing the UI, please add anything else you see fit there. In the meantime, if the functionality of the package is in line, I think we should move this PR forward so we can go ahead and hand it off to the security team. The Kibana changes/polish should not require any changes to the package, or very minimal at most (probably just the header if we decide to support those more generically). As far as grouping, etc goes, everything else should remain unchanged. 

We can finalize the UI while the security team decides how they want to structure their integrations. WDYT?

nimarezainia:
thanks @Supplementing - I will look at the UI issue mentioned. What I ultimately need is a way to show others what we are doing with okta integration and get other teams members to review and agree with the direction before we can get that rolled out. 

To that end looks like there are some feature flags that need to be enabled. but I will wait fr you to provide a new integration with all the changes discussed. is that fair?

Supplementing:
> thanks @Supplementing - I will look at the UI issue mentioned. What I ultimately need is a way to show others what we are doing with okta integration and get other teams members to review and agree with the direction before we can get that rolled out.
> 
> To that end looks like there are some feature flags that need to be enabled. but I will wait fr you to provide a new integration with all the changes discussed. is that fair?

I think there's a little confusion, the changes remaining will come from the Kibana side (https://github.com/elastic/ingest-dev/issues/7320), the integration in its current state will remain mostly unchanged and once the kibana changes from that issue land, it will just look as expected if that makes sense. 

The only change that we may need to the integration will be whatever implementation of a header we land on for the polling settings. If you prefer to wait until thats implemented, I understand. The intent was to try to move this PR towards closed if the only thing left was UI polish, but if thats a blocker, let me know. 

The follow up issue was added/will be added to next iteration, so we will be getting started on it shortly. I expect it to move quickly.

nimarezainia:
> > thanks @Supplementing - I will look at the UI issue mentioned. What I ultimately need is a way to show others what we are doing with okta integration and get other teams members to review and agree with the direction before we can get that rolled out.
> > To that end looks like there are some feature flags that need to be enabled. but I will wait fr you to provide a new integration with all the changes discussed. is that fair?
> 
> I think there's a little confusion, the changes remaining will come from the Kibana side ([elastic/ingest-dev#7320](https://github.com/elastic/ingest-dev/issues/7320)), the integration in its current state will remain mostly unchanged and once the kibana changes from that issue land, it will just look as expected if that makes sense.
> 
> The only change that we may need to the integration will be whatever implementation of a header we land on for the polling settings. If you prefer to wait until thats implemented, I understand. The intent was to try to move this PR towards closed if the only thing left was UI polish, but if thats a blocker, let me know.
> 
> The follow up issue was added/will be added to next iteration, so we will be getting started on it shortly. I expect it to move quickly.

@Supplementing thank you. let me know when we can show this to the other teams involved.

Supplementing:
Quick update, Ive opened another Kibana PR [here](https://github.com/elastic/kibana/pull/262129) which adds the section headers, divider toggling, and reorganizing as discussed. I still need to remove the empty state screen @nimarezainia but that is non-blocking and I can do that separately, just wanted to get this done before I go on PTO for the remainder of the week after today. Ive pushed the changes up here to the integration, Ive also uploaded the built integration on that Kibana PR if you prefer to use it directly that way. 

It doesnt block us, but I will also be opening a package-spec PR in a few minutes as it will be needed for integration developers so that as they develop integrations, they will pass validation. I think we are in pretty decent shape overall. If we need to move things around, that should be fairly trivial, everything we need should now be supported. 

Let me know what you think! cc @kpollich 

(adding screenshot here as well just to be safe)

image

Supplementing:
Another update, added [another PR ](https://github.com/elastic/kibana/pull/263552)with some changes to the underlying way we handle rendering the section headers. Also added an updated integration to that PR and pushed up some changes to this PR for testing. Everything should be in sync now and ready for testing. I will also be pushing a commit to that PR in a few minutes that removes the empty state page @nimarezainia

efd6:
Can we make this a draft until it's ready and has a stack it can run on?

Supplementing:
> Can we make this a draft until it's ready and has a stack it can run on?

Done. Ill move it back to ready once everything is finalized and we have a spec release so CI can pass, etc. Sorry for the noise!

nchaulet:
@efd6 here the current state  (I expended everything) let us know if it's good for you or if there is still changes to made

localhost_5601_app_fleet_integrations_okta_add-integration (14)

efd6 (approved):
This LGTM, though I think it probably should have been two PR: one for the UI changes and one for the agentless change requirements.

Suggest the following for the commit message when you squash (edit as needed)
```
okta: reorganize variables for agentless-first UI

Move input variables from the data_stream manifest up to the
package-level manifest and group them with var_groups so the Kibana
UI presents authentication options (API Key, OAuth2.0, OIN App) as
distinct sections. HTTP polling options move to an advanced group.

The default deployment mode is now agentless, reducing configuration
steps for the common case.

Pipeline changes add tag-based routing guards so processors only run
when relevant tags are present.
```

PR #17523, opened 2026-02-23
Update datastream manifest file categories for misclassified security

JDKurma:
Add datastream manifest file categories for those misclassified as security(inherited from package level categories)

Datastream manifest file categories were added to package spec in https://github.com/elastic/package-spec/pull/1065

## Proposed commit message
Categorize commonly misclassified security datastreams explicitly in their respective manifest files.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


donaherc:
@JDKurma is also adding a [package-spec PR here:

https://github.com/elastic/package-spec/pull/1095/changes

to accommodate these changes and validate the yaml files properly at build/publish time.

packages/gitlab/data_stream/sidekiq/manifest.yml
```diff
@@ -1,4 +1,6 @@
 title: GitLab Sidekiq logs
+categories:
+  - productivity
```
trisch-me, line 3: I have just checked one random data_stream 
Gitlab integration has 
```
categories:
  - security
  - productivity_security
```

So it means we can’t just add this category because global categories don’t include it and it’s a top category according to https://github.com/elastic/package-registry/blob/main/categories/categories.yml

Have you made a sanity check for all these cases?

trisch-me (changes requested):
Categories are not synced between integration itself and their data_streams. This needs to be fixed

chrisberkhout (commented):
How does this work?

E.g. for aws_bedrock, the top level manifest has:
```yaml
categories:
  - aws
  - cloud
  - observability
  - security
```
And the PR adds this in the data stream manifests for `guardrails` and `runtime`, but not for `invocation`:
```
categories:
  - cloud
  - aws
  - observability
```
So the top-level categories are right for `invocation` and the others need to override it?

trisch-me:
@chrisberkhout 
the idea of this PR is to align categories of datastreams with security use cases.

i.e. integration defines categories for all their datastreams. And small integrations are usually coherent in their usage, but big integrations such as aws etc might have completely different datastreams, some are for security, some are for o11y etc. If we take integration categories to describe data streams we will get a lot of false positives. I.e. data from o11y data stream will be classified as security one. 

To solve it we have changed the schema to be able to define categories per datastream.

And this PR is fixing the faulty datastreams that were defined as security despite the fact that they have no connection to security.

Now back to your original question

```
And the PR adds this in the data stream manifests for guardrails and runtime, but not for invocation:

categories:
  - cloud
  - aws
  - observability
So the top-level categories are right for invocation and the others need to override it?
```

guadrails and runtime shouldn’t be classified as security data streams therefore they get just the same classification as integration except security tag
but invocation is security related, therefore it’s not changed

trisch-me (approved):
lgtm

chrisberkhout:
Thanks @trisch-me. That all makes sense.

chrisberkhout (approved):

andrzej-stencel (approved):

taylor-swanson (approved):

navnit-elastic (approved):

JDKurma:
@muthu-mps @elastic/obs-infraobs-integrations  Could I get a review for this PR?

ishleenk17 (approved):

JDKurma:
@mergifyio queue

PR #17526, opened 2026-02-23
[otel collector internal telemetry] Split dashboards per event

gregkalapos:
- Split the whole content pack into separate dashboards. With this, added drop-downs like selecting the receiver, the receiver transport, the processor, or the exporter for the given event 
- Adding some more charts
- Renaming the integration to `otel_collector_internal_telemetry` from `otel_internal_telemetry`. The original idea was to keep it general and later add dashboards for SDK telemetry. The reasoning for the rename is that due to `discovery: datasets - name: collectortelemetry.otel`, in reality this will be installed when collector internal telemetry is installed, but this would not work with SDKs, because those metrics won't be in `collectortelemetry.otel`.
- Also bumping Kibana required version for 9.3, because we use variable chaining in ESQL which required 9.3+ (later an 8.x compatible dashboard will be added as well, but that'll be done in a separate  PR).

## Checklist

## Screenshots

Overview:
Screenshot 2026-02-23 at 20 35 18

Traces: 
Screenshot 2026-02-23 at 20 33 35

Metrics:
Screenshot 2026-02-23 at 20 34 00

Logs:
Screenshot 2026-02-23 at 20 34 09


packages/otel_collector_internal_telemetry/manifest.yml
```diff
@@ -1,18 +1,18 @@
 format_version: 3.5.2
-name: otel_internal_telemetry
```
gregkalapos, line 2: Question: Is this ok to rename this way? I basically renamed the folder and updated the name here in `manifest.yml`.

We are still on version 0.0.x and I'm hoping we don't need to go through some deprecation process for the old name.

gregkalapos:
Changed the overview table to this:

Screenshot 2026-03-02 at 17 21 46


So no more `sent_spans`, `refused_spans`, etc - those were 4 columns per event, in sum 12 columns. Now it's reduced just into 2: `Receiver Failure Rate` (Failure across all receivers), and `Exporter Failure Rate` (same for exporter). Also added coloring to it, once Failure Rate rate goes above 1%, it'll be red, otherwise it's ~this lighter color~ the normal black color. Thinking here to use this, instead of the default color pattern is that once these  Failure Rate are more than 0, then this means data loss - so as the middle ground, as long as it's below 1% we keep it less visible, above that it's then red. Added CPU and Memory usage as well. 

cc @AlexanderWert

PR #17540, opened 2026-02-24
o365: fix stale content types persisting in cursor after config changes

chemamartinez:
## Proposed commit message

```
o365: fix stale content types persisting in cursor after config changes

Fix a bug where the O365 audit integration remains DEGRADED after
changing content_types in the configuration, continuing to make API
requests with the removed content types.

When the CEL program generates listing URLs, the content type is embedded
in the URL query string and stored in cursor.todo_links, which is persisted to disk.
After a configuration change and agent restart, these stale links are fetched
before any new work is generated, producing the error:
"AF20020: The specified content type is not valid"
from the API and keeping the integration in a DEGRADED state.

The fix filters cursor.todo_links and cursor.todo_content on every evaluation,
discarding any entries that reference content types no longer present in the configuration.
Additionally, todo_types is now reconciled against the current config on every
evaluation to guard against the same issue mid-cycle without a restart.
```

## Checklist

### DIFF
--- a/packages/o365/data_stream/audit/_dev/test/policy/test-cel-all.expected
+++ b/packages/o365/data_stream/audit/_dev/test/policy/test-cel-all.expected
@@ -92,7 +92,7 @@ inputs:
                     "cursor": state.?cursor.orValue({}).with(
                       {
                         "todo_links": state.?cursor.todo_links.orValue([]).filter(link,
-                          configured_types.exists(ct, link.contains("contentType=" + ct))
+                          configured_types.exists(ct, link.to_lower().contains("contenttype=" + ct.to_lower()))
                         ),
                         "todo_content": state.?cursor.todo_content.orValue([]).filter(item,
                           configured_types.exists(ct, ct.to_lower() == item.?contentType.orValue("").to_lower())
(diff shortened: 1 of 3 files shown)


packages/o365/data_stream/audit/_dev/test/policy/test-cel-all.expected
```diff
@@ -66,17 +66,43 @@ inputs:
...
+              // Filter stale cursor data that references content types no longer in config.
+              // Listing links have the content type embedded in the URL query string, so
+              // links for removed types would produce API errors if not pruned here.
+              state.base.content_types.split(",").map(t, t.trim_space()).as(configured_types,
+                state.with(
+                  {
+                    "cursor": state.?cursor.orValue({}).with(
+                      {
+                        "todo_links": state.?cursor.todo_links.orValue([]).filter(link,
+                          configured_types.exists(ct, link.to_lower().contains("contenttype=" + ct.to_lower()))
```
efd6, line 95: This filter uses a substring match. If a configured content type were a prefix of another (e.g., hypothetical `Audit` matching a link for `Audit.Exchange`), this could retain links that should be pruned. A regex with a word-boundary or delimiter anchor would avoid this.

In practice this isn't a concern: the actual O365 content types (`Audit.AzureActiveDirectory`, `Audit.Exchange`, `Audit.SharePoint`, `Audit.General`, `DLP.All`) are all distinct and none is a prefix of another. A regex match would add per-evaluation cost and CEL complexity for a scenario that doesn't exist today and is unlikely to arise without broader API changes.

Only including this comment as a note to the future.

PR #17552, opened 2026-02-24
Remove ECS field definitions from CDR transform destinations

maxcold:
## Proposed commit message

Removes redundant ECS field definitions from transform destination index field mappings across 7 integration packages, now covered by the `ecs@mappings` component template added to Fleet transform destination index templates (elastic/kibana#223878).

**Packages updated:** wiz, cloud_security_posture, qualys_vmdr, tenable_io, rapid7_insightvm, aws, microsoft_defender_endpoint

**Changes per package:**
- Deleted `ecs.yml` files containing ECS field definitions from transform field directories
- Created `ecs-overridden.yml` where `constant_keyword` type overrides need to be preserved (qualys_vmdr, rapid7_insightvm, aws)
- Minor version bumps and changelog entries

**Note:** The Wiz package also bumps its Kibana version constraint from `~8.16.6 || ~8.17.4 || ^8.18.0 || ^9.0.0` to `^8.19.0 || ^9.1.0` since `ecs@mappings` was introduced in 8.19.0/9.1.0.

## Checklist

## Author's Checklist

- [ ] 

Simplest to use https://github.com/elastic/security-documents-generator command `csp` to generate some data, build packages from this PR, install them and ingest new data. Everything should work as before. Same for first installing the new version packages and then generating data with the data generator

## Related issues

- Relates elastic/kibana#224221


kcreddy (commented):
Thanks for taking care of this.
It would be nice to also rename `ecs.yml` files which has only overridden fields to `ecs-overridden.yml` to conform to this change.
Example: `aws.config`, `aws.inspector`, `aws_securityhub`, `google_scc` etc.

maxcold:
@kcreddy updated

packages/aws/elasticsearch/transform/latest_cdr_misconfigurations/fields/ecs-overridden.yml
```diff
@@ -0,0 +1,7 @@
+# Define ECS constant fields as constant_keyword
+- name: cloud.provider
+  type: constant_keyword
```
kcreddy, line 3: Without `external: ecs` the field description will be omitted from the docs.

Without  `external: ecs` on this field:
```diff
-| cloud.provider | Name of the cloud provider. Example values are aws, azure, gcp, or digitalocean. | constant_keyword |
+| cloud.provider |  | constant_keyword |
```

maxcold: good to know, it was just inconsistent before, `external: ecs` was missing in one of the ecs-overriden.yml files. Will add it back

maxcold: @kcreddy returned it back

kcreddy (approved):
LGTM. 
But would like to get @chrisberkhout thoughts as per his earlier comment - https://github.com/elastic/elastic-package/issues/1641#issuecomment-2963063729

chrisberkhout (approved):
This is great!

Very nice to remove the redundant defs and a potential source of divergence between source and destination indexes.
I see the overrides remain, which is good.

The one other case I can think of to have the explicit definitions is when there's an issue with field limit and it's important to ensure that certain fields are definitely present (e.g. for rules) before dynamic mappings fight for what remains. But that should be rare and only a workaround rather than a standard practice.

Not critical to do now, but it would be good to make sure that for each transform the files in `elasticsearch/transform/*/fields/*.yml` files are identical in name and content to their source data stream's field definition files in `data_stream/*/fields/*.yml`, except `is-transform-source-{false,true}.yml` and anything else that doesn't match for a good reason. When definitions exist in only once place, it's good to have them in a uniquely named file. All this will get easier to keep in sync once we start using the `elastic-package link` function (although it currently has a bug that I'm about to fix).

seanrathier (approved):

Issue #17581, opened 2026-02-26
crowdstrike: fdr pipeline geoip processors lack if conditions causing unnecessary CPU overhead at scale


efd6:
The five geoip processors in the CrowdStrike FDR ingest pipeline run on every document regardless of whether the relevant IP field exists. They rely on `ignore_missing: true` to avoid errors, but the processor is still invoked — including a [database validity check](https://github.com/elastic/elasticsearch/blob/main/modules/ingest-geoip/src/main/java/org/elasticsearch/ingest/geoip/GeoIpProcessor.java) that traverses cluster state metadata on every call.

The processors in question ([`default.yml` lines 1925–1971](https://github.com/elastic/integrations/blob/a186039859c4e4a81b4c2fac4389b281569305be/packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml#L1925-L1971)):

```yaml
  - geoip:
      tag: geoip_observer_ip_into_observer_geo_0729ba64
      field: observer.ip
      target_field: observer.geo
      ignore_missing: true
  - geoip:
      tag: geoip_source_ip_into_source_geo_fcc86651
      field: source.ip
      target_field: source.geo
      first_only: true
      ignore_missing: true
  - geoip:
      tag: geoip_destination_ip_into_destination_geo_ab5e2968
      field: destination.ip
      target_field: destination.geo
      ignore_missing: true
  - geoip:
      tag: geoip_source_ip_into_source_as_56e63fbc
      field: source.ip
      target_field: source.as
      database_file: GeoLite2-ASN.mmdb
      first_only: true
      properties:
        - asn
        - organization_name
      ignore_missing: true
  - geoip:
      tag: geoip_destination_ip_into_destination_as_8a007787
      field: destination.ip
      target_field: destination.as
      database_file: GeoLite2-ASN.mmdb
      properties:
        - asn
        - organization_name
      ignore_missing: true
```

Not all FDR events have IP fields. CrowdStrike FDR covers a wide range of event types (process, file, registry, network, DNS, authentication, etc.), and `source.ip` / `destination.ip` are only populated for network-related events. Telemetry from a large deployment (>100 billion events/day) shows `destination.ip` is present on only ~4% of events, meaning ~95% of geoip invocations for that field are wasted.

[...]

PR #17584, opened 2026-02-26
crowdstrike.fdr: Add if conditions and opt-out config for geoip processors

kcreddy:
## Proposed commit message
```
crowdstrike.fdr: Add if conditions and opt-out config for geoip processors

Avoid unnecessary geoip processor invocations when IP fields are absent.
Without an "if" guard, each geoip processor is still entered on every
document — triggering a database validity check[1] that traverses cluster 
state metadata on every call — before discovering the field is missing 
and returning early.
Adding "if: ctx.<field> != null" skips the processor entirely, avoiding
this overhead for the majority of documents that lack the relevant IP.

Additionally, expose three new configuration options —
enable_geoip_observer_ip, enable_geoip_source_ip, and
enable_geoip_destination_ip — that allow users to selectively disable
GeoIP enrichment per IP field. All default to true for backward
compatibility.

New test sample log copied from existing sample, but with disabled
new configuration options to verify "geo" fields are not populated.

[1] https://github.com/elastic/elasticsearch/blob/70266af5b6e9bd5a40a918cb34fc5d1c3fe680aa/modules/ingest-geoip/src/main/java/org/elasticsearch/ingest/geoip/GeoIpProcessor.java#L103
```

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/17581

### DIFF
--- a/packages/crowdstrike/changelog.yml
+++ b/packages/crowdstrike/changelog.yml
@@ -1,7 +1,10 @@
 # newer versions go on top
 - version: "3.6.0"
   changes:
-    - description: Add conditional execution to geoip processors in the FDR data stream to improve performance.
+    - description: Add if conditions to geoip processors in the FDR data stream to improve performance.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/17584
+    - description: Added new configuration options to the FDR data stream to enable or disable GeoIP enrichment for `observer.ip`, `source.ip`, and `destination.ip` fields.
       type: enhancement
       link: https://github.com/elastic/integrations/pull/17584
 - version: "3.5.0"


packages/crowdstrike/data_stream/fdr/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1924,24 +1924,28 @@ processors:
   # IP Geolocation Lookup.
   - geoip:
       tag: geoip_observer_ip_into_observer_geo_0729ba64
+      if: ctx.observer?.ip != null
```
efd6: Can we make sure that this is also included in the generator that @navnit-elastic is using for his work?

kcreddy: @navnit-elastic LMK where your current generator lies, I can add this change to be included there.

kcreddy:
> I'm not sure that this is a close; there was extension work suggested in the issue to add options to skip the `observer.ip` (and possibly the others) GeoIP look-up entirely. This needs to be discussed to determine whether we think it's a worthwhile thing to do, or whether the change here is enough.

@efd6, in [2f6b0c0](https://github.com/elastic/integrations/pull/17584/commits/2f6b0c0cf193a9f45e05711f679684246104577d), I added the flags which allows for disabling the `geoip` enrichment. PTAL

packages/crowdstrike/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.6.0"
+  changes:
+    - description: Add conditional execution to geoip processors in the FDR data stream to improve performance.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/17584
```
efd6: We need another change for the config change I think.

efd6 (approved):
LGTM after changelog addition.

PR #17602, opened 2026-02-27
abnormal_security: fix threat collection halting on non-200 detail response

### LINKED ISSUE #17590: [abnormal_security.threat] Integration fails when threat ID returns 404 response
We are experiencing a problem at this point in the code (I believe), when the program attempts to retrieve more information about a specific threat ID from the threats API.  The problem occurs when the request for a specific threat ID returns an HTTP "404 Not Found" response instead of the expected payload:

<https://github.com/elastic/integrations/blob/922ee2a3a14bc7501240e076040c20e3983038ad/packages/abnormal_security/data_stream/threat/agent/stream/cel.yml.hbs#L88>

This happens in our tenant every few days.  We need kill and restart the Elastic Agent to restore the feed, and we lose most of the threat event logs that occurred between the start of the fault and the agent restart. We don't know why this is happening: Abnormal may have deleted the threat ID shortly after creation, or perhaps though some internal fault they have a few threat log entries that simply have no detail.  Either way, the CEL code needs to detect this situation and move on efficiently, rather than getting bogged down and failing to report new threat logs.

I am not a CEL expert AT ALL, but my AI agent suggests adding a handler for 404 responses starting at line 88 of the `cel.yml.hbs`.  I think you may want similar handlers for the later enrichment calls to the `/attachments` and `/links` API endpoints.

[This post has been updated... I originally stated that the threat API was returning an empty array as a payload.  That was incorrect.  It is returning an HTTP 404]

efd6:
## Proposed commit message
```
abnormal_security: fix threat collection halting on non-200 detail response

When the /v1/threats/{id} detail endpoint returns a non-200 status
(e.g. 404 for a deleted threat), the CEL program now emits an error
event and advances to the next threat in the worklist. Previously,
it set want_more to false without advancing the worklist index,
causing an infinite retry loop on the same failing threat that
blocked all further collection until the agent was restarted.

Also guard body.messages access in the success branch so a 200
response with a missing or empty messages field does not crash.
```

> [!NOTE]
> Workaround in second commit to be reverted when elastic/elastic-package#3321 is fixed.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #17590

### DIFF
--- a/packages/abnormal_security/data_stream/threat/_dev/test/scripts/threat_detail_404.txt
+++ b/packages/abnormal_security/data_stream/threat/_dev/test/scripts/threat_detail_404.txt
@@ -8,6 +8,14 @@
 use_stack -profile ${CONFIG_PROFILES}/${PROFILE}
 install_agent -profile ${CONFIG_PROFILES}/${PROFILE} -network_name NETWORK_NAME
 docker_up -profile ${CONFIG_PROFILES}/${PROFILE} -network ${NETWORK_NAME} abnormal-security-404-mock
+
+# Workaround for https://github.com/elastic/elastic-package/issues/3321:
+# add_package passes the package dir as RepositoryRoot instead of the repo
+# root, so the builder can't find licenses/Elastic-2.0.txt. Copy it into
+# the package directory where the builder expects it.
+exec mkdir -p ${PACKAGE_ROOT}/licenses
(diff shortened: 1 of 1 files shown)


andrewkroh (approved):
LGTM.

PR #17616, opened 2026-03-02
crowdstrike: Fix Falcon ingest pipeline when `event.original` is a scalar (e.g. HTTP 400)

### LINKED ISSUE #17556: crowdstrike: falcon pipeline crashes on events where crowdstrike.event is not a map
The `logs-crowdstrike.falcon` ingest pipeline fails for any event where `crowdstrike.event` is a scalar (e.g. an Integer) rather than a Map. All such events hit the `on_failure` handler and are indexed as `pipeline_error` documents. A secondary mapping conflict also occurs when `crowdstrike` itself is a scalar.

## Errors

```
Processor "conditional" with tag "convert-nt-timestamp-to-unix" in pipeline
"logs-crowdstrike.falcon-3.5.0" failed with message "dynamic getter
[java.lang.Integer, event] not found"
```

```
object mapping for [crowdstrike] tried to parse field [crowdstrike] as object,
but found a concrete value
```

## Root cause

Three script processors in [`default.yml`](https://github.com/elastic/integrations/blob/18d6f3f6924aea9b6962fcc7fd51819c875faed8/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml) guard with conditions that check for non-null but not for the correct type:

1. [`convert-nt-timestamp-to-unix`](https://github.com/elastic/integrations/blob/18d6f3f6924aea9b6962fcc7fd51819c875faed8/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml#L62-L103) — condition at [L66](https://github.com/elastic/integrations/blob/18d6f3f6924aea9b6962fcc7fd51819c875faed8/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml#L66):
   ```yaml
   if : ctx.crowdstrike?.event != null
   ```
   The script body does `ctx.crowdstrike.event[field]`, which fails when `crowdstrike.event` is an Integer.

2. [`script_remove_null_event`](https://github.com/elastic/integrations/blob/18d6f3f6924aea9b6962fcc7fd51819c875faed8/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml#L182-L195) — same condition pattern. Calls `ctx.crowdstrike.event.entrySet()`, which fails on non-Map types.

3. [`script_remove_null_metadata`](https://github.com/elastic/integrations/blob/18d6f3f6924aea9b6962fcc7fd51819c875faed8/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml#L196-L208) — same pattern for `crowdstrike.metadata`.

The mapping conflict (`crowdstrike` as concrete value) is a separate but related problem: when `event.original` contains a JSON primitive rather than a JSON object, the [JSON processor at L31–L38](https://github.com/elastic/integrations/blob/18d6f3f6924aea9b6962fcc7fd51819c875faed8/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml#L31-L38) sets `crowdstrike` to a scalar, which conflicts with the object mapping at index time.

## Suggested fix

Replace the null checks with type checks:

```yaml
if: 'ctx.crowdstrike?.event instanceof Map'
```

```yaml
if: 'ctx.crowdstrike?.metadata instanceof Map'
```

For the mapping conflict on `crowdstrike` itself, consider adding a `drop` or early exit after the JSON processor when the result is not a Map — these documents cannot be meaningfully processed by the rest of the pipeline.

## Open question

I have not been able to confirm what CrowdStrike streaming API data produces a non-Map `event` field. It would be worth checking sample `event.original` values whether this is a known event type from the API or an edge case in how the input frames messages.

## Affected versions

Present in 3.5.0 (current). The conditions have been unchanged since at least 3.3.0.

navnit-elastic:
## Proposed commit message

```
crowdstrike: avoid pipeline failure when event.original is a scalar

- Add drop processor after JSON parse when crowdstrike is not a Map
  (e.g. HTTP 400 body parsed as integer). Prevents mapping conflicts
  and avoids "dynamic getter [java.lang.Integer, event] not found"
  when later processors access .event on that scalar.
- Use instanceof Map for convert-nt-timestamp-to-unix,
  script_remove_null_event, and script_remove_null_metadata so
  scripts only run when event/metadata are objects.
```

## Checklist

## Author's Checklist

- [ ] 

Pipeline Tests:

```
--- Test results for package: crowdstrike - START ---
╭─────────────┬─────────────┬───────────┬──────────────────────────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE     │ DATA STREAM │ TEST TYPE │ TEST NAME                                                                    │ RESULT │ TIME ELAPSED │
├─────────────┼─────────────┼───────────┼──────────────────────────────────────────────────────────────────────────────┼────────┼──────────────┤
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-event-stream.log)                             │ PASS   │ 395.320833ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-audit-events.log)                      │ PASS   │  407.42525ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-auth-activity.log)                     │ PASS   │   371.4785ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-cspmioa-streaming.log)                 │ PASS   │ 366.688625ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-cspmsearch-streaming.log)              │ PASS   │ 353.600459ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-customer-ioc-event.log)                │ PASS   │ 393.771958ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-data-protection-detection-summary.log) │ PASS   │  374.11675ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-detection-summary.log)                 │ PASS   │ 377.238292ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-epp-detection-summary.log)             │ PASS   │ 378.336458ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-events.log)                            │ PASS   │ 374.217667ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-firewall.log)                          │ PASS   │ 383.649291ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-identity-protection-incident.log)      │ PASS   │  377.42225ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-incident-summary.log)                  │ PASS   │ 376.302834ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-ipd-summary.log)                       │ PASS   │ 393.694917ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-mobile-detection-summary.log)          │ PASS   │ 380.345125ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-non-object-dropped.log)                │ PASS   │ 377.904083ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-recon-notification.log)                │ PASS   │   368.3105ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-remote-response.log)                   │ PASS   │  385.39675ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-sample.log)                            │ PASS   │ 386.655083ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-tags-list.log)                         │ PASS   │ 396.320625ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-tags.log)                              │ PASS   │ 406.976667ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-user-activity.log)                     │ PASS   │ 379.015833ms │
│ crowdstrike │ falcon      │ pipeline  │ (ingest pipeline warnings test-falcon-xdr-detection-summary.log)             │ PASS   │ 419.145875ms │
│ crowdstrike │ falcon      │ pipeline  │ test-event-stream.log                                                        │ PASS   │ 118.742583ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-audit-events.log                                                 │ PASS   │ 109.917083ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-auth-activity.log                                                │ PASS   │  78.440625ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-cspmioa-streaming.log                                            │ PASS   │  76.508417ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-cspmsearch-streaming.log                                         │ PASS   │  74.696583ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-customer-ioc-event.log                                           │ PASS   │ 128.289833ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-data-protection-detection-summary.log                            │ PASS   │  84.108208ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-detection-summary.log                                            │ PASS   │     85.852ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-epp-detection-summary.log                                        │ PASS   │ 277.652959ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-events.log                                                       │ PASS   │  87.039333ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-firewall.log                                                     │ PASS   │  78.802584ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-identity-protection-incident.log                                 │ PASS   │  75.452958ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-incident-summary.log                                             │ PASS   │  71.463583ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-ipd-summary.log                                                  │ PASS   │  73.481417ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-mobile-detection-summary.log                                     │ PASS   │  76.037167ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-non-object-dropped.log                                           │ PASS   │  66.961833ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-recon-notification.log                                           │ PASS   │  70.821542ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-remote-response.log                                              │ PASS   │  76.592792ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-sample.log                                                       │ PASS   │ 101.653708ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-tags-list.log                                                    │ PASS   │  81.891375ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-tags.log                                                         │ PASS   │  70.659792ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-user-activity.log                                                │ PASS   │     72.008ms │
│ crowdstrike │ falcon      │ pipeline  │ test-falcon-xdr-detection-summary.log                                        │ PASS   │  68.560416ms │
╰─────────────┴─────────────┴───────────┴──────────────────────────────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: crowdstrike - END   ---
Done
```

## Related issues

- Closes #17556


packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -63,7 +67,7 @@ processors:
         tag: convert-nt-timestamp-to-unix
         description: Convert Windows NT timestamps to UNIX timestamps for multiple fields.
         lang: painless
-        if : ctx.crowdstrike?.event != null
+        if: 'ctx.crowdstrike?.event instanceof Map'
```
kcreddy: Also add a test where `ctx.crowdstrike?.event` is an Integer as reported in the issue.

navnit-elastic: Thanks for the suggestion.

In the reported issue, the failure comes from **crowdstrike** being an integer (e.g. when `event.original` is a bare 400 from an HTTP 400 response). The JSON processor then sets `crowdstrike = 400`, and the condition `ctx.crowdstrike?.event != null` evaluates `.event` on that integer, which triggers the "dynamic getter [java.lang.Integer, event] not found" error. So the case we’re fixing is "scalar crowdstrike," not "object crowdstrike with integer event."

We don’t have real customer logs where `crowdstrike` is an object and `crowdstrike.event` is an integer. I’d prefer to keep tests aligned with what we’ve actually seen, but I’m happy to add an extra test that mocks `crowdstrike.event` as an integer if you’d still like that case covered.

**Note:** I’ve updated the commit message so it’s consistent with the above.

kcreddy: > We don’t have real customer logs where crowdstrike is an object and crowdstrike.event is an integer. I’d prefer to keep tests aligned with what we’ve actually seen,

I think its important to add this test. If sample is not yet provided in the SDH, you could ask there or mock it.
cc: @efd6

navnit-elastic: I’m happy to add this test case.

Because `crowdstrike` being an object with `crowdstrike.event` as an integer isn’t something we see in real traffic, asking in SDH won’t give us real samples. I’d need help generating a **mocked** sample that still looks realistic so the test is meaningful.

Could you or @efd6 suggest or provide such a mocked event (e.g. minimal JSON that would produce `ctx.crowdstrike.event` as an Integer after the JSON processor)? Once I have that, I’ll add the test.

packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -38,6 +38,10 @@ processors:
             tag: append_error_message_8bb0f55c
             field: error.message
             value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - drop:
+      tag: drop_non_object_crowdstrike
+      description: Drop documents where parsed event.original is not a JSON object (e.g. HTTP status 400). These cannot be meaningfully processed and would cause mapping conflicts.
+      if: 'ctx.crowdstrike != null && !(ctx.crowdstrike instanceof Map)'
```
efd6:
```suggestion
      if: ctx.crowdstrike != null && !(ctx.crowdstrike instanceof Map)
```
(similar below)

packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml
efd6: This is a minimal change that gives the system test that replicates the original issue.
```diff
diff --git a/packages/crowdstrike/data_stream/falcon/_dev/deploy/docker/files/config.yml b/packages/crowdstrike/data_stream/falcon/_dev/deploy/docker/files/config.yml
index 67b2d5cc77..50cd610814 100644
--- a/packages/crowdstrike/data_stream/falcon/_dev/deploy/docker/files/config.yml
+++ b/packages/crowdstrike/data_stream/falcon/_dev/deploy/docker/files/config.yml
@@ -62,6 +62,7 @@ rules:
         body: |-
           {"metadata":{"customerIDString":"abcabcabc22221","offset":1,"eventType":"RemoteResponseSessionStartEvent","eventCreationTime":1698932494000,"version":"1.0"},"event":{"SessionId":"1111-fffff-4bb4-99c1-74c13cfc3e5a","HostnameField":"UKCHUDL00206","UserName":"admin.rose@example.com","StartTimestamp":1698932494,"AgentIdString":"fffffffff33333"}}
           {"metadata":{"customerIDString":"abcabcabc22222","offset":2,"eventType":"RemoteResponseSessionStartEvent","eventCreationTime":1698932494000,"version":"1.0"},"event":{"SessionId":"1111-fffff-4bb4-99c1-74c13cfc3e5a","HostnameField":"UKCHUDL00206","UserName":"admin.rose@example.com","StartTimestamp":1698932494,"AgentIdString":"fffffffff33333"}}
+          400
           {"metadata":{"customerIDString":"abcabcabc22223","offset":3,"eventType":"RemoteResponseSessionStartEvent","eventCreationTime":1698932494000,"version":"1.0"},"event":{"SessionId":"1111-fffff-4bb4-99c1-74c13cfc3e5a","HostnameField":"UKCHUDL00206","UserName":"admin.rose@example.com","StartTimestamp":1698932494,"AgentIdString":"fffffffff33333"}}
 
   - path: /refresh
```
We could have a complete error path test, but it would be largely duplicating the existing system test, so I don't think it's really worth it.

Note that this is not a completely faithful representation of the original error. In the original case, the bare text is `400 Bad Request`. However, if we include that text we need to put it at the end so that we don't contaminate following events, and we end up with a health status degradation. With the input in the state that it's currently in, the change above is the best we can do.

navnit-elastic: Thanks for the suggestion. I’ve applied the change.

efd6 (approved):
Thanks

PR #17620, opened 2026-03-02
ping_federate: strip brackets from IPv6 in audit CEF before decode_cef

navnit-elastic:
## Proposed commit message

```
ping_federate: strip brackets from IPv6 in audit CEF before decode_cef

PingFederate can emit CEF audit logs with IPv6 in brackets
(e.g. src=[2a01:599:110:e652:...]), which violates CEF spec and causes
decode_cef to fail with "value is not a valid IP address".

Add a script processor before decode_cef in TCP, UDP, and filestream
streams to remove brackets around IPv6 addresses so the decoder
receives a valid IP.
```

## Checklist

## Author's Checklist

- [ ] 

System Tests:

```
--- Test results for package: ping_federate - START ---
╭───────────────┬─────────────┬───────────┬───────────┬────────┬───────────────╮
│ PACKAGE       │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT │  TIME ELAPSED │
├───────────────┼─────────────┼───────────┼───────────┼────────┼───────────────┤
│ ping_federate │ admin       │ system    │ logfile   │ PASS   │    41.068725s │
│ ping_federate │ audit       │ system    │ logfile   │ PASS   │ 36.846062666s │
│ ping_federate │ audit       │ system    │ tcp       │ PASS   │ 40.067340083s │
│ ping_federate │ audit       │ system    │ tls       │ PASS   │ 40.995306209s │
│ ping_federate │ audit       │ system    │ udp       │ PASS   │ 41.058248209s │
╰───────────────┴─────────────┴───────────┴───────────┴────────┴───────────────╯
--- Test results for package: ping_federate - END   ---
Done
```

## Related issues

-

### DIFF
--- a/packages/ping_federate/data_stream/audit/agent/stream/udp.yml.hbs
+++ b/packages/ping_federate/data_stream/audit/agent/stream/udp.yml.hbs
@@ -19,6 +19,9 @@ processors:
 - rename:
     fields:
       - {from: "message", to: "event.original"}
+# Workaround: PingFederate can emit CEF with IPv6 in brackets (e.g. src=[2a01:...]).
+# CEF v25 spec defines src as IPv4 only: https://dn790008.ca.archive.org/0/items/commoneventformatv25/CommonEventFormatV25.pdf
+# decode_cef follows the spec and does not accept bracketed IPv6. Strip brackets before decode_cef.
 - script:
     lang: javascript
     source: |
--- a/packages/ping_federate/changelog.yml
+++ b/packages/ping_federate/changelog.yml
@@ -1,8 +1,8 @@
 # newer versions go on top
-- version: "1.0.4"
+- version: "1.2.0"
   changes:
     - description: Support IPv6 source addresses with brackets in audit data stream.
-      type: bugfix
+      type: enhancement
       link: https://github.com/elastic/integrations/pull/17620
 - version: "1.0.3"
   changes:
--- a/packages/ping_federate/data_stream/audit/agent/stream/filestream.yml.hbs
+++ b/packages/ping_federate/data_stream/audit/agent/stream/filestream.yml.hbs
@@ -20,6 +20,9 @@ processors:
 - rename:
     fields:
       - {from: "message", to: "event.original"}
+# Workaround: PingFederate can emit CEF with IPv6 in brackets (e.g. src=[2a01:...]).
+# CEF v25 spec defines src as IPv4 only: https://dn790008.ca.archive.org/0/items/commoneventformatv25/CommonEventFormatV25.pdf
+# decode_cef follows the spec and does not accept bracketed IPv6. Strip brackets before decode_cef.
 - script:
     lang: javascript
     source: |
--- a/packages/ping_federate/data_stream/audit/agent/stream/tcp.yml.hbs
+++ b/packages/ping_federate/data_stream/audit/agent/stream/tcp.yml.hbs
@@ -22,6 +22,9 @@ processors:
 - rename:
     fields:
       - {from: "message", to: "event.original"}
+# Workaround: PingFederate can emit CEF with IPv6 in brackets (e.g. src=[2a01:...]).
+# CEF v25 spec defines src as IPv4 only: https://dn790008.ca.archive.org/0/items/commoneventformatv25/CommonEventFormatV25.pdf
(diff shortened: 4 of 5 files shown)


efd6 (commented):
Can you add a link to documentation that supports the rationale for this change? The CEF parser that we are using does not support IPv6 (the `src` field described in the [v25 spec](https://dn790008.ca.archive.org/0/items/commoneventformatv25/CommonEventFormatV25.pdf) says that it is an IPv4, "Identifies the source that an event refers to in an IP network. The format **is an IPv4 address**. Example: “192.168.10.1”." ISTM that this is a workaround for a format that's not documented to support IPv6 (so it's not a bugfix, at least not in the processor; it is likely a bug in the data source if we are seeing IPv6 data coming through, though note paragraph below regarding spec versions).

If there is a later version of the spec that _does_ support IPv6, we should probably make a change to the processor in beats. We _can_ apply this workaround, but it should be noted as such.

andrewkroh:
Relates: https://github.com/elastic/beats/issues/40269 (for CEF spec updates, mentions IPv6)

navnit-elastic:
> The CEF parser that we are using does not support IPv6 (the src field described in the [v25 spec](https://dn790008.ca.archive.org/0/items/commoneventformatv25/CommonEventFormatV25.pdf) says that it is an IPv4, "Identifies the source that an event refers to in an IP network. The format is an IPv4 address. Example: “192.168.10.1”." ISTM that this is a workaround for a format that's not documented to support IPv6 (so it's not a bugfix, at least not in the processor; it is likely a bug in the data source if we are seeing IPv6 data coming through

The failure is due to **brackets** around the IP, which are not part of the CEF `src` format in the spec.

The change strips those brackets so the parser receives a valid IP string; it’s a format normalization, not an IPv6‑vs‑IPv4 feature.

The CEF v25 “src is IPv4” point is a separate spec detail; if the processor later supports IPv6 per a newer spec, the same “no brackets” rule would still apply.

efd6:
Yes, it looks like Ping Federate had the same misconception as I did; they have misunderstood IPv6 syntax in a URI as being the general syntax. So what I said about IPv6 updates fixing it was wrong, but the comment about the nature of the change is now more correct; this is not a bugfix. The bug is in Ping Federate.

We should probably relax the CEF processor, I'm sure others will have made the same mistake.

packages/ping_federate/data_stream/audit/agent/stream/udp.yml.hbs
```diff
@@ -19,6 +19,20 @@ processors:
...
+# CEF v25 spec defines src as IPv4 only: https://dn790008.ca.archive.org/0/items/commoneventformatv25/CommonEventFormatV25.pdf
+# decode_cef follows the spec and does not accept bracketed IPv6. Strip brackets before decode_cef.
+- script:
+    lang: javascript
+    source: |
+      function process(event) {
+        var msg = event.Get("event.original");
+        if (msg != null) {
+          // Remove brackets from IPv6 addresses to comply with CEF specification
+          msg = msg.replace(/\[([0-9a-fA-F:]+)\]/gi, "$1");
```
efd6, line 32: Just a note that the IPv6 pattern is more complex than this. An indication of this is from the grok patterns [here](https://github.com/elastic/elasticsearch/blob/c995a172480ab881727d5733526bc7fc6b605197/libs/grok/src/main/resources/patterns/ecs-v1/grok-patterns#L28)

navnit-elastic: Thanks for the note.

This step only strips the brackets; it doesn’t validate the contents. Real validation happens in decode_cef, which will still reject invalid IPs.

efd6: Yeah, my concern (minor) was that if there was anything else in the message that is `[hexdigitsandcolons]` then it would be mutated as well even if it were not an IPv6. I think this is reasonably unlikely, but it is less unlikely than something that specifically looks like an IPv6 and isn't an IPv6.

navnit-elastic: Thanks for clarifying.

Agreed - there's a small risk. We will keep this in mind for the future.

Issue #7641, opened 2023-09-03
sei: packages potentially using incorrect mustache snippet escaping


efd6:
The mustache templating system used by ingest pipelines has two levels of escaping available, not escaped (triple stache) and HTML escaped (double stache) — see [man mustache](http://mustache.github.io/mustache.5.html) under "tag types: variables". This can lead to data corruption, particularly in cases where an operating system has chosen to use a character requiring escaping in its path syntax (example [here](https://github.com/elastic/integrations/issues/7627)).

In general we should not be HTML escaping fields for ingestion, so it is nearly always the case that we should be using the triple stache. This is not the case; this is a list of SEI packages that have at least one instance of a double stache in a template snippet:

- [ ] 1password
- [ ] akamai
- [ ] atlassian_bitbucket
- [ ] atlassian_confluence
- [ ] atlassian_jira
- [ ] auditd
- [ ] aws
- [ ] barracuda
- [ ] bitdefender
- [ ] bluecoat
- [ ] carbonblack_edr
- [ ] cef
- [ ] checkpoint
- [ ] cisco_asa
- [ ] cisco_ios
- [ ] cisco_ise
- [ ] cisco_secure_endpoint
- [ ] cisco_umbrella
- [ ] citrix_waf
- [ ] cloudflare
- [ ] cloudflare_logpush
- [ ] crowdstrike
- [ ] cyberark_pta
- [ ] cylance
- [ ] entityanalytics_entra_id
- [ ] f5
- [ ] fireeye
- [ ] forcepoint_web
- [ ] forgerock
- [ ] fortinet_forticlient
- [ ] fortinet_fortiedr
- [ ] fortinet_fortigate
- [ ] gcp
- [ ] github
- [ ] google_workspace
- [ ] hid_bravura_monitor
- [ ] imperva
- [ ] infoblox_nios
- [ ] jumpcloud
- [ ] juniper_junos
- [ ] juniper_netscreen
- [ ] juniper_srx
- [ ] keycloak
- [ ] lyve_cloud
- [x] m365_defender #7522 #7707
- [ ] mattermost
- [ ] microsoft_defender_endpoint
- [ ] microsoft_dhcp
- [ ] microsoft_sqlserver
- [ ] mimecast
- [ ] modsecurity
- [ ] mysql_enterprise
- [ ] nagios_xi
- [ ] netflow
- [ ] netscout
- [ ] netskope
- [ ] network_traffic
- [ ] o365
- [ ] okta
[...]

PR #17623, opened 2026-03-02
ssi_all: use triple-brace templating in complex templates - Part 2

kcreddy:
## Proposed commit message
```
ssi_all: use triple-brace Mustache templating in ingest pipelines

Fix remaining double-brace Mustache template variables in ingest
pipelines for packages owned by elastic/security-service-integrations.

This is a follow-up to #11314 which addressed most packages but missed
some. Affected packages: admin_by_request_epm,
aws_securityhub, beyondinsight_password_safe, bitsight, cloudflare,
cloudflare_logpush, google_workspace, m365_defender,
microsoft_exchange_online_message_trace, rapid7_insightvm,
sentinel_one, tenable_ot_security, ti_cif3.

Relates: #7641

[git-generate]
for f in $(
	(
		for p in $(
			for f in packages/*/manifest.yml; do
				if yq -e 'select(.owner.github == "elastic/security-service-integrations")' "$f" >/dev/null 2>&1; then
					dirname "$f"
				fi
			done
		); do
			rg -l '(^|[^{])\{\{[^{][ .a-zA-Z0-9_]*[^}]\}\}($|[^}])' -g '*.yml' "$p" || true
		done
	)|grep "elasticsearch/ingest_pipeline"|sort|uniq
); do
	perl -pi -e 's/(?<!\{)(\{\{[^{][ .a-zA-Z0-9_]*[^}]}})(?!\})/{$1}/g' $f
done
for p in $(git diff --name-only HEAD~1|cut -d/ -f1,2|sort|uniq); do
	(
		cd $p
		elastic-package changelog add \
			--description "Use triple-brace Mustache templating when referencing variables in ingest pipelines." \
			--type bugfix \
			--next patch \
			--link https://github.com/elastic/integrations/pull/17623
	)>/dev/null 2>&1
done
```

## Checklist

## Related issues

- Relates https://github.com/elastic/integrations/issues/7641
- Closes https://github.com/elastic/integrations/issues/17673


kcreddy:
> When I squash the four commits into one and give it the commit message that's proposed (with the PR number added), we don't get exactly what is here. There are some inconsequential YAML and whitespace changes, but there is also a reversion of the ti_cif3 changes. Can you check that?
> 
> In cases where git-generate is used, I'd recommend keeping the PR as a single commit so that the state in the PR is essentially as it will be once it's merged.

@efd6, the problem was `rg` wasn't identifying `ti_cif3` file because the original pattern uses `[^}]` after `}}` which requires a character to exist, missing cases where `}}` is at the end of a line. I updated the `rg` regex to match with perl regex to support lookarounds. Now `ti_cif3` is included.
I also retained the whitespace and yaml formatting changes that are being done by `elastic-package test pipeline` and `elastic-package changelog add` and squashed into single commit.

PR #17625, opened 2026-03-02
o365: add script tests for audit data stream

efd6:
## Proposed commit message
```
o365: add script tests for audit data stream

Add four elastic-package script tests covering the CEL-based audit
data stream:

* env: verify that script test environment variables are populated
  correctly.
* invalid_content_type: configure a content type the mock API
  rejects with 403 and verify exactly one error event is emitted
  with no data events.
* subscription_permission_error: verify that a 401 permission error
  on one content type does not prevent collection from the remaining
  types. Migrates the intent of the skipped system test
  test-cel-bad-creds-config.yml.

The tests include two temporary workarounds for bugs in
elastic-package that can be removed once fixes are released:

* elastic/elastic-package#3321: the script test runner passes the
  package directory as RepositoryRoot, so the builder cannot find
  the repository license file. The workaround copies the license
  into the package directory before add_package and removes it
  during cleanup.
* elastic/elastic-package#3303: remove_package fails because
  Fleet's cascade deletion of package policies is asynchronous and
  not reliably complete before the command runs. The workaround is
  to omit remove_package from cleanup and rely on the test runner's
  automatic teardown, which silently handles the failure.
```

> [!NOTE]
> No user-facing change, so no changelog entry.

> [!NOTE]
> The original state of this PR included an update test, but this was incorrect and depends on changes to elastic-package. So it has been backed out until the required infrastructure is added.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/o365/data_stream/audit/_dev/test/scripts/env.txt
```diff
@@ -0,0 +1,28 @@
...
+stdout '/packages/o365$'
+
+exec echo ${DATA_STREAM}
+stdout '^audit$'
+
+exec echo ${DATA_STREAM_ROOT}
+stdout '/packages/o365/data_stream/audit$'
+
+exec echo ${CURRENT_VERSION}
+stdout '^[0-9]+\.[0-9]+\.[0-9]+$'
```
kcreddy: Do you think we should match with [semver regex pattern](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) so that patterns ending with say: `-next` are included?

packages/o365/data_stream/audit/_dev/test/scripts/invalid_content_type.txt
```diff
@@ -0,0 +1,125 @@
...
+# the tenant is not authorised to access.
+#
+# With the only content type rejected, the program should emit one error event
+# and stop (want_more becomes false).
+
+[!external_stack] skip 'Skipping external stack test.'
+[!exec:jq] skip 'Skipping test requiring absent jq command'
+
+# Connect to the running stack.
+use_stack -profile ${CONFIG_PROFILES}/${PROFILE}
```
kcreddy, line 16: `${PROFILE}` is not checked in `env.txt`, maybe we should add it.

packages/o365/data_stream/audit/_dev/test/scripts/invalid_content_type.txt
```diff
@@ -0,0 +1,125 @@
...
+vars: ~
+data_stream:
+  vars:
+    url: http://o365-mock:8080
+    token_url: http://o365-mock:8080
+    preserve_original_event: true
+    client_id: test-cel-client-id
+    client_secret: test-cel-client-secret
+    azure_tenant_id: test-cel-tenant-id
+    content_types: "Audit.Nonexistent"
```
kcreddy, line 69: What happens if one of the valid content types are included?
say: `content_types: "Audit.SharePoint, Audit.Nonexistent"`

packages/o365/data_stream/audit/_dev/test/scripts/invalid_content_type.txt
```diff
@@ -0,0 +1,125 @@
...
+
+# Clean up.
+remove_package_policy -profile ${CONFIG_PROFILES}/${PROFILE} ${DATA_STREAM_NAME}
+uninstall_agent -profile ${CONFIG_PROFILES}/${PROFILE} -timeout 1m
+docker_down o365-mock
+
+# Clean up license workaround (https://github.com/elastic/elastic-package/issues/3321).
+exec rm -rf ${PACKAGE_ROOT}/licenses
+
+-- test_config.yaml --
```
kcreddy, line 58: Is it requirement to embed `test_config.yaml`, `<mock>/docker-compose.yml`, and `<mock>/config.yml` into the script test file or can they be defined in separate files? It is more of a clarification on script testing capability - the overall file size may be reasonably smaller in most cases.

packages/o365/data_stream/audit/_dev/test/scripts/invalid_content_type.txt
```diff
@@ -0,0 +1,125 @@
...
+# root, so the builder can't find licenses/Elastic-2.0.txt. Copy it into
+# the package directory where the builder expects it.
+exec mkdir -p ${PACKAGE_ROOT}/licenses
+exec cp ${PACKAGE_ROOT}/../../licenses/Elastic-2.0.txt ${PACKAGE_ROOT}/licenses/Elastic-2.0.txt
+
+# Add the package resources.
+add_package -profile ${CONFIG_PROFILES}/${PROFILE}
+
+# Add the data stream policy with a single invalid content type.
+add_package_policy -profile ${CONFIG_PROFILES}/${PROFILE} test_config.yaml DATA_STREAM_NAME
```
kcreddy, line 35: Should there be a check that `${DATA_STREAM}` defined in `env.txt` matches the `DATA_STREAM_NAME`?

packages/o365/data_stream/audit/_dev/test/scripts/upgrade.txt
```diff
@@ -0,0 +1,212 @@
+# Test that upgrading the package to the latest version does not break
+# data collection.
+#
+# Flow:
+#   1. Install the current development version.
+#   2. Create a policy and verify data collection works.
+#   3. Run upgrade_package_latest.
+#   4. Verify data collection still works after the upgrade.
```
kcreddy: Just trying to get clarification here.
Are we upgrading from `current latest version` (latest published in EPR) to `current development version` (to be published into EPR) 🤔 ? 
I am curious why install the development version first, if thats the version we wanted to test the upgrade into.

kcreddy (commented):
The script test capabilities are amazing 🎉 and thanks for adding and giving many examples 😄 
Just trying to get some clarifications through the review.

packages/o365/data_stream/audit/_dev/test/scripts/env.txt
efd6: Probably.

packages/o365/data_stream/audit/_dev/test/scripts/invalid_content_type.txt
efd6: I don't think it's necessary. The env.txt is really demonstration only. $PROFILE is always set.

packages/o365/data_stream/audit/_dev/test/scripts/invalid_content_type.txt
efd6: It _is_ possible to use configs from other locations, but it is better to use the locally defined set-ups to keep locality.

packages/o365/data_stream/audit/_dev/test/scripts/invalid_content_type.txt
efd6: I don't think that's necessary, and is formally not a correct assertion; we just happen to have them be the same nearly always.

packages/o365/data_stream/audit/_dev/test/scripts/upgrade.txt
efd6: Good catch. Investigating this, we need some changes to the script testing to get this to work correctly. I'll file an issue to add the needed functionality and back this out until we have that in place.

issue: elastic/elastic-package#3334

packages/o365/data_stream/audit/_dev/test/scripts/invalid_content_type.txt
efd6: This is covered by subscription_permission_error.txt. It configures "Audit.SharePoint, Audit.TypeRequiringAdditionalPermissions, Audit.General" where the middle type gets a 401 while the other two succeed.

kcreddy:
I think we should delete `test-cel-bad-creds-config.yml` now that we replicated that scenario into script tests.

PR #17633, opened 2026-03-03
[checkpoint] Move last update time

ash-darin:
Enhancement

## Proposed commit message

Events may include last_hit_time *and* lastupdatetime.

Lastupdatetime is in all observed cases later than last_hit_time and should therefore mark the end of the event. The current order of processors parses last_hit_time as event.end after lastupdatetime though, therefore overwriting the date parsed from lastupdatetime.

This PR changes this order and additionally removes the attempt to parse the date as ISO, as the dates of last_hit_time and lastupdatetime are always UNIX Timestamps.

This [documentation](https://sc1.checkpoint.com/documents/R81.20/WebAdminGuides/EN/CP_R81.20_LoggingAndMonitoring_AdminGuide/CP_R81.20_LoggingAndMonitoring_AdminGuide.pdf)
 describes "lastupdatetime" as an "integer".

## Checklist

Checking this is problematic, as the sourcefields are both discarded.


ash-darin (commented):
Resolved conflict of versioning with other approved patch.

packages/checkpoint/changelog.yml
```diff
@@ -1,6 +1,10 @@
 # newer versions go on top
-- version: "1.45.4"
+- version: "1.45.5"
   changes:
+    - description: Improve event.end determination
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/17633
+- version: "1.45.4"
```
andrewkroh, line 7:
```suggestion
- version: "1.45.4"
  changes:
```

Changelog is malformed at the moment.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1362,18 +1362,18 @@ processors:
       field: checkpoint._temp_unixms
       target_field: "@timestamp"
       ignore_missing: true
-  - date:
-      tag: date_checkpoint_lastupdatetime_to_event_end_869e28e8
-      field: checkpoint.lastupdatetime
-      formats: ["ISO8601", "UNIX"]
```
kgeller, line 1368: Is dropping the `ISO8601` format for both of these processors intentional?

ash-darin: Yes it is intentional. I mentioned the reasoning in the proposed commit message:

> This PR changes this order and additionally removes the attempt to parse the date as ISO, as the dates of last_hit_time and lastupdatetime are always UNIX Timestamps.
> 
> This [documentation](https://sc1.checkpoint.com/documents/R81.20/WebAdminGuides/EN/CP_R81.20_LoggingAndMonitoring_AdminGuide/CP_R81.20_LoggingAndMonitoring_AdminGuide.pdf) 
> describes "lastupdatetime" as an "integer".

If you think this problematic, I would at least propose that "UNIX" is tried first. Trying ISO first and failing reliably is unnecessary overhead in my mind.

ash-darin:
Apparently I can not trigger tests. 

@kgeller  I forgot to update the manifest, sorry about that. I now rectified it.

kgeller:
> Apparently I can not trigger tests.
> 
> @kgeller I forgot to update the manifest, sorry about that. I now rectified it.

Thanks for fixing up the manifest! 

Looks like now you're running into actual test failures. I'd suggest running those locally and ensuring ✅ and then I can get CI to go green as well. The `elastic-package test` command has some pretty thorough docs https://github.com/elastic/elastic-package/tree/main?tab=readme-ov-file#elastic-package-test

packages/checkpoint/data_stream/firewall/fields/fields.yml
```diff
@@ -49,6 +49,10 @@
       type: keyword
       description: |
         Anti virus type.
+    - name: app_category
```
ash-darin, line 52: Field was previously undefined, but test suite complains:

`field "checkpoint.app_category" is undefined`

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -844,6 +844,11 @@ processors:
       field: checkpoint.last_detection
       target_field: event.end
       ignore_missing: true
+  - convert:
```
ash-darin, line 847: Needs to be float to be accepted by test suite. This apparently was never noted as all test cases so far omitted this field. 

`field "event.risk_score"'s Go type, string, does not match the expected field type: float`

packages/checkpoint/docs/README.md
```diff
@@ -181,6 +181,7 @@ The `firewall` data stream provides events from Check Point devices, including f
 | checkpoint.analyzed_on | Check Point ThreatCloud / emulator name. | keyword |
 | checkpoint.answer_rdata | List of answer resource records to the questioned domains. | keyword |
 | checkpoint.anti_virus_type | Anti virus type. | keyword |
+| checkpoint.app_category | Application category. | keyword |
```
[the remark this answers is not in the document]
kgeller: There is no duplicate

kgeller (approved):
LGTM! 

@ash-darin thanks for doing that additional cleanup of the field addition and convert processor!

PR #17651, opened 2026-03-04
Update control group reference in dashboard JSON

ishleenk17:
## Proposed commit message
Update the control group reference in postgreSQL dashboard


ishleenk17:
> Any idea why the control group name was incorrect during export?

We are figuring that out [here](https://github.com/elastic/obs-integration-team/issues/888).

PR #17656, opened 2026-03-04
Fix issues found while migrating to simplified API

jsoriano:
We are evaluating the migration of elastic-package to the "simplified" Package Policies API in Fleet (see https://github.com/elastic/elastic-package/issues/3038 and https://github.com/elastic/elastic-package/pull/3307).

This API brings benefits like improved validation and better handling of defaults and integration with Fleet functionality.

With the improved, stricter, validation we have found some issues in test configurations, mostly about required variables that seem to be incorrectly set, or not set at all.

This change attempts to fix the discovered issues. 

## Related issues

- https://github.com/elastic/elastic-package/issues/3038
- https://github.com/elastic/elastic-package/pull/3307
- https://github.com/elastic/kibana/issues/255976

### DIFF
--- a/packages/ti_opencti/data_stream/indicator/_dev/test/system/test-default-config.yml
+++ b/packages/ti_opencti/data_stream/indicator/_dev/test/system/test-default-config.yml
@@ -1,5 +1,10 @@
 input: cel
 service: opencti_stub
+# The simplified Fleet API coerces string "false"/"true" to booleans before
+# reaching the select-type validation, so "false" fails include() against the
+# string options. Force the legacy API until Fleet fixes the schema ordering.
+# See https://github.com/elastic/kibana/issues/255976
+policy_api_format: legacy
 vars:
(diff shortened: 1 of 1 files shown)


packages/ti_opencti/data_stream/indicator/_dev/test/system/test-default-config.yml
```diff
@@ -1,5 +1,10 @@
 input: cel
 service: opencti_stub
+# The simplified Fleet API coerces string "false"/"true" to booleans before
+# reaching the select-type validation, so "false" fails include() against the
+# string options. Force the legacy API until Fleet fixes the schema ordering.
+# See https://github.com/elastic/kibana/issues/255976
+policy_api_format: legacy
```
jsoriano, line 7: Support for this added in https://github.com/elastic/package-spec/pull/1103 and https://github.com/elastic/elastic-package/pull/3307.

Issue #17628, opened 2026-03-02
[tenable_ot_security.assets] Some assets are missing


StacieClark-Elastic:
### Integration Name

Tenable OT Security [tenable_ot_security]

### Dataset Name

assets

### Integration Version

1.0.0

### Agent Version

NA

### Agent Output Type

elasticsearch

### Elasticsearch Version

NA

### OS Version and Architecture

NA

### Software/API Version

NA

### Error Message

Request tracing has at least one request where the response body in truncated. This request has a 200 status code.
The cel execution loop has been seen to hit it's maximum. It's unclear whether this is related to the truncated response body.

### Event Original

NA

### What did you do?

Analysis of code and testing show:

- the getAssets graphql query is saving the pagination cursor between periodic runs (intervals). Pagination cursors are not valid when the underlying datastore has been updated. The pagination becomes invalid and subsequent queries stop returning any data.

### What did you see?

Assets that exist on the tenable server do not show up in the elastic dashboard.

### What did you expect to see?

All assets on the tenable server have at least one document in elasticsearch,

StacieClark-Elastic:
possible solutions to this issue: (in order of least likely to most likely to work). The issues we need to address are:
1. Fetching all assets that are returned by the query.
2. Being able to restart an "in progress" query when the maximum number of cel executions is met.
3. Address performance when assets have not been changed since the last query and do not need to be re-fetched.

Solutions considered:
1. Continue to use the getAsset query, but set the pagination cursor to null when all data has been collected. Use a fingerprint on id, firstSeen and lastUpdate to only ingest assets that are "new" or have changes.
    - Does not satisfy criteria '2' . No guarantee that the pagination cursor will be valid on the next interval.
    - Doe not satisfy 3 well as all assets would be fetched on each interval. And if all assets weren't fetched due to max executions being met, it could several intervals to fetch all the assets.
2. Continue to use the getAsset query, sort the responses on some key like 'id', then store the last "id" in the cursor between loops. Use the pagination cursor only within an interval. The id and the pagination cursor would be cleared when hasNextPage = false. If max executions are hit, we clear the pagination cursor. On the next interval if the "id" exists but not pagination cursor exists, we add the "id" as a a filter. 
    - Does not satisfy 3 well as all assets would be fetched on each interval. And if all assets weren't fetched due to max executions being met, it could several intervals to fetch all the assets.
[...]

PR #17669, opened 2026-03-04
tenable_ot_security: fix assets data stream stale cursor and timestamp

efd6:
## Proposed commit message
```
tenable_ot_security: fix assets data stream stale cursor and timestamp

Switch the assets data stream from cursor-based GraphQL pagination to
time-based incremental fetching, following the pattern already used by
the events stream. The old approach stored an opaque GraphQL pagination
cursor that went stale after max_executions interruptions, causing the
API to return empty results indefinitely.

Change @timestamp from firstSeen to ingest time so assets appear in
recent dashboard time ranges. Guard the max_executions template
variable across all three data streams to prevent emitting an empty
YAML value.

The system test mock was rewritten to match the new request structure.
Assets are sorted by lastUpdate ascending across three pages of three.
Each page's request is matched by the lastUpdate value from the
previous page's last asset: page 1 matches a regex timestamp (the
initial fetch), pages 2 and 3 match exact timestamps. The test config
sets initial_interval to 87600h so the mock's hardcoded timestamps
from 2024-2025 fall within the fetch window.

Upgrade note: a follow-up release will change the document fingerprint
from a full-body hash to an id plus timestamps hash. To avoid duplicate
asset documents, upgrade to this version first and allow at least one
successful asset collection interval before upgrading to the
fingerprint release. Do not skip this version.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- For #17628

### DIFF
--- a/packages/tenable_ot_security/changelog.yml
+++ b/packages/tenable_ot_security/changelog.yml
@@ -7,7 +7,7 @@
         appear in recent dashboard time ranges. Guard max_executions template variable
         across all data streams.
         Upgrade note: a follow-up release will change the document fingerprint from
-        a full-body hash to an id-only hash. To avoid duplicate asset documents,
+        a full-body hash to an id plus timestamps hash. To avoid duplicate asset documents,
         upgrade to this version first and allow at least one successful asset collection
         interval before upgrading to the fingerprint release. Do not skip this version.
       type: bugfix
--- a/packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
+++ b/packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
@@ -71,7 +71,10 @@ program: |-
                 }
               ),
               "cursor": {
-                "last_timestamp": (size(nodes) > 0) ? nodes[size(nodes) - 1].lastUpdate : state.?cursor.last_timestamp,
+                "last_timestamp": (size(nodes) > 0) ?
+                  nodes[size(nodes) - 1].lastUpdate 
+                :
+                  state.?cursor.last_timestamp.orValue(timeframe.startTime),
               },
               "want_more": body.data.assets.pageInfo.hasNextPage,
             }
@@ -128,7 +131,6 @@ state:
         }
         pageInfo {
           hasNextPage
-          endCursor
         }
       }
     }


packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
```diff
@@ -31,51 +33,62 @@ processors:
...
+          "variables": {
+            "first": state.limit,
+            "filter": {
+              "op": "Or",
+              "expressions": [
+                {"field": "lastUpdate", "op": "GreaterEqual", "values": timeframe.startTime},
+                {"field": "firstSeen", "op": "GreaterEqual", "values": timeframe.startTime},
+              ]
+            },
+            "sort": [{"field": "lastUpdate", "direction": "AscNullFirst"}],
```
kcreddy, line 58: In the case of `AscNullFirst`, say the first page has all-nulls. What happens in that case?
I'm curious if `startTime` resolves to `null` or resets to initial interval again with `.orValue(`

packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
```diff
@@ -31,51 +33,62 @@ processors:
...
-  want_more: false
   key: {{key}}
   limit: {{limit}}
+  initial_interval: {{initial_interval}}
   preserve_original_event: {{preserve_original_event}}
   query: >-
-    query getAssets($first: Int, $after: String) {
-      assets(first: $first, after: $after) {
+    query getAssets($first: Int, $filter: AssetExpressionsParams, $sort: [AssetSortParams!]) {
+      assets(first: $first, filter: $filter, sort: $sort) {
```
kcreddy, line 94: Inside the query, we are still fetching `endCursor` which is not used in this new time-based incremental fetch (no pagination). 

https://github.com/elastic/integrations/blob/35e7cac4fa0d3aea5fc15d309bf689015c7de808/packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs#L118

Should we just delete it since its not required?

packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
```diff
@@ -31,51 +33,62 @@ processors:
...
-  							?"event.original": state.?preserve_original_event.orValue(false) ? optional.of(e.encode_json()) : optional.none(),
-  						}
-  					),
-  					"cursor": {"end_cursor": (size(nodes) > 0) ? body.data.assets.pageInfo.endCursor : state.?cursor.end_cursor},
-  					"want_more": body.data.assets.pageInfo.hasNextPage,
-  				}
-  			)
-  		)
-  	)
+  {"startTime": state.?cursor.last_timestamp.orValue(timestamp(now() - duration(state.initial_interval)).format(time_layout.RFC3339))}.as(timeframe,
```
kcreddy, line 36: Since the first fetch starts with `initial_interval`, it can pull previous data causing duplicates (since fingerprint is not yet implemented). Should we categorise this as a breaking-change?

packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
```diff
@@ -31,51 +33,62 @@ processors:
...
+        },
+        "Body": {
+          "query": state.query,
+          "variables": {
+            "first": state.limit,
+            "filter": {
+              "op": "Or",
+              "expressions": [
+                {"field": "lastUpdate", "op": "GreaterEqual", "values": timeframe.startTime},
+                {"field": "firstSeen", "op": "GreaterEqual", "values": timeframe.startTime},
```
kcreddy, line 55: Just a clarification - do we know when an asset is created, only the `firstSeen` is populated and not the `lastUpdate`?

packages/tenable_ot_security/changelog.yml
```diff
@@ -1,3 +1,17 @@
...
+    - description: >-
+        Fix assets data stream to use time-based incremental fetching instead of
+        cursor-based pagination, preventing stale cursor errors after max_executions
+        interruptions. Change @timestamp from firstSeen to ingest time so assets
+        appear in recent dashboard time ranges. Guard max_executions template variable
+        across all data streams.
+        Upgrade note: a follow-up release will change the document fingerprint from
+        a full-body hash to an id-only hash. To avoid duplicate asset documents,
+        upgrade to this version first and allow at least one successful asset collection
+        interval before upgrading to the fingerprint release. Do not skip this version.
```
kcreddy, line 12: Do you think it makes sense to add a support knowledge base article on this?

packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
efd6: I've added a `!= null` guard on the cursor update plus an `.orValue(timeframe.startTime)` fallback on the else branch, so the cursor always holds a valid timestamp.

[`AscNullFirst`](https://docs.tenable.com/OT-security/api/sortdirection.doc.html) puts nulls before non-null values, so once we get past the null `lastUpdate` assets, the cursor will get a real timestamp from the first non-null node.

There is an issue that if every asset on a page has null `lastUpdate`, the cursor doesn't advance. With `want_more: true` this re-fetches the same page until `max_executions` budget depletion. Across executions the cursor falls back to `initial_interval` and the cycle repeats. Fixing this properly would need a hybrid approach, e.g. keeping `endCursor` for within-execution pagination alongside the time-based cross-execution cursor, which is a bigger change. Do you think it's worth addressing that now or in a follow-up?

packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
efd6: Removed (it's also in events and system_log where it's already not used, but I've left those in and we can clean up later).

packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
efd6: This doesn't change the fingerprint; [it's still a full hash of `tenable_ot_security.assets`](https://github.com/elastic/integrations/blob/9e7883beeed299079ed5b6bb986c07b945a4fcf3/packages/tenable_ot_security/data_stream/assets/elasticsearch/ingest_pipeline/default.yml#L11). After upgrade, `initial_interval` re-fetches previously collected assets, but if an asset's body is identical the hash and `_id` are the same and the doc is overwritten (no duplicate). If a mutable field like `lastSeen` changed, the hash differs. This is existing behaviour on every normal collection.

It does break in the #17670 follow-up, which changes the `_id` scheme from full hash to id + timestamps. That guarantees duplicates for every existing asset and is labelled as a breaking change.

packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
efd6: I wasn't able to find documentation on whether `lastUpdate` is populated at creation time. The `Or` filter is defensive; if an asset is created but not yet updated, it would only match on `firstSeen`. If `lastUpdate` is always set on creation, the `firstSeen` arm is redundant but harmless. I'd rather keep it than risk missing newly created assets. Have you seen any API docs or testing results that would clarify?

packages/tenable_ot_security/changelog.yml
efd6: Probably. When do we want to do this? Before or after merging the follow-up PR? The KB would basically be the text in the descriptions from the two PRs.

kcreddy: It makes sense to create it after the second PR is merged 👍🏼

packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
kcreddy: I agree, we should go forward with current approach and revisit this as a followup.

packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
kcreddy: Thanks for the clarification.

packages/tenable_ot_security/data_stream/assets/agent/stream/cel.yml.hbs
kcreddy: Couldn't confirm myself from the docs. Agreed on the harmless redundancy.

PR #17670, opened 2026-03-04
tenable_ot_security: change assets fingerprint to id plus timestamps

### LINKED ISSUE #17628: [tenable_ot_security.assets] Some assets are missing

efd6:
## Proposed commit message
```
tenable_ot_security: change assets fingerprint to id plus timestamps

Change the assets document fingerprint from a full-body hash to a hash
of the asset id and all timestamp fields (firstSeen, lastSeen, lastHit,
lastSnapshot, lastUpdate, runStatusTime). This preserves update history;
a new document is created whenever a timestamp changes without the
duplicate-on-every-poll problem caused by the full-body hash.

Change the risk.total_risk field type from keyword to double. The field
carries a floating-point risk score and double is the correct type.
This type change also causes Fleet to trigger an automatic data stream
rollover on upgrade: the putMapping call fails with an
illegal_argument_exception when it tries to change the field type in
the existing backing index, and Fleet responds by rolling over. The
rollover starts the life-cycle clock on the old backing index so
duplicate documents from the _id scheme change age out without manual
intervention.

The Kibana version constraint is tightened to ^8.16.1 || ^9.1.4 so
that 9.x users see the breaking-change call-out. Where possible,
upgrading to 1.1.0 first and allowing at least one asset collection
interval before upgrading to 2.0.0 avoids the overlap period entirely.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Depends #17669 (not to be merged before and only to be merged after reasonable interval)
- Closes #17628

### DIFF
--- a/packages/microsoft_exchange_online_message_trace/manifest.yml
+++ b/packages/microsoft_exchange_online_message_trace/manifest.yml
@@ -9,7 +9,7 @@ categories:
   - email_security
 conditions:
   kibana:
-    version: "^8.19.4 || ^9.1.4"
+    version: "^8.19.4 || ~9.0.7 || ^9.1.4"
   elastic:
     subscription: "basic"
 icons:
--- a/packages/microsoft_exchange_online_message_trace/manifest.yml
+++ b/packages/microsoft_exchange_online_message_trace/manifest.yml
@@ -9,7 +9,7 @@ categories:
   - email_security
 conditions:
   kibana:
-    version: "^8.19.4 || ~9.0.7 || ^9.1.4"
+    version: "^8.19.4 || ^9.1.4"
   elastic:
     subscription: "basic"
 icons:
--- a/packages/tenable_ot_security/changelog.yml
+++ b/packages/tenable_ot_security/changelog.yml
@@ -8,11 +8,17 @@
       type: enhancement
       link: https://github.com/elastic/integrations/pull/17670
     - description: >-
+        Change assets risk.total_risk field type from keyword to double.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/17670
+    - description: >-
         The _id scheme change means every existing asset document will be
-        duplicated until lifecycle policy removes the old versions. Customers
-        upgrading from 1.0.0 should install 1.1.0 first and allow at least
-        one successful asset collection interval before upgrading to this
-        version.
+        duplicated until lifecycle policy removes the old versions. Where
+        possible, upgrade to 1.1.0 first and allow at least one successful
+        asset collection interval before upgrading to this version. Where
+        this is not possible, the risk.total_risk field type change triggers
+        an automatic data stream rollover on upgrade, so the old backing
+        index begins aging out immediately without manual intervention.
       type: breaking-change
       link: https://github.com/elastic/integrations/pull/17670
 - version: "1.1.1"
--- a/packages/tenable_ot_security/changelog.yml
+++ b/packages/tenable_ot_security/changelog.yml
@@ -1,3 +1,20 @@
+- version: "2.0.0"
+  changes:
+    - description: >-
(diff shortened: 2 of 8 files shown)


kcreddy (dismissed):

efd6:
@kcreddy I plan to sit on this for a while. Will convert to draft again until we are ready.

packages/tenable_ot_security/changelog.yml
```diff
@@ -1,3 +1,20 @@
...
+        Change assets document fingerprint from full-body hash to id plus
+        timestamps. Assets now produce a new document only when a timestamp
+        field changes, preserving update history without creating duplicates
+        from non-temporal field noise.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/17670
+    - description: >-
+        The _id scheme change means every existing asset document will be
+        duplicated until lifecycle policy removes the old versions. Customers
+        upgrading from 1.0.0 should install 1.1.0 first and allow at least
```
andrewkroh: I think we might have a problem with these instructions. It looks like on Serverless a user can only install latest if I'm understanding the UI.

Image

kcreddy (approved):
Echoing same concern as Andrew, otherwise looks good.

packages/tenable_ot_security/changelog.yml
efd6: This means that we need to leave this open for a while. I think that is a defect TBH.

packages/microsoft_exchange_online_message_trace/manifest.yml
```diff
@@ -9,7 +9,7 @@ categories:
   - email_security
 conditions:
   kibana:
-    version: "^8.19.4 || ~9.0.7 || ^9.1.4"
+    version: "^8.19.4 || ^9.1.4"
```
andrewkroh: Did you mean to modify the exchange package and the tenable_ot_security package?

packages/tenable_ot_security/data_stream/assets/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -101,6 +108,17 @@ processors:
...
 
+  - convert:
+      tag: convert_total_risk_to_double
+      field: tenable_ot_security.assets.risk.total_risk
+      type: double
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+        - remove:
```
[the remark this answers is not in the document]
efd6: Fixed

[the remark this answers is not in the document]
efd6: Look at the logic, not the syntax.

andrewkroh (approved):

Issue #17286, opened 2026-02-06
[servicenow] Add configurable date format handling date processors


mohitjha-elastic:
**Description:**

The ServiceNow integration date processor currently attempts to parse dates using both `ddMMyyyy` and `MMddyyyy` formats by default. While this works in some cases, it introduces ambiguity for certain dates and can lead to incorrect parsing without any explicit error.

**Problem Statement:**

When both `ddMMyyyy` and `MMddyyyy` formats are enabled implicitly, dates such as:
`01012025`
`02032024`
are ambiguous and can be interpreted differently depending on the parsing order.

For example:
`01012025` could be interpreted as `1st Jan 2025 or Jan 1st 2025`

**Currently:**

The processor does not fail or warn on ambiguous cases.
Users have no control over which date format should be treated as authoritative.
This can result in silently incorrect data being ingested into ServiceNow, which is harder to detect and debug than a hard failure.

**Why This Is an Issue:**

Ambiguous date parsing can lead to data integrity issues.
Silent misinterpretation is riskier than explicit failure.

**Proposed Enhancement:**

Introduce a user-configurable date format option in the ServiceNow integration.
Allow users to explicitly choose the expected date format, for example:

- ddMMyyyy (and its common variants)
- MMddyyyy (and its common variants)
- Unambiguous formats such as ISO8601

**Breaking Change Consideration:**

Since this change alters date parsing behavior, it may need to be treated as a breaking change.

**Expected Outcome:**

Elimination of ambiguous date parsing
Improved data correctness and predictability
Better user control and clearer error feedback during ingestion

efd6:
As a refinement (or possibly clarification) to the proposal, I would suggest two options rather than three:

- ddMMyyyy (and its common variants) and unambiguous formats such as ISO8601
- MMddyyyy (and its common variants) and unambiguous formats such as ISO8601

This is since it is ~always safe to fall back to ISO8601.

It is also worth noting that the current implementation has an effective default of MMddyyyy over ddMMyyyy, so this should be chosen as the default option, _or_ there should be not default and the configuration would be required. Neither of these is the perfect solution; the first potentially silently defaults people into a situation that is incorrect for their data (this is no worse than the current situation), and the second requires that the a new configuration value be entered. In any case, I think that this **should** be marked as a breaking change and the affected data streams should be marked as `enabled: false`.

I think this does need further discussion though.

efd6:
We've had an internal discussion and this is something we can go ahead with as described in my last comment.

PR #17674, opened 2026-03-05
[servicenow] Add Configurable Date Format Handling

mohitjha-elastic:
## Proposed commit message

```
servicenow: Add configurable date format handling.

The ServiceNow date processor previously attempted to parse dates using both ddMMyyyy 
and MMddyyyy formats by default, which could create ambiguity for values such as 03052025 
or 03/05/2025 (e.g., March 5 vs May 3). This behavior could lead to silent misinterpretation of 
dates and potential data integrity issues since no warning or failure was triggered for ambiguous 
cases. To address this, a user-configurable date format preference has been introduced in the 
ServiceNow integration. Users can now explicitly select either ddMM or MMdd as their preferred 
format, with ddMM set as the default. Additionally, unambiguous formats are supported across both
branches, with ISO8601 handled as a fallback in all date processors.
```

## Checklist

- Clone integrations repo.

> elastic-package test -v

## Related Issues

- Closes https://github.com/elastic/integrations/issues/17286


packages/servicenow/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.4.0"
```
efd6: This is a breaking change, so the version should bump major.
```suggestion
- version: "2.0.0"
```

packages/servicenow/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 3.3.2
 name: servicenow
 title: "ServiceNow"
-version: "1.3.3"
+version: "1.4.0"
```
efd6:
```suggestion
version: "2.0.0"
```

packages/servicenow/data_stream/event/manifest.yml
```diff
@@ -123,6 +123,19 @@ streams:
...
+        type: select
+        title: Date Format Preference
+        multi: false
+        required: true
+        show_user: true
+        options:
+          - text: MM DD format
+            value: MMdd
+          - text: DD MM format
+            value: ddMM
```
efd6, line 137: I think we should have a default. Also below.

efd6 (commented):
The commit message says users can choose "unambiguous formats like ISO8601", but the select only offers MMdd and ddMM. Either add ISO8601 as an option with corresponding pipeline support, or correct the commit message. I would also suggest that all date processors should have ISO8601 as a fallback since it is unambiguous.

dd.MM.yyyy formats are missing from the MMdd branch. Dot-separated dates are conventionally day-first; no ServiceNow locale uses MM.dd.yyyy. These formats are unambiguous and should be present in both branches. A user who selects MMdd and receives dot-separated dates will get parse failures that didn't happen before. So this would be a regression.

Only the ddMM path is tested. Every test config sets date_format_preference: "ddMM". There's no coverage for the MMdd path. With roughly 60 duplicated processor blocks for MMdd, a typo or missing format in any of them could go unnoticed.

Also, in the commit message, it would be good to use an example date that is affected by the issue; 01012025 is not affected.

mohitjha-elastic:
@efd6 Thanks, I have updated the changes as well as commit message.

packages/servicenow/data_stream/event/_dev/test/pipeline/test-common-config.yml
efd6, line 1: Can we add pipeline tests for MMdd? It should be straightforward to generate the inputs from the inputs that we have for ddMM.

packages/servicenow/_dev/deploy/docker/files/config_oauth2.yml
```diff
@@ -211,8 +211,8 @@ rules:
                   "value": ""
                 },
                 "sys_updated_on": {
-                  "display_value": "2024-09-23 22:39:40",
-                  "value": "2024-09-24 05:39:40"
+                  "display_value": "06-07-2023 22:39:40",
+                  "value": "06-07-2023 15:39:40"
```
efd6:
```suggestion
                  "display_value": "06-07-2023 22:39:40",
                  "value": "2023-06-08 05:39:40"
```
`value` is [UTC format](https://www.servicenow.com/community/developer-blog/actual-vs-display-value-the-most-misunderstood-concept-in/ba-p/3311327) and the offset between the two values is because the test specifies TZ=America/Los_Angeles.

PR #17694, opened 2026-03-06
[BeyondTrust PRA] Improve custom_attribute handling and correctly parse IP addresses with ports

### LINKED ISSUE #17558: [BeyondTrust PRA]: Ingest pipeline fails converting IP:PORT and conditional access to custom_attribute
### Integration Name

BeyondTrust PRA [beyondtrust_pra]

### Dataset Name

beyondtrust_pra.access_session

### Integration Version

0.3.2

### Agent Version

N/A

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.19.9

N/A

### Software/API Version

N/A

### Error Message

```
"Processor conditional with tag in pipeline logs-beyondtrust_pra.access_session-0.3.2 failed with message: dynamic getter [java.lang.String, custom_attribute] not found"
"Processor convert with tag convert_destination_public_ip_to_ip in pipeline logs-beyondtrust_pra.access_session-0.3.2 failed with message: '<IP_ADDRESS>:<PORT>' is not an IP string literal., Processor conditional with tag in pipeline logs-beyondtrust_pra.access_session-0.3.2 failed with message: dynamic getter [java.lang.String, custom_attribute] not found"
```

### Event Original

The key issue is that json.destination.private_ip contains an IP:PORT value such as "30.30.30.30:30303", and the pipeline attempts to convert it to an ip type without splitting out the port. Some events also appear to not contain custom_attribute in the structure expected by the pipeline.

### What did you do?

- Installed and configured the BeyondTrust PRA integration in Fleet.
- Enabled collection for the access session dataset (beyondtrust_pra.access_session).
- Began ingesting BeyondTrust PRA access session events into Elasticsearch.
- Observed ingest pipeline failures in the integration pipeline logs-beyondtrust_pra.access_session-0.3.2.

### What did you see?

Ingest pipeline errors that prevent documents from being ingested:

- A convert processor (tag: convert_destination_public_ip_to_ip) fails because the source field contains an IP:PORT string (example: "30.30.30.30:30303") which is not a valid IP literal.
- A conditional processor fails with: dynamic getter [java.lang.String, custom_attribute] not found.
Because the pipeline fails, the integration cannot be used reliably (events are dropped / ingestion fails).

Successful ingestion of BeyondTrust PRA access session events, including events where:
- destination/private IP fields may be formatted as IP:PORT (pipeline should either parse/split into separate ip and port fields or skip conversion safely).
- custom_attribute is missing or not in the exact structure expected (pipeline should guard for missing fields and not error).

### Anything else?

The value in json.destination.private_ip is an IP:PORT combination, so conversion to an ip field type fails unless it is split first. Suggested behavior would be to extract IP into a dedicated ip field and port into a separate numeric field (or store the raw value in a non-ip field and conditionally parse).

moxarth-rathod:
## Proposed commit message

```
beyondtrust_pra: fix IP:PORT conversion and custom_attribute access in access_session pipeline

The ingest pipeline failed on two classes of events:

1. IP fields (destination.private_ip, destination.public_ip,
   performed_by.private_ip, performed_by.public_ip) sometimes
   contain an IP:PORT value such as "30.30.30.30:30303". The
   convert processor (type: ip) rejects these because they are
   not bare IP literals. A grok step now splits IP and port
   before conversion, and the port is mapped to new fields.

2. The foreach conditional on custom_attributes.custom_attribute
   used null-safe access (?.) which guards against null but not
   against type mismatches. When custom_attributes is a plain
   string rather than a Map, Painless throws "dynamic getter
   [java.lang.String, custom_attribute] not found". The
   conditionals now check instanceof Map first.

Fixes #17558
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/beyondtrust_pra directory.
> elastic-package test

## Related issues

- Closes #17558

### DIFF
--- a/packages/beyondtrust_pra/data_stream/access_session/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/beyondtrust_pra/data_stream/access_session/elasticsearch/ingest_pipeline/default.yml
@@ -124,10 +124,8 @@ processors:
   - grok:
       field: json.destination.private_ip
       tag: grok_destination_private_ip
-      pattern_definitions:
-        PORT: '[0-9]+'
       patterns:
-        - ^%{IP:json.destination.private_ip}:%{PORT:json.destination.private_port}
+        - ^%{IP:json.destination.private_ip}:%{POSINT:json.destination.private_port}$
       ignore_failure: true
   - convert:
       field: json.destination.private_ip
@@ -163,10 +161,8 @@ processors:
   - grok:
       field: json.destination.public_ip
       tag: grok_destination_public_ip
-      pattern_definitions:
-        PORT: '[0-9]+'
       patterns:
-        - ^%{IP:json.destination.public_ip}:%{PORT:json.destination.public_port}
+        - ^%{IP:json.destination.public_ip}:%{POSINT:json.destination.public_port}$
       ignore_failure: true
   - convert:
       field: json.destination.public_ip
@@ -561,10 +557,8 @@ processors:
   - grok:
       field: json.performed_by.private_ip
       tag: grok_performed_by_private_ip
-      pattern_definitions:
-        PORT: '[0-9]+'
       patterns:
-        - ^%{IP:json.performed_by.private_ip}:%{PORT:json.performed_by.private_port}
+        - ^%{IP:json.performed_by.private_ip}:%{POSINT:json.performed_by.private_port}$
       ignore_failure: true
   - convert:
       field: json.performed_by.private_ip
@@ -600,10 +594,8 @@ processors:
   - grok:
       field: json.performed_by.public_ip
       tag: grok_performed_by_public_ip
-      pattern_definitions:
-        PORT: '[0-9]+'
       patterns:
-        - ^%{IP:json.performed_by.public_ip}:%{PORT:json.performed_by.public_port}
+        - ^%{IP:json.performed_by.public_ip}:%{POSINT:json.performed_by.public_port}$
       ignore_failure: true
   - convert:
       field: json.performed_by.public_ip
--- a/packages/beyondtrust_pra/data_stream/access_session/fields/fields.yml
+++ b/packages/beyondtrust_pra/data_stream/access_session/fields/fields.yml
@@ -45,11 +45,13 @@
               description: The user's private IP address.
             - name: private_port
               type: long
+              description: The user's private port.
             - name: public_ip
               type: ip
               description: The user's public IP address.
             - name: public_port
               type: long
+              description: The user's public port.
             - name: seconds_involved
               type: long
               description: Integer value indicating the number of seconds the user was involved in this session.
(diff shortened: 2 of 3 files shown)


packages/beyondtrust_pra/data_stream/access_session/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -121,6 +121,14 @@ processors:
       tag: rename_destination_os
       target_field: beyondtrust_pra.access_session.destination.os
       ignore_missing: true
+  - grok:
+      field: json.destination.private_ip
+      tag: grok_destination_private_ip
+      pattern_definitions:
+        PORT: '[0-9]+'
+      patterns:
+        - ^%{IP:json.destination.private_ip}:%{PORT:json.destination.private_port}
```
kcreddy:
```suggestion
        - ^%{IP:json.destination.private_ip}:%{POSINT:json.destination.private_port}
```

Use `POSINT` over custom `PORT` pattern.
[Usage](https://github.com/logstash-plugins/logstash-patterns-core/blob/main/patterns/ecs-v1/grok-patterns#L33)
[Example](https://github.com/elastic/integrations/blob/main/packages/apache/data_stream/access/elasticsearch/ingest_pipeline/default.yml#L23)

packages/beyondtrust_pra/data_stream/access_session/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -137,6 +145,29 @@ processors:
...
+      tag: set_destination_nat_port_from_destination_private_port
+      copy_from: beyondtrust_pra.access_session.destination.private_port
+      ignore_empty_value: true
+  - grok:
+      field: json.destination.public_ip
+      tag: grok_destination_public_ip
+      pattern_definitions:
+        PORT: '[0-9]+'
+      patterns:
+        - ^%{IP:json.destination.public_ip}:%{PORT:json.destination.public_port}
```
kcreddy: Same here.

packages/beyondtrust_pra/data_stream/access_session/elasticsearch/ingest_pipeline/default.yml
kcreddy: Also add `$` anchor to grok pattern. [Example](https://github.com/elastic/integrations/blob/main/packages/apache/data_stream/access/elasticsearch/ingest_pipeline/default.yml#L112)

packages/beyondtrust_pra/data_stream/access_session/fields/fields.yml
```diff
@@ -43,9 +43,13 @@
             - name: private_ip
               type: ip
               description: The user's private IP address.
+            - name: private_port
+              type: long
```
kcreddy, line 47: Also add descriptions like other fields.

PR #17695, opened 2026-03-06
More fixes after trying the simplified api

jsoriano:
Follow up to https://github.com/elastic/integrations/pull/17656, more issues found on tests after trying to use the simplified API in https://github.com/elastic/elastic-package/pull/3307.

In m365_defender and microsoft_defender_cloud the mock configuration is fixed for the correct refresh token url.
Variables were not correctly set with previous implementation of elastic-package and it looks like the wrong refresh was never tried, so it was silently ignored.

The test in ti_abusetech was actually missing a required variable.


jsoriano:
@claude 

> **`packages/m365_defender/_dev/deploy/docker/vulnerability-http-mock-config.yml`** (lines 12-22) still uses the old `/tenant_id/oauth2/token` path and includes the `Authorization: Basic` header for the refresh token rule, which appears to be the same pattern being fixed in the alert and incident mock configs. Consider whether this file should also be updated for consistency.

You are right that this config looks also incorrect, but it is not failing at the moment with main or with the changes in https://github.com/elastic/elastic-package/pull/3307, so I would leave it as is by now.

PR #17700, opened 2026-03-06
[watchguard_firebox] Fix grok to allow underscores in cluster member names

ilyannn:
## Summary

- The `member_info` field in the syslog grok pattern used `%{SYSLOGHOST}` (which resolves to `%{HOSTNAME}`) — a pattern that only allows alphanumerics and hyphens
- WatchGuard cluster member names can contain underscores, causing the entire syslog grok to fail with a `pipeline_error`
- Replace `%{SYSLOGHOST}` with the existing custom `%{WATCHGUARDHOSTNAME}` pattern which already supports underscores

## Root cause

The standard grok `HOSTNAME` pattern character class is `[0-9A-Za-z-]` (no underscore). The custom `WATCHGUARDHOSTNAME` defined in the same pipeline already uses `[0-9A-Za-z\-\_]` but was only applied to the hostname field, not to `member_info`.

## Reproducer

```
POST /_ingest/pipeline/logs-watchguard_firebox.log-1.6.0/_simulate
{
  "docs": [
    {
      "_source": {
        "message": "<142>Mar 6 10:53:27 WatchGuard-Firebox-FW2 FVE6035FD3AE3 Cluster_Member-Info1 (2026-03-06T08:53:27) firewall: msg_id=\"3000-0148\" Allow Production External 52 tcp 20 127 10.0.1.2 89.160.20.128 59996 443 offset 8 S 770032243 win 65535 geo_dst=\"SWE\" (HTTPS-proxy-00)"
      }
    }
  ]
}
```

## Checklist

Also:

- [x] All pipeline tests pass (`elastic-package test pipeline`)
- [x] Kept existing expected timestamps at 2025 to avoid unnecessary date churn
- [x] Data anonymized

### DIFF
--- a/packages/watchguard_firebox/changelog.yml
+++ b/packages/watchguard_firebox/changelog.yml
@@ -1,9 +1,11 @@
 # newer versions go on top
-- version: "1.6.1"
+- version: "1.6.2"
   changes:
     - description: Fix grok pattern to allow underscores in cluster member names.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/17700
+- version: "1.6.1"
+  changes:
     - description: Fix Grok Pattern for Watchguard Member Info.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/17856


ilyannn:
I will close this PR as most of the work has already been done – this PR would have added a different test, but it's not worth it.

PR #17742, opened 2026-03-10
[PPS]: Fix parsing for plain usernames and space-padded dates for syslog messages 

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
pps: fix parsing for plain usernames and space-padded dates

Adds a grok pattern variant to handle syslog messages where the username
has no domain suffix (e.g. `jsmith - Admins - Success` vs `user@domain - Success`).

This also fixes the `Backup Restore Service` entries which were previously
matched by the fallback pattern and incorrectly mapped the hostname to
`host.domain` instead of `host.hostname`.

Also adds `MMM  d HH:mm:ss` as an additional date format to handle BSD syslog's
space-padded single-digit days (e.g. `Feb  5`), which caused the date processor
to throw a parse error at index 4.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/pps/data_stream/log/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/pps/data_stream/log/elasticsearch/ingest_pipeline/default.yml
@@ -64,8 +64,7 @@ processors:
       field: event.created
       target_field: "event.created"
       formats:
-        - "MMM dd HH:mm:ss"
-        - "MMM  d HH:mm:ss"
+        - "MMM [dd][ d] HH:mm:ss"
       timezone: "{{{event.timezone}}}"
       if: ctx.event?.timezone != null && ctx.event?.created != null
       on_failure:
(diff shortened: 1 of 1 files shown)


packages/pps/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -64,6 +65,7 @@ processors:
       target_field: "event.created"
       formats:
         - "MMM dd HH:mm:ss"
+        - "MMM  d HH:mm:ss"
```
efd6:
```suggestion
        - "MMM [dd][ d] HH:mm:ss"
```

packages/pps/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -78,6 +80,7 @@ processors:
...
```
efd6:
```suggestion
        - "MMM [dd][ d] HH:mm:ss"
```

ShourieG:
@efd6, resolved all comments

ghost:
Hi @ShourieG , thank you for the fix. However, I have noticed that the new pattern in the date processor struggles when numbers for the day and for the hour are the same. E.g. "Mar 10 10:00:00" leads to an error, while "Mar 11 10:00:00" is working just fine. Would it be possible for you to address this issue? Thanks in advance!

PR #17009, opened 2026-01-20
Updated Logstash Single Pipeline View to display batch size and event count metrics

andsel:
## Proposed commit message
Update a couple of Logstash metrics dashboard to display batch size and event count. This feature is available since Logstash 9.2

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## How to test this PR locally

The test is based on a running Logstash instance monitored by an  ElasticAgent which push metric data to be processed by the integration.

Requirements: Docker must be running on test host.

1. Install `elastic-package`
   1. Download from https://github.com/elastic/elastic-package/releases/tag/v0.117.1
   2. Add permission to the file with 
   
         ```sh
         xattr -r -d com.apple.quarantine elastic-package
         ```
     
   3. Run it 
   
         ```sh
         ./elastic-package
         ```

2. Bring up stack and build/install the integration (from `integration/packages/logstash` of the Elastic Integration local clone)
   1. Launch the stack

         ```sh
         /path/to/elastic-package stack up -v
         ```   

   2. Build the package locally

         ```sh
         /path/to/elastic-package build
         ```  

   3. Install into the package registry running in the Docker, so it can be served to Fleet

         ```sh
         /path/to/elastic-package install
         ```  

   4. Verify the integration is installed, check the listing: https://localhost:5601/app/integrations/installed?currentPage=1

3. Now install and configure a local Logstash that generates some metrics, this is monitored by a local agent that will be installed in next step
   1. Download Logstash `>= 8.2.0` from https://www.elastic.co/downloads/logstash
   2. Once unpacked edit the `config/logstash.yml` to have the following settings:

         ```yaml
         pipeline.batch.metrics.sampling_mode: "full"
         monitoring.enabled: false # this is already the default
         ```
[...]

PR #17785, opened 2026-03-12
[Logstash integration] Covered batch structure's metrics navigation with safe operator

### LINKED ISSUE #17784: Logstash integration 2.9.0 incompatibility with Logstash < 9.2.0
### Integration Name

Logstash [logstash]

### Dataset Name

_No response_

### Integration Version

2.9.0

### Agent Version

8.19.9

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.19.9

n/a

### Software/API Version

Logstash 8.19

### Error Message

This integration, when used alongside Logstash < 9.2.0, puts the Elastic Agent in degraded health, with this error:
```
{"log.level":"error","@timestamp":"2026-03-10T01:01:01.000Z","message":"single event object returned by evaluation","component":{"binary":"filebeat","dataset":"elastic_agent.filebeat","id":"cel-default","type":"cel"},"log":{"source":"cel-default"},"log.logger":"input.cel","service.name":"filebeat","id":"cel-logstash.pipeline-<redacted>","error":{"message":"failed eval: ERROR: <input>:100:15: no such key: batch\n |     \"events\": pipelines.map(pipeline, {\n | ..............^"},"log.origin":{"file.line":459,"file.name":"cel/input.go","function":"github.com/elastic/beats/v7/x-pack/filebeat/input/cel.input.run.func1"},"input_source":"http://localhost:9600/_node","input_url":"http://localhost:9600/_node","ecs.version":"1.6.0","ecs.version":"1.6.0"}
```

### Event Original

n/a

### What did you do?

Upgrade Logstash integration from 2.8.0 to 2.9.0 while usign Logstash 8.19

### What did you see?

Elastic Agent is going into degraded health

Elastic Agent to stay in a healthy state after upgrading the Logstash integration with this <9.2.0 version of Logstash

### Anything else?

Logstash Integration [2.9.0](https://github.com/elastic/integrations/pull/17009) introduced the [following enhancement](https://www.elastic.co/docs/reference/integrations/logstash#changelog):
```
adds visualizations for monitoring batch event count and byte size. Note that these metrics will only show results for Logstash versions 9.2.0 or later.
```

This seems likely from https://github.com/elastic/integrations/pull/17009/changes#diff-41525c3f3f426f522a2880dc04da52deb7b57e18efef2c1fdcdc911c3fd6c0b5R65-R80, where there's no guard for the possibility of `.batch` not existing.

andsel:
## Proposed commit message

Fixes the CEL scripts used to access batch metrics substructure with safe-navigation operator. This fixes ElasticAgent becomes unhealthy when monitor an older Logstash version that doesn't provide those metrics. 

## Checklist

- ~~[ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) ~~

## Author's Checklist

- [ ] 

## Related issues

- Relates #17009
- Closes #17784

### DIFF
--- a/packages/logstash/data_stream/pipeline/agent/stream/cel.yml.hbs
+++ b/packages/logstash/data_stream/pipeline/agent/stream/cel.yml.hbs
@@ -62,12 +62,12 @@ program: |
         },
         "batch":{
           "event_count":{
-            "current": body.pipelines[pipeline_name].?batch.?event_count.?current.orValue([]),
-            "average.lifetime": body.pipelines[pipeline_name].?batch.?event_count.?average.lifetime.orValue([])
+            "current": body.pipelines[pipeline_name].?batch.event_count.current.orValue([]),
+            "average.lifetime": body.pipelines[pipeline_name].?batch.event_count.average.lifetime.orValue([])
           },
           "byte_size":{
-            "current": body.pipelines[pipeline_name].?batch.?byte_size.?current.orValue([]),
-            "average.lifetime": body.pipelines[pipeline_name].?batch.?byte_size.?average.lifetime.orValue([]),
+            "current": body.pipelines[pipeline_name].?batch.byte_size.current.orValue([]),
+            "average.lifetime": body.pipelines[pipeline_name].?batch.byte_size.average.lifetime.orValue([]),
           },
         },
         "events":{


packages/logstash/data_stream/pipeline/agent/stream/cel.yml.hbs
```diff
@@ -62,24 +62,12 @@ program: |
...
-            "current": has(body.pipelines[pipeline_name].batch.event_count.current) ?
-              body.pipelines[pipeline_name].batch.event_count.current
-            :
-              [],
-            "average.lifetime": has(body.pipelines[pipeline_name].batch.event_count.average.lifetime) ?
-              body.pipelines[pipeline_name].batch.event_count.average.lifetime
-            :
-              [],
+            "current": body.pipelines[pipeline_name].?batch.event_count.current.orValue([]),
+            "average.lifetime": body.pipelines[pipeline_name].?batch.event_count.average.lifetime.orValue([])
```
jsvd, line 66: Which nils do we have to protect against?

1. `body.pipelines[pipeline_name].batch` being empty?
2. `body.pipelines[pipeline_name]` being empty?

If we want to accept `body.pipelines[pipeline_name]` being nil, the ? would need to be inside the square brackets:

```suggestion
            "current": body.pipelines[?pipeline_name].batch.event_count.current.orValue([]),
            "average.lifetime": body.pipelines[?pipeline_name].batch.event_count.average.lifetime.orValue([])
```

if `body.pipelines[pipeline_name]` is always valid then we only need `body.pipelines[pipeline_name].?batch`

andsel: Why we have to check for nil of `body.pipelines[pipeline_name]`? There are many places where that access `body.pipelines[pipeline_name]` without nil check without any problem, here we have to be sure that `batch` subtree is not nil.

jsvd: > Why we have to check for nil of body.pipelines[pipeline_name]? 

Just asking, wasn't sure we needed to. if we only have to check if batch is there, then LGTM

andsel: The problem of bug originated from the absence of `batch`, so it got a nil error in evaluating the path containing `batch` on a payload that didn't have it. So I think that the goal here is check the existence of `batch` subtree, which wasn't available in previous payload, it's a sort of backward compatibility check.

Issue #17744, opened 2026-03-10
[Elasticsearch Service Billing]: Credits purchased midway do not add to the balance


eddxavier-elastic:
### Integration Name

Elasticsearch Service Billing [ess_billing]

### Dataset Name

ess_billing.billing

### Integration Version

1.7.0

### Agent Version

9.3.1

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.3.1

### OS Version and Architecture

Debian 6.1.158-1

### Error Message

The integration has an issue adding credits midway through a contract term. For this org id it only shows credits procured on their first contract, where they added credits halfway.

<img width="1011" height="263" alt="Image" src="https://github.com/user-attachments/assets/31927660-ef0e-4031-af10-9bb4199ab859" />

### What did you do?

The integration was configured with the default settings using Elastic Cloud API key.

### What did you see?

Only first year/term credits being added to the balance

### What did you expect to see?

All credits processed for the org id

PR #17793, opened 2026-03-12
ess_billing: fix wrong handling of credit data

3kt:
## Summary

- **Fix credits dashboard**: Rewrite all panels from form-based Lens/KQL to ES|QL queries to correctly handle multiple active credit lines. The previous KQL filters (`ess.billing.end >= now-1M/M AND ess.billing.ecu_balance > 0 AND @timestamp > now-24h`) could miss or double-count credits with overlapping periods. The new approach filters on `ess.billing.active == true` and uses `INLINE STATS latest = MAX(@timestamp)` to always show the most recent balance snapshot.
- **Fix run-rate calculation**: Replace brittle timestamp arithmetic with `COUNT_DISTINCT(DATE_TRUNC(1 day, @timestamp))` for accurate daily averaging.
- **Fix ingest pipeline**: Extract SKU service type into `ess.billing.cloud.service.type` instead of overwriting the ECS `cloud.service.type` field for non-deployment SKUs.
- **Add deployment_name filter** to the billing dashboard.

## Test plan

- [x] Verify the credits dashboard correctly displays active credit balances when multiple credit lines exist
- [x] Verify the run-rate metric shows a reasonable ECUs/day value
- [x] Verify the credit burn-down chart renders correctly
- [x] Verify the billing dashboard deployment_name filter works
- [x] Verify non-deployment SKU records have `ess.billing.cloud.service.type` populated (not `cloud.service.type`)

Fixes https://github.com/elastic/integrations/issues/17744


ishleenk17:
@claude review the PR

3kt:
@claude your previous comments were addressed, can you check?

packages/ess_billing/manifest.yml
```diff
@@ -10,7 +10,7 @@ categories:
   - cloud
 conditions:
   kibana:
-    version: "^8.18.0 || ^9.0.0"
+    version: "^9.3.0"
```
muthu-mps, line 13: This would trigger users to upgrade the stack version to apply this fix. Is there a reason to upgrade the stack version dependency?

muthu-mps:
Can we add latest dashboard screenshots to this PR to verify the panels?

packages/ess_billing/manifest.yml
3kt: Dashboard compatibility - I needed to use ES|QL rather than the old Lens approach to fix the handling, and I'm unsure how nice that would play out on older versions.

eddxavier-elastic:
@muthu-mps here is a couple of dashboards screenshots showing correct credit balance
image

image

packages/ess_billing/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.7.1"
+  changes:
+    - description: Fix wrong handling of credit data
```
muthu-mps:
```suggestion
    - description: Fix incorrect handling of credit data
```

PR #17798, opened 2026-03-13
[doc] Update number_of_workers in aws integration to align with cloudwatch api rate limit

kaiyan-sheng:
## Proposed commit message

Updates AWS CloudWatch–related documentation and integration UI text so `number_of_workers` is no longer described as “one per log group.” The previous guidance led users with many log groups to set a high worker count, which triggers AWS DescribeLogGroups/FilterLogEvents rate limits (5 TPS) and causes `ThrottlingException: Rate exceeded` and DEGRADED status in Fleet (especially after 8.19.x unit health reporting).

## Checklist


efd6 (approved):
Agree with @zmoog's https://github.com/elastic/integrations/pull/17798#pullrequestreview-3952651680. If we don't include it in the docs, we should probably have a KB article for it.

PR #17827, opened 2026-03-15
[AWS-cloudtrail] set host.id field

alexreal1314:
## Proposed commit message

```
Fix CloudTrail host entity classification to only use EC2 instance IDs.

The entity classification logic incorrectly treated AMI, EBS volume, and snapshot IDs as host resources. Only EC2 instance IDs (i-*) are genuine host identifiers.
Removed volume, snapshot, and image from hostResourceTypes, keeping only instance. Removed vol-, snap-, and ami- from hostIdPrefixes, keeping only i-. These non-instance resource IDs now fall through to genericTargets and are stored in entity.target.id instead of host.target.entity.id.
Added host.id as a new field set to a single string value (the first EC2 instance ID) rather than an array, since host.id should represent one host per document.

Updated test expected files accordingly: replaced host blocks with entity.target.id for snapshot and AMI tests, and set host.id as a single string in the five EC2 instance test files.
```

closes https://github.com/elastic/security-team/issues/16343.

---

**Fix host entity classification in CloudTrail ingest pipeline**

**Problem**

The CloudTrail ingest pipeline's entity classification logic treated all resources matching `hostIdPrefixes` (`i-`, `vol-`, `snap-`, `ami-`) and `hostResourceTypes` (`instance`, `volume`, `snapshot`, `image`) as host entities. This caused AMI IDs, EBS volume IDs, and snapshot IDs to be incorrectly stored in `host.target.entity.id` and `host.id`.

These resource types are not hosts in any meaningful sense — only EC2 instance IDs (`i-*`) represent actual compute hosts. For example, `ModifySnapshotAttribute` events were producing documents with `host.id: ["snap-0a392d80692e2526a"]`, which is semantically wrong.

**Changes**

1. **`hostResourceTypes`**: Removed `volume`, `snapshot`, and `image` — only `instance` remains.
2. **`hostIdPrefixes`**: Removed `vol-`, `snap-`, `ami-` — only `i-` remains.
3. **`host.id`**: New field added in this PR to surface the target host identifier. Set to a single string value (first EC2 instance ID) rather than an array, since `host.id` represents one host per document.
4. **Routing**: Non-instance resource IDs (`vol-*`, `snap-*`, `ami-*`) now fall through to `genericTargets` and are stored in `entity.target.id` instead.

**Affected fields**

| Resource type | Before | After |
|---|---|---|
| EC2 instance (`i-*`) | `host.target.entity.id` (array) | `host.target.entity.id` (array), `host.id` (string) |
| AMI (`ami-*`) | `host.target.entity.id` | `entity.target.id` |
| EBS volume (`vol-*`) | `host.target.entity.id` | `entity.target.id` |
| Snapshot (`snap-*`) | `host.target.entity.id` | `entity.target.id` |

**Test updates**

- `test-modify-snapshot-attribute-json.log-expected.json` — `host` block replaced with `entity.target.id`
- `test-modify-image-attribute-json.log-expected.json` — `host` block replaced with `entity.target.id`
- 5 EC2 instance test files — added `host.id` as a single string value

---

## Checklist

## Author's Checklist

- [ ] 

1. `cd packages/aws`
2. run `elastic-package test pipeline -v --data-streams cloudtrail --generate`

## Related issues

-


packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -941,6 +941,7 @@ processors:
           }
           if (!hostTargets.isEmpty()) {
             field("host.target.entity.id").set(hostTargets);
```
andrewkroh, line 937: This is a pre-existing issue, but the `hostTargets` contained non-host entities. The `hostTargets` set includes not just EC2 instance IDs (`i-*`) but also AMI, EBS, and snapshot IDs. These are not "hosts" in any ECS sense. Only EC2 instance IDs are genuinely host identifiers IMO.

The other resource types seem like they should be routed to `entity.target.id` via the `hostIdPrefixes`.

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -941,6 +941,7 @@ processors:
...
+            field("host.id").set(hostTargets);
```
andrewkroh: ECS defines `host.id` as scalar keyword. The pipeline sets it as an array via `TreeSet`. While Elasticsearch accepts arrays in keyword fields, this diverges from the ECS contract and will break consumers expecting a single string value (such as ES|QL queries).

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
alexreal1314: thanks, fixed, only `i-* id` patterns are now assigned to the host.target.entity.id field.

packages/aws/data_stream/cloudtrail/elasticsearch/ingest_pipeline/default.yml
alexreal1314: you are correct, although we have other cases where we set keyword fields as array it might be best to set a single value.

andrewkroh (approved):
The `host.id` value seems better aligned to ECS now.

Regarding the proposed commit message, those are going to be viewed as plain text by tools like `git log` so do not use Markdown. Some recent examples: https://github.com/elastic/integrations/pull/17508, https://github.com/elastic/integrations/pull/17379. So please edit that to avoid markdown, and then be sure to paste it during the squash and merge.

PR #17829, opened 2026-03-16
checkpoint_email: handle 401 on event query by clearing cached token

efd6:
## Proposed commit message
```
checkpoint_email: handle 401 on event query by clearing cached token

When the event query returns 401, the CEL program now sets expiry to
the past, forcing re-authentication on the next polling cycle. Without
this, a server-side token invalidation causes repeated 401s until the
token's 30-minute clock expiry passes naturally.

Add a script test that verifies the recovery: a mock cycles the event
query between 401 and 200, and the test asserts that both an error
event (from the 401) and data events (from the subsequent 200) are
indexed. The 401 response body matches a real Checkpoint response. The
200 event bodies are based on the existing system test mock.
```

> [!NOTE]
> Workaround in second commit to be reverted when elastic/elastic-package#3321 is fixed.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/checkpoint_email/data_stream/event/_dev/test/scripts/env.txt
andrewkroh, line 1: Is this a script test env debugging aid?

efd6: It's a validation. It can be removed.

PR #17858, opened 2026-03-17
[eset_protect] Add new ECOS Detection data stream which uses the /v2/detections endpoint

brijesh-elastic:
## Proposed commit message

```
eset_protect: add new ECOS Detection data stream which uses the /v2/detections endpoint

This PR removes the `API Version` selection option in the detection data stream, which was originally
introduced in version (eset_protect-2.0.0)[1]. It replaces this option with a new ECOS detection
data stream that uses the `/v2/detections` endpoint. These logs are then rerouted to the
main `detection` data stream for parsing and indexing, as they share a similar log structure.

The reason for this change is that the `/v2/detections` endpoint specifically returns detections from
ECOS, whereas `/v1/detection` endpoint cover ESET Protect and ESET Inspect. More details in the
(ESET Connect documentation)[2]. So, introducing this new data stream helps users collect
all types of detection logs more effectively.

This PR updates the query parameter naming convention for the detection data stream from
snake_case to camelCase.
While the (ESET Connect API documentation)[3] specifies camelCase, internal testing indicates that
the API currently supports both formats. This change ensures our implementation aligns strictly
with the official documentation.

[1] https://github.com/elastic/integrations/pull/17447
[2] https://help.eset.com/eset_connect/en-US/incident_management.html
[3] https://help.eset.com/eset_connect/en-US/incident_management_v1_detections_get.html
```

## Checklist

- Clone integrations repo.

> elastic-package test -v


packages/eset_protect/data_stream/ecos_detection/_dev/test/pipeline/test-detection.log
```diff
@@ -0,0 +1 @@
+{"uuid":"eabcdefd-1234-5678-abcd-abcdef123456","typeName":"Spam","category":"DETECTION_CATEGORY_UNSPECIFIED","displayName":"Antispam reason: URL example.com found on cloud blacklist","objectName":"","objectTypeName":"","objectHashSha1":"","objectUrl":"","edrRuleUuid":"","circumstances":"","severityLevel":"SEVERITY_LEVEL_UNSPECIFIED","occurTime":"2026-01-01T12:30:00.000Z","note":"","severityScore":0,"resolved":false,"responses":[],"userName":"","objectSizeBytes":"0","cloudOfficeTenantUuid":"123456789","email":{"reference":"","to":"example@example.com","mailboxUserUuid":"12345678901234567890","cc":"","from":"example@example.com","subject":"Example subject","bodyParts":[],"attachments":[],"headers":[],"containedUrls":[],"isReadOnly":false,"internetMessageId":"12345678901234","senderIpAddress":"","origin":"OBJECT_ORIGIN_GOOGLE_WORKSPACE","mailboxUuid":""},"scanUuid":""}
```
chrisberkhout, line 1: Please make this a v2 ECOS detection. Can be based on the API doc. Should have `triggeringEvent` and possibly some other extra fields (v2 seems to embed more related data) to exercise how the (main) detection pipeline handles that.

chrisberkhout: Exercising the redirect in pipeline tests may be tricky. I think maybe what we've done elsewhere is to put all the tests cases in the destination data stream's pipeline tests, so this one should go there too.

packages/eset_protect/_dev/deploy/docker/files/detection-config.yml
chrisberkhout, line 1: Would be good to have a basic system test that uses v2 the endpoint.

packages/eset_protect/data_stream/ecos_detection/manifest.yml
```diff
@@ -0,0 +1,109 @@
+title: Collect Detection logs detected by ESET Cloud Office Security
+type: logs
+dataset: eset_protect.ecos_detection
+streams:
+  - input: cel
+    title: ECOS Detection Logs
+    description: Collect Detection logs detected by ESET Cloud Office Security.
```
chrisberkhout: Probably need to say here that these will be redirected to the main detection data stream.

packages/eset_protect/data_stream/ecos_detection/elasticsearch/ingest_pipeline/default.yml
chrisberkhout, line 1: I'm wondering whether the fields definitions and ingest pipeline need to be here when we have the routing happening. Is that the case?

I think I've seen other cases with `routing_rules.yml` where those things go away.

packages/eset_protect/data_stream/ecos_detection/routing_rules.yml
```diff
@@ -0,0 +1,7 @@
+- source_dataset: eset_protect.ecos_detection
+  rules:
+    - target_dataset: eset_protect.detection
+      if: ctx.tags != null && ctx.tags.contains('eset_protect-ecos_detection')
+      namespace:
+        - "{{data_stream.namespace}}"
+        - default
```
chrisberkhout, line 7: Does the `- default` line need to be there?

chrisberkhout (commented):
Very nice commit message.

packages/eset_protect/data_stream/ecos_detection/_dev/test/pipeline/test-detection.log
brijesh-elastic: Yes, we already have that [test case](https://github.com/elastic/integrations/blob/main/packages/eset_protect/data_stream/detection/_dev/test/pipeline/test-detection.log#L8) in the detection data stream's pipeline test log, which includes the `triggeringEvent` field. I will add more test cases once we have the live logs.

The log I placed under the `ecos_detection` data stream is just to test the `eset_protect-ecos_detection` tags we are adding to the pipeline.

packages/eset_protect/_dev/deploy/docker/files/detection-config.yml
brijesh-elastic: It isn't possible to add system tests for the routing data stream at this time. See [issue](https://github.com/elastic/elastic-package/issues/1917).

packages/eset_protect/data_stream/ecos_detection/elasticsearch/ingest_pipeline/default.yml
brijesh-elastic: > I'm wondering whether the fields definitions and ingest pipeline need to be here when we have the routing happening. Is that the case?

This is not required, as all logs are redirected to the `detection` data stream. Therefore, no logs will be indexed in the `ecos_detection` data stream, except for any data collection errors that may occur.

packages/eset_protect/data_stream/ecos_detection/routing_rules.yml
brijesh-elastic: It works as fallback, if `{{data_stream.namespace}}` resolves to null, missing, or non-string.

packages/eset_protect/data_stream/ecos_detection/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Okay. So they can be removed?

packages/eset_protect/data_stream/ecos_detection/_dev/test/pipeline/test-detection.log
chrisberkhout: Okay. Am I right that this is not an ECOS one though? Even if that's not relevant to the test of tags, in the future we'll probably look at pipeline test cases here and expect the actual data collected in this data stream matches, so it's good to make match the realistic structure as far as possible.

chrisberkhout (changes requested):
I added to comments from the earlier review where I think those points can still be tidied up.

The other thing is a breaking change changelog message so users get advice on upgrading from 2.0.0 in the upgrade UI. Note it's possible for one changelog entry to have several changes entries, so we can keep the enhancement one and add a breaking-change one with the message we want to show.

packages/eset_protect/data_stream/ecos_detection/_dev/test/pipeline/test-detection.log
brijesh-elastic: Right, It is ECOS detection log.

chrisberkhout: Oh, okay. Thanks!

packages/eset_protect/data_stream/ecos_detection/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Would be good to try, but I realize I was thinking of something I've seen before that's a different case (no ingest pipeline in destination data stream when source has already done the processing).

packages/eset_protect/docs/README.md
```diff
@@ -4,9 +4,11 @@ ESET PROTECT enables you to manage ESET products on workstations and servers in
 
 ## Data streams
 
-The ESET PROTECT integration collects three types of logs: Detection, Device Task and Event.
+The ESET PROTECT integration collects following types of logs:
```
ezzamo:
```suggestion
The ESET PROTECT integration collects the following types of logs:
```

chrisberkhout (commented):
@brijesh-elastic I think this can be closed now.

As I undestand it, the current version is back to where it started, on v1.

With guidance from ESET our plan is to support v3, which is to be released soon.

nestyy1337:
@chrisberkhout Hey! Do we have any ETA on the v3 release from ESET?

chrisberkhout:
@nestyy1337 Not yet. We're watching for updates on https://help.eset.com/eset_connect/en-US/release_notes.html

PR #17859, opened 2026-03-17
Handle misp.attribute.timestamp being an Integer

chrisberkhout:
## Proposed commit message

```
Handle misp.attribute.timestamp being an Integer

Fixes an issue observed in a production system, where the value can be
an integer, not just a string.
```

## Checklist

### DIFF
--- a/packages/ti_misp/changelog.yml
+++ b/packages/ti_misp/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "1.41.2"
   changes:
-    - description: Handle misp.attribute.timestamp being an Integer
+    - description: Handle misp.attribute.timestamp being a number or a string.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/17859
 - version: "1.41.1"


efd6 (approved):

packages/ti_misp/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.41.2"
+  changes:
+    - description: Handle misp.attribute.timestamp being an Integer
```
andrewkroh:
```suggestion
    - description: Handle misp.attribute.timestamp being a number or a string.
```

andrewkroh (approved):
LGTM. I'm assuming that the timestamp attribute is guaranteed to always be present by MISP despite the null guard on line 49 for the `date` processor.

chrisberkhout:
> LGTM. I'm assuming that the timestamp attribute is guaranteed to always be present by MISP despite the null guard on line 49 for the `date` processor.

That's right. MISP will always have it and return it (as seen [here](https://github.com/MISP/MISP/blob/01a0e0230f522bc95b4f19eb030dfc52c79e7f15/app/Model/MispAttribute.php#L700-L703) and elsewhere). The only reason it wouldn't be there is if there's a larger, general failure with the event. I think the guard is just following the pattern of the other date processors.

PR #17869, opened 2026-03-18
abnormal_security: defer ingestion of Scanning items in ai_security_mailbox

### LINKED ISSUE #12932: abnormal_security.ai_security_mailbox: items in scanning state are not properly updated
The integration's data source may include items that are not currently classified; they are returned in with `judgementStatus` state of "Scanning".

The current algorithm for collecting documents from the API is (blurring heavily over the details):
```
start ← currently stored last cursor value or value calculated from user configuration if there is no last cursor
end ← now
while we can paginate
    get data from endpoint between start and end
    publish data
cursor ← end
```

These events progress the cursor (even though we do not gain useful information from them), meaning that the integration never revisits the event to obtain it when the `judgementStatus` _has_ been resolved to a useful vales. We could pin the cursor timestamp to before the earliest `judgementStatus`:"Scanning" to ensure that we always retry collecting these documents (falling back to the current time as we currently do if there are none), but this would result in re-ingestion of all documents after that time, including documents that were in a resolved state. We could instead add a look-back time which would be improve the situation probabilistically; we could get arbitrarily close to 100% _also_ at the cost of potential re-ingestion.

Instead of these approaches, I propose that we change the algorithm to maintain a list of work items that are in the scanning state:
```
start ← currently stored last cursor value or value calculated from user configuration if there is no last cursor
end ← now
scanning list ← [] if not present in cursor, other wise cursor scanning list
while we can paginate including items in scanning list
    get data from endpoint between start and end
    if data is in scanning state
        add to scanning list
    else
        publish data
cursor ← {end, scanning list}
```
This has the potential to have unbounded growth on the scanning list in pathological cases (i.e. where abnormal security have a continued behaviour of not moving items out of scanning), so we may want to have a condition to check that to avoid the issue, publishing and error in the case that we exceed some high water; this would require dumping all events in the scanning list and emitting an error to ingest, but the situation could be improved to dumping only the oldest (or newest) items on the scanning list after the two-parameter `tail`, and `front` list functions become available.

efd6:
## Proposed commit message
```
abnormal_security: defer ingestion of Scanning items in ai_security_mailbox

Rewrite the CEL collection program for the ai_security_mailbox data
stream to defer items with judgementStatus "Scanning" instead of
ingesting them immediately. Scanning items are held in a persistent
cursor list and re-checked on each collection cycle. Once the status
resolves, the item is published normally. Items that remain Scanning
beyond a configurable timeout (default 168h) are published as-is.

This prevents the collection cursor from advancing past unclassified
events that would otherwise never be revisited.

Adds script tests covering both the filtering behaviour and the
timeout expiry path.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #12932

### DIFF
--- a/packages/abnormal_security/data_stream/ai_security_mailbox/_dev/test/scripts/scanning_timeout.txt
+++ b/packages/abnormal_security/data_stream/ai_security_mailbox/_dev/test/scripts/scanning_timeout.txt
@@ -10,11 +10,6 @@
 use_stack -profile ${CONFIG_PROFILES}/${PROFILE}
 install_agent -profile ${CONFIG_PROFILES}/${PROFILE} -network_name NETWORK_NAME
 docker_up -profile ${CONFIG_PROFILES}/${PROFILE} -network ${NETWORK_NAME} timeout-mock
-
-# Workaround for https://github.com/elastic/elastic-package/issues/3321.
-exec mkdir -p ${PACKAGE_ROOT}/licenses
-exec cp ${PACKAGE_ROOT}/../../licenses/Elastic-2.0.txt ${PACKAGE_ROOT}/licenses/Elastic-2.0.txt
-
 add_package -profile ${CONFIG_PROFILES}/${PROFILE}
 add_package_policy -profile ${CONFIG_PROFILES}/${PROFILE} test_config.yaml DATA_STREAM_NAME
 
@@ -34,9 +29,6 @@ stdout '^Scanning$'
 remove_package_policy -profile ${CONFIG_PROFILES}/${PROFILE} ${DATA_STREAM_NAME}
 uninstall_agent -profile ${CONFIG_PROFILES}/${PROFILE} -timeout 1m
 docker_down timeout-mock
-
-# Clean up license workaround (https://github.com/elastic/elastic-package/issues/3321).
-exec rm -rf ${PACKAGE_ROOT}/licenses
 
 -- test_config.yaml --
 input: cel
--- a/packages/abnormal_security/data_stream/ai_security_mailbox/_dev/test/scripts/scanning_filtered.txt
+++ b/packages/abnormal_security/data_stream/ai_security_mailbox/_dev/test/scripts/scanning_filtered.txt
@@ -8,11 +8,6 @@
 use_stack -profile ${CONFIG_PROFILES}/${PROFILE}
 install_agent -profile ${CONFIG_PROFILES}/${PROFILE} -network_name NETWORK_NAME
 docker_up -profile ${CONFIG_PROFILES}/${PROFILE} -network ${NETWORK_NAME} scanning-mock
-
-# Workaround for https://github.com/elastic/elastic-package/issues/3321.
-exec mkdir -p ${PACKAGE_ROOT}/licenses
-exec cp ${PACKAGE_ROOT}/../../licenses/Elastic-2.0.txt ${PACKAGE_ROOT}/licenses/Elastic-2.0.txt
-
 add_package -profile ${CONFIG_PROFILES}/${PROFILE}
 add_package_policy -profile ${CONFIG_PROFILES}/${PROFILE} test_config.yaml DATA_STREAM_NAME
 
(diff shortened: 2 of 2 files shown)


packages/abnormal_security/data_stream/ai_security_mailbox/agent/stream/cel.yml.hbs
```diff
@@ -31,90 +61,190 @@ program: |
...
+            (resp.StatusCode == 200 ?
+              bytes(resp.Body).decode_json().as(body,
+                string(body.?judgementStatus.orValue("")) == "Scanning" ?
+                  // Defer to scanning list.
+                  {
+                    "events": [{"retry": true}],
+                    "cursor": {
+                      "last_timestamp": state.end_time,
+                      "scanning": state.?cursor.scanning.orValue([]) + [{"id": state.worklist[int(state.next)], "first_seen": now.format(time_layout.RFC3339)}],
+                    },
```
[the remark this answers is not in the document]
efd6: That should be fixed now.

packages/abnormal_security/data_stream/ai_security_mailbox/_dev/test/scripts/scanning_timeout.txt
```diff
@@ -0,0 +1,128 @@
...
+use_stack -profile ${CONFIG_PROFILES}/${PROFILE}
+install_agent -profile ${CONFIG_PROFILES}/${PROFILE} -network_name NETWORK_NAME
+docker_up -profile ${CONFIG_PROFILES}/${PROFILE} -network ${NETWORK_NAME} timeout-mock
+add_package -profile ${CONFIG_PROFILES}/${PROFILE}
+add_package_policy -profile ${CONFIG_PROFILES}/${PROFILE} test_config.yaml DATA_STREAM_NAME
+
+# The scanning item should appear once the timeout expires and the
+# re-check publishes it. Subsequent cycles re-defer and re-publish
+# the same item, but the _id fingerprint deduplicates it.
+get_docs -profile ${CONFIG_PROFILES}/${PROFILE} -want 1 -confirm 15s -timeout 5m ${DATA_STREAM_NAME}
```
kcreddy, line 19: Is it possible to check there is no document indexed until scanning timeout is reached?

efd6: Can you clarify?

kcreddy: > The scanning item should appear once the timeout expires

> With scanning_timeout set to 1s and a 10s collection
 interval, the item should be deferred in the first cycle, then
 published during the re-check in the second cycle.

Can we test that the scanning item doesn't appear (deferred) until the timeout expires? Possibly checking `get_docs` gives `0` hits in the first cycle?

efd6: OK. Not reliably. The problem is timing: we would need to query after the first cycle completes but before the second cycle's re-check publishes the expired item. With `interval: 10s` and `scanning_timeout: 1s`, that window is essentially zero; the re-check in the second cycle fires immediately after the interval elapses and the item is already expired by then.

We _could_ widen the window by increasing `scanning_timeout` to 30s+ and asserting 0 docs early, but that makes the test fragile (if the agent is slow to start, we would see no document because nothing has run yet, not because the deferral worked) and slow (we'd have to wait for the timeout to actually expire).

The `scanning_filtered` test already proves deferral directly: it asserts exactly 1 document is indexed when the mock returns both a Safe and a Scanning campaign, and fails if the Scanning item leaks through (this can be, and was, verified by running it against the old code). So the deferral behaviour is covered there; this test is specifically exercising the timeout expiry path.

kcreddy: > if the agent is slow to start, we would see no document because nothing has run yet, not because the deferral worked

Yeah, this is the part I was curious about. Thanks for clarification 👍🏻

PR #17871, opened 2026-03-18
socradar_alert: add SOCRadar integration

Radargoger:
Adds a new integration package that collects security incidents and
alarms from the SOCRadar API using the CEL input. The incidents data
stream polls the SOCRadar platform API on a configurable interval,
normalizes results to ECS fields via an ingest pipeline, and includes
a Kibana dashboard for incident visibility.


packages/ti_socradar_feeds/data_stream/feed/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,131 @@
+config_version: 2
+interval: {{interval}}
+{{#if enable_request_tracer}}
+resource.tracer.filename: "../../logs/cel/http-request-trace-*.ndjson"
+request.tracer.maxbackups: 5
```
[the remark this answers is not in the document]
Radargoger: Updated please check again

Radargoger:
Gentle reminder

packages/socradar-alert/data_stream/incidents/manifest.yml
```diff
@@ -0,0 +1,16 @@
...
+    template_path: httpjson.yml.hbs
+    vars:
+      - name: enable_request_tracer
+        type: bool
+        title: Enable request tracing
+        description: Enable HTTP request tracing for debugging
+        multi: false
+        required: false
+        show_user: true
+        default: true
```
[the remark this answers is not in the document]
Radargoger: resolved

Radargoger:
Hi @andrewkroh, sorry for the ping — gentle nudge on this PR. 🙏

All previously raised review comments from the macroscope bot have been addressed and the latest `main` has been merged in. The PR is currently blocked on:

- `buildkite/integrations` — waiting for a maintainer to trigger `/test` (it doesn't run automatically for external contributors).
- The 5 workflows awaiting maintainer approval.
- A code owner review from `elastic/integrations-triaging`.

Whenever you (or someone from the triaging team) get a chance, could you kindly approve the workflows and trigger the build? Happy to address any further feedback. Thanks!

andrewkroh:
What's the relationship between the two threat intel integrations? Do they provide overlapping indicators? When would a user choose one over the other?

All new development should use CEL over HTTP/JSON. You have 2 out of 3 using CEL. Convert the other one over to CEL.

As a reviewer, this is a large changeset. It's actually three separate integration packages. My recommendation is to start with the smallest package and submit it as its own PR. Then apply any learnings to the subsequent PRs so reviewers don’t repeat the same findings three times. This optimizes for review, which these days is the bottleneck. Thanks

I would also recommend looking over https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments or giving it to your agent as context.

Radargoger:
Hi @andrewkroh, thank you for the detailed feedback — much appreciated.

**Split into separate PRs:**
Done. The combined PR has been broken into three standalone PRs:
- #17871 (this PR) — `socradar-alert`: incidents/alarms via CEL
- #18764 — `ti_socradar_feeds`: curated threat feed collections via CEL
- #18765 — `ti_socradar_taxii`: TAXII-based indicator collection via CEL

**httpjson → CEL conversion:**
`socradar-alert` has been converted from httpjson to CEL. All original functionality is preserved: pagination, cursor tracking, and `initial_lookback` on first run.

**Relationship between the two threat intel integrations:**
- `ti_socradar_feeds`: Simple REST poll against SOCRadar's curated feed collections. Best for users who want categorized indicators (IP, domain, hash, URL, email) with minimal setup.
- `ti_socradar_taxii`: TAXII-protocol collection. Best for environments already using TAXII tooling or needing broader indicator types (ASN, X.509, Windows Registry, etc.). The two do share some indicator types (IP, domain, hash) but come from different SOCRadar endpoints and are intended for different deployment contexts.

**Fleet Package Code Review Comments:**
Reviewed. Several findings from that guide have been applied across the three PRs (e.g. `as.number` type `integer` → `long` in taxii, `initial_interval` respected on first run in feeds).

Happy to address any further feedback. Thanks again!

packages/socradar-alert/data_stream/incidents/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,96 @@
...
+        "start_date": [start_date],
+      }.format_query()).with({
+        "Header": {
+          "Api-Key": [state.api_key],
+        }
+      }).do_request().as(resp,
+        resp.StatusCode == 200 ?
+          bytes(resp.Body).decode_json().as(body,
+            body.?data.?alarms.orValue([]).as(alarms,
+              body.?data.?total_pages.orValue(1).as(total_pages,
```
efd6:
```suggestion
```

packages/socradar-alert/data_stream/incidents/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,96 @@
...
+                page < int(total_pages) ?
```
efd6:
```suggestion
                page < int(body.total_pages.orValue(1)) ?
```

packages/socradar-alert/data_stream/incidents/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,96 @@
...
+                  {
+                    "events": alarms.map(a, {"message": a.encode_json()}),
```
efd6:
```suggestion
                    "events": body.?data.alarms.orValue([]).map(a, {"message": a.encode_json()}),
```

packages/socradar-alert/data_stream/incidents/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,96 @@
...
+                    "current_page": page + 1,
+                    "url": state.url,
+                    "api_key": state.api_key,
+                    "limit": state.limit,
+                    "initial_lookback": state.initial_lookback,
+                    "cursor": {"last_requested_at": last_cursor},
+                  }
+                :
+                  {
+                    "events": alarms.map(a, {"message": a.encode_json()}),
```
efd6:
```suggestion
                    "events": body.?data.alarms.orValue([]).map(a, {"message": a.encode_json()}),
```

packages/socradar-alert/data_stream/incidents/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,96 @@
...
+                    "url": state.url,
+                    "api_key": state.api_key,
+                    "limit": state.limit,
+                    "initial_lookback": state.initial_lookback,
+                    "cursor": {
+                      "last_requested_at": now.format("2006-01-02T15:04:05.000000"),
+                    },
+                  }
+              )
+            )
```
efd6:
```suggestion
```

packages/socradar-alert/data_stream/incidents/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,96 @@
...
+            "cursor": state.?cursor.orValue({}),
+          }
+      )
+    ))
+  ))
+processors:
+  - decode_json_fields:
+      fields: ["message"]
+      target: "alarm"
+      overwrite_keys: true
```
efd6: Please don't do this.

packages/socradar-alert/data_stream/incidents/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,44 @@
...
+  - append:
+      field: tags
+      value: preserve_original_event
+      allow_duplicates: false
+  - append:
+      field: error.message
+      value: >-
+        Processor '{{{ _ingest.on_failure_processor_type }}}'
+        {{#_ingest.on_failure_processor_tag}}with tag '{{{ _ingest.on_failure_processor_tag }}}'
+        {{/_ingest.on_failure_processor_tag}}failed with message '{{{ _ingest.on_failure_message }}}'
```
efd6: Final new line.

packages/socradar-alert/data_stream/incidents/fields/base-fields.yml
```diff
@@ -0,0 +1,83 @@
...
+- name: alarm.alarm_default_risk_level
+  type: keyword
+  description: Default risk level (extracted)
+- name: alarm.date
+  type: date
+  format: yyyy-MM-dd HH:mm:ss
+  description: The date when the alarm was created
+- name: alarm.company_id
+  type: keyword
+  description: ID of the company
```
efd6: Final new line.

packages/socradar_alert/docs/README.md
efd6, line 1: This needs to be done in _dev/build/docs, and it needs to be completed.

packages/socradar-alert/img/sample-logo.svg
efd6, line 1: SOCRadar logo?

packages/socradar-alert/img/sample-screenshot.png
efd6, line 1: Dashboard screenshot.

packages/socradar-alert/kibana/dashboard/socradar-1bd8f7b8-b4eb-48aa-8d33-aaa3d1b75790.json
efd6, line 1: Same comments as for the IOC data stream.

packages/socradar-alert/.gitignore
efd6, line 1: Please delete this file.

packages/socradar-alert/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "0.0.1"
+  changes:
+    - description: Initial draft of the package
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1 # FIXME Replace with the real PR link
```
efd6:
```suggestion
      link: https://github.com/elastic/integrations/pull/17871
```

packages/socradar-alert/guide.md
efd6, line 1: This looks like it's the README. It will not be presented to users as this file.

packages/socradar-alert/LICENSE.txt
efd6, line 1: Delete this.

packages/socradar-alert/manifest.yml
```diff
@@ -0,0 +1,73 @@
...
+title: SocRadar Integration
+version: 0.0.1
+description: Collect security incidents and alarms from SocRadar API
+type: integration
+categories:
+  - security
+  - threat_intel
+conditions:
+  kibana:
+    version: ^8.10.0
```
efd6: This is very old.

packages/socradar-alert/manifest.yml
```diff
@@ -0,0 +1,73 @@
...
+          - name: limit
+            type: integer
+            title: Records per page
+            description: Number of alarms to fetch per API call
+            default: 100
+            required: false
+            show_user: false
+
+owner:
+  github: furkankapukayasoc
```
efd6: ?

elastic/security-service-integrations

packages/socradar-alert/sample_event.json
efd6, line 1: Please delete this.

ShourieG:
@vera-review-bot review

Radargoger:
@vera-review-bot review

Radargoger:
Thank you fro your support @ShourieG ! We updated with your feedbacks!

Radargoger:
@vera-review-bot review

Radargoger:
Thank you fro your support @ShourieG ! We updated with your feedbacks!

Radargoger:
@vera-review-bot review

Radargoger:
@vera-review-bot review

Radargoger:
@vera-review-bot review

Radargoger:
@vera-review-bot review

efd6 (commented):
Can you add system tests?

Radargoger:
Thanks @efd6 — system tests have been added for the `incidents` data stream. I added a Docker deploy setup with mocked SOCRadar incidents API responses (`_dev/deploy/docker/`) and a system test config (`data_stream/incidents/_dev/test/system/test-default-config.yml`).

efd6:
This does not have a CODEOWNERS entry. Please add one.

Radargoger:
@efd6 addressed the remaining review items in c5cb510:

- **README**: now generated from `_dev/build/docs/README.md` (complete template) and regenerated so `docs/README.md` is in sync.
- **Final newlines**: added to `base-fields.yml` and `manifest.yml`.

Also picked up the automated-review findings:
- Added a `preserve_original_event` option with a conditional `event.original` removal (default off) so the raw payload isn't stored twice per document.
- Restructured the CEL program to the code-style standard — cursor/window/page pre-bindings lifted out of `state.with()`, and `resp.Body.decode_json()` used directly.
- Fixed `alarm.company_id` mapping (`keyword` → `long`) to match the numeric API value.

Pipeline tests and `elastic-package check` (lint + build) pass locally.

Radargoger:
Hi @efd6 , I think I've worked through all of your review feedback now, and the package is green on pipeline, system and asset tests locally. A couple of the automated-review suggestions I left in place on purpose (happy to walk through the reasoning). Whenever you get a chance, could you take another look? Thanks a lot!

efd6:
If you can respond to the bot comments explaining why you think they're invalid, that would be really helpful.

Radargoger:
Thanks @efd6! Quick notes on the ones I left as-is:

- `event.original` removal — that's the standard `preserve_original_event` flow. An earlier round of the same bot actually asked me to *add* that removal so we don't store the raw payload twice, so keeping it always would double storage. Happy to always-keep it if you'd rather.
- `terminate` — I tried it, but elastic-package lint rejects it (`Additional property terminate is not allowed`), so I kept `drop`.
- The detection rule is intentional — it's what drives the optional Kibana → SOCRadar status sync.
- The 8 unvalidated fields I just removed (5f6ada7).

All green locally. Let me know your call on `event.original`.

efd6:
The terminate comment is correct (supported from 8.16.0/9.0.0 onward). Please make that change. ([ref](https://www.elastic.co/docs/reference/ingest-processor/terminate-processor)).

Radargoger:
Hi @efd6  , updated status:

Everything from the review should be addressed now, and the bot's flags are showing as almost all resolved. Latest fixes:

Status-sync Watcher condition now reads hits.total.value (it was comparing the hits.total object to 0).
Kept event.original unconditionally and simplified the event fields to just event.kind: alert — dropped event.category/event.type to match ti_recordedfuture/ti_flashpoint, so the alarms aren't labelled as ECS threat indicators.
socradar_alert.date now uses date_format: (the format: key isn't applied by package-spec), and the severity script has a description.
One heads-up on terminate: applied as you asked, but the released elastic-package check still rejects it (Additional property terminate is not allowed) since package-spec's processor list hasn't caught up with the 8.16+ runtime — the pipeline/system tests all pass. Happy to handle the lint side differently if you'd prefer.

Two threads on an older commit (f864f12) still show unresolved — event.type: indicator and the 90-day look-back — but both are already fixed in the current version (event.type is gone, look-back is now-10m).

Many thanks!
Burak

packages/socradar_alert/_dev/build/docs/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+- **Centralized alarm visibility** — View all SOCRadar alarms in Kibana Discover and the included dashboard.
+- **Security alerting** — Automatically create Kibana Security Alerts for every alarm using the included Detection Rule.
+- **Status sync to SOCRadar** — Optionally sync alert status changes made in Kibana back to SOCRadar using an Elasticsearch Watcher.
+
+## What do I need to use this integration?
+
+- A valid **SOCRadar API Key**
+- Your **SOCRadar Company ID**
+- Kibana **8.18 or higher**
```
efd6: Remove these; they go stale and the source of truth is the manifest. Also above in the Compatibility section.

packages/socradar_alert/_dev/build/docs/README.md
```diff
@@ -0,0 +1,248 @@
...
+    }
+  },
+  "actions": {
+    "notify_socradar": {
+      "webhook": {
+        "method": "POST",
+        "url": "https://platform.socradar.com/api/company/alarms/status/change",
+        "headers": {
+          "Content-Type": "application/json",
+          "Api-Key": "<your-api-key>"
```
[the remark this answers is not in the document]
efd6: This is a concern.

packages/socradar_alert/data_stream/incidents/_dev/test/pipeline/test-incidents.log
```diff
@@ -0,0 +1,2 @@
+{"alarm_id": "a1b2c3d4-0000-0000-0000-000000000001", "status": "OPEN", "alarm_risk_level": "HIGH", "alarm_text": "A possible phishing domain targeting your brand was detected.", "alarm_asset": "example.com", "notification_id": 1001, "tags": ["phishing", "brand-protection"], "company_id": 330, "is_approved": false, "date": "2024-01-15 10:30:00", "alarm_type_details": {"alarm_sub_type": "Phishing Domain", "alarm_main_type": "Brand Protection", "alarm_generic_title": "Possible Phishing Domain Detected", "alarm_default_risk_level": "HIGH"}, "title": "Possible Phishing Domain Detected", "alarm_type_id": "1001", "alarm_assignees": ["analyst"], "content": {"summary": "A lookalike domain was registered targeting the brand.", "evidence_url": "https://platform.socradar.com/alarm/1001"}, "history": {"OPEN": "2024-01-15 10:30:00"}, "extra": {"source": "brand-protection-engine"}, "alarm_related_assets": {"domains": ["example.com"]}, "alarm_related_entities": {"brands": ["ExampleBrand"]}}
```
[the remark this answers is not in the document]
efd6: Please do this in the system tests.

packages/socradar_alert/data_stream/incidents/_dev/test/pipeline/test-incidents.log
```diff
@@ -0,0 +1,2 @@
+{"alarm_id":"a1b2c3d4-0000-0000-0000-000000000001","status":"OPEN","alarm_risk_level":"HIGH","alarm_text":"A possible phishing domain targeting your brand was detected.","alarm_asset":"example.com","notification_id":1001,"tags":["phishing","brand-protection"],"company_id":330,"is_approved":false,"date":"2024-01-15 10:30:00","alarm_type_details":{"alarm_sub_type":"Phishing Domain","alarm_main_type":"Brand Protection","alarm_generic_title":"Possible Phishing Domain Detected","alarm_default_risk_level":"HIGH"}}
+{"alarm_id":"a1b2c3d4-0000-0000-0000-000000000002","status":"RESOLVED","alarm_risk_level":"MEDIUM","alarm_text":"Leaked credentials referencing your domain were found on a paste site.","alarm_asset":"corp.example.org","notification_id":1002,"tags":["credential-leak"],"company_id":330,"is_approved":true,"approved_by":"analyst","date":"2024-02-03 08:15:42","alarm_type_details":{"alarm_sub_type":"Credential Leak","alarm_main_type":"Digital Risk Protection","alarm_generic_title":"Leaked Credentials Detected","alarm_default_risk_level":"MEDIUM"}}
```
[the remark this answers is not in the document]
efd6: Please do this.

packages/socradar_alert/data_stream/incidents/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,14 @@
...
+vars:
+  api_url: http://{{Hostname}}:{{Port}}
+  api_key: xxxx
+  company_id: "330"
+  initial_lookback: 72h
+  interval: 1m
+  limit: 100
+data_stream:
+  vars:
+    enable_request_tracer: false
```
efd6:
```suggestion
    enable_request_tracer: true
```
This helps in debugging issues.

packages/socradar_alert/data_stream/incidents/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,87 @@
...
+state:
+  api_key: {{api_key}}
+  limit: {{limit}}
+  initial_lookback: {{initial_lookback}}
+redact:
+  fields:
+    - api_key
+program: |
+  state.with(
+  state.?cursor.last_requested_at.orValue("").as(last_cursor,
```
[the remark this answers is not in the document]
efd6: Please address this. The binding to local scopes unnecessarily complicate the code: there is no need to bind `state.?cursor.last_requested_at` to a new label, same for `state.?query_timestamp` and `state.?start_date`. These can all be used directly at the actual use sites, more simply with optional types.

Also, run `celfmt -s` on the program.

Also, don't use sentinel values to indicate absence, use absence.

packages/socradar_alert/data_stream/incidents/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,111 @@
...
+                    :
+                      string(resp.Status) + " (" + string(resp.StatusCode) + ")"
+                  ),
+                },
+              },
+              "want_more": false,
+              "current_page": 1,
+              "query_timestamp": "",
+              "start_date": "",
+              "cursor": state.?cursor.orValue({}),
```
efd6: This is not effective; if the `events` field is an object the cursor is dropped by the input.

packages/socradar_alert/data_stream/incidents/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,128 @@
...
+      tag: append_related_hosts
+      allow_duplicates: false
+      if: ctx.socradar_alert?.alarm_asset != null
+  - script:
+      lang: painless
+      tag: script_set_event_severity
+      description: Maps the SOCRadar alarm risk level to the numeric ECS event.severity used by the detection rule's risk score mapping.
+      if: ctx.socradar_alert?.alarm_risk_level != null
+      source: >-
+        def m = ['CRITICAL': 99L, 'HIGH': 73L, 'MEDIUM': 47L, 'LOW': 21L];
```
efd6: If you have static look-up tables, put them in the `params` option for the script processor; as the code currently is, this map is allocated every execution while if it's in the `params`, it's only allocated at pipeline compilation.

packages/socradar_alert/kibana/dashboard/socradar_alert-1bd8f7b8-b4eb-48aa-8d33-aaa3d1b75790.json
efd6, line 1: Please export this using `elastic-package export` rather than with the kibana API.

efd6 (commented):
> since package-spec's processor list hasn't caught up with the 8.16+ runtime

You can specify the version in the package manifest. You will need to specify at least v3.3.2.

Radargoger:
Hi @efd6 , pushed a few follow-ups:

Ran celfmt -s on the CEL program.
Widened the pipeline tests: added a CRITICAL, asset-less alarm so both the severity mapping and the related.hosts skip get exercised. (The extra alarm-detail fields stay in the system test, as you suggested.)
One I couldn't resolve cleanly: the Watcher setup has users paste their SOCRadar API key straight into the watch, so it ends up in .watches unencrypted. As far as I can tell Watcher has no keystore for webhook headers — is there a pattern you'd prefer here, or does the security note in the README cover it?

Many thanks,
Burak

packages/socradar-alert/data_stream/incidents/agent/stream/cel.yml.hbs
Radargoger: The CEL no longer pre-binds `url`/`api_key`/`limit`/`initial_lookback` into the event state, and JSON decoding moved out of the input's `processors` block into the ingest pipeline's `json` processor. The program was also run through `celfmt -s`.

packages/socradar-alert/data_stream/incidents/elasticsearch/ingest_pipeline/default.yml
Radargoger: Fixed — the file ends with a single trailing newline.

packages/socradar-alert/data_stream/incidents/fields/base-fields.yml
Radargoger: Fixed — trailing newline added.

packages/socradar-alert/img/sample-logo.svg
Radargoger: Replaced the placeholder with the real SOCRadar logo (`img/logo.svg`).

packages/socradar-alert/img/sample-screenshot.png
Radargoger: Replaced the placeholder with an actual dashboard screenshot (`img/socradar-alert-overview.png`).

packages/socradar-alert/kibana/dashboard/socradar-1bd8f7b8-b4eb-48aa-8d33-aaa3d1b75790.json
Radargoger: Re-exported the dashboard with `elastic-package export` so it's in canonical form, and fixed the per-panel `ignoreGlobalFilters`/pinned time ranges and the global-filter data-view reference that were flagged separately. If there's anything IOC-specific still open here, point me to it and I'll fix it.

packages/socradar-alert/.gitignore
Radargoger: Deleted.

packages/socradar-alert/changelog.yml
Radargoger: Applied — the changelog link now points to #17871.

packages/socradar_alert/data_stream/incidents/_dev/test/pipeline/test-incidents.log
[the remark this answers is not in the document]
Radargoger: The extra alarm-detail fields are now exercised in the system test (the mock returns `title`, `alarm_type_id`, `alarm_assignees`, `content`, `history`, `extra`, `alarm_related_assets`, `alarm_related_entities`), and I dropped the pipeline `test-common-config.yml` so nothing config-like lives in the fixtures.

packages/socradar_alert/data_stream/incidents/_dev/test/pipeline/test-incidents.log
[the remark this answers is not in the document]
Radargoger: Added a CRITICAL, asset-less alarm to the fixture so the severity mapping (`CRITICAL` → 99) and the `related.hosts` skip (no `alarm_asset`) are both exercised now, alongside the existing HIGH/MEDIUM docs.

packages/socradar_alert/data_stream/incidents/_dev/test/system/test-default-config.yml
Radargoger: Done — `enable_request_tracer` is now `true`.

packages/socradar_alert/data_stream/incidents/agent/stream/cel.yml.hbs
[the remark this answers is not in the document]
Radargoger: Reworked the program so there are no local-scope bindings for the cursor/timestamps: `cursor.last_requested_at` is read directly via optional access at the use site, and the `query_timestamp`/`start_date` labels are gone. Also ran `celfmt -s`.

packages/socradar_alert/data_stream/incidents/agent/stream/cel.yml.hbs
Radargoger: Fixed — the error branch now emits `events` as an array (`[{...}]`), so the cursor is no longer dropped by the input.

packages/socradar_alert/data_stream/incidents/elasticsearch/ingest_pipeline/default.yml
Radargoger: Moved the risk-level → severity map into the script processor's `params`, so it's allocated once at pipeline compile time rather than on every execution.

packages/socradar_alert/kibana/dashboard/socradar_alert-1bd8f7b8-b4eb-48aa-8d33-aaa3d1b75790.json
Radargoger: Re-exported the dashboard with `elastic-package export` (canonical form).

packages/socradar_alert/_dev/build/docs/README.md
Radargoger: Removed the stale version info: the Compatibility section now only states SOCRadar API v4 compatibility, and the prerequisites no longer pin a Kibana version. I kept the API Key and Company ID as the actual required inputs, happy to drop those too and point to the manifest, or turn them into short how-to-obtain steps, if you'd prefer.

packages/socradar_alert/_dev/build/docs/README.md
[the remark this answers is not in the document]
Radargoger: This is the unencrypted API key in the Watcher `.watches` doc. I couldn't find a keystore mechanism for Watcher webhook headers, so for now I added a security note in the README. I've also raised it in the main PR thread — open to a better pattern if you have one.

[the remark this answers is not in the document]
Radargoger: I looked into this. There's no way to encrypt a custom webhook header in a stored watch — `xpack.watcher.encrypt_sensitive_data` only encrypts `basic_auth.password`, not arbitrary headers, and SOCRadar's API authenticates exclusively via the `Api-Key` header (no HTTP basic auth), so the key can't be routed through the one field Watcher would encrypt.

That said, I think keeping status sync is reasonable here, because the exposure can be scoped down tightly:

- SOCRadar supports **endpoint-scoped API keys**, so a user can create a key authorized for *only* the status-change endpoint instead of a general key with access to all SOCRadar APIs. Even if the watch (and the key in `.watches`) is exposed, access stays limited to that one endpoint — in line with least-privilege.
- `.watches` is itself a restricted system index (reading it needs `manage_watcher`/superuser).
- I've updated the README security note to recommend exactly this: a dedicated key scoped to just this endpoint, plus regular rotation.

Status sync is a core part of the integration's value, so I'd like to keep it in this release on that basis. Does that work for you?

packages/socradar_alert/data_stream/incidents/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,85 @@
...
+      state.url + "?" + {
+        "limit": [string(state.limit)],
+        "include_total_records": ["true"],
+        "show_page_in_response": ["true"],
+        "page": [string(state.?current_page.orValue(1))],
+        "include_alarm_details": ["true"],
+        "include_company_id": ["true"],
+        "start_date": [
+          state.?cursor.last_requested_at.orValue(
+            (now - duration(state.initial_lookback)).format("2006-01-02 15:04:05")
```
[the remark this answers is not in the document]
efd6: @Radargoger, this is something that interacts with how the API works. I would expect that this is a concern. Can you check?

packages/socradar_alert/data_stream/incidents/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,133 @@
...
+        rejected as a duplicate by the data stream's create semantics instead
+        of being indexed a second time.
+  - set:
+      field: ecs.version
+      tag: set_ecs_version
+      value: 9.3.0
+  - set:
+      field: event.kind
+      tag: set_event_kind
+      value: alert
```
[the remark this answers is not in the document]
efd6: Please address this.

packages/socradar_alert/data_stream/incidents/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,85 @@
...
+    ).with(
+      {"Header": {"Api-Key": [state.api_key]}}
+    ).do_request().as(resp,
+      (resp.StatusCode == 200) ?
+        resp.Body.decode_json().as(body,
+          (state.?current_page.orValue(1) < int(body.?data.?total_pages.orValue(1))) ?
+            {
+              "events": body.?data.alarms.orValue([]).map(a, {"message": a.encode_json()}),
+              "want_more": true,
+              "current_page": state.?current_page.orValue(1) + 1,
```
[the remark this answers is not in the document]
efd6: This is a valid concern.
```
-- src.cel --
{
	"mixed": state.?page.orValue(1) + 1,
}
-- data.json --
{
	"page": 1
}
-- out.json --
failed eval: ERROR: <input>:2:34: no such overload
 |  "mixed": state.?page.orValue(1) + 1,
 | .................................^
```

packages/socradar_alert/data_stream/incidents/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,85 @@
...
+            }
+          :
+            {
+              "events": body.?data.alarms.orValue([]).map(a, {"message": a.encode_json()}),
+              "want_more": false,
+              "current_page": 1,
+              "cursor": {"last_requested_at": now.format("2006-01-02 15:04:05")},
```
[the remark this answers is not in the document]
efd6: This is a valid concern.

packages/socradar_alert/_dev/build/docs/README.md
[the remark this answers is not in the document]
efd6: Thanks. That will do.

efd6 (commented):
Please address the following concerns:

- [ ] [Zero-alarm polls don't persist the cursor — next poll replays from `initial_lookback`](https://github.com/elastic/integrations/pull/17871#discussion_r3748516758)
- [ ] [`start_date` is recomputed from `now` on each paginated page — window slides on first multi-page run](https://github.com/elastic/integrations/pull/17871#discussion_r3748195909)
- [ ] [`event.category` and `event.type` are never set](https://github.com/elastic/integrations/pull/17871#discussion_r3748195926)
- [ ] [`current_page` incremented without `int()` cast — fails with `no such overload` after registry restore](https://github.com/elastic/integrations/pull/17871#discussion_r3748195942)
- [ ] [`kibana/index_pattern/` saved object still shipped](https://github.com/elastic/integrations/pull/17871#discussion_r3717680694)
- [ ] [Pipeline test has no `test-common-config.yml` and no error-path fixture — `terminate` branch untested](https://github.com/elastic/integrations/pull/17871#discussion_r3717805372)
- [ ] [`kibana/security_rule/` still in package — flagged and "resolved" four times without removal](https://github.com/elastic/integrations/pull/17871#discussion_r3721577653)

Addressing means either making the required change or providing a reply comment. Please do not resolve concerns unilaterally.

packages/socradar_alert/data_stream/incidents/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
Radargoger: The bot's suggested `event.category: threat` + `event.type: info` fails ECS validation — in ECS 9.3 `event.category: threat` only permits `event.type: indicator`, and elastic-package's `ensureExpectedEventType` hard-fails the mismatch. SOCRadar alarms are heterogeneous (brand protection, digital risk protection, credential leaks, ransomware mentions, phishing domains) with no single valid ECS category, so — matching the alert data streams already in the repo (ti_flashpoint/alert, ti_recordedfuture/triggered_alert & playbook_alert, ti_rapid7_threat_command/alert), which all set only `event.kind: alert` — we currently set `event.kind: alert` and leave `event.category`/`event.type` unset rather than ship an invalid pairing. (ECS only cross-checks category↔type when both are present.)

That's our reading of it — but we're happy to do it whichever way you'd prefer.

packages/socradar_alert/data_stream/incidents/agent/stream/cel.yml.hbs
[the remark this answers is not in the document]
Radargoger: Followed the suggested approach: when the final page has no alarms the program emits a single `{"cursor_placeholder": true}` event so the cursor write is persisted, and the pipeline drops it right after the JSON decode. I also write the cursor as `now - 5m` rather than `now`: an alarm created just before a poll but not yet visible in the API would otherwise fall permanently behind the cursor; the 5m overlap catches those and the fingerprint `_id` discards the re-fetched duplicates. Happy to switch if you'd prefer another idiom for the zero-event cursor write.

packages/socradar_alert/data_stream/incidents/agent/stream/cel.yml.hbs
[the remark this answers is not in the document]
Radargoger: Checked our API: paging is by page number over the `start_date`-filtered result set, so a `start_date` that moves between pages shifts the window and can skip records. Pinned it for the duration of a pagination sequence, following the pattern other paginated CEL inputs in this repo use — the window start is computed once when a new sequence begins (cursor, or `now - initial_lookback` on the first run), carried in state, and reused while `want_more` is true. Verified with a mito eval that a mid-pagination evaluation reuses the pinned value rather than recomputing from `now`.

packages/socradar_alert/data_stream/incidents/agent/stream/cel.yml.hbs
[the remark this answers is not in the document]
Radargoger: Fixed — `current_page` is cast with `int()` at the page parameter, the `total_pages` comparison and the increment, so a JSON-float restore from the registry no longer hits `no such overload`. Thanks for the repro.

packages/socradar_alert/kibana/index_pattern/socradar_alert-6ede9604-546c-4a6e-a7de-97f16c3991f3.json
```diff
@@ -0,0 +1 @@
+{"attributes": {"name": "Incidents", "timeFieldName": "@timestamp", "title": "logs-socradar_alert.incidents-*"}, "id": "socradar_alert-6ede9604-546c-4a6e-a7de-97f16c3991f3", "type": "index-pattern"}
```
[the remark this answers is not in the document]
Radargoger: Dropped the shipped index_pattern. The dashboard now references the managed `logs-*` data view — the convention other log integrations in this repo follow — instead of a package-owned saved object; asset and system tests pass.

packages/socradar_alert/data_stream/incidents/_dev/test/pipeline/test-incidents.log
```diff
@@ -0,0 +1,2 @@
+{"alarm_id":"a1b2c3d4-0000-0000-0000-000000000001","status":"OPEN","alarm_risk_level":"HIGH","alarm_text":"A possible phishing domain targeting your brand was detected.","alarm_asset":"example.com","notification_id":1001,"tags":["phishing","brand-protection"],"company_id":330,"is_approved":false,"date":"2024-01-15 10:30:00","alarm_type_details":{"alarm_sub_type":"Phishing Domain","alarm_main_type":"Brand Protection","alarm_generic_title":"Possible Phishing Domain Detected","alarm_default_risk_level":"HIGH"}}
```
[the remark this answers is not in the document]
Radargoger: Added `test-common-config.yml` with `preserve_original_event`. On the error-path fixture: elastic-package's pipeline tester (`verifyFieldsInTestResult` → `checkErrorMessage`) unconditionally fails any result document carrying a string `error.message` — exactly what the `terminate` branch preserves — so a pipeline error fixture can't pass; another package using the same `data_collection_error`/terminate pattern ships none for the same reason. Covered it in the system test instead: a mock rule returns 500 for a dedicated company id, the CEL takes its error branch, and the pipeline terminates on `data_collection_error`; the indexed document keeps `error.{code,id,message}` and has no `event.kind`/`socradar_alert`, confirming it stopped at `terminate` rather than falling into `on_failure`. Both system configs pass.

packages/socradar_alert/kibana/security_rule/socradar_alert-alarm-detection-rule.json
```diff
@@ -0,0 +1,119 @@
+{
+  "attributes": {
+    "name": "SOCRadar - Alarm Detection",
```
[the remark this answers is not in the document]
Radargoger: Removed the rule from the package as suggested — and with it the dependent "Alert status sync" (Watcher) section of the README, since it only made sense with the bundled rule. This PR is now collection + dashboard, the same shape as the other threat-intel alert integrations in the repo. Syncing alarm status back to SOCRadar will be a separate, standalone follow-up so it can be designed properly rather than bolted onto this integration.

packages/socradar_alert/_dev/build/docs/README.md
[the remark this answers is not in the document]
Radargoger: With the rule removal the whole Watcher/status-sync section is gone from this PR, so this one is resolved by removal — thanks for the earlier look at it.

Radargoger:
Hello @efd6  , 

Thanks, all seven are addressed now, each with a reply on its own thread (I haven't resolved any of them):

1. **Zero-alarm cursor** — fixed. An empty final page now emits a single placeholder event so the cursor write persists, and the pipeline drops it. The cursor is written as `now - 5m` so an alarm that becomes visible slightly late isn't skipped.
2. **Sliding `start_date`** — fixed. Checked our API: paging is by page number over the `start_date`-filtered result set, so the window start is now pinned for the whole pagination sequence and carried in state.
3. **`event.category` / `event.type`** — replied. `threat` + `info` fails ECS validation (`threat` only permits `indicator`), and the alarms are heterogeneous with no single valid category, so we set only `event.kind: alert`. Happy to do it differently if you'd prefer.
4. **`current_page` `int()` cast** — fixed at the page parameter, the `total_pages` comparison and the increment.
5. **`kibana/index_pattern/`** — removed; the dashboard now references the managed `logs-*` data view.
6. **Pipeline tests** — `test-common-config.yml` added. A pipeline error-path fixture can't pass (the tester fails any result document carrying a string `error.message`, which is exactly what `terminate` preserves), so the terminate branch is covered by a system test that mocks a 500 instead.
7. **`kibana/security_rule/`** — removed, together with the dependent "Alert status sync" (Watcher) section of the README. The package is now collection + dashboard; syncing alarm status back to SOCRadar will be a separate follow-up.

The full suite (asset, pipeline, and both system configs) passes against a freshly created stack.

Radargoger:
The CI failure was the dashboard's migration stamp rather than the package logic: it had been re-exported from a 9.x Kibana at some point, which left `typeMigrationVersion: 10.3.0` on the saved object. CI boots the stack at the oldest version the manifest supports (8.18.0), where the highest known dashboard version is 10.2.0, so the package could not be installed there at all — the run died at installation rather than in a test.

Lowered the stamp back to 10.2.0 (`65452d4c`) and re-verified against a real 8.18.0 stack: check, installation, asset, pipeline, static and both system configs all pass. Ready for another `/test` whenever you get a chance.

Radargoger:
Hi @efd6 
Thank you for the test

we’ve made updates based on your findings. Could you please run another /test when you get a chance?

Best Regards.

packages/socradar_alert/data_stream/incidents/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,68 @@
+---
+description: Pipeline for processing alarm events
+processors:
```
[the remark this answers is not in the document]
efd6: I don't see `event.category` or `event.type` being set.

packages/socradar_alert/data_stream/incidents/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
efd6: Yeah, that's fair. I think this is a gap in ECS.

packages/socradar_alert/data_stream/incidents/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,95 @@
...
+      tag: set_alarm_generic_title
+      ignore_empty_value: true
+      ignore_failure: true
+  - set:
+      field: alarm.alarm_default_risk_level
+      copy_from: alarm.alarm_type_details.alarm_default_risk_level
+      tag: set_alarm_default_risk_level
+      ignore_empty_value: true
+      ignore_failure: true
+  - remove:
```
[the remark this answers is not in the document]
efd6: Please to this.

packages/socradar_alert/data_stream/incidents/agent/stream/cel.yml.hbs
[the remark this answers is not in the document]
efd6: Conventionally, we use either `{"message": "retry"}` and a drop in the ingest pipeline, or `{"retry": true}` and a beats processor in the input definition in the processors block.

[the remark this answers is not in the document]
Radargoger: Switched to the conventional form: the program now emits `{"message": "retry"}` on a poll that returns no alarms, and the ingest pipeline drops it. The drop sits before the rename/JSON decode, since the marker isn't JSON. Cursor persistence is unaffected — the event is still published, it just never gets indexed.

packages/socradar_alert/data_stream/incidents/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
Radargoger: Done — the main processor list now ends with the conditional `append` of `preserve_original_event` when `error.message` is set. The trailing `remove` of `event.original` had already been dropped, so `event.original` is always kept.

packages/socradar_alert/data_stream/incidents/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,150 @@
...
+      params:
+        risk_severity:
+          CRITICAL: 99
+          HIGH: 73
+          MEDIUM: 47
+          LOW: 21
+      source: >-
+        def v = params.risk_severity.get(ctx.socradar_alert.alarm_risk_level);
+        if (v != null) { ctx.event = ctx.event == null ? [:] : ctx.event; ctx.event.severity = v; }
+  - date:
```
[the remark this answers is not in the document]
Radargoger: Added the guard — the `date` processor now runs only when `socradar_alert.date` is present, so an alarm without a date keeps the collection timestamp and still gets the rest of the pipeline instead of becoming a `pipeline_error`. `ignore_failure: false` stays, so a date that is present but malformed still surfaces. Covered by a new pipeline fixture: the date-less alarm comes out with `event.kind: alert`, its severity, and its flattened alarm-type fields.

Radargoger:
Hi @efd6,

Thanks for the test run and feedback! Both points are addressed:

* The `{"message": "retry"}` placeholder is dropped before rename/JSON decoding, without affecting cursor persistence.
* The main processor list now ends with a conditional `append` of `preserve_original_event` for documents with errors.

Missing dates now retain the collection timestamp and continue through the pipeline, while malformed dates still raise errors. A fixture covers the missing-date case.

The full suite passed on stack 8.18.0, our minimum supported version. Three commits have landed since the last CI run. Could you trigger another `/test` when you have a moment? Thanks in advance.

Please let me know if anything still needs attention.
BG

efd6 (approved):
Suggested commit message:
```
socradar_alert: add SOCRadar integration

Adds a new integration package that collects security incidents and
alarms from the SOCRadar API using the CEL input. The incidents data
stream polls the SOCRadar platform API on a configurable interval,
normalizes results to ECS fields via an ingest pipeline, and includes
a Kibana dashboard for incident visibility.
```

Can you provide text for the origin of the test samples? (e.g. constructed from documentation, obtained from a live instance [with sanitization], ... etc.)

Radargoger:
Hello @efd6 

Thanks for continous support! 
The samples come from a live instance — a demo tenant, not a customer account — and were sanitized before being committed. The response envelope, field names and alarm-type taxonomy are as the API returns them; anything identifying was replaced: alarm IDs with sequential placeholders (`a1b2c3d4-0000-…`), monitored assets and brands with RFC 2606 example domains and placeholder names (`example.com`, `corp.example.org`, `ExampleBrand`), approver and assignee names with a generic `analyst`, and notification IDs and dates with fixed values.

Thanks again for the review. Anything else you'd like from me here?

Issue #17403, opened 2026-02-12
Add support for an additional level of directories under packages/


jsoriano:
In the context of composable packages project, and with the introduction of OTel packages, there are going to be more packages related to the same technology.

Allow to have an additional level of directories, so packages can be grouped by technology. See a POC in https://github.com/elastic/integrations/pull/16966.

The main changes will be:
* Allow to store packages with an additional nesting level under the packages directory, so for example the `nginx_input_otel` package can be stored under `packages/nginx`, as `input_otel`, along with other packages related to nginx.
* Code owners can be defined at the level of a directory containing multiple packages for the same technology. So if the `packages/nginx` directory contains multiple packages related to NGINX, the `CODEOWNERS` file only needs to specify a single entry, for `packages/nginx`, instead of an entry for each package. Each package still needs to define its owner accordingly in the manifest.
* It should be still possible to define more specific owners, so it is possible to define a different owner for a single package, even if the rest of packages for the same technology are owned by another team.
* In PRs, selective triggering tests per package should also trigger tests of dependent packages.
* Packages can be released individually from their directory, independently of its nested position.
* Packages will be grouped by technology with the help of their maintainer teams.
* Previous model should keep working.

This approach relies on existing features in `elastic-package`, and only requires changes in the scripts in the integrations repository. These changes need to take into account backwards compatibility with backport branches.

- [x] Update dev tools to validate package names and codeowners. https://github.com/elastic/integrations/pull/18543
- [ ] Update Buildkite scripts to trigger tests, publish and create backport branches. https://github.com/elastic/integrations/pull/18577/
- [ ] Update github workflow to automate the creation of the issue templates. https://github.com/elastic/integrations/pull/18549
    - https://github.com/elastic/integrations/blob/a710cd1d0e0edaf562ddf86a41f1a9ac9d395320/.github/workflows/updatecli/updatecli.d/sync-packages-to-bug-issue-template.yml#L18-L43

mrodm:
Once the support for the additional level of directories is added, it needs to be revisited also the CI workflows/pipelines that rely on just one level of directories.
As example of one of these workflows: https://github.com/elastic/integrations/pull/17879

Check other Github Workflows in case they need to be updated too.

theletterf:
@alaudazzi @jmikell821 @bmorelli25 FYI. Docs locations could change.

alaudazzi:
@theletterf @jsoriano 
Is this going to impact the content in the integrations-docs repo?

[...]

PR #17879, opened 2026-03-18
[ci] Add blocking PR check for package docs using docs-builder

theletterf:
## Summary

- Adds `.github/workflows/validate-package-docs.yml` — a new blocking PR check that runs `elastic/docs-builder` (metadata-only + strict mode) against any `packages/**/docs/*.md` files changed in a PR
- Generates a temporary `docset.yml` at the repo root listing only the changed files, so no static manifest needs to be maintained
- Excludes `_dev/build/` paths (those contain `{{ fields }}` template syntax not valid for docs-builder)
- Exits early if no package docs changed, so unrelated PRs pay zero cost

## Why

Vale linting already runs but is non-blocking (`fail_on_error: false`). Badly formatted Markdown or invalid docs-builder directives can slip into `packages/*/docs/README.md` files and break downstream consumers (EPR / `integration-docs` repo) silently. This check catches those issues at PR time.

## Test plan

- [ ] Open a PR modifying a `packages/*/docs/README.md` with intentionally broken Markdown — confirm `Validate Package Docs (docs-builder)` check appears and fails
- [ ] Fix the Markdown — confirm the check passes
- [ ] Open a PR touching only non-docs files — confirm the check is skipped

### DIFF
--- a/.github/workflows/validate-package-docs.yml
+++ b/.github/workflows/validate-package-docs.yml
@@ -4,7 +4,8 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
     paths:
-      - 'packages/**/docs/*.md'
+      - 'packages/*/docs/*.md'
+      - 'packages/*/*/docs/*.md'
 
 permissions:
   contents: read


theletterf:
+CC @elastic/docs-engineering @elastic/integration-docs

shmsr (approved):

.github/workflows/validate-package-docs.yml
```diff
@@ -0,0 +1,53 @@
+name: Validate Package Docs
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'packages/**/docs/**'
```
mjwolf: This pattern will also catch file that are in the knowledge_base directory, which is used for [AI assistant knowledge](https://github.com/elastic/package-spec/blob/main/spec/integration/docs/spec.yml#L14-L22), not just the files that become published documentation.

I'm not sure if that matters or would cause any problems though

theletterf:
Good catch — fixed in the latest commit. Both the path trigger and the grep filter now exclude subdirectories under `docs/`: the trigger uses `packages/**/docs/*.md` (single-level `*` instead of `**`) and the grep uses `[^/]*\.md$` to reject any path with a slash after `docs/`, so `knowledge_base/` and similar subdirs are ignored.

theletterf:
@elastic/ecosystem Please have a look, thanks!

theletterf:
@elastic/ecosystem Could you have a look / approve?

.github/workflows/validate-package-docs.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Find changed package docs
+        id: changed-docs
+        run: |
+          FILES=$(git diff --name-only origin/${{ github.base_ref }}...HEAD \
+            | grep '^packages/.*/docs/[^/]*\.md$' \
```
mrodm, line 27: This would be needed to be updated once this issue is done: https://github.com/elastic/integrations/issues/17403

Once that issue is completed, there could be packages at different levels. There is an example in this PoC https://github.com/elastic/integrations/pull/16966

theletterf: Good heads-up! Both the path trigger (`packages/**/docs/*.md`) and the grep filter (`^packages/.*/docs/[^/]*\.md$`) already handle nested package paths like `packages/nginx/nginx/docs/README.md`, so this should work without changes once #17403 lands.

theletterf: @mrodm Keep me honest here though! :)

theletterf: Actually I was wrong — `packages/**/docs/*.md` doesn't match nested paths. Fixed in the latest commit: now uses two explicit patterns (`packages/*/docs/*.md` and `packages/*/*/docs/*.md`) so both current and nested package structures (per #17403) are covered. The grep filter was already fine since `.*` in regex has no depth limit.

mrodm: Thanks!
Checking again this regex and I think it works as it should, once it is added the other level of directories:
```
 $ echo packages/nginx/nginx_otel/docs/README.md |  grep '^packages/.*/docs/[^/]*\.md$'
packages/nginx/nginx_otel/docs/README.md
```

PR #17668, opened 2026-03-04
Initial commit of Dataminr Pulse integration for elastic

sgunanathan:
This is the very first integration for Dataminr Pulse with Elastic for cyber alerts.
We have created an integration that uses Dataminr Pulse V4 APIs and pull Dataminr alerts periodically into Elastic.
The integration creates a data stream with Dataminr alert fields mapped to ECS specific and custom fields. The ReadMe lists all the field mappings done for the integration.
The integration also provides a couple of dashboards to view the results.

## Checklist
- [ x] I have reviewed [tips for building integrations](https://www.elastic.co/docs/extend/integrations/tips-for-building) and this pull request is aligned with them.
- [ x] I have verified that all data streams collect metrics or logs.
- [ x] I have added an entry to my package's `changelog.yml` file.
- [ x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ x] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally
One can run static, pipeline, asset and policy tests locally to verify the integration.
Commands:
elastic-package test static
elastic-package test pipeline --data-streams alerts
elastic-package test asset
elastic-package test policy

## Screenshots
Screenshot of Integration set up page:
Screenshot 2026-03-04 at 4 47 50 PM

Screenshot of dashboards:
Screenshot 2026-03-04 at 4 41 50 PM
Screenshot 2026-03-04 at 4 46 13 PM

PR #17886, opened 2026-03-18
Initial commit of Dataminr Pulse integration for elastic

pgifford-dataminr:
This is a new PR that covers the same changes as this [PR](https://github.com/elastic/integrations/pull/17668). The original submitter on that PR is no longer at the company so I am opening this new PR and the other one can be closed.

This is the very first integration for Dataminr Pulse with Elastic for cyber alerts.
We have created an integration that uses Dataminr Pulse V4 APIs and pull Dataminr alerts periodically into Elastic.
The integration creates a data stream with Dataminr alert fields mapped to ECS specific and custom fields. The ReadMe lists all the field mappings done for the integration.
The integration also provides a couple of dashboards to view the results.

## Checklist

One can run static, pipeline, asset and policy tests locally to verify the integration.
Commands:
elastic-package test static
elastic-package test pipeline --data-streams alerts
elastic-package test asset
elastic-package test policy

## Screenshots

Screenshot of Integration set up page:
Screenshot 2026-03-18 at 12 43 23 PM

Screenshot of dashboards:
Screenshot 2026-03-18 at 12 43 38 PM
Screenshot 2026-03-18 at 12 43 51 PM


efd6:
The issues raised by the review bot are valid. Please address them.

packages/dataminr_pulse/data_stream/alerts/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1064 @@
...
+      source: |
+        boolean drop(Object o) {
+          if (o == null || o == '') {
+            return true;
+          } else if (o instanceof Map) {
+            ((Map) o).values().removeIf(v -> drop(v));
+            return (((Map) o).size() == 0);
+          } else if (o instanceof List) {
+            ((List) o).removeIf(v -> drop(v));
+            return (((List) o).length == 0);
```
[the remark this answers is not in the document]
TimofteA: Addressed

packages/dataminr_pulse/data_stream/alerts/agent/stream/stream.yml.hbs
```diff
@@ -0,0 +1,331 @@
...
+                          "fetched_alerts": has(api_body.alerts) ? size(api_body.alerts) : 0,
+                          "fetch_timestamp": fetch_ts
+                        }.encode_json()
+                      }],
+                    "next_cursor": has(api_body.nextPage) && api_body.nextPage != "" ? string(api_body.nextPage) : "",
+                    "access_token": auth_result.new_state.access_token,
+                    "expires_at": auth_result.new_state.expires_at,
+                    "last_fetch_timestamp": fetch_ts,
+                    "retry_count": 0,
+                    "want_more": false
```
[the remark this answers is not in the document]
TimofteA: Addressed

packages/dataminr_pulse/_devtests/program.cel
```diff
@@ -0,0 +1,223 @@
...
+                                "http_status_code": api_resp.StatusCode,
+                                "cursor_used": prep.updated_state.next_cursor != "" ? prep.updated_state.next_cursor : "initial",
+                                "fetched_alerts": has(response_body.alerts) ? size(response_body.alerts) : 0,
+                                "fetch_timestamp": fetch_ts
+                              }.encode_json()
+                            }],
+                          "next_cursor": has(response_body.nextPage) && response_body.nextPage != "" ? string(response_body.nextPage) : "",
+                          "access_token": prep.updated_state.access_token,
+                          "expires_at": prep.updated_state.expires_at,
+                          "last_fetch_timestamp": int(fetch_ts.getDate()),
```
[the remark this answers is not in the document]
TimofteA: Addressed

packages/dataminr_pulse/data_stream/alerts/agent/stream/stream.yml.hbs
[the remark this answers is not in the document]
TimofteA: Even after the code changes in the seconds commit I cannot resolve this thread. Clicking "Resolve thread" results in an error: "Failed to resolve thread".

Image

TimofteA (approved):

packages/dataminr_pulse/data_stream/alerts/agent/stream/stream.yml.hbs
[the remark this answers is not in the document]
efd6: This has been fixed.

packages/dataminr_pulse/_devtests/program.cel
efd6, line 1: Please delete this.

.github/CODEOWNERS
```diff
@@ -204,6 +202,7 @@
 /packages/cyera @elastic/security-service-integrations
 /packages/cylance @elastic/security-service-integrations
 /packages/darktrace @elastic/security-service-integrations
+/packages/dataminr_pulse @elastic/security-external-integrations
```
efd6:
```suggestion
/packages/dataminr_pulse @elastic/security-service-integrations
```

packages/dataminr_pulse/data_stream/alerts/agent/stream/stream.yml.hbs
```diff
@@ -0,0 +1,331 @@
...
+  access_token: ""
+  expires_at: 0
+  last_fetch_timestamp: 0
+  retry_count: 0
+redact:
+  fields:
+    - client_id
+    - client_secret
+    - access_token
+program: |
```
efd6, line 41: I think this is an unmanageably large CEL program. Can you describe what the intention of the program is? In general, CEL programs should get the data from the API/data source, split it into parts if necessary, translate it to JSON if necessary and then send it to ingest, possibly as a JSON encoded string, possibly as a JSON message. It should not do the majority of the processing of the data.

packages/dataminr_pulse/_devtests/state.json
efd6, line 1: Please delete this.

packages/dataminr_pulse/config/kibana.yml
efd6, line 1: Please delete this.

packages/dataminr_pulse/data_stream/alerts/_dev/deploy/docker/files/mock_server.py
efd6, line 1: Can you use https://github.com/elastic/stream instead. You can see its use in action in many other integrations.

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-alert-events.json
efd6, line 1: If this is renamed to be a .log file, then the unquoted strings that are the values for the "message" fields in the objects in the "events" array can be used. This is easier to maintain.

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-common-config.yml
efd6, line 1: This should not be necessary.

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-entity-events.json
efd6, line 1: Same situation here; rename to a .log file and just have the message strings.

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-fingerprint-events.json
efd6, line 1: And here.

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-log-events.json
efd6, line 1: And here.

packages/dataminr_pulse/data_stream/alerts/_dev/test/policy/test-custom-config.expected
```diff
@@ -0,0 +1,64 @@
...
...next_cursor : \"initial\",\n                          \"fetched_alerts\": 0,\n                          \"fetch_timestamp\": fetch_ts,\n                          \"response_body\": string(api_resp.Body)\n                        }\n                      ).encode_json()\n                    }],\n                    \"access_token\": auth_result.new_state.access_token,\n                    \"expires_at\": auth_result.new_state.expires_at,\n                    \"next_cursor\": (api_resp.StatusCode == 400 || api_resp.StatusCode == 404) ? \"\" : auth_result.new_state.next_cursor,\n                    \"retry_count\": is_retryable ? (has(state.retry_count) ? int(state.retry_count) : 0) + 1 : 0,\n                    \"want_more\": is_retryable\n                  })\n                )\n              )\n            )\n          )\n        )\n      :\n        auth_result.new_state.with({\n          \"events\": [auth_result.auth_event],\n          \"retry_count\": (has(auth_result.want_more) && auth_result.want_more) ? (has(state.retry_count) ? int(state.retry_count) : 0) + 1 : 0,\n          \"want_more\": has(auth_result.want_more) ? auth_result.want_more : false\n        })\n    )\n  )\n)\n"
```
efd6: That the program is quoted indicates that there is trailing whitespace on one of the lines. Please remove it. But also, I really think the program needs to be simplified.

packages/dataminr_pulse/data_stream/alerts/_dev/test/scripts/pipeline_simulate.txt
efd6, line 1: I don't think we need a script test for this.

packages/dataminr_pulse/data_stream/alerts/fields/ecs.yml
efd6, line 1: This is not necessary since ECS fields are automatically identified.

packages/dataminr_pulse/data_stream/alerts/manifest.yml
```diff
@@ -0,0 +1,121 @@
...
+        description: Preserves a raw copy of the original event, added to the field `event.original`
+        default: false
+        required: false
+        show_user: true
+      - name: processors
+        type: yaml
+        title: Processors
+        multi: false
+        required: false
+        show_user: false
```
efd6: Please add final new line.

packages/dataminr_pulse/img/screenshot.png
efd6, line 1: Normally these would be dashboard screenshots.

packages/dataminr_pulse/changelog.yml
```diff
@@ -0,0 +1,5 @@
+- version: "1.0.0"
+  changes:
+    - description: Initial release
+      type: enhancement
+      link: https://github.com/elastic/integrations/issues/1
```
efd6:
```suggestion
      link: https://github.com/elastic/integrations/issues/17886
```

packages/dataminr_pulse/scripts/setup-fleet.sh
efd6, line 1: Please delete this.

packages/dataminr_pulse/changelog.yml
```diff
@@ -0,0 +1,5 @@
+- version: "1.0.0"
```
efd6:
```suggestion
- version: "0.1.0"
```
Suggest not releasing to GA in initial release.

packages/dataminr_pulse/manifest.yml
```diff
@@ -0,0 +1,30 @@
...
+policy_templates:
+  - name: dataminr_pulse
+    title: Dataminr Pulse Alerts
+    description: Collect real-time alerts from Dataminr Pulse API
+    inputs:
+      - type: cel
+        title: Dataminr Pulse Alerts via CEL
+        description: Collect alerts using Common Expression Language (CEL) input
+owner:
+  github: elastic/security-external-integrations
```
efd6:
```suggestion
  github: elastic/security-service-integrations
```

packages/dataminr_pulse/manifest.yml
```diff
@@ -0,0 +1,30 @@
+format_version: 3.0.3
+name: dataminr_pulse
+title: Dataminr Pulse
+version: 1.0.0
```
efd6:
```suggestion
version: 0.1.0
```

packages/dataminr_pulse/docker-compose.yml
efd6, line 1: Please delete this.

packages/dataminr_pulse/kibana/dashboard/dataminr_pulse-3364fad8-deb4-40cf-9dfe-50bf4d5a1823.json
efd6, line 1: Suggest having links widget between the two dashboards. https://www.elastic.co/docs/explore-analyze/visualize/link-panels

packages/dataminr_pulse/data_stream/alerts/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1064 @@
...
+      field: event.kind
+      tag: set_pipeline_error_to_event_kind
+      value: pipeline_error
+  - append:
+      field: error.message
+      value: >-
+        Processor {{{_ingest.on_failure_processor_type}}}
+        with tag '{{{_ingest.on_failure_processor_tag}}}'
+        in pipeline {{{_ingest.pipeline}}}
+        failed with message: {{{_ingest.on_failure_message}}}
```
efd6: Add a final new line.

packages/dataminr_pulse/data_stream/alerts/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1064 @@
...
+        - dataminr_pulse.discovered_entity.name
+        - dataminr_pulse.discovered_entity.products
+        - dataminr_pulse.discovered_entity.publishedDate
+        - dataminr_pulse.discovered_entity.type
+        - dataminr_pulse.entity_type
+        - dataminr_pulse.intel_agents_raw
+      tag: remove_unwanted_fields
+      ignore_missing: true
+
+on_failure:
```
efd6, line 666:
```suggestion
  - set:
      field: event.kind
      value: pipeline_error
      tag: set_pipeline_error_into_event_kind
      if: ctx.error?.message != null
  - append:
      field: tags
      value: preserve_original_event
      allow_duplicates: false
      if: ctx.error?.message != null
on_failure:
```
This ensures that errors in processors that have their own `on_failure` option that populates the `error.message` array are still handled helpfully.

packages/dataminr_pulse/data_stream/alerts/agent/stream/stream.yml.hbs
TimofteA: This CEL program handles three concerns — all of which must happen agent-side:

  1. Authentication — OAuth token management against /auth/2/token, with automatic refresh before expiry and retry logic for transient failures.
  2. API fetch + pagination + error handling — GET requests to /pulse/v1/alerts with cursor-based pagination, 401 token-regeneration, and retryable/fatal error classification.
  3. Document splitting — Fans out each alert into multiple events by entity type. This must happen in CEL because the ingest pipeline processes one document at a time and cannot turn one document into many. The logic is a simple router:
    - One event per vulnerability, threatActor, or malware entity (with the entity attached)
    - One combined event for all url/ipAddress entities
    - One combined event for any other entity types
    - A fallback event if no entities or no intelAgents exist

packages/dataminr_pulse/data_stream/alerts/_dev/test/policy/test-custom-config.expected
TimofteA: The CEL program has already been simplified in this iteration.
The remaining complexity is auth management, API fetch with pagination/error handling, and a minimal document splitter — all of which must live in CEL. The ingest pipeline now handles all data processing.

TimofteA:
I can't mark the comments above as resolved - getting "Failed to resolve thread". Not sure if this is expected and only certain users can mark them as resolved.

pgifford-dataminr:
@efd6 Any chance we could get another look on this?

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-entity-events.log-expected.json
efd6, line 1: Can you add a "preserve_original_events" tag to the pipeline test? This makes it easier to review confidently.

packages/dataminr_pulse/manifest.yml
```diff
@@ -0,0 +1,30 @@
...
+    title: Dataminr Pulse Logo
+    size: any
+    type: image/svg+xml
+type: integration
+categories:
+  - security
+  - threat_intel
+conditions:
+  kibana:
+    version: "^8.13.0"
```
efd6: Why are we supporting such an old stack?

packages/dataminr_pulse/manifest.yml
efd6: This has not been changed.

packages/dataminr_pulse/manifest.yml
```diff
@@ -0,0 +1,30 @@
...
+  - name: dataminr_pulse
+    title: Dataminr Pulse Alerts
+    description: Collect real-time alerts from Dataminr Pulse API
+    inputs:
+      - type: cel
+        title: Dataminr Pulse Alerts via CEL
+        description: Collect alerts using Common Expression Language (CEL) input
+owner:
+  github: elastic/security-external-integrations
+  type: elastic
```
efd6: Add final new line.

packages/dataminr_pulse/data_stream/alerts/manifest.yml
```diff
@@ -0,0 +1,121 @@
...
+            text: "15 minute"  
+      - name: page_size
+        type: select
+        title: Page Size
+        description: Maximum number of alerts to return per request (default 40, max 100)
+        default: "40"
+        required: true
+        multi: false
+        show_user: true
+        options:
```
efd6: Is there a reason not to just use text or integer?

packages/dataminr_pulse/data_stream/alerts/manifest.yml
```diff
@@ -0,0 +1,121 @@
...
+        secret: false
+        show_user: false
+      - name: token_expiry_buffer_ms
+        type: integer
+        title: Token Expiry Buffer (milliseconds)
+        description: Buffer time in milliseconds before token expiry to trigger refresh (default 300000 = 5 minutes)
+        default: 300000
+        required: true
+        show_user: false
+        secret: false
```
efd6: Make this a duration and calculate the milliseconds in the CEL using `getMilliseconds` (see https://github.com/google/cel-spec/blob/master/doc/langdef.md).

efd6 (commented):
I have a patch file that I would like you to look at for an alternative approach.

packages/dataminr_pulse/manifest.yml
```diff
@@ -0,0 +1,30 @@
+format_version: 3.0.3
```
efd6: This needs to be updated to handle the `terminate` processor.
```suggestion
format_version: 3.3.2
```

packages/dataminr_pulse/data_stream/alerts/manifest.yml
```diff
@@ -0,0 +1,82 @@
...
+        default: https://userauth.dataminr.com/auth/2/token
+        required: true
+        show_user: false
+      - name: token_expiry_buffer
+        type: text
+        title: Token Expiry Buffer
+        description: How long before the token expires to refresh it (e.g. 5m, 30s).
+        default: 5m
+        required: true
+        show_user: false
```
efd6, line 36: This is being flagged as requiring a secret marker because of a sub-string match with "token". I think that package-spec should understand that a duration is very unlikely to be a secret, but in the meantime we can do one of two things, mark this as `secret: false` or change the name to not include the "token" sub-string.

cc @jsoriano

pgifford-dataminr: Added secret: false

packages/dataminr_pulse/data_stream/alerts/agent/stream/stream.yml.hbs
```diff
@@ -1,10 +1,10 @@
 config_version: 2
 interval: {{interval}}
 max_executions: 10
```
kcreddy: This is significantly lower than the default value of [`1000`](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-cel#max_executions-cel). Is this intended?

packages/dataminr_pulse/data_stream/alerts/agent/stream/stream.yml.hbs
```diff
@@ -0,0 +1,166 @@
...
+            ).do_request().as(resp,
+              (resp.StatusCode == 401) ?
+                {
+                  "events": [
+                    {
+                      "error": {
+                        "message": "GET " + req_url + ": 401 Unauthorized",
+                      },
+                    },
+                  ],
```
kcreddy, line 113: `events` is a single element array, does it not keep Fleet input health non-DEGRADED? Is this intended, so that user is supposed to check the 401 from the ingested docs?

packages/dataminr_pulse/data_stream/alerts/agent/stream/stream.yml.hbs
```diff
@@ -18,244 +18,135 @@ resource.url: {{url}}
...
-                          }
+            ).do_request().as(resp,
+              (resp.StatusCode == 401) ?
+                {
+                  "events": [
+                    {
+                      "error": {
+                        "message": "GET " + req_url + ": 401 Unauthorized",
+                      },
+                    },
```
kcreddy, line 112: `events` is a single element array, does it not keep Fleet input health non-DEGRADED? Is this intended, so that user is supposed to check the 401 from the ingested docs?

packages/dataminr_pulse/data_stream/alerts/fields/fields.yml
```diff
@@ -1,267 +1,120 @@
...
+  description: Estimated event location.
   fields:
     - name: location
       type: geo_point
-      description: Geo-point location (lat/lon)
+      description: Geo-point location (lat/lon).
     - name: name
       type: keyword
-      description: Location name
+      description: Location name.
```
kcreddy: I'm not sure if this was flagged before and discussed in the previous review comments - but `geo` fields are supposed to be only nested, and not at the root level.

https://www.elastic.co/docs/reference/ecs/ecs-geo
> Note also that the geo fields are not expected to be used directly at the root of the events.

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-entity-events.log-expected.json
```diff
@@ -55,7 +49,7 @@
             },
             "event": {
                 "category": [
-                    "vulnerability"
+                    "web"
```
kcreddy: This change seems buggy, as the original event seems to be indicating a vulnerability.

packages/dataminr_pulse/manifest.yml
```diff
@@ -14,7 +14,7 @@ categories:
   - threat_intel
 conditions:
   kibana:
-    version: "^8.13.0"
+    version: "^8.19.0"
```
kcreddy, line 17: Is it also supposed to support `9.x`?

packages/dataminr_pulse/data_stream/alerts/manifest.yml
armina-dm: The reason was to limit the possibility of strange extreme values the user could enter for api polling times and page size since we couldn't find a way to validat the input. Is there an alternative way to validate the input?

packages/dataminr_pulse/manifest.yml
armina-dm: I am not aware of any specific versioning restrictions. However, we encountered issues when working with the dashboards due to inconsistencies between the Kibana version installed via elastic-package stack up and the Kibana version on our Cloud instance. Making dashboards work cross-platform proved to be more complicated than expected.

I attempted to create the dashboards in the UI and then use the elastic-package export dashboards command. I also recall finding a script in the documentation that was supposed to generate dashboard JSONs without object references, making them ready to add to the project, but this approach failed as well.

Is there documentation available on best practices for creating dashboards and integrating them into the project? Specifically, guidance on:
    - Which Kibana version to use for dashboard development
    - How to ensure cross-version compatibility
    - The recommended workflow for exporting and preparing dashboards for integration packages
    
  CC: @efd6 (Similar question to the one you posed before)

packages/dataminr_pulse/data_stream/alerts/fields/fields.yml
armina-dm: moved to source.geo.location

packages/dataminr_pulse/data_stream/alerts/agent/stream/stream.yml.hbs
armina-dm: Regarding the 401 handling ,the current implementation is intentional and designed for auto-recovery. When a 401 occurs, the integration clears the expired token, creates an error document as an audit trail, and continues running. On the next interval, it detects the missing token, re-authenticates automatically, and resumes polling. This allows the integration to self-heal from transient auth failures without manual intervention.

packages/dataminr_pulse/data_stream/alerts/agent/stream/stream.yml.hbs
armina-dm: Regarding the 401 handling ,the current implementation is intentional and designed for auto-recovery. When a 401 occurs, the integration clears the expired token, creates an error document as an audit trail, and continues running. On the next interval, it detects the missing token, re-authenticates automatically, and resumes polling. This allows the integration to self-heal from transient auth failures without manual intervention.

packages/dataminr_pulse/data_stream/alerts/agent/stream/stream.yml.hbs
armina-dm: Not as far as I am aware. Changed back to default

packages/dataminr_pulse/data_stream/alerts/manifest.yml
armina-dm: Seems addressed by the patch

packages/dataminr_pulse/manifest.yml
armina-dm: done

packages/dataminr_pulse/manifest.yml
armina-dm: Responded in a more recent comment

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-entity-events.log-expected.json
armina-dm: Added a new suite of pipeline tests with preserve_original_events in the tags

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-entity-events.log-expected.json
armina-dm: Identified and fixed

packages/dataminr_pulse/manifest.yml
efd6: @armina-dm If you generate the dashboards in the lowest version that you are requiring in the `kibana.version` constraint, it will work in later versions. To do this, ensure that you are running the stack version (using `elastic-package`) that is the minimum version specified in the manifest.

packages/dataminr_pulse/data_stream/alerts/_dev/test/policy/test-custom-config.expected
```diff
@@ -0,0 +1,178 @@
...
+                        }
+                    )
+                  :
+                    {
+                      "token": state.access_token,
+                      "expires": state.expires_at,
+                    }
+                ).as(auth,
+                  has(auth.error) ?
+                    {"events": {"error": {"message": auth.error}}}
```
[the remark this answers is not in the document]
efd6: @armina-dm Is this intentional?

packages/dataminr_pulse/data_stream/alerts/sample_event.json
```diff
@@ -0,0 +1,63 @@
...
+            ]
+        }
+    },
+    "geo": {
+        "location": {
+            "lat": 35.263265,
+            "lon": -80.854384
+        },
+        "name": "Mecklenburg County, NC, USA"
+    }
```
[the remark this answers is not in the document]
efd6: @armina-dm Was this file generated using `elastic-package test system -g`?

packages/dataminr_pulse/data_stream/alerts/fields/fields.yml
```diff
@@ -0,0 +1,123 @@
...
+  - name: geo
+    type: group
+    description: Estimated event location.
+    fields:
+      - name: location
+        type: geo_point
+        description: Geo-point location (lat/lon).
+      - name: name
+        type: keyword
+        description: Location name.
```
kcreddy: Since these are ECS fields, these explicit field definitions can be removed from here. 
All ECS mappings are already added using component template.

packages/dataminr_pulse/manifest.yml
kcreddy: +1 to above comment from Dan.
You can for example if your minimum targeted stack version is `8.19.0`, then run: `elastic-package stack up --version=8.19.0 -d -v` and export the dashboards. These exported dashboard should also work with `9.x` versions.

But for your integration to appear in the `9.x` stack versions you need to add `|| ^9.0.0` to the `manifest.yml` kibana version constraints:

```yaml
conditions:
  kibana:
    version: "^8.19.0 || ^9.0.0"
```

packages/dataminr_pulse/data_stream/alerts/fields/fields.yml
armina-dm: Ok, thank you. Removed from fields definition

packages/dataminr_pulse/data_stream/alerts/sample_event.json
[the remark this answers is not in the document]
armina-dm: It doesn't look like it was. Now it is. Regenerated, and updated with most recent push.

packages/dataminr_pulse/data_stream/alerts/_dev/test/policy/test-custom-config.expected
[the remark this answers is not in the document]
armina-dm: It looks unintentional. Made the change.

packages/dataminr_pulse/manifest.yml
armina-dm: @efd6 How can I check what version of Kibana is on the cloud instance? Is it the one you updated in the patch, 8.19? That would be the logical restriction in my opinion as a minimum version requirement to create them on. 

Regarding the dashboard generation, I tried to follow the following flow for their creation:
- elastic stack up with version 8.19
- created a simple dashboard
- Added the version constraints to the manifest
- elastic package export dashboards -d dashboard_id
- added a filter to pass validation and created a build
- Installing the package locally working fine -> rendering the dashboard correctly. 
- Installing the package on the cloud instance -> there are cross version rendering issues, and I was unable to find the right version to avoid this inconsistency. 

For example, a simple Control element. The control added to on the local stack and worked fine, then I installed the package on a cloud project and its just not there anymore. (attaching screnshots with this control example; no data in the second photo is OK, there isnt any flowing on the project) 

The main problem is the inconsistency between what I can create and what will render because of the version, and not being able to identify which Kibana version would save this headache. 

Im rather sure there's has to be a better way that i am simply unaware of on how to avoid these issues. 

image
image

efd6: > How can I check what version of Kibana is on the cloud instance?

I would suggest don't. Use `elastic-package` and start with a principled position of what to support. I would suggest `^8.19.0` since it's the earliest currently supported version.

armina-dm: Ok, so we'll keep it as is was updated by the patch. Thank you!

Then, for further reference: Is there any way to avoid these cross platform rendering problems?

In current version everything is OK we made sure every element of our dashboards is working OK on all platforms, this is just a question to ease our future developments.

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-entity-events.log-expected.json
efd6: Conventionally, this is done by using a test config, for example, [here](https://github.com/elastic/integrations/blob/main/packages/abnormal_security/data_stream/ai_security_mailbox/_dev/test/pipeline/test-common-config.yml).

packages/dataminr_pulse/data_stream/alerts/manifest.yml
efd6: No, there is not, but a failure should be presented to the user early. What is the API response to a value that is out of range?

armina-dm: The API enforces a maximum of 100 alerts per page regardless if the page size value specified when making the request is greater than that. Multiple polling requests in short succession don't break the API, but according to the API documentation: "Polling every 30 seconds to 30 minutes is recommended depending on your system's alert latency tolerance and user needs around real-timeness." I improved the description to include this recommendation for polling time

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-entity-events.log-expected.json
armina-dm: added common config for pipeline tests

packages/dataminr_pulse/data_stream/alerts/manifest.yml
armina-dm: After conversing with my team, I learned we use options for our other connectors so customers don't get confused or make too many requests by mistake. I made a new commit and re-added the options based on the recommended values by product.

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-common-config.yml
```diff
@@ -0,0 +1,3 @@
+fields:
+  tags:
+    - preserve_original_event
```
efd6: Add final new line.

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-field-mappings.json
efd6, line 1: Can we have these as a log test (literal JSON-encoded objects as lines in a .log file) rather than a .json test? See test-alert-events.log for the preferred pattern.

The reason for this is that that format of test is easier to maintain.

packages/dataminr_pulse/data_stream/alerts/manifest.yml
efd6: @cpascale43 Do you have a view on this?

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-field-mappings.json
```diff
@@ -0,0 +1,30 @@
...
+
+        }
+    ]
+}
```
efd6: Add final new line.

packages/dataminr_pulse/data_stream/alerts/_dev/test/scripts/pipeline_simulate.txt
efd6, line 1: We don't need script tests for pipeline testing here; the same validation can be done in a standard pipeline test.

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-common-config.yml
armina-dm: done in most recent commit

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-field-mappings.json
armina-dm: done in most recent commit

packages/dataminr_pulse/data_stream/alerts/_dev/test/pipeline/test-field-mappings.json
armina-dm: done in most recent commit

packages/dataminr_pulse/data_stream/alerts/_dev/test/scripts/pipeline_simulate.txt
armina-dm: removed in most recent commit

armina-dm:
@efd6 can we please /test again? Had to push to include the regenerate of the pipeline tests.

efd6 (approved):
Thanks

efd6:
Will wait for @kcreddy before merge.

kcreddy (approved):
LGTM. 
@efd6 please feel free to merge.

PR #17892, opened 2026-03-19
o365: add upgrade test

efd6:
This tests that an already populated index can have the integration upgraded and accept more documents successfully.

## Proposed commit message
```
o365: add upgrade test

This tests that an already populated index can have the integration
upgraded and accept more documents successfully.
```

> [!NOTE]
> No user-facing change, so no changelog entry.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Depends elastic/elastic-package#3388

### DIFF
--- a/packages/o365/data_stream/audit/_dev/test/scripts/upgrade.txt
+++ b/packages/o365/data_stream/audit/_dev/test/scripts/upgrade.txt
@@ -1,0 +1,358 @@
+# Test that upgrading from the latest EPR version to the dev version does not
+# break data collection.
+#
+# Installs the EPR version from the registry, creates a policy on it, verifies
+# data collection, then installs the dev version, upgrades the policy, and
+# verifies the upgraded integration collects new data.
+
+[!external_stack] skip 'Skipping external stack test.'
+[!exec:jq] skip 'Skipping test requiring absent jq command'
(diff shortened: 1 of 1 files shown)


kcreddy (approved):
LGTM. I've done a few tests that simulates the upgrade path:
- `dev` with a version bump - Passed ✅ 
- `dev` with no version bump (no changes to manifest) - Skipped ✅ 
- Breaking-change - Skipped ✅ 

There is one scenario I would like to flag (not a blocker): 

The `[breaking_change] skip` guard is the right call — you can't meaningfully upgrade across a breaking change without manual intervention. But it does leave a gap: there's no automated way to verify that a breaking change *works* at all when a user follows the documented upgrade path (reinstall, reconfigure, etc.).

Could there be a future variant of `upgrade_package_latest` that would let breaking-change tests verify that:

1. The old data stream remains searchable after the new version is installed fresh.
2. The new version collects data successfully with its new schema.
3. No mapping conflicts arise between old and new backing indices.

This wouldn't simulate a real upgrade (since breaking changes require user reconfiguration), but it would catch mapping conflicts and data stream continuity issues that currently have no test coverage for major version bumps.

efd6:
Yeah, I'm not sure what we can do for breaking changes in the general case. I think there we just have to go back to manual testing.

PR #17899, opened 2026-03-19
[azure_application_insights] add client secret

jakubgalecki0:
## Proposed commit message

This pull request adds support for client secret authentication https://github.com/elastic/beats/pull/48880

- Tested new auth method 
- Tested backward compatibility with api_key auth

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- 

## Screenshots

image

image

#### API KEY 

image

image


packages/azure_application_insights/manifest.yml
lucabelluccini: I think this constraint should be refined to the versions of Beats which include the api key change onwards https://github.com/elastic/beats/pull/48880.

packages/azure_application_insights/manifest.yml
```diff
@@ -20,7 +20,7 @@ categories:
   - observability
 conditions:
   kibana:
-    version: "^8.13.0 || ^9.0.0"
+    version: "^8.19.14 || ^9.2.8"
```
muthu-mps: As we have back-ported to 8.19.0 and 9.2.0. Should we use them instead?
CC: @tommyers-elastic

packages/azure_application_insights/_dev/build/docs/README.md
```diff
@@ -8,7 +8,15 @@ The Application Insights Integration allows users to retrieve application insigh
 
 `Application ID`:: (_[]string_) ID of the application. This is Application ID from the API Access settings blade in the Azure portal.
 
-`Api Key`:: (_[]string_) The API key which will be generated. See [Azure Monitor Log Analytics API Overview](https://learn.microsoft.com/en-us/azure/azure-monitor/logs/api/overview) for more information.
+`Authentication Type`:: (_string_) Optional. The authentication method to use. Accepted values: `api_key` or `client_secret`. Defaults to `api_key` if not set.
+
```
muthu-mps, line 12: Can we add a note in the README highlighting that users currently authenticating via API Key should migrate to Client Secret?

lucabelluccini: ++

packages/azure_application_insights/manifest.yml
```diff
@@ -20,7 +20,7 @@ categories:
   - observability
 conditions:
   kibana:
-    version: "^8.13.0 || ^9.0.0"
+    version: "^8.19.0 || ^9.2.0"
```
lucabelluccini: This would be actually:
- >= 8.19.31
- >= 9.2.8
- >= 9.3.3

zmoog: We probably need something like this:

```yaml
conditions:
  kibana:
    version: "~8.19.14 || ~9.2.8 || ^9.3.3"
```

lucabelluccini: Except if we add a comment on the authentication method that selecting `Authentication Type` is allowed only on 8.19.31, 9.2.8, 9.3.3 onwards

packages/azure_application_insights/manifest.yml
```diff
@@ -30,12 +30,46 @@ vars:
     multi: false
     required: true
     show_user: true
+  - name: auth_type
+    type: select
+    title: Authentication Type
+    description: The authentication type to use. Either "api_key" or "client_secret".
```
lucabelluccini: Can we put a warning that this setting can be used only on 8.19.14, 9.2.8, 9.3.3 onwards?

`kibana.version` is only a "proxy" of the recommended version and checks only Kibana.
Maybe it is better to be explicit here.

lucabelluccini:
```suggestion
    description: Only for Elastic Agent 8.19.14, 9.2.8, 9.3.3 or more recent: the authentication type to use. Either "api_key" or "client_secret".
```

This is a suggestion, not the actual "text" to display.

packages/azure_application_insights/manifest.yml
zmoog: we should use these constraints:

```yaml
conditions:
  kibana:
    version: "~8.19.14 || ~9.2.8 || ^9.3.3"
```

for `major.minor.patch`: 

- the `~` modifier allows newer **patch** releases
- the `^` modifier also allows new **minor** releases

packages/azure_application_insights/_dev/build/docs/app_insights.md
```diff
@@ -6,7 +6,15 @@ The Application Insights Integration allows users to retrieve application insigh
 
 `Application ID`:: (_[]string_) ID of the application. This is Application ID from the API Access settings blade in the Azure portal.
 
-`Api Key`:: (_[]string_) The API key which will be generated, more on the steps here https://dev.applicationinsights.io/documentation/Authorization/API-key-and-App-ID.
+`Authentication Type`:: (_string_) Optional. The authentication method to use. Accepted values: `api_key` or `client_secret`. Defaults to `api_key` if not set.
```
zmoog: nit: should we flip the default auth to `client_secret` for new installs? Microsoft is discontinuing API key support soon, right?

packages/azure_application_insights/_dev/build/docs/README.md
```diff
@@ -8,7 +8,17 @@ The Application Insights Integration allows users to retrieve application insigh
...
+NOTE: API Key authentication is deprecated. Users currently authenticating via API Key are encouraged to migrate to Client Secret authentication. Client Secret authentication uses Azure Active Directory (Azure AD) and provides more robust security through short-lived tokens.
```
zmoog: Is MSFT still actively using the "Azure Active Directory"? I guess they moved to Entra ID naming now.

packages/azure_application_insights/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: "1.10.0-next"
+  changes:
+    - description: Add support for Azure AD client secret authentication. New optional parameters `auth_type`, `client_id`, `client_secret`, and `tenant_id` have been added. `api_key` is now optional.
```
zmoog: As before, i suggest to replace "Azure AD" with Entra ID.

packages/azure_application_insights/manifest.yml
```diff
@@ -30,12 +30,46 @@ vars:
...
+    secret: true
+    multi: false
+    required: false
+    show_user: true
+  - name: tenant_id
+    type: text
+    title: Tenant ID
+    description: The tenant ID for Azure AD authentication.
+    multi: false
+    required: false
```
zmoog: I suggest replacing Azure AD with Entra ID.

PR #17908, opened 2026-03-19
[Watchguard]: Pipeline failure on traffic logs

### LINKED ISSUE #17863: [Watchguard]: Pipeline failure on traffic logs
### Integration Name

WatchGuard Firebox [watchguard_firebox]

### Dataset Name

watchguard.log

### Integration Version

1.6.0

### Agent Version

9.3.0

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.3.0

ECH + Ubuntu Agent

### Software/API Version

n/a

### Error Message

An error is received on `logs-watchguard_firebox.log-1.6.0-pipeline_traffic` indicating an issue with the processor tagged `src_dst_community_id`.  The specific error is:
```
could not convert string [ip] to transport protocol
```

### Event Original

This appears to trigger when the connection is based on the `esp` protocol as opposed to `tcp`/`udp` sessions.
```
<142>Mar 17 14:20:26 Firebox_M670 (2026-03-17T20:20:26) firewall: msg_id="3000-0151" Allow Firebox External_UPN esp 1.2.3.4 168.179.100.100 geo_src="USA" geo_dst="USA" duration="697" sent_pkts="2" rcvd_pkts="67635" sent_bytes="1600" rcvd_bytes="7220344" (Any From Firebox-00)
```

### What did you do?

n/a

### What did you see?

There are docs flagged as `pipeline_error` indicating unexpected processing.

No `pipeline_error` messages.

### Anything else?

The pipeline definition has a condition to run only if the source/dest port is `!=0`, however in this case the source/dest port is null, and therefore should also not trigger this processor.

The pipeline needs to be updated to perform null checking on the source ip/port fields.

qcorporation:
## Summary

Fixes ingest pipeline failures on ESP traffic logs where community_id ran with missing or zero ports (Painless treats null != 0 as true).

## Changes

pipeline_traffic.yml: Tighten the community_id processor if so it only runs when source/destination IP and ports are present and ports are non-zero.
fields.yml: Add watchguard_firebox.log.geo_src (keyword) so pipeline output from KV-parsed geo_src="..." passes field validation.

## Tests

Append an ESP traffic sample to test-traffic.log; regenerate test-traffic.log-expected.json with elastic-package test pipeline --generate.

## Checklist

## Related issues
- Closes #17863
Bumps package to 1.6.2 (changelog: bugfix).


packages/watchguard_firebox/data_stream/log/elasticsearch/ingest_pipeline/pipeline_traffic.yml
```diff
@@ -1145,7 +1145,13 @@ processors:
...
   - community_id:
       tag: src_dst_community_id
-      if: 'ctx.watchguard_firebox.log.source_port != 0 && ctx.watchguard_firebox.log.destination_port != 0'
+      if: >-
+        ctx.watchguard_firebox?.log?.source_ip != null
+        && ctx.watchguard_firebox?.log?.destination_ip != null
+        && ctx.watchguard_firebox?.log?.source_port != null
+        && ctx.watchguard_firebox.log.source_port != 0
+        && ctx.watchguard_firebox?.log?.destination_port != null
+        && ctx.watchguard_firebox.log.destination_port != 0
```
taylor-swanson: Remaining `?` are probably redundant and can be removed.

Also, it's odd to me that we are running it against the vendor fields and not the ECS fields...

qcorporation: @taylor-swanson sorry for the delay. I've updated it to remove the redundant checks.
To answer your question regarding not using the ECS fields: The reason we switched to vendor fields is that the ECS fields (source.ip, destination.ip, source.port, destination.port) get turned into arrays by append processors earlier in the pipeline — NAT/control values (ctl_dst_ip, ctl_src_ip, ctl_dst_port, ctl_src_port, host_dest_ip, etc.) are appended to them. The community_id processor requires scalar values, which is exactly what caused the original errors

Issue #15295, opened 2025-09-11
Input packages: index template patterns don’t match dotted datastream names


leandrojmp:
Similarly to other input packages, we want to add the ability to customize the dataset name for better data organization.

eg https://github.com/elastic/integrations/blob/main/packages/winlog/manifest.yml#L33

<details>

<summary>Edited, original issue below</summary>

### Integration Name

Custom macOS Unified Logs [unifiedlogs]

### Integration Version

0.4.0

### Agent Version

It is not relevant

### Agent Output Type

elasticsearch

### Elasticsearch Version

It is not relevant

### OS Version and Architecture

It is not relevant

### Error Message

The Index Template for the UnifiedLogs integration is not using the correct index pattern, the template `logs-unifiedlogs` is using the index pattern as `logs-unifiedlogs-*`, this index pattern is wrong and does not follows the data stream naming scheme of integrations.

By using the wrong index pattern, the data from the integration will match the catch all index template `logs`, and will ignore both the integration custom template `logs-unifiedlogs@custom` and also the integration ingest pipeline, it will also not have the correct `index.default_pipeline` and `index.final_pipeline` set.

<img width="1065" height="988" alt="Image" src="https://github.com/user-attachments/assets/f21d0ca1-483e-4111-a194-d7dc5b9ee558" />

### What did you do?

Add the Custom UnifiedLogs integration to an ingest pipeline.

### What did you see?

The index pattern does not match the data stream name

### What did you expect to see?

The index pattern should match the correct data stream name, so it needs to be changed to `logs-unifiedlogs.*`.

</details>

marc-gr:
`unifiedlogs` is an input package, so it does not have datastreams. This means the index template pattern is as expected and should not cause issues. Have you experienced any?

OTOH we allow to customize the datastream in other input packages for better organization, so I still think this is a valid enhancement to do.

I will edit this issue to this ER if that is fine.

leandrojmp:
Hello @marc-gr 

As mentioned in the original issue, the template does not work, it does not match the datastream created by the integration, I'm not sure what you mean with it does not have datastreams.

Also the changed description is unrelated to the original issue, the user already can customize the dataset name.

When configuring the integration you select the data stream type and the data set name, so it will create a datastream following the naming scheme for integration datastreams.

<img width="704" height="312" alt="Image" src="https://github.com/user-attachments/assets/0bdf3571-5c05-4708-a937-c7921b4b85d4" />

[...]

PR #17924, opened 2026-03-20
Input packages: dataset var and dotted index patterns

marc-gr:
## Summary

Addresses **[#15295](https://github.com/elastic/integrations/issues/15295)** for the **Custom macOS Unified Logs (`unifiedlogs`)** input package by making the **dataset** configurable and visible in Fleet, and by ensuring the agent input config receives **`data_stream.dataset`** (same idea as other input packages, e.g. winlog).

### Changes (unifiedlogs only)

- **`manifest.yml`**: add **`data_stream.dataset`** (default `unifiedlogs.generic`, required, user-visible).
- **`agent/input/unifiedlogs.yml.hbs`**: emit **`data_stream.dataset`** under **`data_stream:`** so Fleet-generated policies pass the dataset through to the agent.
- **`changelog.yml`**: version bump + entry (see PR).
- **Policy tests** (`_dev/test/policy/`): default dataset vs custom dataset (`elastic_agent.test`) to lock in the generated Fleet/agent policy shape.

## Related

- Closes / relates to: [#15295](https://github.com/elastic/integrations/issues/15295)


packages/unifiedlogs/agent/input/unifiedlogs.yml.hbs
```diff
@@ -1,4 +1,6 @@
 condition: ${host.platform} == 'darwin'
+data_stream:
+  dataset: {{data_stream.dataset}}
```
jsoriano, line 3: I think this is the only change needed here to fix unifiedlogs.

packages/unifiedlogs/manifest.yml
```diff
@@ -127,6 +127,16 @@ policy_templates:
...
         description: |
           The Ingest Node pipeline ID to be used by the integration.
+      - name: data_stream.dataset
+        type: text
+        title: Dataset name
+        description: >-
+          Dataset to write data to. Changing the dataset will send the data to a different index.
+          You can't use `-` in the name of a dataset and only valid characters for
+          Elasticsearch index names are allowed.
+        default: unifiedlogs.unifiedlogs
```
jsoriano: Defining the variable is only needed in principle if we want or need to replace the default value. The default value before this change should be the policy template name, `unifiedlogs`, what should be fine.

packages/filestream/elasticsearch/index_template/template.json
```diff
@@ -0,0 +1,7 @@
+{
+  "index_patterns": [
+    "logs-filestream-*",
+    "logs-filestream.*-*"
+  ],
+  "data_stream": {}
+}
```
jsoriano: We should not need to define index patterns in the package, this should be handled by Fleet when creating policies for this package.

packages/unifiedlogs/elasticsearch/index_template/template.json
jsoriano, line 1: Should not be needed.

packages/winlog/elasticsearch/index_template/template.json
jsoriano, line 1: Should not be needed.

marc-gr:
Yes I realized it was missing after the fact. This feels like prone to error though, would be interesting to enforce it somehow

belimawr (approved):

PR #17932, opened 2026-03-20
fix(qualys_vmdr): use join for ID_SET in KB response error path

andrewkroh:
## Proposed commit message

```
The CEL expression at line 244 calls string() on the result
of ID_SET.orValue([]) which is a list. When the knowledge
base API response is missing expected ID values, this path
triggers string(list) which has no overload.

Replace string() with join(" ") to properly convert the list
to a string for the error message
```

## Checklist

### DIFF
--- a/packages/qualys_vmdr/changelog.yml
+++ b/packages/qualys_vmdr/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "6.17.1"
+  changes:
+    - description: Fix intermittent evaluation error in asset_host_detection when the knowledge base API response is missing expected ID values.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/17932
 - version: "6.17.0"
   changes:
     - description: Add navigation links across all dashboards.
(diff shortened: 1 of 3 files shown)


packages/qualys_gav/data_stream/asset/agent/stream/cel.yml.hbs
```diff
@@ -142,13 +142,13 @@ program: |
               rate_limit(
                 headers.with(
                   {
-                    "x-ratelimit-Reset": [string(headers[?"x-ratelimit-towait-sec"][0].orValue("3600"))],
+                    "x-ratelimit-Reset": [headers[?"x-ratelimit-towait-sec"].orValue(["3600"]).map(v, string(v)).join("")],
```
kortschak: I don't think this is the cause. I think the problem is at https://github.com/elastic/integrations/blob/e5b6ee9c63772b101f2d9d18e2b0c28a124ece21/packages/qualys_vmdr/data_stream/asset_host_detection/agent/stream/input.yml.hbs#L244

Which should be
```
"response_error": "response IDs values are missing: " + kb_body.?doc.KNOWLEDGE_BASE_VULN_LIST_OUTPUT.RESPONSE.ID_SET.orValue([]).join(" "),
```
or something similar.

The code for the limit reset is correct AFAICS.

andrewkroh: Thanks! This solution was way off the mark. Re-based with that as the only fix.

efd6 (approved):
LGTM as an alter ego. You may want another approval.

PR #17936, opened 2026-03-21
Feature/openclaw integration

oldcodeoberyn:
- Enhancement

## Proposed commit message

**WHAT:** 
- Introduced the `sessions` data stream for the OpenClaw integration.
- Implemented an Elasticsearch ingest pipeline to parse OpenClaw session logs and extract key observability metrics, including session operations, token consumption (prompt/completion), and LLM cost.
- Added a comprehensive Kibana dashboard (`OpenClaw Overview`) to visualize session activities, usage trends, and token costs.
- Included pipeline test cases (`test-sessions.log` and expected JSON outputs) to ensure reliable log parsing.

**WHY:** 
- To provide full visibility into AI agent (OpenClaw) operations. 
- This initial phase focuses on helping administrators and developers monitor session behavior, audit token consumption, and accurately track LLM usage costs within their Elasticsearch clusters.

## Checklist

## Author's Checklist

- [x] Pipeline tests pass successfully (`elastic-package test pipeline`).
- [x] Dashboards render correctly with the newly extracted token and cost fields.
- [x] Package builds successfully (`elastic-package build`).

1. Navigate to the integration directory:
   ```bash
   cd packages/openclaw
   ```
2. Run pipeline tests to verify the ingest pipeline logic against the provided log samples:
   ```bash
   elastic-package test pipeline
   ```

## Related issues
 ⦁N/A
## Screenshots
image


.github/CODEOWNERS
```diff
@@ -378,6 +377,7 @@
 /packages/okta @elastic/security-service-integrations
 /packages/openai @elastic/obs-infraobs-integrations
 /packages/opencanary @elastic/security-service-integrations
+/packages/openclaw @elastic/ecosystem
```
kpollich: Howdy from @elastic/ecosystem - Can we discuss more about the future of this integration and how you're expecting ownership to work here? This is the first we're seeing this and we're on the hook at codeowners here. We don't use or understand OpenClaw enough to be considered subject matter experts and maintain this package.

oldcodeoberyn: Hi @kpollich , Thanks for your reply. Actually, I’m not quite sure how to determine the ownership for an integration like this. Although I’m an Elastican, I see this more as a community-contributed integration that should be maintained by the community going forward. It’s not a formal, production-grade integration, so it doesn’t need to be treated with the same strict ownership rules.
However, there doesn’t seem to be a matching category in the current CODEOWNERS rules. Do you think assigning it to @elastic/obs-infraobs-integrations would be more appropriate? Should I update the CODEOWNERS entry to this team?

kpollich: I'm not sure either who the official GitHub team should be here in this case. The ecosystem team owns developer tooling for integrations, so it doesn't seem appropriate here. 

Some thoughts
- `type: community` is already set for this package which is correct in this case
- If we don't intend this integration to ever be supported in production, it should exist as a version lower than 1.0.0 and be classified as prerelease/beta.

oldcodeoberyn: OK, let me change the version and tag it to @elastic/obs-infraobs-integrations since it is now provide the visibility of openclaw session log

PR #18032, opened 2026-03-25
Lowercase Vidyard thumbnail URLs in package docs

theletterf:
## Summary

- Lowercases Vidyard video thumbnail image URLs in 4 packages: `crowdstrike`, `m365_defender`, `microsoft_defender_endpoint`, `sentinel_one`
- Updates both `docs/` and `_dev/build/docs/` files

## Why

docs-builder rejects image filenames with uppercase characters. When integration-docs downloads these thumbnails from `play.vidyard.com`, the mixed-case filenames (e.g. `VKKWSpg4sDEk1DBXATkyEP.jpg`) cause the build to fail.

Vidyard serves thumbnails case-insensitively, so lowercasing the URLs is safe.

See: https://github.com/elastic/integration-docs/actions/runs/23537107911/job/68515036838?pr=1152

### DIFF
--- a/packages/crowdstrike/changelog.yml
+++ b/packages/crowdstrike/changelog.yml
@@ -2,7 +2,7 @@
 - version: "3.12.1"
   changes:
     - description: Lowercase Vidyard video thumbnail URL for docs-builder compatibility
-      type: enhancement
+      type: bugfix
       link: https://github.com/elastic/integrations/pull/18032
 - version: "3.12.0"
   changes:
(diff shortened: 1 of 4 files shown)


packages/crowdstrike/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.12.1"
+  changes:
+    - description: Lowercase Vidyard video thumbnail URL for docs-builder compatibility
+      type: enhancement
```
kcreddy: This should be `bugfix`

kcreddy: same for the rest

theletterf: Fixing

theletterf: Fixed in https://github.com/elastic/integrations/pull/18032/commits/4d275e5481a0f667be12dac91e769b27477f7cf9

PR #18053, opened 2026-03-25
[ci] Fix docs validation notification: use Jobs API instead of artifacts

theletterf:
## Summary

Reworks the notification mechanism for the package docs validation workflow. When docs-builder finds errors, a comment is posted on the PR mentioning `@elastic/integration-docs`.

## How it works

1. **Validation workflow** (`validate-package-docs.yml`): When docs-builder exits non-zero, uploads a `docs-validation-failed` artifact containing the PR number. Uses `upload-artifact@v7`. Stays non-blocking (`continue-on-error: true`) with `contents: read` only.

2. **Notification workflow** (`notify-package-docs-failure.yml`): Triggered by `workflow_run`. Downloads the artifact via `download-artifact@v8` with `run-id` and `github-token` for cross-workflow access. If artifact exists, posts a comment. Permissions at job level (`pull-requests: write`, `actions: read`).

## Security

- Validation workflow: `pull_request` trigger, `contents: read` only — safe for forks
- Notification workflow: `workflow_run` trigger, runs in base branch context — safe for forks
- Permissions at job level per least-permissive approach
- All GitHub context values passed via env vars, not inline interpolation

## Testing

Artifact upload confirmed in test PR #18049.

### DIFF
--- a/.github/workflows/notify-package-docs-failure.yml
+++ b/.github/workflows/notify-package-docs-failure.yml
@@ -5,51 +5,46 @@ on:
     workflows: ["Validate Package Docs"]
     types: [completed]
 
-permissions:
-  pull-requests: write
-  actions: read
-
 jobs:
   notify:
     name: Comment on PR
     if: github.event.workflow_run.event == 'pull_request'
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      actions: read
     steps:
-      - name: Check for validation failure
-        id: check
-        env:
-          GH_TOKEN: ${{ github.token }}
-          REPO: ${{ github.repository }}
-          RUN_ID: ${{ github.event.workflow_run.id }}
-          HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
-        run: |
-          # Check if the docs-builder step failed (even though the job succeeded via continue-on-error)
-          FAILED=$(gh api "repos/${REPO}/actions/runs/${RUN_ID}/jobs" \
-            --jq '[.jobs[].steps[] | select(.name == "Validate with docs-builder" and .conclusion == "failure")] | length')
-          if [ "$FAILED" = "0" ]; then
-            echo "No validation failure detected"
-            echo "notify=false" >> $GITHUB_OUTPUT
-            exit 0
-          fi
-          # Find the PR number from the head SHA
-          PR_NUMBER=$(gh api "repos/${REPO}/commits/${HEAD_SHA}/pulls" \
-            --jq '.[0].number' 2>/dev/null || true)
-          if [ -z "$PR_NUMBER" ] || [ "$PR_NUMBER" = "null" ]; then
-            echo "Could not find PR for SHA ${HEAD_SHA}"
-            echo "notify=false" >> $GITHUB_OUTPUT
-            exit 0
-          fi
-          echo "pr_number=$PR_NUMBER" >> $GITHUB_OUTPUT
-          echo "notify=true" >> $GITHUB_OUTPUT
+      - name: Download failure signal
+        id: artifact
+        uses: actions/download-artifact@v8
+        with:
+          name: docs-validation-failed
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ github.token }}
+          path: validate-result
+        continue-on-error: true
 
       - name: Comment on PR
-        if: steps.check.outputs.notify == 'true'
+        if: steps.artifact.outcome == 'success'
         env:
           GH_TOKEN: ${{ github.token }}
           RUN_URL: ${{ github.event.workflow_run.html_url }}
           REPO: ${{ github.repository }}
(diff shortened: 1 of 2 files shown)


.github/workflows/validate-package-docs.yml
```diff
@@ -9,6 +9,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: write
```
mrodm: @v1v is there any inconvenient on setting this permission in Github workflows?

I see there are other workflows setting that permission, but I'd like to double check it for this workflow.

This would be needed to post comment in Pull Requests (keep me honest here @theletterf )

theletterf: Worst thing that might happen from a fork is someone writing a different comment on their own PR.

v1v: We follow the least-permissive approach, so permissions are at the job level rather than at the top-level. That's my suggestion here.


OTOH, this won't work for forked PRs, this GitHub workflow runs for PRs in an unprivileged environment. Maybe it will be interesting to provide a workaround for those forked PRs with a job summary:
- https://github.blog/news-insights/product-news/supercharging-github-actions-with-job-summaries/

Or, we could support the pull-request/workflow_run approach:
- https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

theletterf: Let me check — might split it again.

theletterf: Reworked — dropped `pull-requests: write` from the validation workflow entirely. Now using the two-workflow split: the notification workflow downloads the validation run's logs via the API and greps for docs-builder's error/warning summary line. Permissions are at job level.

theletterf: @v1v Thanks for the feedback — reworked the approach:

- **Validation workflow** stays read-only (`contents: read` only), keeps `continue-on-error: true`, no `pull-requests: write`
- **Notification workflow** (`workflow_run` trigger) downloads the validation run's logs via the API and greps for docs-builder's summary line (`N Errors / N Warnings`). This works regardless of `continue-on-error` masking step conclusions
- Permissions are at job level per least-permissive approach
- Works for fork PRs since the `workflow_run` trigger runs in the base branch context

Could you take another look?

v1v: This won't add any GitHub comments with the outcome. If you are looking for that feature, you could use the upload-artefact/download-artifact steps between the two different workflows.

I recall I saw something like that in the past

theletterf: @v1v Reworked per your suggestions:

- **Validation workflow** stays read-only (`contents: read` only), no `pull-requests: write`. Uses `upload-artifact@v7` to upload a signal file with the PR number when docs-builder finds errors
- **Notification workflow** uses `download-artifact@v8` with `run-id` and `github-token` to download the artifact cross-workflow. Permissions at job level (`pull-requests: write`, `actions: read`)
- Artifact upload confirmed working in test PR #18049

Could you take another look?

theletterf: @v1v Updated again based on further review:

- **Validation workflow**: `contents: read` only, no write permissions. Uses `upload-artifact@v7` to upload a signal file when docs-builder finds errors. `PR_NUMBER` passed as env var (no inline `${{ }}` in shell)
- **Notification workflow**: `download-artifact@v8` with `run-id` + `github-token` for cross-workflow download. Permissions at job level (`pull-requests: write`, `actions: read`)
- **Fork safety**: The notification workflow verifies the PR number from the artifact matches the actual PR for the head SHA (`GET /repos/{repo}/commits/{sha}/pulls`), preventing a forked PR from targeting a different PR's comments
- **Fork support**: Works for fork PRs — artifacts are stored in the upstream repo, and `download-artifact@v8` can access them from the `workflow_run` context

Artifact upload confirmed in test PR #18049.

.github/workflows/notify-package-docs-failure.yml
```diff
@@ -5,51 +5,46 @@ on:
...
-  actions: read
-
 jobs:
   notify:
     name: Comment on PR
     if: github.event.workflow_run.event == 'pull_request'
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      actions: read
```
v1v, line 15: 🤟

.github/workflows/validate-package-docs.yml
```diff
@@ -78,6 +79,18 @@ jobs:
...
+          fi
+
+      - name: Upload failure signal
+        if: steps.changed-docs.outputs.skip != 'true'
+        uses: actions/upload-artifact@v7
+        with:
+          name: docs-validation-failed
+          path: validate-result/pr_number
+          if-no-files-found: ignore
+          retention-days: 1
```
v1v, line 96: 💯

PR #18103, opened 2026-03-27
ti_crowdstrike: fix data_stream.namespace field type in transform destination indices

chemamartinez:
## Proposed commit message

```
fix data_stream.namespace field type in transform destination indices

The data_stream.namespace field in the latest_intel and latest_ioc transform
destination indices was mapped as constant_keyword. This is incorrect because
these transforms consume data from all namespaces (logs-ti_crowdstrike.*-*),
meaning documents with different namespace values are merged into the same
destination index. Elasticsearch silently drops or ignores values that differ from the
constant, leading to incorrect field values in the destination index.

Per the Fleet Package Code Review guidelines[1]:

    If your transform consumes data from multiple data streams, avoid using constant_keyword
    in the transform's mappings if it results in documents with varying values being merged into
    the destination index. For instance, use a keyword type instead of constant_keyword for fields
    like data_stream.dataset.

Changed the type to keyword in both transform field definitions. As this is a mapping
change, the destination index versions and fleet_transform_version values have been
bumped to trigger recreation of the transform and its destination index on upgrade.

[1] https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#optimizing-indices-using-constant-keyword-fields
```

## Checklist

### DIFF
--- a/packages/ti_crowdstrike/changelog.yml
+++ b/packages/ti_crowdstrike/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "2.7.1"
   changes:
-    - description: Update data stream namespace field type to keyword in transforms.
+    - description: Change `data_stream.namespace` field type from constant_keyword to keyword in transform destination indices to support multiple namespaces.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/18103
 - version: "2.7.0"


packages/ti_crowdstrike/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.7.1"
+  changes:
+    - description: Update data stream namespace field type to keyword in transforms.
```
andrewkroh:
```suggestion
    - description: Change `data_stream.namespace` field type from constant_keyword to keyword in transform destination indices to support multiple namespaces.
```

chemamartinez:
> One related concern, if multiple namespaces are being used, is the `unique_key` in the transform set up to work correctly? Are the `event.id` values either unique across namespaces OR when there is an ID collision are they both talking about the same indicator?

Taking a look at the `event.id` field it refers to the indicator ID, it seems to me a strong ID (probably a hash?), but in practice I think it is very unlikely that two different indicators share the same ID from the same Crowdstrike account. Especially from the intel feed which is a global intel database.

PR #17835, opened 2026-03-16
Make cyberark_epm.raw_event.evidences a flattened field

chrisberkhout:
## Proposed commit message

```
Map `cyberark_epm.raw_event.evidences` as flattened

The documentation[1] is not specific about the structure of the objects.
A pipeline test has had a value added based on real-world data.

[1]: https://docs.cyberark.com/epm/latest/en/content/webservices/getdetailedrawevents.htm
```

## Checklist

- [ ] I have reviewed [tips for building integrations](https://www.elastic.co/docs/extend/integrations/tips-for-building) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices)

PR #18105, opened 2026-03-27
cyberark_epm: update raw_event.exposed_users field type from keyword to flattened

chemamartinez:
## Proposed commit message

```
Map `cyberark_epm.raw_event.exposed_users` as flattened.

The documentation[1] is not specific about the structure of the objects.
Based on the documentation[1], this field and evidences already fixed
should be the only ones that are expected to be arrays of objects.

A pipeline test has had a value added based on real-world data.

[1]: https://docs.cyberark.com/epm/latest/en/content/webservices/getdetailedrawevents.htm
```

## Checklist

## Related issues

- Relates https://github.com/elastic/integrations/pull/17835

### DIFF
--- a/packages/cyberark_epm/data_stream/raw_event/_dev/test/pipeline/test-raw-event.log-expected.json
+++ b/packages/cyberark_epm/data_stream/raw_event/_dev/test/pipeline/test-raw-event.log-expected.json
@@ -1189,7 +1189,7 @@
                     "943E0A8C840430E53D8D641CF4CDC1660C75FFE0"
                 ],
                 "user": [
-                    "{accountName=SOMENAME, domain=SOMEDOMAIN, username=USERNAME}",
+                    "USERNAME",
                     "caadmin"
                 ]
             },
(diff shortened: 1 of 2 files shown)


packages/cyberark_epm/data_stream/raw_event/_dev/test/pipeline/test-raw-event.log-expected.json
```diff
@@ -1182,6 +1189,7 @@
                     "943E0A8C840430E53D8D641CF4CDC1660C75FFE0"
                 ],
                 "user": [
+                    "{accountName=SOMENAME, domain=SOMEDOMAIN, username=USERNAME}",
```
efd6: I think we can do better than this; if the element of `cyberark_epm.raw_event.exposed_users` is not a string, I think we need to be more careful. I know that the commit message says that we don't have any documentation for the structure, but are there heuristics that we can reasonably lean against? Maybe all fields that do not match "domain" get appended into `user`? Alternatively, just omit values if they are not a string.

chemamartinez: Agreed, I missed that assignment. I've taken an intermediate approach, just adding the field that we are sure it is an user (username) to related.user.

PR #18135, opened 2026-03-30
microsoft_exchange_online_message_trace: fix pagination with broken skiptoken

### LINKED ISSUE #18134: microsoft_exchange_online_message_trace: endpoint returns broken skiptoken causing infinite 400 loop
The Microsoft Graph beta `/admin/exchange/tracing/messageTraces` endpoint intermittently returns an `@odata.nextLink` containing a `$skiptoken` with malformed date values. The base64-decoded skiptoken JSON has dates wrapped in extra literal quotes:

```json
{
  "StartDate": "\"2026-03-26T15:43:20.2718704Z\"",
  "EndDate": "\"2026-03-28T13:04:13.0580000Z\"",
  "StartingRecipientAddress": "\"user@example.com\"",
  "ResultSize": "1000"
}
```

When this token is sent back to Microsoft (as the OData spec requires; clients follow `@odata.nextLink` verbatim), their backend fails to deserialise the quoted dates and returns a 400 BadRequest: "Cannot process argument transformation on parameter 'StartDate'".

This is a Microsoft-side serialisation defect. The integration does not construct or modify the skiptoken.

#### Impact

The integration gets stuck in a permanent retry loop. The CEL error handler uses the single-object error form, which rolls the cursor back to the last successfully persisted state. That state contains the poisoned `next_link`, so every polling cycle retries the same broken URL and gets the same 400 error.

The only manual workaround is to stop the agent, delete the input's registry entry, and restart, losing data from the skipped time range.

#### Fix

Sanitise the skiptoken when storing it in the cursor. The fix decodes the base64 JSON, walks each value with `transformMapEntry`, and uses `parse_time` with a quoted RFC3339Nano layout as a semantic discriminator:

- If a value parses as a quoted timestamp, it was double-quoted → replace it with the unquoted form.
- If parsing fails, the value is either a clean timestamp or not a timestamp at all → leave it unchanged.

The fix is applied at cursor storage time (during the successful 200 response) so the persisted cursor always contains a clean URL. This matters because the error path rolls back the cursor by design; fixing the URL only at use time would still persist the broken version.

When the defect is absent, all values fail the quoted parse and pass through unchanged.

#### Notes

- The messageTraces Graph API is the new replacement for the deprecated Reporting Webservice (decommissioned April 6, 2026). There is no GA endpoint yet; the beta endpoint is the only option. See #17043.
- The Graph API responses include `Deprecation` and `Sunset` headers, but these apply to the `/beta` namespace in general, not to the messageTraces endpoint specifically.
- There is a secondary issue: when the CEL program emits an error event, the ingest pipeline's `json` processor fails because `event.original` does not exist on error events.

efd6:
## Proposed commit message
```
microsoft_exchange_online_message_trace: fix pagination with broken skiptoken

The Microsoft Graph beta messageTraces endpoint sometimes returns a
$skiptoken containing double-JSON-encoded string values (e.g.
"\"2026-03-26T...\"" instead of "2026-03-26T...") that its own
backend rejects with a 400 BadRequest on the next page request.
Because the CEL error handler uses the single-object error form,
the cursor rolls back to the last successful state — which contains
the poisoned next_link — creating an infinite retry loop.

Fix by sanitising the skiptoken when reading from the cursor. The
stored next_link URL is parsed with parse_url/parse_query, the
base64-encoded skiptoken is decoded to a JSON map, and each value
that matches ^".*"$ (indicating double-JSON-encoding) is decoded
one level with decode_json. Values that don't match the pattern or
fail to decode pass through unchanged. The URL is reassembled with
format_query (preserving other query parameters) and format_url.

Fixing on read rather than on write means existing poisoned cursors
are repaired automatically on upgrade — no manual registry deletion
required.

Base64 padding is stripped before decoding (trim_right +
base64_raw_decode) because base64_decode requires padding and the
API sometimes returns unpadded tokens. Output uses padded base64.

The system test exercises a three-page pagination chain. Page 1
returns a broken skiptoken (unpadded, double-encoded values); the
expected page 2 match token was computed by running the fix logic
through mito. Page 2 returns a clean skiptoken (padded); the page 3
match token was computed the same way — clean values pass through
unchanged but keys are reordered by encode_json.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #18134

### DIFF
--- a/packages/microsoft_exchange_online_message_trace/changelog.yml
+++ b/packages/microsoft_exchange_online_message_trace/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "2.0.3"
   changes:
-    - description: Work around Microsoft Graph beta API defect where double-quoted dates in the skiptoken cause pagination to fail with 400 errors.
+    - description: Work around Microsoft Graph beta API defect where double-quoted values encoded in the skiptoken cause pagination to fail with 400 errors.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/18135
 - version: "2.0.2"


packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
```diff
@@ -87,7 +87,36 @@ program: |-
...
-                  ?"next_link": body[?"@odata.nextLink"],
+                  // Work around Microsoft Graph beta API defect where
+                  // skiptoken date values are double-quoted, causing
+                  // 400 errors on the next page request.
+                  ?"next_link": body[?"@odata.nextLink"].optMap(link,
+                    link.parse_url().as(u,
+                      u.RawQuery.parse_query().as(q,
+                        q[?"$skiptoken"].orValue([""]).as(tokens,
+                          (tokens.size() > 0 && tokens[0] != "") ?
+                            try(tokens[0].trim_suffix("==").trim_suffix("=").base64_raw_decode().decode_json()).as(token,
```
chrisberkhout: `.trim_suffix("==").trim_suffix("=").base64_raw_decode()`
could be
`.base64_decode()`

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
```diff
@@ -87,7 +87,36 @@ program: |-
...
+                              !is_error(token) ?
+                                u.with_replace({
+                                  "RawQuery": "$skiptoken=" + bytes(
```
chrisberkhout: Would be nicer to modify `$skiptoken` without assuming it's the only query string parameter. It is, but they give us the whole URL.

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
```diff
@@ -87,7 +87,36 @@ program: |-
...
+                                    token.transformMapEntry(k, v,
+                                      {
+                                        k: v.parse_time("\"" + time_layout.RFC3339Nano + "\"").as(qt,
+                                          is_error(qt) ? v : qt.format(time_layout.RFC3339Nano)
+                                        )
+                                      }
+                                    ).encode_json()
```
chrisberkhout: Important: this needs to be more general. We're seeing values like:
```json
{
  "StartDate": "\"2026-03-26T15:43:20.2718704Z\"",
  "EndDate": "\"2026-03-28T13:04:13.0580000Z\"",
  "StartingRecipientAddress": "\"user@example.com\"",
  "ResultSize": "1000"
}
```
And the `StartingRecipientAddress` is also used for pagination, so that also needs to have the quotes stripped.

If these values are being serialized as JSON, which seems likely, then an email address string could have characters escaped within it, as well as the outer quotes. I don't think we necessarily need to handle those cases, but one way to do so would be to try to JSON decode the values:

```bash
mito -data <(echo '{
  "StartDate": "\"2026-03-26T15:43:20.2718704Z\"",
  "EndDate": "2026-03-28T13:04:13.0580000Z",
  "StartingRecipientAddress": "\"user@example.com\"",
  "ResultSize": "1000"
}') <(echo '
  state.transformMapEntry(k, v, { k: try(v.decode_json()).as(dv, string(dv).has_prefix("failed") ? v : dv) })
')
```

That example doesn't use `is_error(dv)` because that was coming out as false even when there was a JSON parsing error (like for `EndDate`).

---
For future reference: `try()` is not needed to check for an error:
```bash
mito <(echo '(0/0).as(v, is_error(v) ? "error" : "ok")')
```

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
```diff
@@ -87,7 +87,36 @@ program: |-
                   [],
                 "want_more": want_more,
                 "cursor": {
```
chrisberkhout, line 119: The idea of storing good values sounds reasonable, but there are some benefits to saving what we get from the API and fixing it on read are:
- bad old values in saved cursor data can be corrected
- if we need to change the logic we can do that with a package upgrade only

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
efd6: Indeed.

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
efd6: I'll see what we can do.

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
efd6: Agree.

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
efd6: Based on diagnostics, it appears (one instance only) that the base64 can be sent unpadded, so this is safer.

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
efd6: Done

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
chrisberkhout: Oh, with a length that's not already divisible by 3?
Really keeping its users on their toes.

packages/microsoft_exchange_online_message_trace/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.0.2"
+  changes:
+    - description: Work around Microsoft Graph beta API defect where double-quoted dates in the skiptoken cause pagination to fail with 400 errors.
```
chrisberkhout:
```suggestion
    - description: Work around Microsoft Graph beta API defect where double-quoted values encoded in the skiptoken cause pagination to fail with 400 errors.
```

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
```diff
@@ -38,8 +38,38 @@ program: |-
...
+                  is_error(token) ? state.cursor.next_link :
+                  u.with_replace({
+                    "RawQuery": q.with_replace({
+                      "$skiptoken": [bytes(
+                        token.transformMapEntry(k, v,
+                          {
+                            k: (v.has_prefix('"') && v.has_suffix('"')) ?
+                              v.decode_json().as(dv, is_error(dv) ? v : dv)
+                            :
+                              v
```
chrisberkhout, line 62: This is good. I was thinking all values could have the transform attempt, but the `ResultSize` value is a string and we don't want to turn it into a number like my earlier example did.

PR #18141, opened 2026-03-31
[Auth0] Add agentless deployment

### LINKED ISSUE #17534: [auth0] Add Agentless Deployment Support
# Description
Add agentless support to the auth0.

# Requirements
Following the established Phase I pattern, for each integration see the [Onboarding Integration Guide](https://github.com/elastic/security-team/blob/main/docs/cloud-security-team/agentless/onboard-integration.mdx).

**1. Technical implementation**
- [ ] Update integration manifest.yml to enable agentless deployment mode ()
- [ ] Update integration documentation with agentless deployment instructions
- [ ] Update changelog.md 

**Example reference: https://github.com/elastic/integrations/pull/13367**

**2. Performance documentation**
- [ ] Test and document throughput in agentless (requires access to vendor environment and/or sample data)
- [ ] Document specific metrics for each integration. For example/where possible:
  - **API response time:** Average time for vendor API calls to complete
  - **Events processed per minute:** How many log entries/events the integration can handle
  - **Error rates:** Percentage of failed API calls or data processing errors
  - **Container resource usage:** CPU and memory consumption under typical load
  - **Vendor-specific limits:** Rate limiting thresholds and API quotas

**Example documentation format:** "Crowdstrike Falcon Intelligence: 200ms avg API response, 5,000 events/min, 0.1% error rate, 512MB RAM/0.5CPU, 1000 API calls/hour limit"

# Dependencies
- Agentless infrastructure GA readiness (still in beta)
- Input compatibility: Currently agentless is optimized for httpjson and cel inputs
- UX enhancement: Add agentless deployment filter/toggle to integrations catalog page for better discoverability

moxarth-rathod:
## Proposed commit message

```
auth0: add agentless support and update kibana version constraint

Adds support for agentless and updates the Kibana version constraint to avoid a known issue 
where selecting agentless deployment mode hides configuration fields and blocks 
deployment if inputs default to `enabled: false`.

Ref: https://github.com/elastic/kibana/pull/265106
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/auth0 directory.

> elastic-package test -v

## Related issues

- Closes #17534

### DIFF
--- a/packages/auth0/changelog.yml
+++ b/packages/auth0/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.24.0"
+  changes:
+    - description: Enable Agentless deployment.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
 - version: "1.23.1"
   changes:
     - description: Remove duplicate security-solution-default tag references
--- a/packages/auth0/changelog.yml
+++ b/packages/auth0/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Enable Agentless deployment.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/1
+      link: https://github.com/elastic/integrations/pull/18141
 - version: "1.23.1"
   changes:
     - description: Remove duplicate security-solution-default tag references
--- a/packages/auth0/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/auth0/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
@@ -7,6 +7,17 @@ processors:
 - set:
     field: auth0.logs.data
     copy_from: json.data
+- remove:
+    field:
+      - organization
+      - division
+      - team
+    ignore_missing: true
+    if: ctx.organization instanceof String && ctx.division instanceof String && ctx.team instanceof String
+    tag: remove_agentless_tags
+    description: >-
+      Removes the fields added by Agentless as metadata,
+      as they can collide with ECS fields.
 - date:
     field: auth0.logs.data.date
     formats:
--- a/packages/auth0/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/auth0/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
@@ -7,17 +7,6 @@ processors:
 - set:
     field: auth0.logs.data
     copy_from: json.data
-- remove:
-    field:
-      - organization
-      - division
-      - team
-    ignore_missing: true
-    if: ctx.organization instanceof String && ctx.division instanceof String && ctx.team instanceof String
-    tag: remove_agentless_tags
-    description: >-
-      Removes the fields added by Agentless as metadata,
-      as they can collide with ECS fields.
 - date:
     field: auth0.logs.data.date
     formats:
--- a/packages/auth0/data_stream/logs/manifest.yml
+++ b/packages/auth0/data_stream/logs/manifest.yml
@@ -4,7 +4,6 @@ streams:
   - input: http_endpoint
     title: Auth0 log events via Webhooks
     description: Receives log events from Auth0 via Webhooks
-    enabled: false
     template_path: http_endpoint.yml.hbs
     vars:
       - name: listen_address
@@ -78,7 +77,6 @@ streams:
   - input: cel
     title: Auth0 log events via API requests
     description: Collects log events from Auth0 via API requests.
-    enabled: false
     template_path: cel.yml.hbs
     vars:
       - name: url
--- a/packages/auth0/data_stream/logs/manifest.yml
+++ b/packages/auth0/data_stream/logs/manifest.yml
@@ -4,6 +4,7 @@ streams:
   - input: http_endpoint
     title: Auth0 log events via Webhooks
     description: Receives log events from Auth0 via Webhooks
+    enabled: false
     template_path: http_endpoint.yml.hbs
     vars:
       - name: listen_address
--- a/packages/auth0/data_stream/logs/manifest.yml
+++ b/packages/auth0/data_stream/logs/manifest.yml
@@ -79,6 +79,7 @@ streams:
     title: Auth0 log events via API requests
     description: Collects log events from Auth0 via API requests.
     template_path: cel.yml.hbs
+    enabled: false
     vars:
       - name: url
         type: text
--- a/packages/auth0/data_stream/logs/manifest.yml
+++ b/packages/auth0/data_stream/logs/manifest.yml
@@ -78,8 +78,8 @@ streams:
   - input: cel
     title: Auth0 log events via API requests
     description: Collects log events from Auth0 via API requests.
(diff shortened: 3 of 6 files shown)


packages/auth0/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -7,6 +7,17 @@ processors:
 - set:
     field: auth0.logs.data
     copy_from: json.data
+- remove:
+    field:
+      - organization
+      - division
+      - team
```
andrewkroh: This should not be needed if the minimum kibana matches where https://github.com/elastic/kibana/pull/230479 was applied.

packages/auth0/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.24.0"
+  changes:
+    - description: Enable Agentless deployment.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
```
andrewkroh:
```suggestion
      link: https://github.com/elastic/integrations/pull/18141
```

packages/auth0/data_stream/logs/manifest.yml
```diff
@@ -4,7 +4,6 @@ streams:
   - input: http_endpoint
     title: Auth0 log events via Webhooks
     description: Receives log events from Auth0 via Webhooks
-    enabled: false
```
andrewkroh, line 7: Why this change?

packages/auth0/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
moxarth-rathod: We need the minimum kibana version - `8.19.2` to fix https://github.com/elastic/kibana/pull/230479.

andrewkroh: Why was the minimum version selected as `^8.18.0`? I think we can safely move to at least 8.19.2, and drop this `remove` processor.

The 8.18 Elastic stack is unsupported / unmaintained. And it has been since 9.2 was released (Oct 23 2025)[^1].

[^1]: https://www.elastic.co/support/eol

moxarth-rathod: Got it, Thanks for the suggestions!

packages/auth0/data_stream/logs/manifest.yml
moxarth-rathod: This prevents a UI bug in serverless where having all available options disabled causes an issue. I'll create a separate issue for it and attach it here.

moxarth-rathod: Issue link: https://github.com/elastic/beats/issues/49942

andrewkroh: So this is a workaround for a UI bug? There needs to be a UI bug issue (elastic/kibana) associated as well.

I think the reason both inputs are disabled is to make the user choose their ingestion method. Changing the `enabled` default value does affect new users experience the onboarding flow, both in the UI and especially via the package_policy API.

Let's say you are an API user who was only configuring the CEL input in their requests. Then the http_endpoint becomes enabled by default in the integration package. The next time you try to reproduce an API request for auth0 you will get new behavior. They might not notice it in this case because there are no mandatory variables for the http_endpoint stream, but their agents will now have an HTTP server listening which is very surprising.

Or let's say, they were using the http_endpoint only. Now the CEL input becomes enabled by default. This is different from the earlier case because the CEL input does have two mandatory variables, client_id and client_secret, so now the package_policy API request that they used to make fails because they did not set the mandatory variables. This will be confusing because their request didn't ask for the CEL input at all.

All of this is to say that changing `enabled` can impact users.

If we must have a workaround and the requirement is to have an input enabled, then it should only enable the CEL input. The http_endpoint input is not supported in agentless. And ideally we should return the integration back to its original state (enabled=false) after Kibana addresses the UX issue.

efd6: There is #18157 which looks related (though auth0 is not listed in that issue) and which links to elastic/kibana#260500.

packages/auth0/data_stream/logs/manifest.yml
```diff
@@ -78,7 +78,6 @@ streams:
   - input: cel
     title: Auth0 log events via API requests
     description: Collects log events from Auth0 via API requests.
-    enabled: false
```
andrewkroh, line 81: Kibana has addressed https://github.com/elastic/kibana/issues/261788. Let's revert this change.

moxarth-rathod: Sure, let me check

mohitjha-elastic: @andrewkroh 
The [PR](https://github.com/elastic/kibana/pull/265106) has been merged with the `9.5.0` label, and the backport has been skipped. As per the current schedule, this `9.5.0` is planned for the July release. Shall we proceed with the workaround until then, as reverting the changes may cause the issue in the current serverless setup?
Let me know your thoughts on this.

andrewkroh: Is this a problem only affecting agentless deployments on Serverless? If so, the PR that was merged to main will be deployed to Serverless soon. Serverless uses artifacts produced from `main` not traditional tagged stack releases.

But if this is a problem affecting agentless on ECH too, then the fix should be backported and we should wait for that to be available on ECH.

kcreddy: > But if this is a problem affecting agentless on ECH too, then the fix should be backported and we should wait for that to be available on ECH.

The changes are now backported: 
- https://github.com/elastic/kibana/pull/266203
- https://github.com/elastic/kibana/pull/266201
- https://github.com/elastic/kibana/pull/266103

Once they are available on ECH, we should update the stack version constraints accordingly.

packages/auth0/manifest.yml
```diff
@@ -1,15 +1,15 @@
...
+version: "1.24.0"
 description: Collect logs from Auth0 with Elastic Agent.
 type: integration
 categories:
   - security
   - iam
 conditions:
   kibana:
-    version: "^8.13.0 || ^9.0.0"
+    version: "^8.19.16 || ~9.3.5 || ^9.4.1"
```
efd6, line 12: Wait until these exist?

moxarth-rathod: yes, moving to draft

packages/auth0/data_stream/logs/manifest.yml
```diff
@@ -78,8 +78,8 @@ streams:
...
     template_path: cel.yml.hbs
+    enabled: false
```
efd6, line 81: Move this to the original location.

packages/auth0/manifest.yml
efd6: The commit message should explain why this is necessary.

efd6 (approved):

Issue #17561, opened 2026-02-25
Conflicting `event.kind` values in constant_keyword fields


andrewkroh:
The following data streams define `event.kind` as a `constant_keyword` field but attempt to set it to conflicting values within their ingest pipelines. This configuration will cause indexing errors because `constant_keyword` fields must have a single, immutable value for all documents in the index.

### Summary of Issues

*   **Fixed Value Conflict:** The field definition explicitly sets a constant value (e.g., `alert`), but the ingest pipeline sets it to a different value (e.g., `pipeline_error`) during error handling or other conditional logic.
*   **Multiple Value Conflict:** The field definition is `constant_keyword` but does not specify a value (meaning the first indexed document sets the constant for the index). However, the pipeline sets the field to multiple different values (e.g., `state` normally, but `pipeline_error` on failure), ensuring that one of those cases will fail to index.

### Affected Data Streams

| Package | Data Stream | Field Definition | Conflicting Pipeline Processor |
| :--- | :--- | :--- | :--- |
| **aws** | `securityhub_findings_full_posture` | [fields/ecs.yml:4](https://github.com/elastic/integrations/blob/29c5ee80540d5423a6b393912dd5edad73c9b637/packages/aws/data_stream/securityhub_findings_full_posture/fields/ecs.yml#L4) (value: unset) | [default.yml:18](https://github.com/elastic/integrations/blob/29c5ee80540d5423a6b393912dd5edad73c9b637/packages/aws/data_stream/securityhub_findings_full_posture/elasticsearch/ingest_pipeline/default.yml#L18) (sets `state`) <br> [default.yml:2668](https://github.com/elastic/integrations/blob/29c5ee80540d5423a6b393912dd5edad73c9b637/packages/aws/data_stream/securityhub_findings_full_posture/elasticsearch/ingest_pipeline/default.yml#L2668) (sets `pipeline_error`) |
| **blacklens** | `alerts` | [fields/ecs.yml:2](https://github.com/elastic/integrations/blob/29c5ee80540d5423a6b393912dd5edad73c9b637/packages/blacklens/data_stream/alerts/fields/ecs.yml#L2) (value: `alert`) | [default.yml:99](https://github.com/elastic/integrations/blob/29c5ee80540d5423a6b393912dd5edad73c9b637/packages/blacklens/data_stream/alerts/elasticsearch/ingest_pipeline/default.yml#L99) (sets `pipeline_error`) |
[...]

PR #18143, opened 2026-03-31
[aws, blacklens, github, microsoft_sentinel] Resolve conflicting event.kind values in constant_keyword fields

mohitjha-elastic:
## Proposed commit message

```
aws, blacklens, github, microsoft_sentinel: fix conflicting event.kind values.

Fix conflicting event.kind mappings where pipelines set pipeline_error (or state) while the field 
was defined as constant_keyword, which Elasticsearch treats as a single immutable value per index. 
The securityhub_findings_full_posture data stream now references ECS for event.kind, matching the
sibling Security Hub stream. 
For blacklens alerts, GitHub issues/code_scanning/dependabot/secret_scanning, 
and Microsoft Sentinel alert/event/incident streams, event.kind is mapped as keyword and the 
ingest pipelines set the normal ECS value (alert or event) so successful documents stay unchanged
while failure handlers can still set pipeline_error. 
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/<integration_name> directory.

Replace the <integration_name> with the actual name of integration. 

> elastic-package test -v

## Related issues

- Closes https://github.com/elastic/integrations/issues/17561


packages/blacklens/data_stream/alerts/fields/ecs.yml
```diff
@@ -1,7 +1,6 @@
 # Define ECS constant fields as constant_keyword
 - name: event.kind
-  type: constant_keyword
-  value: alert
+  type: keyword
```
andrewkroh: `event.kind` is an ECS field and should be defined using `external: ecs` rather than `type: keyword`.

https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#defining-an-ecs-field-without-using-an-external-definition

----

fwiw You can detect and fix (add `-fix`) this class of issue automatically using

`go run github.com/andrewkroh/fydler@main -a useecs packages/blacklens/data_stream/alerts/fields/ecs.yml`

mohitjha-elastic: Updated all occurrences of `event.kind` to be defined using `external: ecs`.

PR #18155, opened 2026-03-31
[m365_defender] Introduce Include Alerts configuration option

### LINKED ISSUE #18144: m365_defender.incident: Make `$expand=alerts` configurable
### Summary
The M365 Defender incident data stream hardcodes `$expand=alerts` in its Microsoft Graph Security API query. On large tenants, this combination of `$expand=alerts` and `$filter` triggers persistent HTTP 500 InternalServerError responses from Microsoft when incidents have many linked alerts. The HTTP JSON input exhausts its retry budget, enters a Degraded state, and the cursor never advances — creating a permanent ingestion stall.

### Observed behavior
`GET /v1.0/security/incidents?$expand=alerts&$filter=lastUpdateDateTime ge <cursor>&$top=25` returns HTTP 500 consistently on a large tenant (~45k users).
The HTTP JSON input retries 6 times, then gives up. Because no events are emitted on failure, the cursor is never advanced.
Every subsequent poll sends the identical failing request. The input is permanently stuck until manually recreated.
`m365_defender.incident` events stop while `m365_defender.alert` continues normally.
Smaller tenants (~2k users) are unaffected.

### Root cause
The ingest template sets `$expand=alerts` unconditionally:

```yaml
- set:
    target: url.params.$expand
    value: 'alerts'
```

Removing `$expand=alerts` from the query causes incidents to return normally. The issue appears specific to the combination of `$filter` and `$expand=alerts` when one or more incidents in the result page has a large number of linked alerts.

### Proposed fix
Make `$expand=alerts` a configurable boolean option on the incident data stream, defaulting to true for backward compatibility.

Users who experience the 500 stall can disable the option. Because `alerts_v2` collects alerts independently, disabling `$expand` on incidents does not result in data loss.

brijesh-elastic:
## Proposed commit message

```
m365_defender: Introduce Include Alerts configuration option

This PR adds `Include Alerts` configuration option to to prevent ingestion stalls on large tenants. This
allows users to disable alert expansion in the Incident data stream, bypassing Microsoft Graph API
500 errors caused by high alert volume from single incident.
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/m365_defender directory.

> elastic-package test -v

## Related issues

- Closes #18144

### DIFF
--- a/packages/m365_defender/data_stream/incident/agent/stream/httpjson.yml.hbs
+++ b/packages/m365_defender/data_stream/incident/agent/stream/httpjson.yml.hbs
@@ -54,9 +54,11 @@ response.pagination:
 response.split:
   target: body.value
   ignore_empty_value: true
+{{#if include_alerts}}
   split:
     target: body.alerts
     keep_parent: true
+{{/if}}
 cursor:
   last_update_time:
     value: >-
--- a/packages/m365_defender/data_stream/incident/manifest.yml
+++ b/packages/m365_defender/data_stream/incident/manifest.yml
@@ -68,7 +68,7 @@ streams:
         required: true
         show_user: true
         title: Include Alerts
-        description: Include associated alerts with the incident. When enabled, the alerts are split so that each generated event represents a single alert from the incident.
+        description: Include associated alerts with the incident. When enabled, the alerts are split so that each generated event represents a single alert from the incident. If disabled, enable the Alerts (alerts_v2) data stream to ensure alert data is still collected.
         type: bool
         multi: false
         default: true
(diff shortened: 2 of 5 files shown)


packages/m365_defender/data_stream/incident/agent/stream/httpjson.yml.hbs
kcreddy, line 60: Should we also add a condition `{{#if include_alerts}}` here, to make it cleaner and explicit?

packages/m365_defender/data_stream/incident/manifest.yml
```diff
@@ -64,6 +64,14 @@ streams:
         type: bool
         multi: false
         default: false
+      - name: include_alerts
+        required: true
+        show_user: true
+        title: Include Alerts
+        description: Include associated alerts with the incident. When enabled, the alerts are split so that each generated event represents a single alert from the incident.
```
kcreddy: Also add `If disabled, enable the Alerts (alerts_v2) data stream to ensure alert data is still collected.`

packages/m365_defender/data_stream/incident/agent/stream/httpjson.yml.hbs
brijesh-elastic: Sure

PR #18156, opened 2026-03-31
[fortinet_fortigate] Improve escaped quote handling in painless script

taylor-swanson:
## Proposed commit message

- Fixed painless script to properly handle escaped quotes in key/value pairs
- Added pipeline test case

## Checklist

```
cd packages/fortinet_fortigate
elastic-package test
```


packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -50,7 +50,9 @@ processors:
           char quote = (char)"\"";
           for (def currentPosition = 0; currentPosition < input.length(); currentPosition++) {
               if (input.charAt(currentPosition) == quote) {
+                if (currentPosition == 0 || input.charAt(currentPosition - 1) != (char)'\\') {
```
mjwolf: I don't think this will handle all cases, for example `\\\\"`, where there's two escaped `\` and then an unescaped `"`, isInQuotes shouldn't flip, but it will in this code that only looks at the one previous char.

taylor-swanson: Good catch. Added proper escape handling (same method used in cisco_ise). I added another test case for this:

```
msg="web_misc: HTTP.X-Forwarded-For.Header.XSS\\"
```

taylor-swanson:
@mjwolf, could you take another look with the updated changes?

PR #18158, opened 2026-03-31
Map crowdstrike.alert.detection_context as flattened

chrisberkhout:
## Proposed commit message

```
Map crowdstrike.alert.detection_context as flattened

The flattened type was chosen because a complete list of subfield names
is not known.

Also restructures the alert pipeline to deserialize JSON straight into
`crowdstrike.alert.*` rather than putting it elsewhere and renaming
individual fields to their final destinations. Since there are a number
of known fields that were intentionally not mapped into the vendor
namespace, those are explicitly removed.

This may lead to additional dynamically mapped fields if there are more
unknown fields in the API's alert payloads.
```

## Note to reviewers

Can be read commit-by-commit.

If the last point in the proposed commit message is a problem, we can change to a Painless processor that retains only whitelisted fields under `crowdstrike.alert.*`.

## Checklist


packages/crowdstrike/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.15.0"
+  changes:
+    - description: Map `crowdstrike.alert.detection_context` as flattened.
```
kcreddy: I think we should also callout this in the changelog: new/unknown fields from the CrowdStrike API will now surface in the index (dynamically mapping) rather than being silently dropped. Because this behavior change may lead to mapping explosion. WDYT?

kcreddy (approved):
LGTM, just a note on changelog (which is also not a blocker).

packages/crowdstrike/changelog.yml
chrisberkhout: Agreed. Done during rebase, as:

> Map `crowdstrike.alert.detection_context` as flattened. Keep new/unknown alert fields under the `crowdstrike.alert` prefix.

PR #18185, opened 2026-04-01
Fixing integration url.full missing from fortigate

### LINKED ISSUE #14463: Fortigate Integration url.full missing
While reviewing the rule **Threat Indicator URL Match** I realized that the `url.full` field (which is the field against which the indicator is compared) is missing for fortigate logs.
I think it's worth adding this field in the main integration in order be able to use this rule
In the meanwhile I've add to my custom pipeline this
```
      {
        "set": {
          "field": "url.full",
          "value": "{{url.scheme}}://{{url.domain}}{{url.path}}",
          "if": "ctx?.url?.scheme != null && ctx?.url?.domain != null && ctx?.url?.path != null && ctx?.url?.full == null && ctx?.url?.query == null"
        }
      },
            {
        "set": {
          "field": "url.full",
          "value": "{{url.scheme}}://{{url.domain}}{{url.path}}?{{url.query}}",
          "if": "ctx?.url?.scheme != null && ctx?.url?.domain != null && ctx?.url?.path != null && ctx?.url?.full == null && ctx?.url?.query != null"
        }
      }
```

Thank you

qcorporation:
## Proposed commit message
Populate url.full when the individual URL components (url.scheme, url.domain, url.path, url.query) are present but url.full is missing. This enables detection rules like Threat Indicator URL Match that rely on url.full for indicator comparison.

## WHAT

- Added 4 processors to utm.yml: infer url.scheme from network.protocol, copy url.full from fortinet.firewall.url when it's already a full URL, and assemble url.full from parts (with and without query string).
- Added the same 2 assembly processors to default.yml as a fallback for non-UTM log types.
- Declared url.full in ecs.yml so static field validation passes.
- Added 2 pipeline test lines to test-fortinet-7-4.log exercising the with-query and without-query assembly paths, using documented logid=1059028704 (LOGID_APP_CTRL_IPS_PASS) and real field values from the FortiOS 7.4 Log Message Reference.

## WHY: 
Fortigate logs parse url.domain, url.path, and url.scheme individually but never assembled url.full. The Elastic detection rule "Threat Indicator URL Match" compares indicators against url.full, so FortiGate events were invisible to that rule. The approach follows the same pattern the issue reporter validated in their custom pipeline.

## Checklist

## Related issues
- Closes https://github.com/elastic/integrations/issues/14463


packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -353,6 +353,16 @@ processors:
       tag: pipeline_461401ce
       name: '{{ IngestPipeline "utm" }}'
       if: ctx.fortinet?.firewall?.type == 'utm' || ctx.fortinet?.firewall?.type == 'dns'
+  - set:
+      tag: set_url_full_without_query_3f3ba2f9
+      field: url.full
+      value: "{{{url.scheme}}}://{{{url.domain}}}{{{url.path}}}"
+      if: ctx?.url?.scheme != null && ctx?.url?.domain != null && ctx?.url?.path != null && ctx?.url?.full == null && ctx?.url?.query == null
```
taylor-swanson:
```suggestion
      if: ctx.url?.scheme != null && ctx.url.domain != null && ctx.url.path != null && ctx.url.full == null && ctx.url.query == null
```

packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -353,6 +353,16 @@ processors:
...
+  - set:
+      tag: set_url_full_with_query_0d8b7ec0
+      field: url.full
+      value: "{{{url.scheme}}}://{{{url.domain}}}{{{url.path}}}?{{{url.query}}}"
+      if: ctx?.url?.scheme != null && ctx?.url?.domain != null && ctx?.url?.path != null && ctx?.url?.full == null && ctx?.url?.query != null
```
taylor-swanson:
```suggestion
      if: ctx.url?.scheme != null && ctx.url.domain != null && ctx.url.path != null && ctx.url.full == null && ctx.url.query != null
```

packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/utm.yml
```diff
@@ -360,6 +360,28 @@ processors:
...
+      tag: set_url_full_from_original_8ef44009
+      field: url.full
+      copy_from: fortinet.firewall.url
+      ignore_empty_value: true
+      if: ctx.url?.full == null && ctx.fortinet?.firewall?.url instanceof String && (ctx.fortinet.firewall.url.startsWith('http://') || ctx.fortinet.firewall.url.startsWith('https://'))
+  - set:
+      tag: set_url_full_with_query_ff61f80b
+      field: url.full
+      value: "{{{url.scheme}}}://{{{url.domain}}}{{{url.path}}}?{{{url.query}}}"
+      if: ctx.url?.scheme != null && ctx.url?.domain != null && ctx.url?.path != null && ctx.url?.full == null && ctx.url?.query != null
```
taylor-swanson:
```suggestion
      if: ctx.url?.scheme != null && ctx.url.domain != null && ctx.url.path != null && ctx.url.full == null && ctx.url.query != null
```

packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/utm.yml
```diff
@@ -360,6 +360,28 @@ processors:
...
+  - set:
+      tag: set_url_full_without_query_0a3eb6ad
+      field: url.full
+      value: "{{{url.scheme}}}://{{{url.domain}}}{{{url.path}}}"
+      if: ctx.url?.scheme != null && ctx.url?.domain != null && ctx.url?.path != null && ctx.url?.full == null && ctx.url?.query == null
```
taylor-swanson:
```suggestion
      if: ctx.url?.scheme != null && ctx.url.domain != null && ctx.url.path != null && ctx.url.full == null && ctx.url.query == null
```

packages/fortinet_fortigate/data_stream/log/_dev/test/pipeline/test-fortinet-6-2.log-expected.json
```diff
@@ -3756,7 +3769,10 @@
             },
             "tags": [
                 "preserve_original_event"
-            ]
+            ],
+            "url": {
+                "scheme": "https"
+            }
```
taylor-swanson: `url.scheme` with nothing else seems odd. The event has to do with certificates, not a URL, so I'm not sure we should have this? `network.protocol` is set to `https`, which I think is good enough.

mjwolf: I've fixed this. Now url.scheme will only be set when there's other url domain or firewall url information available, so this won't be set for certificates anymore.

taylor-swanson (approved):

PR #18187, opened 2026-04-01
[CI] Add validation to prevent duplicate package names

### LINKED ISSUE #17206: [CI] Add validation to prevent duplicate package names
Currently, it is possible to define multiple packages with the same package name. This can lead to ambiguity and unexpected behavior during builds, dependency resolution, or runtime usage.

There is no validation in place to detect or prevent this situation early.

Introduce a validation/check that ensures all package names are unique.

The validation should:
- Scan all defined packages
- Detect duplicate package names
- Fail with a clear and actionable error message when duplicates are found

This validation should be running also as part of the command `mage check`.

mrodm:
## Proposed commit message

Add a new `dev/packagenames` Go package that validates no two packages
share the same `name` field across the `packages/` directory.

WHAT: Introduces a recursive scanner that reads `manifest.yml` files,
filtering only valid package manifests (those with `format_version`,
`name`, `type`, and `version` fields, where `type` is one of
`integration`, `input`, or `content`). All duplicate names are
reported at once with their directory paths. The check runs as part of
`mage check`. The `dev/packagenames` folder is also added to the
backport script so it is available on backport branches.

WHY: It is currently possible to define multiple packages with the same
`name` in their manifest, leading to ambiguity and unexpected behavior
during builds and in the package registry. This validation catches the
problem early in CI.

## Author's Checklist

- [x] Revert the temporary test filter in `.buildkite/scripts/common.sh` before merging.
- [x] Revert the temporary debug messages in `dev/packagenames/packagenames.go` before merging.
- [x] Remove test packages.
- [x] Ensure `backport_branch.sh` includes the required changes to copy the new files.
- [x] Run backport pipeline in dry run mode and ensure that the new files are copied.
    - https://buildkite.com/elastic/integrations-backport/builds/232 
- [x] Validate that this new check works with elastic-package < v0.118.0
    - Tested locally:
      ```shell
      go get github.com/elastic/elastic-package@v0.109.1
      mage -v check
      # Output:
      # Running dependency: github.com/elastic/integrations/dev/packagenames.Check
      # ELASTIC_PACKAGE_BIN environment variable found, checking for elastic-package binary in CI path
      # elastic-package binary found in ELASTIC_PACKAGE_BIN path: /home/mariorodriguez/Coding/work/integrations/build/elastic-package
      # elastic-package 0.109.1 < 0.118.0, using recursive walk to find packages
      # Error: error validating package names in directory 'packages': found duplicate package names:
      # duplicate package name "elastic_package_registry" found in: packages/elastic_package_registry, packages/other/elastic_package_registry2
      ```
- [x] Validate that this script works without any elastic-package in the PATH:
    - https://buildkite.com/elastic/integrations/builds/40724
- [x] Validate that `elastic-package` under `${WORKSPACE}/build` is used
    - https://buildkite.com/elastic/integrations/builds/40730
- [x] Validate that it can detect duplicated names in nested folders:
    - https://buildkite.com/elastic/integrations/builds/40734#019d4a1d-9df6-4823-97c6-cbf76befae64/L688
    - `packages/elastic_package_registry` and `packages/other/elastic_package_registry2` 

```bash
# Running mage check in the current working copy should report no errors
mage -v check

# Simulate error:
# To simulate a duplicate, create two packages with the same name field
# in their manifest.yml and run mage check — it should fail with a
# clear error listing both directories.

cp -r packages/elastic_package_registry packages/elastic_package_registry2

# If this new test package is included, then codeowners check should be disabled in mage check
cp -r packages/elastic_package_registry packages/other/elastic_package_registry2

# update .github/CODEOWNERS to add an entry for elastic_package_registry2. Example:
# /packages/elastic_package_registry @elastic/ecosystem
# /packages/elastic_package_registry2 @elastic/ecosystem
# /packages/other/elastic_package_registry2 @elastic/ecosystem

mage -v check
```

## Related issues

- Closes #17206

---

_Generated with the assistance of Claude_

### DIFF
--- a/.buildkite/scripts/check_sources.sh
+++ b/.buildkite/scripts/check_sources.sh
@@ -6,7 +6,6 @@ set -euo pipefail
 
 add_bin_path
 with_mage
-use_elastic_package
 
 echo "--- Run mage check"
 mage -v check
@@ -29,6 +28,7 @@ fi
 
 if [[ "$run_links_command" == "true" ]] ; then
     # links command require at least v0.113.0
+    use_elastic_package
     ${ELASTIC_PACKAGE_BIN} links check
 else
     echo "Skip elastic-package links. Unsupported for this elastic-package version."


andrewkroh:
Did you consider requiring package dir names to match the manifest `name`? That would naturally force unique names within the repo and help improve consistency and prevent surprises.

mrodm:
> Did you consider requiring package dir names to match the manifest `name`? That would naturally force unique names within the repo and help improve consistency and prevent surprises.


Requiring package directory names to match the manifest `name` could definitely help with consistency for our current scenario of folders/packages layout. That said, it would likely require updating some existing packages (folders) in the repository, such as `sql_input` and possible others.

Also, with the plan to support an additional directory level for grouping packages (relates to https://github.com/elastic/integrations/issues/17403), I’m not sure this alone would fully address duplicate package names, since the same manifest name could still appear in different subdirectories.

So I think we still need a solution like this one to ensure that there are no duplicated package names. And moreover, this solution would not require to update any package (folder) present in the repository now.

.buildkite/scripts/common.sh
```diff
@@ -829,7 +829,7 @@ teardown_test_package() {
 }
 
 list_all_directories() {
-    find . -maxdepth 1 -mindepth 1 -type d | xargs -I {} basename {} | sort
+    find . -maxdepth 1 -mindepth 1 -type d | xargs -I {} basename {} | sort | grep -E '^elastic_package_registry$'
 }
```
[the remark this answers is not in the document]
mrodm: This is just for testing purposes. It needs to be removed before merging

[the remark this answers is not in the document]
mrodm: I'll handle it myself when I finish the tests

dev/packagenames/packagenames.go
```diff
@@ -0,0 +1,153 @@
...
+
+func checkPackageNames(dir string) error {
+	paths, err := findPackagePaths(dir)
+	if err != nil {
+		return fmt.Errorf("error finding packages: %w", err)
+	}
+	return checkDuplicateNames(paths)
+}
+
+func findPackagePaths(dir string) ([]string, error) {
```
mrodm: Currently, this method relies on `elastic-package find` command and if it is not supported (elastic-package <0.118.0) it finds all package manifests recursively using `filepath.WalkDir`.

As it has to be supported (because of backport branches) finding all packages via a walk path , WDYT about removing the support of `elastic-package find` here? @elastic/ecosystem

teresaromero: From what i understand, `elastic-package find` is still in technical preview. I understand it comes handy to list all packages under a given folder, but as a technical preview it could lead to unexpected errors if the command is changed... and we use it as dependency on this mage command.
With this said, i think we could remove the dependency and the code around validating versions etc. We can keep the walk, or use Glob with a fixed deep pattern (`packages/*/manifest.yml`) and extend it to any deep we might have in the future, to obtain the same list of directories. This way we have a single way of collecting the list and verifying the names. If we ever need to filter or use the features of the `find` command, other than listing all the packages, then i think it would be worth on adding the command...

mrodm: Agree with that, I'll remove the usage of `elastic-package` for that purpose for now, so the process looks for packages (based on the `manifest.yml` files) following the `WalkDir` process.

.buildkite/scripts/backport_branch.sh
```diff
@@ -147,23 +148,29 @@ updateBackportBranchContents() {
...
+    git add "${COVERAGE_SCRIPTS_FOLDER}"
 
     echo "Copying $CODEOWNERS_SCRIPTS_FOLDER from $SOURCE_BRANCH..."
     git checkout "$SOURCE_BRANCH" -- "${CODEOWNERS_SCRIPTS_FOLDER}"
-    git add ${CODEOWNERS_SCRIPTS_FOLDER}
+    git add "${CODEOWNERS_SCRIPTS_FOLDER}"
+
+    echo "Copying $PACKAGENAMES_SCRIPTS_FOLDER from $SOURCE_BRANCH..."
+    git checkout "$SOURCE_BRANCH" -- "${PACKAGENAMES_SCRIPTS_FOLDER}"
+    git add "${PACKAGENAMES_SCRIPTS_FOLDER}"
```
mrodm, line 173: These are the main changes in this file.
The new folder `dev/packagenames` needs to be copied into the backport branch too.

Issue #18186, opened 2026-04-01
Add documentation on agent version conditions in packages


juliaElastic:
Relates https://github.com/elastic/package-spec/issues/165

Document how integrations can define:
- package level agent version condition
- input template level agent version condition

Explain how the agent version conditions are being handled by Fleet.

PR #18203, opened 2026-04-02
docs: Agent version conditions

juliaElastic:
## Summary

Adds documentation explaining how integration packages can define agent version conditions

## Related issues

- Closes https://github.com/elastic/integrations/issues/18186
- Relates https://github.com/elastic/package-spec/issues/165

### DIFF
--- a/docs/extend/agent-version-conditions.md
+++ b/docs/extend/agent-version-conditions.md
@@ -32,7 +32,7 @@ conditions:
 Use this approach when **all** inputs in the package rely on agent capabilities introduced in a specific version.
 
 ::::{note}
-`conditions.agent.version` requires `format_version: 3.6.0` or later in the package manifest and Kibana 9.4 or later.
+`conditions.agent.version` requires Kibana 9.4 or later.
 ::::
 
 
--- a/docs/extend/tips-for-building.md
+++ b/docs/extend/tips-for-building.md
@@ -579,7 +579,7 @@ elastic-package test [pipeline|static|asset|policy|system] -v -g
 
 6. If a package relies on features available only in a specific Elastic Agent version, use an agent version condition. There are two approaches depending on your needs:
 
-    **Package-level** — restricts all inputs; requires `format_version: 3.6.0` or later and Kibana 9.4 or later:
+    **Package-level** — restricts all inputs; requires Kibana 9.4 or later:
 
     ```yaml
     conditions:


docs/extend/agent-version-conditions.md
```diff
@@ -0,0 +1,93 @@
...
+conditions:
+  agent:
+    version: "^9.3.0"
+```
+
+Use this approach when **all** inputs in the package rely on agent capabilities introduced in a specific version.
+
+::::{note}
+`conditions.agent.version` requires `format_version: 3.6.0` or later in the package manifest.
+::::
```
mrodm, line 36: According to the spec, this field could be set even in previous spec versions. There is no JSON patch removing that field for spec <3.6.0.

Related package-spec Pull Requests:
- https://github.com/elastic/package-spec/pull/999
- https://github.com/elastic/package-spec/pull/1063

But, would these fields require some specific kibana or elastic-agent versions to be taken into account?

mrodm: Maybe should a JSON patch have been added to remove that field in previous specs (<3.6.0) ? @jsoriano @teresaromero

juliaElastic: Yeah, the feature requires kibana >= 9.4, we should mention that in the docs

teresaromero: there was an initial intent to allow this since 3.6.0, but the patch was removed to allow all package versions to adopt the feature, if this was ok in the fleet side [here](https://github.com/elastic/package-spec/pull/999#discussion_r2436592156). @juliaElastic will this break fleet for previous versions? i think the docs should mention that any version of a package can contain this information, but fleet will be able to use it from 9.4

juliaElastic: Yeah for Kibana versions < 9.4, the agent version conditions on the package level won't be used, and the `semverSatisfied` conditions on the template level won't be compiled, so these packages would be invalid when used in an agent policy.

juliaElastic: updated to remove mention of format version 3.6.0

docs/extend/agent-version-conditions.md
```diff
@@ -0,0 +1,99 @@
...
+      ...
+    })
+  )
+{{/semverSatisfies}}
+```
+
+Here the `program:` block is only included in the rendered config when the agent version is `^9.3.0`. Agents on earlier versions receive the configuration without the `program:` key.
+
+
+## How Fleet handles agent version conditions [how-fleet-handles-agent-version-conditions]
```
andrewkroh, line 75: The docs don't currently call out an important edge case for template-level `semverSatisfies` conditions: what happens when the agent binary version changes between restarts.

When the agent starts it uses a cached copy of the policy. There is no agent-side check that compares the version the cached policy was rendered for against the current release version, so the new binary runs the *previously* rendered policy until the first check-in completes and Fleet returns a freshly rendered one.

For most forward upgrades within a major this is harmless because newer agents accept older configs. The risky cases are downgrades and major-version upgrades that remove an input option, or processor that the cached rendering depends on.

I think we should add an `## Upgrade and downgrade considerations` section explains this to package authors. Perhaps we advise them to use the package-level `conditions.agent.version` mechanism when the feature is critical to the integration's correctness (Fleet will exclude the integration entirely from incompatible agents rather than ship a config that may break on a version mismatch.)

docs/extend/agent-version-conditions.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Package-level conditions
+
+When a policy is created or updated, Fleet evaluates the `conditions.agent.version` constraint against the versions of enrolled agents in the policy. If an agent does not satisfy the constraint, Fleet surfaces a warning in the Fleet UI indicating the version mismatch.
+The policy assigned to the earlier version of agents will not include the incompatible integrations.
+
+### Input template-level conditions
+
+Fleet renders `.hbs` templates for each enrolled agent, passing the agent's version as `_meta.agent.version`. The `{{#semverSatisfies}}` helper evaluates the semver constraint at render time:
```
andrewkroh, line 84: When does Fleet render the per-agent template? On every check-in if the reported `elastic.agent.version` changed? Only when the policy revision is bumped? Something else?

Could we add a sentence or two to clarify the mechanics? Even something like "Fleet re-renders the policy when the agent reports a new version on check-in, producing a new per-agent policy revision" would be enough to anchor the reader.

docs/extend/agent-version-conditions.md
```diff
@@ -0,0 +1,99 @@
...
+    post(state.url + "/oauth/token", "application/json", {
+      "client_id": state.client_id,
+      "client_secret": state.client_secret,
+      ...
+    })
+  )
+{{/semverSatisfies}}
+```
+
+Here the `program:` block is only included in the rendered config when the agent version is `^9.3.0`. Agents on earlier versions receive the configuration without the `program:` key.
```
andrewkroh, line 72: As written, this example shows a CEL input where the entire `program:` key is omitted on agents older than 9.3.0. That isn't graceful degradation, it's a CEL input with no `program`, which fails to start. The reader is left thinking older agents don't run this input which is probably not what we want to teach.

I think it would be more useful to show an `{{else}}` branch that supplies an older, working `program:`, e.g.:

    redact:
      fields:
        - client_secret
    program: |
      {{#semverSatisfies _meta.agent.version "^9.3.0"}}
      // uses a built-in available in 9.3+
      state.with(
        post(state.url + "/oauth/token", "application/json", { ... })
      )
      {{else}}
      // older equivalent, valid on agents <9.3
      state.with(
        request("POST", state.url + "/oauth/token").with({ ... }).do_request()
      )
      {{/semverSatisfies}}


This demonstrates the mechanism and shows the recommended pattern that every branch produces a runnable config.

docs/extend/agent-version-conditions.md
```diff
@@ -0,0 +1,99 @@
...
+Both mechanisms are used together when a package supports multiple agent versions but uses newer capabilities when available.
+
+
+## Package-level agent version condition [package-level-agent-version-condition]
+
+Use the `conditions.agent.version` field in `manifest.yml` to require a minimum Elastic Agent version for the entire package. Fleet will warn users if enrolled agents in a policy do not meet this requirement. Agents incompatible with the version condition will not run the integration.
+
+```yaml
+conditions:
+  kibana
```
andrewkroh, line 26: This is malformed YAML. Missing colon.

PR #18205, opened 2026-04-02
[akamai] Fix Pipeline for Populating Http Message Headers

mohitjha-elastic:
## Proposed commit message

```
akamai: fix pipeline for populating http message headers.

The Ingest pipeline failed when responseHeaders or requestHeaders could not be parsed 
as Name: value lines, including Akamai’s {p} placeholder and RFC 7230 obs-fold continuations
(e.g. folded Content-Security-Policy). Non-parseable string values are dropped with a remove 
processor when the field contains no colon, so the KV step is not run on placeholders.
Obs-fold sequences (\r\n followed by space or tab) are normalized to a single space with 
gsub before KV, which matches the spec’s treatment of folded linear whitespace.
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/akamai directory.

> elastic-package test -v

### DIFF
--- a/packages/akamai/data_stream/siem/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/akamai/data_stream/siem/elasticsearch/ingest_pipeline/default.yml
@@ -130,28 +130,23 @@ processors:
       tag: remove_http_message_response_headers_non_kv
       field: json.httpMessage.responseHeaders
       if: ctx.json?.httpMessage?.responseHeaders instanceof String && ctx.json.httpMessage.responseHeaders.contains(":") == false
-  - set:
-      tag: copy_response_headers_for_kv
-      if: ctx.json?.httpMessage?.responseHeaders != null && ctx.json.httpMessage.responseHeaders != ""
-      field: _tmp.akamai_response_headers_kv
-      copy_from: json.httpMessage.responseHeaders
   - gsub:
-      tag: gsub_tmp_response_headers_obs_fold
+      tag: gsub_http_message_response_headers_obs_fold
       description: >
-        On a temp copy only: replace RFC 7230 obs-fold (CRLF + space/tab) with a single SP so
-        each physical line is a valid Name: value for KV. json.httpMessage.responseHeaders is unchanged.
-      field: _tmp.akamai_response_headers_kv
+        Replace RFC 7230 obs-fold (CRLF + space/tab) with a single SP so each line is a valid
+        Name: value for KV. The json subtree is removed later in the pipeline.
+      field: json.httpMessage.responseHeaders
       pattern: '\r\n[\t ]+'
       replacement: ' '
       ignore_missing: true
   - kv:
       tag: kv_httpMessage_responseHeaders
-      field: _tmp.akamai_response_headers_kv
+      field: json.httpMessage.responseHeaders
       target_field: akamai.siem.response.headers
       field_split: '\r\n'
       value_split: ':\s*'
       ignore_missing: true
-      if: ctx._tmp?.akamai_response_headers_kv != null && ctx._tmp.akamai_response_headers_kv != ""
+      if: ctx.json?.httpMessage?.responseHeaders != null && ctx.json.httpMessage.responseHeaders != ""
       trim_key: ' '
       trim_value: ' '
   - urldecode:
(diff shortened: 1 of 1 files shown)


packages/akamai/data_stream/siem/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -130,8 +130,26 @@ processors:
...
+        }
+  - gsub:
+      tag: gsub_http_message_response_headers_unfold
+      field: json.httpMessage.responseHeaders
+      pattern: '\r\n[\t ]+'
+      replacement: ' '
+      ignore_missing: true
   - kv:
-      if: ctx.json.httpMessage?.responseHeaders != ""
+      if: ctx.json?.httpMessage?.responseHeaders != ""
```
chrisberkhout: When you use the null-safe operator you can always get a `null`. So this if condition will be true if you get a null or a non-empty string. If you need to narrow it down further:

```suggestion
      if: ctx.json?.httpMessage?.responseHeaders != null && ctx.json.httpMessage.responseHeaders != ""
```

This form is also possible but only use it when you have longer lists of items you want to check for:
```suggestion
      if: ![null, ""].contains(ctx.json?.httpMessage?.responseHeaders)
```

packages/akamai/data_stream/siem/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -130,8 +130,26 @@ processors:
...
+      tag: script_drop_non_kv_response_headers
+      if: ctx.json?.httpMessage?.responseHeaders != null
+      source: |
+        def rh = ctx.json.httpMessage.responseHeaders;
+        if (rh instanceof String) {
+          def s = (String) rh;
+          if (s.indexOf(':') < 0) {
+            ctx.json.httpMessage.remove('responseHeaders');
+          }
+        }
```
chrisberkhout: This could be simpler as a remove processor. Something like this (condition true if it's not null and it doesn't contain `:`):

```
- remove
    field: ctx.json.httpMessage.responseHeaders
    if: ctx.json?.httpMessage?.responseHeaders?.contains(":") == false
```

packages/akamai/data_stream/siem/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -142,8 +160,26 @@ processors:
       tag: urldecode_httpMessage_requestHeaders
       field: json.httpMessage.requestHeaders
       ignore_missing: true
+  - script:
+      lang: painless
+      tag: script_drop_non_kv_request_headers
+      if: ctx.json?.httpMessage?.requestHeaders != null
```
chrisberkhout: As above.

packages/akamai/data_stream/siem/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -142,8 +160,26 @@ processors:
...
+        }
+  - gsub:
+      tag: gsub_http_message_request_headers_unfold
+      field: json.httpMessage.requestHeaders
+      pattern: '\r\n[\t ]+'
+      replacement: ' '
+      ignore_missing: true
   - kv:
-      if: ctx.json.httpMessage?.requestHeaders != ""
+      if: ctx.json?.httpMessage?.requestHeaders != ""
```
chrisberkhout: As above

mohitjha-elastic:
Thanks @chrisberkhout 
I’ve switched to the `gsub + kv`–only style for obs-fold handling, without a follow-up script. 

Non-parseable placeholders (e.g. {p}) are dropped with a remove processor when the string contains no :``, replacing the earlier painless blocks.
This keeps the pipeline to ingest processors only for the folding path (no marker round-trip script) while still avoiding KV failures on folded CSP-style headers.

Please review the updated pipeline logic and let me know what you think.

packages/akamai/data_stream/siem/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Suggestion applied.

packages/akamai/data_stream/siem/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Handled as suggested above.

packages/akamai/data_stream/siem/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Done as
```
if: ctx.json?.httpMessage?.responseHeaders instanceof String && ctx.json.httpMessage.responseHeaders.contains(":") == false
```
I guess that's slightly safer, because my version could theoretically try to run `.contains(":")` on something that's not a string.

packages/akamai/data_stream/siem/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: Also changed to remove processor.

packages/akamai/data_stream/siem/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -127,29 +127,54 @@ processors:
...
-      tag: remove_http_message_response_headers_empty_placeholder
+      tag: remove_http_message_response_headers_non_kv
       field: json.httpMessage.responseHeaders
-      if: ctx.json.httpMessage?.responseHeaders == "{p}"
+      if: ctx.json?.httpMessage?.responseHeaders instanceof String && ctx.json.httpMessage.responseHeaders.contains(":") == false
+  - gsub:
+      tag: gsub_http_message_response_headers_obs_fold
+      description: >
+        Replace RFC 7230 obs-fold (CRLF + space/tab) with a single SP so each line is a valid
+        Name: value for KV. The json subtree is removed later in the pipeline.
```
chrisberkhout, line 137: Very helpful information. Thanks.

PR #18210, opened 2026-04-02
Bugfix: Safely Handle Malformed UAC events

brian-mckinney:
## Proposed commit message

Bugfix for unchecked decoding of long values in UAC events

## Related issues
Fixes: https://github.com/elastic/sdh-beats/issues/6727

## Testing
```bash
elastic-package test pipeline -v -C packages/windows -d forwarded
elastic-package test pipeline -v -C packages/system -d security
```


andrewkroh:
@macroscope-app review

packages/system/data_stream/security/_dev/test/pipeline/test-security-4738-malformed.json
andrewkroh, line 1: All test files are named test-security-4378 but the event ID is [4738](https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4738). All 8 need renamed.

packages/system/data_stream/security/elasticsearch/ingest_pipeline/standard.yml
```diff
@@ -1213,13 +1213,20 @@ processors:
...
+        Long newUacValue;
+        try {
+          newUacValue = Long.decode(ctx.winlog.event_data.NewUacValue.trim());
+        } catch (Exception e) {
           return;
         }
-        Long newUacValue = Long.decode(ctx.winlog.event_data.NewUacValue);
         ArrayList uacResult = new ArrayList();
         for (entry in params.entrySet()) {
-          Long flag = Long.decode(entry.getKey());
```
andrewkroh, line 1222: The params keys are hardcoded hex literals. If someone were to introduce an invalid key in the future, we'd want the pipeline to fail loudly rather than silently `continue`. I think we should remove this try/catch and keep the original `Long flag = Long.decode(entry.getKey());` so that an invalid params key surfaces as a pipeline error.

andrewkroh (commented):
Both packages need a changelog entry and a patch version bump.

belimawr (approved):
I'm reviewing only the file owned by the data-plane team: `packages/windows/changelog.yml`.

andrewkroh:
As reported by the failing CI, the `version` in the manifest.yml files for each package needs updated to match the changelog version. This is handled automatically if you use `elastic-package -C packages/<package_dir_name> changelog add ...` to add changelog messages.

andrewkroh (approved):

PR #18218, opened 2026-04-03
[Fortinet] Add support for Fortigate 7.6 events and misc improvements

dot-mike:
## Proposed commit message

Adds definitions for 7.6 event fields
Adds 7.6 events test coverage
Adds `event.code` mapping to `event.action` for Fortigate 7.6
Improve `event.outcome` & `event.type` for failure & error events.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/logid.yml
```diff
@@ -0,0 +1,1377 @@
+---
+description: Pipeline for FortiGate logid to event.action mapping
+processors:
+  - script:
+      tag: script_logid_event_action
+      lang: painless
+      description: Map logid to event.action
+      if: ctx.event?.code != null
```
andrewkroh: Move this `if` down so that is is closer to the script `source` to improve the code locality. It's helpful to be able to see the `if` and `source` together on the screen.

packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/logid.yml
```diff
@@ -0,0 +1,1377 @@
...
+        "2401064611": "FortiGate-virtualpatch-localin-vpatch-log"
+        "2500010000": "FortiGate-casb-access-blocked"
+        "2500010001": "FortiGate-casb-access-bypass"
+        "2500010002": "FortiGate-casb-access-monitor"
+      source: |
+        def logid = ctx.event.code;
+        if (params.containsKey(logid)) {
+          if (ctx.fortinet == null) { ctx.fortinet = new HashMap(); }
+          if (ctx.fortinet.firewall == null) { ctx.fortinet.firewall = new HashMap(); }
+          if (ctx.event.action == null) {
```
andrewkroh: This should be elevate to the `if` condition to avoid running the script if `event.action` is already set.

packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/logid.yml
```diff
@@ -0,0 +1,1377 @@
...
+        "2401064610": "FortiGate-virtualpatch-localin-vpatch-block"
+        "2401064611": "FortiGate-virtualpatch-localin-vpatch-log"
+        "2500010000": "FortiGate-casb-access-blocked"
+        "2500010001": "FortiGate-casb-access-bypass"
+        "2500010002": "FortiGate-casb-access-monitor"
+      source: |
+        def logid = ctx.event.code;
+        if (params.containsKey(logid)) {
+          if (ctx.fortinet == null) { ctx.fortinet = new HashMap(); }
+          if (ctx.fortinet.firewall == null) { ctx.fortinet.firewall = new HashMap(); }
```
andrewkroh: The two lines initializing `ctx.fortinet` and `ctx.fortinet.firewall` should be removed since the script only writes to `event.action`.

packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/logid.yml
```diff
@@ -0,0 +1,1377 @@
...
+        "2400064600": "FortiGate-virtualpatch-ot-vpatch-block"
+        "2400064601": "FortiGate-virtualpatch-ot-vpatch-log"
+        "2401064610": "FortiGate-virtualpatch-localin-vpatch-block"
+        "2401064611": "FortiGate-virtualpatch-localin-vpatch-log"
+        "2500010000": "FortiGate-casb-access-blocked"
+        "2500010001": "FortiGate-casb-access-bypass"
+        "2500010002": "FortiGate-casb-access-monitor"
+      source: |
+        def logid = ctx.event.code;
+        if (params.containsKey(logid)) {
```
andrewkroh: Avoid the double `params` hash lookup (containsKey + get) by doing a single lookup and null-checking the result. e.g.

```
def action = params[ctx.event.code];
if (action != null) {
  ctx.event.action = action;
}
```

packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/logid.yml
```diff
@@ -0,0 +1,1377 @@
+---
+description: Pipeline for FortiGate logid to event.action mapping
+processors:
+  - script:
+      tag: script_logid_event_action
+      lang: painless
+      description: Map logid to event.action
+      if: ctx.event?.code != null
+      params:
```
andrewkroh, line 10: Can you add a comment about where this data comes from. This will help reviewers verify the completeness and could help future maintainers the next time this needs an refresh.

andrewkroh (commented):
The change also needs a changelog entry and version bump in manifest.yml. You can use `elastic-package changelog add` to accomplish this.

dot-mike:
@andrewkroh done. Thanks!

andrewkroh (approved):
LGTM.

andrewkroh:
The CI failure is because the readme needs to be regenerated with the newly added field, so you need to run

`elastic-package build`

to update it.

dot-mike:
Done, thanks @andrewkroh ! I learned something new

dot-mike:
I have one request @andrewkroh 
Is it acceptable to rewrite the "event.action" fields for login as well? Today it's just mapped to "login" from https://github.com/dot-mike/elastic-integrations/blob/4da8616fdc19c6dcb057cc773966bbfa77e5f418/packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/login.yml#L8-L12

I would like to set event.action from logid.yml mapping so we get proper action, either success or failed:
https://github.com/dot-mike/elastic-integrations/blob/4da8616fdc19c6dcb057cc773966bbfa77e5f418/packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/logid.yml#L330-L331

I know this may break peoples integration when `event.action` gets a new value. So this is why I'm asking

andrewkroh:
That would be a breaking change to modify the existing `event.action`.

I would not do it in this pull request, but if we wanted to make it consistent, then you can probably do that as a separate breaking-change PR and bump the major version. We should do a bit of research to check the impact, like check the elastic/detection-roles to see if there's anything in there that's already depending on the value. Similarly, check if we have any visualizations or dashboards in the package itself that are dependent upon specific `event.action` values.

dot-mike:
> That would be a breaking change to modify the existing event.action.

I thought so. Thanks for the clarification. Will revise in the future. For now I will apply this to my `@custom` pipeline.

dot-mike:
One thing to add, the log ID mapping is as per 7.6 log ID message reference. 7.6 is now considered mature as of 7.6.5 released December 2025. With the 7.6.x train being EOL in January 2030. Version 7.4 was released May 2023 with EOL in 2028.

So I think it's fair to say the log ID mapping should stay on version 7.6

7.2 is due for EOL in September 2026. So I think it's fair to say that having log ID mapping for latest version is fine. Should maybe be considered update when a new major or minor version gets released.

vinit-chauhan:
Hey @dot-mike - First of all, Thanks for the PR. 

Would you mind resolving the merge conflicts, other than that it and It looks good to me.

dot-mike:
@vinit-chauhan  | @andrewkroh 
I never got around to resolve merge conflicts. Busy IRL.

Please re-open so I can move forward with this. Thanks.

PR #18235, opened 2026-04-07
entityanalytics_ad: add support for empty Active Directory groups

### LINKED ISSUE #16519: [Active Directory Entity Analytics]: Add Groups and Group Policies Datasets
### Integration Name

Active Directory Entity Analytics [entityanalytics_ad]

### Dataset Name

entityanalytics_entra_id.group, entityanalytics_entra_id.gpo

### Integration Version

0.17.1

### Agent Version

9.1.5

Ubuntu 20.04

### User Goal

Ability to leverage the integration instead of using custom scripts

### Existing Features

It just needs more datasets.

### What did you see?

Expected:

Image

### Anything else?

_No response_

efd6:
## Proposed commit message

```
entityanalytics_ad: add group data stream and pipeline

The beats entity-analytics input already collects all Active Directory
groups for membership enrichment; their names and SIDs appear within
user and device entities via the user.group fields. Groups with no
direct members are invisible to this process because they have no
member links to follow.

The include_empty_groups option (elastic/beats#49093) makes the input
emit standalone documents for these otherwise-invisible groups so they
appear in the asset inventory. This change adds the integration-side
support: a group ingest pipeline, a routing rule to direct group
documents to a dedicated data stream, and ECS and vendor field
definitions.

The entity pipeline dispatch is updated to recognise group documents
(ctx.group.id != null) and the fallback entity pipeline condition
now excludes them. Pipeline tags are corrected to distinguish user,
device, and group dispatch.
```

Also adds system testing, including document routing tests (second commit).

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #16519


chrisberkhout (changes requested):
The commit description and the new group pipeline description both mention empty groups specifically.

I see the new option in the input config to include empty groups. Do we get information about non-empty groups? If so, is that existing data that's getting new support (pipeline and data stream) or is that data new?

Can you please rewrite the commit description and pipeline description to clarify the difference between existing and new support for groups, and empty vs non-empty groups.

packages/entityanalytics_ad/data_stream/entity/_dev/test/scripts/env.txt
chrisberkhout, line 1: Was this intended to be committed?

efd6: Yes

efd6:
> Can you please rewrite the commit description and pipeline description to clarify the difference between existing and new support for groups, and empty vs non-empty groups.

[Done](https://github.com/elastic/integrations/compare/322c534d0ade127405e9e906878ac0312ebbfcd9..feef661179a125e3d7d7aed7c926e4889dd2db36). PTAL

packages/entityanalytics_ad/data_stream/entity/_dev/test/pipeline/test-group.json-expected.json
chrisberkhout, line 1: All these events have `asset.type: "activedirectory_user"` - both the markers and the groups.

It looks like two issues:
- For these marker events, with `labels.identity_source: "entity-analytics-entityanalytics_ad.entity-test-group"` in the incoming data, `asset.type: "activedirectory_user"` is probably not the right value. If they're marking start and completion of a sync of users and groups, then maybe the values should change? In the pipeline tests I don't see different markers for devices.
- The group pipeline renames `activedirectory` to `entityanalytics_ad`, but then the maker pipeline condition is true and so the group document is run through that as well.

chrisberkhout:
Sorry for the closer/reopen. I hit the wrong button.

packages/entityanalytics_ad/data_stream/entity/_dev/test/scripts/routed_data_streams.txt
chrisberkhout, line 1: Nice.

packages/entityanalytics_ad/_dev/deploy/docker/ldap-mock-service/main.go
chrisberkhout, line 1: This is nice and clear to read.

packages/entityanalytics_ad/data_stream/group/fields/beats.yml
chrisberkhout, line 1: These are the same for all data streams. You could use `elastic-package link` if you want. Not important though.

packages/entityanalytics_ad/data_stream/entity/_dev/test/scripts/env.txt
chrisberkhout: Oh, I see there are a number of those elsewhere.
It's only testing the test framework though, isn't it?
If these values are helpful maybe `elastic-package test script` should log them?
No problem with adding it for now. Just interested in the use case.

packages/entityanalytics_ad/data_stream/group/manifest.yml
```diff
@@ -0,0 +1,3 @@
+title: Collect Group logs from Active Directory
```
chrisberkhout: Nit: I think the data stream titles could all drop the word "Collect".

Also, `data_stream/entity/manifest.yml:8`, which has the line that appears in the policy editor UI: "Collect User Identities logs from Active Directory.", should no longer mention Users specifically.

chrisberkhout (changes requested):
> > Can you please rewrite the commit description and pipeline description to clarify the difference between existing and new support for groups, and empty vs non-empty groups.
> 
> [Done](https://github.com/elastic/integrations/compare/322c534d0ade127405e9e906878ac0312ebbfcd9..feef661179a125e3d7d7aed7c926e4889dd2db36). PTAL

Thanks! That's much clearer.

Another possibility would be to have the empty group docs in the user data stream, just without the user fields, so they line up with existing group data. But keeping them separate probably is better, and it's well explained in the README, etc.

---

It's separate from this change, but the LLM caught a point where the proposed commit message differs from the implementation. The proposed commit message says:

> their names and SIDs appear within user and device entities via the user.group and device.group fields

However, for devices, they're [put](https://github.com/elastic/integrations/blob/b2d65afe848c6a3774b9cded13e2616a832690b6/packages/entityanalytics_ad/data_stream/entity/elasticsearch/ingest_pipeline/device.yml#L206-L220) into `user.group`, and that's what we see in the existing pipeline test [output](https://github.com/elastic/integrations/blob/b2d65afe848c6a3774b9cded13e2616a832690b6/packages/entityanalytics_ad/data_stream/entity/_dev/test/pipeline/test-device.json-expected.json#L180-L195).

packages/entityanalytics_ad/data_stream/entity/_dev/test/scripts/env.txt
efd6: It's a smoke test for the script test environment variables. Useful when writing new script tests. You can run it alone to confirm the test harness is wiring things correctly before debugging your actual test.

packages/entityanalytics_ad/data_stream/entity/_dev/test/pipeline/test-group.json-expected.json
efd6: Yes, this is a pre-existing issue.

The entity sub-pipelines rename `activedirectory` to `entityanalytics_ad`, which makes the marker pipeline's condition (`ctx.activedirectory == null`) true on entity documents as well. It affects users and devices too. The device pipeline test expectations show the same `asset.type: "activedirectory_user"` on device documents.

I'll fix this in a follow-up.

chrisberkhout (approved):

PR #18244, opened 2026-04-07
[Pulse Connect Secure]: Expand parser coverage and ECS mappings

IOITI:
**CHEOPS Cyberdéfense** give you this enhancement of Pulse Connect Secure integration.

## Proposed commit message

Pulse Secure Connect: expand parser coverage and ECS mappings : 
- Expand ingest pipeline coverage for namespace-prefixed authentication, VPN session, web request, Host Checker, and administrative system logs.
- Normalize parsed events more fully to ECS and add the required Pulse-specific field definitions.
- Upgrade the Pulse Connect Secure integration to version 2.7.0.

## What changed

- Reworked the main ingest pipeline to support a much broader set of Pulse log families.
- Added parsing for:
  - primary and secondary authentication
  - login failures
  - realm restriction checks
  - unauthenticated connection events
  - VPN session start/end
  - session create/delete/timeout/resume
  - connection close and client connection completion
  - web requests
  - Host Checker policy and compliance results
  - virus signature operations
  - integrity scan events
  - syslog configuration changes
  - user account changes on authentication servers
- Added dynamic ECS mapping for event.action, event.category, event.type, event.outcome, event.reason, related.*, source.*, destination.*, url.*, http.*, and network metrics.
- Expanded the Pulse vendor fields under pulse_secure.* for auth servers, bandwidth, Host Checker, integrity, restrictions, syslog settings, transport mode, web request metrics, and session metadata.
- Refreshed the sample event and expanded pipeline fixtures and expected outputs to cover the added log families.
- Rewrote the package README template to document the broader behavior and deployment/validation flow.
- Added the 2.7.0 changelog entry.

## Breaking changes / behavior changes

- No package input or deployment configuration changes.
- Parsed output changed for many events:
  - event.category is now dynamic instead of defaulting to network
  - event.type is now set dynamically for more event families
  - event.action is now populated for recognized events
  - event.outcome and event.reason are populated more consistently
  - empty placeholder values may now be omitted instead of emitted as empty strings
  - user.roles is now derived from Pulse role data
  - source.nat.ip is no longer implicitly copied from client.nat.ip
  - unauthenticated connection events now use the actual connection IP from the message body
  - MAC addresses are normalized to RFC 7042 format (uppercase with hyphens)

## Checklist


packages/pulse_connect_secure/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1,137 +1,683 @@
...
-      copy_from: client
-  - set:
-      field: observer.ip
-      value: ['{{{observer.ip}}}']
-      if: ctx.observer?.ip instanceof String
-  - remove:
-      field:
-        - _tmp
-      ignore_missing: true
+    - set:
```
kgeller: Can we correct the spacing format in this file? 

The change makes it inconsistent with the rest of the repo, but also makes this file more challenging to review since the diff is much larger.

packages/pulse_connect_secure/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -1,137 +1,683 @@
...
+            patterns:
+                - '^%{IP:destination.ip}$'
+                - '^%{DATA:pulse_secure.connection.target} port %{INT:destination.port:long}$'
+            ignore_failure: true
+            if: ctx._tmp?.connection_target != null && ctx._tmp.connection_target != ''
+    - set:
+            field: destination.address
+            copy_from: _tmp.connection_target
+            if: ctx._tmp?.connection_target != null
+    - script:
```
kgeller: This ~300-line script acts as both message classifier and field transformer — setting `event.kind`, running a 25-branch `if/else if` chain on `message.startsWith(...)`, assigning ECS categorization fields, computing derived values, and handling host checker/URL/user enrichment.

The established pattern in this repo is to use **native ingest processors** for ECS categorization:
- **CrowdStrike, Fortinet, PAN-OS** — sub-pipelines per event family with `set`/`append`
- **Cisco FTD** — conditional `set`/`append` by message ID with a small Painless lookup table

A monolithic script makes debugging harder (errors lack branch context), prevents testing event families in isolation, and means every new event type requires editing one large script.

**At minimum:** pull `event.kind = "event"` out into its own `set` processor before the script — if the script fails, `event.kind` is never set.

**Ideally:** split into sub-pipelines per event family (auth, vpn, session, host-checker, config, web-request), route from the grok stage, and use native `set`/`append` for ECS fields. Keep Painless only for what genuinely needs it (string splitting, duration math, bandwidth aggregation).

packages/pulse_connect_secure/changelog.yml
```diff
@@ -1,4 +1,16 @@
 # newer versions go on top
+- version: "2.7.0"
```
kgeller, line 2:
```suggestion
- version: "3.0.0"
```

Breaking changes typically correspond to a bump in the major

kgeller (changes requested):
Thanks for the contribution! The expanded parser coverage, proper ECS categorization, structured field definitions, and rewritten docs are a significant improvement to this integration.

I've left a few comments that need to be addressed before we can merge.

~Since this is a breaking change, we'll also need to do some additional validation on our side before we can approve. We'll follow up once that's done.~

packages/pulse_connect_secure/changelog.yml
kgeller: After discussing internally, the team thinks this isn't a breaking change after all. Will add another comment to get it to an enhancement

packages/pulse_connect_secure/changelog.yml
```diff
@@ -1,4 +1,16 @@
...
+  changes:
+    - description: >
+        Expand Pulse Connect Secure parsing for namespace-prefixed authentication, VPN session, web request, Host Checker, and administrative system logs. Add the related ECS and vendor fields.
+        Breaking changes:
+          - event.category changed from mostly network to per-event dynamic ECS values
+          - event.type is now dynamic and present for far more events
+          - event.action is now populated
+          - some previously empty fields are now omitted
+          - some MAC values are normalized according to RFC 7042
+      type: breaking-change
```
kgeller, line 12:
```suggestion
    - description: >
        Expand Pulse Connect Secure parsing for namespace-prefixed authentication, VPN session, web request, Host Checker, and administrative system logs. Add the related ECS and vendor fields.
      type: enhancement
```

After some internal review, we think this is instead an enhancement

IOITI: Some ECS fields didn't have the same value between the 2.6.0 and this PR, ex `event.category` can be different : see https://github.com/elastic/integrations/pull/18244/changes#diff-4fd413006d84f9876188bcb70c34b4bc915451152ca3d3fb679e61f792136a29R33 

Example: An user having an alerting rules based on `event.category` field value, will have an impact by change proposed with my PR. So it's a breaking change for me.

And [this part doc said:](https://www.elastic.co/docs/extend/integrations/_publish_an_integration#:~:text=Event%20value%20changes%3A%20Changing%20standardized%20values%20(e.g.%2C%20event.outcome%20from%20%22Succeeded%22%20to%20%22success%22))

> Document breaking changes Use breaking-change type in changelog and clearly describe impact on existing users. Breaking changes require a major version bump.
>
> The following changes are considered breaking:
> - ...
> - Event value changes: Changing standardized values (e.g., event.outcome from "Succeeded" to "success")
> - ...

packages/pulse_connect_secure/data_stream/log/elasticsearch/ingest_pipeline/default.yml
IOITI: Corrected in commit bd010f102a000fa779d491ec3420214b8cd5b86d

packages/pulse_connect_secure/data_stream/log/elasticsearch/ingest_pipeline/default.yml
IOITI: Corrected in commit bd010f102a000fa779d491ec3420214b8cd5b86d

packages/pulse_connect_secure/changelog.yml
kgeller: You're totally correct in that the values of `event.category` will change, but the difference between this case and that provided documentation example essentially boils down to backward compatibility. Rules would still trigger for a network type category, but now we simply just have more values as options, not changed values.

IOITI:
@kgeller any news for the review? 😃

IOITI:
Reopen ?

PR #18245, opened 2026-04-07
jumpcloud: fix start_time errors when source_lag_time is configured

kcreddy:
## Proposed commit message
```
jumpcloud: fix start_time errors when source_lag_time is configured

PR #15432 (v1.17.0) added source_lag_time with a guard condition to
clamp start_time when the cursor is ahead of now - source_lag_time.
Two bugs in the implementation cause 400 Bad Request errors from the
JumpCloud API:

1. parseDate uses the default RFC3339 layout, which cannot parse
   JumpCloud's fractional-second timestamps (e.g.
   2023-01-14T08:16:06.495Z). When parsing fails, parseDate silently
   returns zero time, disabling the guard condition entirely.

2. The YAML >- block scalar introduces a trailing newline in the
   if-true branch output. When the guard fires, start_time is sent
   as "2026-03-09T10:46:26Z\n", which JumpCloud rejects as invalid.

Fix (1) by passing "RFC3339Nano" layout to parseDate so it handles
fractional seconds. Fix (2) by adding the -]] trim marker to strip
the trailing newline.

Exercise the source_lag_time code path in the existing system test.
```

## Checklist

### DIFF
--- a/packages/jumpcloud/changelog.yml
+++ b/packages/jumpcloud/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "1.17.1"
   changes:
-    - description: Fix `start_time` guard condition failing to parse JumpCloud's fractional-second timestamps by using RFC3339Nano layout.
+    - description: Fix `start_time` guard condition failing to parse JumpCloud's fractional-second timestamps by using a RFC3339 nanosecond-resolution fractional second layout.
       type: bugfix
       link: https://github.com/elastic/integrations/pull/18245
     - description: Fix trailing newline in `start_time` when the guard condition fires.


packages/jumpcloud/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "1.17.1"
+  changes:
+    - description: Fix `start_time` guard condition failing to parse JumpCloud's fractional-second timestamps by using RFC3339Nano layout.
```
efd6:
```suggestion
    - description: Fix `start_time` guard condition failing to parse JumpCloud's fractional-second timestamps by using a RFC3339 nanosecond-resolution fractional second layout.
```
This moves it from code description to human-facing intention declaration.

PR #18266, opened 2026-04-08
[kafka_log] Update integration to input type to provide multi-signal support

### LINKED ISSUE #18264: [Kafka Log]: Configurable data_stream type
### Integration Name

Kafka [kafka]

### Dataset Name

_No response_

### Integration Version

3.0.2

### Agent Version

9.3.2

Windows Server 2022

### User Goal

Allow users to select the data_stream type upon creation. 

### Existing Features

The current data_stream type is fixed as logs. 

### What did you see?

Ingested data land in the logs-* pattern as the data_stream type is fixed

### Anything else?

_No response_

adrianchen-es:
## Proposed commit message
Enhance the Kafka log integration to allow ingestion of metrics. 

The pattern implemented aligns with other integrations that allow both log and metric ingestions. 
A dynamic type is not used as it will require existing users to recreate the integration whereas a separate datastream reduces the potential complication and allows existing integrations to enable metrics or switch. 

## Checklist

## Author's Checklist

- [ ] 

## Related issues
Closes https://github.com/elastic/integrations/issues/18264

-


agithomas:
Could you also add the recommended pipeline tests as part the PR.

++ @stefans-elastic for reviewing the PR.

cc @lalit-satapathy

adrianchen-es:
I am not 100% clear on the test failures for the integration test for kafka_log - https://buildkite.com/elastic/integrations/builds/41142#019d7164-02ba-44b9-96d7-9e5ed4848465, as a local test succeeds. 
image

adrianchen-es:
@stefans-elastic, any pointers on the integration test error or other changes that are required to progress this?
Just re-ran `elastic-package test system -v` within the kafka_log package locally, and it passes. 

image

stefans-elastic:
@adrianchen-es

Unfortunately I wasn't able to test it locally (because `The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested`). And I don't see obvious reason for the failure. However I got AI to look at it and here is the result(I hope it is helpful):
 
Root Cause: kafka_log.metrics CI test failure

  In CI, all datastream tests in a package share the same Kafka broker, while locally each test gets its own fresh broker. The generic test runs first and consumes all messages from
  testTopic under group_id: system_test, committing the offset to the end of the log. When the metrics test starts with the identical topics: [testTopic] + group_id: system_test, Kafka
  returns the committed offset — nothing to read — and the test times out after 642s.

  Fix: At minimum, change group_id: system_test → group_id: system_test_metrics in data_stream/metrics/_dev/test/system/test-kafka-config.yml. Ideally also add a dedicated kafka-metrics
  docker service writing metric-shaped JSON ({"metric":{"name":"cpu.usage","value":0.42}}) to a separate testTopicMetrics topic to fully isolate the two tests.

adrianchen-es:
> @adrianchen-es
> 
> Unfortunately I wasn't able to test it locally (because `The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested`). And I don't see obvious reason for the failure. However I got AI to look at it and here is the result(I hope it is helpful):
> 
> Root Cause: kafka_log.metrics CI test failure
> 
> In CI, all datastream tests in a package share the same Kafka broker, while locally each test gets its own fresh broker. The generic test runs first and consumes all messages from testTopic under group_id: system_test, committing the offset to the end of the log. When the metrics test starts with the identical topics: [testTopic] + group_id: system_test, Kafka returns the committed offset — nothing to read — and the test times out after 642s.
> 
> Fix: At minimum, change group_id: system_test → group_id: system_test_metrics in data_stream/metrics/_dev/test/system/test-kafka-config.yml. Ideally also add a dedicated kafka-metrics docker service writing metric-shaped JSON ({"metric":{"name":"cpu.usage","value":0.42}}) to a separate testTopicMetrics topic to fully isolate the two tests.

@stefans-elastic, that was already done, but the results were the same - https://github.com/elastic/integrations/pull/18266/changes/c787fbba1f34aff5cda2fffcb749a98698002488
I reverted that as it made no material difference (at least for the non-local test). 

Both local logs and the CI test logs suggest new containers were created.

image

image

I will nevertheless update the consumer group and see if the CI improves.

adrianchen-es:
Tried with a different consumer group again with the same CI results. 
image

packages/kafka_input/manifest.yml
```diff
@@ -0,0 +1,276 @@
...
+      - name: wait_close
+        type: text
+        title: Wait Close
+        description: When shutting down, how long to wait for in-flight messages to be delivered and acknowledged.
+        required: false
+        show_user: false
+      - name: isolation_level
+        type: text
+        title: Wait Close
+        description: This configures the Kafka group isolation level, supports the values "read_uncommitted" which returns all messages in the message channel and "read_committed" which hides messages that are part of an aborted transaction. The default is "read_uncommitted".
```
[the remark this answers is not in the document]
adrianchen-es: Resolved for the new metrics data stream and the same error in the generic data stream

adrianchen-es:
@stefans-elastic , any insight? The test container is now at the data stream level, the test data is now tailored, there is a different topic and consumer group. I've reduced the `wait_for_data_timeout` since successes are within 10-15s

stefans-elastic:
@adrianchen-es not much yet. But I've figured a way to run the tests locally the way they are run in CI (and they indeed fail). I'm trying to troubleshoot.
I can share the setup in case you want to reproduce this locally:
file `local-test.env`:
```
export YQ_VERSION=v4.35.2
export SETUP_GVM_VERSION=v0.6.0
export JQ_VERSION=1.7
export GH_CLI_VERSION=2.29.0
export KIND_VERSION=v0.27.0
export K8S_VERSION=v1.33.0
export SERVERLESS=false
export STACK_VERSION=8.17.0
export UPLOAD_SAFE_LOGS=0
```
then run the command:
```
source local-test.env && .buildkite/scripts/test_one_package.sh kafka_log
```

stefans-elastic:
@adrianchen-es 
I've managed to get it working locally. Here is patch:
```
diff --git a/packages/kafka_log/data_stream/metrics/_dev/test/system/test-kafka_metric-config.yml b/packages/kafka_log/data_stream/metrics/_dev/test/system/test-kafka_metric-config.yml
index c428ce9a9b..451896f0c2 100644
--- a/packages/kafka_log/data_stream/metrics/_dev/test/system/test-kafka_metric-config.yml
+++ b/packages/kafka_log/data_stream/metrics/_dev/test/system/test-kafka_metric-config.yml
@@ -1,6 +1,6 @@
 service: kafka-svc
 input: kafka
-wait_for_data_timeout: 30s
+wait_for_data_timeout: 5m
 assert:
   min_count: 2
 data_stream:
diff --git a/packages/kafka_log/data_stream/metrics/fields/fields.yml b/packages/kafka_log/data_stream/metrics/fields/fields.yml
index d97c400b0d..1be49ebe18 100644
--- a/packages/kafka_log/data_stream/metrics/fields/fields.yml
+++ b/packages/kafka_log/data_stream/metrics/fields/fields.yml
@@ -6,11 +6,9 @@
   fields:
     - name: name
       type: keyword
-      dimension: true
       description: Metric name extracted from the JSON payload.
     - name: value
       type: double
-      metric_type: gauge
       description: Metric value extracted from the JSON payload.
 - name: kafka.headers
   description: Included Kafka headers
diff --git a/packages/kafka_log/data_stream/metrics/manifest.yml b/packages/kafka_log/data_stream/metrics/manifest.yml
index ab31ed01e0..14ffa3fc3d 100644
--- a/packages/kafka_log/data_stream/metrics/manifest.yml
+++ b/packages/kafka_log/data_stream/metrics/manifest.yml
@@ -1,5 +1,5 @@
 title: Custom Kafka Metrics
-type: metrics
+type: logs
 streams:
   - input: kafka
     description: Collect metric data from Kafka topic with Elastic Agent.
@@ -272,5 +272,3 @@ streams:
 elasticsearch:
   dynamic_dataset: true
   dynamic_namespace: true
-  source_mode: synthetic
-  index_mode: time_series
```

Comments on the changes (AI generated):
---
manifest.yml: type: metrics → type: logs

  The Filebeat kafka input is classified by Fleet as a logs-type input regardless of what the manifest declares. Fleet's processor always injects data_stream.type: logs into every event.
  Meanwhile, elastic-package generates a component template from the manifest's declared type, setting data_stream.type as a constant_keyword with value: metrics. When a document arrives
  with type: logs but the index template expects type: metrics, Elasticsearch rejects it with a 400. Because data streams are created lazily on the first successful document, no document
  ever lands, the data stream is never created, and the test sees 0 hits. Changing the declared type to logs aligns the component template with what Fleet actually sends.

  ---
  manifest.yml: remove source_mode: synthetic and index_mode: time_series

  index_mode: time_series (TSDB) requires every document to contain all routing dimension fields — missing dimensions are rejected. source_mode: synthetic reconstructs _source from stored
  fields rather than storing it verbatim, which requires strict mapping compatibility. Both are advanced modes intended for purpose-built metrics pipelines (counters, gauges with fixed
  cardinality). A kafka log input that relays arbitrary JSON messages is not that pipeline — these modes only add rejection surface without benefit.

  ---
  fields.yml: remove dimension: true and metric_type: gauge

  dimension: true is the TSDB routing path annotation — it only makes sense when index_mode: time_series is active. With TSDB removed, the field is just a plain keyword. metric_type: gauge
   is an ES mapping parameter that only has meaning (and causes issues) inside a TSDB index. Removing both keeps the field definitions clean and avoids mapping conflicts if the package is
  ever installed against a stack that enforces stricter TSDB validation.

  ---
  test-kafka_metric-config.yml: wait_for_data_timeout: 30s → 5m

  On a fresh stack, the test environment needs to: start Kafka, produce messages, have Kibana finish uploading the package (component templates + ingest pipeline), have Fleet push the
  policy to the agent, have the agent connect to Kafka and consume, and have Elasticsearch index at least 2 documents. 30 seconds is not enough headroom for all of that, especially on CI
  where Kibana's Fleet upload rate-limiter can delay package installation. 5 minutes gives the stack enough time to stabilize before the assertion fires.

adrianchen-es:
Hey @stefans-elastic .
Changing it to log type defeats the purpose of the PR? Or would the data stream type remain? 

If the failure is due to hardcoding in filebeat, it should also fail in my local environment? Given it isn't customised.

I'll tweak it aligning to your patch and see what the CI test returns with

adrianchen-es:
This does override the data stream back to logs - logs-kafka_log.metrics-60924.

adrianchen-es:
@stefans-elastic , not to waste your time. 

Is it currently impossible to have metrics as the data_stream.type? 
I've attempted to mimic how the WMI input works, and the UI appears to have a selector for the data_stream.type. 
But it seems that it ends up as logs still.

stefans-elastic:
@adrianchen-es 
it seems to me that `kafka_input` package is pretty much the same as `kafka_log` so is there a reason to introduce it?

adrianchen-es:
> @adrianchen-es it seems to me that `kafka_input` package is pretty much the same as `kafka_log` so is there a reason to introduce it?

Hey @stefans-elastic, could you help with the question from [here](https://github.com/elastic/integrations/pull/18266#issuecomment-4236240567)?
> @stefans-elastic , not to waste your time.
> 
> Is it currently impossible to have metrics as the data_stream.type? I've attempted to mimic how the WMI input works, and the UI appears to have a selector for the data_stream.type. But it seems that it ends up as logs still.

The aim is to have a method to ingest metrics from the Kafka input. If that is hard-coded and not possible. I will close this (or pause it and look at the package-spec/filebeat).

stefans-elastic:
@adrianchen-es 
Unfortunately I don't have a straight answer to this. But given that kafka input is filebeat's feature (not metricbeat's) it is possible that the answer is no. But I'm not sure on this so 'second opinion' is needed here (maybe @agithomas you can help here?)

adrianchen-es:
> @adrianchen-es Unfortunately I don't have a straight answer to this. But given that kafka input is filebeat's feature (not metricbeat's) it is possible that the answer is no. But I'm not sure on this so 'second opinion' is needed here (maybe @agithomas you can help here?)

@agithomas any chance you could help us clarify this?

agithomas:
@stefans-elastic , @adrianchen-es 

Please take a look at the references mentioned below

- https://github.com/elastic/kibana/issues/255486
- https://github.com/elastic/kibana/pull/255732
- https://github.com/elastic/kibana/pull/214216

Based on my understanding, this applies to input packages only. Also, to take advantage, the stack must be `^9.4.0`.

adrianchen-es:
> @stefans-elastic , @adrianchen-es 
> 
> 
> 
> Please take a look at the references mentioned below
> 
> 
> 
> - https://github.com/elastic/kibana/issues/255486
> 
> - https://github.com/elastic/kibana/pull/255732
> 
> - https://github.com/elastic/kibana/pull/214216
> 
> 
> 
> Based on my understanding, this applies to input packages only. Also, to take advantage, the stack must be `^9.4.0`.
> 
> 

Thank you @agithomas

adrianchen-es:
Revert to Draft pending 9.4.0

adrianchen-es:
It appears the current test infra can't validate the change for data stream type until this is merged https://github.com/elastic/elastic-package/pull/3509

packages/kafka_log/_dev/deploy/docker/docker-compose.yml
```diff
@@ -1,21 +1,33 @@
...
-      start_period: 10s
-    hostname: kafka-service
-    environment:
-      KAFKA_LISTENERS: "INTERNAL://kafka-service:9092,EXTERNAL://:9094, CONTROLLER://:9093"
-      KAFKA_ADVERTISED_LISTENERS: "INTERNAL://kafka-service:9092,EXTERNAL://kafka-service:9094"
-      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: "INTERNAL:PLAINTEXT,EXTERNAL:PLAINTEXT, CONTROLLER:PLAINTEXT"
-      KAFKA_INTER_BROKER_LISTENER_NAME: INTERNAL
+    # Official Apache Kafka image (KRaft). See https://hub.docker.com/r/apache/kafka
+    image: apache/kafka:4.1.2
+    hostname: kafka-input-svc
```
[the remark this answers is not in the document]
adrianchen-es: Resolved

packages/kafka_log/manifest.yml
```diff
@@ -1,25 +1,287 @@
...
 title: Custom Kafka Logs
 description: Collect data from kafka topic with Elastic Agent.
-type: integration
-version: "1.9.1"
+type: input
+version: "2.0.0"
 conditions:
   kibana:
-    version: "^8.13.0 || ^9.0.0"
+    version: "^9.4.0"
```
tommyers-elastic: do we need to bump the version at all? input packages are supported in the stack since ~8.5.0

packages/kafka_log/docs/README.md
```diff
@@ -1,17 +1,107 @@
...
-## Ingest Pipelines
-Custom ingest pipelines may be added by adding the name to the pipeline configuration option, creating custom ingest pipelines can be done either through the API or the [Ingest Node Pipeline UI](/app/management/ingest/ingest_pipelines/).
+Elastic Agent connects to your cluster using the **bootstrap hosts** you configure, joins a **consumer group** (`group_id`), and subscribes to one or more **topics**. Messages are read by the agent, enriched with Kafka metadata (for example topic, partition, offset), and written to Elasticsearch using the **dataset** name you choose (default `kafka_log.generic`). Optional **SASL**, **Kerberos**, and **TLS** settings secure the connection to the brokers. Optional **parsers** and **processors** adjust the payload on the agent before ingest.
 
-**ECS Field Reference**
+## What data does this integration collect?
 
-Please refer to the following [document](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) for detailed information on ECS fields.
\ No newline at end of file
+The integration collects **log events** derived from Kafka messages:
```
tommyers-elastic: with this change, this is no longer true. we can now support collection of any signal type from kafka.

packages/kafka_log/manifest.yml
```diff
@@ -1,25 +1,287 @@
...
+        show_user: false
+      - name: wait_close
+        type: text
+        title: Wait Close
+        description: When shutting down, how long to wait for in-flight messages to be delivered and acknowledged.
+        required: false
+        show_user: false
+      - name: isolation_level
+        type: text
+        title: Isolation Level
```
tommyers-elastic, line 211: i validated that this config is an exact copy-paste of the orignal data stream manifest, and it is, with the exception of this one line, which was an error in the original 😄. nice !

packages/kafka_log/manifest.yml
ishleenk17: Is there some functionality added for input package in 9.4 that we need this change ?

tommyers-elastic (commented):
i'm generally happy with this change. we just need to validate that existing package users can upgrade to the new input package with no breaking changes or weird behaviour.

the only other thing that's a potential blocker for me is the kibana compatibility changes.

ishleenk17:
One thing which we should verify is that the upgrade scenario works. Since the input config file moved from datastream level to the root level of the folder. 

@stefans-elastic : Could you pls help with this validation ?

stefans-elastic:
I've tested "Upgrade" scenario locally. Everything went well - no errors occurred.
I ingested 32 events before upgrade and 33 events post upgrade and all 65 documents are present in ES:
Screenshot 2026-04-30 at 16 47 56

cc @ishleenk17  @tommyers-elastic

packages/kafka_log/manifest.yml
adrianchen-es: I bumped the version due to this

https://github.com/elastic/kibana/pull/255732

packages/kafka_log/docs/README.md
adrianchen-es: I'll update it to just events

packages/kafka_log/manifest.yml
tommyers-elastic: got it. 

i don't think this should affect our compatibility guarantees. the package still works on the existing stack versions. it just means that this particular bug is fixed in 9.4 so if you want to use the data stream type override you gotta use 9.4+.

adrianchen-es: Should we bring this back down? 

My concern is that a customer tries to use this and it doesn't behave as expected if they're on versions older than 9.4. 

Where gating would mitigate this

tommyers-elastic: i understand your rationale, but i don't think this bug justifies the compatibility jump. just because there's a bug in <9.4 for an advanced option doesn't mean we should stop people on 9.2 using the package entirely. by that logic we would have to bump all our other input packages to 9.4 compatibility, which is something i don't think we want to do.

adrianchen-es: That's a fair point. Version bump reverted.

packages/kafka_log/_dev/build/docs/README.md
```diff
@@ -1,17 +1,107 @@
-# Custom Kafka Log integration
+# Custom Kafka Logs Integration for Elastic
```
ishleenk17: Shall we not use the word Logs here ? Since I suppose we want to collect data other than logs as well

ishleenk17 (approved):

ishleenk17:
1 nit, else looks good!

packages/kafka_log/_dev/build/docs/README.md
adrianchen-es: Custom Kafka Event integration?

adrianchen-es: Changed it to `# Custom Kafka Event Integration`

adrianchen-es:
@ishleenk17 , @tommyers-elastic or @stefans-elastic 
Thank you for the review and approval.

Could I please get a final look before I merge?
I updated the README per the last comment.

adrianchen-es:
Thanks @ishleenk17 
Renamed it to `Custom Kafka Integration`

packages/kafka_log/_dev/build/docs/README.md
```diff
@@ -1,17 +1,107 @@
-# Custom Kafka Log integration
+# Custom Kafka Integration
 
-The custom Kafka log integration is used to read from topics in a Kafka cluster.
+## Overview
 
-To configure this integration, specify a list of one or more hosts in the cluster to bootstrap the connection with, a list of topics to track, and a group_id for the connection.
+The Custom Kafka Logs integration is an **input** package for Elastic Agent. It runs the Filebeat **Kafka** input so agents can **consume** records from Apache Kafka topics and ship them to Elasticsearch. Use it when applications or pipelines already publish logs or events to Kafka and you want Elastic to read from those topics without an intermediate forwarder.
```
ishleenk17:
```suggestion
The Custom Kafka integration is an **input** package for Elastic Agent. It runs the Filebeat **Kafka** input so agents can **consume** records from Apache Kafka topics and ship them to Elasticsearch. Use it when applications or pipelines already publish logs or events to Kafka and you want Elastic to read from those topics without an intermediate forwarder.
```

Issue #10784, opened 2024-08-13
[aws]: inspector2/guarduty/securityhub - no role_arn or session_token


jdnurmi:
### Integration Name

AWS [aws]

### Dataset Name

guardduty

### Integration Version

v2.22.1

### Agent Version

8.15.0

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.15.0

### OS Version and Architecture

(Cloud)

### Error Message

```
{"log.level":"error","@timestamp":"2024-08-13T19:58:17.344Z","message":"Error while processing http request: failed to collect first response: failed to execute http POST: server responded with status code 403: {\"message\":\"The security token included in the request is invalid.\"}","component":{"binary":"filebeat","dataset":"elastic_agent.filebeat","id":"httpjson-31962c63-cb07-4dfd-8008-c45dda472331","type":"httpjson"},"log":{"source":"httpjson-31962c63-cb07-4dfd-8008-c45dda472331"},"log.origin":{"file.line":181,"file.name":"httpjson/input.go","function":"github.com/elastic/beats/v7/x-pack/filebeat/input/httpjson.run.func1"},"log.logger":"input.httpjson-cursor","service.name":"filebeat","id":"httpjson-aws.guardduty-23af25d5-57be-472e-84d1-85d7bea197ff","input_source":"https://guardduty.us-east-2.amazonaws.com/detector/18c0b8b5740bb519ba47e3361517d029/findings","input_url":"https://guardduty.us-east-2.amazonaws.com/detector/18c0b8b5740bb519ba47e3361517d029/findings","ecs.version":"1.6.0","ecs.version":"1.6.0"}
```

Same for inspector2 and security hub.

Best I can determine, if https://github.com/elastic/integrations/blob/e6362aa697abb3c2b2ff0f6e200a4b7bb5fcb638/packages/aws/data_stream/securityhub_findings/agent/stream/httpjson.yml.hbs#L37 is the source of signing, that's wrong for ephemeral credentials (missing x-amz-security-token / SESSION_TOKEN) , which I discovered while trying to use an assumed role (from valid instance credentials) - it doesn't appear that signing method takes role_arn into account at all either.

### What did you do?

Nodes in EKS, have a valid pod role assigned;  Want _this_ integration to use those credentials to assume `role_arn` into a foreign (in-org) account to access securityhub and other services.

### What did you see?

Nothing, but digging the logs yielded the above

### What did you expect to see?

Happy security dashboards!

[...]

PR #18333, opened 2026-04-10
aws/config: validate credentials before SigV4 signing

### LINKED ISSUE #18332: [aws] AWS Config: improve error when credentials are missing and clarify Role ARN limitation in docs
### Integration Name

AWS [aws]

### Dataset Name

config

### Description

The AWS Config data stream uses the CEL input with hand-rolled AWS SigV4 request signing. This means it **requires** static credentials (`access_key_id` + `secret_access_key`) and does not support Role ARN, IAM instance profiles, or any other AWS SDK credential resolution method.

Two problems stem from this:

#### 1. Unhelpful error when credentials are missing

When `secret_access_key` is not provided, the CEL program crashes with:

```
ERROR: 
:215:31: no such key: secret_key | !has(config_rules.worklist) ? // Exit early due to POST failure. | ..............................^
```

This gives no indication that the user needs to provide static AWS credentials. The CEL program accesses `state.secret_key` and `state.access_key` without `has()` guards in multiple places ([cel.yml.hbs lines 67, 98, 115, 196, 227, 252, 332, 358, 371](https://github.com/elastic/integrations/blob/main/packages/aws/data_stream/config/agent/stream/cel.yml.hbs)), while `session_token` is correctly handled as optional throughout using `state.?session_token` and `has()` checks.

The program should detect missing credentials early and emit a clear error like:

> "AWS Config requires access_key_id and secret_access_key. Role ARN and other credential methods are not currently supported for this data stream."

#### 2. Documentation does not mention Role ARN limitation

The [Config docs](https://github.com/elastic/integrations/blob/main/packages/aws/docs/config.md) say:

> For the current integration package, it is compulsory to add Secret Access Key and Access Key ID.

This is correct but insufficient. Users coming from other AWS data streams (S3, CloudWatch, metrics) where Role ARN works expect it to work here too. The docs should explicitly state that Role ARN / IAM instance profile authentication is not supported for this data stream and explain why (the CEL input does not use the AWS SDK for authentication).

The package-level manifest marks `access_key_id` and `secret_access_key` as `required: false`, which is correct for the AWS package as a whole (most data streams support Role ARN via native AWS inputs), but misleading for this specific data stream.

### Context

This same hand-rolled SigV4 limitation affects five other data streams: `guardduty`, `inspector`, `securityhub_findings`, `securityhub_findings_full_posture`, and `securityhub_insights`. Those use httpjson rather than CEL, but the authentication approach is identical.

Work is underway to add `auth.aws` support to both the httpjson and CEL inputs ([elastic/beats#49956](https://github.com/elastic/beats/pull/49956), [#16985](https://github.com/elastic/integrations/pull/16985)), which will resolve this limitation long-term by delegating authentication to the AWS SDK. This issue covers interim improvements until that work lands:

- Add `has()` guards to the CEL program so missing credentials produce a clear error instead of a crash
- Update the Config data stream docs to explicitly state that Role ARN is not supported

### Related

- #10784 — same SigV4 limitation for GuardDuty/Inspector/SecurityHub (httpjson)
- [elastic/beats#49956](https://github.com/elastic/beats/pull/49956) — backporting `auth.aws` to 8.19.x
- #16985 — adding `auth.aws` to GuardDuty httpjson and `var_groups` for credential type selection

kcreddy:
## Proposed commit message
```
aws/config: validate credentials before SigV4 signing

The AWS Config CEL program accesses state.secret_key and
state.access_key without guards. When a user configures the
integration with Role ARN or no credentials (both valid for other
AWS data streams), the program crashes with the unhelpful error
"no such key: secret_key".

Add a credential check at the top of the CEL program that detects
missing or empty access_key_id / secret_access_key and returns a
clear error explaining that static credentials are required and
Role ARN is not supported for this data stream. The check uses
optional field access (state.?secret_key.orValue("")) to handle
both absent and empty-string cases safely.

Update the data stream docs and manifest description to state the
Role ARN limitation upfront. Add a script test that verifies the
error message is emitted when credentials are missing.

Fixes #18332
```

## Checklist

**Script tests run sucessfully**
```
# Test that the AWS Config data stream emits a clear error event when
# access_key_id and secret_access_key are not provided. (41.941s)
# The credential check fires before any HTTP request, so exactly one error
# event should be indexed per evaluation cycle. Wait for at least 1. (8.027s)
# Verify the error message names the missing credentials. (0.039s)
# Verify no data events were produced — only error events. (0.023s)
# Clean up. (12.967s)
PASS

--- Test results for package: aws - START ---
╭─────────┬─────────────┬───────────┬─────────────────────┬────────┬────────────────╮
│ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME           │ RESULT │   TIME ELAPSED │
├─────────┼─────────────┼───────────┼─────────────────────┼────────┼────────────────┤
│ aws     │ config      │ script    │ env                 │ PASS   │    32.725625ms │
│ aws     │ config      │ script    │ missing_credentials │ PASS   │ 1m3.002499542s │
╰─────────┴─────────────┴───────────┴─────────────────────┴────────┴────────────────╯
--- Test results for package: aws - END   ---
Done
```

## Related issues

- Closes https://github.com/elastic/integrations/issues/18332


packages/aws/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "6.4.4"
+  changes:
+    - description: Add credential validation to AWS Config CEL program to emit a clear error when access_key_id or secret_access_key are missing.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/18333
+    - description: Document that the AWS Config data stream requires static credentials and does not support Role ARN or other SDK-based credential methods.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/18333
```
ShourieG: Both are pointing to the same pr here, both are labelled as a bugfix. Can just condense this into one description?

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -44,6 +44,27 @@ redact:
...
+            "message": "AWS Config requires access_key_id and secret_access_key. Role ARN, IAM instance profiles, and other credential methods are not supported for this data stream.",
+          },
+        },
+        "want_more": false,
+        ?"session_token": state.?session_token,
+        ?"access_key": state.?access_key,
+        ?"secret_key": state.?secret_key,
+        "batch_size": state.batch_size,
+        "aws_region": state.aws_region,
+        "tld": state.tld,
```
ShourieG: Since we are doing a hard stop/ erroring out here, any reason to keep the state vars ?

packages/aws/changelog.yml
kcreddy: Updated in [fae22ac](https://github.com/elastic/integrations/pull/18333/commits/fae22acaaeafbad9887b0304e50311a3fc00064a)

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
kcreddy: Agreed. The single-object error form should reset the state from the initial config. These vars are unnecessary. Removed them in [fae22ac](https://github.com/elastic/integrations/pull/18333/commits/fae22acaaeafbad9887b0304e50311a3fc00064a).

Note: the existing error returns at lines 255 and 360 carry the same unnecessary state vars. I've left those as-is to keep this PR focused, but they could be cleaned up separately.

ShourieG (approved):

efd6:
Note that v8.19.4 and v9.2.1 both have access to [AWS SigV4 signing functions](https://pkg.go.dev/github.com/elastic/mito/lib#AWS) which offer env, shared and static credentials.

kcreddy:
> Note that v8.19.4 and v9.2.1 both have access to [AWS SigV4 signing functions](https://pkg.go.dev/github.com/elastic/mito/lib#AWS) which offer env, shared and static credentials.

In another discussion https://github.com/elastic/integrations/issues/10784#issuecomment-4214979899, there was another feature `auth.aws` (now backported into `8.19`: https://github.com/elastic/beats/pull/50141) which can also do this. The idea here was to use this `auth.aws` once it is available in `8.19`.

Is there any preferred method between the two (mito signing functions vs `auth.aws`)?

efd6:
I completely forgot about that. The discussion [here](https://github.com/elastic/security-team/issues/13112#issuecomment-3139188187) is good for knowing which. If it was your plan to wait for that, that's fine.

PR #18337, opened 2026-04-10
EntityAnalytics AD: add support for ECS entity fields

chemamartinez:
## Proposed commit message

```
Maps lastLogonTimestamp, UAC flags, and DN-valued LDAP attributes
(directReports, managedObjects) to user.entity.* and host.entity.* fields:

- user.entity.lifecycle.last_activity from lastLogonTimestamp.
- user.entity.attributes.mfa_enabled set to true when the SMARTCARD_REQUIRED
  flag is present in userAccountControl. No additional permissions or
  configuration required, as userAccountControl is always fetched.
- user.entity.relationships.supervises ({id, name, domain}) from the
  directReports attribute. Requires adding directReports to the User Attributes
  setting. Note: specifying any value in User Attributes switches the LDAP
  request from "all attributes" to an explicit list, so all previously implicit
  attributes must be added manually.
- user.entity.relationships.administers ({host.id, host.name, user.domain})
  from the managedObjects attribute. Requires adding managedObjects to the User
  Attributes setting, with the same caveat as above.
- host.entity.lifecycle.last_activity from lastLogonTimestamp.
- host.entity.attributes.managed set to true when WORKSTATION_TRUST_ACCOUNT
  or SERVER_TRUST_ACCOUNT is present in userAccountControl.
- host.entity.relationships.supervises ({id, name, domain}) from the
  directReports attribute.
- host.entity.relationships.administers ({host.id, host.name, user.domain})
  from the managedObjects attribute.

DN values in directReports and managedObjects are parsed into structured
objects extracting the CN component as name, the full DN as id, and the
DC components joined as a dotted domain string.

Test samples are AI-generated logs with random data.
```

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/18288


packages/entityanalytics_ad/data_stream/entity/elasticsearch/ingest_pipeline/device.yml
```diff
@@ -232,6 +250,71 @@ processors:
...
+          if (!dcParts.isEmpty()) {
+            def domain = dcParts[0];
+            for (int i = 1; i < dcParts.size(); i++) {
+              domain += '.' + dcParts[i];
+            }
+            result['domain'] = domain;
+          }
+          return result;
+        }
+        def buildRel(def dns) {
```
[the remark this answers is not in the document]
chemamartinez: We're aware of this limitation, but there are a few reasons why we've opted to keep flat parallel arrays rather than restructuring to nested objects:

- Nested mappings are not an option as ES|QL doesn't support nested field types (see [Elasticsearch docs](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/nested)).

- Index correlation is already broken at the storage layer — since the sub-fields (id, name, domain) are mapped as flat keyword arrays under a group, Elasticsearch provides no positional guarantee between them regardless of how the script emits the data.

- Consumers of these fields only need the full set of values for each attribute independently (e.g. all related IDs, all related names). Correlating a specific name to a specific ID within the same relationship group is not a requirement we need to support right now.

Given these constraints, emitting a flat list of each value without cross-field correlation is the right trade-off for us.

chemamartinez:
Moved to draft again because of an ongoing discussion about subfields in relationships fields.

packages/entityanalytics_ad/data_stream/device/fields/ecs.yml
```diff
@@ -102,3 +102,49 @@
...
+            - name: owns
+              type: group
+              fields:
+                - name: id
+                  type: keyword
+                - name: name
+                  type: keyword
+                - name: domain
+                  type: keyword
+            - name: administered_by
```
uri-weisman: Entity store already uses `administers` and not `administered_by`, so we'll stick with that (that was one of Andrew's suggestions).

packages/entityanalytics_ad/data_stream/entity/elasticsearch/ingest_pipeline/device.yml
```diff
@@ -232,6 +250,93 @@ processors:
...
+          def rel = ['host': [:]];
+          def dnList = (dns instanceof List) ? dns : [dns];
+          for (def dn : dnList) {
+            if (!(dn instanceof String) || dn.isEmpty()) continue;
+            def p = parseDn(dn);
+            if (!rel['host'].containsKey('id')) rel['host']['id'] = new ArrayList();
+            rel['host']['id'].add(p['id']);
+            if (p.containsKey('name')) {
+              if (!rel['host'].containsKey('name')) rel['host']['name'] = new ArrayList();
+              rel['host']['name'].add(p['name']);
```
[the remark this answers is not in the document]
chemamartinez: This is intended, host.domain is not part of the current schema. On the other hand, populating user.domain helps Entity Store to correlate to existing entities.

packages/entityanalytics_ad/data_stream/device/fields/ecs.yml
```diff
@@ -102,3 +102,61 @@
...
+          type: group
+          fields:
+            - name: managed
+              type: boolean
+        - name: lifecycle
+          type: group
+          fields:
+            - name: last_activity
+              type: date
+        - name: relationships
```
efd6: This is not yet defined in ECS. Is there an issue or PR for it?

packages/entityanalytics_ad/data_stream/device/fields/ecs.yml
efd6, line 1: Suggest bumping the ECS version to v9.3.0, then we have `entity.attributes` and `entity.lifecycle` defined.

efd6 (commented):
Can you update the proposed commit message to more accurately reflect the change?

packages/entityanalytics_ad/data_stream/device/fields/ecs.yml
chemamartinez: There is an RFC at the moment: https://github.com/elastic/ecs/pull/2598.

Once the RFC is merged, they are going to add to ECS the missing fields.

packages/entityanalytics_ad/data_stream/device/fields/ecs.yml
chemamartinez: Even though we bump ECS version, `entity.attributes` and `entity.lifecycle` are defined right now as an object without any subfield. This is not the desired final form of these fields, where fields inside should be defined as well.

efd6: This in conjunction with https://github.com/elastic/integrations/pull/18337#discussion_r3186854608 makes me reluctant to approve yet.

chemamartinez: I understand and agree it is not the ideal scenario. Same happens with https://github.com/elastic/integrations/pull/18525 and https://github.com/elastic/integrations/pull/18621.

The RFC has been merged already and the next step is to push the decided entity schema into ECS.

efd6: The RFC has been merged, and the fields have been incorporated into ECS v9.4.0 at alpha, but IMO that's essentially non-existence from the perspective of a user who is expecting a solid contract has none. I'm still reluctant to approve this yet.

packages/entityanalytics_ad/docs/README.md
```diff
@@ -214,6 +216,14 @@ This is the `User` dataset.
 | user.account.status.password_expired |  | boolean |
 | user.account.status.recovery |  | boolean |
 | user.account.status.suspended |  | boolean |
+| user.entity.attributes.mfa_enabled |  | boolean |
+| user.entity.lifecycle.last_activity |  | date |
+| user.entity.relationships.administers.host.id |  | keyword |
+| user.entity.relationships.administers.host.name |  | keyword |
+| user.entity.relationships.administers.user.domain |  | keyword |
```
alexreal1314: @chemamartinez user.domain is not picked up by entity store v2 extraction for the creation of raw_identifiers.

@romulets @uri-weisman correct me if im wrong please. if thats the case what is the purposed solution?

chemamartinez: I think it should be still be populated by the integration, as it is information for the administer that comes from the source and it is defined in ECS.

Also, if it is picked in the future by entity store it is already available.

packages/entityanalytics_ad/data_stream/device/fields/ecs.yml
chemamartinez: Main entity fields that have been added at have been marked as beta (https://www.elastic.co/docs/reference/ecs/ecs-entity).

I don't know why fields inside https://www.elastic.co/docs/reference/ecs/ecs-entity_reference have been marked as alpha instead, they have been discussed, approved and I have a strong opinion they won't be modified as they quite make sense and are already being used in entity workflows. Most probably scenario is that they are extended in the future.

However, if you don't feel confortable about approving it, we can look into the case further and evaluate the available options.

efd6 (approved):

alexreal1314:
@chemamartinez do you have an env with some real data ingested by the updated pipeline?

Based on current changes i see some issue with current approach and im wondering if we could bypass it, the issue:
The raw_identifiers contain values like `ALEJANDR-PC210` (bare CN) and `CN=ALEJANDR-PC210,OU=Computers,DC=acmecrm,DC=com` (full DN). Neither of these values will ever appear as a host.name in any source log that the entity store ingests — so no host entity with those values will ever be created. The entity store only creates host entities from documents where host.name is the FQDN like alejandr-pc210.acmecrm.com.

The EUID the maintainer would try to construct — host:ALEJANDR-PC210 or host:CN=ALEJANDR-PC210,... — simply doesn't exist in the entity index.

The questions is whether the pipeline could write the FQDN (dns_host_name) as host.name in the administers field, not the bare CN.

If that is not possible the maintainer will have to do the heavy lifting by also querying the logs-* and searching for
`activedirectory.device.distinguished_name = "CN=ALEJANDR-PC210,OU=Computers,DC=acmecrm,DC=com"`
read the entity.id and generate the EUID.

cc @uri-weisman @romulets @niros1 @seanrathier

chemamartinez:
@alexreal1314 unfortunately we don't have any available test env at the moment.

Regarding your concerns, I think it is a legitimate gap.

The root fields for hosts already follow a clear three-way split: `host.name` = lowercased FQDN (from `dns_host_name`), `host.hostname` = original-case CN (from `activedirectory.device.cn`), `host.domain` = domain part extracted from the FQDN. Therefore, the relationships fields should follow the same convention for a good correlation, meaning `host.name` in administers should also be the lowercased FQDN, not the bare CN.

The fix for the administers field is straightforward, we can synthesize the FQDN as `cn.toLowerCase() + '.' + domain`.

We may have a deeper problem with `supervises` field. For user-to-user relationships from directReports the CN in the DN is the display name (e.g., CN=Jane Smith,...), not the `sAMAccountName` (jsmith) which is what the entity store uses for `user.name`. There's no way to derive sAMAccountName from a DN alone. So for supervises, the `user.name` in the relationship may not match the entity store identity.

I'd like to hear @uri-weisman's opinion on this.

uri-weisman:
Hey @alexreal1314 @chemamartinez 
Regarding `administers`, that's a good suggestion, let's go for it.
As for `supervises`, is there something we can do in the beats repo to populate the `sAMAccountName ` instead, or it will be expensive?

chemamartinez:
@uri-weisman It should be feasible in beats without extra LDAP queries.

Since the entity analytics input already does a full user enumeration before emitting events, all DN→sAMAccountName mappings should be available in memory at emit time.

The change would be: build a `distinguishedName: sAMAccountName` lookup map during the fetch phase, then resolve each DN in directReports through that map before emitting, the only cost is the in-memory map. The one thing to confirm is whether the current code streams events during pagination or collects all users first; if it streams, a two-pass approach would be needed. In either case, unresolvable DNs (out-of-scope OUs, cross-domain refs) would fall back to the raw DN.

However, this should be treated in a separate issue. I'll explore this issue deeper and open an issue for the Entity Analytics input.

PR #18366, opened 2026-04-13
[aws_sqs] enable agentless deployment for aws sqs metrics

jakubgalecki0:
## Proposed commit message

Enable agentless deployment for aws sqs metrics

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/aws/manifest.yml
zmoog, line 588: IIRC, this metricset doesn't support load balancing—so if you run two instances with the same config they'll collect the same data.

Does agentless handle this scenario?

jakubgalecki0: As far as I understand in agentless mode we are responsible for running the agent but other than that it shouldn't differ much from standard agent based approach.

cc: @kpollich

ishleenk17: @zmoog: Are we talking about the scenario of multiple agents linked to 1 agent policy running the same config, polling the same CloudWatch namespace in the same region at the same time ? 
Curioud how is duplication handled in case of agent based mode ?

One thing to note for agentless: It runs as 1 policy per pod. So 1 agent linked to 1 policy per pod. So if we have another instance of agent it would be part of a different policy and pod.

zmoog: > @zmoog: Are we talking about the scenario of multiple agents linked to 1 agent policy running the same config, polling the same CloudWatch namespace in the same region at the same time?

Yes, this is the case: 1 agent policy assigned to 2+ agents.

> Curioud how is duplication handled in case of agent based mode?

AFAIK, it's not handled. I think there aren't many integrations that can be assigned to multiple agents to scale out (in the cloud space, I know integrations based on aws-s3/sqs or azure-eventhub inputs can).

> One thing to note for agentless: It runs as 1 policy per pod. So 1 agent linked to 1 policy per pod. So if we have another instance of agent it would be part of a different policy and pod.

Got it, thanks! Looking forward to learn more about agentless.

ishleenk17 (approved):
Looks good!

packages/aws/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "6.6.0"
+  changes:
+    - description: Enable agentless deployment for AWS SQS metrics.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/18365
```
kcreddy:
```suggestion
      link: https://github.com/elastic/integrations/pull/18366
```

kcreddy (approved):
LGTM after this minor change. 
codeowner approval for `aws` package.

zmoog (approved):
LGTM.

Nit: We seem to have two different patterns emerging for agentless docs. Some use a dedicated 'Agentless Enabled Integration' section, and others bundle it under 'How do I deploy this?'. We should probably standardize this at some point to improve discoverability.

ishleenk17:
> LGTM.
> 
> Non-blocking: Should we add the 'Agentless Enabled Integration' section to the docs for consistency with our other agentless integrations?

https://github.com/elastic/integrations/pull/18366/changes#diff-f343caa5586e738446eae34c3f3b12c7a7a5142e0c770d5230afefbe03f51e79R37

Something other than this ?

zmoog:
> > LGTM.
> > Non-blocking: Should we add the 'Agentless Enabled Integration' section to the docs for consistency with our other agentless integrations?
> 
> https://github.com/elastic/integrations/pull/18366/changes#diff-f343caa5586e738446eae34c3f3b12c7a7a5142e0c770d5230afefbe03f51e79R37
> 
> Something other than this ?

Sorry, I updated the comment but you've been faster: we seem to have two different patterns emerging for agentless docs. Some integrations use a single 'Agentless Enabled Integration' section, and others bundle it under 'How do I deploy this?' with agent and agentless sub-sections. We should probably standardize this at some point to improve discoverability.

Not required in this PR, a follow-up one is just fine, it this makes sense.

PR #18367, opened 2026-04-13
[integration/kubernetes] add ability to create single input

### LINKED ISSUE #17250: [kubernetes/container_logs]: Allow creating single input to monitor all log files
### Integration Name

Kubernetes [kubernetes]

### Dataset Name

kubernetes.container_logs

### Integration Version

1.84.1

### Agent Version

9.2.1

Darwin 25.2.0

### User Goal

Add an opt-in config option to create one filestream input to monitor the entire cluster. The `paths` would be set to `/var/log/pods/*/*/*.log` by default.

This feature will be opt-in and will NOT be the default.

### Existing Features

The kubernetes.container_logs data stream creates one filestream input per container. For 100s of pod and potentially 1000s of containers, this becomes an issue because each input comes with a baseline memory usage. 
This can lead to OOMs and restarts in production environment.

### What did you see?

I manually installed a policy with one filestream input monitoring `/var/log/containers/*.log` as a whole and tested it against the latest kubernetes integration. That showed remarkable reduction in overall memory usage.

Attaching images:

1. Latest kubernetes integration. 
Image

2. Custom policy with one input and rest of the config same as previous one.
Image

### Anything else?

If any customer wants to switch to this feature, they'd face data duplication as the input id will change. We'll be documenting this caveat.

VihasMakwana:
## Proposed commit message

Add ability to create single input to monitor all the log files. Right now, a new input is created for every file discovered. This has some with a memory cost. See https://github.com/elastic/integrations/issues/17250#issuecomment-4235756466

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/17250

## Screenshots

See https://github.com/elastic/integrations/issues/17250


packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -104,6 +104,26 @@ streams:
...
+      - name: singleInput
+        type: bool
+        title: Single input
+        multi: false
+        required: false
+        show_user: false
+        default: false
+        description: >-
+          When selected, Elastic Agent will create a single input to collect all container logs on the node. 
+          This is useful when you want to avoid creating a separate input for each container.
```
orestisfl, line 117: Do we want to export an internal implementation detail like this? Is there a negative to this approach?

VihasMakwana: we can rephrase this a little bit and highlight the memory savings. I'll do that.

VihasMakwana: > Is there a negative to this approach?

Do you mean  negative to "existing" approach i.e. one input per container? Or the approach added by this PR.

packages/kubernetes/data_stream/container_logs/agent/stream/stream.yml.hbs
```diff
@@ -50,12 +62,38 @@ processors:
...
+      - elastic.co/preserve_original_event
+    host: ${NODE_NAME}
+    default_indexers.enabled: false
+    default_matchers.enabled: false
+    indexers:
+      - pod_uid:
+    matchers:
+      - logs_path:
+        logs_path: "/var/log/pods/"
+        resource_type: "pod"
```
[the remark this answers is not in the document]
VihasMakwana: Fix it for me

packages/kubernetes/data_stream/container_logs/manifest.yml
belimawr: We should mention this will cause the existing files to be re-ingested once. That one of the biggest challenges of doing this type of migration.

However, I've been thinking about it and I believe we solve that by allowing take over without specifying IDs: https://github.com/elastic/beats/issues/51307

packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -104,6 +104,26 @@ streams:
...
+      - name: pathsSingleInput
+        type: text
+        title: Kubernetes container log path for single input
+        default: 
+          - /var/log/pods/*/*/*.log
+        multi: true
+        required: false
+        show_user: false
+        description: >-
+          When "Single input" is selected, use this configuration to set the path for the input.
```
belimawr, line 127: I wish we could use the same `paths` variable here... I wish there was a better solution to it.

Could you add a note in the `paths` variable explaining it will be ignored when this is enabled?

packages/kubernetes/data_stream/container_logs/agent/stream/stream.yml.hbs
```diff
@@ -50,12 +62,38 @@ processors:
...
+      - elastic.co/namespace
+      - elastic.co/preserve_original_event
+    host: ${NODE_NAME}
+    default_indexers.enabled: false
+    default_matchers.enabled: false
+    indexers:
+      - pod_uid:
+    matchers:
+      - logs_path:
+          logs_path: "/var/log/pods/"
```
belimawr: I've been thinking about it, maybe it is better to have this configurable as well? It has to match what is added in `paths`.

packages/kubernetes/data_stream/container_logs/manifest.yml
```diff
@@ -104,6 +104,26 @@ streams:
...
+        show_user: false
+        default: false
+        description: >-
+          When selected, Elastic Agent will create a single input to collect all container logs on the node. 
+          This is useful when you want to avoid creating a separate input for each container.
+      - name: pathsSingleInput
+        type: text
+        title: Kubernetes container log path for single input
+        default: 
+          - /var/log/pods/*/*/*.log
```
belimawr, line 122: I've been doing some testing with the `add_kubernetes_metadata` for a support case and if we use `/var/log/containers/*.log` and use `resource_type: container` in the matcher, we can get the same metadata as in the 'multiple inputs' version.

I know it has the drawback of not exposing rotated files and gzipped files to Filebeat... Do we know how often container metadata is required for customers?

packages/kubernetes/data_stream/container_logs/agent/stream/stream.yml.hbs
VihasMakwana: Yeah, we should do that.

packages/kubernetes/data_stream/container_logs/manifest.yml
VihasMakwana: > Do we know how often container metadata is required for customers?

Not really. 

> I've been doing some testing with the add_kubernetes_metadata for a support case and if we use /var/log/containers/*.log and use resource_type: container in the matcher, we can get the same metadata as in the 'multiple inputs' version.

I think this can be a good enhancement as a follow-up.

packages/kubernetes/data_stream/container_logs/manifest.yml
VihasMakwana: Sure. I'll add a note

packages/kubernetes/data_stream/container_logs/agent/stream/stream.yml.hbs
VihasMakwana: Done.

packages/kubernetes/data_stream/container_logs/manifest.yml
VihasMakwana: Done.

packages/kubernetes/data_stream/container_logs/manifest.yml
belimawr: My personal opinion here: I rather delay this a bit than change it afterwards, because it will be kinda of a breaking change or give even more knobs to our users, which ends up becoming confusing and dangerous (there is so much we can do to have a set of mutually exclusive options on an integration).

My suggestion: seek some product input on the trade offs here.

PR #18379, opened 2026-04-14
Add integration for tailsamplingprocessor

gregkalapos:
Adding a content pack to visualize metrics from the tailbasedprocessor OTel Collector component. 

Metrics are based on the [upstream docs here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/documentation.md). 

## Checklist

## Screenshots

sample-screenshot


packages/otel_collector_tailsamplingprocessor/manifest.yml
```diff
@@ -0,0 +1,32 @@
...
+  - monitoring
+  - opentelemetry
+conditions:
+  kibana:
+    version: "9.3.0"
+  elastic:
+    subscription: "basic"
+discovery:
+  datasets:
+    - name: collectortelemetry.otel
```
gregkalapos, line 19: I'd like some input on this: Problem is that this dataset is present as soon as internal telemetry in an OTel collector config is turned on, but there is no guarantee that the tailsamplingprocessor is present.

This means, with this potentially we'd install this integration even if the data we visualize here is not present.

Options I see:
- either proceed as is (and then potentially the integration is installed even if there is no data to visualize)
- we try to find something that identifies data from tailsamping processor (so far, I don't see anything obvious)
- drop this part and there is no auto discovery (this'd be sad :( )

gregkalapos:
Closing this over https://github.com/elastic/integrations/pull/18527

PR #18390, opened 2026-04-14
[system_otel] Remove host.ip and host.mac

### LINKED ISSUE #18334: [system_otel]: Remove host.ip and host.mac from otel system dashboards
Current system_otel content pack dashboard display `host.ip` and `host.mac` values. With the push to disable those in order to reduce costs, the values should also be removed from the visualization tools. If not removed, this would break the tables displaying these values and negatively impact user experience.

Below is a screenshot from such a table that is impacted by this change:

Image

jmmcorreia:
## Proposed commit message

Removed `host.ip` and `host.mac` from system_otel content pack as those values are set to no longer be collected by default due to the high volume of data they might generated in k8s environments.

Check related issues:
- https://github.com/elastic/elastic-agent/issues/13575
- https://github.com/elastic/opentelemetry-dev/issues/1311

Please explain:

Removed `host.ip` and `host.mac` from ESQL queries. 

These values were not made optional due to the fact that if they were never ingested into ES, the mappings won't be know, causing the whole query to fail. 

Point `elastic-package` to the right ES instance (local or ECS) 

`elastic-package build`
`elastic-package install`

helm-install otel-operator with one of the value files from PR: https://github.com/elastic/elastic-agent/pull/13566

Wait for content pack to get enabled.

Point 

## Related issues

- Closes #18334 

## Screenshots

Screenshot 2026-04-14 at 15 25 40

Screenshot 2026-04-14 at 15 25 46

### DIFF
--- a/packages/system_otel/changelog.yml
+++ b/packages/system_otel/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "0.2.7"
+  changes:
+    - description: Remove `host.ip` and `host.mac` from dashboard
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/
 - version: "0.2.6"
   changes:
     - description: Use `host.id` so cloud metadata works with default attributes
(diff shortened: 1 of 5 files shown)


osullivandonal (approved):
LGTM 🚀

rogercoll (approved):

PR #18394, opened 2026-04-14
[jamf_protect, lyve_cloud] Add region variable in aws-s3 input

### LINKED ISSUE #18279: Multiple integrations fail on Agent 8.19.12+ when using non-AWS S3-compatible buckets — missing `region` field
## Summary

Elastic Agent 8.19.12 added a validation in the Filebeat aws-s3 input (elastic/beats#48534) that requires the `region` config field when `non_aws_bucket_name` is set. Several integrations that support non-AWS S3-compatible storage only render `default_region` in their templates. These are different fields: `region` is the S3 input-level region, `default_region` is the AWS SDK fallback region. The result is a startup failure:

```
region must be configured when using non_aws_bucket_name
```

The same bug was reported and fixed for `cloudflare_logpush` in #18278 (closes #17790). This issue tracks the remaining affected integrations.

## Affected integrations

All are owned by @elastic/security-service-integrations.

### jamf_protect (5 data streams)

`alerts`, `telemetry`, `telemetry_legacy`, `web_threat_events`, `web_traffic_events`

Templates render `default_region` but not `region`. The data streams support `non_aws_bucket_name` via the `jamf_protect_bucket_name` / `global_bucket_name` variables.

### tencent_cloud (4 data streams)

`audit`, `clb`, `cos`, `scf`

Same problem — templates render `default_region` but not `region`. The data streams support `non_aws_bucket_name` directly.

### lyve_cloud (1 data stream)

`audit`

Different variant of the bug: the template has a variable called `region` but maps it to the wrong config key:

```hbs
{{#if region}}
default_region: {{region}}
{{/if}}
```

This should be `region: {{region}}`. The `default_region` key does not satisfy the Filebeat validation.

## Root cause

elastic/beats#48534 (merged 2026-02-10, shipped in 8.19.12) added this validation in `x-pack/filebeat/input/awss3/config.go`:

```go
if c.NonAWSBucketName != "" && c.RegionName == "" {
    return errors.New("region must be configured when using non_aws_bucket_name")
}
```

`RegionName` is the `region` config key. `DefaultRegion` (the `default_region` config key) is a separate field in the AWS common config used for SDK initialisation and does not satisfy this check.

## Fix

Each integration needs a `region` variable added to its manifest and the aws-s3 stream templates updated to render `region: {{region}}`. See #18278 for the pattern used in the `cloudflare_logpush` fix, which matches the approach already used by the `tanium` and `netskope` integrations.

## References

- elastic/beats#48534 — Filebeat PR that introduced the validation
- #18278 — Fix for `cloudflare_logpush`
- #17790 — Original report for `cloudflare_logpush`

brijesh-elastic:
## Proposed commit message
```
jamf_protect, lyve_cloud: add region variable to aws-s3 input

Elastic Agent 8.19.12 introduced validation in the Filebeat aws-s3
input that requires the region configuration field whenever
non_aws_bucket_name is set (elastic/beats#48534).

Currently, the jamf_protect and lyve_cloud integration only renders "default_region",
which is a separate field used for AWS SDK initialization; therefore,
I am adding the "region" variable to ensure compatibility.
```

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/18279

### DIFF
--- a/packages/jamf_protect/manifest.yml
+++ b/packages/jamf_protect/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.0.3
 name: jamf_protect
 title: Jamf Protect
-version: "3.3.0"
+version: "3.3.1"
 description: Receives events from Jamf Protect with Elastic Agent.
 type: integration
 categories:
@@ -158,6 +158,14 @@ policy_templates:
             show_user: false
             default: ""
             description: URL of the entry point for an AWS web service.
+          - name: region
+            type: text
+            title: Region
+            multi: false
+            required: false
+            show_user: false
+            default: ""
+            description: The name of the region of the end point. Required when using a non-AWS S3-compatible bucket.
           - name: default_region
             type: text
             title: Default AWS Region
--- a/packages/lyve_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
+++ b/packages/lyve_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
@@ -5,7 +5,10 @@ non_aws_bucket_name: {{non_aws_bucket_name}}
 {{#if region}}
 region: {{region}}
 {{/if}}
-
+{{#if region}}
+default_region: {{region}}
+region: {{region}}
+{{/if}}
 {{#if credential_profile_name}}
 credential_profile_name: {{credential_profile_name}}
 {{/if}}
--- a/packages/lyve_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
+++ b/packages/lyve_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
@@ -3,9 +3,6 @@ non_aws_bucket_name: {{non_aws_bucket_name}}
 {{/if}}
 
 {{#if region}}
-region: {{region}}
-{{/if}}
-{{#if region}}
 default_region: {{region}}
 region: {{region}}
 {{/if}}
--- a/packages/lyve_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
+++ b/packages/lyve_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
@@ -3,7 +3,7 @@ non_aws_bucket_name: {{non_aws_bucket_name}}
 {{/if}}
 
 {{#if region}}
-default_region: {{region}}
+region: {{region}}
 {{/if}}
 
 {{#if credential_profile_name}}
--- a/packages/tencent_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
+++ b/packages/tencent_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
@@ -78,6 +78,9 @@ api_timeout: {{api_timeout}}
 {{#if endpoint}}
 endpoint: {{endpoint}}
 {{/if}}
+{{#if region}}
+region: {{region}}
+{{/if}}
 {{#if default_region}}
 default_region: {{default_region}}
 {{/if}}
--- a/packages/tencent_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
+++ b/packages/tencent_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
@@ -78,9 +78,6 @@ api_timeout: {{api_timeout}}
 {{#if endpoint}}
 endpoint: {{endpoint}}
 {{/if}}
-{{#if region}}
-region: {{region}}
-{{/if}}
 {{#if default_region}}
 default_region: {{default_region}}
 {{/if}}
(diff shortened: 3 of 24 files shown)


packages/lyve_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
```diff
@@ -3,7 +3,7 @@ non_aws_bucket_name: {{non_aws_bucket_name}}
 {{/if}}
 
 {{#if region}}
-default_region: {{region}}
+region: {{region}}
 {{/if}}
```
kcreddy: Add `default_region` as well. It is used during client initialization.

packages/jamf_protect/manifest.yml
```diff
@@ -158,6 +158,14 @@ policy_templates:
...
             default: ""
             description: URL of the entry point for an AWS web service.
+          - name: region
+            type: text
+            title: Region
+            multi: false
+            required: false
+            show_user: false
+            default: ""
+            description: The name of the region of the end point. Required when using a non-AWS S3-compatible bucket.
```
kcreddy, line 168: Lets add a note into the README for all these integrations that it is required.

packages/tencent_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
kcreddy, line 31: There are following concerns from Claude with regards to `tencent_cloud`:

### Is the `tencent_cloud` change actually exercised?

I'm not sure the new 8.19.12 validation actually fires for `tencent_cloud` today. The stream templates reference `{{#if non_aws_bucket_name}}`, but `non_aws_bucket_name` isn't declared as a variable anywhere — not in the four `data_stream/*/manifest.yml` files and not at the input level in `packages/tencent_cloud/manifest.yml`. `bucket_arn` is `required: true` in each data stream manifest, so the template always takes the `bucket_arn` path. The Filebeat validation is:

```go
if c.NonAWSBucketName != "" && c.RegionName == "" {
    return errors.New("region must be configured when using non_aws_bucket_name")
}
```

If `non_aws_bucket_name` is never rendered into the config, the check never triggers, and adding `region` is harmless but unnecessary for `tencent_cloud`. Could you confirm you've reproduced the failure against the `tencent_cloud` integration on 8.19.12+? If it doesn't repro, I'd either (a) drop `tencent_cloud` from this PR, or (b) scope it to the underlying template/manifest gap (i.e. expose `non_aws_bucket_name` properly first).

### `region` description is misleading for `tencent_cloud`

All three integrations copy the same wording:

> The name of the region of the end point. Required when using a non-AWS S3-compatible bucket.

For `tencent_cloud`, the integration's only supported storage is Tencent Cloud COS, which is itself non-AWS. "When using a non-AWS S3-compatible bucket" reads oddly there — it's always the case. Something like "The region the Tencent Cloud COS bucket is located in" would be clearer. (This ties into the question above — if the tencent_cloud change stays in the PR.)

packages/lyve_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
brijesh-elastic: The `lyve_cloud` package only supports Lyve Cloud's Managed S3; it doesn't provide support for AWS S3 buckets. Do we still need to provide the `default_region` parameter?

packages/tencent_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
brijesh-elastic: It appears the `aws-s3` input for the `tencent_cloud` package is not functioning correctly. Based on the manifest and README, the package fetches data from the Tencent Cloud Object Storage (COS). So, the `non_aws_bucket_name` should be a required field; however, this variable is currently missing from the manifest.

I will address these concerns in another PR. For now, I am removing the `tencent_cloud` changes from this PR.

packages/lyve_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
kcreddy: Yes for `<8.19.12` cases, the `default_region` is used inside `InitializeAWSConfig` so that it handles initialisation of S3 client. In order not to break this since the integration is still on `^8.13.0`, we should still use the existing `default_region`.


```suggestion
{{#if region}}
default_region: {{region}}
region: {{region}}
{{/if}}
```

packages/jamf_protect/manifest.yml
brijesh-elastic: Done, https://github.com/elastic/integrations/pull/18394/changes/54c87056edd445cdd63915c7b0b762f45ac38653.

packages/lyve_cloud/data_stream/audit/agent/stream/aws-s3.yml.hbs
kcreddy: @brijesh-elastic, this needs addressing. Can you take a look?

kcreddy (approved):

PR #18395, opened 2026-04-14
[awsfirehose] Fix VPC flow log detection in firehose routing pipeline

### LINKED ISSUE #18110: bad if condition - firehose pipeline -> vpc flow log
https://github.com/elastic/integrations/blame/68b822662604844675c944e9cb68ea1edadf4ccd/packages/awsfirehose/data_stream/logs/elasticsearch/ingest_pipeline/default.yml#L36

this condition is wrong:
else if (message_lower.contains('"eni-') && message_lower.contains('"vpc-') && message_lower.contains('accept') || message_lower.contains('reject'))

it should be
else if (
  message_lower.contains('"eni-') &&
  message_lower.contains('"vpc-') &&
  (message_lower.contains('accept') || message_lower.contains('reject'))
) {
  ctx.event.dataset = 'aws.vpcflow';
}

i have created a ticket with your support.
i have seen this issue in action (latest elastic version, latest integration version).
also, this simple fix is suggest by Copilot on Github..

please fix this soon.

zmoog:
## Summary

- Fixed broken VPC flow log detection condition in the firehose ingest pipeline that had two bugs: operator precedence (`&&` vs `||`) and incorrect string matching (`"eni-` with leading quote never matched inside JSON-wrapped messages)
- Replaced the fragile condition with robust detection covering all VPC flow log variants (v2-v10):
  - `tgw-` + `transitgateway` — transit gateway flow logs (default format includes `resource-type: TransitGateway`)
  - `eni-` — standard VPC flow logs, requires a secondary marker (`accept`, `reject`, `nodata`, `skipdata`)
  - `nat-` — v9 regional NAT gateway flow logs (where `interface-id` is `-`), requires a secondary marker
- Expanded VPC flow log test suite from 1 case to 16, covering all versions (v2-v10) and all field formats supported by the downstream vpcflow pipeline (14, 17, 18, 21, 29, 30, 36, 39, 40 tokens), using examples from the [official AWS VPC Flow Logs documentation](https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-records-examples.html)
- Added a regression test to the firewall test suite: a firewall log with `"action":"reject"` that must route to `aws.firewall_logs`, not `aws.vpcflow` — this is the exact scenario the original operator precedence bug caused
- Added a ["How log routing works"](https://github.com/elastic/integrations/blob/fix/firehose-vpcflow-operator-precedence/packages/awsfirehose/docs/README.md#how-log-routing-works) section to the docs explaining the detection method and format requirements for each log type
- Bumped package version to 1.9.1

### Known limitation

VPC flow logs using custom formats that exclude both the `action` and `log-status` fields (e.g., the [6-field zonal NAT gateway format](https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-records-examples.html#flow-log-example-nat)) cannot be detected through content matching alone. The default format always includes `action` and `log-status`, so this only affects deliberately minimal custom configurations. Users with such formats should use the `es_datastream_name` delivery stream parameter for routing instead.

## Test plan

- [x] `elastic-package test pipeline -v -d logs` passes all 22 tests (16 vpcflow + 2 firewall + 8 other log types unchanged)
- [x] No regressions in WAF, CloudTrail, Firewall, Route53, API Gateway, S3, CloudFront, or ELB detection
- [x] Token counts verified against downstream `aws.vpcflow` pipeline dissect patterns

Closes #18110


tetianakravchenko (commented):
please add changelog record and bump version of the package

packages/awsfirehose/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -32,8 +32,8 @@ processors:
             ctx.event.dataset = 'aws.cloudtrail';
         }
 
-        // AWS VPC Flow Logs
-        else if (message_lower.contains('"eni-') && message_lower.contains('"vpc-') && message_lower.contains('accept') || message_lower.contains('reject')) {
+        // AWS VPC Flow Logs (standard, transit gateway, and regional NAT gateway)
```
tetianakravchenko: zonal NAT gateway is also covered, isn't it?
can be added some clarification comment, like:
```suggestion
        // AWS VPC Flow Logs:
        //   eni-  → standard ENI-based records AND zonal NAT gateway (interface-id = eni-xxxxx)
        //   nat-  → regional NAT gateway v9+ (interface-id = "-", resource-id = nat-xxxxx)
        //   tgw-  → transit gateway (no action field; log-status may be OK with no ACCEPT/REJECT)
```
?
(if it is correct)

packages/awsfirehose/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -32,8 +32,8 @@ processors:
...
+        else if (message_lower.contains('tgw-') || ((message_lower.contains('eni-') || message_lower.contains('nat-')) && (message_lower.contains('accept') || message_lower.contains('reject') || message_lower.contains('nodata') || message_lower.contains('skipdata')))) {
```
tetianakravchenko: wondering if `tgw-` as a standalone condition has a risk of matching non-vpc logs (that follow after this check in ingest pipeline)? 
in https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-records-examples.html#flow-log-example-tgw is mentioned that `action` should be available

packages/awsfirehose/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
zmoog: Yeah, it seems so. Thanks for checking, taking a closer look at this use case.

packages/awsfirehose/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
zmoog: Good point, checking this as well!

zmoog: Tightened the transit gateway detection rule adding the resource-type.

packages/awsfirehose/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
zmoog: For the `eni-` part, we can only support the standard format (which includes accept/reject). We probably need to note this in the docs, if it's not there yet.

And yes, we should update the comment to reflect what we support now.

zmoog: Updated the comment.

zmoog:
I added a new section in the docs to suggest how to successfully route logs using this integration.

packages/awsfirehose/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
tetianakravchenko: btw, where transit gateway log samples are coming from? I see that test samples 1-8 are from https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-records-examples.html, but the rest? are these real or reconstructed from spec?

zmoog: I used examples from the docs as much as possible, but some test cases have been reconstructed from the spec.

Let me see if I can find more examples, including outside the AWS docs. We can also set up a couple of VPCs with a transit gateway to collect some samples.

packages/awsfirehose/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -14,10 +14,11 @@ processors:
         if (ctx.event == null) {
             ctx.event = [:];
         }
-        
-        // AWS WAF Logs
+
+        // Initialize lower case version of the message to
```
tetianakravchenko: this sentence seems to be not finished, or?

packages/awsfirehose/docs/README.md
```diff
@@ -18,6 +18,33 @@ Here is a list of log types that are supported by this integration:
 | VPC Flow           | Firehose, CloudWatch, S3 |
 | WAF                | Firehose, CloudWatch. S3 |
 
+### How log routing works
+
+This integration uses content-based detection to automatically identify the log type and route
+each record to the appropriate integration dataset. The detection relies on known patterns in the
+log message, such as field names, resource identifiers, or structural characteristics.
+
+The following table describes how each log type is detected and what is required for routing
```
tetianakravchenko: this table can get quickly outdated, not sure if we should ducplicate in readme filtering logic
I think mentioning about the detecting, maybe 1 specific example and leaving a note about auto-detection not working should be good for the doc

tetianakravchenko (commented):
@zmoog build is failing:
```
Error: checking package failed: checking readme files are up-to-date failed: files do not match
```
I believe `elastic-package build` should fix it

packages/awsfirehose/docs/README.md
zmoog: Yeah, after a second look I thought rh same: I'll only leave the recommendations on the formats.

packages/awsfirehose/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
tetianakravchenko: lets start with what we have now and iterate over it 👍

tetianakravchenko (approved):

PR #18414, opened 2026-04-15
[cloud_asset_inventory] Fix duplicated 'Discovery' in data stream descriptions

SAY-5:
Addresses one item from #18242 (text-auditor batch).

`packages/cloud_asset_inventory/data_stream/asset_inventory/manifest.yml` contained `Asset Discovery Discovery` for all three cloud providers:

- AWS — line 14
- Azure — line 151
- GCP — line 310

The duplicated word renders in the Kibana data stream picker and the integrations catalog. Drop the duplicate so each entry reads `Asset Discovery for <provider>`.

- Bumped `cloud_asset_inventory` patch version to `1.5.1` and added a bugfix `changelog.yml` entry as required by the integrations release process.
- Verified the upstream still contains the duplication on `main` before opening this PR.
- Scoped to a single package (per #18242 the auditor flagged many packages — happy to follow up with separate one-package-per-PR fixes if this is the preferred style).

### DIFF
--- a/packages/cloud_asset_inventory/changelog.yml
+++ b/packages/cloud_asset_inventory/changelog.yml
@@ -8,6 +8,11 @@
 # 1.1.x - 9.2.x
 # 1.0.x - 9.1.x
 # 0.1.x - 8.15.x
+- version: "1.5.1"
+  changes:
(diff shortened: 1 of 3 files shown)


SAY-5:
Closing — elastic CLA blocks merge. Apologies.

PR #18419, opened 2026-04-16
[microsoft_intune] Update changelog and version in manifest

janvi-elastic:
## Proposed commit message

Update changelog and version in manifest for Microsoft Intune.


efd6 (approved):
Thanks

PR #18430, opened 2026-04-16
docs: update agent-version-conditions doc

juliaElastic:
Follow up after https://github.com/elastic/integrations/pull/18203 to address comments

### DIFF
--- a/docs/extend/agent-version-conditions.md
+++ b/docs/extend/agent-version-conditions.md
@@ -114,7 +114,7 @@ For most forward upgrades within a major version this is harmless: newer agents 
 - **Downgrades**: the cached policy was rendered for a newer agent and may reference inputs, processors, or built-ins not available in the older binary.
 - **Major-version upgrades** that remove an input option or processor the cached rendering depends on.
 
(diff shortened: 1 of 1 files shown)


docs/extend/agent-version-conditions.md
```diff
@@ -81,14 +97,26 @@ The policy assigned to the earlier version of agents will not include the incomp
...
 
 
+## Upgrade and downgrade considerations [upgrade-downgrade-considerations]
+
+When an agent binary is upgraded or downgraded, the new binary starts with a cached copy of the previously rendered policy. The agent runs this cached policy until the first successful check-in, at which point Fleet re-renders the template against the new version and returns a fresh policy revision.
+
+For most forward upgrades within a major version this is harmless: newer agents accept older rendered configs while they wait for the first check-in. The risky cases are:
+
+- **Downgrades**: the cached policy was rendered for a newer agent and may reference inputs, processors, or built-ins not available in the older binary.
+- **Major-version upgrades** that remove an input option or processor the cached rendering depends on.
```
cmacknz, line 115: There is a 3rd case, during an upgrade where an agent can automatically roll back. Similar to a downgrade or a user initiated rollback except that Fleet doesn't know it is going to happen before hand. In that case, agent might need to make the cached policy specific to an agent version. The agent revision would go backwards in this case but so would the version.

For remotely triggered downgrades or major version upgrades, probably Fleet would have to do a policy change to a compatible version before or even as part of the upgrade in the same action. The latter option would fit with the idea that the agent version is part of the policy.

andrewkroh (approved):

mrodm (approved):

PR #18436, opened 2026-04-16
[Elasticsearch] Fix authentication failure when password contains YAML special characters

### LINKED ISSUE #18434: [Elasticsearch]: Authentication fails when password contains YAML special characters (e.g. double quotes)
### Integration Name

Elasticsearch [elasticsearch]

### Dataset Name

All metrics datasets (`cluster_stats`, `node_stats`, `node`, `index`, `index_recovery`, `index_summary`, `ccr`, `enrich`, `ingest_pipeline`, `ml_job`, `pending_tasks`, `shard`)

### Integration Version

1.20.1

### Agent Version

8.19.1

### Agent Output Type

elasticsearch

### Elasticsearch Version

8.19.1

Linux (x86_64)

### Software/API Version

_N/A_

### Error Message

The `elasticsearch/metrics` input enters a **Degraded** state with HTTP 401 errors when the configured password contains YAML special characters such as double quotes (`"`):

```
Degraded

Error fetching data for metricset elasticsearch.cluster_stats: error determining if
connected Elasticsearch node is master: HTTP error 401 in : 401 Unauthorized
```

The same error pattern appears across all metrics datasets that share the same template structure.

### Event Original

_N/A — this is an authentication/configuration issue, not a pipeline/parsing issue._

### What did you do?

Configured the **Elasticsearch** integration's metrics input (`elasticsearch/metrics`) with a `username` and `password` where the password value contains double-quote characters, for example:

```
password: "mysecret"
```

(i.e. the literal password string includes the `"` characters as part of the value, not as YAML delimiters)

### What did you see?

All metrics datasets immediately enter a **Degraded** state with `401 Unauthorized` responses from Elasticsearch.

Inspecting the compiled agent configuration (`components-actual.yaml`) shows the password is stored **without** the surrounding quote characters:

```yaml
# With a password value of: "mysecret"
password:
  kind:
    stringvalue: mysecret   # <-- double quotes stripped by YAML parser
```

This happens because the Handlebars template renders the password directly as an unquoted YAML scalar:

```yaml
# stream.yml.hbs (all metrics datasets)
{{#if password}}
password: {{password}}
{{/if}}
```

When `{{password}}` expands to `"mysecret"`, the rendered YAML line becomes:

```yaml
password: "mysecret"
```

YAML interprets the double quotes as **string delimiters**, so the parsed value is `mysecret` — the quotes are silently dropped. The agent then authenticates with the wrong password.

Verification with `curl` confirms the distinction:
```bash
# Fails (quotes stripped by shell — same as what the agent experiences):
curl -k -u monitoring_user:"mysecret" -X GET 'https://localhost:9200/_cluster/stats?pretty'
# {"type":"security_exception","reason":"unable to authenticate user..."}

# Succeeds (quotes preserved via single-quoting):
curl -k -u 'monitoring_user:"mysecret"' -X GET 'https://localhost:9200/_cluster/stats?pretty'
```

Logging in via Kibana UI with the same password works correctly because the UI handles the value as a raw string.

Authentication should succeed regardless of whether the password contains YAML special characters. The compiled agent config should preserve the password value exactly as entered, including any double quotes, backslashes, or other special characters.

### Anything else?

**Root cause:** The `.hbs` stream templates for every Elasticsearch metrics dataset use `password: {{password}}` without quoting, so the YAML parser interprets characters like `"` as structural syntax rather than literal content.

**Fix:** Replace `{{password}}` with `{{escape_string password}}` in all 12 metrics dataset stream templates. The `escape_string` Handlebars helper (registered in Fleet's template engine) wraps the value in YAML single quotes and escapes any interior single quotes by doubling them, safely preserving all special characters.

For example, a password of `"mysecret"` would render as:

```yaml
password: '"mysecret"'
```

which YAML correctly parses back to the string `"mysecret"` (with the double quotes intact).

The same pattern is already used correctly in other integrations (e.g. `ibmmq`, `apache_tomcat`, `websphere_application_server`) and with other sensitive fields in `cloudflare` (`escape_string auth_key`).

**Workaround:** Use a password that does not contain YAML special characters, or authenticate via API key instead of username/password.

maramos-elastic:
## Summary

Fixes #18434

Passwords containing YAML special characters (e.g. double quotes `"`) cause `401 Unauthorized` errors in all Elasticsearch metrics datasets because the compiled agent configuration strips those characters during YAML parsing.

### Root cause

All 12 Elasticsearch metrics dataset stream templates used the bare Handlebars expression:

```yaml
{{#if password}}
password: {{password}}
{{/if}}
```

When a password such as `"mysecret"` is rendered, the template produces:

```yaml
password: "mysecret"
```

YAML treats the double quotes as **string delimiters**, so the parsed value is `mysecret` — the characters are silently dropped. The Beat then authenticates with a truncated password and receives a `401 Unauthorized` response.

### Fix

Replace `{{password}}` with `{{escape_string password}}` in all 12 metrics dataset stream templates:

- `data_stream/cluster_stats/agent/stream/stream.yml.hbs`
- `data_stream/node_stats/agent/stream/stream.yml.hbs`
- `data_stream/node/agent/stream/stream.yml.hbs`
- `data_stream/index/agent/stream/stream.yml.hbs`
- `data_stream/index_recovery/agent/stream/stream.yml.hbs`
- `data_stream/index_summary/agent/stream/stream.yml.hbs`
- `data_stream/ccr/agent/stream/stream.yml.hbs`
- `data_stream/enrich/agent/stream/stream.yml.hbs`
- `data_stream/ingest_pipeline/agent/stream/stream.yml.hbs`
- `data_stream/ml_job/agent/stream/stream.yml.hbs`
- `data_stream/pending_tasks/agent/stream/stream.yml.hbs`
- `data_stream/shard/agent/stream/stream.yml.hbs`

The `escape_string` helper (registered in Fleet's Handlebars engine) wraps the value in YAML single quotes and escapes any interior single quotes by doubling them — safely preserving all special characters including `"`, `\`, `:`, `#`, etc.

**Before** (password `"mysecret"` → stored as `mysecret`):
```yaml
password: "mysecret"
```

**After** (password `"mysecret"` → stored correctly as `"mysecret"`):
```yaml
password: '"mysecret"'
```

This pattern is already used correctly by other integrations for similar fields (e.g. `ibmmq`, `apache_tomcat`, `websphere_application_server`, `cloudflare`).

### Testing

Configure an Elasticsearch integration policy with a password containing `"` characters and confirm the `components-actual.yaml` preserves the full password value, and that metrics collection succeeds without `401` errors.

## Checklist

- [x] I have reviewed the [guidelines for contributing](https://github.com/elastic/integrations/blob/main/CONTRIBUTING.md) to this repository.
- [x] I have verified that my code works as expected.
- [x] I have incremented the package's version per the [versioning guidelines](https://github.com/elastic/integrations/blob/main/docs/developer_workflow_design_build_test_integration.md#package-versioning) and updated the [changelog](packages/elasticsearch/changelog.yml).
- [x] I have updated the package's [changelog](packages/elasticsearch/changelog.yml).

### DIFF
--- a/packages/elasticsearch/data_stream/ccr/agent/stream/stream.yml.hbs
+++ b/packages/elasticsearch/data_stream/ccr/agent/stream/stream.yml.hbs
@@ -8,7 +8,7 @@ scope: {{scope}}
 username: {{username}}
 {{/if}}
 {{#if password}}
-password: {{password}}
+password: {{escape_string password}}
 {{/if}}
 {{#if api_key}}
 api_key: {{api_key}}
--- a/packages/elasticsearch/data_stream/ccr/agent/stream/stream.yml.hbs
+++ b/packages/elasticsearch/data_stream/ccr/agent/stream/stream.yml.hbs
@@ -5,7 +5,7 @@ hosts:
 {{/each}}
 scope: {{scope}}
 {{#if username}}
-username: {{username}}
+username: {{escape_string username}}
 {{/if}}
 {{#if password}}
 password: {{escape_string password}}
--- a/packages/elasticsearch/elasticsearch/transform/index_pivot/transform.yml
+++ b/packages/elasticsearch/elasticsearch/transform/index_pivot/transform.yml
@@ -119,7 +119,7 @@ pivot:
         script: "Math.max(0, params.end-params.start)"
 dest:
   index: "monitoring-indices"
-  pipeline: "1.20.1-monitoring_indices"
+  pipeline: "1.20.2-monitoring_indices"
 description: This transform runs every 10 minutes to compute extra metrics for the Elasticsearch indices.
 frequency: 10m
 settings:
--- a/packages/elasticsearch/elasticsearch/transform/index_pivot/transform.yml
+++ b/packages/elasticsearch/elasticsearch/transform/index_pivot/transform.yml
@@ -119,7 +119,7 @@ pivot:
         script: "Math.max(0, params.end-params.start)"
 dest:
   index: "monitoring-indices"
-  pipeline: "1.20.2-monitoring_indices"
+  pipeline: '{{ ingestPipelineName "monitoring_indices" }}'
 description: This transform runs every 10 minutes to compute extra metrics for the Elasticsearch indices.
 frequency: 10m
 settings:
--- a/packages/elasticsearch/changelog.yml
+++ b/packages/elasticsearch/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.20.2"
+  changes:
+    - description: Fix authentication failure when password contains YAML special characters (e.g. double quotes) by using the escape_string Handlebars helper.
+      type: bugfix
+      link: https://github.com/elastic/integrations/issues/18434
 - version: "1.20.1"
   changes:
     - description: Make Elasticsearch server log pipeline tolerate missing elasticsearch fields
--- a/packages/elasticsearch/changelog.yml
+++ b/packages/elasticsearch/changelog.yml
@@ -1,7 +1,7 @@
 # newer versions go on top
 - version: "1.20.2"
   changes:
-    - description: Fix authentication failure when password contains YAML special characters (e.g. double quotes) by using the escape_string Handlebars helper.
+    - description: Fix authentication failure when username or password contains YAML special characters (e.g. double quotes) by using the escape_string Handlebars helper.
       type: bugfix
       link: https://github.com/elastic/integrations/issues/18434
 - version: "1.20.1"
--- a/packages/elasticsearch/data_stream/cluster_stats/agent/stream/stream.yml.hbs
+++ b/packages/elasticsearch/data_stream/cluster_stats/agent/stream/stream.yml.hbs
@@ -8,7 +8,7 @@ scope: {{scope}}
 username: {{username}}
 {{/if}}
 {{#if password}}
-password: {{password}}
+password: {{escape_string password}}
 {{/if}}
 {{#if api_key}}
 api_key: {{api_key}}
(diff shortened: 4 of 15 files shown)


packages/elasticsearch/elasticsearch/transform/index_pivot/transform.yml
```diff
@@ -119,7 +119,7 @@ pivot:
         script: "Math.max(0, params.end-params.start)"
 dest:
   index: "monitoring-indices"
-  pipeline: "1.20.1-monitoring_indices"
+  pipeline: "1.20.2-monitoring_indices"
```
andrewkroh: There is a helper for this that some other packages are using. I'd test with this so we don't forget to bump the version.

`pipeline: '{{ ingestPipelineName "monitoring_indices" }}'`

packages/elasticsearch/data_stream/ccr/agent/stream/stream.yml.hbs
```diff
@@ -8,7 +8,7 @@ scope: {{scope}}
 username: {{username}}
```
andrewkroh: To be defensive we would probably also escape the username everywhere.

consulthys (approved):
LGT Stack Monitoring

packages/elasticsearch/elasticsearch/transform/index_pivot/transform.yml
maramos-elastic: Nice call, thx

maramos-elastic: Just inplemented
```diff
-  pipeline: "1.20.2-monitoring_indices"
+  pipeline: '{{ ingestPipelineName "monitoring_indices" }}'
```

packages/elasticsearch/data_stream/ccr/agent/stream/stream.yml.hbs
maramos-elastic: done

packages/elasticsearch/elasticsearch/transform/index_pivot/transform.yml
consulthys: @3kt this might solve the hardcoded version thingie that we've been fighting against all this time

pickypg (approved):
LGTM. Thanks @andrewkroh for the 👀

PR #18503, opened 2026-04-17
Jamf Pro: add support for ECS entity fields

chemamartinez:
## Proposed commit message

```
Adds ECS entity field support to the Jamf Pro integration for both
the inventory and events data streams.

Maps managed status, OS version, last activity timestamp, group
memberships, and local admin accounts to host.entity.* and
user.entity.* fields:

- host.entity.attributes.managed from general.remote_management.managed
- host.os.* (name, version, full, platform) moved from standalone
  os.* to the host.os.* namespace in both pipelines
- host.entity.lifecycle.last_activity from general.last_contact_time
  (inventory) and webhook.event_timestamp (events)
- user.group.id and user.group.name from group_memberships[*].
  It requires the GROUP_MEMBERSHIPS section to be enabled.
- user.entity.attributes.administered_by from local_user_accounts[*]
  where admin == true, collecting {user_id, username, email} per entry.
  It requires the LOCAL_USER_ACCOUNTS section to be enabled.

Test sample are AI-generated logs with random data.
```

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/18293


packages/jamf_pro/data_stream/inventory/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -199,17 +207,20 @@ processors:
       ctx.jamf_pro.inventory.operating_system.version = normalize(ctx.jamf_pro.inventory.operating_system.version);
       String name = full_name(ctx.jamf_pro.inventory.operating_system.version);
       if (name != '') {
-        if (ctx.os == null) {
-          ctx.os = [:];
+        if (ctx.host == null) {
+          ctx.host = [:];
         }
```
efd6:
```suggestion
        ctx.host = ctx.host ?: [:];
```
(similar below)

packages/jamf_pro/data_stream/inventory/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -224,6 +235,49 @@ processors:
...
+    ignore_missing: true
+    ignore_failure: true
+- script:
+    tag: script_collect_admin_accounts
+    lang: painless
+    if: ctx.jamf_pro.inventory?.local_user_accounts != null
+    source: |-
+      List admins = new ArrayList();
+      for (def acct : ctx.jamf_pro.inventory.local_user_accounts) {
+        if (acct.admin == true) {
```
efd6: Prefer to invert the condition and continue in that case. This moves the code left.

efd6 (commented):
Add the provenance of the test inputs to the commit message.

efd6 (approved):

alexreal1314:
@chemamartinez is adding *.relationships should be part of this PR? is it something that could be added or is there another PR that should address it?

chemamartinez:
@alexreal1314 for the administers field, it happens the same as for the other integrations, we decided to revert the change after it switched the direction (see https://github.com/elastic/integrations/pull/18503/changes/3616e5d6d31802fa5e3109c0ef6584caee1b271d), as from the source we only can get the administrator for each entity.

It happens the same with the `owns` field. The API only retrieved the owners of each entity, so adding the field would require a `owner` ECS field.

That data is still available in dedicated custom fields in events, so these fields can be correlated and populated in later steps with transforms or ESQL.

PR #18508, opened 2026-04-17
[elastic_agent] Add TSDB dimensions and fix metric_type declarations

### LINKED ISSUE #17349: [Elastic Agent] Evaluate if we should increase the dimensions for the metrics data streams
https://github.com/elastic/integrations/issues/6938 and https://github.com/elastic/integrations/issues/16510 have turned the metrics data streams into TSDB. It uses basically `agent.id`, `component.id` and `metricset.name` as dimensions. ECS fields like `cloud.*`, `container.*` and `host.*` aren't part of the dimensions. This task is to evaluate if it'd bring value to add other fields, like `cloud.*`, `container.*` and `host.*`, to the dimensions. What impact adding more fields would have in storage and query performance.

AndersonQ:
## Proposed commit message

```
[elastic_agent] Add TSDB dimensions and fix metric_type declarations

Add `dimension: true` to fields across the 12 elastic_agent metrics data streams. The added dimensions are redundant with the existing dimensions (1-to-1 with `agent.id` or `component.id`) and therefore do not increase time-series cardinality.

Fix metric_type on:
- beat.stats.libbeat.pipeline.events.active: counter -> gauge
- beat.stats.libbeat.output.events.active: counter -> gauge
- filebeat_input.*.histogram.count: gauge -> counter

Add metric_type on numeric fields that were missing it: system stats, cpu ticks, memstats, handles, runtime.goroutines, uptime, cgroup stats, libbeat pipeline/config/output metrics, write-latency histogram, system.process.cgroup.{memory.mem.failures, cpuacct.percpu}, and filebeat_input.{cel_executions, system_packet_drops}.

Assisted by Claude Code
```
## Checklist

## Author's Checklist

- [ ] 

## How to review the PR

In order to better see what fields are being changes I suggest using git to see the diff with more context:
 - fetch/switch to the branch
 - run `git show -U10 HEAD`

the flag `-U` controls how much context (lines before and after the change) to show.

- spin up a stack. (I found an [issue with the agent hostname](https://github.com/elastic/integrations/issues/18523)) filter on 8.15.5, so, use the 8.19.x to have the dashboard working properly)
- install agents (I installed with fleet-server, defend and packetbeat)
- check the dashboard
- look for errors

## Related issues

- Closes #17349

### DIFF
--- a/packages/elastic_agent/data_stream/apm_server_metrics/fields/ecs.yml
+++ b/packages/elastic_agent/data_stream/apm_server_metrics/fields/ecs.yml
@@ -9,8 +9,10 @@
   dimension: true
 - name: agent.name
   external: ecs
+  dimension: true
 - name: agent.type
   external: ecs
+  dimension: true
 - name: agent.version
   external: ecs
 - name: log.level
--- a/packages/elastic_agent/changelog.yml
+++ b/packages/elastic_agent/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.9.0"
+  changes:
+    - description: Add TSDB dimensions and complete/fix metric_type declarations across elastic_agent metrics data streams.
+      type: enhancement
+      link: "https://github.com/elastic/integrations/pull/18508"
 - version: "2.8.0"
   changes:
     - description: Replace markdown navigation links with Links panels for space-aware dashboard navigation
--- a/packages/elastic_agent/data_stream/apm_server_metrics/fields/agent.yml
+++ b/packages/elastic_agent/data_stream/apm_server_metrics/fields/agent.yml
@@ -38,8 +38,10 @@
       external: ecs
     - name: domain
       external: ecs
+      dimension: true
     - name: hostname
       external: ecs
+      dimension: true
     - name: id
       external: ecs
     - name: ip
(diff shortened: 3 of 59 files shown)


packages/elastic_agent/data_stream/apm_server_metrics/fields/ecs.yml
```diff
@@ -9,8 +9,10 @@
   dimension: true
 - name: agent.name
   external: ecs
+  dimension: true
```
leehinman, line 12: seems suspicious.  `agent.id` is already a dimension.  Do we have examples were the `agent.id` is the same but `agent.name` and/or `agent.type` is different?

AndersonQ: it should not happen. This PR is indeed adding redundant dimensions, that's the idea. The only situation where the agent name and the agent ID still the same is if the hostname changes, like enabling/disabling FQDN. I cannot imagine why agent.type would ever change

AndersonQ: here's why: https://github.com/elastic/integrations/pull/17379#issuecomment-3997888147

leehinman: Interesting, thanks for the history.

leehinman (approved):

PR #18510, opened 2026-04-18
[O365] - Prune stale listing URLs from cursor to recover from long outages

ShourieG:
## Type of change
- Bug

## Proposed commit message

```
packages/o365: prune stale listing URLs from cursor to recover from long outages

A listing URL queued in cursor.todo_links before a prolonged outage
can age past the Management API's 7-day startTime window while parked
there, producing a permanent AF20055 retry loop: the listing-error
branch cannot evict it because its map-shaped error return is treated
by the CEL runtime as a freeze signal that discards cursor mutations.

Add a pre-flight filter that drops listing URLs whose embedded
startTime is older than now - maximum_age, tolerating both startTime
and starttime spellings (NextPageUri responses have been observed
with the lowercase form). Includes a system test that queues a
NextPageUri with a 2019 startTime against a 1h maximum_age and
asserts it is not followed.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -174,6 +174,43 @@ program: |-
...
+                      st[0].parse_time(time_layout.RFC3339) >= oldest_allowed_start
+                  )
+                )
+              ),
+            }
+          ),
+        }
+      )
+    )
+  ).as(state,
```
efd6: This could be folded into the previous block.
```
state.base.content_types.split(",").map(t, t.trim_space()).as(configured_types,
  state.with(
    {
        {
          // Listing links have the content type embedded in the URL query
          // string, so links for removed types would produce API errors
          // if not pruned here.
          //
          // Links whose startTime has aged past maximum_age are also
          // pruned. The Office 365 Management API rejects requests whose
          // startTime is more than 7 days in the past with AF20055; a
          // link queued before a prolonged period without successful
          // fetches (agent downtime, auth failures, credential rotation
          // gaps, etc.) may have aged past that boundary while parked in
          // the cursor. The listing-error branch cannot evict it reliably
          // because its map-shaped error return is treated by the runtime
          // as a freeze signal that discards cursor mutations. Once
          // pruned, the bottom branch generates a fresh link clamped to
          // the valid window on the next iteration.
          "todo_links": state.?cursor.todo_links.orValue([]).filter(link,
            configured_types.exists(ct, link.to_lower().contains("contenttype=" + ct.to_lower()))
            &&
            link.parse_url().RawQuery.parse_query().as(q,
              // Listing URLs generated by this program use "startTime",
              // but NextPageUri values returned by the API have been seen
              // with the lowercase "starttime" variant (see #15325), so
              // accept either. If no startTime can be extracted, keep the
              // link and let the API itself decide.
              q.?startTime.orValue(q.?starttime.orValue([])).as(st,
                (st.size() == 0) ?
                  true
                :
                  now() - st[0].parse_time(time_layout.RFC3339) <= duration(state.base.maximum_age)
              )
            )
          ),
          "todo_content": state.?cursor.todo_content.orValue([]).filter(item,
            configured_types.exists(ct, ct.to_lower() == item.?contentType.orValue("").to_lower())
          ),
        }
      ),
    }
  )
)
```

ShourieG: @efd6, made the suggested changes

efd6 (approved):

chrisberkhout (commented):
I notice that this throws away the listing link, but doesn't rewrite where it's up to (in `state.cursor.last_for`, which will be set to the listing link's end time).

That means that if the range of the listing link (default 1h, max 24h) overlaps the range where data is available (7d), it can create a gap.

For example, if I have a listing link based on a 24h `batch_interval`, starting 7.5 days ago and ending 6.5 days ago, this logic will throw away the link because it starts > 7 days ago, and there will be an avoidable gap from 7 days ago to 6.5 days ago.

I didn't file an issue because with default settings the data loss chance and amount are small, and not very recent.

PR #18528, opened 2026-04-20
[okta] Fix user.target.email never being set for target users

seanrathier:
## Summary

`user.target.email` is never populated for Okta target user entities due to a field name mismatch in the ingest pipeline.

**Root cause:** The pipeline's Painless script (tag: `okta-target-modifications`) renames `alternateId` → `alternate_id` on every entry in `okta.target`. The first `User`-type entry is then stored as `okta_target_user`. However, the subsequent `set` processor references `okta_target_user.login` — a field that was never created — instead of `okta_target_user.alternate_id`.

```yaml
# Before (broken): condition is always false, email never set
- set:
    field: user.target.email
    copy_from: okta_target_user.login        # ← field does not exist
    if: ctx.okta_target_user?.login != null

# After (fixed):
- set:
    field: user.target.email
    copy_from: okta_target_user.alternate_id  # ← field set by the Painless script
    if: ctx.okta_target_user?.alternate_id != null
```

**Impact:** Any feature that reads `user.target.email` from Okta system logs (e.g. Entity Analytics `communicates_with` relationship maintainer) receives no data from this index, even when hundreds of thousands of qualifying events are present.

## Test plan

- [ ] Ingest an Okta `application.user_membership.add` or similar event with a User-type target
- [ ] Confirm `user.target.email` is now populated with the target user's login/email (previously `null`)
- [ ] Confirm `user.target.full_name` and `user.target.id` are unaffected

### DIFF
--- a/packages/okta/data_stream/system/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/okta/data_stream/system/elasticsearch/ingest_pipeline/default.yml
@@ -545,8 +545,8 @@ processors:
       if: ctx.okta_target_user?.id != null
   - set:
       field: user.target.email
-      copy_from: okta_target_user.login
-      if: ctx.okta_target_user?.login != null
+      copy_from: okta_target_user.alternate_id
+      if: ctx.okta_target_user?.alternate_id != null
   - set:
       field: user.target.group.name
       copy_from: okta_target_group.display_name


chrisberkhout (changes requested):
Hi @seanrathier ,

What's the status of the PR draft?

Looking at the API doc and the pipeline, I think the change is correct.

To merge we'd want to regenerate the pipeline tests and see the change in expected output. I believe this case is covered, so those will change with the corrected logic.

There should also be a changelog entry.

Issue #18318, opened 2026-04-09
[Github] Make the PAT settings in the Github Security Advisories dataset optional


leandrojmp:
Hello,

The [Github Security Advisories](https://docs.github.com/en/rest/security-advisories/global-advisories?apiVersion=2026-03-10) endpoint can be used without authenticaton, but the Elastic Agent integration requires that a PAT is set.

This configuration should be optional and only required if needed as described in Github documentation:

> This endpoint can be used without authentication if only public resources are requested.

brijesh-elastic:
Hi @leandrojmp,
The [PR](https://github.com/elastic/integrations/pull/18545) has been merged. Starting with version `2.23.1`, you will be able to pull `Security Advisory` data without providing authentication credentials.

PR #18545, opened 2026-04-21
[github] change API key variable to optional for security advisories data stream

brijesh-elastic:
## Proposed commit message

```
github: change API key variable to optional for security advisories data stream,
as authentication is not required for public repositories

[1] https://docs.github.com/en/rest/security-advisories/global-advisories?apiVersion=2026-03-10#list-global-security-advisories
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/github directory.

> elastic-package test -v

## Related issues

- #18318

### DIFF
--- a/packages/github/data_stream/security_advisories/agent/stream/cel.yml.hbs
+++ b/packages/github/data_stream/security_advisories/agent/stream/cel.yml.hbs
@@ -11,9 +11,14 @@ state:
 {{/if}}
   advisory_type: {{advisory_type}}
   batch_size: {{batch_size}}
+{{#if api_key}}
 redact:
   fields:
     - api_key
+{{else}}
+redact:
+  fields: ~
+{{/if}}
 regexp:
   github_link_next: '<([^>]+)>; rel="next"'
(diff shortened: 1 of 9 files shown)


packages/github/data_stream/security_advisories/agent/stream/cel.yml.hbs
kcreddy, line 17: Any error during `redact` when field doesn't exist?
I think its good to have a system test for non-authentication case.

brijesh-elastic: > Any error during redact when field doesn't exist?

No, But keeping it safe, https://github.com/elastic/integrations/pull/18545/changes/68b8906b29d1db021b36716f8e415a8811515fda

> I think its good to have a system test for non-authentication case.

Yes, https://github.com/elastic/integrations/pull/18545/changes/68b8906b29d1db021b36716f8e415a8811515fda

kcreddy (approved):

PR #18548, opened 2026-04-21
Update default index pattern for ML jobs to use the transform index

sodhikirti07:
## Proposed commit message
Updated default index pattern for ML jobs to use the transform index directly

## Checklist

## Author's Checklist

- [ ] updated changelog.yml and manifest.yml
- [ ] updated README.md and removed the step to create multi data view for ml jobs
- [ ] updated ml.json files for LMD, DED and PAD to include hardcoded index patterns for jobs that use transform destination indices
- [ ] bumped versions in transforms

- Tested using [ITP](https://github.com/elastic/security-ml/tree/main/src/integrations-test-pipeline)

## Related issues

- 

## Screenshots
image


packages/ded/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: "3.1.0"
+  changes:
+    - description: Update default index pattern for ML jobs to use the transform-generated index directly
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/17626
```
susan-shu-c: Should be https://github.com/elastic/integrations/pull/18548/?

packages/ded/elasticsearch/transform/pivot_transform_ea/transform.yml
jmcarlock:
```suggestion
  index: ml_network_ded_ea-3.1.0
```

jmcarlock (approved):

susan-shu-c (approved):
Could we get Tom V to approve or look also?

sodhikirti07:
@tveasey @peteharverson Could any one of you review this PR as well?

tveasey:
Can you provide a little context into the change: it looks like you updating the source directory for ML job inputs. My main thought would be around backwards compatibility and the upgrade path. What would happen to jobs which were pointing at the old location after upgrade? Is it the case that this only takes effect on newly created jobs? If so and there is no migration path, will the data source for the old jobs still exist and receive data?

sodhikirti07:
> Is it the case that this only takes effect on newly created jobs? If so and there is no migration path, will the data source for the old jobs still exist and receive data?

@tveasey This change will take effect only after a user upgrades to this particular version of the package. Once upgraded (and the ML jobs are updated), users will see the new index pattern input for their jobs. Existing jobs will continue to receive data from their original source indices.

jgowdyelastic (approved):
ML job changes LGTM

PR #18549, opened 2026-04-21
Update sync-packages workflow to support nested package directories

mrodm:
## Proposed commit message

Update the GitHub Workflow to update the issue templates (`sync-packages-to-bug-issue-template` via updatecli) to support packages that are located in nested directories.

**WHAT:** Update the `sync-packages-to-bug-issue-template` updatecli script and the `bump-elastic-stack-version` workflow to support packages living in multiple (nested) directories.
- Replace the manual `ls packages/` loop with `mage listPackages`, which returns the full path to each package regardless of which directory it lives in.
- Read `title` from each `manifest.yml` using the path returned by `mage listPackages`, and display the full package path (e.g. `packages/foo/bar`) in the issue template options list instead of just the top-level folder name.
- Add Go and Mage installation steps to the `bump-elastic-stack-version` workflow so that `mage listPackages` is available at runtime.
                
**WHY:** The existing script assumed all packages live directly under `packages/`, breaking for integrations that use nested directories. Delegating package discovery to `mage listPackages` makes the workflow resilient to any directory layout.

## Checklist

## Author's Checklist

- [x] Verify that the `bump-elastic-stack-version` workflow correctly installs Go and Mage before the Updatecli step runs.
- [x] Confirm the issue template options list shows full package paths after this change.

1. Ensure `mage` is installed and `mage listPackages` returns the expected list of package paths (including any nested ones).
2. Copy the script in the `.github/workflows/updatecli/updatecli.d/sync-packages-to-bug-issue-template.yml` locally to a script, and try to run it.
   ```
   cat <<EOF > sync_issues.sh
   #!/usr/bin/env bash
   set -eu

   package_paths=$(mage listPackages)

   pkgs=$(echo "$package_paths}" | while IFS= read -r package_path; do
     title=$(yq .title "${package_path}/manifest.yml")
     echo "- $title [$package_path]"
   done | LC_ALL=C sort -s -f | yq -o json .)
        
   yq eval ".body[1].attributes.options = $pkgs" -i '.github/ISSUE_TEMPLATE/integration_bug.yml'
   yq eval ".body[1].attributes.options = $pkgs" -i '.github/ISSUE_TEMPLATE/integration_feature_request.yml'
      
   git diff --name-only HEAD
   EOF
   
   bash sync_issues.sh
   git diff .github/ISSUE_TEMPLATE
   ```
3. Inspect `.github/ISSUE_TEMPLATE/integration_bug.yml` and `integration_feature_request.yml` to confirm
the options list contains - <title> [<package_path>] entries for all packages, including nested ones. 
                
## Related issues
- Part of #17403
- Requires #18543

---

> This PR was generated with the assistance of Claude.


.github/workflows/updatecli/updatecli.d/sync-packages-to-bug-issue-template.yml
```diff
@@ -23,19 +23,20 @@ targets:
...
-          title=$(yq .title packages/$i/manifest.yml)
-          pkg=$i
-          echo "- $title [$pkg]"
-        done | sort -s -f | yq -o json .)
+        package_paths=$(mage listPackages)
+
+        pkgs=$(echo "$package_paths" | while IFS= read -r package_path; do
+          title=$(yq .title "${package_path}/manifest.yml")
+          echo "- $title [$package_path]"
+        done | LC_ALL=C sort -s -f | yq -o json .)
```
mrodm, line 39: Instead of writing the package folder name (e.g `1password`), I propose to use the full package path (e.g. `packages/1password`) instead here.

As there could be packages in nested directories, that could help to locate the corresponding package. Moreover, it could happen that the same package folder name is repeated in a different nested directory even if the package names (from manifest.yml file).

WDYT @andrewkroh @teresaromero @kpollich ?

Example:
- Current approach in main:
```yaml
  - type: dropdown
    id: integration_name
    attributes:
      label: Integration Name
      description: Which integration package are you having issues with?
      options:
        - 1Password [1password]
        - Abnormal AI [abnormal_security]
        - abuse.ch [ti_abusech]
        - Active Directory Entity Analytics [entityanalytics_ad]
        - ActiveMQ OpenTelemetry Assets [activemq_otel]
```
- Change proposed in this PR:
```yaml
  - type: dropdown
    id: integration_name
    attributes:
      label: Integration Name
      description: Which integration package are you having issues with?
      options:
        - 1Password [packages/1password]
        - Abnormal AI [packages/abnormal_security]
        - abuse.ch [packages/ti_abusech]
        - Active Directory Entity Analytics [packages/entityanalytics_ad]
        - ActiveMQ OpenTelemetry Assets [packages/activemq_otel]
```

.github/workflows/bump-elastic-stack-version.yml
```diff
@@ -24,6 +24,14 @@ jobs:
...
       - uses: actions/checkout@v6
 
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: .go-version
+
+      - name: Install mage
+        run: go install github.com/magefile/mage
+
```
mrodm, line 34: Required to execute `mage listPackages` in the updatecli workflow.

.github/workflows/updatecli/updatecli.d/sync-packages-to-bug-issue-template.yml
mrodm: cc @lucabelluccini for awareness in case it is changed that list in the templates

kpollich: I think including the full path is relatively harmless and only adds clarity in cases where integration names might collide w/ composable packages. +1 from me on switching to `packages/*` for the options here.

andrewkroh: That seems good if it helps remove ambiguity, given the new option to put packages in more places.

teresaromero: +1 to clarity. However, reading the [parent](https://github.com/elastic/integrations/issues/17403) issue i find this example

>so for example the nginx_input_otel package can be stored under packages/nginx, as input_otel, along with other packages related to nginx.

in this case, the name on the manifest will be telling us the folder structure, is this coincidence of the example or could it lead to other combination? i mean this because, if the naming convention is like that, the name at the manifest will be unique? cc @jsoriano

mrodm: > if the naming convention is like that, the name at the manifest will be unique?

Package names in the manifest are ensured to be unique via the changes introduced in this PR (`dev/packagenames` tool).
https://github.com/elastic/integrations/pull/18187


> in this case, the name on the manifest will be telling us the folder structure, is this coincidence of the example or could it lead to other combination?

Currently, there is no restriction about the folder names it could be created for new packages. That's why it was introduced this new validation, specially when they could be created in nested directories.

For instance, it could be created the nested packages as (even it could be redundant to repeat `nginx`) or other different folder names:
- packages/nginx/nginx
- packages/nginx/nginx_input_otel


@teresaromero would this answer your doubts? Not sure if I'm missing anything

.github/workflows/bump-elastic-stack-version.yml
jsoriano: Should we pin a version of mage here?

mrodm: I think this is not needed, since this step is running after the checkout action. I think in that case, this action is running in the root of the working directory and therefore the `go install` command would install the mage version specified in the `go.mod` file.

I did a test in this PR by forcing to use another mage version (`v1.17.1`) instead of the latest one (`v1.17.2`) here: https://github.com/elastic/integrations/pull/18549/commits/bc1f22bcd5a4b53c655b3f4ac3d6b1b385af3a94

And the workflow/step installed the v1.17.1 ([workflow link](https://github.com/elastic/integrations/actions/runs/25068772019/job/73443027058?pr=18549#step:4:7)):
```
Run go install github.com/magefile/mage
  go install github.com/magefile/mage
  shell: /usr/bin/bash -e {0}
  env:
    GOTOOLCHAIN: local
go: downloading github.com/magefile/mage v1.17.1
```

jsoriano: Ah you are right, it is already in the `go.mod` 👍

jsoriano (approved):

.github/workflows/updatecli/updatecli.d/sync-packages-to-bug-issue-template.yml
teresaromero: thanks!

teresaromero (approved):

PR #15005, opened 2025-08-21
[prisma_access] Handle Malformed URL and Parsing Issue

mohitjha-elastic:
## Proposed commit message
```
prisma_cloud: Handle malformed URL in CEF string and parsing issue in pipeline processors.

Add `*.client.to_firewall_str` and `*.firewall_to_client_str` fields to capture original string values when 
boolean conversion fails in PanOSFirewallToClient and PanOSClientToFirewall fields.
Add support for UNIX timestamp format in `*.time.not_after` and `*.time.not_before` fields.
Handle malformed URL in CEF string.
```

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [X] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally

- Clone integrations repo.
- Install elastic package locally.
- Start elastic stack using elastic-package.
- Move to integrations/packages/prisma_access directory.
- Run the following command to run tests.

> elastic-package test -v

PR #18551, opened 2026-04-21
[prisma_access] Fix url escape issues

vinit-chauhan:
## Proposed commit message

**[prisma_access] Fix escaped `=` handling in CEF extension keys to prevent decoding errors**

In v1.6.1, a `replace` processor was added to strip backslash-escaped `=` characters (`\=`) from CEF messages before decoding, as a workaround for URLs containing literal `\=` sequences (e.g. query parameters like `?gtm\=value`). However, the processor was overly aggressive — it replaced `\=` with bare `=`, which corrupted the CEF extension key-value structure. The `decode_cef` processor then failed to parse the resulting message because the unescaped `=` characters inside values were misinterpreted as key-value delimiters.

This fix corrects the `replace` processor to:
- Only trigger when the message actually contains `\\=` (double-backslash-escaped equals).
- Replace `\\=` with `\=` (single-backslash-escaped), preserving the CEF structure rather than stripping the escape entirely.

This ensures the `decode_cef` processor can correctly parse CEF extension values that contain escaped `=` without breaking the key-value parsing.

A new test log line with URL query parameters containing escaped `=` has been added to validate the fix end-to-end.

## Checklist

1. Build and deploy the `prisma_access` package using `elastic-package`:
   ```bash
   cd packages/prisma_access
   elastic-package build
   elastic-package test system
   ```
2. Verify the system test passes with 21 hit counts (previously 20).
3. Send a CEF log line containing URL query parameters with `\=` escapes (like the new test log in `sample_logs/prisma_access.log`) over TCP and confirm:
   - The `decode_cef` processor does not error out.
   - The `request` field correctly contains the URL with `\=` sequences preserved.

## Related issues

- Relates https://github.com/elastic/integrations/pull/15005 (v1.6.1 — original "Handle malformed URL in CEF string" fix that introduced the regression)

## Screenshots
Value with `\\=` 
image

Value with `\=`
image


packages/prisma_access/data_stream/event/agent/stream/tcp.yml.hbs
```diff
@@ -25,15 +25,17 @@ processors:
         to: event.original
     fail_on_error: true
     ignore_missing: false
+
 # The decode_cef processor treats '\=' as a CEF escape sequence inside
-# values, which causes parse failures for URLs containing literal '\='
-# (e.g. query parameters like '?gtm\=value'). Strip these before decoding.
+# values, and is implemented in decode_cef processor. There was a bugfix ( v1.6.1 of this package )
+# where customer had `\\=` in values, which is not a valid CEF escape sequence.
```
andrewkroh:
```suggestion
# where user had `\\=` in values, which is not a valid CEF escape sequence.
```

kgeller (approved):
LGTM

PR #16650, opened 2025-12-19
osquery_manager: update ECS mapping

mmahacek:
A number of ECS fields are not defined within the integration field mapping, which is causing them to map as object instead of the expected nested type. These fields are:

dll.pe.sections
file.macho.sections
file.pe.sections
process.parent.macho.sections
process.parent.pe.sections
threat.enrichments.indicator.file.pe.sections
threat.indicator.file.pe.sections

Additionally, the integration defines an error field as text, however this conflicts with the ECS error field that is expected to be an object.

## Proposed commit message

osquery_manager: update ECS mapping and rename error to error.message

Please explain:

- WHAT: patterns used, algorithms implemented, design architecture, message processing, etc.
- WHY:  the rationale/motivation for the changes

This text will be pasted into the squash dialog when the change is committed and will be
a long term historical record of the change to help future contributors understand the
change, please help them by making it clear and comprehensive, they may be you.

If the commit title is adequate to describe both of these things, The text here may be omitted
or replaced with "See title". The title of the PR will be used as the commit message title when
the merge is made and the "See title" marker will be removed if present.

The text here and the PR title will be subject to the PR review process.
-->
Update ECS mapping for the osquery_manager.results datastream.
Update osquery_manager.action_responses ingest pipeline to rename `error` to `error.message`

## Checklist

- [X] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [X] I have verified that all data streams collect metrics or logs.
- [X] I have added an entry to my package's `changelog.yml` file.
- [X] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [X] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## How to test this PR locally

There are no special tests to run outside of `elastic-package test`

## Related issues

- Closes https://github.com/elastic/integrations/issues/16643

PR #18608, opened 2026-04-23
[osquery_manager] Add `query_profile` data stream

marc-gr:
## Proposed commit message

Adds the query_profile data stream to enable receiving profile from shceduled and live queries if configured.

It includes a dashboard.

Query profiles dashboard

## Checklist


szwarckonrad (commented):
Leaving this Claude Opus 4.7 generated review since I lack expertise in this project to have any meaningful comments myself 😄 .


 Nice to see profiling exposed as its own stream. Pulled the branch and diffed it against `result/` and `action_responses/`. The skeleton matches the existing convention — `base-fields.yml` and `elastic_agent.yml` are byte-identical to `action_responses/`, the `action_data` group mirrors `action_responses/fields/action-responses.yml:21-42`, and `host.os.build` is correctly placed in the package-custom file (it's listed as non-ECS in `_dev/scripts/osquery-gen/generator.go:38`).

A few things worth a second look before merge:

**Likely needs action**

- **Changelog references a dashboard that isn't in the diff.** `changelog.yml` says "Add `query_profile` data stream and a Kibana dashboard for query profile events (Lens on `logs-*`)", but no `kibana/dashboard/*.json` file is added. PR body checklist also has the dashboard-good-practices box checked. Either ship the dashboard or trim the changelog wording.
- **PR is not rebased.** Diff inserts `1.29.0` directly above `1.28.0`, but `main` already has `1.28.1` at the top (#16650). Rebase needed.
- **Top-level `type` field (`query-profile.yml:1-4`) is a footgun.** Bare `type: keyword` collides visually with `data_stream.type` in field-picker UIs and is non-ECS. Consider `event.kind` / `event.category` / `event.type`, or fold the discriminator into the existing `osquery_profile.source` field (`query-profile.yml:46-49`), which already distinguishes `scheduled` vs `live`.
- **Field redundancy: `query.name` (`query-profile.yml:6-12`) and `osquery_profile.query_name` (`query-profile.yml:50-53`) are the same datum** under two paths. Pick one.
- **`osquery_profile.query` (`query-profile.yml:54-58`) is `text` with no `keyword` multi-field.** Without it, "top N slowest queries by SQL" can't aggregate. Peer field `error` in `action_responses/fields/action-responses.yml:68-75` adds the keyword sub-field — same pattern would help here.

**Worth confirming with reviewers**

- **No ingest pipeline.** `result/` and `action_responses/` both ship one (`action_responses/elasticsearch/ingest_pipeline/default.yml:1-19`) that overrides `data_stream.{type,dataset}` and copies `agent.id` → `agent_id`. The new stream has neither the pipeline nor an `agent_id` top-level field. Intentional, or oversight? If anything cross-streams on `agent_id`, profile docs will be the odd one out.
- **`stream.yml.hbs` content `enabled: true` is unique in this repo** — grepped, no other stream uses just that. The other two `osquery_manager` streams render `query: {{query}}` from a hidden `query` var in their manifests. Is `enabled` actually consumed by osquerybeat, or is this a placeholder that could follow the existing `osquery_manager` idiom for consistency?
- **No scheduled-query correlation keys.** `action_responses/fields/action-responses.yml` carries `schedule_id` (line 27), `pack_id` (line 31), `schedule_execution_count` (line 39), and `planned_schedule_time` (line 44) — the keys to pin a scheduled response to its iteration. The new stream has only `action_id` (described as live-only) and `response_id`. If scheduled profiles are meant to join to `action_responses` or result documents, `schedule_id` / `pack_id` may be missing.

**Smaller**

- Description body has a typo ("shceduled") and is single-sentence; the WHAT/WHY split used in #18305's body would help reviewers.
- `query-profile.yml:66-68` says `duration` is milliseconds, but osquery's `osquery_schedule.total_duration` is microseconds. If osquerybeat converts before shipping, fine — worth a one-liner in the description so it's documented. Same instinct for `memory` (bytes) and `cpu_time` (pre-summed vs computed in ES).
- `event.module` was added to `query_profile/fields/ecs.yml:18-19` — likely consistent with how dashboards already filter (`event.module: "osquery_manager"`); note it's missing from `action_responses/fields/ecs.yml`, which is probably an oversight there rather than a problem here.

packages/osquery_manager/data_stream/query_profile/fields/query-profile.yml
```diff
@@ -0,0 +1,88 @@
+- name: query
+  title: Query reference
+  type: group
+  fields:
+    - name: name
+      type: keyword
+      ignore_above: 1024
```
tomsonpl, line 7: Just checking on this one since `query.name` can fall back to the raw SQL (`action_data.query`), and it's a `keyword` with `ignore_above: 1024`, would SQL longer than 1024 chars end up not indexed? If so I think those executions would quietly drop out of the "Top queries" table and the Query control. Wondering if a `keyword` sub-field on a `text` field (or an explicit truncate) might be safer here? 🙂

packages/osquery_manager/data_stream/query_profile/fields/query-profile.yml
```diff
@@ -0,0 +1,88 @@
...
+      enabled: false
+    - name: id
+      type: keyword
+      ignore_above: 1024
+    - name: platform
+      type: keyword
+      ignore_above: 1024
+    - name: query
+      type: keyword
+      ignore_above: 1024
```
tomsonpl, line 42: Same small question here as on `query.name` - `action_data.query` is `keyword`/`ignore_above: 1024` and feeds the "Live query SQL" control and "Top live queries" table, so I think longer ad-hoc SQL might silently drop from those. Does that match what you're seeing? Happy to be wrong!

packages/osquery_manager/data_stream/query_profile/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,47 @@
...
+    ignore_empty_value: true
+    override: false
+- set:
+    field: query.name
+    copy_from: action_data.saved_query_id
+    ignore_empty_value: true
+    override: false
+- set:
+    field: query.name
+    copy_from: action_data.query
```
tomsonpl: Curious about the trade-off here: folding name / `saved_query_id` (UUID) / raw SQL into one `query.name` means the "Top queries" bucket mixes readable names with UUIDs and SQL blobs. Totally fine if that's the intent — just wondering whether keeping the source-specific fields and computing a display label in Lens might read a bit cleaner? No strong opinion. 🙂

packages/osquery_manager/data_stream/query_profile/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,47 @@
...
+    ignore_empty_value: true
+    override: false
+# The per-profile query name is now consolidated into query.name.
+- remove:
+    field: osquery_profile.query_name
```
tomsonpl: Small thought - once `osquery_profile.query_name` is removed we lose the original scheduled name and which branch populated `query.name`, and since it's at ingest time it can't be backfilled later. Would it be worth keeping it around just in case, or is it safely redundant?

packages/osquery_manager/data_stream/query_profile/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,47 @@
...
+    value: "osquery_manager.query_profile"
+    ignore_empty_value: true
+    override: true
+- set:
+    field: agent_id
+    copy_from: "agent.id"
+    ignore_empty_value: true
+    override: true
+# Consolidate the query identity into a single, aggregatable query.name across every
+# profile source (scheduled, rrule, live). osquerybeat already sets query.name for
```
tomsonpl: Tiny doc nit to double-check: the comment says osquerybeat sets `query.name` directly for scheduled/RRULE, but the first processor copies from `osquery_profile.query_name`. Just want to make sure the comment matches what osquerybeat actually emits so the next reader isn't confused. 🙂

tomsonpl (approved):
Thanks @marc-gr , I love it! Really nice addition! The consolidated `query.name` and the dashboard are great. Left a few small things I wanted to check on, mostly around long SQL and one doc nit. Nothing blocking from me. 🙂

One more I couldn't easily anchor: since the stream is `experimental` and off until enabled, on first open the controls/terms panels may show "no options" until the first profile doc lands - might be worth a line in the About panel so it doesn't look empty/broken. Just a thought!

marc-gr:
Thanks for the review! Pushed 65e1054 addressing the feedback:

Truncate raw-SQL fallback in the pipeline so query.name stays aggregatable
Keep osquery_profile.query_name (no longer removed)
Fix pipeline comments
Remove live-only SQL dashboard panels; consolidated on query.name

raqueltabuyo:
@marc-gr review from UX regarding the dashboard has finished, could you please apply the changes? https://www.figma.com/design/PDU6zLir3pcE7xKiPe3gXS/-9.6--Dashboard---Osquery-Profiling?node-id=6439-142468&t=QeLaLzfKrWZhdSdv-0

PR #18611, opened 2026-04-23
feat: add cursor tracking to agentless_hello_world generic data stream

rubenruizdegauna:
## Summary

- Adds `cursor.request_count` (incrementing integer) and `cursor.last_collected` (timestamp) fields to the `agentless_hello_world` generic data stream's CEL program
- Both values are emitted inside each event document and persisted via the CEL cursor mechanism
- This enables testing whether cursor state survives across beats-process-to-receivers mode restarts

## How to verify

After deploying, query the logs index to see cursor values incrementing:

```
FROM logs-agentless_hello_world.generic-*
| SORT @timestamp DESC
| KEEP @timestamp, cursor.request_count, cursor.last_collected, http.response.status_code
| LIMIT 20
```

- **Cursor preserved**: `cursor.request_count` continues from where it left off after restart
- **Cursor lost**: `cursor.request_count` resets to 1

## Changes

- `data_stream/generic/agent/stream/cel.yml.hbs` — added `state.cursor` with `request_count` and `last_collected`; emit in events and persist via CEL cursor
- `data_stream/generic/fields/fields.yml` — added `cursor.request_count` (long) and `cursor.last_collected` (date)
- `data_stream/generic/sample_event.json` — updated with cursor fields
- `data_stream/generic/_dev/test/pipeline/` — updated pipeline test input and expected output
- `manifest.yml` — bumped version to 0.3.0
- `changelog.yml` — added 0.3.0 entry

### DIFF
--- a/packages/agentless_hello_world/data_stream/generic/agent/stream/cel.yml.hbs
+++ b/packages/agentless_hello_world/data_stream/generic/agent/stream/cel.yml.hbs
@@ -4,6 +4,9 @@ resource.timeout: 15s
 resource.url: "{{url}}"
 state:
   url: "{{url}}"
+  cursor:
+    request_count: 0
+    last_collected: "1970-01-01T00:00:00Z"
 program: |
   request("GET", state.url)
     .do_request()
@@ -12,9 +15,17 @@ program: |
             "http": {
               "response": {
                 "status_code": resp.StatusCode
+              }
+            },
+            "cursor": {
+              "request_count": int(state.cursor.request_count) + 1,
+              "last_collected": now
             }
-          }
-        }]
+        }],
+        "cursor": {
+          "request_count": int(state.cursor.request_count) + 1,
+          "last_collected": now
+        }
     })
 tags:
   - agentless-hello-world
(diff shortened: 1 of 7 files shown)


packages/agentless_hello_world/data_stream/generic/agent/stream/cel.yml.hbs
```diff
@@ -12,9 +15,17 @@ program: |
...
+              "request_count": int(state.cursor.request_count) + 1,
+              "last_collected": now
             }
-          }
-        }]
+        }],
+        "cursor": {
+          "request_count": int(state.cursor.request_count) + 1,
+          "last_collected": now
+        }
```
straistaru: Is there a way to compute the updated values for the cursor only once? The `request_count` formula is simple enough, but I am not sure that the timestamp provided by `now` will be the same.

It would be nice if there was a way to do those operations at the beginning of the `program` and asign the values to some variables which then can be used in both places to update the cursor.

straistaru (approved):
LGTM. Just 1 minor question/nit, but not a blocker.

packages/agentless_hello_world/data_stream/generic/agent/stream/cel.yml.hbs
rubenruizdegauna: Good point. Done in b151e283d7ca2a2c0c9f2a3ca173c0a9fa18365a

PR #18617, opened 2026-04-23
feat: add HTTPJSON input variant with cursor tracking to agentless_hello_world

rubenruizdegauna:
## Summary

- Adds an HTTPJSON input variant alongside the existing CEL input for the `agentless_hello_world` generic data stream
- Implements timestamp-based cursor tracking (`cursor.last_collected` and `cursor.previous_collected`) to detect whether cursor state persists across restarts
- Updates the ingest pipeline to parse httpjson message JSON and extract cursor/http fields to top-level, so both input types produce consistent document structure

## Depends on

- #18611 (CEL cursor tracking)

## How to verify

Deploy and select the HTTPJSON input type. After a restart, check `cursor.previous_collected`:
- **Cursor preserved**: shows a recent timestamp from before the restart
- **Cursor lost**: resets to `1970-01-01T00:00:00Z`

```
FROM logs-agentless_hello_world.generic-*
| SORT @timestamp ASC
| KEEP @timestamp, cursor.last_collected, cursor.previous_collected
| LIMIT 50
```

### DIFF
--- a/packages/agentless_hello_world/changelog.yml
+++ b/packages/agentless_hello_world/changelog.yml
@@ -1,3 +1,8 @@
+- version: "0.3.0"
+  changes:
+    - description: Add cursor tracking (request_count, last_collected) to the generic data stream for testing cursor persistence across restart modes.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/99999
 - version: "0.2.0"
   changes:
     - description: Add mock counter metrics data stream with constant and spike modes for high-rate ingestion testing.
--- a/packages/agentless_hello_world/changelog.yml
+++ b/packages/agentless_hello_world/changelog.yml
@@ -2,7 +2,7 @@
   changes:
     - description: Add cursor tracking (request_count, last_collected) to the generic data stream for testing cursor persistence across restart modes.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/99999
+      link: https://github.com/elastic/integrations/pull/18611
 - version: "0.2.0"
   changes:
     - description: Add mock counter metrics data stream with constant and spike modes for high-rate ingestion testing.
--- a/packages/agentless_hello_world/changelog.yml
+++ b/packages/agentless_hello_world/changelog.yml
@@ -2,7 +2,7 @@
   changes:
     - description: Add HTTPJSON input variant for the generic data stream with timestamp-based cursor tracking.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/18611
+      link: https://github.com/elastic/integrations/pull/18617
 - version: "0.3.0"
   changes:
     - description: Add cursor tracking (request_count, last_collected) to the generic data stream for testing cursor persistence across restart modes.
--- a/packages/agentless_hello_world/changelog.yml
+++ b/packages/agentless_hello_world/changelog.yml
@@ -1,3 +1,8 @@
+- version: "0.4.0"
+  changes:
+    - description: Add HTTPJSON input variant for the generic data stream with timestamp-based cursor tracking.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/18611
 - version: "0.3.0"
   changes:
     - description: Add cursor tracking (request_count, last_collected) to the generic data stream for testing cursor persistence across restart modes.
--- a/packages/agentless_hello_world/data_stream/generic/agent/stream/httpjson.yml.hbs
+++ b/packages/agentless_hello_world/data_stream/generic/agent/stream/httpjson.yml.hbs
@@ -1,0 +1,21 @@
+config_version: 2
+interval: 20s
+request.method: GET
+request.url: "{{url}}"
+request.timeout: 15s
+
+response.transforms:
+  - set:
+      target: body.cursor.last_collected
+      value: '[[formatDate (now) "RFC3339"]]'
+  - set:
+      target: body.cursor.previous_collected
+      value: "[[.cursor.last_collected]]"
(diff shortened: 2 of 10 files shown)


packages/agentless_hello_world/changelog.yml
```diff
@@ -1,3 +1,13 @@
+- version: "0.4.0"
+  changes:
+    - description: Add HTTPJSON input variant for the generic data stream with timestamp-based cursor tracking.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/18611
```
straistaru: Wrong PR linked

packages/agentless_hello_world/data_stream/generic/agent/stream/httpjson.yml.hbs
```diff
@@ -0,0 +1,21 @@
+config_version: 2
```
straistaru, line 1: Since this is added to the existing generic data stream, does it mean that both inputs will run at the same time now?
If this is the case, how will this impact our current usages of this integration (e.g., synthetics)?
Also, I am not sure that setting 2 inputs to run for the same integration aligns with our long term goals (but I might be wrong on this one).

packages/agentless_hello_world/manifest.yml
```diff
@@ -31,8 +31,12 @@ policy_templates:
...
     inputs:
       - type: cel
-        title: Collect data from EPR endpoint
-        description: Fetches https://epr.elastic.co every minute.
+        title: Collect data from EPR endpoint (CEL)
+        description: Fetches https://epr.elastic.co every minute using CEL input.
+        vars: []
+      - type: httpjson
+        title: Collect data from EPR endpoint (HTTPJSON)
+        description: Fetches https://epr.elastic.co every minute using HTTPJSON input.
```
straistaru: `every minute` is wrong. It does so every 20s.
I see this is a recurring issue in this document (also to lines out of the PR scope). Could you please address all of them?

packages/agentless_hello_world/data_stream/generic/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -14,6 +14,28 @@ processors:
...
+  - set:
+      tag: extract_http_from_message
+      if: ctx._parsed?.http != null
+      field: http
+      copy_from: _parsed.http
+  - remove:
+      tag: cleanup_parsed
+      field:
+        - _parsed
+        - message
```
straistaru, line 37: Is removal of the message intentional?

packages/agentless_hello_world/data_stream/generic/agent/stream/httpjson.yml.hbs
rubenruizdegauna: Good point. Right both work at the same time and both send metrics. I don't think it will affect the current usage, as it's just adding extra logs, but still it's a good point. I'll check if I can change it to select one or the other.

packages/agentless_hello_world/manifest.yml
rubenruizdegauna: Good catch. Many [C&P or AI have been around this](https://github.com/elastic/integrations/blob/main/packages/agentless_hello_world/manifest.yml#L35)... Fixing it

packages/agentless_hello_world/data_stream/generic/elasticsearch/ingest_pipeline/default.yml
rubenruizdegauna: I re-tested it and everything looks fine. Explanation which I cannot take credit for:

> Yes, removing message there is intentional and correct. Here's why:                                                                                                                                                                                                                                                                             
>                                                                                                                                                                                                                                                                                                                                                  
 > The pipeline uses message as a transport field — the CEL/HTTPJSON input dumps the raw HTTP response body into it. The processor sequence is:                                                                                                                                                                                                    
>                                                                                                                                                                                                                                                                                                                                                  
>  1. Line 17–21 — parse message (raw JSON string) → _parsed                                                                                                                                                                                                                                                                                       
> 2. Lines 23–32 — extract cursor and http from _parsed into proper fields
>  3. Lines 33–38 — cleanup: remove both _parsed (intermediate scratch) and message (now redundant raw payload)                                                                                                                                                                                                                                    
>                                                                                                                                                                                                                                                                                                                                                  
>  After step 3, message is gone because the structured data has been lifted into proper fields. Keeping it would leave a redundant raw JSON blob in every document.

> CEL builds the event explicitly in the program — it produces structured fields directly:                                                                                                                                                                                                                                                        
>   {                                                                                                                                                                                                                                                                                                                                               
>     "http": { "response": { "status_code": 200 } },                                                                                                                                                                                                                                                                                               
>     "cursor": { "request_count": 1, "last_collected": "2026-..." }                                                                                                                                                                                                                                                                                
>   }                                                                                                                                                                                                                                                                                                                                               
>   No message field. http and cursor are already at the top level.                                                                                                                                                                                                                                                                                 
>                                                                  
>   HTTPJSON works differently — it fetches the URL, the transforms inject cursor fields into the response body, and then the entire response body gets serialized a

packages/agentless_hello_world/data_stream/generic/agent/stream/httpjson.yml.hbs
rubenruizdegauna: Now CEL is enabled by default and HTTPJSON disabled. They can be enabled in the UI (Ignore the naming)

straistaru (approved):

PR #18659, opened 2026-04-27
[mongodb_atlas] Fix source.address handling in alert pipeline

### LINKED ISSUE #18441: [mongodb_atlas] source.address incorrectly overwritten, causing mapping conflict on IP/FQDN mix
## Summary

MongoDB Atlas `alert` data stream ingest pipeline converts `source.address` to `ip`, which is not ECS-aligned and can cause indexing failures when `source.address` contains FQDNs/hostnames (especially after data stream rollover / when mappings expect `ip`).

Per ECS, `source.address` should keep the raw value (hostname/FQDN/IP) and IPs should be duplicated into `source.ip` (type `ip`).

## Impact

- `response.hostnameAndPort` can contain hostname/FQDN or IPv4/IPv6.
- Pipeline groks host into `source.address`.
- Pipeline then runs `convert` to `ip` on `source.address`.
- If `source.address` becomes mapped as `ip` in the data stream backing index, subsequent events with FQDN in `source.address` are rejected with mapper/parsing errors.
- This is often observed after rollover, but can happen anytime mappings treat `source.address` as `ip`.

## Current pipeline location

`packages/mongodb_atlas/data_stream/alert/elasticsearch/ingest_pipeline/default.yml`

Problematic block:

```yaml
- grok:
    field: mongodb_atlas.alert.host_name_and_port
    patterns:
      - '%{HOSTNAME:source.address}:%{POSINT:source.port}'
    ignore_missing: true

- convert:
    field: source.address
    type: ip
    ignore_missing: true
    on_failure:
      - append:
          field: related.hosts
          value: '{{{source.address}}}'
          allow_duplicates: false

- append:
    field: related.ip
    value: '{{{source.address}}}'
    allow_duplicates: false
    if: "ctx.related?.hosts == null"
```

## Expected behavior (ECS-aligned)

Keep `source.address `as the raw value.
Only set source.ip when source.address is parseable as an IP. If FQDN, we can put it in source.domain (should check this).
related.ip should only contain IP values (e.g., from source.ip), not hostnames/FQDNs.

mykola-elastic:
## Summary

The MongoDB Atlas `alert` ingest pipeline grokked the host portion of
`response.hostnameAndPort` into `source.address` and then ran
`convert: type: ip` **in place** on that same field. This caused two problems:

- **Mapping corruption / indexing failures.** When the conversion succeeded,
  the document's `source.address` value became an `ip`-typed value. After
  data-stream rollover, `source.address` could be dynamically mapped as
  `ip`, and subsequent FQDN events were rejected with mapper/parsing errors
  (issue #18441).
- **`related.ip` polluted with hostnames.** `related.ip` was appended from
  `source.address` (gated on `related.hosts == null`), so its semantics
  depended on the side effect of the in-place convert rather than on whether
  the value was actually an IP.

Per ECS, `source.address` should stay as the raw value (hostname, FQDN, or
IP), parseable IPs belong in `source.ip`, and hostnames/FQDNs belong in
`source.domain`.

## Changes

- **`data_stream/alert/elasticsearch/ingest_pipeline/default.yml`** —
  Replaced the in-place `convert` + `related.ip` blocks with the
  ECS-aligned pattern (mirrors
  `packages/opencanary/data_stream/events/elasticsearch/ingest_pipeline/default.yml`):
  - `convert source.address → target_field: source.ip, type: ip`
  - `on_failure: set source.domain copy_from source.address`
  - `append related.ip` gated on `ctx.source?.ip != null`
  - `append related.hosts` gated on `ctx.source?.domain != null`
  - `source.address` is no longer mutated.

- **`data_stream/alert/_dev/test/pipeline/test-alert-logs.json`** —
  Added a second event with an IPv4 `hostnameAndPort` (`10.0.0.5:27017`) so
  the test suite exercises the IP branch (it was previously FQDN-only).

- **`data_stream/alert/_dev/test/pipeline/test-alert-logs.json-expected.json`** —
  Updated the FQDN expectation to include `source.domain`; added a new
  expectation for the IPv4 event asserting raw `source.address`,
  `source.ip`, and `related.ip` (and no `source.domain` / `related.hosts`).

## Behavior matrix

| Input `hostnameAndPort` | `source.address` | `source.ip` | `source.domain` | `related.ip` | `related.hosts` |
| --- | --- | --- | --- | --- | --- |
| `host.example.com:27017` | `host.example.com` | — | `host.example.com` | — | `[host.example.com]` |
| `10.0.0.5:27017` | `10.0.0.5` | `10.0.0.5` | — | `[10.0.0.5]` | — |

## Notes

- IPv6 inputs are still not handled by the existing
  `%{HOSTNAME:source.address}` grok pattern (it only matches IPv4 and
  hostnames) — that's a pre-existing limitation and out of scope for this
  bugfix.
- `fields/fields.yml` description for
  `mongodb_atlas.alert.host_name_and_port` is left unchanged.

## Test plan

- [x] `elastic-package test pipeline --data-streams alert -v` passes (FQDN
      and new IPv4 fixture).
- [x] `elastic-package check`, `elastic-package format`,
      `elastic-package lint` are clean.
- [x] Manual: ingest a synthetic alert with FQDN
      `hostnameAndPort` and confirm the destination index does not map
      `source.address` as `ip` (it should stay `keyword`).
- [x] Manual: ingest a synthetic alert with IPv4 `hostnameAndPort` and
      confirm `source.ip` is populated and `related.ip` contains the IP.

## Related issues

- Closes https://github.com/elastic/integrations/issues/18441

### DIFF
--- a/packages/mongodb_atlas/changelog.yml
+++ b/packages/mongodb_atlas/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.2.1"
+  changes:
+    - description: Fix `source.address` handling in alert pipeline to be ECS-aligned (write parsed IPs to `source.ip`, hostnames to `source.domain`).
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/1
 - version: "1.2.0"
   changes:
     - description: Enable Agentless deployment.
--- a/packages/mongodb_atlas/changelog.yml
+++ b/packages/mongodb_atlas/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Fix `source.address` handling in alert pipeline to be ECS-aligned (write parsed IPs to `source.ip`, hostnames to `source.domain`).
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/1
+      link: https://github.com/elastic/integrations/pull/18659
 - version: "1.2.0"
   changes:
     - description: Enable Agentless deployment.
--- a/packages/mongodb_atlas/data_stream/alert/_dev/test/pipeline/test-alert-logs.json
+++ b/packages/mongodb_atlas/data_stream/alert/_dev/test/pipeline/test-alert-logs.json
@@ -42,6 +42,49 @@
                     "threshold"
                 ]
             }
+        },
+        {
+            "response": {
+                "alertConfigId": "6683cb42d12516670f38a3ef",
+                "clusterName": "IntegrationsDevBuildCluster",
+                "created": "2024-07-02T09:55:24Z",
+                "currentValue": {
+                    "number": 1.2666666666666666,
+                    "units": "RAW"
+                },
+                "eventTypeName": "OUTSIDE_METRIC_THRESHOLD",
+                "groupId": "646f4379c47da356740d14ad",
+                "hostnameAndPort": "10.0.0.5:27017",
+                "id": "6683ce8c9558e8655626e1ee",
+                "lastNotified": "2024-07-02T10:03:21Z",
+                "metricName": "FTS_PROCESS_CPU_USER",
+                "orgId": "646f418c72f24c07d430aaca",
+                "replicaSetName": "atlas-ccx4uc-shard-0",
+                "resolved": "2024-07-02T10:03:21Z",
+                "status": "CLOSED",
+                "updated": "2024-07-02T10:03:21Z",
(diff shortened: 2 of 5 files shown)


stefans-elastic (approved):
LGTM

jakubgalecki0 (approved):

giorgi-imerlishvili-elastic (approved):

tommyers-elastic (approved):
lgtm

PR #18674, opened 2026-04-28
ci(docs): scope pull_request triggers to integration branches

Mpdreamz:
Scopes `docs-build` and `docs-preview-cleanup` `pull_request*` workflows to integration branches ([assembler.yml](https://github.com/elastic/docs-builder/blob/main/config/assembler.yml)).

Related: https://github.com/elastic/docs-eng-team/issues/569

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,6 +1,7 @@
 name: docs-build
 on:
   pull_request:
+    branches: [main]
     types: [opened, synchronize, reopened]
   push:
     branches: [main]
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-build.yml
```diff
@@ -1,6 +1,7 @@
 name: docs-build
 on:
   pull_request:
+    branches: [main]
```
mrodm, line 4: Should it be added also backport branches here ? 
PRs merged in `backport-*` will also publish new package versions.

mrodm: If docs are just generated based on main, maybe it is not needed.

PR #18678, opened 2026-04-28
[POC] Hostmetrics_otel package using new dashboard api

ishleenk17:
POC of a package using new dashboard api. 

This is just to POC how the package structure would look like if we use the new dashboard api structure.


ishleenk17:
some example dashboards are here. Note: Dashboard as api doesn't support link panels yet. 

Screenshot 2026-04-28 at 8 51 55 PM


Screenshot 2026-04-28 at 8 50 29 PM

PR #18747, opened 2026-04-30
[OTLP input otel] Remove signal toggle

teresaromero:
## Proposed commit message

The package was initially configured with signal toggles. For traces, spans-events also need logs to be enabled. This PR removes the toggle so the package supports the 3 signals

## Checklist

- [x] I have verified that all data streams collect metrics logs and traces.


packages/otlp_input_otel/_dev/test/system/test-only-traces-config.yml
```diff
@@ -3,6 +3,7 @@ service_notify_signal: SIGHUP
 vars:
   grpc_endpoint: 0.0.0.0:4317
   http_endpoint: 0.0.0.0:4318
+# the test service sends all signals, this test only checks for traces
 signal_types:
   - traces
```
mrodm, line 8: For traces, I think it should be added here metrics and logs, since there will be docs created in those data streams.

mrodm (approved):

PR #18748, opened 2026-04-30
[Network Beaconing Identification] Add process related fields to beaconing transform

sodhikirti07:
## Proposed commit message
Add process related fields to beaconing transform

## Checklist

## Author's Checklist

- [ ] 

- Tested using [ITP](https://github.com/elastic/security-ml/tree/main/src/integrations-test-pipeline)

## Related issues

- 

## Screenshots
Transform and detection alerts preview

image

image

### DIFF
--- a/packages/beaconing/elasticsearch/ingest_pipeline/ml_beaconing_ingest_pipeline.yml
+++ b/packages/beaconing/elasticsearch/ingest_pipeline/ml_beaconing_ingest_pipeline.yml
@@ -12,3 +12,50 @@ processors:
         - UNIX
         - ISO8601
       target_field: event.ingested
+  - script:
+      description: Casts flat underscore process fields in group-by to ECS nested types.
+      lang: painless
+      source: |
+        if (ctx.process_pid != null) {
+          String v = ctx.process_pid.toString();
+          if (!v.isEmpty()) {
+            if (ctx.process == null) { ctx.process = new HashMap(); }
+            ctx.process.pid = Long.parseLong(v);
+          }
+        }
+        if (ctx.process_parent_pid != null) {
+          String v = ctx.process_parent_pid.toString();
+          if (!v.isEmpty()) {
+            if (ctx.process == null) { ctx.process = new HashMap(); }
+            if (ctx.process.parent == null) { ctx.process.parent = new HashMap(); }
+            ctx.process.parent.pid = Long.parseLong(v);
+          }
+        }
+        if (ctx.process_code_signature_exists != null) {
+          String v = ctx.process_code_signature_exists.toString();
+          if (!v.isEmpty()) {
+            if (ctx.process == null) { ctx.process = new HashMap(); }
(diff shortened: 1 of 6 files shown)


jmcarlock:
Looks good. My suggestion is to make `process.executable` fallback to having an empty string like the other new fields to guarantee backward compatibility/use with other integrations. Otherwise it is required to run the transform.

jmcarlock (approved):

jmcarlock:
LGTM! Agree with the macroscopeapp, the field definitions also need updating

mattkime (approved):

packages/beaconing/elasticsearch/transform/pivot_transform/fields/fields.yml
```diff
@@ -1,9 +1,33 @@
...
+  name: process.code_signature.subject_name
+- external: ecs
+  name: process.hash.sha256
+- external: ecs
+  name: process.parent.hash.sha256
+- external: ecs
+  name: process.working_directory
+- external: ecs
+  name: process.pid
+- external: ecs
```
susan-shu-c, line 29: Have you tested what the output of using the ECS fields results in?
As mentioned the transform `tostring()` may go to text or keyword
But does using `external: ecs` automatically then handle the process to store them in `long`?
Just to be sure how it actually works, and where that translation is being done

sodhikirti07: Thank you Susan for mentioning this. I've added a processor in the ingest pipeline that will correctly cast the process related 4 fields to correct ECS type.

packages/beaconing/elasticsearch/ingest_pipeline/ml_beaconing_ingest_pipeline.yml
```diff
@@ -11,4 +11,51 @@ processors:
...
+            if (ctx.process == null) { ctx.process = new HashMap(); }
+            ctx.process.pid = Long.parseLong(v);
+          }
+        }
+        if (ctx.process_parent_pid != null) {
+          String v = ctx.process_parent_pid.toString();
+          if (!v.isEmpty()) {
+            if (ctx.process == null) { ctx.process = new HashMap(); }
+            if (ctx.process.parent == null) { ctx.process.parent = new HashMap(); }
+            ctx.process.parent.pid = Long.parseLong(v);
```
susan-shu-c, line 31: Is there any possibility of `ctx.process_parent_pid` missing and how would that be handled?

sodhikirti07: These fields can be missing, and that is handled by the null and empty checks in the ingest pipeline. If the value is missing, the entire block (e.g. lines 26–33 for process.parent.pid) is skipped, so the field is never written to the output document, it is simply do not exist from the transform's output.

jmcarlock (approved):

susan-shu-c (approved):
Thanks for addressing review/questions!

PR #18751, opened 2026-05-01
aws/config: fix null EvaluationResults crash in CEL program

kcreddy:
## Proposed commit message
```
aws/config: fix null EvaluationResults crash in CEL program

The GetComplianceDetailsByConfigRule AWS API can return
{"EvaluationResults": null} instead of an empty array when no
compliance results exist for a rule. The CEL program used
has(body.EvaluationResults) && size(body.EvaluationResults) > 0
to check for results, but has() returns true for null values and
size(null) has no overload in CEL, causing a runtime crash.

Add type(body.EvaluationResults) == list into the condition to 
safely handle both missing and null values.

Update the system test mock to return null EvaluationResults for
one config rule, covering this scenario.

```

## Checklist

Added system test scenario passes.

### DIFF
--- a/packages/aws/changelog.yml
+++ b/packages/aws/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "6.14.1"
+  changes:
+    - description: Fix CEL evaluation failure in AWS Config when GetComplianceDetailsByConfigRule returns null EvaluationResults.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/1
 - version: "6.14.0"
   changes:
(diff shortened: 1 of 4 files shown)


efd6:
This looks like a dup of #18787.

Except I'm not convinced by the type check. Formally, this change here is more correct, but I think the change there is easier to read with approximately no additional risk.

kcreddy:
> This looks like a dup of #18787, but you were here first.

I like that you were also fixing in other places. Lets go with your change. I like that one. Thanks for flagging, I thought I fixed it already last week :(

Closing as duplicate of https://github.com/elastic/integrations/pull/18787

PR #18758, opened 2026-05-01
Google Workspace: strip domain prefix from username in Chrome events

### LINKED ISSUE #18393: [google_workspace] Remove hostname prefix from `user.name` field in Chrome `content_transfer` events
# Description
In GWorkspace Chrome logs, `user.name` is sometimes populated with a `hostname\user.name` format rather than just the username. The hostname component should be mapped to `host.hostname` or `host.name` instead, keeping `user.name` clean and consistent with ECS expectations.

# Current behavior
In certain Chrome log events, most notably `event.action: content_transfer`, `user.name` is populated in the format `hostname\username` (ie **MYHOST\firstname.lastname**). This conflates host identity with user identity in a single field, making it difficult to filter, correlate, or alert on user activity across log sources.

# Requested behavior
When `user.name` contains a `hostname\username` pattern, the integration should map fields as 
```
"user": {
  "name": "firstname.lastname"
},
"host": {
  "hostname": "MYHOST"
}
```
Sample `event.original` is available upon request. 

# User context
This issue was raised internally. The hostname prefixed `user.name` value prevents consistent user correlation across log sources (similar to this CloudTrail `user.name`/ARN session name parsing issue: https://github.com/elastic/integrations/issues/18346)

chemamartinez:
## Proposed commit message

```
[google_workspace] Strip domain prefix from username in Chrome events

On Windows machines joined to an Active Directory domain, the OS reports
user identity as DOMAIN\username. The Google Workspace Reports API surfaces
this verbatim in the DEVICE_USER parameter (documented as "the user's name
as reported by the OS"), so composite values like MYHOST\firstname.lastname
were landing in user.name, breaking cross-source user correlation.

The pipeline now dissects the backslash format before the related.user
appends. For the DOMAIN\user form, the prefix maps to user.domain — the
closest ECS fit for the Windows auth domain. For the DOMAIN\user@dns.domain
form (a NetBIOS prefix stapled to a UPN), the prefix is discarded via a
_tmp field since it has no clean ECS home; user.name and user.domain are
extracted from the UPN parts instead. The raw composite value is preserved
in google_workspace.chrome.device_user for reference.

Adds two mocked CONTENT_TRANSFER pipeline test fixtures covering the
Windows domain-user scenarios (plain DOMAIN\user and DOMAIN\user@domain).
```

## Checklist

## Related issues

- Closes https://github.com/elastic/integrations/issues/18393


packages/google_workspace/data_stream/chrome/_dev/test/pipeline/test-chrome.log-expected.json
```diff
@@ -428,6 +430,209 @@
...
+                    "remote_ip": [
+                        "198.51.100.2"
+                    ],
+                    "timestamp": "2025-03-02T14:00:00.000Z",
+                    "trigger_type": "FILE_UPLOAD",
+                    "type": "CONTENT_TRANSFER_TYPE"
+                }
+            },
+            "host": {
+                "hostname": "CORP",
```
kcreddy: I'm not sure if this should be truly mapped to `host.hostname`.

I got Claude checking on this based on other examples in the repo. It agrees that the logic should be similar to how [`bitdefender`](https://github.com/elastic/integrations/blob/main/packages/bitdefender/data_stream/push_notifications/elasticsearch/ingest_pipeline/default.yml#L1033-L1041) handles it. WDYT?

---

> **PROMPT: In the real-world case, what would the pattern look like? In `CORP\jdoe@example.com`, is there any ECS field that "CORP" be mapped into because "user.name: jdoe" and "user.domain: example.com"**

>: 

> Good question. Let me think through what `CORP\jdoe@example.com` actually represents in the real world.
> 
> This is a hybrid identity — two different naming systems combined:
> 
> - `CORP` = the **NetBIOS domain name** (the legacy AD domain identifier)
> - `jdoe@example.com` = the **UPN (User Principal Name)** — a separate, email-style login
> 
> So the full value `CORP\jdoe@example.com` is a mashup: the NetBIOS `DOMAIN\` prefix stapled onto a UPN. This happens in Google Workspace Chrome reporting because the OS reports the user identity verbatim, and on domain-joined Windows machines you get the down-level logon format mixed with UPN.
> 
> The meaningful fields are:
> 
> | Component | What it is | Where it could go |
> |-----------|-----------|-------------------|
> | `CORP` | NetBIOS domain name | Nowhere clean in ECS. It's *not* a hostname, *not* the DNS domain. It's the short AD domain. |
> | `jdoe` | Username | `user.name` |
> | `example.com` | UPN suffix / DNS domain | `user.domain` |
> 
> The NetBIOS domain name (`CORP`) doesn't map neatly to any ECS field. It's not `host.hostname` (that's the machine). It's not `user.domain` (that's the DNS domain, which is `example.com` from the UPN). It's not `host.domain` either (that's already set from `owner_domain` in this pipeline).
> 
> This is exactly why the bitdefender integration maps it to `_tmp.user_leading_domain` — it extracts it just to get past it, then discards it in cleanup. The NetBIOS domain is useful for the raw value (preserved in `google_workspace.chrome.device_user`) but doesn't belong in any ECS field.
> 
> For the simpler case `MYHOST\firstname.lastname` (no `@`), the prefix *might* be a hostname or might be a NetBIOS domain — you can't tell from the string alone. The bitdefender pattern handles this by mapping the prefix to `user.domain` in the no-`@` case (`%{DATA:user.domain}\\%{GREEDYDATA:user.name}`), which treats it as the Windows domain the user authenticated against. That's a reasonable pragmatic choice, even if `MYHOST` isn't technically a "domain" — it's the local machine's authority, which is the closest fit.
> 
> So the recommended pattern ordering would be:
> 
> ```yaml
> # 1. DOMAIN\user@dns.domain — discard NetBIOS prefix, keep user and DNS domain
> - '%{DATA:_tmp.user_leading_domain}\\%{DATA:user.name}@%{GREEDYDATA:user.domain}'
> 
> # 2. DOMAIN\user — prefix is the auth domain, no DNS domain present
> - '%{DATA:user.domain}\\%{GREEDYDATA:user.name}'
> 
> # 3. user@domain — plain UPN/email
> - '%{DATA:user.name}@%{GREEDYDATA:user.domain}'
> 
> # 4. bare username — no domain info at all
> - '%{GREEDYDATA:user.name}'
> ```
> 
> The key takeaway: `CORP` in the real-world case is a short AD identifier that has no clean ECS home. Discard it to `_tmp` and let the raw value in `google_workspace.chrome.device_user` preserve it for anyone who needs it.

chemamartinez: You're right, good catch. CORP is a NetBIOS domain name and doesn't belong in host.hostname. Applying the bitdefender pattern makes sense to me: the leading domain goes to _tmp (discarded) for the `DOMAIN\user@dns.domain` form, and to user.domain for the plain `DOMAIN\user` form.

Added in https://github.com/elastic/integrations/pull/18758/commits/137ff3b8a99910c8801ffa5eca95be6365c690c3

kcreddy (approved):

PR #18822, opened 2026-05-05
[azure_metrics] Fix compute_vm stream YAML indentation

zmoog:
## Summary

Multi-line flow sequence (`[...]`) continuations in `compute_vm/agent/stream/stream.yml.hbs` were indented to column 5, while the parent `- name:` list item is at column 9. Older YAML parsers (go-yaml v2/v3, js-yaml, PyYAML) silently accept this, but Kibana 9.5+ uses [`eemeli/yaml`](https://github.com/eemeli/yaml) which strictly enforces YAML 1.2 rule [9C9N](https://matrix.yaml.info/details/9C9N.html). On 9.5.0 the integration fails to install with:

```
Flow sequence in block collection must be sufficiently indented and end with a ]
at line N, column 5
```

This PR re-indents the affected continuations to column 13, matching the convention already used in the `resource_ids` and `unless` branches of the same template. Confirmed with `eemeli/yaml`:
- before fix → reproduces the error verbatim;
- after fix → parses cleanly.

Affected lines: 40, 45–56, 107–118, 169–180 in `compute_vm/agent/stream/stream.yml.hbs`. Whitespace-only changes (no metric names added/removed).

## Test plan

- [ ] `elastic-package check` passes for `packages/azure_metrics`.
- [ ] Install `azure_metrics` (Virtual Machines policy template) on Elastic Cloud 9.5.0 — installation succeeds.
- [ ] Verify VM metrics flow end-to-end (Windows GuestMetrics + Linux GuestMetrics).


MichaelKatsoulis (approved):
LGTM

PR #18852, opened 2026-05-06
[cloudflare_logpush] fix breaking-change bugs

brijesh-elastic:
## Proposed commit message

```
cloudflare_logpush: fix breaking-change bugs

Fix several field mapping and ingest pipeline correctness bugs across data streams:
- Fix `cloudflare_logpush.gateway_http.quarantined` mapping by changing its type from
`keyword` to `boolean` to match the actual values returned by Cloudflare.
- Fix `http_request` HTTP header field naming by renaming `header` to `headers` to align
with the field definitions and the underlying Cloudflare payload.
- Fix `gateway_dns` resolved IP modeling by replacing the flat
`cloudflare_logpush.gateway_dns.resolved_ip` field with the nested
`cloudflare_logpush.gateway_dns.resolved_ip_details.ips`, so per-IP category
metadata can be captured alongside the IP values.
- Fix `dlp_forensic_copies` `preserve_duplicate_custom_fields` toggle so that the correct
custom fields (`triggered_rule_id`, `datetime`) — which have ECS counterparts — are
removed when the toggle is OFF. Previously the remove processor referenced fields
the pipeline never produced, so the toggle had no effect for these fields.
- Ensure identifier fields (`*ID`/`*IDs`) are consistently mapped as `keyword` and stored
as strings by replacing `rename` (and integer `convert`) processors with `convert`
processors of `type: string`. Fields whose mapping type changes in this release are:
`cloudflare_logpush.gateway_dns.application_id`,
`cloudflare_logpush.gateway_dns.matched.category.ids`,
`cloudflare_logpush.gateway_dns.question.category.ids`, and
`cloudflare_logpush.gateway_dns.question.indicator_feed.ids` (`long` → `keyword`);
`cloudflare_logpush.http_request.bot.detection_ids`,
`cloudflare_logpush.http_request.edge.rate.limit.id`, and
`cloudflare_logpush.http_request.zone.id` (`long` → `keyword`);
`cloudflare_logpush.magic_ids.signature.id` (`long` → `keyword`); and
`cloudflare_logpush.network_analytics.rule.id` and
`cloudflare_logpush.network_analytics.rule.set.override.id` (`text` → `keyword`).
The already-`keyword` `*IDs` fields in `gateway_dns`, `gateway_http`, and
`gateway_network` continue to be cast to strings so their values are stored
consistently rather than as the raw JSON integers Cloudflare emits.
- Fix `dns.response_code` in the `dns_firewall` and `gateway_dns` data streams to
correctly populate IANA keyword values instead of numeric representations.
- Change the `cloudflare_logpush.network_analytics.tcp.options` and
`cloudflare_logpush.network_analytics.ipv6.extension_headers` field types from `text`
to `long`, matching Cloudflare's documented semantics (consistent with
`cloudflare_logpush.network_analytics.ipv4.options`). Their comma-separated values
continue to be parsed into numeric arrays.

The fixes above are user-visible breaking changes:
- `cloudflare_logpush.gateway_http.quarantined` mapping changes from `keyword` to `boolean`.
Existing indices will reject the new value type, so users must roll over the
data stream (or reindex) for the new mapping to take effect.
- The `http_request` HTTP header fields are renamed from `header` to `headers`. Saved searches,
dashboards, queries, and detection rules referencing `cloudflare_logpush.http_request.*.header`
will no longer return data and must be updated to use `headers`.
- The `cloudflare_logpush.gateway_dns.resolved_ip` field is removed; resolved IPs now live under
`cloudflare_logpush.gateway_dns.resolved_ip_details.ips`. Saved searches, dashboards, queries,
and detection rules referencing the old field must be updated.
- In the `dlp_forensic_copies` data stream, when `preserve_duplicate_custom_fields` is
OFF (the default), `cloudflare_logpush.dlp_forensic_copies.triggered_rule_id` and
`cloudflare_logpush.dlp_forensic_copies.datetime` are no longer present in documents.
Users that depend on these custom fields must turn `preserve_duplicate_custom_fields`
ON, or migrate to the corresponding ECS fields.
- Identifier fields (`*ID`/`*IDs`) are now indexed as `keyword` strings rather than the raw
JSON integers Cloudflare emits (or `text`). The following fields change mapping type:
`cloudflare_logpush.gateway_dns.application_id`,
`cloudflare_logpush.gateway_dns.matched.category.ids`,
`cloudflare_logpush.gateway_dns.question.category.ids`,
`cloudflare_logpush.gateway_dns.question.indicator_feed.ids`,
`cloudflare_logpush.http_request.bot.detection_ids`,
`cloudflare_logpush.http_request.edge.rate.limit.id`,
`cloudflare_logpush.http_request.zone.id`, `cloudflare_logpush.magic_ids.signature.id`,
`cloudflare_logpush.network_analytics.rule.id`, and
`cloudflare_logpush.network_analytics.rule.set.override.id`. In addition, the
already-`keyword` `*IDs` fields in `gateway_dns`, `gateway_http`, and `gateway_network`
now contain string values instead of integers. Existing indices will reject the changed
value types, so users must roll over the data stream (or reindex), and queries, dashboards,
and detection rules comparing these fields against numeric values must be updated to use
string values.
- `dns.response_code` in the `dns_firewall` and `gateway_dns` data streams now contains IANA
keyword values instead of numeric representations. Queries, dashboards, and detection rules
referencing this field must be updated to use the new keyword values.
- The `cloudflare_logpush.network_analytics.tcp.options` and
`cloudflare_logpush.network_analytics.ipv6.extension_headers` fields change from `text` to
`long`. Existing indices will reject the new value type, so users must roll over the data
stream (or reindex) for the new mapping to take effect.

```

## Checklist

- Clone integrations repo.

> elastic-package test -v


packages/cloudflare_logpush/data_stream/dns_firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -336,17 +336,44 @@ processors:
...
-      copy_from: cloudflare_logpush.dns_firewall.upstream.response_code
-      ignore_empty_value: true
-      if: ctx.cloudflare_logpush?.dns_firewall?.response?.code == null
-      tag: set_dns_upstream_response_code
+  - script:
+      lang: painless
+      tag: script_map_dns_response_code
+      params:
+        "0": NoError
+        "1": FormErr
```
kcreddy, line 344: We may need similar handling in [`gateway_dns`](https://github.com/elastic/integrations/blob/main/packages/cloudflare_logpush/data_stream/gateway_dns/elasticsearch/ingest_pipeline/default.yml#L239-L243) data stream.

packages/cloudflare_logpush/data_stream/gateway_dns/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -591,6 +583,22 @@ processors:
       field: json.ResolvedIPCountryCodes
       target_field: cloudflare_logpush.gateway_dns.resolved_ip_details.country_codes
       ignore_missing: true
+  - convert:
+      tag: convert_json_ResolvedIPs_to_cloudflare_logpush_gateway_dns_resolved_ip_details_ips_d1c23f97
+      field: json.ResolvedIPs
+      target_field: cloudflare_logpush.gateway_dns.resolved_ip_details.ips
+      type: ip
+      ignore_missing: true
```
kcreddy, line 621: Do we also need to guard against `!= ''`? [Example](https://github.com/elastic/integrations/blob/main/packages/cloudflare_logpush/data_stream/gateway_dns/elasticsearch/ingest_pipeline/default.yml#L634-L645)

packages/cloudflare_logpush/data_stream/http_request/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -721,14 +722,8 @@ processors:
       tag: convert_json_EdgeRateLimitID_to_cloudflare_logpush_http_request_edge_rate_limit_id_6b98f73a
       field: json.EdgeRateLimitID
       target_field: cloudflare_logpush.http_request.edge.rate.limit.id
-      if: ctx.json?.EdgeRateLimitID != ''
```
kcreddy, line 724: Do we need to keep this guard?

packages/cloudflare_logpush/data_stream/http_request/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -993,12 +988,12 @@ processors:
...
       tag: rename_json_RequestHeaders_to_cloudflare_logpush_http_request_request_header_88514ca5
       field: json.RequestHeaders
-      target_field: cloudflare_logpush.http_request.request.header
+      target_field: cloudflare_logpush.http_request.request.headers
       ignore_missing: true
   - rename:
       tag: rename_json_ResponseHeaders_to_cloudflare_logpush_http_request_response_header_a8900b0d
       field: json.ResponseHeaders
-      target_field: cloudflare_logpush.http_request.response.header
+      target_field: cloudflare_logpush.http_request.response.headers
```
kcreddy, line 996: Change the `tag` as well for consistency.

kcreddy (commented):
Looks like changelog and commit message has some differences.
Commit message doesn't describe these changes, can you please verify:
> It omits 3 areas (dns.response_code, network_analytics text→long, and the magic_ids/network_analytics/http_request identifier type changes)

packages/cloudflare_logpush/data_stream/http_request/elasticsearch/ingest_pipeline/default.yml
brijesh-elastic: No, Since we're converting it to string, it will never fail the conversion.

packages/cloudflare_logpush/data_stream/gateway_dns/elasticsearch/ingest_pipeline/default.yml
brijesh-elastic: I think no. As, `ResolvedIPs` is an array. No need to check against `!= ''` condition.

kcreddy (approved):
LGTM

leandrojmp:
Hello @kcreddy and @brijesh-elastic 

Just saw the breaking change message in Fleet while upgrading some integrations and while it explain what will happen, I don't think it correctly reflects the breaking changes issues, mostly this part:

> Existing indices will reject the changed
value types, so users must roll over the data stream (or reindex), and **queries, dashboards,
and detection rules comparing these fields against numeric values must be updated to use
string values.**

Rolling over the data streams would allow the fields where the mapping was changed to be indexed, but it will introduce a Kibana mapping conflict for those fields and as far as I know fields with conflict in Kibana cannot be used in Dashboards and some Security rule, it breaks them.

So all those fields cannot be used in some Security Rules/Dashboards until the Mapping Conflict in Kibana persists.

I think KQL rules may still work, but ES|QL rules breaks if the same field has conflicting mapping in the indices matching the query range.

kcreddy:
Thanks @leandrojmp, you're right, and the wording in that commit message is misleading. Two corrections to it.

**Rolling over isn't something you need to do, and it isn't the remediation.** Fleet should already do it, i.e., a lazy rollover on the affected data streams itself during the upgrade when it encounters mismatched type (different family). So "users must roll over the data stream" is misleading.

**The conflict lasts as long as the old backing indices do.** A field's type can't be changed in an index that already exists, so every backing index written before the upgrade keeps the old mapping, and any data view matching them reports those fields as `conflict` until the last one is reindexed or deleted by ILM. Your read on the impact is mostly accurate: Lens rejects conflict-typed fields, they aren't offered in the filter bar, and they aren't sortable. Discover still renders the values and KQL still runs.

So the actual choice for affected users is to wait for retention to age the old indices out, or reindex them. [Reindexing data streams due to mapping conflicts](https://www.elastic.co/search-labs/blog/elasticsearch-mapping-conflicts-reindex-data-streams) walks through the latter. One caveat if you follow it: skip the step that adds the corrected mapping to the `@custom` component template. That step is for cases where the package's own mapping is wrong, which isn't the case here, and the override would only have to be unwound later.

The notice should have said to expect a mapping conflict on existing data until it's reindexed or ages out, rather than implying the rollover resolves it. We'll word it that way for future type changes.

cc: @brijesh-elastic

leandrojmp:
Thanks @kcreddy 

I think that a bigger problem is that sometimes breaking changes are being introduced without not much improvement for the users, in this case for example, some fields are changing mapping just to align with the recommendation to have ID fields mapped as keywords instead of numeric types.

While I understand that this is a nice change to correct a mapping mistake, I'm not sure that the problems the change introduces justify it.

Reindex in theory is a solution, but it does not scale, in my case I have close to 500 GB of Cloudflare per day, reindex is not an option.

Maybe leaving the mapping as they are, specially if it something that already exists for a long time, or create a new field with the new mapping, would be better.

There is an old issue to discuss some alternatives to those breaking changes, how to communicate them, how to reduce their occurrence, but unforunately there was not much activity on it, it is this one #13861, would be nice if this discussion improved.

For example, at the moment I have 3 major integrations (AWS, CrowdStrike and Cloudflare) that I paused updated because all 3 are introducing breaking changes that require a validation of all security rules and dashboards for those indices, which takes a lot of extra time.

PR #18856, opened 2026-05-06
[chore][internal] Add build & read access to buildkite pipelines for members of integration-experience team

Niceplace:
## Proposed commit message

For members (i.e me) of the [integration-experience Github team](https://github.com/orgs/elastic/teams/integration-experience), pipeline access is limited i.e. read-only, as defined by [this general setting](https://github.com/elastic/integrations/blob/main/catalog-info.yaml#L108-L109). [As advised by the platform engineering team](https://elastic.slack.com/archives/C01DJ3DC0AX/p1778090913662489?thread_ts=1778090687.438969&cid=C01DJ3DC0AX), I'm adding build & read permissions to the integration-experience team so we can all read & execute builds.

The main impact here is that if a PR is submitted to a repository we steward, some of our team members would not have access to trigger builds from Buildkite's UI. It would be strictly readonly. This might be helpful in situations where flaky tests occur and we need to re-trigger things (although it would be preferable to fix the problem at the source instead)

Here is the platform engineering documentation reference: https://docs.elastic.dev/ci/getting-started-with-buildkite-at-elastic#2-adjust-your-pipeline-settings, more specifically the `Assigning a Pipeline to Teams` collapsible section (which I cannot provide a URL to unfortunately)

It's possible we have other repositories impacted by this but I'm seeing if this will fix it here and will open PRs in other repositories we steward afterwards.

N/A


catalog-info.yaml
```diff
@@ -105,6 +105,8 @@ spec:
           access_level: BUILD_AND_READ
         sec-windows-platform:
           access_level: BUILD_AND_READ
+        integration-experience:
+          access_level: BUILD_AND_READ
```
mrodm: Nit.
To try to keep all the teams using BUILD_AND_READ sorted, could you move this entry after the `logstash` team entry? Around line 80.

mrodm (approved):
LGTM.
Just a minor comment about inserting this following the sorted order for BUILD_AND_READ permission.
TIA!!

catalog-info.yaml
bhapas: before `logstash` probably

bhapas (approved):
LGTM.. Just resolve the comment from Mario before push.

catalog-info.yaml
Niceplace: @mrodm Thanks for pointing that out, I agree that it's good to keep things tidy. Will fix !

PR #18880, opened 2026-05-07
[Azure Logs] Add dedicated aadgraphactivitylogs data stream

### LINKED ISSUE #18879: [Azure]: Extend Support to AAD Graph Activity Logs
### Integration Name

Azure Logs [packages/azure]

### Dataset Name

azure.aadgraphactivitylogs (new)

### Integration Version

1.36.1

### Agent Version

9.3.0

macOS 26.4.1 

### User Goal

Ingest the `AzureADGraphActivityLogs` diagnostic category from Microsoft Entra ID into a dedicated, schema-aware data stream so detection rules and dashboards can target legacy Azure AD Graph (`graph.windows.net`) activity directly alongside the existing first-class support for `MicrosoftGraphActivityLogs`.

### Existing Features

`azure.graphactivitylogs` data stream parses the modern `MicrosoftGraphActivityLogs` diagnostic category. There is no equivalent for the sibling `AzureADGraphActivityLogs` category that Microsoft Entra ID also emits on the same diagnostic settings.

### What did you see?

`AzureADGraphActivityLogs` events are emitted by Microsoft Entra ID, reach the Elastic Agent through the configured Event Hub, and ARE indexed but they fall through to the `logs-azure.platformlogs-default` catch-all dataset because the integration's events router has no specific case for this category.

In `platformlogs`, events are stored with only generic platform-log parsing: the AAD-Graph-specific properties (`requestUri`, `appId`, `userAgent`, `actorType`, `scopes`, `sessionId`, etc.) survive only inside the raw `event.original` JSON. Detection rules and dashboards cannot target the schema cleanly without falling back to full-text search.

Observed in our research cluster, 24-hour window:

| Dataset | Count | Notes |
|---|---|---|
| `logs-azure.graphactivitylogs-default` | ~570 | `MicrosoftGraphActivityLogs` from this tenant — correctly routed (control) |
| `logs-azure.aadgraphactivitylogs-default` | (does not exist) | no dataset defined |
| `logs-azure.platformlogs-default` filtered to `AzureADGraphActivityLogs` | 28 | fallback bucket — mix of test traffic and background tenant activity |

The traffic in the fallback bucket includes calls from Microsoft's own first-party Azure Portal admin experience (appId `74658136-14ec-4630-ad9b-26e160ff0fc6`, `env_cloud_role: restdirectoryservice`) hitting endpoints like `/v2/myorganization/isDirectoryFeatureEnabled` and `/{tenantId}/$batch` > i.e., normal admin sessions in the Entra Portal still produce AAD Graph activity. Even Microsoft hasn't fully migrated, so this isn't a niche legacy concern.

#### Why this matters

Legacy Azure AD Graph (`graph.windows.net`) is deprecated by Microsoft but still in active use across production tenants in three concrete ways:

1. Microsoft's own first-party tooling. See the Azure Portal traffic noted above.
2. Older line-of-business apps and third-party SaaS connectors that haven't been ported to Microsoft Graph.
3. Adversary tooling. ROADtools (`roadrecon`, `roadtx`), AzureHound v1, AADInternals, PingCastle, and Stormspotter all default to legacy AAD Graph endpoints. Microsoft introduced the `AzureADGraphActivityLogs` diagnostic category specifically so defenders can detect that activity — but the visibility only translates into actionable detection if the events are properly parsed.

References:
- [Azure Monitor Logs reference - `AADGraphActivityLogs` table](https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/aadgraphactivitylogs) (canonical schema reference; describes the table as *"AAD Graph Activity Logs provide details of legacy API requests made to Azure Active Directory Graph for resources in the tenant"*)
- [Logs available for streaming from Microsoft Entra ID](https://learn.microsoft.com/en-us/entra/identity/monitoring-health/concept-diagnostic-settings-logs-options) (diagnostic-setting context - note as of writing this page does not yet enumerate `AzureADGraphActivityLogs` by name even though the diagnostic category is exposed in the portal)

#### Current behavior — code references

Tested against `packages/azure` v1.36.1 with the v2 (router-based) processor architecture.

1. The events ingest pipeline at `data_stream/events/elasticsearch/ingest_pipeline/default.yml` maps `routing.category` to `event.dataset`. There is no branch for `AzureADGraphActivityLogs`, so the event matches only the generic fallback (`set: event.dataset = azure.platformlogs if routing.category != null`).
2. `data_stream/events/routing_rules.yml` reroutes `azure.platformlogs`-tagged events to `logs-azure.platformlogs-default`. There is no `azure.aadgraphactivitylogs` target.
3. There is no `data_stream/aadgraphactivitylogs/` directory — no manifest, no field mappings, no ingest pipeline, no agent stream config.
4. As a result, AAD Graph events land in `logs-azure.platformlogs-default` indexed with the generic platform-log mappings only.

#### Reproduction steps

1. In a Microsoft Entra tenant, configure a diagnostic setting that exports `AzureADGraphActivityLogs` to an Event Hub. (Optionally include `MicrosoftGraphActivityLogs` on the same setting as a control — those events arrive in Elastic correctly today.)
2. Generate authenticated traffic against `graph.windows.net`:
   ```bash
   TOKEN=$(az account get-access-token --resource https://graph.windows.net --query accessToken -o tsv)
   TID=<your-tenant-id>
   curl -sS -H "Authorization: Bearer $TOKEN" \
        -H "User-Agent: aadgraph-repro/1.0" \
        "https://graph.windows.net/$TID/users?api-version=1.6&\$top=5"
   ```
3. Confirm the events appear on the Azure side: query `AADGraphActivityLogs` in your Log Analytics workspace (note the table name drops the `Azure` prefix from the diagnostic-setting category).
4. In Elastic, observe events landing in `logs-azure.platformlogs-default` without AAD-Graph-specific field extraction:
   ```esql
   FROM logs-azure.platformlogs-default
   | WHERE @timestamp > NOW() - 1 hour
     AND event.original LIKE "*AzureADGraphActivityLogs*"
     AND event.original LIKE "*aadgraph-repro/1.0*"
   | KEEP @timestamp, data_stream.dataset, event.original
   | LIMIT 10
   ```
5. Confirm the dedicated dataset doesn't exist:
   ```esql
   FROM logs-azure.aadgraphactivitylogs-default
   | LIMIT 1
   ```
   Returns "Failed to resolve table".

#### Proposed solution

Add a dedicated `azure.aadgraphactivitylogs` data stream parallel to `azure.graphactivitylogs`. The schemas are similar but not identical (see schema differences below) so a separate dataset is cleaner than overloading `graphactivitylogs`.

Required changes:

1. New data stream: `packages/azure/data_stream/aadgraphactivitylogs/` with manifest, agent stream config, ingest pipeline (modeled on `graphactivitylogs/default.yml`), fields, sample event, and pipeline test fixtures.
2. Events router (`data_stream/events/elasticsearch/ingest_pipeline/default.yml`):
   ```yaml
   - set:
       field: event.dataset
       value: azure.aadgraphactivitylogs
       if: 'ctx.routing?.category == "AzureADGraphActivityLogs"'
   ```
3. Reroute rule (`data_stream/events/routing_rules.yml`):
   ```yaml
   - target_dataset: azure.aadgraphactivitylogs
     if: ctx.event?.dataset == 'azure.aadgraphactivitylogs'
     namespace:
       - "{{data_stream.namespace}}"
       - default
   ```

#### Schema differences vs. `MicrosoftGraphActivityLogs`

The Azure resource log envelope is identical, but inside `properties` there are differences worth handling explicitly:

| Field | MS Graph | AAD Graph | Notes |
|---|---|---|---|
| `properties.actorType` | absent | `User` / `ServicePrincipal` / etc. | needs `azure.aadgraphactivitylogs.properties.actor_type` |
| `properties.deviceId` (top level) | uses `cDeviceId` | uses flat `deviceId` | pipeline's `device.id` set should `copy_from properties.device_id`, not `c_device_id` |
| `properties.sessionId` | uses `cSid` | uses flat `sessionId` | same — `session.id` set from `properties.session_id` |
| `properties.requestUri` | absolute URL `https://graph.microsoft.com/v1.0/...` | relative path `/v2/{tenantId}/...` | the `uri_parts` processor needs to handle a relative URI shape |
| `properties.scopes` | space-separated friendly names | space-separated GUIDs (e.g. `62e90394-...`) | same `split` logic works, value semantics differ |
| `operationName` | `Microsoft Graph Activity` | `AAD Graph Activity` | always literal — not useful for detection alone |
| `category` | `MicrosoftGraphActivityLogs` | `AzureADGraphActivityLogs` | the routing key |
| `cloud.service.name` (set in pipeline) | `Microsoft Graph` | `Azure AD Graph` | distinguishes datasets in queries |

A real captured event for reference (UA / IP / IDs redacted; from a research tenant):

```json
{
  "Level": 4,
  "callerIpAddress": "<ip>",
  "category": "AzureADGraphActivityLogs",
  "correlationId": "<request-id>",
  "durationMs": 59,
  "operationName": "AAD Graph Activity",
  "operationVersion": "1.6",
  "properties": {
    "actorType": "User",
    "apiVersion": "1.6",
    "appId": "04b07795-8ddb-461a-bbee-02f9e1bf7b46",
    "clientAuthMethod": "0",
    "durationMs": 59,
    "identityProvider": "https://sts.windows.net/{tid}/",
    "ipAddress": "<ip>",
    "operationId": "<request-id>",
    "requestId": "<request-id>",
    "requestMethod": "GET",
    "requestUri": "/v2/{tenantId}/users",
    "responseSizeBytes": 54662,
    "responseStatusCode": 200,
    "roles": "",
    "scopes": "62e90394-69f5-4237-9190-012177145e10",
    "servicePrincipalId": "",
    "sessionId": "<session-id>",
    "signInActivityId": "<sign-in-id>",
    "tenantId": "<tid>",
    "timeGenerated": "2026-05-07T15:19:33.5368860Z",
    "tokenIssuedAt": "2026-05-07T13:50:39Z",
    "userAgent": "<ua>",
    "userId": "<uid>",
    "wids": ""
  },
  "resourceId": "/TENANTS/<TID>/PROVIDERS/MICROSOFT.AADIAM",
  "resultSignature": "200",
  "tenantId": "<tid>",
  "time": "2026-05-07T15:19:33.5368860Z"
}
```

#### Migration consideration

Existing tenants that already have `AzureADGraphActivityLogs` flowing to the integration today have those events in `logs-azure.platformlogs-default`. After this change, new events will land in `logs-azure.aadgraphactivitylogs-default`. The previously-indexed events in `platformlogs` won't be backfilled. Worth a one-line note in the integration docs / changelog so customers know to query both indices for a transition window.

#### Local prototype

I've prototyped this fix locally against the current `main` and validated it end-to-end:

- New `data_stream/aadgraphactivitylogs/` with the changes outlined above
- Edits to `data_stream/events/elasticsearch/ingest_pipeline/default.yml` and `data_stream/events/routing_rules.yml`
- Pipeline test fixtures (`test-aadgraph-activity.log`) using 4 real captured events covering `/users` 200, `/directoryRoles` 400, `/tenantDetails` 200, `/applications` 200
- `elastic-package check` passes; `elastic-package test pipeline -d aadgraphactivitylogs` passes both the warnings test and the deterministic field-comparison test

The pipeline produces ECS-compliant output that's directly usable for detection rules:

| ECS field | Source / derivation |
|---|---|
| `event.action` | derived from HTTP method + collection in `requestUri` (e.g. `users-read`, `applications-update`, `oauth2-permission-grants-execute`, `batch-execute`) — the raw `operationName` is the literal string `"AAD Graph Activity"` and isn't useful for detection on its own |
| `event.outcome` | `success` / `failure` / `unknown` from `responseStatusCode` (2xx / 4xx-5xx / 3xx) |
| `event.category` | `["iam", "web"]` |
| `event.type` | `["access", "info"]` for GET/HEAD; `["access", "change"]` for POST/PATCH/PUT; `["access", "deletion"]` for DELETE |
| `event.id` | `properties.requestId` (UUID) |
| `event.duration` | `properties.durationMs` * 1,000,000 (ECS nanoseconds) |
| `source.ip`, `client.ip`, `related.ip[]` | from `callerIpAddress`, with `source.geo.*` / `client.geo.*` / `source.as.*` enrichment |
| `user.id`, `client.user.id`, `related.user[]` | `properties.userId` (or falls back to `servicePrincipalId` when an SP is acting alone) |
| `service.id` | `properties.servicePrincipalId` |
| `session.id` | `properties.sessionId` (AAD Graph schema; differs from MS Graph's nested `c_sid`) |
| `device.id` | `properties.deviceId` (AAD Graph schema; differs from MS Graph's `c_device_id`) |
| `http.request.method`, `http.response.status_code`, `http.response.bytes` | from `properties.*` |
| `url.original`, `url.path`, `url.scheme`, etc. | `uri_parts` on `properties.requestUri` |
| `user_agent.original`, `user_agent.name`, `user_agent.os.*` | from `properties.userAgent` |
| `cloud.account.id` | `properties.tenantId` |
| `cloud.service.name` | `"Azure AD Graph"` (distinguishes from MS Graph in cross-dataset queries) |
| `azure.tenant_id`, `azure.correlation_id`, `azure.resource.id` / `.provider` | from envelope |
| `azure.aadgraphactivitylogs.properties.app_id`, `.actor_type`, `.scopes[]`, `.roles[]`, `.wids[]`, `.identity_provider`, `.client_auth_method`, `.token_issued_at`, `.sign_in_activity_id` | kept in dataset namespace (no clean ECS equivalent) |

Two design choices worth flagging:

1. `event.action` derivation. A small painless script extracts the directory collection name from `requestUri` (handling both `/v2/{tenantId}/{collection}` and `/v2/myorganization/{collection}` shapes), CamelCase-splits it (`servicePrincipals` → `service-principals`), and combines with a verb derived from the HTTP method (`GET`/`HEAD` → `read`, `POST` → `execute`, `PATCH`/`PUT` → `update`, `DELETE` → `delete`). Falls back to the generic `operationName` if the URI doesn't match.
2. `related.user[]` correlation. The `properties.app_id` is appended to `related.user` alongside `user.id` so a single ECS pivot finds events whether the AppId appears in the calling-user position (SP acting alone) or as the OAuth client context (user driving an app). Useful for both bulk-recon and credential-add detection rules. ECS doesn't have a clean canonical field for "OAuth client id" — sibling integrations (AWS CloudTrail's `aws.cloudtrail.user_identity.*`, Okta's `okta.actor.*`) keep these in the dataset namespace, which is the same choice we made here.

Happy to open a PR with this implementation if there's interest. Wanted to file the issue first to align on the dataset name and on whether maintainers prefer a dedicated dataset vs. extending `azure.graphactivitylogs` to handle both categories.

#### Open questions for maintainers

1. Dedicated dataset vs. shared with `graphactivitylogs`? The schemas overlap ~80% but the `requestUri` shape, `actorType` field, and semantics differ. Dedicated feels cleaner; happy to be talked out of it.
2. Naming. `azure.aadgraphactivitylogs` follows the naming pattern (concatenated, lowercase) used by `graphactivitylogs`. Acceptable, or prefer `azure.aad_graph_activity_logs` / `azure.entra_id_graph_legacy` / similar?
3. Migration story for already-ingested events. Existing tenants have AAD Graph events sitting in `logs-azure.platformlogs-default`. After the fix, new events go to the new dataset. Worth a one-line release-note callout, or do you want a back-compat reroute that picks up historical events too?
4. Deprecation banner in the docs. Microsoft is sunsetting AAD Graph; the diagnostic category itself may eventually be retired but there is no true indication of when. Should the dataset's docs include a deprecation banner pointing users to migrate workloads to MS Graph?

terrancedejesus:
Adds the `azure.aadgraphactivitylogs` data stream to ingest the `AzureADGraphActivityLogs` diagnostic category from Microsoft Entra ID, parallel to `azure.graphactivitylogs` for Microsoft Graph. Without this, AAD Graph events fall through to `azure.platformlogs` and the AAD-Graph-specific properties survive only inside `event.original`.

## Proposed commit message

```
azure: add aadgraphactivitylogs data stream

Add a dedicated data stream for the AzureADGraphActivityLogs
diagnostic category from Microsoft Entra ID. Without this,
legacy Azure AD Graph (graph.windows.net) events fall through
to the platformlogs catch-all and lose schema-aware parsing.

The events router maps routing.category ==
"AzureADGraphActivityLogs" to the new dataset. The ingest
pipeline extracts ECS fields: event.action from HTTP method +
URI collection, event.outcome from response status,
event.category [iam, web], and related.user including the
OAuth app_id for client correlation.

Legacy AAD Graph is still actively used by Microsoft first-party
tooling, older line-of-business apps, and adversary tooling
(ROADtools, AzureHound v1, AADInternals). The dedicated dataset
makes these events available for detection rules and dashboards.
```

## Checklist

- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) — *N/A, no dashboards added.*

## Author's Checklist

- [x] PII redacted from pipeline test fixtures (tenant IDs, user/session/request UUIDs, sign-in activity IDs, internal user-agents). Synthetic test values throughout.
- [x] `elastic-package check` passes; `elastic-package test pipeline -d aadgraphactivitylogs` passes.
- [x] End-to-end stack test: events POSTed to `logs-azure.events-default` are correctly rerouted to `logs-azure.aadgraphactivitylogs-default` with full ECS field extraction.

```bash
cd packages/azure
elastic-package check
elastic-package install
elastic-package test pipeline -d aadgraphactivitylogs
```

Optional end-to-end:

```bash
# Pipe the captured fixtures through the live events router
python3 -c "
import json
with open('data_stream/aadgraphactivitylogs/_dev/test/pipeline/test-aadgraph-activity.log') as f:
    for line in f:
        line = line.strip()
        if line:
            print(json.dumps({'create': {'_index': 'logs-azure.events-default'}}))
            print(json.dumps({'message': line}))
" > /tmp/aadgraph-bulk.ndjson

curl -sk -u user:pass -X POST \
  "https://localhost:9200/_bulk?refresh=wait_for" \
  -H "Content-Type: application/x-ndjson" \
  --data-binary @/tmp/aadgraph-bulk.ndjson

curl -sk -u user:pass  \
  "https://localhost:9200/logs-azure.aadgraphactivitylogs-default/_count"
# expected: {"count":4,...}
```

1. Open Discover at `https://localhost:5601` (`user` / `pass`)
2. data view `logs-azure.aadgraphactivitylogs-*`
3. confirm `event.action`, `event.outcome`, `http.*`, `url.path`, `azure.aadgraphactivitylogs.properties.*`, and `related.user` all populate.

## Related issues

- Closes #18879 

## Screenshots

**Pipeline tests passing locally**

Screenshot 2026-05-07 at 2 05 54 PM

**Discover view of the new dataset with ECS-parsed events**
**Expanded document showing the full ECS field tree**

Screenshot 2026-05-07 at 2 07 25 PM

Screenshot 2026-05-07 at 2 09 42 PM


packages/azure/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: "1.37.0-preview1"
```
terrancedejesus:
```suggestion
- version: "1.37.0"
```

packages/azure/manifest.yml
```diff
@@ -1,6 +1,6 @@
 name: azure
 title: Azure Logs
-version: "1.36.1"
+version: "1.37.0-preview1"
```
terrancedejesus:
```suggestion
version: "1.37.0"
```

efd6 (commented):
Please update the proposed commit message so that it's something that can be used in context of git; no Markdown, appropriately wrapped etc.

For example based on the current code here (update as needed):
```
azure: add aadgraphactivitylogs data stream

Add a dedicated data stream for the AzureADGraphActivityLogs
diagnostic category from Microsoft Entra ID. Without this,
legacy Azure AD Graph (graph.windows.net) events fall through
to the platformlogs catch-all and lose schema-aware parsing.

The events router maps routing.category ==
"AzureADGraphActivityLogs" to the new dataset. The ingest
pipeline extracts ECS fields: event.action from HTTP method +
URI collection, event.outcome from response status,
event.category [iam, web], and related.user including the
OAuth app_id for client correlation.

Legacy AAD Graph is still actively used by Microsoft first-party
tooling, older line-of-business apps, and adversary tooling
(ROADtools, AzureHound v1, AADInternals). The dedicated dataset
makes these events available for detection rules and dashboards.
```

terrancedejesus:
@efd6 proposed commit message updated. Thank you!
Any specific labeling or additional checks?

efd6:
The build is complaining:

> Error: error validating packages in directory 'packages': error checking data streams from 'packages/azure': package "packages/azure" shares ownership across data streams but these ones [packages/azure/data_stream/aadgraphactivitylogs] lack owners

I think you will need to add a line before [this](https://github.com/elastic/integrations/blob/cd8d67e56ebce425b361b9d526cd3b4cfe58a41f/.github/CODEOWNERS#L103). Who will be the owner of this data stream?

terrancedejesus:
> Who will be the owner of this data stream?

Yes, I noticed the buildkite fail related to owners. There are a few owners across the Azure package data streams it seems so I am not sure what team should be the owner/maintainer? I assume since these are the legacy data stream for Microsoft Graph and it was provisioned to write threat detection rules on, we mirror that so @elastic/security-service-integrations?

packages/azure/data_stream/aadgraphactivitylogs/agent/stream/log.yml.hbs
chrisberkhout, line 1: This file shouldn't be here, because the `log` input isn't used. It should be removed for all data streams. If you don't want to include that in this PR, at least don't add the new one.

packages/azure/changelog.yml
chrisberkhout, line 1: The list of data streams in the README should be updated.

If it's intended that this data stream has its own item in the integrations browser, separate from the full integration, as is the case for several of the other data streams, then there should be a policy template entry in the top level manifest, and an extra file of documentation in `_dev/build/docs/`.

packages/azure/data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
chrisberkhout, line 1: This pipeline does match about half the other pipelines with the same name:
```
find -name 'azure-shared-pipeline.yml' | xargs md5sum | sort
```
```
28624170d9ba87d593c9aef7dd72284a  ./data_stream/application_gateway/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
28624170d9ba87d593c9aef7dd72284a  ./data_stream/firewall_logs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
352b4a0232fcf818b45a174958b161e8  ./data_stream/identity_protection/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
352b4a0232fcf818b45a174958b161e8  ./data_stream/provisioning/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
8fb30aa0822189b990f17ba026aeb928  ./data_stream/platformlogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
d9d42c87fb050f4264f140fe5c4fddd0  ./data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
d9d42c87fb050f4264f140fe5c4fddd0  ./data_stream/activitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
d9d42c87fb050f4264f140fe5c4fddd0  ./data_stream/auditlogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
d9d42c87fb050f4264f140fe5c4fddd0  ./data_stream/eventhub/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
d9d42c87fb050f4264f140fe5c4fddd0  ./data_stream/graphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
d9d42c87fb050f4264f140fe5c4fddd0  ./data_stream/signinlogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
d9d42c87fb050f4264f140fe5c4fddd0  ./data_stream/springcloudlogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
```
It's a good time to use the [new links functionality](https://github.com/elastic/elastic-package/blob/main/docs/howto/links.md), at least for the ones that do still match.

packages/azure/data_stream/aadgraphactivitylogs/_dev/test/pipeline/test-common-config.yml
```diff
@@ -0,0 +1,7 @@
+fields:
+  tags:
+    - preserve_original_event
+dynamic_fields:
+  # This can be removed after ES 8.14 is the minimum version.
+  # Relates: https://github.com/elastic/elasticsearch/pull/105689
+  url.extension: '^.*$'
```
chrisberkhout: Looks like 8.19 is the minimum, so this comment's advice can be followed.

packages/azure/data_stream/aadgraphactivitylogs/fields/fields.yml
```diff
@@ -0,0 +1,116 @@
...
+        The legacy Azure AD Graph API version of the event (e.g. `1.6`).
+    - name: operation_name
+      type: keyword
+      description: |
+        Operation name.
+    - name: result_signature
+      type: keyword
+      description: |
+        Result signature.
+    - name: properties
```
chrisberkhout, line 12: Are all the fields documented in https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/aadgraphactivitylogs retained here or mapped to common fields outside of this prefix or to ECS fields?

chrisberkhout: It would be helpful for future maintenance if that API doc was linked either in the README or in a comment in the ingest pipeline.

packages/azure/data_stream/events/routing_rules.yml
chrisberkhout, line 1: The  description in `data_stream/events/manifest.yml`, starting with

> Collect all the supported (see list below) Azure logs from Event Hub to a target data stream.
> 
> ✨ **New in version 1.20.0+:** by enabling this integration, you can collect all the logs from the following Azure services and route them to the appropriate data stream:

should also be updated.

chrisberkhout (changes requested):

packages/azure/data_stream/aadgraphactivitylogs/agent/stream/log.yml.hbs
terrancedejesus: removed: [30046ce36184e5916a2b5f69455b6c5b0fe976f6](https://github.com/elastic/integrations/pull/18880/changes/30046ce36184e5916a2b5f69455b6c5b0fe976f6)

packages/azure/changelog.yml
terrancedejesus: Thanks Chris! Intended for AAD Graph Activity logs to have its own item. Here are the updated commits.

- [0a4f6973f89ad1c070d269aeab138e7be02bd14b](https://github.com/elastic/integrations/pull/18880/changes/0a4f6973f89ad1c070d269aeab138e7be02bd14b): Adds docs
- [20ba136297e2bc2d0e950cfdb17032c56efcc328](https://github.com/elastic/integrations/pull/18880/changes/20ba136297e2bc2d0e950cfdb17032c56efcc328): Adds a manifest for Azure AD Graph Activity Logs
- [8e1dc867400bfc1b8fac53dee0560715e0a538df](https://github.com/elastic/integrations/pull/18880/changes/8e1dc867400bfc1b8fac53dee0560715e0a538df): I had logs streaming in over the weekend to have more of a sample dataset to test mapping on. These adjustments address findings.

After changes, streamed all logs into my local stack to check mappings and confirm accuracy. Also verified the integration showed up based on the manifest and doc changes.

Screenshot 2026-05-11 at 11 07 23 AM

packages/azure/data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
terrancedejesus: Added per recommendations: [7be1278fcf4bcd9e2723782c95c43f5a55f45a2d](https://github.com/elastic/integrations/pull/18880/changes/7be1278fcf4bcd9e2723782c95c43f5a55f45a2d)

Note, I did not adjust all of them that could be as that seems out of scope for this. Maybe a separate issue/pr for this?

packages/azure/data_stream/aadgraphactivitylogs/_dev/test/pipeline/test-common-config.yml
terrancedejesus: Removed: [a62a1acfa031743851f6da084009f8701383be18](https://github.com/elastic/integrations/pull/18880/changes/a62a1acfa031743851f6da084009f8701383be18)

packages/azure/data_stream/aadgraphactivitylogs/fields/fields.yml
terrancedejesus: > Are all the fields documented in https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/aadgraphactivitylogs retained here or mapped to common fields outside of this prefix or to ECS fields?

yes confirmed they are all mapped or routed to ECS. Updated example data for testing if you'd like to see locally.

> It would be helpful for future maintenance if that API doc was linked either in the README or in a comment in the ingest pipeline.

`ingest_pipeline/default.yml` now has a header comment with the MS docs URL and a note explaining the envelope-vs-projection distinction. (The README intro already mentions the same URL, so future maintainers see it from either entry point.): [99758fd2ab81c636797e80c1b72efaf6890483fe](https://github.com/elastic/integrations/pull/18880/changes/99758fd2ab81c636797e80c1b72efaf6890483fe)

packages/azure/data_stream/events/routing_rules.yml
terrancedejesus: updated: [7f1b4b693985486e62b524ba0b62643b78cfd95f](https://github.com/elastic/integrations/pull/18880/changes/7f1b4b693985486e62b524ba0b62643b78cfd95f)

terrancedejesus:
@chrisberkhout - Thank you again for the feedback and guidance. All comments have been addressed. For each comment, I added the commit for easier reference. The latest buildkit [build](https://github.com/elastic/integrations/pull/18880#issuecomment-4422767006) has succeeded.

Additionally, I've rebuilt locally a new stack and ingested AAD Graph Activity Logs to verify logs are as expected and integration updates are accurate. Please let me know if if there is anything else to add/adjust.

packages/azure/data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,643 @@
...
+  - rename:
+      field: azure.aadgraphactivitylogs.properties.user_id
+      target_field: user.id
+      tag: rename-properties_user_id
+      if: ctx.azure?.aadgraphactivitylogs?.properties?.user_id != null && ctx.azure.aadgraphactivitylogs.properties.user_id != ''
+  - set:
+      field: client.user.id
+      value: '{{user.id}}'
+      tag: set-client_ip
+      ignore_empty_value: true
```
zmoog, line 362:
```suggestion
      field: client.user.id
      value: '{{user.id}}'
      tag: set-client_user_id
      ignore_empty_value: true
```

packages/azure/data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,643 @@
...
+  - set:
+      field: cloud.region
+      value: '{{destination.geo.region_name}}'
+      tag: set-cloud_region
+      ignore_empty_value: true
+  - set:
+      field: cloud.service.name
+      value: 'Azure AD Graph'
+      tag: set-cloud_region
+      ignore_empty_value: true
```
zmoog, line 563:
```suggestion
  - set:
      field: cloud.service.name
      value: 'Azure AD Graph'
      tag: set-cloud_service_name
      ignore_empty_value: true
```

packages/azure/data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,643 @@
...
+        else { verb = method.toLowerCase(); }
+        ctx.event.action = hyphenated.toString() + '-' + verb;
+      ignore_failure: true
+
+  - pipeline:
+      name: '{{ IngestPipeline "azure-shared-pipeline" }}'
+      tag: pipeline-azure-shared-pipeline
+  - fingerprint:
+      fields: 
+        - azure.aadgraphactivitylogs.properties.request_uri
```
zmoog: The pipeline removes the `azure.aadgraphactivitylogs.properties.request_uri` field a few lines before if `url.original` is not null. Should we keep it or replace this with a different field?

zmoog (approved):
LGTM, just a few nits.

packages/azure/data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/default.yml
terrancedejesus: @zmoog Good catch and thank you for all the review! If I understand correctly, the remove a few lines up was stripping properties.request_uri before the fingerprint got to use it, so that input was silently dropping out (ignore_missing: true masked it). The hash was effectively only using http.request.id, azure.tenant_id, and properties.time_generated and confirmed this against the data.

Swapped it to use url.original instead, which holds the same value (the uri_parts step keeps a copy via keep_original: true) and is still around at fingerprint time. Re-streamed live events through twice end-to-end and confirmed the _ids are stable now.

fix ref: [8fae9c52502f6fa6d3153d78ea16fa265dc37851](https://github.com/elastic/integrations/pull/18880/changes/8fae9c52502f6fa6d3153d78ea16fa265dc37851)

muthu-mps (approved):
code owner approval!

packages/azure/data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
chrisberkhout: It's a step in the right direction. I think it would be good to move that version of the file to the top level and change the identical copies to links. Could be done in this PR.

```
# place the shared copy at the top level
mkdir -p ./_dev/shared/
mv ./data_stream/graphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml ./_dev/shared/

# update link to the shared copy
echo "../../../../_dev/shared/azure-shared-pipeline.yml d9410a8b01f785a2e328560cc5d9a2286c8ca5b4e6ab5ec0edfaba385ddb0fb9" > \
  ./data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link

# remove other hard copies
rm ./data_stream/activitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
rm ./data_stream/auditlogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
rm ./data_stream/eventhub/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
rm ./data_stream/signinlogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml
rm ./data_stream/springcloudlogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml

# add other links to the shared copy
cp ./data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link ./data_stream/activitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link
cp ./data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link ./data_stream/auditlogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link
cp ./data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link ./data_stream/eventhub/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link
cp ./data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link ./data_stream/graphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link
cp ./data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link ./data_stream/signinlogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link
cp ./data_stream/aadgraphactivitylogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link ./data_stream/springcloudlogs/elasticsearch/ingest_pipeline/azure-shared-pipeline.yml.link
```

chrisberkhout (approved):
Thanks. Looks good.

I gave a script to adopt links for the other uses of the same version of the ingest pipeline.

PR #18828, opened 2026-05-05
[#17491][cisco-ftd]: Parse user authentication rejection reasons

Niceplace:
## Proposed commit message

Add grok processor support in cisco FTD integration to accept four variants of user authentication rejection reasons: Invalid Password, Password is expiring, Password malformed, Password has expired. The support for "Invalid Password" is scoped to address issue #17491 whereas the other three rejection reasons are a backport of https://github.com/elastic/integrations/pull/8015 

## Checklist

- [x] I have reviewed [tips for building integrations](https://www.elastic.co/docs/extend/integrations/tips-for-building) and this pull request is aligned with them.
- [x] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [x] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally

From the root of the repo

1. Go to `packages/cisco_ftd`
2. Run `elastic-package stack up -d` 
3. Run `elastic-package test pipeline`
4. All tests should pass

Sample logs used for pipeline tests are in `/packages/cisco_ftd/data_stream/log/_dev/test/pipeline/test-ftd-fix-message-113005-user-auth-rejected.log`

## Related issues

Fixes: https://github.com/elastic/integrations/issues/17491 and as recommended in https://github.com/elastic/integrations/issues/17491#issuecomment-3933219500, implements the same support for authentication reason rejections as https://github.com/elastic/integrations/pull/8015

PR #18881, opened 2026-05-07
Handle self-signed certs expiration in package tests

Niceplace:
## Summary of changes

Re-generate expired self-signed certificates used to test integration packages and add a monitoring script in the CI that will run before every package is tested to flag expired and close-to-expiry certificates.

The CI will fail the test step for a given package if its certificates expires in 6 months or less.
The CI will log a warning in the test step for a given package if the certificate expires in more than 6 months but less than a year.

Here's how the diagnosis was done for the `aws.config` integration:

1. `aws.config` tests were regularly failing (timing out) in https://github.com/elastic/integrations/pull/18828
2. Local execution with `elastic-package up -d;elastic-package test system -d config` from `packages/aws` folder showed the same behavior locally
3. Container log investigation showed errors related to x509
```
elastic-agent-1  | {"log.level":"error","@timestamp":"2026-05-07T16:33:55.327Z","message":"Error dialing x509: certificate has expired or is not yet valid: current time 2026-05-07T16:33:55Z is after 2026-05-06T06:27:43Z","component.id":"cel-default","component.type":"cel","component.binary":"filebeat","component.dataset":"elastic_agent.filebeat","log.source":"cel-default","log.origin":{"file.line":39,"file.name":"transport/logging.go","function":"github.com/elastic/elastic-agent-libs/transport/httpcommon.(*HTTPTransportSettings).RoundTripper.LoggingDialer.func2"},"id":"cel-aws.config-569bd14e-3f72-4aed-835d-c40421343ac3","network.transport":"tcp","server.address":"config.xxxx.amazonaws.com:443","log.logger":"input.cel","service.name":"filebeat","input_source":"https://config.xxxx.amazonaws.com/","input_url":"https://config.xxxx.amazonaws.com/","ecs.version":"1.6.0","ecs.version":"1.6.0"}
```
4. Checked certificate expiry date with `openssl x509 -enddate -noout -in $CERT_FILE_PATH` 
5. Re-generated certificate with expiry 10 years from now
6. Run the same test again `elastic-package up -d;elastic-package test system -d config` from `packages/aws` and it passes almost immediately instead of timing out after 10 minutes

## Proposed commit message

  Re-generate expired and close to be expired self-signed certificates for aws.config and cybereason integrations. Add a monitoring script in the CI that will run before every integration package is tested to flag expired and close-to-expiry certificates and fail the test with a comprehensive JUNIT report.

## Checklist

1. From the root of the repo: `cd  packages/aws`
2. Run `elastic-package stack up -d`
3. Run `elastic-package test system -d config` so it only runs the config tests
4. All tests pass 

If you want to reproduce the failure, try with a commit from `main` such as bebb005b5848848aaef614b653d22c5e7af48af1

## Related pipeline failures

https://buildkite.com/elastic/integrations/builds/42437/canvas?sid=019dfe60-06c0-4b86-a3f2-b404855cfa40&tab=output


.buildkite/scripts/test_one_package.sh
```diff
@@ -26,6 +26,11 @@ with_kubernetes
 
 use_elastic_package
 
+echo "--- [${package_name}] Check TLS certificate expiry"
```
Niceplace, line 31: I assumed this was the right place to put the check, this way it will run once for each package immediately before the tests, at least that's how I understand the pipeline structure.

This way only tests that have expired or close-to-be-expired tests will fail, and not the entire job.

Niceplace:
Question: Does it make sense to bump package changelogs in this case ? I did not do it because this doesn't fix a bug in the package's code, it only modifies the setup used to test it. I'm happy to revisit this though

efd6 (approved):
LGTM, but please update the proposed commit message to be the text that [will be included as the commit message](https://github.com/elastic/integrations/blob/cd8d67e56ebce425b361b9d526cd3b4cfe58a41f/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L19-L21) (no Markdown since `git log` is not Markdown-aware).

.buildkite/scripts/check_certificates.sh
```diff
@@ -0,0 +1,110 @@
...
+
+echo "Summary: ${error_count} error(s), ${warning_count} warning(s)"
+
+if [ "$error_count" -gt 0 ]; then
+    echo ""
+    echo "Renew the certificates above, then sync test configs:"
+    echo "  openssl req -x509 -newkey rsa:2048 -keyout <key> -out <cert> \\"
+    echo "    -subj '<subject>' [-addext 'subjectAltName=DNS:<hostname>'] -days 3650 -noenc"
+    echo "  .buildkite/scripts/update-test-cert.sh <cert>"
+    exit 1
```
mrodm, line 136: Wondering if this could create a XML file following the jUnit format, so it can be then reported as a failure and follow the automation of flaky-tests to create GitHub issues with this error.

.buildkite/scripts/check_certificates.sh
```diff
@@ -0,0 +1,110 @@
...
+done
+
+echo "Summary: ${error_count} error(s), ${warning_count} warning(s)"
+
+if [ "$error_count" -gt 0 ]; then
+    echo ""
+    echo "Renew the certificates above, then sync test configs:"
+    echo "  openssl req -x509 -newkey rsa:2048 -keyout <key> -out <cert> \\"
+    echo "    -subj '<subject>' [-addext 'subjectAltName=DNS:<hostname>'] -days 3650 -noenc"
+    echo "  .buildkite/scripts/update-test-cert.sh <cert>"
```
mrodm: This file does not exist in the PR.

.buildkite/scripts/check_certificates.sh
mrodm: Some example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="system" tests="20" failures="1">
    <!--test suite for system tests-->
    <testcase name="system test: default" classname="aws.cloudtrail" time="105.619672145"></testcase>
    <testcase name="system test: data_granularity" classname="aws.ec2_metrics" time="717.594866274"></testcase>
    <testcase name="system test: default" classname="aws.ec2_metrics" time="1022.428637031"></testcase>
    <testcase name="system test: default" classname="aws.securityhub_findings_full_posture" time="39.192594525"></testcase>
    <testcase name="system test: default" classname="aws.cloudfront_logs" time="109.69616598"></testcase>
    <testcase name="system test: default" classname="aws.elb_logs" time="100.8600316"></testcase>
    <testcase name="system test: default" classname="aws.securityhub_insights" time="36.60892882"></testcase>
    <testcase name="system test: default" classname="aws.waf" time="108.81979344"></testcase>
    <testcase name="system test: default" classname="aws.s3access" time="106.250401394"></testcase>
    <testcase name="system test: default" classname="aws.apigateway_logs" time="194.243282454"></testcase>
    <testcase name="system test: default" classname="aws.ec2_logs" time="103.762434038"></testcase>
    <testcase name="system test: default" classname="aws.firewall_logs" time="107.836685465"></testcase>
    <testcase name="system test: default" classname="aws.guardduty" time="44.097627039"></testcase>
    <testcase name="system test: default" classname="aws.inspector" time="39.243202065"></testcase>
    <testcase name="system test: default" classname="aws.route53_resolver_logs" time="103.599879859"></testcase>
    <testcase name="system test: default" classname="aws.securityhub_findings" time="42.964267548"></testcase>
    <testcase name="system test: default" classname="aws.config" time="643.350771599">
      <failure>test case failed: could not find the expected hits in logs-aws.config-28774 data stream</failure>
    </testcase>
    <testcase name="system test: default" classname="aws.emr_logs" time="111.049148092"></testcase>
    <testcase name="system test: default" classname="aws.redshift" time="812.692963297"></testcase>
    <testcase name="system test: default" classname="aws.vpcflow" time="105.466743095"></testcase>
  </testsuite>
</testsuites>
```

mrodm (commented):
@elastic/ecosystem WDYT about adding this new check in the CI  ?

.buildkite/scripts/check_certificates.sh
Niceplace: Oh ! Nice catch thank you, will fix that

.buildkite/scripts/check_certificates.sh
Niceplace: I like that ! I'm thinking maybe I could shift the responsibility of this script in the tests, something like in the "beforeEach" hook, this way we get the actual failure in Junit natively. 

WDYT ?

mrodm: These files are generated by `elastic-package` via the `elastic-package test` commands, one file per test type (static, system, pipeline...). There is no support for hooks there. And moreover, as this is detected as part of the CI scripts, before any `elastic-package` command related to testing, `elastic-package` cannot do anything here.

If you meant to add this functionality to `elastic-package`, I'm not sure this should belong there. I think `elastic-package` should be agnostic to the files being present in the docker deployer folder. WDYT  @elastic/ecosystem ?

Niceplace: Yeah I understand your concerns, adding the check for expired certs within `elastic-package` is where I was going next, since there are currently [14 diferent certificates](https://elastic.slack.com/archives/C08JBEAUTKR/p1778178551185009) that we would need to monitor. It felt like approaching this from the test runner's perspective would give us a scalable solution and require less maintenance.

This condition needs to be caught before a test runs though because once the test runs, the root cause (the log entry that mentions certificate expiry) will only be logged in the container logs and those are not included in buildkite artifacts at the moment. This means that automated investigation tools such as the PR Detective won't have the necessary data to flag it properly and we are left with an error message that is vague (Assertions failed, test timed out).

I'll see what I can do to generate a JUNIT file in the current script, if you have any other pointers, I'm definitely interested :D

Niceplace: Sweet ! So generating the junit report makes the root cause visible [to the PR detective](https://github.com/elastic/integrations/pull/18881#issuecomment-4407653529) so that's nice

.buildkite/scripts/write_junit_cert_report.py
```diff
@@ -0,0 +1,69 @@
...
+    print(f"write_junit_cert_report.py: expected cert_path/message pairs, got {len(items)} argument(s)", file=sys.stderr)
+    sys.exit(1)
+
+pairs = [(items[i], items[i + 1]) for i in range(0, len(items), 2)]
+
+testsuites = ET.Element("testsuites")
+testsuite = ET.SubElement(testsuites, "testsuite")
+testsuite.set("name", "system")
+testsuite.set("tests", str(len(pairs)))
+testsuite.set("errors", str(len(pairs)))
```
mrodm, line 51: Could it be that not all tests are failing ? For instance, from 5 tests containing SSL certificates, just 2 of them show issues related to expiration dates.

.buildkite/scripts/write_junit_cert_report.py
```diff
@@ -0,0 +1,69 @@
...
+tree = ET.ElementTree(testsuites)
+try:
+    ET.indent(tree, space="  ")
+except AttributeError:
+    pass  # ET.indent added in Python 3.9
+
+with open(output_path, "w", encoding="utf-8") as f:
+    f.write('<?xml version="1.0" encoding="UTF-8"?>\n')
+    tree.write(f, encoding="unicode")
+    f.write("\n")
```
mrodm, line 69: Could you add some tests , at least to be run locally as it is done for the `find_oldest_supported_version.py`?

.buildkite/scripts/write_junit_cert_report.py
```diff
@@ -0,0 +1,69 @@
...
+import xml.etree.ElementTree as ET
+
+if os.environ.get("_CERT_CHECK_CALLER") != "check_certificates.sh":
+    print("write_junit_cert_report.py must be called from check_certificates.sh", file=sys.stderr)
+    sys.exit(1)
+
+def classname_from_path(cert_path, fallback):
+    """packages/aws/data_stream/config/_dev/... → 'aws.config'
+    packages/cybereason/_dev/...             → 'cybereason'"""
+    m = re.search(r"packages/([^/]+)/data_stream/([^/]+)/", cert_path)
```
mrodm, line 28: This regex should support nested directories for packages (relates to https://github.com/elastic/integrations/issues/17403). There is no package yet following that, but it would be good that these scripts support it.
For instance, there could be packages in the future like:
- packages/nginx/nginx_otel/...
- packages/nginx/nginx_input_otel/...

.buildkite/scripts/write_junit_cert_report.py
mrodm: Currently, these tests are not running in CI but it would be interesting to do so.

I created a PR to include tests for the current python script:  https://github.com/elastic/integrations/pull/18942

If that PR is merged before this PR, it would be good that you add the execution of tests for this script too there.

.buildkite/scripts/write_junit_cert_report.py
mrodm: Could it be used the package_name that it is set as a parameter for this script here for that? 

Or maybe even the package_name from the `manifest.yml` that can be obtained in the check_certificates.sh 
```sh
yq -r '.name' "${search_root}/manifest.yml"
```

Package folder names are not forced to be unique (when using nested directories), package names in their manifest.yml files are validated to be unique.

If you use as parameters for this python script:
- package_name: as a first part of the `class_name`. 
- search_root: to remove the prefix from the cert paths checked to retrieve (I hope) easier the data_stream name.

I think you could use as `class_name`: `<package_name>` or `<package_name>.<data_stream_name>` if there is a data_stream folder in the path

.buildkite/scripts/test_one_package.sh
```diff
@@ -26,6 +26,11 @@ with_kubernetes
...
+if ! .buildkite/scripts/check_certificates.sh "${package_path}" ; then
+    exit 1
+fi
```
mrodm, line 35: Instead of existing and failing the buildkite build. I think it would be better to continue the execution of the tests as it now.

As this new check also creates a jUnit file and it will be reported/notified via GitHub issues, I think it is ok to continue testing the packages as usual.

While the time the certs are not fixed, there will no be tests running and it could hide potential issues. It's likely that these certs are not updated in the next few days of being the GitHub issue created.


```
if ! .buildkite/scripts/check_certificates.sh "${package_path}" ; then
    echoerr "Failed validation of SSL certificates for ${package_path}"
fi
```

Or maybe creating an annotation (but there are already a few):
```
if ! .buildkite/scripts/check_certificates.sh "${package_path}" ; then
     buildkite-agent annotate "Failed validation of SSL certificates for ${package_path}" --style "warning" --context "ctx-ssl-${package_path//\//-}"
fi
```

.buildkite/scripts/check_certificates.sh
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env bash
+# Scans TLS certificates used by test mock servers under packages/_dev/ directories.
+#
+# Many integrations use a local HTTPS mock server (via docker.elastic.co/observability/stream)
+# to simulate vendor APIs during system tests. Each mock server is configured with a
+# self-signed TLS certificate checked into the repo under:
+#   packages/<integration>/_dev/deploy/docker/files/*.crt
```
mrodm, line 7: There could be another directory level of directories:

- `packages/<package>/_dev/deploy/docker/files/*.crt`
- `packages/<technology>/<package>/_dev/deploy/docker/files/*.crt`

Relates to https://github.com/elastic/integrations/issues/17403

.buildkite/scripts/check_certificates.sh
```diff
@@ -0,0 +1,132 @@
...
+#   check_certificates.sh [package_path]
+#
+#   package_path  Optional path to a single package (e.g. packages/cybereason).
+#                 When omitted, all packages under packages/ are scanned.
+#
+# Thresholds:
+#   ERROR   — certificate is expired or expires within 6 months (180 days)
+#   WARNING — certificate expires within 1 year (365 days)
+#
+# Exits non-zero if any ERROR is found. Warnings are printed but do not fail the build.
```
mrodm: As mentioned in the other comment, I think it would be better to run the tests to try to detect as much errors as possible in one run. There could be some data streams in the package that don't use any SSL certificates. Or some tests that do not require SSL certificates in the same data stream.

mrodm:
@Niceplace I think there are now some issues already reported for the packages failing (netbox, aws...) , so before retrying the whole CI build (and therefore testing all packages again) let's check that first.
Those errors should not be related to the changes introduced here , at least if all the SSL certs are fixed. And if needed, we can force merge this PR.

.buildkite/scripts/check_certificates.sh
Niceplace: Totally valid ! I'm going to remove that exit code in `test_one_package.sh` and log an error instead. I think we should leave the `exit 1` in place for `check_certificates.sh` as we use this exit code in `test_one_package.sh` to determine the outcome of the cert check. Does that make sense ?  

Also. to save on CI execution time, what would be ideal is to skip over the tests that fail the cert check and not even attempt to run them. This avoid having tests that hang because the connection between the mock server and the elastic agent is never established and this can add 10 to 20 minutes on the pipeline run. I'll see what I can do about that.

Niceplace: After some investigation on skipping tests that fail the cert check, given we have no way to exclude specific tests using arguments or env vars with `elastic-package` I'm going to pass on that idea for now. The proposed solution adds too much complexity and would modify `test-*-config.yml` files in place, which is not a nice side effect for the CI.

This is not a downgrade from the status quo but an optimization we would be choosing not to do.

.buildkite/scripts/test_one_package.sh
Niceplace: Gotcha ! I'm going with a buildkite annotation. If you think this will be too noisy let me know I can fallback to echo instead.

PR #18891, opened 2026-05-08
[m365_defender] Unique `entity_id` values

chrisberkhout:
## Proposed commit message

```
[m365_defender] Unique `entity_id` values

Set values in `entity_id` that uniquely identify a process, rather than
using the OS PID value, which may not be unique. Since PIDs can repeat
across hosts and be reused on a single host, additional fields are
necessary to identify the process.

This change should prevent irrelevant documents being shown in the Event
Analyzer due to ID clashes.

ECS defines `process.entity_id` as:

> Unique identifier for the process.
>
> The implementation of this is specified by the data source, but some
> examples of what could be used here are a process-generated UUID,
> Sysmon Process GUIDs, or a hash of some uniquely identifying
> components of a process.
>
> Constructing a globally unique identifier is a common practice to
> mitigate PID reuse as well as to identify a specific process over
> time, across multiple monitored hosts.

Although that definition says it should identify a process globally, in
practice, other EDR integrations such as `crowdstrike` and
`sentinel_one_cloud_funnel` use globally unique IDs assigned by an
agent, not a value based on the OS PID. That means the ID for a process
won't clash with another, but may also not match the ID used for the
same process in data from a different source.

The Resolver Schemas[1] document in Kibana refers to `entity_id` but
provides no further guidance on this point.

[1]: https://github.com/elastic/kibana/blob/fb26c1c/x-pack/solutions/security/plugins/security_solution/public/resolver/documentation/schema.md
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/m365_defender/changelog.yml
+++ b/packages/m365_defender/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "5.13.1"
+  changes:
+    - description: Unique `entity_id` values to avoid irrelevant documents in the Event Analyzer.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/18891
 - version: "5.13.0"
   changes:
     - description: Add timestamp range filter and storage tier exclusion to latest transform source queries to reduce scan scope and improve performance.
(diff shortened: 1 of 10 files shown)


efd6 (commented):
Is there a reason to make the entity_id opaque rather than leaving it as the input to the hash?

In the commit message: `s/defintion/definition/` and remove the square brackets around "Resolver Schemas".

chrisberkhout:
> Is there a reason to make the entity_id opaque rather than leaving it as the input to the hash?

It's probably good to discourage parsing things out of that value. It could work unhashed though. EDR integrations use values assigned by an agent (CrowdStrike, SentinelOne), so they're not fingerprinting.

> In the commit message: `s/defintion/definition/`

:+1: 

> and remove the square brackets around "Resolver Schemas".

Also did it. But just because it's Markdown? I do keep hearing that commit messages shouldn't have Markdown, but I think light Markdown is good in commit messages. It's a plain text format that's supposed to be read as plain text, as well as having other formatting possibilities. I wouldn't put tables without aligned columns or numbered lists without increasing numbers, or HTML, or images. Maybe you could say a basic link is a borderline case.

efd6 (approved):

efd6:
> But just because it's Markdown? I do keep hearing that commit messages shouldn't have Markdown, but I think light Markdown is good in commit messages.

Markdownish is fine where it helps (for example the `>` you have, though this predates Markdown in email and has been conventionally also represented as spaced indentation in commit messages), but I don't think the square brackets helped there.

PR #18894, opened 2026-05-08
ci: handle all package documentation edits

shmsr:
## Summary
- Expand the Documentation edit helper workflow to cover all flat package source docs under `packages/*/_dev/build/docs/*.md`, including per-data-stream files like `cloudtrail.md`.
- Require both `manifest.yml` and `changelog.yml` before considering a docs package complete, and report partial metadata updates separately.
- Harden generated changelog commands by safely passing custom `changelog:` text between steps and shell-quoting it in the suggested command.

## Test plan
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/docs-edit-automation.yml`
- Focused shell check for custom `changelog:` parsing with shell-sensitive text.
- Focused shell check for metadata classification: no metadata, complete metadata, missing manifest, and missing changelog.


.github/workflows/docs-edit-automation.yml
```diff
@@ -6,7 +6,11 @@ on:
       - opened
       - synchronize
     paths:
-      - 'packages/*/_dev/build/docs/README.md'
+      # Match any markdown source file under a package's _dev/build/docs/
+      # directory. This covers both the package-level README.md and
+      # per-integration/per-data-stream files such as cloudtrail.md,
+      # activitylogs.md, gke.md, etc.
+      - 'packages/*/_dev/build/docs/*.md'
```
mrodm: Would this pattern match packages in nested directories ?
Like `packages/nginx/nginx_otel/` ?  

There are still no packages located in nested directories, but it would be good that all these processes support that too to be prepared.

Related to this issue: https://github.com/elastic/integrations/issues/17403

shmsr: Fixed. Thank you!

PR #18927, opened 2026-05-09
[juniper_srx] extract IP from RPD_IFD_NOTIFICATION events

ilyannn:
## Summary
- Add a tag-specific `grok` in the `system` pipeline to extract the interface address from `RPD_IFD_NOTIFICATION` messages (pattern `address #N <IP>`) into `observer.ip`, `host.ip` and `related.ip`. The address is configured on the SRX's own interface, so `observer.ip` is its natural home (the data stream declared that field but never populated it); mirroring onto `host.ip` follows the existing `observer.name` → `host.name` copy.
- Map `juniper.srx.interface_index` (the Junos internal IFD index) to ECS `observer.ingress.interface.id`, for consistency with the existing `interface_name` → `observer.ingress.interface.name` rename (interface up/down events carry no flow direction). This is a **different numbering space** from the SNMP ifIndex that `SNMP_TRAP_LINK_UP`/`SNMP_TRAP_LINK_DOWN` report as `snmp-interface-index` — Junos reports both per interface and they routinely differ (e.g. `ge-0/0/0` → IFD 130 / SNMP ifIndex 501) — so the two are deliberately *not* merged into one field, and `juniper.srx.snmp_interface_index` stays in the vendor namespace.
- Drop the redundant `juniper.srx.system.if_trace`: the generic k-v extractor surfaces it from `IF_TRACE:` messages, but it just duplicates `message` (message minus the `IF_TRACE: ` prefix). Also drop `juniper.srx.system` when the cleanup leaves it empty.
- Guard the `juniper.srx.message` → `message` rename with `ctx.message == null` so it does not fail (`rename` errors on an existing target) or overwrite the `message` already populated from the event's free text.
- Add a `RPD_IFD_NOTIFICATION` test case with fictional values.
- Bump version to `1.28.0`.

## Test plan
- [x] `elastic-package test pipeline -C packages/juniper_srx --data-streams log` — all 14 cases PASS (Elastic stack 9.4.3).
- [x] New `RPD_IFD_NOTIFICATION` test event in `test-system.log` produces `observer.ip: ["175.16.199.0"]`, `host.ip: ["175.16.199.0"]`, `related.ip: ["175.16.199.0"]`, and `observer.ingress.interface.id: "508"`.
- [x] `elastic-package check -C packages/juniper_srx` passes (lint + build, README regenerated with no diff).
- [x] Merged `main` (`da58441c2f`) in `40ba7f7a99` — no conflicts, and `main` has not touched `packages/juniper_srx`, so the PR diff is unchanged. Full pipeline suite and `check` re-run green on the merged tree.

## Review notes
All review threads are addressed:
- `if_trace` dropped, and `interface-index` mapped to an ECS field, per @taylor-swanson's comments.
- Vera's `observer.ip` suggestion taken. Its suggestion to also map `snmp_interface_index` into `observer.ingress.interface.id` was **not** taken, because the two indexes are different numbering spaces (see above); the misleading comment it flagged has been reworded.
- Vera's multi-address `grok` note is left as-is: I have not observed an `RPD_IFD_NOTIFICATION` carrying more than one address, and the `#N` in the pattern only mirrors the literal text Junos emits. Happy to revisit if a sample with `address #1` turns up.

One open question for reviewers, noted [in thread](https://github.com/elastic/integrations/pull/18927#discussion_r3778480485): ECS glosses `observer.ingress.interface.id` as "typically SNMP interface ID", so arguably the genuine SNMP ifIndex is the better tenant for that field and the IFD index belongs in the vendor namespace. Left as-is for now since it would invert the mapping originally requested; happy to switch.

This PR was largely generated with 🤖 Claude Code/Opus 5 under my supervision.


packages/juniper_srx/data_stream/log/fields/fields.yml
```diff
@@ -423,6 +423,8 @@
           type: integer
         - name: direction
           type: keyword
+        - name: if_trace
+          type: keyword
```
taylor-swanson: Do we really need `if_trace`? Could we drop the field instead? It seems to basically be the same as what's in `message`

packages/juniper_srx/data_stream/log/fields/fields.yml
```diff
@@ -522,6 +524,8 @@
       type: keyword
     - name: function_name
       type: keyword
+    - name: interface_index
+      type: keyword
```
taylor-swanson: I would prefer this to go to an ECS field, if possible, but the only two options for interface index would be (as a closest match):

- observer.egress.interface.id
- observer.ingress.interface.id

Two issues I see:
- Is interface.id mean the same thing as interface index?
- Can we determine the direction of traffic flow from this log?

packages/juniper_srx/data_stream/log/elasticsearch/ingest_pipeline/system.yml
```diff
@@ -901,7 +923,7 @@ processors:
     field: juniper.srx.message
     target_field: message
     ignore_missing: true
-    if: "ctx.juniper?.srx?.message != null"
+    if: "ctx.juniper?.srx?.message != null && ctx.message == null"
```
taylor-swanson, line 935: Are we getting complaints about overwriting an existing message?

taylor-swanson (commented):
A few comments, and merge conflicts will need to be resolved.

packages/juniper_srx/data_stream/log/elasticsearch/ingest_pipeline/system.yml
ilyannn: No, I just did not want to let `juniper.srx.message` field:

```
"message": "<Up Broadcast Multicast> agg_parent 0 address #0 175.16.199.0",
```

clobber the `message` field that is set in a different part:

```
"message": "IF_TRACE: EVENT <UpDown> st0.60 index 508 <Up Broadcast Multicast> agg_parent 0 address #0 175.16.199.0",
```

They come from two different places, one the key `message` and one the free text after the `[]`:

```
<30>1 ... RPD_IFD_NOTIFICATION [junos@1111.1.1.1.1.111 ... message="<Up Broadcast Multicast> ..."] IF_TRACE: EVENT <UpDown> st0.60 index 508 <Up Broadcast Multicast> ...
```

Without the guard the `rename` just fails.

ilyannn: We can fix the rename failure (clear the `message` first), but since we do have `juniper.srx.message` field in `fields.yml` it looked like we might want to use that field as well. The description of that field is not very helpful:

```yaml
    - name: message
      type: keyword
      description: |
        mesagge
```

This field is present in `test-atp.log`, for example:

```
            "juniper": {
                "srx": {
                    "message": "malware analysis detected host downloaded a malicious_file with score 9, sha256 ABC123",
```
and we don't rename it to `message` there.

packages/juniper_srx/data_stream/log/fields/fields.yml
ilyannn: Good point, I've dropped it.

packages/juniper_srx/data_stream/log/fields/fields.yml
ilyannn: Sure, it seems to be the IFD/SNMP index which matches `interface.id`. This seems to be an interface up/down event without a direction, but since we don't have `interface.id` at the root level, I've used `observer.ingress.interface.id` for consistency with the existing processor:

```
  - rename:
      tag: rename_juniper_srx_interface_name_to_observer_ingress_interface_name_4023c7b6
      field: juniper.srx.interface_name
      target_field: observer.ingress.interface.name
      ignore_missing: true
```

taylor-swanson (commented):
There are a few vera review bot comments that need addressing. No complaints otherwise.

packages/juniper_srx/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -221,6 +221,13 @@ processors:
       field: juniper.srx.interface_name
       target_field: observer.ingress.interface.name
       ignore_missing: true
+  # interface-index is the SNMP/IFD interface index, which maps to ECS observer.*.interface.id.
+  # Interface up/down notifications carry no flow direction, so it sits under ingress alongside the interface name.
+  - rename:
+      tag: rename_juniper_srx_interface_index_to_observer_ingress_interface_id
+      field: juniper.srx.interface_index
```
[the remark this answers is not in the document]
ilyannn: Looked this up, and the two indexes are **not** the same numbering space, so I'm deliberately not merging them.

Junos reports `Interface index` (the internal IFD index) and `SNMP ifIndex` as separate values for the same interface, and they routinely differ — from Juniper's own `show interfaces` documentation: `so-2/1/1` → IFD 149 / SNMP 33, `ge-0/0/0` → IFD 130 / SNMP 501, `ge-0/0/8` → IFD 146 / SNMP 520. There's also a dedicated `show interfaces snmp-index` command precisely because you can't derive one from the other.

So mapping `snmp_interface_index` into the same ECS field would put two different index spaces in one field — a pivot on `observer.ingress.interface.id` would then conflate an IFD 508 with an SNMP 508 for unrelated interfaces. `snmp_interface_index` stays in the vendor namespace.

I have taken the sub-point about the misleading comment — it did conflate the two. Reworded in cc5ee4e83f.

Worth noting the fixture is a coincidence and I should have caught it: all three events show `508` for `st0.60` because I picked that value for the new line, but on a real device the IFD index and SNMP ifIndex for that interface would differ.

@taylor-swanson one thing I'd like your call on, since it pushes back on the ECS mapping you asked for: ECS glosses `observer.ingress.interface.id` as "typically SNMP interface ID", so given the above, the genuine SNMP ifIndex is arguably the better fit for that field and the IFD index arguably belongs in the vendor namespace. Happy to keep it as-is (index in ECS, useful and populated) or invert it. Out of scope for this PR either way, but say the word if you'd rather I switch.


---
_This comment is written by :robot: Claude Code/Opus 5 under my supervision._

packages/juniper_srx/data_stream/log/elasticsearch/ingest_pipeline/system.yml
```diff
@@ -308,6 +308,28 @@ processors:
...
+    tag: grok_rpd_ifd_notification_address
+    if: "ctx.juniper?.srx?.tag == 'RPD_IFD_NOTIFICATION' && ctx.juniper?.srx?.message != null"
+    field: juniper.srx.message
+    patterns:
+    - 'address #%{NONNEGINT} %{IP:_temp_.rpd_ifd.address}'
+    ignore_failure: true
+- append:
+    tag: append_rpd_ifd_address_to_host_ip
+    if: "ctx._temp_?.rpd_ifd?.address != null"
+    field: host.ip
```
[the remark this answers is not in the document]
ilyannn: Agreed, taken in cc5ee4e83f. The address is configured on the SRX's own interface, so `observer.ip` is its natural home, and mirroring onto `host.ip` matches the existing `observer.name` -> `host.name` copy. It also means the field stops being permanently empty.

Expected event updated, and the full `juniper_srx` pipeline suite passes locally against a 9.4.3 stack (14/14).


---
_This comment is written by :robot: Claude Code/Opus 5 under my supervision._

packages/juniper_srx/data_stream/log/elasticsearch/ingest_pipeline/system.yml
```diff
@@ -308,6 +308,28 @@ processors:
...
             value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
 
+# RPD_IFD_NOTIFICATION
+# Extract interface address (e.g. "address #0 175.16.199.0") into host.ip and related.ip
+- grok:
+    tag: grok_rpd_ifd_notification_address
+    if: "ctx.juniper?.srx?.tag == 'RPD_IFD_NOTIFICATION' && ctx.juniper?.srx?.message != null"
+    field: juniper.srx.message
+    patterns:
+    - 'address #%{NONNEGINT} %{IP:_temp_.rpd_ifd.address}'
```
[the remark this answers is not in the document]
ilyannn: Correct that grok stops at the first match, so only `address #0` is captured — but I'd rather leave this as-is for now.

I have not seen a real `RPD_IFD_NOTIFICATION` carrying more than one address; the `#N` in the pattern is just mirroring the literal text Junos emits, so it isn't evidence that a list actually occurs. Swapping a one-line grok for a painless regex loop to handle a case I can't demonstrate seems like the worse trade — more surface area, harder to read, no confirmed benefit.

Leaving it, and happy to revisit the moment a sample with `address #1` turns up.


---
_This comment is written by :robot: Claude Code/Opus 5 under my supervision._

PR #18942, opened 2026-05-11
[CI] Add Buildkite step to run Python CI scripts unit tests

mrodm:
## Proposed commit message

Add a generic Buildkite pipeline step (`check-ci-python-scripts`) that runs
unit tests for Python scripts used in CI.

**NOTE**: It is used the provider gcp in Buildkite to ensure that the same python3 version as in the integration steps is used.

**WHAT:**
- New step `:snake: Python CI scripts unit tests` in `pipeline.yml` that runs
  `run_ci_python_scripts_tests.sh` on a GCP Ubuntu agent.
- The step runs in parallel with the existing `check` (Go sources) step, keeping
  CI feedback fast.
- `test-integrations` is gated on both `check` and `check-ci-python-scripts`
  passing before triggering package tests.
- The shell script sets up an isolated Python venv, installs pinned deps from
  `requirements-ci-python-scripts.txt`, and invokes `find_oldest_supported_version.py --test`.
- `.gitignore` updated to exclude the generated venv directory.
- `list_all_directories` in `common.sh` temporarily filtered to
  `packages/elastic_package_registry` for focused testing (to be reverted).

**WHY:**
`find_oldest_supported_version.py` contains logic critical to how the CI
pipeline selects which stack versions to test against. Without a CI gate, a
regression in that script would silently affect all integration test runs.
This PR adds that safety net and lays the groundwork for testing other Python
CI scripts in the same step.

## Author's Checklist

- [x] Revert the temporary `grep` filter in `list_all_directories` (`common.sh`) before merging.
- [x] Trigger a Buildkite build on this branch and verify the new `:snake: Python CI scripts unit tests` step appears and passes.
- [x] Confirm `test-integrations` only starts after both `check` and `check-ci-python-scripts` succeed.
- [x] Run `.buildkite/scripts/run_ci_python_scripts_tests.sh` locally to verify the venv setup and tests pass.

---

### DIFF
--- a/.buildkite/scripts/run_ci_python_scripts_tests.sh
+++ b/.buildkite/scripts/run_ci_python_scripts_tests.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+#
+# Runs Python unit tests for CI scripts in an isolated venv.
+# Intended for Buildkite (see pipeline.yml).
+set -euo pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+VENV_DIR="${REPO_ROOT}/.buildkite/.venv-ci-python-scripts"
+REQ_FILE="${REPO_ROOT}/.buildkite/scripts/requirements-ci-python-scripts.txt"
+FIND_OLDEST_SCRIPT="${REPO_ROOT}/.buildkite/scripts/find_oldest_supported_version.py"
+
+if [[ ! -f "${VENV_DIR}/bin/activate" ]]; then
+    rm -rf "${VENV_DIR}"
+    if ! python3 -m venv "${VENV_DIR}"; then
+        if command -v apt-get >/dev/null 2>&1 && [[ "$(id -u)" -eq 0 ]]; then
+            echo "python3 -m venv unavailable; installing python3-venv"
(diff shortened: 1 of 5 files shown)


.buildkite/scripts/run_ci_python_scripts_tests.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+#
+# Runs Python unit tests for CI scripts in an isolated venv.
+# Intended for Buildkite (see pipeline.yml).
+set -euo pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+VENV_DIR="${REPO_ROOT}/.buildkite/.venv-ci-python-scripts"
+REQ_FILE="${REPO_ROOT}/.buildkite/scripts/requirements-ci-python-scripts.txt"
+FIND_OLDEST_SCRIPT="${REPO_ROOT}/.buildkite/scripts/find_oldest_supported_version.py"
```
mrodm, line 10: To execute these tests:

https://github.com/elastic/integrations/blob/22fb8844e49e51223531b74685f464c6eec3202a/.buildkite/scripts/find_oldest_supported_version.py#L126-L274

teresaromero (approved):

PR #18962, opened 2026-05-12
[CI] Add features/* branch support to changeset detection and publish pipeline

mrodm:
### WHAT

Extends CI branch handling to treat `feature/*` branches the same as `backport-*` branches in two places:

**`common.sh`** — `get_from_changeset` / `get_to_changeset`

- The outer guard in both functions previously only matched `main` and `backport-*`, making the inner `feature/` check inside `get_from_changeset` unreachable (dead code) — `feature/*` branches never entered the block.
- Both outer guards are updated to `^(backport-|feature/)` so `feature/*` branches benefit from the same previous-successful-commit logic.
- On the first push of a `feature/*` branch (no previous successful build), `from` is set to `BUILDKITE_COMMIT` rather than `origin/${BUILDKITE_BRANCH}^`, avoiding spurious full-repo diffs that would trigger unnecessary package testing.

**`catalog-info.yaml`** — `integrations` pipeline

- `branch_configuration` extended from `"main backport-*"` to `"main backport-* feature/*"` so the integrations pipeline is triggered for `feature/*` branches.
- `cancel_intermediate_builds_branch_filter` and `skip_intermediate_builds_branch_filter` updated to include `!feature/*` so intermediate builds on `feature/*` branches are properly cancelled/skipped, consistent with `main` and `backport-*`.

**`docs/ci_pipelines.md`**

- Updated the pipeline trigger description, build-cancellation bullet, and changeset-detection bullet to include `feature/*` alongside `main` and `backport-*`.
- Added a new "Feature branches" section documenting that changes are integrated via Pull Requests with CI checks, that changeset detection follows the same logic as `backport-*` branches, and that publishing is restricted to `main` and `backport-*` only.

### WHY

`feature/*` branches are used for long-lived feature development that requires the same CI protections as `backport-*` branches: accurate changeset detection against a baseline and CI build triggering. The `integrations-publish` pipeline intentionally remains restricted to `main` and `backport-*` branches only.

## Author's Checklist

- [x] Verified that `feature/*` branches were previously excluded from the `get_from_changeset`/`get_to_changeset` logic (dead code — inner check was unreachable)
- [x] Verified that `integrations-publish` pipeline does NOT trigger on `feature/*` branches
- [x] Updated `docs/ci_pipelines.md` to document the new `feature/*` branch workflow

---

> This PR was drafted with the assistance of [Claude](https://claude.ai) (claude-sonnet-4-6).

### DIFF
--- a/.buildkite/scripts/common.sh
+++ b/.buildkite/scripts/common.sh
@@ -648,7 +648,7 @@ get_from_changeset() {
         from="${previous_successful_commit}"
         if [[ "${previous_successful_commit}" == "null" ]]; then
             if [[ "${BUILDKITE_BRANCH}" =~ ${LONG_RUNNING_BRANCH_PATTERN} ]]; then
-                # First push of a new backport or feature/* branch: only CI infrastructure files
+                # First push of a new long-running branch (matches LONG_RUNNING_BRANCH_PATTERN): only CI infrastructure files
                 # changed, no package code was modified — skip package testing.
                 from="${BUILDKITE_COMMIT}"
             else
--- a/.buildkite/scripts/common.sh
+++ b/.buildkite/scripts/common.sh
@@ -10,6 +10,8 @@ platform_type_lowercase="${platform_type,,}"
 
 SCRIPTS_BUILDKITE_PATH="${WORKSPACE}/.buildkite/scripts"
 
+readonly LONG_RUNNING_BRANCH_PATTERN="^(backport-|feature/)"
+
 export ELASTIC_PACKAGE_BIN=${WORKSPACE}/build/elastic-package
 
 API_BUILDKITE_PIPELINES_URL="https://api.buildkite.com/v2/organizations/elastic/pipelines/"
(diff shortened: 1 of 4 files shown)


.buildkite/scripts/common.sh
```diff
@@ -662,7 +662,7 @@ get_to_changeset() {
     # Changeset that triggered the build
     local to="${BUILDKITE_COMMIT}"
 
-    if [[ "${BUILDKITE_BRANCH}" == "main" || ${BUILDKITE_BRANCH} =~ ^backport- ]]; then
+    if [[ "${BUILDKITE_BRANCH}" == "main" || ${BUILDKITE_BRANCH} =~ ^(backport-|feature/) ]]; then
```
teresaromero: could we extract the pattern into a const so we reference it and keep the string only in one place?

teresaromero (approved):

.buildkite/scripts/common.sh
mrodm: Thanks!
Added a new variable (readonly) with this pattern for long running branches:
https://github.com/elastic/integrations/pull/18962/commits/54eab69bb4d6ae9543dbeff939d1d768cd6239de

I've also updated/added some comments

PR #18964, opened 2026-05-12
[Cisco ISE] Keep cisco-av-pair values with embedded 'equals' as keywords

haetamoudi:
## Proposed commit message

Keep cisco-av-pair values with embedded 'equals' as keywords

Changes:
Fixes the Cisco ISE passed authentications ingest pipeline so `cisco-av-pair` values that contain extra `=` characters are parsed correctly (kept as full keyword strings). Adds pipeline tests, field/docs updates, changelog entry, and a package version bump.

## Checklist


packages/cisco_ise/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.32.4"
+  changes:
+    - description: Keep cisco-av-pair values with embedded 'equals' as keywords.
```
Niceplace: Nitpick: IMHO, the changelog description should point out that the fix targets top-level keys that are not `mdm-tlv`. 

```suggestion
    - description: Handle the values of top-level keys under cisco-av-pair as a scalar mapped to a keyword field. Maintain separate process for mdm-tlv as it can contain sub-keys.
```

packages/cisco_ise/data_stream/log/elasticsearch/ingest_pipeline/pipeline_passed_authentications.yml
```diff
@@ -171,14 +171,30 @@ processors:
...
         ctx.cisco_ise.log.log_details.get("cisco-av-pair")?.forEach((v) -> {
+          def firstEq = v.indexOf('=');
+          if (firstEq <= 0) {
+            return true;
+          }
+          def topKey = v.substring(0, firstEq).trim();
+          def rest = v.substring(firstEq + 1);
+
+          // Only mdm-tlv uses nested subkeys (device-platform=win, ...). Other pairs
+          // (e.g. FQSubjectName=...cn=...,ou=...) may contain '=' in the value and must stay scalar.
```
Niceplace, line 182: Question: Would it make sense to link to official cisco docs or spec that would justify what top-level keys may contain `=` in the value and which ones may not ? This could help with future similar issues and possibly allow us to pre-emptively fix other fields that would be in the same situation.

Niceplace (approved):
Given we paired on this I feel comfortable giving my 👍 LGTM and all tests are green so that's a good sign.  I have one question and one nitpick, but nothing blocking.

vinit-chauhan (approved):

PR #18977, opened 2026-05-12
trend_micro_vision_one: reduce default batch_size to 200

### LINKED ISSUE #18975: trend_micro_vision_one: reduce default batch_size for CEL data streams
The `endpoint_activity` and `network_activity` CEL data streams both default `batch_size` to 1000, which is passed as the `top` query parameter to the Trend Micro Vision One search API. Agentless profiling shows the pod peaking at 575 MB, above a proposed 512 MB cgroup limit.

### Memory breakdown

The pod runs three separate processes sharing a single cgroup:

| Process | Binary | Heap in-use at snapshot |
|---------|--------|------------------------|
| elastic-agent | `elastic-agent` | 12 MB |
| CEL component | `elastic-otel-collector` | 44 MB |
| HTTPJSON component | `elastic-otel-collector` | 30 MB |

Total heap in-use is ~87 MB. The cgroup baseline is ~454 MB. The ~367 MB gap is structural overhead: binary text segments loaded three times, Go runtime per process (`runtime.allocm` alone is 12 MB per collector), `HeapSys` retention from previous allocation peaks, and `init()` allocations from linked-but-unused modules (AWS SDK endpoint tables, GCP protobuf descriptors, govmomi types, Apache Arrow, Envoy control plane, MongoDB drivers, Metricbeat modules).

Batch processing adds a transient ~102 MB heap peak on top of the 454 MB baseline, bringing the cgroup to ~575 MB. Reducing `batch_size` directly reduces this transient peak. It may also reduce the baseline over time by lowering `HeapSys` retention from smaller peaks, but this is unconfirmed.

### Change

Reduce the default `batch_size` from 1000 to a lower value (200 is a reasonable starting point) for both `endpoint_activity` and `network_activity`. Update the variable description to note that increasing the value above the default may cause out-of-memory events on agentless deployments with constrained memory limits.

The default should not be enforced as a hard cap. This integration runs on both agentless (constrained) and self-managed (relaxed) hosts. Users on hosts with sufficient memory should remain free to increase the value.

### Limitations

The integration-level `batch_size` knob addresses ~100 MB of a ~575 MB problem. The ~400 MB structural baseline is outside the integration's control. Whether reducing `batch_size` is sufficient to fit within 512 MB depends on whether the baseline also compresses with smaller peaks, which needs measurement.

The Trend Micro Vision One API documentation is login-gated, so we cannot verify what values of `top` the API accepts. If the API rejects or silently ignores small values, this will be visible in logs.

efd6:
## Proposed commit message
```
trend_micro_vision_one: reduce default batch_size to 200

Lower the default batch_size for endpoint_activity and
network_activity from 1000 to 200. The CEL input holds the entire
API response in memory during processing, and at 1000 the peak
allocation can push memory-constrained agents past their cgroup
limit. The description now warns that values above 200 risk
out-of-memory failures on such agents.

Closes #18975
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #18975

### DIFF
--- a/packages/trend_micro_vision_one/changelog.yml
+++ b/packages/trend_micro_vision_one/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.12.0"
+  changes:
+    - description: Reduce default batch_size from 1000 to 200 for endpoint_activity and network_activity data streams to lower peak memory use on memory-constrained agents.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/18977
 - version: "2.11.0"
   changes:
     - description: Update deprecated visualization charts and implemented best practices in the existing dashboards.
--- a/packages/trend_micro_vision_one/data_stream/endpoint_activity/_dev/test/policy/test-default.expected
+++ b/packages/trend_micro_vision_one/data_stream/endpoint_activity/_dev/test/policy/test-default.expected
@@ -128,7 +128,7 @@ inputs:
           resource.url: https://example.visionone.trendmicro.com
           state:
             api_token: ${SECRET_0}
-            batch_size: 1000
+            batch_size: 200
             initial_interval: 24h
           tags:
             - forwarded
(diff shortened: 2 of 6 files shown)


andrewkroh (approved):
LGTM. 

Existing installations are unaffected. Fleet's package_policy API copies the current input defaults into the policy at creation time, so only package policies created after this change will pick up the new batch_size default.

PR #18987, opened 2026-05-13
[rubrik] Move request templating out of CEL programs

chrisberkhout:
## Proposed commit message

```
[rubrik] Move request templating out of CEL programs

Move page size and filter configuration into CEL input state so the
program bodies no longer contain Handlebars templates. The programs now
read optional state values and build the GraphQL variables in CEL before
encoding the request body.

For filter configuration, pass the yaml variable directly into state and
normalize it in CEL. This keeps the template simple while allowing both
a bare list of filter objects and the older wrapped filter shape. Update
the filter defaults to show the bare list shape that the filters
variable is intended to hold.

Existing filter configs keep working because CEL still accepts values in
the filters state, while also accepting texts there to match the
documented example.

Use an empty filter list for the tasks stream, which does not declare a
filters variable, so its request body stays equivalent without carrying
unused state.
```

## Checklist

This makes `celfmt -extract -i data_stream/NAME/agent/stream/cel.yml.hbs` work.

It fixes the tasks stream to not use a variable that doesn't exist.

It fixes filter variable default values to show a format that works with with the previous (and new) usage.

## Checklist

### DIFF
--- a/packages/rubrik/changelog.yml
+++ b/packages/rubrik/changelog.yml
@@ -1,4 +1,10 @@
 # newer versions go on top
+- version: "0.9.4"
+  changes:
+    - description: |
+        Move request templating out of CEL programs.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/18987
 - version: "0.9.3"
   changes:
     - description: |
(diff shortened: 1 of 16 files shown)


ritalwar (approved):
LGTM!

PR #18989, opened 2026-05-13
osquery_manager: generate ECS section mappings from config

marc-gr:
## Proposed commit message

osquery_manager: generate ECS section mappings from config

Move the osquery-gen ECS keep list into `config.yml` and pass it through the generator instead of embedding a separate text file. This lets the generated `ecs.yml` include object/nested parent mappings such as `dll.pe.sections`, `file.pe.sections`, and related `threat.*.file.pe.sections` fields consistently when artifacts are regenerated.

Refresh generated osquery artifacts for osquery 5.23.0 and bump the package to 1.29.0 so the changelog captures both the ECS generation fix and osquery version update.

## Checklist

## Author's Checklist

- [x] Regenerated osquery_manager artifacts with osquery-gen.
- [x] Confirmed the PR #16650 ECS parent `*.sections` fields are generated in `data_stream/result/fields/ecs.yml`.

```bash
cd packages/osquery_manager/_dev/scripts/osquery-gen
go run . -config ./config.yml -skip-package-check
go test ./...
```

Also run:

```bash
git diff --check
```

Not run locally: full `elastic-package check`.

## Related issues

- Relates #16650

## Screenshots

N/A


packages/osquery_manager/_dev/scripts/osquery-gen/generator.go
```diff
@@ -550,12 +534,16 @@ func collectECSFieldNames(root any, parent string, keepFields map[string]struct{
...
+		if kept && name != "@timestamp" {
+			*ecsFields = append(*ecsFields, fieldName)
+		}
 		if childFields, hasFields := m["fields"]; hasFields {
 			collectECSFieldNames(childFields, joinPath(parent, name), keepFields, ecsFields)
 		} else {
-			fieldName := joinPath(parent, name)
 			typ, _ := m["type"].(string)
-			if (inSet(keepFields, fieldName) || isAllowedECSType(typ)) && name != "@timestamp" {
+			if !kept && isAllowedECSType(typ) && name != "@timestamp" {
```
tomsonpl, line 546: nit: the inverted condition (`!kept && isAllowedECSType(typ)`) reads as "skip the type check when kept," but the actual intent is "kept-list takes priority and also emits the parent object node; the leaf type check still applies to non-kept children." A one-line comment above the `kept` block on line 538 would save the next reader from re-deriving this.

Also, `name != "@timestamp"` is now checked on both branches — minor duplication, harmless, but could be lifted to a single guard at the top of the loop body.

packages/osquery_manager/_dev/scripts/osquery-gen/generator.go
```diff
@@ -550,12 +534,16 @@ func collectECSFieldNames(root any, parent string, keepFields map[string]struct{
...
 				*ecsFields = append(*ecsFields, fieldName)
```
tomsonpl, line 547: Suggest adding a unit test for `collectECSFieldNames` (or `generateECSFieldsYAML`) that feeds a synthetic ECS YAML with a `file.pe.sections`-shaped object node + a `keep_fields` containing that parent path, and asserts the parent appears in the output.

This is the substance of the fix and right now it's only validated indirectly via the regenerated `ecs.yml`. The repo already runs `go test ./...` for this tool, so a `generator_test.go` would be cheap and lock the contract in.

packages/osquery_manager/_dev/scripts/osquery-gen/main.go
```diff
@@ -93,7 +98,7 @@ func main() {
 	}
 
 	log.Printf("Resolved versions: osquery=%s beats=%s ecs=%s", osqueryVersion, beatsRef, ecsVersion)
-	if err := generateArtifacts(outputRoot, osqueryVersion, ecsVersion, beatsRef, !*skipPackageCheck); err != nil {
+	if err := generateArtifacts(outputRoot, osqueryVersion, ecsVersion, beatsRef, cfg.ECS.KeepFields, !*skipPackageCheck); err != nil {
```
tomsonpl, line 101: If `ecs.keep_fields` is omitted from `config.yml`, `cfg.ECS.KeepFields` is `nil` and behavior silently degrades to "leaves + `isAllowedECSType` only" — i.e. the parent-object fix this PR introduces disappears with no warning.

A `log.Printf("ECS keep fields: %d entries", len(cfg.ECS.KeepFields))` line near the existing `Resolved versions:` log would catch accidental empty configs in CI output.

packages/osquery_manager/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "1.29.0"
+  changes:
+    - description: Generate ECS `*.sections` parent mappings from osquery-gen config
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/16650
+    - description: Upgrade osquery version to 5.23.0
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/16650
```
tomsonpl, line 9: Minor: the "Upgrade osquery version to 5.23.0" entry links to #16650, but #16650 tracks the ECS `*.sections` mapping issue, not the osquery version bump. Consider linking this PR (#18989) itself, or leaving it without a link if there's no standalone tracking issue for the osquery bump.

tomsonpl (approved):
Solid fix for a real bug (#16650 follow-through). The generator refactor is clean and the config-driven model is consistent with the rest of the tool. Left a few comments inline — main ask is (1) a unit test covering the parent-object keep-list path.

PR #19000, opened 2026-05-14
abnormal_security: skip deleted items during detail fetch instead of halting

### LINKED ISSUE #18968: Abnormal AI Security Mailbox CEL code fails when abuse campaign GUID query returns 404
There is a problem with the `abnormal_security.ai_security_mailbox` stream.  Like all other Abnormal AI CEL-based data collectors, this stream takes a two-stage query process:

1. Query the base API to obtain a "work list"
2.  Query the API for each GUID in the work list to obtain details on the work list item.

There is a problem at line 195 of the `cel.yml.hbs` Elastic Agent code for this stream.  The problem occurs when the API returns an HTTP 404 error when querying for a GUID that was gathered in step 1, but no longer exists when querying for the work item:
<https://github.com/elastic/integrations/blob/4edf61cd8123fba0f4feffe146a5ef2c320c8f07/packages/abnormal_security/data_stream/ai_security_mailbox/agent/stream/cel.yml.hbs#L195>

When a 404 occurs, the CEL code declares `"want_more": false` and the query loop terminates, leaving the work list unprocessed.

We observed this same problem previously in the `abnormal_security.threats` stream.  The fix is to apply some guard code for 404 responses so that the CEL program does not hang on deleted/missing security mailbox items.  We have applied the following patch code in our environment, but a permanent fix is needed so that future updates to the integration will no longer have this bug:

```cel
: // Campaign deleted before we could fetch it — skip silently.
  (resp.StatusCode == 404) ?
    {
      "events": [{"retry": true}],
      "cursor": {
        "scanning": state.?cursor.scanning.orValue([]),
      },
      "worklist": (int(state.next) + 1 < size(state.worklist)) ? state.worklist : [],
      "next": (int(state.next) + 1 < size(state.worklist)) ? (int(state.next) + 1) : 0,
      "next_page": (int(state.next) + 1 < size(state.worklist)) ? state.next_page : (int(state.next_page) + 1),
      "want_more": true,
    }
```

I also think it would be beneficial to review all of the Abnormal AI streams to ensure that adequate guard code exists to prevent this same problem from hanging up processing of other event streams.

efd6:
## Proposed commit message
```
abnormal_security: skip deleted items during detail fetch instead of halting

When the Abnormal Security API returns HTTP 404 for a worklist item
that was deleted between the list and detail fetch, the CEL program
now skips the item and continues processing the remaining worklist.

Previously, ai_security_mailbox, case, and vendor_case streams
treated any non-200 detail response as fatal, setting want_more to
false and halting collection for the entire worklist. The threat
stream already continued past errors but still emitted a noisy
error event for 404s.

All four worklist-based streams now emit a dummy retry event on 404
(dropped by a drop_event processor) so the input's publish-ack
cycle completes promptly instead of waiting for the next periodic
evaluation. The case, vendor_case, and threat streams gain the
drop_event processor that ai_security_mailbox already had.

System test mock configs include phantom worklist items that return
404, verifying the skip-and-continue behaviour as a regression test.

Fixes #18968
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #18968


packages/abnormal_security/data_stream/case/agent/stream/cel.yml.hbs
```diff
@@ -92,6 +92,18 @@ program: |
             "next_page": int(state.next) + 1 < size(state.worklist) ? state.next_page : int(state.next_page) + 1,
             "want_more": true,
           })
+        : resp.StatusCode == 404 ?
+          // Case deleted before detail fetch: skip.
+          {
+            "events": [{"retry": true}],
```
efd6, line 98: Note that we're changing behaviour here; we are now dropping 404s. I'm open to reverting this behaviour since there are arguments both ways.

We do get 4xx metrics, so the information is mostly still available (all 4xx rather than a specific count of 404s).

packages/abnormal_security/data_stream/threat/_dev/test/scripts/threat_detail_404.txt
chrisberkhout, line 1: Maybe this can be removed? A system test should do the job because we don't get the 404 error documents anymore.

chrisberkhout (approved):
Looks good.

I don't have a strong opinion about whether to have 404 docs in this case. Skipping them seems fine.

The CEL programs could be have nicer worklist handling, but that's beyond the scope of this fix. I mean taking items from the worklist as they are processed rather than indexing into it, and simplifying how the `next_page` number is advanced.

packages/abnormal_security/data_stream/threat/_dev/test/scripts/threat_detail_404.txt
efd6: Done

PR #19002, opened 2026-05-14
aws: fix config system test (mock response + expired TLS cert)

### LINKED ISSUE #18713: [Stack 9.5.0-SNAPSHOT] [aws] Failing test daily: system test: default in aws.redshift
- Stack version: 9.5.0-SNAPSHOT
- Package: aws
- Failing test: system test: default
- DataStream: redshift
- Owners:
    - @elastic/obs-infraobs-integrations

Error:
```
could not setup service: Terraform deployer is unhealthy: container (ID: 61c40976e777) exited with code 1
```

First build failed: https://buildkite.com/elastic/integrations/builds/42098

stefans-elastic:
## Proposed commit message

aws/config: handle null EvaluationResults and renew mock TLS cert

Fixes two independent causes of system test failures in the
aws.config data stream.

When GetComplianceDetailsByConfigRule returns null or empty
EvaluationResults, the CEL program previously forwarded bare
events to the ingest pipeline. These lacked an event.original
field, causing pipeline_error documents that broke the
hit_count: 4 assertion. The CEL program now emits a
{"message": "retry"} sentinel for null/empty results; a
drop_event processor discards them before indexing.

The aws.config mock server's self-signed certificate had a
1-year validity and expired on 2026-05-06. Regenerated with
10-year validity (expires 2036-05-13), matching the pattern
used by guardduty, inspector, and securityhub mock servers.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes #18713


packages/aws/data_stream/config/_dev/deploy/docker/files/config.yml
```diff
@@ -200,7 +200,7 @@ rules:
         body: |-
           {{ minify_json `
           {
-            "EvaluationResults": null
+            "EvaluationResults": []
```
efd6: Please don't do this. This changes the behaviour of the mock to make it less restrictive, allowing tests to pass when the real data may cause the actual input to fail. The correct fix is to change the CEL program to tolerate the values that can be here. From a survey of the AWS documentation for this document, `EvaluationResults` does not have a `required` constraint, so it may be absent completely (the documentation says that this is what *should* happen, but by Postel, we should probably tolerate a `null` or a `[]`.

Instead, drop the event.
```diff
diff --git a/packages/aws/data_stream/config/agent/stream/cel.yml.hbs b/packages/aws/data_stream/config/agent/stream/cel.yml.hbs
index faab5eb989..6101af66f5 100644
--- a/packages/aws/data_stream/config/agent/stream/cel.yml.hbs
+++ b/packages/aws/data_stream/config/agent/stream/cel.yml.hbs
@@ -322,7 +322,7 @@ program: |
                 }
               )
             :
-              [{}],
+              [{"retry":true}],
             "next_page": {
               ?"result_token": body.?NextToken,
               ?"rule_token": config_rules.?next_page.rule_token,
@@ -401,7 +401,8 @@ tags:
 {{#contains "forwarded" tags}}
 publisher_pipeline.disable_host: true
 {{/contains}}
-{{#if processors}}
 processors:
+- drop_event.when.equals.retry: true
+{{#if processors}}
 {{processors}}
 {{/if}}
```

stefans-elastic: @efd6 thank you for your comment. I had to tweak the fix a little as it refused working with boolean type. In the end I've done it as it is done in [rapid7_insightvm](https://github.com/elastic/integrations/blob/e7a1cd483f0bad86c22f735ab4f1acd6f0af00ec/packages/rapid7_insightvm/data_stream/asset_vulnerability/agent/stream/cel.yml.hbs#L259) package.

efd6: The alternative is to convert the type to a `dyn` to get type agreement, `dyn([{"retry":true}])`, but this will do.

efd6:
Please update the proposed commit message for the new changes.

stefans-elastic:
> Please update the proposed commit message for the new changes.

thanks for reminding me (I keep forgetting to do it).
updated

efd6 (approved):
Thanks

PR #19013, opened 2026-05-14
imperva_cloud_waf: fix CEL type guard for events field access

chemamartinez:
## Proposed commit message

```
imperva_cloud_waf: fix CEL type guard for events field access

The guard has(v.?events.error) at the outer .as(v, ...) expression is
intended to detect error responses, where events is a map
{"error": {...}}. When the worklist has multiple entries, the
worklist-continuation branch returns state wholesale, carrying the
previous iteration's list-valued events into v. CEL then attempts
string-key field access (.error) on that list and throws "unsupported
index type 'string' in list", cycling the input unit HEALTHY→DEGRADED
on every poll and halting ingestion entirely.

This is triggered in practice when the Imperva API delivers non-WAF
event types (e.g. Client-Side Protection SIEM events) alongside WAF
logs through the same endpoint, as the additional log volume increases
the likelihood of a multi-file worklist.

Add an explicit type check before the field access so that .error is
only accessed when v.events is actually a map.
```

## Checklist

### DIFF
--- a/packages/imperva_cloud_waf/data_stream/event/agent/stream/cel.yml.hbs
+++ b/packages/imperva_cloud_waf/data_stream/event/agent/stream/cel.yml.hbs
@@ -67,7 +67,7 @@ program: |
           "password": state.password,
          }
       )
-  ).as(v, has(v.?events.error) ?
+  ).as(v, has(v.?events) && type(v.events) == map && has(v.?events.error) ?
     v
   : v.next < size(v.worklist) ?
     (
--- a/packages/imperva_cloud_waf/data_stream/event/agent/stream/cel.yml.hbs
+++ b/packages/imperva_cloud_waf/data_stream/event/agent/stream/cel.yml.hbs
@@ -67,7 +67,7 @@ program: |
           "password": state.password,
          }
       )
-  ).as(v, has(v.?events) && type(v.events) == map && has(v.?events.error) ?
+  ).as(v, type(v.?events.orValue(null)) == map && has(v.events.error) ?
     v
   : v.next < size(v.worklist) ?
     (
--- a/packages/imperva_cloud_waf/changelog.yml
+++ b/packages/imperva_cloud_waf/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.13.2"
+  changes:
+    - description: Fix CEL type guard to handle `events` as a list type.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/99999
 - version: "1.13.1"
   changes:
     - description: Downgrade the `format_version` to the minimum version that supports all the necessary features for the package.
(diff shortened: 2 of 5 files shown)


packages/imperva_cloud_waf/data_stream/event/agent/stream/cel.yml.hbs
```diff
@@ -67,7 +67,7 @@ program: |
           "password": state.password,
          }
       )
-  ).as(v, has(v.?events.error) ?
+  ).as(v, has(v.?events) && type(v.events) == map && has(v.?events.error) ?
```
chrisberkhout: I think the optional syntax is redundant. Checking field presence should be enough.
```suggestion
  ).as(v, has(v.events) && type(v.events) == map && has(v.events.error) ?
```

chrisberkhout (changes requested):
The commit message seems to contradict itself:

> On the success path **events is always a list**, so the check is normally safe.
> However, when ..., **v.events can be a list** at the point of the guard.

The change of checking the type of the `events` makes sense.

packages/imperva_cloud_waf/data_stream/event/agent/stream/cel.yml.hbs
efd6: I think (tested) this test can be reduced to `type(v.?events.orValue(null)) == map`; if it's absent, the type is null, if it present then the type gets made concrete and we can compare to the literal type.

chrisberkhout (commented):
CEL change looks good.

For a second I wondered if I was wrong about removing the optional syntax on the latter part of the condition, but that seems fine:

```
mito -data <(echo '{"events": []}')            <(echo 'type(state.?events.orValue(null)) == map && has(state.events.error)')
mito -data <(echo '{"events": {"error": {}}}') <(echo 'type(state.?events.orValue(null)) == map && has(state.events.error)')
mito -data <(echo '{"events": null}')          <(echo 'type(state.?events.orValue(null)) == map && has(state.events.error)')
```
```
false
true
false
```

The build error is:
> test case failed: one or more errors found in document: [0] field "source.geo.location.coordinates" is undefined, could be a multifield
[1] field "source.geo.location.type" is undefined, could be a multifield

chemamartinez:
> CEL change looks good.
> 
> For a second I wondered if I was wrong about removing the optional syntax on the latter part of the condition, but that seems fine:
> 
> ```
> mito -data <(echo '{"events": []}')            <(echo 'type(state.?events.orValue(null)) == map && has(state.events.error)')
> mito -data <(echo '{"events": {"error": {}}}') <(echo 'type(state.?events.orValue(null)) == map && has(state.events.error)')
> mito -data <(echo '{"events": null}')          <(echo 'type(state.?events.orValue(null)) == map && has(state.events.error)')
> ```
> 
>  ```
> false
> true
> false
> ```
> 
> The build error is:
> 
> > test case failed: one or more errors found in document: [0] field "source.geo.location.coordinates" is undefined, could be a multifield
> > [1] field "source.geo.location.type" is undefined, could be a multifield

Thanks for checking, exactly, regenerating `sample_event.json` created some non-ECS fields. I've restored the old version as it is unrelated to this change.

chrisberkhout (approved):

efd6 (approved):

PR #18994, opened 2026-05-13
Add nginx_integration_otel package

teresaromero:
## Proposed commit message

```
Add nginx_integration_otel package

Introduce the nginx_integration_otel package — an OTel-based integration
that collects Nginx access/error logs via the filelog_otel input package,
stub_status metrics via the nginx_otel_input input package, and declares
a dependency on the nginx_otel content package for dashboards and related
assets.

The package ships three data streams (access, error, stubstatus) each
with manifests, base fields, field definitions, sample events, policy
expectations, system test wiring, and Docker-based _dev deploy assets.
Documentation describes OTel-based collection setup end-to-end.

Package dependencies:
  - filelog_otel input  v0.2.0
  - nginx_otel_input input v0.2.0
  - nginx_otel content v0.5.0

Requires Kibana ≥ 9.4.0. This work supports Fleet dependency resolution
for composable/OTel integrations and is tracked alongside ingest-dev work
on composable packages.
```

## Checklist

- [ ] I have reviewed [tips for building integrations](https://www.elastic.co/docs/extend/integrations/tips-for-building) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) (integration pulls dashboards via `nginx_otel` content dependency rather than shipping dashboards in-repo).

## Author's Checklist

- [ ] Replace the placeholder PR link in `packages/nginx_integration_otel/changelog.yml` (`# FIXME Replace with the real PR link`) with this pull request's URL once merged.
- [ ] Confirm whether dataset naming alignment between `nginx_otel` content package assets and this integration is expected to be handled in a follow-up (see manual testing notes below).

## How to test this PR locally

1. Build and validate the package with [elastic-package](https://github.com/elastic/elastic-package) from the integrations repo:
   ```
   elastic-package lint
   elastic-package test
   ```

2. **Manual testing (Kibana + Fleet dependency resolution)**
   - In `kibana.dev.yml`, set:
     ```yaml
     xpack.fleet.experimentalFeatures.enableResolveDependencies: true
     ```
[...]

PR #19024, opened 2026-05-15
Add updatecli workflow to auto-bump package requires versions from EPR

### LINKED ISSUE #17404: Automatically update package dependencies
In the context of the composable packages project, integration packages are going to start having dependencies on other packages (see https://github.com/elastic/package-spec/issues/1083).

It would be good to have a way to automate updates, so when a new input package is published, the integration packages that depend on it are also updated automatically.

For convenience, it would be good if the changes on packages are grouped in PRs by owner.

teresaromero:
## What

Adds a GitHub Actions workflow that automatically bumps pinned dependency versions in integration package `requires:` blocks and creates changelog entries for affected packages.

The workflow runs two updatecli stages in sequence:

1. **bump-requires** — scans `packages/*/manifest.yml` for `requires:` blocks, checks each pinned `input`/`content` package version against EPR, and updates stale pins in place. Changes are recorded to a JSON file consumed by the next stage.

2. **add-changelogs** — for each bumped package, calls `elastic-package changelog add` with a bump level derived from the highest semver level across all dep bumps for that package (major → minor → patch). Packages still at v0.1.0 are capped at minor to avoid jumping to 1.0.0 before the first release.

On pull requests the workflow runs in `diff` mode (dry-run, no commits). On schedule (`0 1 * * 1-5`) and `workflow_dispatch` it runs in `apply` mode and opens or updates a PR.

## Why

`requires:` version pins were bumped manually. As more packages adopt this pattern the process becomes error-prone without automation.

## How to test

```bash
updatecli --experimental diff --config .github/workflows/updatecli/package-requires.d --values .github/workflows/updatecli/values.d/scm.yml
```

To force a bump: temporarily lower a version pin in a `packages/*/manifest.yml`, re-run, confirm the bump and changelog entry are reported. Revert with `git checkout packages/`.

## Related issues

- closes https://github.com/elastic/integrations/issues/17404


teresaromero:
@jsoriano tested this locally using the branch of https://github.com/elastic/integrations/pull/18994. the issue suggested to create a pr per owner, i've thought that as we increase the integrations using this requirements, we could open a single pr with the updates, this will likely tag the teams involved on the changes? should we update changelog for each integration updated?

jsoriano:
> we could open a single pr with the updates, this will likely tag the teams involved on the changes?

Yes, and this may be a problem, requiring the review of many teams may delay the updates also for the teams that review quickly. Separating by team lets each team have their own pace.

In any case we can start with a single PR, and make the change to one per team later if we find problems.

> should we update changelog for each integration updated?

Yes, this is a change that potentially modifies the built packages or the assets they use. A changelog entry for a new version should be used.

We will need to decide if it is a new patch or a new minor, or even a new major, maybe we could base it on the version updated in the dependency. If the update is because of a new minor in the dependency then the new version for the affected package will also be a new minor.

`elastic-package changelog add` has flags to help adding changelog entries to next patch, minor, or major.

.github/workflows/updatecli/package-requires.d/bump-package-requires-versions.yml
```diff
@@ -0,0 +1,61 @@
...
+                # Match "package: <pkg>" line then replace version on the next line.
+                sed -i.bak "/package: ${pkg}[[:space:]]*$/{n; s/version: \"[^\"]*\"/version: \"${latest}\"/;}" "$manifest"
+                rm -f "${manifest}.bak"
+              fi
+              i=$((i + 1))
+            done
+          done
+        done
+
+        git diff --name-only HEAD
```
jsoriano: Consider moving the script to its own file if possible.

.github/workflows/bump-package-requires-versions.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: echo "UPDATECLI_ACTION=diff" >> $GITHUB_ENV
+
+      - name: Select apply action
+        if: ${{ github.event_name != 'pull_request' }}
+        run: echo "UPDATECLI_ACTION=apply" >> $GITHUB_ENV
+
+      - name: Bump package requires versions
+        run: updatecli ${{ env.UPDATECLI_ACTION }} --config .github/workflows/updatecli/package-requires.d
+
+      - name: Create Pull Request
```
jsoriano: updatecli should be able to open the PR itself with the `openPullRequest` action.

teresaromero: i separated the flow like that to be able to run the diff locally with the cli

teresaromero: updated this so it runs with updatecli options

.github/workflows/bump-package-requires-versions.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Install Updatecli in the runner
+        uses: updatecli/updatecli-action@2c3221bc5f4499a99fec2c87d9de4a83cb30e990 #v3.1.3
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: '.go-version'
+
+      - name: Install elastic-package
+        run: go install github.com/elastic/elastic-package@$(grep 'github.com/elastic/elastic-package' go.mod | awk '{print $2}')
```
mrodm, line 60: If this is running in the working directory, it could be run directly 
```
go install  github.com/elastic/elastic-package
```

And that command should install the version defined in the go.mod

.github/scripts/package-requires/bump-package-requires-versions.sh
```diff
@@ -0,0 +1,93 @@
...
+    printf 'minor'
+  elif [ "$from_patch" != "$to_patch" ]; then
+    printf 'patch'
+  else
+    printf 'none'
+  fi
+}
+
+# Discover integration package manifests that have a top-level requires: block.
+manifests=$(grep -l '^requires:' packages/*/manifest.yml 2>/dev/null || true)
```
mrodm, line 45: There could be packages defined in nested directories.

You could try to install also `mage` and run `mage listPackages`. As it has been done for the workflow in charge of syncing issue templates:

https://github.com/elastic/integrations/blob/80e62b9dbdddeea379baf661786ff75275822f45/.github/workflows/updatecli/updatecli.d/sync-packages-to-bug-issue-template.yml#L34


https://github.com/elastic/integrations/blob/80e62b9dbdddeea379baf661786ff75275822f45/.github/workflows/bump-elastic-stack-version.yml#L32-L33

.github/scripts/package-requires/bump-package-requires-versions.sh
```diff
@@ -0,0 +1,93 @@
...
+  result=$(printf '%s' "$epr_input" | yq ".[] | select(.name == \"$1\") | .version" | head -n1)
+  if [ -z "$result" ] || [ "$result" = "null" ]; then
+    result=$(printf '%s' "$epr_content" | yq ".[] | select(.name == \"$1\") | .version" | head -n1)
+  fi
+  printf '%s' "$result"
+}
+
+# Returns the semver bump level (major|minor|patch|none) between two version strings.
+# Strips leading non-numeric prefix (^, ~, v) before comparing.
+semver_level() {
```
mrodm, line 24: Define the variables used in this function as local.

```
semver_level() {
  local from
  local to
  local from_major
  local from_minor
  ...
```

.github/scripts/package-requires/bump-package-requires-versions.sh
```diff
@@ -0,0 +1,93 @@
...
+# type= filters reduce the payload from ~457 to ~86 entries combined.
+# all=false returns one entry per package (the latest version).
+# prerelease=true includes RC/beta releases.
+epr_input=$(curl -fsSL "https://epr.elastic.co/search?prerelease=true&all=false&type=input")
+epr_content=$(curl -fsSL "https://epr.elastic.co/search?prerelease=true&all=false&type=content")
+
+# Look up the latest EPR version for a given package name.
+# Two separate lookups avoid multi-document parsing ambiguity when
+# concatenating the two JSON arrays without a YAML --- separator.
+epr_latest() {
```
mrodm, line 14: Nit. Define variables in this function as local.

```suggestion
epr_latest() {
  local result=""
```

.github/scripts/package-requires/bump-package-requires-versions.sh
```diff
@@ -0,0 +1,93 @@
...
+  result=$(printf '%s' "$epr_input" | yq ".[] | select(.name == \"$1\") | .version" | head -n1)
+  if [ -z "$result" ] || [ "$result" = "null" ]; then
+    result=$(printf '%s' "$epr_content" | yq ".[] | select(.name == \"$1\") | .version" | head -n1)
+  fi
+  printf '%s' "$result"
```
mrodm, line 19: Why not using echo directly ?

.github/scripts/package-requires/bump-package-requires-versions.sh
```diff
@@ -0,0 +1,93 @@
...
+
+for manifest in $manifests; do
+  pkg_name=$(basename "$(dirname "$manifest")")
+  for kind in input content; do
+    count=$(yq ".requires.${kind} | length // 0" "$manifest")
+    i=0
+    while [ "$i" -lt "$count" ]; do
+      pkg=$(yq ".requires.${kind}[$i].package" "$manifest")
+      current=$(yq ".requires.${kind}[$i].version" "$manifest")
+      latest=$(epr_latest "$pkg")
```
mrodm, line 58: This latest version should be compatible with the given kibana constraint of the package too, no ?

.github/scripts/package-requires/bump-package-requires-versions.sh
```diff
@@ -0,0 +1,93 @@
...
+
+      if [ -z "$latest" ] || [ "$latest" = "null" ]; then
+        echo "WARN: ${pkg} not found on EPR" >&2
+      elif [ "$current" != "$latest" ]; then
```
mrodm, line 62: It could be re-ordered the `ifs` to avoid nested ifs.


```sh
      if [ -z "$latest" ] || [ "$latest" = "null" ]; then
        echo "WARN: ${pkg} not found on EPR" >&2
        continue
      if
      
      if [[ "$current" = "$latest" ]]; then
        continue
      fi

      # Bump version
```

.github/scripts/package-requires/bump-package-requires-versions.sh
```diff
@@ -0,0 +1,93 @@
...
+        level=$(semver_level "$current" "$latest")
+        if [ "$bumps_first" = "1" ]; then
+          bumps_first=0
+          printf '[\n  {"pkg":"%s","dep":"%s","kind":"%s","from":"%s","to":"%s","level":"%s"}' \
+            "$pkg_name" "$pkg" "$kind" "$current" "$latest" "$level" > "$bumps_tmp"
+        else
+          printf ',\n  {"pkg":"%s","dep":"%s","kind":"%s","from":"%s","to":"%s","level":"%s"}' \
+            "$pkg_name" "$pkg" "$kind" "$current" "$latest" "$level" >> "$bumps_tmp"
+        fi
+      fi
```
mrodm, line 78: Could this logic be moved to another function ? Maybe update_requirement or similar ?

.github/workflows/bump-package-requires-versions.yml
```diff
@@ -0,0 +1,77 @@
...
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Install Updatecli in the runner
+        uses: updatecli/updatecli-action@2c3221bc5f4499a99fec2c87d9de4a83cb30e990 #v3.1.3
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
```
mrodm, line 55: There is a newer version v6 also used in other workflows:

https://github.com/elastic/integrations/blob/80e62b9dbdddeea379baf661786ff75275822f45/.github/workflows/bump-elastic-stack-version.yml#L28


```suggestion
        uses: actions/setup-go@v6
```

.github/scripts/package-requires/add-changelogs.sh
```diff
@@ -0,0 +1,80 @@
...
+entry_count=$(jq 'length' "$BUMPS_JSON")
+if [ "$entry_count" -eq 0 ]; then
+  echo "Bumps JSON is empty; skipping changelog step."
+  exit 0
+fi
+
+pkg_names=$(jq -r '.[].pkg' "$BUMPS_JSON" | sort -u)
+
+for pkg_name in $pkg_names; do
+  manifest="packages/$pkg_name/manifest.yml"
```
mrodm, line 30: There could be packages that could be located in nested directories.

Relates to https://github.com/elastic/integrations/issues/17403

mrodm: Maybe using `mage listPackages` and look for the right path there ?

.github/scripts/package-requires/add-changelogs.sh
```diff
@@ -0,0 +1,80 @@
...
+  has_major=$(jq "[.[] | select(.pkg == \"$pkg_name\" and .level == \"major\")] | length" "$BUMPS_JSON")
+  has_minor=$(jq "[.[] | select(.pkg == \"$pkg_name\" and .level == \"minor\")] | length" "$BUMPS_JSON")
+
+  if [ "$has_major" -gt 0 ]; then
+    max_level="major"
+  elif [ "$has_minor" -gt 0 ]; then
+    max_level="minor"
+  else
+    max_level="patch"
+  fi
```
mrodm, line 48: Would this mean that if the update of the requirement is from "3.1.0" to "4.0.0" would involve bumping a major version in the package too ?

Should we update just versions of the required packages that do not bump the major version of the package for now ? @jsoriano

.github/workflows/bump-package-requires-versions.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: echo "UPDATECLI_ACTION=apply" >> $GITHUB_ENV
+
+      - name: Bump package requires versions
+        # --experimental is required for the commitusingapi SCM option.
+        run: updatecli --experimental pipeline ${{ env.UPDATECLI_ACTION }} --config .github/workflows/updatecli/package-requires.d --values .github/workflows/updatecli/values.d/scm.yml
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BUMPS_OUTPUT: ${{ runner.temp }}/bumps.json
+          BUMPS_JSON: ${{ runner.temp }}/bumps.json
```
mrodm, line 77: When running on the schedule basis, I think it would trigger all the workflows again as in this other workflow:

https://github.com/elastic/integrations/blob/80e62b9dbdddeea379baf661786ff75275822f45/.github/workflows/bump-elastic-stack-version.yml#L48-L52

WOuld it be needed to merge both workflows in one?

.github/scripts/package-requires/bump-package-requires-versions.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/sh
+set -eu
+
+# Fetch input and content packages from EPR (two targeted calls).
+# type= filters reduce the payload from ~457 to ~86 entries combined.
+# all=false returns one entry per package (the latest version).
+# prerelease=true includes RC/beta releases.
+epr_input=$(curl -fsSL "https://epr.elastic.co/search?prerelease=true&all=false&type=input")
+epr_content=$(curl -fsSL "https://epr.elastic.co/search?prerelease=true&all=false&type=content")
```
mrodm, line 9: To avoid issues with these requests it would be good to add some kind of retry mechanism here.

In buildkite we have this function as a helper

https://github.com/elastic/integrations/blob/80e62b9dbdddeea379baf661786ff75275822f45/.buildkite/scripts/common.sh#L31-L48

Example of usage:
https://github.com/elastic/integrations/blob/80e62b9dbdddeea379baf661786ff75275822f45/.buildkite/scripts/backport_branch.sh#L55

.github/scripts/package-requires/add-changelogs.sh
teresaromero: The initial proposal was to update 1:1 (patch to patch, minor to minor) except for majors, which was thought to be mapped to a minor update on the package. This is a miss implementation here.

I think versions should be bumped for all dependencies updates... whether that is done with the script or not i guess is up to the developer, as releasing a major version of the package should involve more changes than a dependency.

.github/workflows/bump-package-requires-versions.yml
teresaromero: which "both"  workflows?; you mean merging the stack version workflow with this one? or chaning the schedule of them? i used this as base for the new workflow to update requires, but schedule can be modified

.github/scripts/package-requires/bump-package-requires-versions.sh
teresaromero: >For portable #!/bin/sh, keep printf so JSON and version strings are emitted exactly (no extra newline, no -/escape quirks) and formatted output like bumps.json stays reliable.

it makes sense to keep printf so the format is not changed

ref https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html

.github/workflows/bump-package-requires-versions.yml
mrodm: Ah! sorry. I stand corrected. I thought this new pipeline was added into the same config folder as the current workflow `--config .github/workflows/updatecli/updatecli.d ` , but you are adding a new folder `--config .github/workflows/updatecli/package-requires.d`.

Given that, this workflow you're adding here and the one already defined in the repository ([link](https://github.com/elastic/integrations/blob/e0a9d9877f230e5ddc91ac5eb5a867a8bd98821c/.github/workflows/bump-elastic-stack-version.yml)) will be independent.

:+1:

teresaromero:
closing this in favor of https://github.com/elastic/integrations/pull/19217

PR #19033, opened 2026-05-15
[Pfsense] The fix adds null checks before calling `.toLowerCase()` on `ctx._tmp?.action` in two `set` processors within the php-fpm ingest pipeline.

haetamoudi:
**Suggested label:** Enhancement _(pick the matching GitHub label)_
## Proposed commit message

**Squash subject line** (from Sentinel / primary PR):
```
[Pfsense] The fix adds null checks before calling `.toLowerCase()` on `ctx._tmp?.action` in two `set` processors within the php-fpm ingest pipeline.
```
## Executive summary
The fix adds null checks before calling `.toLowerCase()` on `ctx._tmp?.action` in two `set` processors within the php-fpm ingest pipeline. Previously, when the `action` field was absent from a grok capture (e.g., for configuration change log lines), the pipeline would throw a null pointer error because the safe-navigation operator `?.` only prevents NPE on the field access itself, not on subsequent method calls. A new test case was added covering a configuration change log line where `action` is not present, confirming the pipeline processes such events without error.
## Root cause
In php-fpm.yml, the Painless `?.` null-safe operator only short-circuits on `ctx._tmp` itself; when `ctx._tmp` exists (holding `tz_offset` from the default pipeline) but `ctx._tmp.action` is null (because the message matched the grok fallback `%{GREEDYDATA}` pattern instead of a login/logout/error pattern), `ctx._tmp?.action` evaluates to null and the chained `.toLowerCase()` call immediately throws a NullPointerException before `.contains()` is ever reached.
## Approach
In php-fpm.yml, extend null-safe guarding on both `set` processors that check `ctx._tmp?.action.toLowerCase()`. The `?.` only null-guards `ctx._tmp` itself; if `_tmp` exists (e.g., from `tz_offset` set by the default pipeline) but `_tmp.action` was never populated by grok (because the message matched the fallback `%{GREEDYDATA}` pattern instead of a login/logout/error pattern), then `ctx._tmp?.action` evaluates to null and the subsequent `.toLowerCase()` throws a NullPointerException. Fix both conditions using explicit null checks: `ctx._tmp?.action != null && ctx._tmp.action.toLowerCase().contains(...)`. Also add the sanitized event (a Configuration Change php-fpm message with no login/logout/error pattern) as a new test case to prevent regression.
## Pipeline changes
- Modify set processor tag=set_event_outcome_b39ddcc4: change `if` condition from `ctx._tmp?.action.toLowerCase().contains("success")` to `ctx._tmp?.action != null && ctx._tmp.action.toLowerCase().contains("success")`
- Modify set processor tag=set_event_outcome_83ced4a6: change `if` condition from `ctx._tmp?.action.toLowerCase().contains("authentication error")` to `ctx._tmp?.action != null && ctx._tmp.action.toLowerCase().contains("authentication error")`
## Field / mapping changes
—
## Sanitized log (`event_sanitized` excerpt)
```json
<27>1 2026-05-12T19:57:53.381917+08:00 host-1.example.local php-fpm 18801 - - /firewall_rules.php: Configuration Change: alice.johnson@198.51.100.10 (Local Database): Firewall: Rules - disabled a firewall rule.
```
## Reviewer concerns
- The new test case entry is missing `event.outcome` in the expected output, which is correct since the action field is absent — but reviewers should confirm this matches expected behavior (no outcome set for config-change events).
- The expected output for the new log line includes `event.category: ["authentication"]` — worth verifying whether a configuration change event should really be categorized as authentication rather than `configuration` or `change`.
## Risk and classification
- **Plan risk level**: low
- **Tags**: pipeline, processors, test-fixture, ingest
- **Impact**: medium
## Links
- **Issue**: (no issue number)
- **Issue title**: pfsense.log [PIPELINE_FIX]: Processor 'set' with tag 'set_event_outcome_b39ddcc4' in pipeline 'logs-…
- **Pipeline case**: `23dde1681035b4d6` (Integration Sentinel)
---
_This pull request was opened from the Integration Sentinel agent fix flow._
## Checklist
## Author's Checklist

- [ ] 

Add the sanitized RFC5424 php-fpm Configuration Change event to test-pfsense-phpfpm.log and its expected JSON counterpart (no event.outcome, no source fields, host.name from observer.name rename, event.category=["authentication"]). Run `elastic-package test pipeline` against the pfsense package to confirm all four phpfpm test events pass. Verify the three existing cases still produce the correct event.outcome values (success/success/failure) and the new case produces no event.outcome.
From a checkout of this branch:
```bash
cd packages/pfsense
elastic-package test
```
## Screenshots

_None attached from Sentinel — add if applicable._


packages/pfsense/data_stream/log/elasticsearch/ingest_pipeline/php-fpm.yml
```diff
@@ -21,12 +21,12 @@ processors:
       tag: set_event_outcome_b39ddcc4
       field: event.outcome
       value: success
-      if: 'ctx._tmp?.action.toLowerCase().contains("success")'
+      if: 'ctx._tmp?.action != null && ctx._tmp.action.toLowerCase().contains("success")'
```
bhapas:
```suggestion
      if: 'ctx._tmp?.action?.toLowerCase().contains("success")'
```

Does this not work this way?

haetamoudi: @bhapas when `action` is null, `?.toLowerCase()` returns null, but then `.contains(...)` is called on that null, which throws the error. The ?. only protects the immediate member access, not subsequent method calls in the chain.

bhapas: Sure .. yeah.. Then `ctx._tmp?.action?.toLowerCase()?.contains("success") == true` ?

Since the nesting depth is deeper?

bhapas (approved):

PR #19059, opened 2026-05-19
[Darktrace] Handle setTag not being of type Boolean.

ash-darin:
**Suggested label:** Bug _(pick the matching GitHub label)_

## Proposed commit message

Changed handling of the field 

`json.model.actions.setTag `

This field is always 'false' except in the cases when it is not, then it is *not* 'true' but a numeric value, most common occurance is "63", probably one of the tag values itself. This makes the convert processor to fail with

> [63] is not a boolean value, cannot convert to boolean

This has been confirmed by the vendor as being the expected behaviour.

This PR checks for 'false' or 'not false' and sets `darktrace.model_breach_alert.model.actions.is_tag_set` accordingly.

No other instances of similar behaviour are yet known.

## Checklist

## Author's Checklist

- [ ] Check if your "darktrace.model_breach_alert" datastream contains messages in an `error.message` field as described above. Apply patch and check result.

Create an event where `model.actions.setTag` is not 'false' but a numeric value.

## Related issues

- none found


packages/darktrace/data_stream/model_breach_alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -602,13 +602,19 @@ processors:
...
       target_field: darktrace.model_breach_alert.model.actions.is_tag_set
       type: boolean
       ignore_missing: true
       on_failure:
         - append:
             field: error.message
             value: '{{{_ingest.on_failure_message}}}'
+  - set:
+      field: darktrace.model_breach_alert.model.actions.is_tag_set
+      if: ctx.json?.model?.actions?.setTag != null && ctx.json?.model?.actions?.setTag != 'false'
```
efd6:
```suggestion
      if: ctx.json?.model?.actions?.setTag != null && ctx.json.model.actions.setTag != 'false'
```
By the time we reach the second branch, we know that it is present and not null.

packages/darktrace/data_stream/model_breach_alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -602,13 +602,19 @@ processors:
...
+      value: True
```
efd6:
```suggestion
      value: true
```

packages/darktrace/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.1.1
+  changes:
+    - description: Workaround over boolean mismatch
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/18991
```
efd6:
```suggestion
      link: https://github.com/elastic/integrations/pull/19059
```

efd6 (commented):
Do we lose information from this? Should we keep the value if it's not false?

Can you add pipeline test that exercises this change.

ash-darin:
> Do we lose information from this? 

Currently the field is not populated because of the error. So maybe..

> Should we keep the value if it's not false?

My guess is, that it is a value already seen elsewhere in the message but I have no confirmation of this from the vendor. You can see for yourself when I add an example.

> Can you add pipeline test that exercises this change.

Working on it.

packages/darktrace/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.1.1
```
andrewkroh: Missing quote is causing malformed YAML.

ash-darin: fixed

packages/darktrace/data_stream/model_breach_alert/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -602,13 +602,19 @@ processors:
             value: '{{{_ingest.on_failure_message}}}'
   - convert:
       field: json.model.actions.setTag
+      if: ctx.json?.model?.actions?.setTag == 'false'
```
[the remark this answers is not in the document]
ash-darin: testing now against concrete boolean values, using a string comparison was a mistake

ash-darin:
@andrewkroh Is someone willing to pick this up from here? I can not take care of this any more.

PR #19071, opened 2026-05-19
[auditd] Add use_filebeat_parser toggle for go-libaudit pre-parsing

Tacklebox:
## Summary

- Adds a `use_filebeat_parser` boolean toggle (default `false`) to the auditd log data stream. When enabled, Filebeat's new `auditd` filestream parser (backed by `go-libaudit`/auparse) pre-parses each `audit.log` line and populates `auditd.log.*` fields before the event reaches Elasticsearch, bypassing the grok-heavy ingest pipeline.
- Guards the four grok processors with `if: ctx.auditd?.log?.record_type == null` so pre-parsed events skip them, while the normal pipeline path is unaffected.
- Fixes a bug where the same guard was incorrectly applied to downstream KV, date, lowercase, and rename processors — those processors have `ignore_missing`/`ignore_failure` and must always run; the incorrect guards were breaking timestamp extraction and ECS field mapping on the normal path.
- Adds pipeline test fixtures for the pre-parsed path (SYSCALL+execve and USER_LOGIN), exercising that grok is skipped, auparse-resolved field names (`arch`, `syscall`) are preserved, and ECS mapping still fires correctly.

## Relationship to beats PR

The `auditd` filestream parser itself lives in `elastic/beats` (libbeat/reader/auditd). This PR wires up the integration side: the manifest toggle, the HBS template conditional, and the ingest pipeline guards. The beats PR must be merged and released before `use_filebeat_parser: true` is usable by end users.

The old grok path remains the default (`use_filebeat_parser: false`) until the new path has seen production validation, at which point the default will flip and the old path will eventually be removed.

## Test plan

- [x] `elastic-package test pipeline -C packages/auditd -d log` — all 12 tests pass (6 existing + 6 new pre-parsed tests including warnings checks)
- [ ] Manual validation with a real Elastic Agent using `use_filebeat_parser: true` once the beats parser is available


ricardoamaro (commented):
At the top of `packages/auditd/manifest.yml` I would add a `conditions.elastic.subscription` / `agent.version` floor pinned to the release that contains beats#50791. Without that, flipping the toggle on an older agent yields would go: "no such parser: auditd".

packages/auditd/data_stream/log/manifest.yml
```diff
@@ -27,6 +27,20 @@ streams:
         type: bool
         multi: false
         default: false
+      - name: use_filebeat_parser
```
ricardoamaro: note... the description claims schema equivalence, but `event.outcome` is missing for auth records on the new path (your own fixture admits this). Either fix the parser (see my comment in beats#50791 ) or rewrite this description to call out to divergence.

marc-gr:
```suggestion
      - name: use_auditd_parser
```

I would also avoid direct reference to filebeat in the description and title. Something like `When enabled, the integration assumes each log line is an auditd log, populating auditd.log.*....`

marc-gr (commented):
I would check system tests in addition or instead of the current pipeline tests to validate parser execution. Current pipeline tests are only validating pipeline bypass conditions, which is valid but we might want to be sure the config reaches filebeat correctly and it does what is intended

packages/auditd/data_stream/log/manifest.yml
andrewkroh: Why do we need a configuration option at all? Few options is better IMO.

1. I would use the [agent version templating](https://github.com/elastic/integrations/blob/main/docs/extend/agent-version-conditions.md) to inject the configuration to enable the parser when it is available (based on agent version).
2. Modify the pipeline to be an effective no-op when the data is already coming in parsed. You could shift the parsing into a sub-pipeline that is guarded by a single `if`.

andrewkroh (commented):
This need a changelog.yml entry and a manifest.yml minor version bump.

packages/auditd/data_stream/log/manifest.yml
```diff
@@ -27,6 +27,19 @@ streams:
...
+        type: bool
+        title: Use Filebeat auditd parser (experimental)
```
ricardoamaro: Are we sticking with `use_filebeat_parser` or going with the `use_auditd_parser` rename you suggested in discussion_r3281869517? `show_user: false` and no released users make the rename cheap if we want it.

packages/auditd/data_stream/log/_dev/test/system/test-default-config.yml
ricardoamaro, line 1: Per your comment in review #4337728045 : once beats#50791 ships in an Agent release we can pull, want to add a sibling `test-preparsed-config.yml` here that flips `use_filebeat_parser: true` to exercise the parser end-to-end?

ricardoamaro (approved):
approved pending changes and tests

packages/auditd/data_stream/log/manifest.yml
marc-gr: oh yeah missed my own suggestion at the end :facepalm:

packages/auditd/data_stream/log/manifest.yml
marc-gr: since this is experimental we want it to be opt in, once it is out of the experimental phase I agree that we should just stick to the semver check to enable it, currently I used both to avoid errors or having to bump the manifest version if that makes sense

packages/auditd/data_stream/log/_dev/test/system/test-preparsed-config.yml
```diff
@@ -0,0 +1,15 @@
+data_stream:
```
ricardoamaro, line 4:
```suggestion
skip:
  reason: "Requires Elastic Agent >= 9.5.0 with the go-libaudit auditd filestream parser (beats#50791). CI runs the package's minimum supported stack, which has neither the parser nor the Fleet semverSatisfies helper (needs Kibana >= 9.4). Un-skip once the test matrix can boot a >= 9.5 stack."
  link: https://github.com/elastic/beats/pull/50791
data_stream:
```

packages/auditd/_dev/deploy/docker/sample_logs/avc.log
```diff
@@ -0,0 +1,3 @@
+type=AVC msg=audit(1226874073.147:96): avc:  denied  { getattr } for  pid=2465 comm="httpd" path="/var/www/html/file1" dev=dm-0 ino=284133 scontext=unconfined_u:system_r:httpd_t:s0 tcontext=unconfined_u:object_r:samba_share_t:s0 tclass=file
```
ricardoamaro, line 1: I think that for tests to pass we need to declare all five of these in the `auditd.log` group in `packages/auditd/data_stream/log/fields/fields.yml` (for example just before the `# log_format = ENRICHED fields` block):

```yaml
    - name: audit_backlog_limit
      type: long
    - name: denied_mask
      type: keyword
    - name: peer
      type: keyword
    - name: requested_mask
      type: keyword
    - name: seresult
      type: keyword
```

PR #19074, opened 2026-05-19
[backport] Copy .github/CODEOWNERS from main to backport branch

mrodm:
## Proposed commit message

Copy `.github/CODEOWNERS` from the `main` branch into newly created backport branches.

**WHAT:** In `updateBackportBranchContents`, after restoring `.github/workflows` from `main`, also restore `.github/CODEOWNERS` from `main` and stage it for the commit.

**WHY:** The backport branch is created from a historical `BASE_COMMIT`, so its `.github/CODEOWNERS` can be outdated — missing owners or packages added after that commit. GitHub uses the CODEOWNERS file from the **base branch** of a pull request to assign reviewers, so any PR targeting the backport branch would use the stale file. Copying from `main` ensures reviewer assignment is correct and consistent. It also provides a fresh, up-to-date starting point for `removeOtherPackages`, which subsequently strips CODEOWNERS entries for removed packages.

From GitHub documentation ([link](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)):
> Each CODEOWNERS file assigns the code owners for a single branch in the repository. Thus, you can assign different code owners for different branches, such as @octo-org/codeowners-team for a code base on the default branch and @ octocat for a GitHub Pages site on the gh-pages branch.

## Author's Checklist

- [x] Verify backport branch creation in dry-run mode shows correct `.github/CODEOWNERS` diff - https://buildkite.com/elastic/integrations-backport/builds/247

Run the backport pipeline in dry-run mode (`DRY_RUN=true`) for any package. Confirm the diff output includes `.github/CODEOWNERS` and that its contents match `main`.

---

### DIFF
--- a/.buildkite/scripts/backport_branch.sh
+++ b/.buildkite/scripts/backport_branch.sh
@@ -210,6 +210,11 @@ updateBackportBranchContents() {
     git checkout "$SOURCE_BRANCH" -- ".github/workflows"
     git add .github/workflows
 
+    # This should be copied from the main branch before running "removeOtherPackages" function
+    echo "--- Copying .github/CODEOWNERS from $SOURCE_BRANCH..."
+    git checkout "$SOURCE_BRANCH" -- ".github/CODEOWNERS"
+    git add .github/CODEOWNERS
+
     # Copy tools.go so we have the dev scripts dependencies required
     echo "--- Copying tools.go from $SOURCE_BRANCH..."
     git checkout "$SOURCE_BRANCH" -- "tools.go"


jsoriano (dismissed):

.buildkite/scripts/backport_branch.sh
```diff
@@ -210,6 +210,11 @@ updateBackportBranchContents() {
     git checkout "$SOURCE_BRANCH" -- ".github/workflows"
     git add .github/workflows
 
+    # This should be copied from the main branch before running "removeOtherPackages" function
+    echo "--- Copying .github/CODEOWNERS from $SOURCE_BRANCH..."
+    git checkout "$SOURCE_BRANCH" -- ".github/CODEOWNERS"
+    git add .github/CODEOWNERS
```
jsoriano, line 216: We have to take into account that the owner in a package may have changed, though probably this should be updated too in backport branches.

teresaromero (approved):

.buildkite/scripts/backport_branch.sh
mrodm: That's true, this could cause that CI in backport branches fail because some owners do not match with the `owner` set in their manifest (that could be the previous ones), with errors like:

```
Error: error validating packages in directory 'packages': error checking manifest 'packages/elastic_package_registry': owner "elastic/infra-obs-integrations" defined in "packages/elastic_package_registry/manifest.yml" is not in ".github/CODEOWNERS"
```

Until now, all the changes performed when creating the backport branches were under CI scripts or tooling, not in any file related to packages. So, not sure we should do that.

Should we then update the package manifests too ? I think I'd prefer not to update package manifests. But if we don't update them, `.github/CODEOWNERS` file should not be updated neither :thinking: 

@jsoriano @teresaromero WDYT ? Should we keep for now the same behavior and not merge this PR?

Once the automation in backport is in-place, those codeowners could be updated also via backport Pull Requests. But the same issue remains , when new backport branches are created.

jsoriano: Well, if this fails I guess we are covered as this won't be silently ignored. 

If ownership of a package changed in main it should also change in maintained backport branches. Developers will have to update the owners in the package when trying to backport any change, as CI will fail.

mrodm: > Well, if this fails I guess we are covered as this won't be silently ignored.

This is true, this will be blocking CI. But probably for packages unrelated to the one that this backport branch was created.

> If ownership of a package changed in main it should also change in maintained backport branches. Developers will have to update the owners in the package when trying to backport any change, as CI will fail.

@jsoriano My concern here is that it could fail the ownership set for packages unrelated to the backport branch just created.

It could be created a backport branch aimed for `elastic_package_registry`, and in main it was updated the ownership of another package, for instance `nginx`. Then when it is created the backport branch just the `.github/CODEOWNERS` file is updated but not the `owner` defined in the manifest for nginx. This would cause that in the backport branch it should be fixed the nginx package, involving other teams in that backport branch.

jsoriano: _[After rewriting this comment three times]_ I am starting to think that we need a mechanism to sync owners, and this needs to run not only on branch creation but during all the lifetime of the branch.

If we only copy the owners on branch creation we may miss updates later, there won't be anything forcing a change of owners in main to be backported.

I think the moment to sync owners is when backporting changes, or in general when opening PRs against the backport branches. That is the important moment, because a new version of the package will be likely released if the PR gets merged.

This implies two things:
* When the automation to create backports runs, it looks for modified packages, and syncs its owners if needed. In CODEOWNERS and in the manifest file.
* A new CI check is added to all PRs, to check that the owners of the modified packages are synced with main. For convenience it is also included in main, but with a condition to don't run, though it would always pass as main is synced with main.

mrodm: Agreed with your proposal.

Updating Codeowners once during the creation process of the backport branches do not solve the root cause of the issue described here.

Created a new issue to handle this with the approach mentioned above https://github.com/elastic/integrations/issues/19686

mrodm:
Closing PR in favor of this new issue to handle the update of codeowners https://github.com/elastic/integrations/issues/19686

PR #19097, opened 2026-05-20
[ti_google_threat_intelligence] - Respect threat list availability delay

ShourieG:
## Type of change
- Bug
- Enhancement

## Proposed commit message
```
tti_google_threat_intelligence: respect threat list availability delay

Google Threat Intelligence hourly threat lists are only available after
a two-hour delay. The package defaulted the initial interval to 2h, but
the CEL cursor still advanced toward the current hour and could request
T-1h packages that Google rejects as NotAvailableYet.

Stop hourly threat-list streams at the documented T-2h boundary and
treat NotAvailableYet responses as dropped retry events so the same
hour can be retried later without degrading the input. Update system
tests to start at the T-2h boundary so an accidental immediate T-1h
request fails the hit count assertion.

Move the availability delay into each threat list manifest so future API
timing changes do not require CEL code changes. Clarify that the 1h
interval is safe because the CEL guard enforces the availability delay,
and document why the time value is formatted and parsed before
comparison.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
+++ b/packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
@@ -24,54 +24,89 @@ redact:
   fields:
     - access_token
 program: |
-  state.?cursor.last_timestamp.orValue(
-    (now - duration(state.initial_interval)).format("2006010215")
-  ).as(start_time, state.with(
-    request(
-      "GET",
-      state.url.trim_right("/") + "/api/v3/threat_lists/cryptominer/" + start_time + "?" + {
-        ?"query": has(state.query) ? optional.of([state.query]) : optional.none(),
-        "limit": ["4000"],
-        "x-tool": ["Elastic"],
-        "User-Agent": ["v0.8.0"], // Keep this in sync with 'version' in package level manifest.yml.
-      }.format_query()
-    ).with({
-      "Header": {
-        "x-apikey": [state.access_token]
-      }
-    }).do_request().as(resp, resp.StatusCode == 200 ?
-      resp.Body.decode_json().as(body, {
-        "events": (
-          has(body.iocs) && size(body.iocs) > 0 ?
-            body.iocs.map(e, {
-              "message": e.encode_json(),
+  state.?cursor.last_timestamp.orValue("").as(cursor,
+    (now - duration("2h")).format("2006010215").parse_time("2006010215").as(latest_available_time,
+      (
+        cursor != "" ?
+          cursor
+        :
+          (now - duration(state.initial_interval)).format("2006010215")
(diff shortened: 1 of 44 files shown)


packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
```diff
@@ -24,54 +24,89 @@ redact:
...
-        "x-apikey": [state.access_token]
-      }
-    }).do_request().as(resp, resp.StatusCode == 200 ?
-      resp.Body.decode_json().as(body, {
-        "events": (
-          has(body.iocs) && size(body.iocs) > 0 ?
-            body.iocs.map(e, {
-              "message": e.encode_json(),
+  state.?cursor.last_timestamp.orValue("").as(cursor,
+    (now - duration("2h")).format("2006010215").parse_time("2006010215").as(latest_available_time,
```
kcreddy: Do you think we should add a manifest variable with `2h` default? If Google changes this delay in future (or different threat lists have different delays), every data stream needs a code change.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
kcreddy, line 1: WDYT on this identified issue by review bot? I think there is chance for improvement due to ambiguity.

----

#### MEDIUM — `domain:manifest` — Manifest `interval` default still `1h` with misleading description

**File:** `data_stream/cryptominer/manifest.yml` (and all 13 others)

The description says:

> It must be set to a value greater than 1 hour as threat lists are generated hourly and the API returns errors when requesting data prematurely.

But the default is `1h`, which contradicts the description. With the new code, 1h is actually fine (the guard prevents premature requests), but the description creates confusion. It should either:
- Keep the "must be > 1h" guidance and change the default to `2h`, or
- Acknowledge that 1h is safe because the CEL program handles the delay internally, and remove the warning.

packages/ti_google_threat_intelligence/data_stream/cryptominer/agent/stream/cel.yml.hbs
kcreddy: Can you add a comment why a format-then-parse-back is required here: `.format("2006010215").parse_time("2006010215")` ?

ShourieG: Yea we could do that, makes sense

ShourieG: Yea sure

ShourieG:
@kcreddy, all the suggestions have been addressed.

kcreddy (approved):
LGTM

PR #19107, opened 2026-05-21
[ti_custom] - Fix TAXII initial interval requests to use millisecond precision in the `added_after` timestamp

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
ti_custom: use microsecond TAXII added_after timestamps

Format the initial TAXII added_after value with fixed microsecond
precision now that the customer confirmed the vendor accepts six
fractional digits. This keeps the generated timestamp aligned with the
TAXII 2.1 timestamp format.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/ti_custom/_dev/deploy/docker/files/config.yml
+++ b/packages/ti_custom/_dev/deploy/docker/files/config.yml
@@ -1,4 +1,36 @@
 rules:
+  - path: /taxii2/root/collections/b4f80a3f-2e68-4f3b-a3f6-1b6e4a0a7c31/objects/
+    methods: ["GET"]
+    request_headers:
+      Authorization:
+        - 'Bearer abcd1234'
+    query_params:
+      added_after: "{added_after:[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]T[0-9][0-9]:[0-9][0-9]:[0-9][0-9]\\.[0-9][0-9][0-9]Z}"
+      next: null
+    responses:
+      - status_code: 200
+        headers:
+          X-TAXII-Date-Added-Last: "2024-05-15T09:12:16.432Z"
+          Content-Type: "application/taxii+json;version=2.1"
+        body: |-
+          {
+            "more": false,
+            "objects": [
+              {
+                "id": "indicator--2ec88a3e-7c5d-4e10-9f27-ea3a4f54b3e0",
+                "spec_version": "2.1",
+                "type": "indicator",
+                "created": "2024-05-15T09:12:16.432Z",
+                "modified": "2024-05-15T09:12:16.432Z",
+                "name": "added-after.example.com",
+                "description": "Indicator used to verify added_after timestamp precision.",
+                "pattern": "[domain-name:value = 'added-after.example.com']",
+                "pattern_type": "stix",
+                "pattern_version": "2.1",
+                "valid_from": "2024-05-15T09:12:16.432Z"
+              }
+            ]
+          }
   - path: /taxii2/root/collections/e554c1d6-a37a-4267-9ff0-e8b9806b48c4/objects/
     methods: ["GET"]
     request_headers:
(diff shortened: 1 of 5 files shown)


packages/ti_custom/data_stream/indicator/agent/stream/cel.yml.hbs
```diff
@@ -84,7 +84,7 @@ program: |
                         ?"added_after": state.?cursor.last_timestamp.optMap(ts,
                             [ts]
                         ).or(
-                            state.?initial_interval.optMap(i, [(now() - duration(i)).format(time_layout.RFC3339)])
+                            state.?initial_interval.optMap(i, [(now() - duration(i)).format("2006-01-02T15:04:05.000Z07:00")])
```
efd6: The [spec](https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html) says microseconds. Has this been checked against the user's endpoint?

> `added_after`: A single timestamp that filters objects to only include those objects added after the specified timestamp. The value of this parameter is a **timestamp**.

and

> The **timestamp** type defines how timestamps are represented in TAXII and is represented in serialization as a string.
> - The timestamp type MUST be a valid RFC 3339-formatted timestamp [[RFC3339](https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html#kix.dxm6kbk8toqp)] using the format YYYY-MM-DDTHH:MM:SS.ssssssZ Unlike the STIX timestamp type, the TAXII timestamp MUST have microsecond precision.
> - The timestamp MUST be represented in the UTC timezone and MUST use the “Z” designation to indicate this.

ShourieG: The pdf Doc that the user provided, shows millisec in examples and the postman requests that the user has used all show millisec, hence why chose to go with milli. The postman requests are working examples so I'm guessing this will work.

ShourieG: After discussions, this change might cause a regression due to official [taxii spec](https://docs.oasis-open.org/cti/taxii/v2.1/os/taxii-v2.1-os.html) specifying micro second precision. Waiting on customer to confirm if microsecond works for their vendor or not.

ShourieG:
@efd6, The customer has confirmed that their servers accept `microsecond timestamps` so I've updated the PR accordingly and this now aligns with the official TAXII 2.1 spec, so we should be good.

packages/ti_custom/data_stream/indicator/agent/stream/cel.yml.hbs
ShourieG: The customer has confirmed that their servers accept microsecond timestamps so I've updated the PR accordingly.

packages/ti_custom/_dev/deploy/docker/files/config.yml
```diff
@@ -1,4 +1,36 @@
 rules:
+  - path: /taxii2/root/collections/b4f80a3f-2e68-4f3b-a3f6-1b6e4a0a7c31/objects/
+    methods: ["GET"]
+    request_headers:
+      Authorization:
+        - 'Bearer abcd1234'
+    query_params:
+      added_after: "{added_after:[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]T[0-9][0-9]:[0-9][0-9]:[0-9][0-9]\\.[0-9][0-9][0-9][0-9][0-9][0-9]Z}"
```
efd6:
```suggestion
      added_after: '{added_after:[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]{6}Z}'
```
(check whether this is compatible with gorillamux)

efd6 (approved):
nit only

PR #19110, opened 2026-05-21
ti_threatconnect: fix precision loss of indicator IDs above 2^53

### LINKED ISSUE #19109: ti_threatconnect: indicator IDs above 2^53 lose precision due to JSON float64 decoding
## Description

The ThreatConnect integration's indicator data stream loses precision on large indicator IDs returned by the ThreatConnect API. The `id` field in ThreatConnect API responses is a JSON number (not a string). When the CEL program decodes the response with `decode_json()`, these numbers pass through a float64 intermediate representation. IDs that exceed the IEEE 754 safe integer limit (2^53 = 9007199254740992) get silently rounded.

For example, three sequential ThreatConnect indicator IDs:

- `9007199254918487`
- `9007199254918488`
- `9007199254918489`

all round to `9007199254918488` under float64 precision loss (IEEE 754 round-to-nearest-even). This causes `threat_connect.indicator.id` to contain the same value for what should be distinct indicators.

The `webLink` field (a string like `https://.../#/details/indicators/9007199254918489`) preserves the correct ID because strings are not subject to numeric precision loss. This confirms the data is correct at the API level.

## Impact

- `threat_connect.indicator.id` shows incorrect, duplicate values across distinct indicators.
- Queries filtering on `threat_connect.indicator.id` return wrong results.
- Document deduplication based on `_id` is **not** affected, since the fingerprint uses `json.summary` + `json.lastModified` rather than the indicator ID.

## Root cause

The CEL program uses `bytes(resp.Body).decode_json()` which decodes JSON numbers as float64. The mito library provides `decode_json_string_numbers()` which decodes all JSON numbers as their literal string representation, preserving precision.

## Proposed fix

Replace `decode_json()` with `decode_json_string_numbers()` in the CEL program (`cel.yml.hbs`). The ingest pipeline already uses `convert` processors with explicit type coercions for all numeric fields (`confidence` to long, `rating` to double, etc.), so string-valued numbers from the CEL layer will be handled correctly.

This is the same approach used in elastic/integrations#15336 for the `abnormal_security` integration.

## Affected versions

All versions of the `ti_threatconnect` package. Only indicators with IDs at or above 2^53 (9007199254740992) are affected.

kcreddy:
## Proposed commit message
```
ti_threatconnect: fix precision loss of indicator IDs above 2^53

The ThreatConnect API returns indicator IDs as JSON numbers. When
these exceed the IEEE 754 safe integer limit (2^53), the CEL
program's decode_json() rounds them through float64, causing
distinct IDs like 9007199254918487, 9007199254918488, and
9007199254918489 to all collapse to 9007199254918488.

Replace decode_json() with decode_json_string_numbers() so that
all JSON numbers are preserved as their literal string form. The
ingest pipeline already uses convert processors with explicit type
coercions for numeric fields, so string-valued numbers are handled
correctly downstream.

Add a script test that verifies indicator IDs above 2^53 are
preserved correctly. The test mock returns two Address indicators
whose IDs both round to the same float64 value and asserts that
both distinct values appear in threat_connect.indicator.id after
ingestion.

The minimum stack version is raised to 8.19.0/9.1.0, which is
the first version shipping a mito release that includes the
decode_json_string_numbers function.

Fixes elastic/integrations#19109
```

## Checklist

Added script test (with large numbers) passes.
```
--- Test results for package: ti_threatconnect - START ---
╭──────────────────┬─────────────┬───────────┬────────────────────┬────────┬─────────────────╮
│ PACKAGE          │ DATA STREAM │ TEST TYPE │ TEST NAME          │ RESULT │    TIME ELAPSED │
├──────────────────┼─────────────┼───────────┼────────────────────┼────────┼─────────────────┤
│ ti_threatconnect │ indicator   │ script    │ env                │ PASS   │      46.96475ms │
│ ti_threatconnect │ indicator   │ script    │ large_id_precision │ PASS   │ 1m26.988814666s │
╰──────────────────┴─────────────┴───────────┴────────────────────┴────────┴─────────────────╯
--- Test results for package: ti_threatconnect - END   ---
Done
```

## Related issues

- Closes https://github.com/elastic/integrations/issues/19109

## Screenshots

**Before (`id` is rounded)**
Screenshot 2026-05-21 at 2 32 20 PM

**After (`id` is precise)**
Screenshot 2026-05-21 at 2 40 02 PM

### DIFF
--- a/packages/ti_threatconnect/_dev/deploy/docker/files/config-indicator.yml
+++ b/packages/ti_threatconnect/_dev/deploy/docker/files/config-indicator.yml
@@ -352,7 +352,7 @@ rules:
           {
             "data": [
               {
-                "id": 738669,
+                "id": 9007199254918487,
                 "dateAdded": "2023-08-24T06:28:17Z",
                 "securityLabels": {
                   "data": [
@@ -368,7 +368,7 @@ rules:
                 },
                 "ownerId": 51,
                 "ownerName": "Elastic",
-                "webLink": "https://app.threatconnect.com/#/details/indicators/738667/overview",
+                "webLink": "https://app.threatconnect.com/#/details/indicators/9007199254918487/overview",
                 "tags": {
                   "data": [
                     {
@@ -391,14 +391,16 @@ rules:
                     }
                   ]
                 },
-                "type": "URL",
+                "type": "Address",
                 "lastModified": "2023-12-05T06:47:59Z",
-                "threatAssessRating": 0,
-                "threatAssessConfidence": 0,
-                "threatAssessScore": 281,
+                "confidence": 75,
+                "rating": 4,
+                "threatAssessRating": 4,
+                "threatAssessConfidence": 75,
+                "threatAssessScore": 502,
                 "threatAssessScoreObserved": 0,
                 "threatAssessScoreFalsePositive": 0,
-                "summary": "http://www.testingmcafeesites.com/testcat_pc1.html",
+                "summary": "67.43.156.12",
                 "privateFlag": false,
                 "active": true,
                 "activeLocked": false,
@@ -528,11 +530,11 @@ rules:
                     }
                   ]
                 },
-                "text": "http://www.testingmcafeesites.com/testcat_pc.html",
-                "legacyLink": "https://app.threatconnect.com/auth/indicators/details/url.xhtml?orgid=738667&owner=Elastic"
+                "ip": "67.43.156.12",
+                "legacyLink": "https://app.threatconnect.com/auth/indicators/details/address.xhtml?address=67.43.156.12&owner=Elastic"
               },
               {
-                "id": 736759,
+                "id": 9007199254918489,
                 "dateAdded": "2023-08-24T06:19:58Z",
                 "securityLabels": {
                   "data": [
@@ -548,7 +550,7 @@ rules:
                 },
                 "ownerId": 51,
                 "ownerName": "Elastic",
-                "webLink": "https://app.threatconnect.com/#/details/indicators/736758/overview",
+                "webLink": "https://app.threatconnect.com/#/details/indicators/9007199254918489/overview",
                 "tags": {
                   "data": [
                     {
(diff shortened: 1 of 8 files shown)


packages/ti_threatconnect/data_stream/indicator/agent/stream/cel.yml.hbs
```diff
@@ -90,7 +90,7 @@ program: |
               "Timestamp": [string(int(now))],
           }
       }))).do_request().as(resp, resp.StatusCode == 200 ?
-          bytes(resp.Body).decode_json().as(body, {
+          bytes(resp.Body).decode_json_string_numbers().as(body, {
```
efd6:
```suggestion
          resp.Body.decode_json_string_numbers().as(body, {
```

packages/ti_threatconnect/_dev/deploy/docker/files/config-indicator.yml
```diff
@@ -528,11 +530,11 @@ rules:
...
                   ]
                 },
-                "text": "http://www.testingmcafeesites.com/testcat_pc.html",
-                "legacyLink": "https://app.threatconnect.com/auth/indicators/details/url.xhtml?orgid=738667&owner=Elastic"
+                "ip": "67.43.156.12",
+                "legacyLink": "https://app.threatconnect.com/auth/indicators/details/address.xhtml?address=67.43.156.12&owner=Elastic"
               },
               {
-                "id": 736759,
+                "id": 9007199254918489,
```
efd6, line 537: I think we might need a script test to ensure that this is being handled correctly; as it is, we don't know whether we're getting a numeric truncation.

kcreddy: Done in [74e813f](https://github.com/elastic/integrations/pull/19110/commits/74e813fa31511e6b36e2945cf493fa5a52026fba)

packages/ti_threatconnect/data_stream/indicator/agent/stream/cel.yml.hbs
kcreddy: Updated in [74e813f](https://github.com/elastic/integrations/pull/19110/commits/74e813fa31511e6b36e2945cf493fa5a52026fba)

efd6 (approved):
LGTM (float64 collision confirmation [here](https://play.golang.com/p/JPImAcylibl))

PR #19125, opened 2026-05-21
[ti_eset] Use cel templates for existing data streams

polakovicp:
## [ti_eset] Use cel templates for existing data streams

PR replaces httpjson templates with cel in existing data streams in ti_eset integration


efd6 (changes requested):
What is the intention here?

This changes the name of the files and changes the input, but does not actually change the implementation from an HTTP JSON input to a CEL input, or change the system tests to point to the cel input. So we would expect this to break. Also note that even if the implementation were changed, it would cause problems for users; there is more than changing the implementation to be done to avoid loss of cursor state.

polakovicp:
@efd6 I updated template files. also did some updates to system tests, which are now passing

efd6:
The problem with this is still that the cursors that exist in existing set-ups will not be used by the new input. This means that when the switch is made, the agent will behave as if it was a new install, and collect from the look-back time.

polakovicp:
@efd6 does this mean that this change should not be done for existing data streams?

efd6:
It's something that we are working towards, but it needs to be discussed.

polakovicp:
@efd6 in that case please decline this PR

efd6:
I'll raise this PR internally.

efd6:
@polakovicp We will use this as a prototype package to migrate using the [HTTP JSON `run_as_cel` option](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-httpjson#run-as-cel). If you are interested in participating, I'll help you through this. At the moment, this will leave the integration named as an `input.type: httpjson`, but the runtime will actually be the CEL input. We are working on ideas for making the full translation, but at the moment that is not possible.

Let me know here if this is something that you are interested in.

polakovicp:
@efd6 sure, what should be done on our side?

efd6:
@polakovicp Here's what needs to happen. The approach uses the [`run_as_cel` option](https://www.elastic.co/docs/reference/beats/filebeat/filebeat-input-httpjson#run-as-cel) on the httpjson input. This keeps the declared input type as `httpjson` — Fleet never sees a type change — but internally beats redirects execution to the CEL input and migrates the cursor from the old httpjson state. This preserves cursor continuity for existing users on upgrade.

**What to change in this PR:**

1. **Keep the manifest input type as `httpjson`.** Revert the `input: cel` changes in each data stream's `manifest.yml` back to `input: httpjson`, and set `template_path: httpjson_as_cel.yml.hbs`.

2. **Delete the `cel.yml.hbs` files** and rename the `httpjson.yml.hbs` templates to `httpjson_as_cel.yml.hbs`. This makes the pattern visible in the file tree — the input type is httpjson but the runtime is CEL. Update `template_path` in each manifest to match.

3. **Rewrite the `httpjson_as_cel.yml.hbs` templates** to use `run_as_cel`. The structure should be (essentially your CEL code with some modifications in an httpjson context):

~~~yaml
interval: {{interval}}
request.method: "GET"
request.url: {{url}}
{{#if ssl}}
request.ssl: {{ssl}}
{{/if}}
{{#if http_client_timeout}}
request.timeout: {{http_client_timeout}}
{{/if}}
{{#if proxy_url}}
request.proxy_url: {{proxy_url}}
{{/if}}
request.tracer:
  enabled: {{enable_request_tracer}}
  filename: "../../logs/cel/http-request-trace-*.ndjson"
  maxbackups: 5

run_as_cel: true

cel.state:
  url: {{url}}
  username: {{username}}
  password: {{password}}
  page_size: {{page_size}}
  want_more: false

cel.redact:
  fields:
    - username
    - password

cel.program: |-
  state.with({
    "added_after": state.?want_more.orValue(false) ?
      state.added_after
    :
      state.?cursor.timestamp.orValue(
        (now - duration(state.initial_interval)).format(time_layout.RFC3339)
      ),
  }).as(state, state.with(
    request(
      "GET",
      state.url.trim_right("/") + "/?" + {
        "match[type]": ["indicator"],
        "limit": [string(int(state.page_size))],
        "added_after": [state.added_after],
      }.format_query()
    ).with({
      "Header": {
        "Accept": ["application/taxii+json;version=2.1"],
        "Content-Type": ["application/taxii+json;version=2.1"],
        "Authorization": ["Basic " + (state.username + ":" + state.password).base64()],
      },
    }).do_request().as(resp, (resp.StatusCode == 200 || resp.StatusCode == 206) ?
        body.?objects.orValue([]).as(objects, {
          "events": (size(objects) > 0) ?
            objects.map(e, {"message": dyn(e.encode_json())})
          :
            [{"retry": dyn(true)}],
          "added_after": resp.Header[?"X-Taxii-Date-Added-Last"][0].orValue(state.added_after),
          "want_more": body.?more.orValue(false) && size(objects) > 0,
          "cursor": {
            "timestamp": resp.Header[?"X-Taxii-Date-Added-Last"][0].orValue(state.added_after),
          },
        })
      )
    :
      {
        "events": {
          "error": {
            "id": string(resp.Status),
            "message": "GET " + state.url.trim_right("/") + ": " + (
              (size(resp.Body) != 0) ?
                string(resp.Body)
              :
                string(resp.Status) + " (" + string(resp.StatusCode) + ")"
            ),
          },
        },
        "want_more": false,
      }
    )
  ))

tags:
  - run_as_cel
{{#if preserve_original_event}}
  - preserve_original_event
{{/if}}
{{#each tags as |tag|}}
  - {{tag}}
{{/each}}
{{#contains "forwarded" tags}}
{{/contains}}
processors:
  - drop_event.when.equals.retry: true
{{#if processors}}
{{processors}}
{{/if}}
~~~

Key points about this template:
- Auth is handled entirely in the CEL program (`cel.state` carries the credentials, the program constructs the Authorization header). The `auth.basic` config is not needed since `.do_request()` doesn't pick up config-level basic auth.
- The cursor key **must** be `timestamp` (not `last_timestamp`) to match the existing httpjson cursor field name. This is what makes cursor migration work — on first run after upgrade, the redirector injects the old httpjson cursor into `state.cursor`, and the CEL program finds the value at `state.cursor.timestamp`.
- `cel.state` provides the initial state for new installs (no cursor). For existing installs, the stored cursor overwrites `state.cursor` at runtime.
- `cel.redact` lists the state field names (`username`, `password`) to prevent them from appearing in debug logs.
- The tracer path must use `../../logs/cel/` not `../../logs/httpjson/`. After redirect, the CEL input validates the path is within its own log directory.
- `enable_request_tracer` needs `default: false` in the data stream manifest so the template always has a value to render.
- The `run_as_cel` tag is added unconditionally in the `tags:` block. Published documents will show `input.type: httpjson` which could confuse anyone investigating issues. The tag makes it easy to identify that the document was actually produced by the CEL runtime via the redirect, without altering the `input.type` field itself.
- `config_version`, `request.transforms`, `response.pagination`, `response.split`, `auth.basic`, and the old `cursor:` block are all removed — the CEL program replaces that logic.

4. **Keep system tests at `input: httpjson`.** The redirector is transparent — the input type in the policy is still httpjson.

5. **Update the Kibana version constraint** in `packages/ti_eset/manifest.yml`. The `run_as_cel` feature requires beats ≥ 9.4.0 (it is not available in 8.x), so the condition should be `"^9.4.0"`.

6. **Update sample events** — the `input.type` field in sample_event.json should remain `httpjson` (the redirector preserves it in published events).

Let me know if any of this is unclear and I'll help work through it.

efd6:
Just curious, what was the reason that you wanted to change these to CEL?

polakovicp:
@efd6 Thank you for the feedback. I've updated the PR as requested.
I wanted to clarify the current setup: existing data streams are using the `run_as_cel` option, while newly added data streams are using the CEL template. Correct?

> Just curious, what was the reason that you wanted to change these to CEL?

Not really wanted from our side as I suggested to simply decline this PR and keep existing data stream using httpjson. Originally, I did this change to maintain consistency with the new data streams

packages/ti_eset/data_stream/apt/agent/stream/httpjson_as_cel.yml.hbs
```diff
@@ -0,0 +1,103 @@
...
+  - {{tag}}
+{{/each}}
+{{#contains "forwarded" tags}}
+publisher_pipeline.disable_host: true
+{{/contains}}
+processors:
+  - drop_event.when.equals.retry: true
+{{#if processors}}
+{{processors}}
+{{/if}}
```
efd6: Add final new line. Same for the other files like this.

packages/ti_eset/manifest.yml
```diff
@@ -35,7 +35,7 @@ policy_templates:
     title: ETI feeds (TAXII version 2)
     description: Collect data from ETI feeds (TAXII version 2)
     inputs:
-      - type: httpjson
+      - type: cel
```
efd6:
```suggestion
      - type: httpjson
```
(sorry; I missed this in my instructions)

efd6:
When you make the changes you will also need to run `elastic-package build`.

packages/ti_eset/manifest.yml
polakovicp: @efd6 will this change be compatible with the new datastreams targeting [ti_set_polakovicp_enhancements](https://github.com/elastic/integrations/tree/ti_set_polakovicp_enhancements) ? For example
https://github.com/elastic/integrations/pull/19132

polakovicp:
@efd6 code updated, please see my concern here https://github.com/elastic/integrations/pull/19125#discussion_r3422737398

packages/ti_eset/manifest.yml
efd6: You will need to add the cel type to those PRs; the result should be a manifest with an httpjson input and a cel input.

efd6 (approved):
Thanks

packages/ti_eset/manifest.yml
polakovicp: @efd6, so I duplicated input in manifest.yml file and the result is quite unacceptable:
```yml
policy_templates:
  - name: eset
      ...
      - type: httpjson
        title: ETI feeds (TAXII version 2)
        description: Collect data from ETI feeds (TAXII version 2)
        vars:
          - name: username
            ...
          - name: password
            ...

      - type: cel
        title: ETI feeds (TAXII version 2)
        description: Collect data from ETI feeds (TAXII version 2)
        vars:
          - name: username
            ...
          - name: password
            ...
```
UI elements are duplicated for data streams with CEL template type. We of course do not want users to provide the same credentials multiple times. Is there any other way how to make this work?

efd6: Variables can be declared at the package level if they are common. An example of this can be seen [here](https://github.com/elastic/integrations/blob/f8ef9d776af58a221acdbcc4635359134e508b07/packages/aws/manifest.yml#L35-L97).

polakovicp: > Variables can be declared at the package level if they are common. An example of this can be seen [here](https://github.com/elastic/integrations/blob/f8ef9d776af58a221acdbcc4635359134e508b07/packages/aws/manifest.yml#L35-L97).

Yes, but I still have to define 2 entries for policy templates entries, which visually separates data streams into two groups. Can't we just make everything httpjson (old data streams and new data streams)? :D

efd6: > Can't we just make everything httpjson

Sorry. No.

PR #19139, opened 2026-05-21
[ESET PROTECT] Add device inventory and CDR vulnerability findings data streams (v2.3.0)

m11tch:
Suggested label: Enhancement (pick the matching GitHub label)

**WHAT**

Adds two new data streams to the ESET PROTECT integration at v2.3.0:

**`device`** — Collects managed device inventory from the ESET Device Management API (`/v1/devices`). Maps device identity, OS, hardware profiles, and network adapters to ECS (`host.*`). Uses an `_id` fingerprint on `device.uuid` to deduplicate inventory records across polls. Includes MAC address extraction, OS type/platform/family normalisation, and hardware profile camelCase → snake_case renaming via Painless scripts.

**`device_vulnerability`** — Collects per-vulnerability records from the ESET Vulnerability Management API (`/v1/device-vulnerabilities`). Supports all three ESET vulnerability types (application, OS, package) and maps them to ECS `vulnerability.*` and `package.*` fields. Includes a `latest` transform (`latest_cdr_vuln`) that writes to the `security_solution-eset_protect.vulnerability_latest` alias consumed by Kibana's CDR vulnerability findings panel. Each poll emits fresh documents (no `_id` fingerprint) so the transform destination stays current and the 2h retention policy correctly ages out remediated vulnerabilities.

**`detection` (updated)** — The detection stream previously only carried `host.id` (device UUID). Since the ESET detection API returns only a `deviceUuid` with no host metadata, a two-phase CEL collection cycle is introduced: Phase 1 pre-fetches the full device inventory and builds an in-memory UUID → device info map; Phase 2 fetches detections and embeds `host.name`, `host.ip`, and `host.os.*` into each event at collection time. This makes host context visible in the Kibana host flyout without requiring the entity store. The device pre-fetch is best-effort — a Phase 1 API failure falls through to Phase 2 with an empty map so detections continue to flow without enrichment.

**WHY**

ESET PROTECT customers using Kibana Security need:
1. A device inventory view for asset management (`device` stream)
2. Vulnerability findings visible in the Kibana CDR vulnerability panel (`device_vulnerability` stream + transform)
3. Host context (IP, OS) in the Kibana host flyout when investigating detection alerts (`detection` stream enrichment)

The ESET API design requires the two-phase detection approach because — unlike CrowdStrike, SentinelOne, and Microsoft Defender — the ESET detection endpoint returns only a `deviceUuid` reference with no host metadata inline. The device pre-fetch is the only way to enrich detections without a separate lookup at query time.

The vulnerability transform does not set `vulnerability.score.base` from ESET's `riskScore` field because ESET's score uses a 0–100 proprietary scale that is not comparable to CVSS (0–10). The field is declared in the transform destination mapping to satisfy the CDR panel's sort requirement, but left unpopulated. ESET's risk score is preserved in `eset_protect.device_vulnerability.risk_score`.

**Note for reviewers:** The `latest_cdr_vuln` transform destination index requires the `kibana_system` built-in role to have read access to `logs-eset_protect.device_vulnerability-*` and manage/write access to `security_solution-eset_protect.vulnerability_latest*`. A follow-up PR to the Elasticsearch repository will be needed to update the role after this integration merges.

## Checklist

1. Configure the ESET PROTECT integration in Kibana Fleet with valid credentials and region
2. Enable all three data streams: `detection`, `device`, `device_vulnerability`
3. Verify device inventory appears in `logs-eset_protect.device-*` and host metadata (IP, OS) is visible in the Kibana host flyout from detection events
4. For the CDR vulnerability panel, manually create the transform destination index template and transform as described in the note above — this step will be automatic post `kibana_system` role update
5. Navigate to Kibana Security → Findings → Vulnerabilities and confirm ESET findings appear with correct title, severity, CVE, and vendor

## Screenshots
image
image


m11tch:
I've signed the CLA in the meantime.

efd6:
Pipeline test failure:
```
test case failed: one or more problems with fields found in documents: [0] field "host.mac"'s value, 11:22:33:44:55:66, does not match the expected pattern: ^[A-F0-9]{2}(-[A-F0-9]{2}){5,}$
[1] field "host.mac"'s value, AA:BB:CC:DD:EE:01, does not match the expected pattern: ^[A-F0-9]{2}(-[A-F0-9]{2}){5,}$
[2] field "host.mac"'s value, AA:BB:CC:DD:EE:FF, does not match the expected pattern: ^[A-F0-9]{2}(-[A-F0-9]{2}){5,}$
```

m11tch:
ESET API already provides mac adres in hyphen-delimited format, it was incorrectly formatted in the sample data by accident.

efd6:
The pipeline tests are failing with
```
test case failed: Expected results are different from actual ones: --- want
+++ got
@@ -93,7 +93,7 @@
                     "203.0.113.10"
                 ],
                 "mac": [
-                    "AA-BB-CC-DD-EE-FF"
+                    "AA:BB:CC:DD:EE:FF"
                 ],
                 "name": "WORKSTATION01",
                 "os": {
@@ -204,8 +204,8 @@
                     "10.0.0.50"
                 ],
                 "mac": [
-                    "11-22-33-44-55-66",
-                    "AA-BB-CC-DD-EE-01"
+                    "11:22:33:44:55:66",
+                    "AA:BB:CC:DD:EE:01"
                 ],
                 "name": "SERVER01",
                 "os": {
```
This is because the test inputs have not been updated to use dashes in place of colons. Can you point to documentation that shows that ESET PROTECT does always use the RFC format for MAC addresses?

m11tch:
unfortunately documentation only specifies that the field is in string format. 
but data collected from multiple devices always shows it in the correct format:

image

m11tch:
all tests are passing for me now:


```--- Test results for package: eset_protect - START ---
╭──────────────┬──────────────────────┬───────────┬──────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE      │ DATA STREAM          │ TEST TYPE │ TEST NAME                                                │ RESULT │ TIME ELAPSED │
├──────────────┼──────────────────────┼───────────┼──────────────────────────────────────────────────────────┼────────┼──────────────┤
│ eset_protect │ detection            │ pipeline  │ (ingest pipeline warnings test-detection.log)            │ PASS   │ 1.023447382s │
│ eset_protect │ detection            │ pipeline  │ test-detection.log                                       │ PASS   │ 1.007469944s │
│ eset_protect │ device               │ pipeline  │ (ingest pipeline warnings test-device.log)               │ PASS   │ 1.542446716s │
│ eset_protect │ device               │ pipeline  │ test-device.log                                          │ PASS   │ 212.694357ms │
│ eset_protect │ device_task          │ pipeline  │ (ingest pipeline warnings test-device-task.log)          │ PASS   │ 865.842293ms │
│ eset_protect │ device_task          │ pipeline  │ test-device-task.log                                     │ PASS   │  181.48806ms │
│ eset_protect │ device_vulnerability │ pipeline  │ (ingest pipeline warnings test-device-vulnerability.log) │ PASS   │ 872.065815ms │
│ eset_protect │ device_vulnerability │ pipeline  │ test-device-vulnerability.log                            │ PASS   │ 155.984891ms │
│ eset_protect │ event                │ pipeline  │ (ingest pipeline warnings test-event.log)                │ PASS   │ 859.066187ms │
│ eset_protect │ event                │ pipeline  │ test-event.log                                           │ PASS   │ 1.328480924s │
╰──────────────┴──────────────────────┴───────────┴──────────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: eset_protect - END   ---


--- Test results for package: eset_protect - START ---
╭──────────────┬──────────────────────┬───────────┬──────────────────────────┬────────┬──────────────╮
│ PACKAGE      │ DATA STREAM          │ TEST TYPE │ TEST NAME                │ RESULT │ TIME ELAPSED │
├──────────────┼──────────────────────┼───────────┼──────────────────────────┼────────┼──────────────┤
│ eset_protect │ detection            │ static    │ Verify sample_event.json │ PASS   │ 361.598656ms │
│ eset_protect │ device               │ static    │ Verify sample_event.json │ PASS   │ 129.135389ms │
│ eset_protect │ device_task          │ static    │ Verify sample_event.json │ PASS   │ 296.664722ms │
│ eset_protect │ device_vulnerability │ static    │ Verify sample_event.json │ PASS   │ 182.678423ms │
│ eset_protect │ event                │ static    │ Verify sample_event.json │ PASS   │ 292.269128ms │
╰──────────────┴──────────────────────┴───────────┴──────────────────────────┴────────┴──────────────╯
--- Test results for package: eset_protect - END   ---
```

m11tch:
Sorry, still finding my way around all the different tests/checks that need to pass, I think I've covered all of them now.

packages/eset_protect/_dev/deploy/docker/files/device-config.yml
```diff
@@ -0,0 +1,19 @@
...
...{"devices":[{"uuid":"device-uuid-XXXX-5678","displayName":"SERVER01","etag":"etag-value-5678","deployedComponents":[{"displayName":"ESET Server Security","version":{"name":"11.0.5.0"}}],"lastSyncTime":"2024-01-22T09:15:00Z","originalDisplayName":"SERVER01","parentGroupUuid":"group-uuid-XXXX-5678","isMobile":false,"enrollmentStatus":"DEVICE_ENROLLMENT_STATUS_ENROLLED","functionalityStatus":"DEVICE_FUNCTIONALITY_STATUS_ATTENTION_REQUIRED","isMuted":true,"functionalityProblemCount":2,"isMaster":true,"activeProducts":["SERVER_SECURITY"],"deviceType":"DEVICE_TYPE_SERVER","primaryLocalIpAddress":"10.0.0.50","hardwareProfiles":[{"bios":{"serialNumber":"BIOS-SN-5678"},"processors":[{"caption":"Intel Xeon E5-2680","manufacturer":"Intel","id":"CPU0","architecture":"x86_64"}],"hardDrives":[{"serialNumber":"HD-SN-5678","capacityBytes":1000000000000,"hardwareEncryptionSupported":false,"manufacturerName":"Western Digital","displayName":"HDD","driveType":"DRIVE_TYPE_HDD"}],"networkAdapters":[{"macAddress":"11-22-33-44-55-66"},{"macAddress":"AA-BB-CC-DD-EE-01"}]}],"operatingSystem":{"displayName":"Ubuntu Linux 22.04 LTS","version":{"name":"22.04"}},"managementDomain":"example.com"}],"totalSize":1}
```
efd6: Prefer to have this pretty printed and use the `minify-json` helper, for example [here](https://github.com/elastic/integrations/blob/9004e054f9614e1c94f98742663101a5057de63c/packages/abnormal_security/_dev/deploy/docker/files/config.yml#L71-L87). This reduces the risk of future diff noise. (also below)

packages/eset_protect/_dev/deploy/docker/files/device_vulnerability-config.yml
```diff
@@ -0,0 +1,19 @@
+rules:
+  - path: /v1/device-vulnerabilities
+    methods: ["GET"]
+    query_params:
+      pageSize: 1
+      pageToken: xxxxxx
+    responses:
+      - status_code: 200
+        body: |
+          {"vulnerabilities":[{"deviceUuid":"device-uuid-XXXX-5678","deviceGroupUuid":"group-uuid-XXXX-0002","availablePatchUuids":["patch-uuid-XXXX-0001"],"packageVulnerability":{"package":{"displayName":"OpenSSL","name":"openssl","packageManagerType":"PACKAGE_MANAGER_TYPE_APT","versionName":"3.0.2-0ubuntu1.12"},"cveNumber":"CVE-2024-0727","firstDetectTime":"2024-01-12T09:00:00Z","lastDetectTime":"2024-01-22T14:30:00Z","patchAvailable":true,"riskScore":40,"severity":"SEVERITY_LEVEL_LOW","vulnerabilityId":11111}}],"totalSize":3}
```
efd6: Here too and below.

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
```diff
@@ -44,108 +51,236 @@ program: |
...
+                "3rd-integration": ["Elastic"],
+              }
+            })
+          :
+            request
+        )).do_request().as(resp,
+          resp.StatusCode == 200
+          ?
+            resp.Body.decode_json().as(body,
+              (has(body.nextPageToken) && body.nextPageToken != null && body.nextPageToken != "").as(has_more, {
```
efd6: What are the possible values that `nextPageToken` can take?

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
```diff
@@ -44,108 +51,236 @@ program: |
...
+                "end_time": state.end_time,
+                "page_token": state.page_token,
+                "want_more": true,
+                "initial_interval": state.initial_interval,
+                "cursor": state.?cursor.orValue({"last_timestamp": null, "response_id": null}),
+                "device_url": state.device_url,
+                "device_keys": state.?device_keys.orValue([]) + (
+                  has(body.devices) && body.devices != null
+                  ? body.devices.map(d, d.uuid)
+                  : []
```
efd6: Please [`celfmt`](https://github.com/elastic/celfmt) the code. This arrangement is non-standard.

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
```diff
@@ -44,108 +51,236 @@ program: |
...
+                "devices_fetched": true,
+                "device_keys": [],
+                "device_values": [],
+                "device_page_token": "",
+                "device_response_id": "",
+              })
+        )
+      :
+        // Phase 2: Fetch detections, embedding device display name from the lookup map.
+        zip(state.?device_keys.orValue([]), state.?device_values.orValue([])).as(device_map,
```
efd6: Is it always the case that `state.device_keys` and `state.device_values` are set together and always the same length? If not, this is unsafe.

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
```diff
@@ -44,108 +51,236 @@ program: |
...
-              ?
-                (
-                  has(state.cursor) && has(state.cursor.last_timestamp)
+              request
+          )).do_request().as(resp,
+            resp.StatusCode == 200
+            ?
+              resp.Body.decode_json().as(body, {
+                "events": (
+                  has(body.detections) && body.detections != null
```
efd6: Do we know that `detections` is ever null? or is it signalled by actual absence?

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
```diff
@@ -44,108 +51,236 @@ program: |
...
                   ?
-                    state.cursor.last_timestamp
+                    body.detections.map(e, {
+                      "message": e.with({
+                        "deviceDisplayName": (
+                          has(e.context) && has(e.context.deviceUuid) && e.context.deviceUuid in device_map
```
efd6:
```suggestion
                          has(e.?context.deviceUuid) && e.context.deviceUuid in device_map
```

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
```diff
@@ -44,108 +51,236 @@ program: |
...
+                          ? device_map[e.context.deviceUuid].displayName
+                          : ""
+                        ),
+                        "deviceInfo": (
+                          has(e.context) && has(e.context.deviceUuid) && e.context.deviceUuid in device_map
```
efd6:
```suggestion
                          has(e.?context.deviceUuid) && e.context.deviceUuid in device_map
```

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
efd6, line 1: The program here is very long. I need to look at whether this can be expressed with less complexity.

packages/eset_protect/data_stream/device/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,124 @@
...
+  user: {{escape_string username}}
+  password: {{escape_string password}}
+  token_url: https://{{region}}.business-account.iam.eset.systems/oauth/token
+  endpoint_params:
+    grant_type:
+      - password
+state:
+  page_size: {{batch_size}}
+  want_more: false
+  page_token: ""
```
efd6: Rather than setting it to a zero value, just leave it absent and check for presence in the program.

packages/eset_protect/data_stream/device/agent/stream/cel.yml.hbs
efd6, line 1: Can you format this with [`celfmt`](https://github.com/elastic/celfmt)?

packages/eset_protect/data_stream/device/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,124 @@
...
+program: |
+  (
+    request(
+      "GET",
+      state.url.trim_right("/") + "/v1/devices?" + {
+        "pageSize": [string(state.page_size)],
+        "pageToken": [state.page_token]
+      }.format_query()
+    ).as(request, (
+      has(state.cursor) && has(state.cursor.response_id) && state.cursor.response_id != null
```
efd6:
```suggestion
      has(state.?cursor.response_id) && state.cursor.response_id != null
```
but if we can just use absence rather than null, that would be better.

packages/eset_protect/data_stream/device/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,124 @@
...
+          "cursor": {}
+        })
+      :
+        (
+          resp.StatusCode == 202 ?
+            state.with({
+              "events": [{"message":"retry"}],
+              "want_more": true,
+              "cursor": {
+                "response_id": resp.Header["Response-Id"][0]
```
efd6: Is this always something or null (that is something as opposed to absent)?

packages/eset_protect/data_stream/device/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,124 @@
...
+              })
+            :
+              []
+          ),
+          "page_token": (
+            has(body.nextPageToken) && body.nextPageToken != null
+            ?
+              body.nextPageToken
+            :
+              ""
```
efd6: Same question about null versus absence. If this is absent rather than null, can we make use of that?

packages/eset_protect/data_stream/device_vulnerability/_dev/test/pipeline/test-common-config.yml
```diff
@@ -0,0 +1,4 @@
+fields:
+  tags:
+    - preserve_original_event
+    - preserve_duplicate_custom_fields
```
efd6: This is no longer recommended.

packages/eset_protect/data_stream/device_vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,219 @@
...
+state:
+  page_size: {{batch_size}}
+  want_more: false
+  page_token: ""
+  device_url: https://{{region}}.device-management.eset.systems
+  devices_fetched: false
+  device_keys: []
+  device_values: []
+  device_page_token: ""
+  device_response_id: ""
```
efd6: Same comment about setting zero values.

packages/eset_protect/data_stream/device_vulnerability/agent/stream/cel.yml.hbs
efd6, line 1: Please format this.

And similar questions as for the device template.

packages/eset_protect/data_stream/device_vulnerability/manifest.yml
```diff
@@ -0,0 +1,81 @@
...
+        multi: false
+        default: false
+      - name: preserve_duplicate_custom_fields
+        required: true
+        show_user: false
+        title: Preserve duplicate custom fields
+        description: Preserve eset_protect.device_vulnerability fields that were copied to Elastic Common Schema (ECS) fields.
+        type: bool
+        multi: false
+        default: false
```
efd6: No longer recommended.

packages/eset_protect/data_stream/device/manifest.yml
```diff
@@ -0,0 +1,81 @@
...
+        multi: false
+        default: false
+      - name: preserve_duplicate_custom_fields
+        required: true
+        show_user: false
+        title: Preserve duplicate custom fields
+        description: Preserve eset_protect.device fields that were copied to Elastic Common Schema (ECS) fields.
+        type: bool
+        multi: false
+        default: false
```
efd6: No longer recommended.

packages/eset_protect/_dev/deploy/docker/files/device-config.yml
m11tch: Done

packages/eset_protect/_dev/deploy/docker/files/device_vulnerability-config.yml
m11tch: Done, the other config files in the directory also do not follow this recommendation (`detection-config.yml`, `device_task-config.yml`, `oauth-config.yml`) but those are pre-existing and not touched by this PR, so I've left them as-is.

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
m11tch: Confirmed with the real API: `nextPageToken` is always present in the response body. It is an empty string `""` when there are no more pages, and a non-empty token string otherwise. It is never absent or `null`. Simplified all checks accordingly across all three streams.

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
m11tch: Done

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
m11tch: Yes, they are always set together and always the same length. Both are accumulated in the same code path by mapping over the same `body.devices` array in a single response, so they can never become misaligned. The 202 path preserves both via `state.with()` without touching either, and the error path resets both to `[]` together.

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
m11tch: Confirmed: `detections` is always present. The API returns `"detections": []` when there are no results in the requested time window — it is never absent or `null`. Removed the `has()` guard. Same applies to `devices` in the device pre-fetch phase.

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
m11tch: Applied

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
m11tch: Applied

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
m11tch: The complexity is largely driven by two ESET API constraints: (1) the detections endpoint only returns a `deviceUuid` with no display name, so a full device pre-fetch is required upfront to build a lookup map for enrichment; (2) both the device and detection endpoints can respond with `202 Accepted` and a `Response-Id` header, requiring async retry handling in each phase independently. Happy to simplify the structure further if you have a specific approach in mind.

packages/eset_protect/data_stream/device/agent/stream/cel.yml.hbs
m11tch: Removed `page_token: ""` from state. The program now uses `has(state.page_token) ? state.page_token : ""` in the query. Applied the same pattern to `device_vulnerability` and `detection`, and also removed `device_keys`, `device_values`, `device_page_token`, and `device_response_id` zero-value initialisations where the program already handles absence via `.orValue()`.

packages/eset_protect/data_stream/device/agent/stream/cel.yml.hbs
m11tch: Done

packages/eset_protect/data_stream/device/agent/stream/cel.yml.hbs
m11tch: Applied: `has(state.?cursor.response_id) && state.cursor.response_id != null`. Also applied the equivalent for the `detection` and `device_vulnerability` streams.

packages/eset_protect/data_stream/device_vulnerability/manifest.yml
m11tch: Removed from both `device` and `device_vulnerability` (the two new streams in this PR). The existing streams (`detection`, `event`, `device_task`) are left unchanged to avoid a breaking change for existing users.

packages/eset_protect/data_stream/device/manifest.yml
m11tch: Removed from both `device` and `device_vulnerability` (the two new streams in this PR). The existing streams (`detection`, `event`, `device_task`) are left unchanged to avoid a breaking change for existing users.

packages/eset_protect/data_stream/device_vulnerability/agent/stream/cel.yml.hbs
m11tch: Formatted with `celfmt`. For the null/absence questions: same API contract as `device` — `nextPageToken` is always present (empty string = end of pages), `vulnerabilities` and `devices` are always present (empty array when no data). Simplified accordingly.

packages/eset_protect/data_stream/device_vulnerability/_dev/test/pipeline/test-common-config.yml
m11tch: Removed from both device and device_vulnerability (the two new streams in this PR). The existing streams are left unchanged to avoid a breaking change for existing users.

packages/eset_protect/data_stream/device_vulnerability/agent/stream/cel.yml.hbs
m11tch: Done. Removed page_token, device_keys, device_values, device_page_token, and device_response_id zero-value initialisations from state. The program already handled absence via .orValue() so no other changes were needed.

packages/eset_protect/data_stream/device/agent/stream/cel.yml.hbs
m11tch: Confirmed: nextPageToken is always present in the response (empty string "" when no more pages, never absent or null), and devices is always present (empty array [] when there are no results). Simplified both checks accordingly — removed the has() guards and != null conditions.

packages/eset_protect/data_stream/device/agent/stream/cel.yml.hbs
m11tch: Yes — the Response-Id header is always present in a 202 response per the [ESET API documentation](https://help.eset.com/eset_connect/en-US/how_api_works.html). A 202 Accepted is returned when a query takes longer than 30 seconds to process; the server always includes a response-id header which the client must use to poll for the result by resending the identical request ([error codes reference](https://help.eset.com/eset_connect/en-US/error_codes.html)). Without it the async polling pattern would be broken, so accessing resp.Header["Response-Id"][0] is safe.

packages/eset_protect/_dev/deploy/docker/files/device_vulnerability-config.yml
efd6: Thanks

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
efd6: Thanks

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
```diff
@@ -25,127 +25,235 @@ state:
...
+                "events": body.detections.map(e,
+                  {
+                    "message": e.with(
+                      {
+                        "deviceDisplayName": (
+                          has(e.?context.deviceUuid) && e.context.deviceUuid in device_map
+                        ) ?
+                          device_map[e.context.deviceUuid].displayName
+                        :
+                          "",
```
efd6: We can further refine this.
```suggestion
                        "deviceDisplayName": has(e.?context.deviceUuid) ?
                          device_map[?e.context.deviceUuid].displayName.orValue("")
                        :
                          "",
```
(similar below)

packages/eset_protect/data_stream/device/agent/stream/cel.yml.hbs
efd6: Rather than `has(state.page_token) ? state.page_token : ""`, suggest `state.?page_token.orValue("")`.

packages/eset_protect/data_stream/device_vulnerability/_dev/test/pipeline/test-common-config.yml
efd6: Thanks

efd6 (commented):
I have some suggestions in the attached patch. It doesn't reduce the size of the programs, but it does separate the field zeroing from the active field updates, and I think that helps clarify the logic.

[0001-separate-state-cleanup-from-progression-logic-and-si.patch](https://github.com/user-attachments/files/28332466/0001-separate-state-cleanup-from-progression-logic-and-si.patch)

packages/eset_protect/data_stream/detection/agent/stream/cel.yml.hbs
m11tch: Applied.

m11tch:
> I have some suggestions in the attached patch. It doesn't reduce the size of the programs, but it does separate the field zeroing from the active field updates, and I think that helps clarify the logic.
> 
> [0001-separate-state-cleanup-from-progression-logic-and-si.patch](https://github.com/user-attachments/files/28332466/0001-separate-state-cleanup-from-progression-logic-and-si.patch)

Hi @efd6, the patch link returns a 404 for me, so I'm currently unable to review it. 
edit: I can access it now.

packages/eset_protect/data_stream/device_vulnerability/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,87 @@
...
+  region: xx
+  username: xxxx
+  password: xxxx
+  enable_request_tracer: true
+data_stream:
+  vars:
+    batch_size: 1
+    preserve_original_event: true
+    ssl: |
+      certificate_authorities:
```
efd6, line 13: Rather than adding multiple CAs, can we just use a common cert/key pair for all the mocks.

efd6 (commented):
minor query only, then LGTM

packages/eset_protect/data_stream/device_vulnerability/_dev/test/system/test-default-config.yml
m11tch: Done!

efd6 (commented):
Proposed commit message (needs additional information):
```
eset_protect: add device inventory and CDR vulnerability data streams

Add two new data streams:

- device: collects managed device inventory from the ESET Device
  Management API (/v1/devices) and maps to ECS host.* fields with
  fingerprint-based deduplication on device UUID.

- device_vulnerability: collects per-device vulnerability records
  from the ESET Vulnerability Management API. A latest transform
  (latest_cdr_vuln) feeds the Kibana CDR vulnerability findings
  panel. Each poll emits fresh documents so the 2h retention policy
  correctly ages out remediated entries.
```
The additional information required is the provenance of the test inputs. That is, are they sanitised events from a live endpoint, constructed from documentation, etc.

LGTM and mergeable after that.

m11tch:
> - device_vulnerability: collects per-device vulnerability records

test inputs are sanatized events from live endpoints, is this sufficient info or do I need to perform additional steps?

efd6 (approved):

efd6:
Thanks

PR #19144, opened 2026-05-21
aws: restore stable endInclusive bound to findings filter for inspector data stream

efd6:
## Proposed commit message
```
aws: restore stable endInclusive bound to findings filter for inspector data stream

The previous fix (cd8d67e) removed endInclusive entirely from the
Inspector lastObservedAt filter to prevent nextToken invalidation
caused by (now) re-evaluation during pagination. However, Inspector
may enforce the same oneOf schema constraint as Security Hub, which
rejects filters missing the upper bound.

Use a static far-future epoch value (2032-01-01T00:00:00Z) instead.
Inspector validates that dates fall within [1970-01-01, 2033-01-01),
likely due to a 32-bit epoch representation, so the value must stay
below that ceiling. The static value never changes between pagination
pages, avoiding the original nextToken invalidation.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/aws/changelog.yml
+++ b/packages/aws/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "6.19.2"
+  changes:
+    - description: Restore a stable endInclusive bound to Inspector findings filter to prevent potential nextToken invalidation.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/19144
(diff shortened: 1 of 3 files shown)


kcreddy (approved):
Fix LGTM (merge conflicts needs to be addressed)

PR #19147, opened 2026-05-22
[claroty_xdome] - Fix CEL handling for nullable API response arrays

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
packages/claroty_xdome: handle nullable CEL response arrays

Branch on non-empty Claroty response arrays before mapping them or
advancing pagination state so null responses do not degrade CEL streams.
Return no worklist when top-level alert and vulnerability responses are
empty or null, and reset parent offsets so stale pagination state does
not carry into the next collection interval.

Preserve device-page handling for parent-only events and worklist
advancement, and add nullable device-response coverage alongside the
existing empty-array coverage.

Mark synthetic device-page events as retries and drop them with a
Filebeat processor. This keeps placeholder events from reaching ingest
while still allowing CEL state to advance across empty device pages.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/claroty_xdome/data_stream/alert/agent/stream/cel.yml.hbs
+++ b/packages/claroty_xdome/data_stream/alert/agent/stream/cel.yml.hbs
@@ -87,22 +87,25 @@ program: |
             }
           ).do_request().as(resp, (resp.StatusCode == 200) ?
             resp.Body.decode_json().as(body,
-              ((has(body.alerts) && body.alerts != null) ? body.alerts : []).as(alerts,
+              (has(body.alerts) && body.alerts != null && size(body.alerts) > 0) ?
                 {
                   "worklist": {
-                    "alerts": alerts,
+                    "alerts": body.alerts,
                   },
                   "next": 0,
-                  "offset": (size(alerts) > 0) ? (int(state.offset) + int(state.batch_size)) : 0,
-                  "want_more": size(alerts) > 0,
+                  "offset": int(state.offset) + int(state.batch_size),
+                  "want_more": true,
                   "cursor": {
-                    ?"last_timestamp": (size(alerts) > 0) ?
-                      optional.of(alerts[size(alerts) - 1].updated_time)
-                    :
-                      state.?cursor.last_timestamp,
+                    "last_timestamp": body.alerts[size(body.alerts) - 1].updated_time,
                   },
                 }
-              )
+              :
+                {
+                  "want_more": false,
+                  "cursor": {
+                    ?"last_timestamp": state.?cursor.last_timestamp,
+                  },
+                }
             )
           :
             {
--- a/packages/claroty_xdome/data_stream/alert/agent/stream/cel.yml.hbs
+++ b/packages/claroty_xdome/data_stream/alert/agent/stream/cel.yml.hbs
@@ -306,19 +306,29 @@ program: |
           }
         ).do_request().as(resp, (resp.StatusCode == 200) ?
           bytes(resp.Body).decode_json().as(body,
-            ((has(body.devices) && body.devices != null) ? body.devices : []).as(devices,
+            (has(body.devices) && body.devices != null && size(body.devices) > 0) ?
               {
-                "events": (size(devices) > 0) ?
-                  devices.map(evt,
-                    {
-                      "message": evt.with(
-                        {
-                          "alert_info": state.worklist.alerts[int(state.next)],
-                        }
-                      ).encode_json(),
-                    }
-                  )
-                : (has(state.device_offset) && int(state.device_offset) == 0) ?
+                "events": body.devices.map(evt,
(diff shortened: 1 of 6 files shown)


packages/claroty_xdome/data_stream/alert/agent/stream/cel.yml.hbs
```diff
@@ -87,18 +87,22 @@ program: |
...
+                  "offset": (size(alerts) > 0) ? (int(state.offset) + int(state.batch_size)) : 0,
+                  "want_more": size(alerts) > 0,
+                  "cursor": {
+                    ?"last_timestamp": (size(alerts) > 0) ?
+                      optional.of(alerts[size(alerts) - 1].updated_time)
+                    :
+                      state.?cursor.last_timestamp,
+                  },
+                }
+              )
```
efd6: Rather that evaluating to `[]` and threading that through, I'd suggest you just return `{"want_more":false,"cursor":{?"last_timestamp": (size(alerts) > 0) ? optional.of(alerts[size(alerts) - 1].updated_time) : state.?cursor.last_timestamp}}`.

Similar in the other location like this.

ShourieG: @efd6, updated the logic now to not even create a worklist at the top level when `null` or `[]`, I think that will be more efficient and cleaner.

packages/claroty_xdome/data_stream/alert/agent/stream/cel.yml.hbs
```diff
@@ -298,40 +305,42 @@ program: |
...
         ).do_request().as(resp, (resp.StatusCode == 200) ?
           bytes(resp.Body).decode_json().as(body,
-            {
-              "events": (has(body.devices) && size(body.devices) > 0) ?
-                body.devices.orValue([]).map(evt,
-                  {
-                    "message": evt.with(
-                      {
+            ((has(body.devices) && body.devices != null) ? body.devices : []).as(devices,
+              {
```
efd6, line 310: Can you do the same kind of thing here too?

packages/claroty_xdome/data_stream/event/agent/stream/cel.yml.hbs
```diff
@@ -84,25 +84,27 @@ program: |
...
-              optional.of(body.ot_activity_events[size(body.ot_activity_events) - 1].detection_time)
-            :
-              state.?cursor.last_timestamp,
-          },
-          "start_time": state.start_time,
-          "initial_interval": state.initial_interval,
-          "batch_size": state.batch_size,
-          "api_token": state.api_token,
-        }
+        ((has(body.ot_activity_events) && body.ot_activity_events != null) ? body.ot_activity_events : []).as(events,
```
efd6: Also here.

packages/claroty_xdome/data_stream/vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -313,40 +320,42 @@ program: |
           }
         ).do_request().as(resp, (resp.StatusCode == 200) ?
           bytes(resp.Body).decode_json().as(body,
-            {
-              "events": (has(body.devices) && size(body.devices) > 0) ?
-                body.devices.orValue([]).map(evt,
-                  {
-                    "message": evt.with(
-                      {
+            ((has(body.devices) && body.devices != null) ? body.devices : []).as(devices,
```
efd6: And here.

packages/claroty_xdome/data_stream/vulnerability/agent/stream/cel.yml.hbs
```diff
@@ -313,40 +320,42 @@ program: |
...
-              "initial_interval": state.initial_interval,
-              "batch_size": state.batch_size,
-              "api_token": state.api_token,
-              "offset": state.offset,
-            }
+                      }.encode_json(),
+                    },
+                  ]
+                :
+                  [{}],
```
efd6: Is this a retry? If that's the case, mark it as such and delete it in a filebeat processor.

ShourieG: Technically not a `retry` but It is a placeholder event used to advance state when a later device page is empty. I feel we can leave this as is atm.

efd6: That's a retry. If it's added here, it should be handled here. We've seen other cases where these empty objects cause issues.

ShourieG: @efd6, made the changes

efd6 (approved):

PR #19158, opened 2026-05-22
[Cato Networks] Update changelog, version in manifest and Sync CODEOWNERS 

sharad-elastic:
## Proposed commit message
 Update changelog entry, manifest version and sync CODEOWNERS file with main branch.


ShourieG (approved):

ShourieG:
We need to keep a forward upgrade path to versioning from next time, the current approach seems to break conventions.

PR #19168, opened 2026-05-22
Add case for PAM locked account message

Tacklebox:
## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/system/data_stream/auth/elasticsearch/ingest_pipeline/message.yml
+++ b/packages/system/data_stream/auth/elasticsearch/ingest_pipeline/message.yml
@@ -35,6 +35,7 @@ processors:
       ignore_failure: true
       patterns:
         - 'for user %{QUOTE}?%{DATA:_temp.foruser}%{QUOTE}? by %{QUOTE}?%{DATA:_temp.byuser}%{QUOTE}?(?:\(uid=%{NUMBER:_temp.byuid}\))?$'
+        - 'pam_faillock\([^)]+\): %{DATA} for user %{USERNAME:_temp.foruser} account temporarily locked'
         - 'for user %{QUOTE}?%{DATA:_temp.foruser}%{QUOTE}?$'
         - 'by user %{QUOTE}?%{DATA:_temp.byuser}%{QUOTE}?$'
         - '%{BOUNDARY} user %{QUOTE}%{DATA:_temp.user}%{QUOTE}'
@@ -50,7 +51,8 @@ processors:
       ignore_failure: true
       patterns:
         - '^pam_unix(%{DATA}:%{WORD:_temp.category})'
-      if: ctx.message != null && ctx.message.contains('pam_unix(')
+        - '^pam_faillock(%{DATA}:%{WORD:_temp.category})'
+      if: ctx.message != null && (ctx.message.contains('pam_unix(') || ctx.message.contains('pam_faillock('))
   - grok:
       description: Extract rhost and user from PAM key-value messages.
       tag: grok-pam-kv
@@ -138,6 +140,11 @@ processors:
       field: event.outcome
       value: failure
       if: ctx.message != null && ctx.message.contains('authentication failure')
+  - set:
+      tag: set_outcome_failure_account_locked
+      field: event.outcome
+      value: failure
+      if: ctx.message != null && ctx.message.contains('account temporarily locked')
   - append:
       tag: append_category_authentication_2
       field: event.category
@@ -154,6 +161,11 @@ processors:
       field: event.action
       value: 'logged-off'
       if: ctx.message != null && ctx.message.contains('session closed')
+  - set:
+      tag: set_action_account_locked
+      field: event.action
+      value: account_locked
+      if: ctx.message != null && ctx.message.contains('pam_faillock(') && ctx.message.contains('account temporarily locked')
   - rename:
       tag: rename_byuser
       field: _temp.byuser


packages/system/data_stream/auth/elasticsearch/ingest_pipeline/message.yml
```diff
@@ -35,6 +35,7 @@ processors:
       ignore_failure: true
       patterns:
         - 'for user %{QUOTE}?%{DATA:_temp.foruser}%{QUOTE}? by %{QUOTE}?%{DATA:_temp.byuser}%{QUOTE}?(?:\(uid=%{NUMBER:_temp.byuid}\))?$'
+        - 'pam_faillock\([^)]+\): %{DATA} for user %{USERNAME:_temp.foruser} account temporarily locked'
```
ricardoamaro, line 38: **[Critical] `%{DATA}` in unguarded processor, root cause of the EPS regression**

This pattern is added to the `grok-pam-users` processor, which runs on **every** auth message (the only guard is `ctx.message != null && ctx.message != ""`). `%{DATA}` resolves to `.*?` and requires backtracking on every non-matching message.

The fix is a dedicated processor placed **before** `grok-pam-users`, gated on `pam_faillock(`, using `%{USERNAME}` instead of `%{DATA}`:

```yaml
- grok:
    description: Grok username from pam_faillock locked account messages.
    tag: grok-pam-faillock-users
    field: message
    ignore_missing: true
    ignore_failure: true
    if: ctx.message != null && ctx.message.contains('pam_faillock(')
    patterns:
      - 'for user %{USERNAME:_temp.pam_user} account temporarily locked'
```

This also avoids a correctness risk: the existing `for user %{DATA:_temp.foruser}$` pattern (pattern 3 in this same processor) would still fire for pam_faillock messages that don't match the new pattern, potentially writing `"test account temporarily locked"` into `user.name`. Inserting before `grok-pam-users` prevents that entirely.

packages/system/data_stream/auth/elasticsearch/ingest_pipeline/message.yml
```diff
@@ -138,6 +140,11 @@ processors:
       field: event.outcome
       value: failure
       if: ctx.message != null && ctx.message.contains('authentication failure')
+  - set:
+      tag: set_outcome_failure_account_locked
+      field: event.outcome
+      value: failure
+      if: ctx.message != null && ctx.message.contains('account temporarily locked')
```
ricardoamaro, line 147: **[Minor] Outcome condition is less specific than the action condition**

This fires on any message containing `'account temporarily locked'`, including potential non-pam_faillock sources. The action setter a few processors later (tagged `set_action_account_locked`) correctly also requires `pam_faillock(`. Make these consistent:

```yaml
if: ctx.message != null && ctx.message.contains('pam_faillock(') && ctx.message.contains('account temporarily locked')
```

packages/system/data_stream/auth/elasticsearch/ingest_pipeline/message.yml
```diff
@@ -154,6 +161,11 @@ processors:
       field: event.action
       value: 'logged-off'
       if: ctx.message != null && ctx.message.contains('session closed')
+  - set:
+      tag: set_action_account_locked
+      field: event.action
+      value: account_locked
```
ricardoamaro, line 167: **[Minor] Underscore vs hyphen. Inconsistent with existing action values**

All other action values in this pipeline use hyphens: `logged-on`, `logged-off`. This should be:

```yaml
value: account-locked
```

ricardoamaro (changes requested):
The intent is correct and the field values produced are right, but there are two blockers before this can merge.

**Blocker 1: Performance regression (inline comment on line 38)**

The new grok pattern uses `%{DATA}` inside `grok-pam-users`, which runs on every auth message without a `pam_faillock(` guard. The benchmark already shows -17.71% EPS on `auth` and -26.43% on `security`. The fix: a dedicated processor inserted _before_ `grok-pam-users`, gated on `ctx.message.contains('pam_faillock(')`, using `%{USERNAME}` instead of `%{DATA}`. See inline comment for the exact YAML.

**Blocker 2: No test cases**

`changed_files: 1` — only `message.yml` was modified. There are no new pipeline test fixtures for a pam_faillock locked-account message. Without a test, there is no CI-level verification that `user.name`, `event.outcome`, `event.action`, and `event.category` are populated correctly. Add at least one `.log` + `-expected.json` pair to `packages/system/data_stream/auth/_dev/test/pipeline/` with a message like:

```
pam_faillock(sshd:auth): Consecutive login failures for user testuser account temporarily locked
```

**Also:**
- `changelog.yml` entry is missing (checklist box is ticked but the diff does not include any changelog change).
- Two minor issues flagged inline: outcome condition specificity (line 147) and `account_locked` → `account-locked` (line 167).

packages/system/data_stream/auth/elasticsearch/ingest_pipeline/message.yml
Tacklebox: I'm a bit confused how having a `ctx.message.contains('pam_faillock(')` could improve performance in a meaningful way. `%{DATA}` is not greedy like `%{GREEDYDATA}` and it comes after a literal `pam_faillock(.)` in the pattern which already fails to match after O(n) characters if the message doesn't contain `"pam_faillock"` so the amount of backtracking required is the same in either case and not related to the presence of `"pam_faillock"` in the message. Could you maybe give an example where there is backtracking?

I think there is potential opportunity to split out the user parsing for faillock messages into a new grok block but I'm still puzzling on how it would look and how it might help.

packages/system/data_stream/auth/elasticsearch/ingest_pipeline/message.yml
Tacklebox: I think the outcome of any auth event regardless of what logged it where the account is locked should be considered a failed event.outcome? At least that was my thinking on setting the condition. I don't know if there are other messages with that string to be honest, but if it should be specific to pam_faillock I can add that.

packages/system/data_stream/auth/elasticsearch/ingest_pipeline/message.yml
Tacklebox: I don't think there's a consistent format for event.action. we also have several event actions with underscores like `ssh_login` in this pipeline, and the other pipelines seem to mostly use underscore actions `file_create_event`, `file_.*_event`, `exec_event`, and more. But there's also actions in UpperCamel, shish-kebab, etc.

ricardoamaro:
Thanks for the replies. The technical details are worth a follow-up thread if we need them, but the immediate path is simpler.

Nick's #19276 merged this on 2026-06-03: dedicated `grok-pam-faillock` processor, fixtures, and `account-locked` are already on `main` in v2.20.0.

Next step: rebase this branch onto `main` and check whether the functional diff is now redundant. If yes, close #19168 and link #19276. If rebase surfaces something #19276 missed, open a small follow-up PR with a fixture and we will review it.

Ping me if rebase hits conflicts you want a second pair of eyes on.

Tacklebox:
This was already dealt with in https://github.com/elastic/integrations/pull/19276

PR #19169, opened 2026-05-22
[Cisco IOS] Add support for severity letter in cisco logs

bhapas:
## Proposed commit message

Fix cisco_ios log parsing for minimal syslog format and letter-based severity codes.

## Summary

Some Cisco devices (notably Small Business switches) emit syslog messages in a minimal
`<PRI>%FACILITY-SEVERITY-MNEMONIC: message` format with no timestamp, hostname, or
sequence number after the PRI tag. Additionally, these devices use single-letter severity
codes (A/C/E/W/N/I/D) instead of the standard numeric 0-7 values.

Both issues caused the ingest pipeline's grok processors to fail:

1. `grok_header` — all 8 existing patterns required at least a timestamp, IP, or hostname
   after the `<PRI>` tag, so the minimal format matched none of them. Fixed by adding a
   fallback pattern `'^%{CISCO_PRIORITY_MSGCOUNT}%{GREEDYDATA:_temp_.message}$'` at the
   end of the pattern list. `CISCO_PRIORITY_MSGCOUNT` is required (not optional) so the
   fallback only fires when a PRI tag is present.

2. `grok_message` — `%{POSINT:event.severity}` only matched digits. Fixed by adding a
   second pattern that captures a single uppercase letter into a new `cisco.ios.mnemonic`
   field, preserving the original letter value (A through D) without translation. This
   avoids conflating letter-based and numeric severity semantics.

Example logs now parsed correctly:
## Checklist

## Author's Checklist

- [x] All 28 pipeline tests pass (14 existing + 14 new, including warnings tests)
- [x] No regressions in existing test cases
- [x] Letter severity mapping covers all documented Cisco IOS letter codes

```bash
cd packages/cisco_ios
elastic-package test pipeline -v --data-streams log

### DIFF
--- a/packages/cisco_ios/changelog.yml
+++ b/packages/cisco_ios/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Handle severity letter mapping for logs.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/18969
+      link: https://github.com/elastic/integrations/pull/19169
 - version: "1.35.4"
   changes:
     - description: Remove top level note from docs
--- a/packages/cisco_ios/changelog.yml
+++ b/packages/cisco_ios/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.36.0"
+  changes:
+    - description: Handle severity letter mapping for logs.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/18969
 - version: "1.35.4"
   changes:
     - description: Remove top level note from docs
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: packages/cisco_ios/changelog.yml [hunk 1]
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.36.0"
+  changes:
+    - description: Handle severity letter mapping for logs.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/18969
```
qcorporation: this should be updated to 19169

qcorporation (approved):
besides the changelog change that needs to be updated. it looks fine

PR #19163, opened 2026-05-22
[cloudflare_logpush] Ingest Pipeline Improvments

brijesh-elastic:
## Proposed commit message

```
cloudflare_logpush: Ingest Pipeline Improvments

- Update ECS to 9.3.0 and add new fields across multiple data streams
- Replace `rename` processors with typed `convert` processors for fields whose `fields.yml` mapping
  is `ip`, `long`, `double`, or `boolean`. This validates incoming values against the declared mapping
  and prevents off-type values from being silently indexed.
- Replace grok processors with dissect for simple delimiter-based pattern.
- Add `tag` key to each processor in the ingest pipelines.
- Standardize the error message format and use the latest null-removal script.
- Improve ECS mapping across multiple data streams.
- Consolidate timestamp handling to a single script across all data streams.
- Normalize severity handling in CASB data stream.
```

## Checklist

- [ ] I have reviewed [tips for building integrations](https://www.elastic.co/docs/extend/integrations/tips-for-building) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [x] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## How to test this PR locally

- Clone integrations repo.
- Install elastic package locally.
- Start elastic stack using elastic-package.
- Move to integrations/packages/cloudflare_logpush directory.
- Run the following command to run tests.

> elastic-package test -v

## Related issues

- Relates https://github.com/elastic/integrations/issues/17516

PR #19186, opened 2026-05-25
[cloudflare_logpush] fix conflicts in feature branch

brijesh-elastic:
## Proposed commit message

```
cloudflare_logpush: fix conflicts in feature branch
```


brijesh-elastic:
This PR was originally opened because the `feature` branch had conflicts with `main` (in https://github.com/elastic/integrations/pull/19163). Since `feature/*` branches are protected, we couldn't push conflict resolutions directly.

However, given the numerous CI failures and other issues encountered here, we have decided to move away from using long-lived feature branches. Because of this shift in approach, this fix is no longer needed, and I am closing the PR.

PR #19120, opened 2026-05-21
[CI][docs] Fix typos in messages and docs

mrodm:
Fix typos in messages and docs

## Related issues

- Part of https://github.com/elastic/integrations/issues/19098
- Part of https://github.com/elastic/integrations/issues/18743
- Part of https://github.com/elastic/integrations/issues/18672
- Part of https://github.com/elastic/integrations/issues/18427
- Part of https://github.com/elastic/integrations/issues/18779

PR #19190, opened 2026-05-25
Fix text auditor typo batch

Lucas-FManager:
## Summary
- Fixes the repeated `rsa.network.alias_host` punctuation/contraction typo across the listed package field definitions and generated docs.
- Fixes the remaining `cloud.account.id` and `Elasticsearch` typos in `docs/extend`.

## Root cause
The issue grouped concrete user-facing text defects: a duplicated sentence typo family plus two remaining contributor-doc typos.

## Code shape
This is a mechanical text-only change. I kept the edit scoped to the current actionable items and intentionally did not touch `dev/import-beats/variables_compact.go` because that tool was noted as deprecated in the issue thread. I also left `general-guidelines.md` alone because the issue thread says it was already fixed in #19120.

## Security assessment
Low risk: no executable code, dependencies, generated credentials, auth flow, or data parsing changed. The update only corrects static documentation and field-description text.

## Verification
- `rg -n "clear\.Also|isnt ad\.computer|cloud\.acount\.id|Elasticseach" packages docs` returns no matches.
- `rg -n "clear\. Also|isn't ad\.computer|cloud\.account\.id|Elasticsearch instance" packages docs`
- `git diff --check`

Note: `elastic-package` is not installed in my local environment, so I could not run package-level checks here.

Fixes #18743

### DIFF
--- a/docs/extend/add-mapping.md
+++ b/docs/extend/add-mapping.md
@@ -94,7 +94,7 @@ This can be done in two ways:
     Example explicit field definition:
 
     ```yaml
-    - name: cloud.acount.id
+    - name: cloud.account.id
       level: extended
       type: keyword
       ignore_above: 1024
--- a/docs/extend/developer-workflow-import-beat.md
+++ b/docs/extend/developer-workflow-import-beat.md
@@ -34,7 +34,7 @@ The import procedure heavily uses on the *import-beats* script. If you are inter
 
 3. Use the `elastic-package stack up -v -d` command to boot up required dependencies:
 
-    1. Elasticseach instance:
+    1. Elasticsearch instance:
 
         * Kibana’s dependency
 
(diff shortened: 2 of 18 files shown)


teresaromero (commented):
I've run locally `elastic-package foreach -- check` and changes look good. Did this to check the readmes where correctly generated from the tool.

Can you use the template provided for prs? I think there is no need for a changelog entry for these changes, but would @elastic/ecosystem could confirm.

qcorporation (approved):
LGTM

efd6 (commented):
The changes here need to be made in the packages' _dev/build/docs/README.md and then built into the docs/README.md. If the fixes are wanted to be published, the versions of the packages need to be bumped (patch level bump since these are bug fixes).

PR #19193, opened 2026-05-25
[Enhancement] Windows: Add scheduled task parsing support

w0rk3r:
## Proposed commit message

```
windows, system: normalize Security scheduled task definitions

Decode scheduled task XML from Security lifecycle events and normalize selected
task configuration into winlog.scheduled_task.*. Add correlation-safe action
and trigger definitions, tests, docs, and a pipeline benchmark.
```

## Summary

Related issue: https://github.com/elastic/ia-trade-team/issues/892

Windows Security scheduled task events store most task configuration as raw XML in `winlog.event_data.TaskContent` / `TaskContentNew`, so detections and tuning lean too heavily on attacker-controlled `TaskName`.

This PR decodes that XML on Elastic Agent and normalizes selected configuration into `winlog.scheduled_task.*` for Security scheduled task events in the System `security` and Windows `forwarded` data streams. It also adds correlation-safe action/trigger `definition` fields so multi-attribute queries can bind to the same configured item.

## Why

Windows scheduled tasks are a common mechanism for persistence, remote execution, privilege abuse, and ransomware deployment. Security events already capture useful scheduled task metadata, but most of the task definition is stored as raw XML in `winlog.event_data.TaskContent` and `winlog.event_data.TaskContentNew`, with `TaskName` being the only readily queryable field today.

That XML contains the behavioral details needed to understand the task: what it runs, when it runs, which account it uses, and whether it is configured in suspicious ways such as hidden execution or elevated run level. Today, those details are not available as structured fields for search, visualization, tuning, or reliable detection logic.

As a result, some detections have to rely on `TaskName`, which is attacker-controlled and easy to make look benign. This weakens detection quality and can make detections bypassable.

This PR extracts scheduled task metadata into structured `winlog.scheduled_task.*` fields. That makes scheduled task activity easier to investigate and enables more durable detections based on what the task actually does, not just what it is called.

## What changed

Normalization covers Security events `4698`, `4699`, `4700`, `4701`, and `4702`.

- Agent `decode_xml` processors decode `TaskContent` / `TaskContentNew` into temporary maps.
- The Elasticsearch ingest pipeline normalizes selected values into `winlog.scheduled_task.*`. The ingest script does not parse raw XML.

Normalized fields include:

- task name and registration URI
- principals (run-as user/group, logon type, run level)
- Exec and COM handler actions (command, arguments, working directory, class ID)
- triggers (type, effective enabled state, repetition settings)
- selected settings (`hidden`, `enabled`)

These fields describe configured task definition, not evidence that an action ran or a trigger fired. `event.action` still identifies the audited lifecycle operation.

If XML is missing or cannot be decoded, the pipeline still sets `winlog.scheduled_task.name` from `TaskName`. Delete events work with or without task XML.

## Action and trigger correlation

Elasticsearch flattens arrays of objects, so separate predicates on component fields can match different actions or triggers in the same event.

Component fields remain available for single-attribute searches. This PR also emits one canonical `wildcard` definition per action and trigger:

```
v=1|type=exec|context=Author|command=powershell.exe|arguments=-EncodedCommand ...|working_directory=|class_id=
```

```
v=1|type=time|enabled=false|interval=PT30M|duration=|stop_at_duration_end=false
```

Definitions use a fixed key order. An empty value means the source element was omitted or empty after normalization. `%`, `|`, `=`, C0 controls, and DEL are escaped so values cannot inject extra components.

Put every same-item condition in one predicate, for example:

```
winlog.scheduled_task.actions.definition: *type=*exec|context=*|command=*powershell.exe|arguments=*EncodedCommand*
```

That supports same-action / same-trigger searches in KQL, Lucene, EQL, ES|QL, and Query DSL without nested mappings. Two separate predicates against the multivalued `definition` field are not correlation-safe.

## Defaults and missing values

An omitted trigger `Enabled` element defaults to `true` in Task Scheduler, so both the structured field and the definition expose `true`. For example, this trigger has no `Enabled` element:

```xml
<BootTrigger />
```

Other optional values stay absent from structured fields and empty in the definition when omission is meaningful. An empty `stop_at_duration_end=` means the XML element was omitted (Task Scheduler defaults that to `false` when repetition applies); that remains distinct from explicit `stop_at_duration_end=false`.

Omitted:

```xml
<Repetition>
  <Interval>PT30M</Interval>
</Repetition>
```

Explicit:

```xml
<Repetition>
  <Interval>PT30M</Interval>
  <StopAtDurationEnd>false</StopAtDurationEnd>
</Repetition>
```

## Validation

Validated with:

- System and Windows package build + lint
- System pipeline suite (312 checks) and Windows pipeline suite (202 checks)
- matching scheduled-task pipelines across both packages
- create / update / enable / disable / delete fixtures
- multiple actions and triggers; Exec and COM handlers
- explicit and omitted booleans; malformed XML and `TaskName` fallback
- delimiter, control-character, Unicode, and wildcard escaping
- long definitions without `_ignored` fields

Indexed-query checks matched real action/trigger combinations and returned zero hits for impossible combinations across Query DSL, KQL, Lucene, EQL, and ES|QL. An intentionally unsafe two-predicate control still matched, confirming the documented boundary.

End-to-end: Elastic Agent on Windows Server 2019 to a cloud stack. A controlled task with two Exec actions and three triggers produced two action definitions and three trigger definitions on create and update.

The Windows `forwarded` scheduled-task pipeline is byte-identical to System and passed its full local suite.

## Checklist

System:

```bash
cd packages/system
elastic-package build
elastic-package lint
elastic-package test policy --data-streams security
elastic-package test pipeline --data-streams security
```

Windows:

```bash
cd packages/windows
elastic-package build
elastic-package lint
elastic-package test policy --data-streams forwarded
elastic-package test pipeline --data-streams forwarded
```

Benchmark (ingest normalization only: fixtures supply decoded XML in `_tmp`, so Agent `decode_xml` is not included; every document takes the scheduled-task path):

```bash
cd packages/system
elastic-package stack up -d --services=elasticsearch
elastic-package benchmark pipeline --data-streams security --use-test-samples=false
elastic-package stack down
```


packages/system/data_stream/security/elasticsearch/ingest_pipeline/scheduled_task.yml
```diff
@@ -0,0 +1,470 @@
+---
+description: Pipeline for parsing Windows Security scheduled task XML.
+processors:
+  - script:
```
leehinman, line 7: Instead of parsing this in the ingest node with a custom painless script, would it be possible to use the [decode xml](https://www.elastic.co/docs/reference/beats/filebeat/decode-xml) processor on the agent?  Renames and cleanup could still be done at the ingest node, but it should greatly reduce the size of this script and mitigate the worse cast performance impact.

marc-gr: we should test it as the xml processor in filebeat has some conventions that might need attention IIRC. I tried to make https://github.com/elastic/elasticsearch/pull/130337 but it is stalled due to the performance of xpath evaluation. I might try to revisit that and see if we can make it happen on the pipelines and if not I'll see what needs to be done to use the filebeat processor instead.

w0rk3r: For now, what action should I take? As the XML processor is also something "more experimental" and the script is tested and restricted to a smaller subset of event codes, should we just continue with this approach? @marc-gr @leehinman

marc-gr: I'd rather try using filebeat xml processor and have the pipeline make any needed transformations as @leehinman suggested. We can later on switch to the xml decode in the pipeline if we end up merging the ingest processor. The performance hit of such a script is hard to tell and it looks like it would be also a bit harder to maintain than if we do it in filebeat.

w0rk3r: Updated the PR to use decode_xml. Should be ready for another round of reviews.

leehinman (approved):
LGTM

marc-gr (approved):
Nice work on this, approving it to unblock you but just wanted to mention two small things:

1. Raw `|` characters in the `definition` field descriptions might break the generated README tables. Worth confirming they don't break it, and if they do maybe wording as "pipeline separated key=value" or similar as an alternative.

2. Unrecognized trigger types are silently dropped, maybe we could consider a catch-all fallback that at least emits the raw element name as `type`, or documenting the omission in the `triggers` group description the way the actions docs already do.

w0rk3r:
Hmmmm, it seems like a git bug, I didn't do these changes and the revert didn't remove them?! Taking a look

w0rk3r:
@marc-gr I removed the raw `|` characters so they don’t break the README tables and added a note to the triggers group explaining that unrecognized trigger elements are omitted from normalization. Since the allowlist covers the current Microsoft-defined trigger types, I didn’t add a fallback.

ishleenk17:
Giving code owner approavl

ishleenk17 (approved):
Code owner approval

PR #19203, opened 2026-05-26
[wiz] Various minor improvements

chrisberkhout:
## Proposed commit message

```
[wiz] Various minor improvements

- Fix the logo file.

  It was raster data with formatting that wasn't strictly correct. Now
  it's vector data.

- Set the right User-Agent header.

  A workaround for https://github.com/elastic/beats/issues/50867.

- Don't use the demo env URL as a default value.

- Make `cloud_configuration_finding_full_posture`'s 24h interval an
  advanced options.

  That may be helpful for debugging or workarounds.

- Use `updatedAt` rather than `analyzedAt` in queries for cloud
  configuration finding data.

  The `analyzedAt` field remains and is mapped in the same way, but the
  `updatedAt` field is added and that's what's used for query filtering,
  as recommended by Wiz

- Tolerate URLs with a `/graphql` or `/graphql/` path.

  We expect just the base: `https://api.somedc.app.wiz.io`. Wiz provides
  an "API Endpoint" with the path:
  `https://api.somedc.app.wiz.io/graphql`, so that should work too.

- Set a static request rate of 0.5 rps per data stream, increase
  retries.

  Testing shows that 5 rps should be okay (4 rps is fine for 10s, 8 rps
  occasionally returns 429). We have 5 data streams using the GraphQL
  API.  Four run frequently (by default every 5 mins), one is a daily
  sync.
    
  Setting a request rate limit of 0.5 rps for each data stream means we
  should never exceed 2.5 rps in total, and never reach the rate limit.
  
  The only feedback from the server about the rate limit is that it
  responds with 429, with retry-after / reset information, once the
  limit has been exceeded. If a 429 response is received, the input will
  retry the request, respecting the retry-after period.
  
  We set a conservative static rate because the different data streams
  are unable to coordinate on the client side, and they receive no
  request budget feedback from the server side until its too late.
  
  We don't expect retries, but since they are possible due to rate
  limiting rather than just request or server errors, the maximum number
  of retry attempts is raised from 5 to 10.

- Make Token URL a non-advanced setting.

  Since apparently it can vary, and will be provided by Wiz along with
  the API Endpoint and credentials.

- Tell the user the interval setting 'must be 5m or longer'.

- Add a screenshot of the policy editor UI.

  This was specifically requested by Wiz.

- README: Clarify information about transforms and their outputs.

- Dashboards: Use panel-level filters, fix the nav/overview panel
  formatting, and some other minor tidying.
```

## Note to reviewers

This is best read commit-by-commit.

It's batch of updates based on discussion with Wiz and a general review of the current state.

I'm open to breaking it into multiple PRs, if you think that's worthwhile.

## Checklist


packages/wiz/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -59,9 +61,9 @@ max_executions: {{max_executions}}
 {{/if}}
 program: |
   state.with(
-    post_request(
-      state.url.trim_right("/") + "/graphql",
-      "application/json",
+    request(
+      "POST",
+      state.url.trim_right("graphql").trim_right("/") + "/graphql",
```
kcreddy: 🔴 **CRITICAL** `data_stream/audit/.../cel.yml.hbs:66`

**Incorrect trim_right() usage in audit CEL template**

The audit stream uses `state.url.trim_right("graphql").trim_right("/") + "/graphql"` which is incorrect. The CEL `trim_right()` function removes characters from a character set, not a suffix string. This means `trim_right("graphql")` removes any occurrence of the characters 'g', 'r', 'a', 'p', 'h', 'q', or 'l' from the right side of the URL, which can corrupt URLs that contain these letters. If the URL ends with any of these characters (common in domain names like `.app`), they will be incorrectly stripped.

**Recommendation:**

Remove the erroneous `trim_right("graphql")` call:

```cel
request(
  "POST",
  state.url.trim_right("/") + "/graphql",
  {
    "query": state.query,
```

---
🤖 _AI-Generated Review_ | _Vera Review Bot_
> ⚠️ Automated review — verify suggestions before applying.

packages/wiz/data_stream/cloud_configuration_finding/manifest.yml
```diff
@@ -18,7 +18,7 @@ streams:
       - name: interval
         type: text
         title: Interval
-        description: Duration between requests to the Wiz API. Supported units for this parameter are h/m/s.
+        description: Duration between requests to the Wiz API. Must be 5m or longer. Supported units for this parameter are h/m/s.
```
kcreddy, line 21: 🔵 **LOW** `data_stream/cloud_configuration_finding/manifest.yml:1`

**Typo in cloud_configuration_finding manifest title**

The title contains a typo: "Collet Cloud Configuration Finding logs from Wiz." should be "Collect Cloud Configuration Finding logs from Wiz."

**Recommendation:**

Fix the typo:

```yaml
title: Collect Cloud Configuration Finding logs from Wiz.
```

---
🤖 _AI-Generated Review_ | _Vera Review Bot_
> ⚠️ Automated review — verify suggestions before applying.

packages/wiz/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "4.3.0"
+  changes:
+    - description: Various minor improvements.
```
mohitjha-elastic: Could we make the changelog entry more descriptive?
This PR introduces several user-visible changes. Since consumers rely on changelog entries to understand upgrade impact and behavior changes, it may help to capture these updates more explicitly.

packages/wiz/data_stream/cloud_configuration_finding/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -115,6 +115,17 @@ processors:
       tag: set_event_created
       copy_from: wiz.cloud_configuration_finding.analyzed_at
       ignore_empty_value: true
+  - date:
```
mohitjha-elastic, line 118: Incremental collection now filters and cursors on `updatedAt`, but `event.created` is still derived from `wiz.cloud_configuration_finding.analyzed_at.` 
If the ingestion of new log entries is driven by `updatedAt`, should `event.created` also be aligned with the same timestamp to maintain consistency?

mohitjha-elastic (changes requested):

packages/wiz/data_stream/audit/agent/stream/cel.yml.hbs
chrisberkhout: Good catch! Rewrote the commit to fix it, using `state.url.trim_right("/").trim_suffix("/graphql")`.

packages/wiz/data_stream/cloud_configuration_finding/manifest.yml
chrisberkhout: Not in this change, but also a good catch. Fixed in a new commit.

packages/wiz/changelog.yml
chrisberkhout: Sure. I've split the entry into multiple changes that should lay everything out in a way that's readable for users.

packages/wiz/data_stream/cloud_configuration_finding/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: I think analysis creates the finding, but there may be other changes to it later.
I think `analyzedAt` is still the best value for `event.created`.
Wiz specifically asked that we use `updatedAt` for pagination, so I changed that and also keep the extra value in `wiz.cloud_configuration_finding.updated_at`.

kcreddy (approved):
LGTM, thanks!

packages/wiz/data_stream/cloud_configuration_finding/elasticsearch/ingest_pipeline/default.yml
mohitjha-elastic: Thanks for the clarification. 

In this case, if `analyzedAt` represents the original finding creation time, it could likely be the better approximation.

mohitjha-elastic (approved):
Thanks!

PR #19217, opened 2026-05-26
Add requires-update workflow - manually triggered

### LINKED ISSUE #17404: Automatically update package dependencies

teresaromero:
## Proposed commit message

Add a manually-triggered workflow that runs `mage RequiresUpdate` across all integration packages
and opens one PR (or, when every proposal is Kibana-version-blocked, one GitHub issue) per package
with a `requires:` block, bumping `requires.input`/`requires.content` pins to the latest registry
versions.

**What changed:**

`mage RequiresUpdate` walks `packages/`, and for each `type: integration` package that declares a
`requires:` block, calls `elastic-package requires update --changelog --format json`. That command
resolves version bumps from the package registry, applies them to `manifest.yml`, adds a
`changelog.yml` entry per bump, and bumps the package version by the largest semver tier across all
applied bumps (major > minor > patch) — all delegated to `elastic-package`, nothing reimplemented
here. The Go code parses the JSON result, resolves the owning team(s) via `.github/CODEOWNERS`
(falling back to the manifest's `owner.github`, and finally to `elastic/ecosystem` if neither
resolves), and publishes:

- **One PR per package** with applied bumps — branch `automated/requires-update-<package>`, stable
  and force-pushed so re-running updates the existing PR in place rather than duplicating it.
- **One GitHub issue per package** whose proposals were entirely skipped (Kibana version constraint
  too old), so the owning team has something actionable without a PR that can't merge.

Set `DRY_RUN=true` to see proposals without writing files (also skips publishing). Set
`PREVIEW=true` to print what would be published without touching git or GitHub.

**Key decisions:**

- **One PR per package, not per team, for now.** `requires:` adoption is just starting — at most one
  package has it today. Per-package keeps each PR's diff and review scope obvious. Once enough teams
  are getting a steady trickle of these PRs, batching by codeowner team (one PR covering all of a
  team's packages) is a natural follow-up if that team would rather review one PR than several small
  ones — the JSON contract and owner resolution already group by team, so that change is additive.
- **Delegate version bump and changelog to `elastic-package --changelog`**, not reimplemented in this
  repo — one less place for the semver-tier logic to drift from `elastic-package`'s own.
- **CODEOWNERS resolution over the manifest fallback**, with a logged mismatch, because a package can
  have multiple owning teams in CODEOWNERS but only one `owner.github` in its manifest — losing
  co-owners would mean they never see the update.
- **Manual-only trigger (`workflow_dispatch`), no schedule yet.** This is intentionally the smallest
  end-to-end slice: prove the scan → apply → PR/issue → changelog-link-fixup path works against a
  real package before running it unattended on a cadence. Enabling the weekly schedule is a
  deliberate follow-up once a manual run has been verified against a real `requires:` package.

## Checklist

- [ ] I have verified that all data streams collect metrics or logs. _(N/A — dev tooling, no package data streams changed)_
- [ ] I have added an entry to my package's `changelog.yml` file. _(N/A — dev tooling, not a package change)_
- [ ] I have verified that Kibana version constraints are current. _(N/A)_
- [ ] I have verified that any added dashboard complies with Kibana's Dashboard good practices. _(N/A)_

```bash
# Preview proposals without writing files:
DRY_RUN=true mage RequiresUpdate

# Apply changes locally but skip git/GitHub (prints what would be published):
PREVIEW=true mage RequiresUpdate

# Full run (writes files, opens PRs/issues) — requires GITHUB_TOKEN:
mage RequiresUpdate
```

Output from local test merging into a custom branch this and the elastic-package-registry update;

```bash
➜ integrations git:(test-local) DRY_RUN=true mage RequiresUpdate


=== Requires Update Summary (proposed (dry run)) ===

## elastic/ecosystem
  elastic_package_registry
    [input] prometheus_input: 1.0.0 → 1.0.1

Scanned 469 packages, 1 eligible (requires: present), 1 with proposals, 0 errored.
DRY_RUN=true — skipping PR/issue creation.
```
```
➜  integrations git:(test-local) ✗ PREVIEW=true mage RequiresUpdate

=== Requires Update Summary (applied) ===

## elastic/ecosystem
  elastic_package_registry
    [input] prometheus_input: 1.0.0 → 1.0.1

Scanned 469 packages, 1 eligible (requires: present), 1 with proposals, 0 errored.
======================================== PR
Package: elastic_package_registry
Owners:  @elastic/ecosystem
Branch:  automated/requires-update-elastic_package_registry
Files:
  packages/elastic_package_registry/manifest.yml
  packages/elastic_package_registry/changelog.yml

PR title: [automation] Update required package versions for `elastic_package_registry`

PR body:
## Applied

- **prometheus_input** (`input`): `1.0.0` → `1.0.1`
========================================
```

Once merged, trigger the workflow manually via `workflow_dispatch` from the Actions tab.

## Related issues

- Resolves https://github.com/elastic/integrations/issues/17404


teresaromero:
drafting this until i do a full review now the command is merged into elastic-package

teresaromero:
@mrodm opening this for review. Since the first proposal, I've changed two things to deliver this first approach of the update:

PRs/issues are created per package — the grouping by codeowner was adding extra complexity. Since this is being adopted by teams, we can evaluate grouping PRs by codeowner later.
Version bump behavior for content packages — these packages allow version constraints, but the update is currently done with pinned versions. I think this is something we should adjust and revisit on the elastic-package side. Should the update of content packages happen whenever the newest version falls outside the constraint? If so, should the required version then become a new version constraint?

This PR as-is will need an immediate follow-up to enable the automatic workflow. So far I've only tested it locally using the mage commands with the dry-run and preview flags. I'd like to test it against a real package (https://github.com/elastic/integrations/pull/19719) by triggering the workflow manually. My plan is to merge that package PR with a lower version first, then use the merged PR to open the update PR. Do you think there will be an issue other than have a new version published with a lower prometheus input?

WDYT? Thanks!

.github/workflows/requires-update.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Install elastic-package
+        # Installs the version pinned in go.mod, rather than floating to the
+        # latest release, since this job runs unattended and depends on exact
+        # elastic-package JSON output fields (kind, codeowner,
+        # kibana_constraint, new_version, skip_reason). Dependabot already
+        # bumps this pin deliberately via PR (gomod ecosystem).
+        run: |
+          version=$(go list -m -f '{{.Version}}' github.com/elastic/elastic-package)
+          go install "github.com/elastic/elastic-package@${version}"
+          echo "ELASTIC_PACKAGE_BIN=$(go env GOPATH)/bin/elastic-package" >> $GITHUB_ENV
```
mrodm: As this step runs in the working copy of the integrations , I think it could be just run 

```sh
go install "github.com/elastic/elastic-package"
echo "ELASTIC_PACKAGE_BIN=$(go env GOPATH)/bin/elastic-package" >> $GITHUB_ENV
```

and it will install the version from the `go.mod` file.

dev/requiresupdate/requiresupdate.go
```diff
@@ -0,0 +1,281 @@
...
+// adds changelog entries for packages with actual updates, opens one PR (or,
+// if every proposal was skipped, one issue) per package via Publish, and
+// prints a summary grouped by codeowner.
+//
+// Set DRY_RUN=true to see proposals without applying (this also skips
+// publishing, since no files were written). Set PREVIEW=true to print what
+// Publish would do without touching git or GitHub.
+func Run() error {
+	dryRun := os.Getenv("DRY_RUN") == "true"
+	preview := os.Getenv("PREVIEW") == "true"
```
mrodm: This could be set as mage parameters and then these functions would receive these values via parameters too.

Mage now it also allows optional parameters/flags https://magefile.org/targets/


```
func RequiresUpdate(dryRun, preview *bool) error {
	return requiresupdate.Run()
}
```

dev/requiresupdate/requiresupdate.go
```diff
@@ -0,0 +1,281 @@
...
+const packagesDir = "packages"
+
+// elasticPackageBin returns the elastic-package binary to use.
+// Override with ELASTIC_PACKAGE_BIN for local builds that carry new commands.
+func elasticPackageBin() string {
+	if bin := os.Getenv("ELASTIC_PACKAGE_BIN"); bin != "" {
+		return bin
+	}
+	return "elastic-package"
+}
```
mrodm, line 31: Could it be assumed that `elastic-package` would be present in the PATH variable ?

dev/requiresupdate/publish.go
mrodm, line 1: This new folder must be copied now in the backport_branch.sh when new backport branches are created , since these functions are used in the `magefile.go`

https://github.com/elastic/integrations/blob/7601d01dccadc36768f43b5aca329832cbbcf805/.buildkite/scripts/backport_branch.sh#L198-L200

Wondering if it would be better to just copy the whole `dev/` folder instead of adding new folders each time.

dev/requiresupdate/publish.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	return b.String()
+}
+
+func gitExec(args ...string) error {
+	cmd := exec.Command("git", args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
```
mrodm: There is a `dev/backports/gitutil` that looks it could be reused here for these functions. As it could be used in other packages probably it could be moved one level-up.

If doing so, that folder should also be included in the `backport_branch.sh` script :/ 

It would be worth if there is any folder that it is not copied. If all of them are copied in the backport process, I guess it could be just copied the whole `dev/` folder into each new backport branch.

.github/workflows/requires-update.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Configure git
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Fetch remote branch states
+        run: git fetch origin
```
mrodm: Why is this needed ?
IIUC this workflow would need just the current branch. What other branches does it require ?

dev/requiresupdate/publish.go
```diff
@@ -0,0 +1,265 @@
...
+
+// Publish opens one PR per package with applied changes — creating or
+// updating a stable per-package branch — and one GitHub issue per package
+// whose proposals were entirely skipped. Deliberately kept simple while
+// adoption of `requires:` is low; revisit batching by codeowner once weekly
+// volume per team justifies the added complexity.
+//
+// In preview mode nothing is written to git or GitHub; actions are only
+// printed.
+func Publish(summaries []packageSummary, preview bool) error {
```
mrodm: It seems that this method is not used outside of `requiresupdate` package. I think it can be set as private.

dev/requiresupdate/publish.go
```diff
@@ -0,0 +1,265 @@
...
+			"========================================\n\n",
+			s.name, mentions(s.codeowners), branch, strings.Join(s.files, "\n  "), title, body)
+		return nil
+	}
+
+	// Reset HEAD to main without discarding the dirty working tree.
+	// "checkout -B" moves HEAD but does not touch untracked/modified files,
+	// so other packages' pending changes survive subsequent calls.
+	if err := gitExec("checkout", "-B", branch, "origin/main"); err != nil {
+		return fmt.Errorf("creating branch: %w", err)
```
mrodm: Nit.
Should this error (and the others) add `failed` in the message?




```suggestion
		return fmt.Errorf("creating branch failed: %w", err)
```

dev/requiresupdate/publish.go
```diff
@@ -0,0 +1,265 @@
...
+		"pr", "create",
+		"--base", "main",
+		"--head", branch,
+		"--title", title,
+		"--label", "automation",
+		"--body", body,
+	}
+	for _, owner := range codeowners {
+		args = append(args, "--reviewer", "@"+owner)
+	}
```
mrodm: I think the required owners will be added as a reviewers automatically depending on the updated files and the `.github/CODEOWNERS` file.

Is this automatic process adding more teams as a reviewers than the ones set in that file ? Do you want to add all the teams defined as owners for one package along with the teams defined as owners in the data streams related to that package ?

dev/requiresupdate/publish.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+		data, err := os.ReadFile(f)
+		if err != nil {
+			return fmt.Errorf("reading %s: %w", f, err)
+		}
+		fixed := strings.ReplaceAll(string(data), "pull/REPLACE_ME", "pull/"+prNumber)
+		if fixed == string(data) {
+			continue
+		}
+		if err := os.WriteFile(f, []byte(fixed), 0644); err != nil {
```
mrodm: Use `os.Stat` to get the current `info.Mode()` and use the same file permissions instead of hardcoding `0644`.

.github/workflows/requires-update.yml
```diff
@@ -0,0 +1,69 @@
...
+
+jobs:
+  requires-update:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
+    steps:
+      - uses: actions/checkout@v6
```
mrodm: Other workflows are using `v7`


```suggestion
      - uses: actions/checkout@v7
```

dev/requiresupdate/publish.go
```diff
@@ -0,0 +1,265 @@
...
+	for _, s := range sorted {
+		if len(s.files) == 0 {
+			if err := publishIssue(s, preview); err != nil {
+				return fmt.Errorf("%s: %w", s.name, err)
+			}
+			continue
+		}
+		if err := publishPR(s, preview); err != nil {
+			return fmt.Errorf("%s: %w", s.name, err)
+		}
```
mrodm, line 59: Each iteration of this loop is just scoped to just one package, is that right?

dev/requiresupdate/publish.go
```diff
@@ -0,0 +1,265 @@
...
+		if fixed == string(data) {
+			continue
+		}
+		if err := os.WriteFile(f, []byte(fixed), 0644); err != nil {
+			return fmt.Errorf("writing %s: %w", f, err)
+		}
+		if err := gitExec("add", "--", f); err != nil {
+			return err
+		}
+		changed = true
```
mrodm: If this method is just in the scope of one package, when `changelog.yml` is validated, it can be terminated the loop.

WDYT ?

```
		fixed := strings.ReplaceAll(string(data), "pull/REPLACE_ME", "pull/"+prNumber)
		if fixed == string(data) {
			break
		}
		if err := os.WriteFile(f, []byte(fixed), 0644); err != nil { 
			return fmt.Errorf("writing %s: %w", f, err)
		}
		if err := gitExec("add", "--", f); err != nil {
			return err
		}
		changed = true
		break
```

dev/requiresupdate/publish.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	if !changed {
+		return nil
+	}
```
mrodm: Should this scenario be considered an error ? If it is not found a line with `REPLACE_ME`, could it be that `elastic-package` finished with some unexpected data ?

dev/requiresupdate/requiresupdate.go
```diff
@@ -0,0 +1,281 @@
...
+		files:         writtenFiles,
+	}, nil
+}
+
+// ownerResolution is the result of reconciling CODEOWNERS against the
+// package manifest's own owner.github fallback field.
+type ownerResolution struct {
+	owners   []string // bare "org/team" entries, no leading '@'; never empty
+	mismatch string   // non-empty if CODEOWNERS and the fallback disagreed; describes both values
+}
```
mrodm, line 185: IIUC in Pull Requests there is already a check for this validation. If there is a mismatch between the the `owner.github` and the teams set in the `.github/CODEOWNERS`.

This validation runs as part of the `mage -v check` call.

dev/requiresupdate/publish.go
```diff
@@ -0,0 +1,265 @@
...
+		"--base", "main",
+		"--head", branch,
+		"--title", title,
+		"--label", "automation",
+		"--body", body,
+	}
+	for _, owner := range codeowners {
+		args = append(args, "--reviewer", "@"+owner)
+	}
+	stdout, _, err = gh.Exec(args...)
```
mrodm: What about checking if there is any `stderr` output to raise an error if that happens ?

.github/workflows/requires-update.yml
```diff
@@ -0,0 +1,69 @@
...
+        # bumps this pin deliberately via PR (gomod ecosystem).
+        run: |
+          version=$(go list -m -f '{{.Version}}' github.com/elastic/elastic-package)
+          go install "github.com/elastic/elastic-package@${version}"
+          echo "ELASTIC_PACKAGE_BIN=$(go env GOPATH)/bin/elastic-package" >> $GITHUB_ENV
+
+      - name: Configure git
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
```
mrodm: Nit. it could be used this action from observability

```
      - name: set up git user
        uses: elastic/oblt-actions/git/setup@v1
        with:
          username: 'github-actions[bot]'
          email: 'github-actions[bot]@users.noreply.github.com'
```

Based on 
https://github.com/elastic/package-spec/blob/f608d1e1094685ec7844878bded1ddad91c76b01/.github/workflows/post-dependabot.yml#L27-L31

dev/requiresupdate/requiresupdate.go
```diff
@@ -0,0 +1,281 @@
...
+		} else if p.Warning != "" {
+			skipped = append(skipped, p)
+		}
+	}
+
+	if len(applied) == 0 && len(skipped) == 0 {
+		return nil, nil
+	}
+
+	res := resolveOwner(owners, pkgName, result.Codeowner)
```
mrodm: Here there could be some conflicts when `PackageOwners` is called inside this function.

Currently, `PackageOwners` is based on the package name retrieved from the folder name.

This method was used to get the owners from the failing tests (`xUnit` files) in the daily jobs. And currently, those tests are setting as package name the folder name.

This is related to this issue  https://github.com/elastic/elastic-package/issues/3586


But, if IIUC this code assumes that `pkgName` is the name from the manifest and not the folder.


I would create a new method for this use case for now, to be sure it is working as expected. Specially, because it would be needed to work with packages located in nested directories.

Or maybe use/copy this function that I plan to introduce in this PR https://github.com/elastic/integrations/pull/19997/changes#diff-7fefe180136c03143a5c5b39d022850a58a95e13c35a776bf2501a997e70a416R26

so you get the paths with their package names defined as in the manifest

dev/codeowners/codeowners.go
```diff
@@ -19,26 +19,85 @@ import (
...
 		return fmt.Errorf("error validating packages in directory '%s': %w", packagesDir, err)
 	}
 
 	return nil
 }
 
+// PackageOwners returns the owning team(s) for packageName (and, if
+// dataStream is set, the more specific data-stream-level owner when one is
+// defined) from the CODEOWNERS file at codeownersPath.
 func PackageOwners(packageName, dataStream, codeownersPath string) ([]string, error) {
```
mrodm, line 34: It must be taken into account that this `packageName` currently for this function represents the package folder name.

Relates to https://github.com/elastic/elastic-package/issues/3586

dev/requiresupdate/publish.go
teresaromero: its not every day we update this directory, but perhaps to copy the dev folder is convenient. If this folder is exclusive for mage commands, lets just copy it, ~~i was thinking on a rename so its explicit mage scripts should use this as "internal"~~?

.github/workflows/requires-update.yml
teresaromero: Done, bumped to v7.

dev/requiresupdate/publish.go
teresaromero: Good catch — went with copying the whole `dev/` folder instead of a per-subfolder list. `updateBackportBranchContents` now does one `git checkout $SOURCE_BRANCH -- dev` instead of the eight separate `dev/<subfolder>` copies it had. That also picks up the `gitutil` relocation from the next comment for free.

dev/requiresupdate/publish.go
teresaromero: Agreed on reuse — moved it to a new top-level `dev/gitutil` (sibling of `backports`/`requiresupdate`) so both packages can import it without a parent/child dependency, and `publish.go` now uses `gitutil.Git{}` instead of its own `gitExec`. And yeah, ended up doing exactly what's suggested here: `backport_branch.sh` now copies the whole `dev/` folder in one shot instead of a per-subfolder list, so this new location (and any future one) is covered automatically.

dev/requiresupdate/requiresupdate.go
teresaromero: You're right that this assumed `pkgName` (manifest name) matches the folder name, which isn't guaranteed for nested categories. Rather than wait on #19997, I fixed the call site directly — it now passes `filepath.Base(pkgPath)` (the actual directory `ListPackages` walked) into `resolveOwner`/`PackageOwners`, so it's exactly the folder name regardless of whether the manifest name agrees. Manifest name is still used everywhere else (display, messages), just not for this lookup.

dev/codeowners/codeowners.go
teresaromero: Thanks for flagging — fixed the call site in `requiresupdate.go` to pass the folder basename instead of the manifest name into this function, so it matches what `PackageOwners` expects here.

.github/workflows/requires-update.yml
teresaromero: Switched to `elastic/oblt-actions/git/setup@v1`, kept the `github-actions[bot]` identity since that's who's actually opening these PRs/issues via `GITHUB_TOKEN`.

dev/requiresupdate/publish.go
teresaromero: Confirmed both cases are real bugs here, not expected no-ops — `s.files` always has exactly one `changelog.yml` on this path. Made both hard errors now (missing placeholder, missing changelog.yml) instead of silently continuing.

dev/requiresupdate/publish.go
teresaromero: Added a small `ghExec` wrapper that all the call sites go through now — it surfaces stderr in the error when `gh` fails, and prints a warning line when it succeeds but still wrote to stderr. Covers exactly the case you're describing.

dev/requiresupdate/publish.go
teresaromero: Done, renamed to `publish`.

dev/requiresupdate/publish.go
teresaromero: Applied your suggested wording to the block.

dev/requiresupdate/publish.go
teresaromero: Now grabs the existing mode via `os.Stat` before rewriting instead of hardcoding it.

dev/requiresupdate/publish.go
teresaromero: Added, bundled with the changelog fix above since it's the same loop.

dev/requiresupdate/requiresupdate.go
teresaromero: It falls back to bare `elastic-package` on PATH unless `ELASTIC_PACKAGE_BIN` is set — and that's exactly what CI sets after `go install`. No change needed, just confirming the intent.

dev/requiresupdate/publish.go
teresaromero: Confirmed, yes — each loop iteration handles exactly one package.

.github/workflows/requires-update.yml
teresaromero: `publishPR` does `git checkout -B <branch> origin/main` per package, to reset onto a clean base without disturbing the working tree (which still holds other packages' pending changes). That needs `origin/main` available locally, which `checkout@v7`'s default shallow single-branch clone doesn't give us — hence the fetch.

dev/requiresupdate/publish.go
teresaromero: Confirmed empirically on #19719 — GitHub auto-requests the CODEOWNERS team the moment a PR leaves draft, independent of branch protection. Dropped the explicit `--reviewer` loop; CODEOWNERS is the authoritative source for reviewers either way, and if our resolved owner ever disagrees with it, that's already surfaced as an `ownerMismatch` in the PR/issue body rather than something to patch over with a second reviewer list.

dev/requiresupdate/requiresupdate.go
teresaromero: Not fully redundant — `codeowners.Check` enforces this on human-authored PRs at CI-check time, whereas this runs inside the automation itself, before a PR even exists, and only reports (doesn't block). Since `Check` should already prevent mismatches from landing on `main`, this is belt-and-suspenders for a state that shouldn't happen. Keeping it for now as a cheap safety net — happy to drop it later if it never fires.

dev/requiresupdate/requiresupdate.go
teresaromero: Done — switched to `mage RequiresUpdate -dryRun -preview` flags, and wired workflow_dispatch inputs through so they're still reachable from the Actions UI. Smaller change than I expected, turned out to be a clean signature swap.

.github/workflows/requires-update.yml
teresaromero: Tested it — you're right, bare `go install github.com/elastic/elastic-package` inside this module resolves to the go.mod pin, not `@latest`. Simplified to that, dropped the `go list -m` lookup.

.github/workflows/requires-update.yml
teresaromero: Although there are still other actions using the stale one, is this update tracked by any robot?

dev/gitutil/git.go
mrodm, line 1: I think this file should be added into this list 
https://github.com/elastic/integrations/blob/main/.buildkite/scripts/non_package_patterns.txt

dev/requiresupdate/publish.go
mrodm: Thanks for updating this! 👍

.github/workflows/requires-update.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Run RequiresUpdate
+        # Scans packages, applies proposals, and opens one PR (or issue) per
+        # package — see dev/requiresupdate.
+        id: requires_update
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: >-
+          mage RequiresUpdate
+          -dryRun=${{ github.event.inputs.dry_run || 'false' }}
+          -preview=${{ github.event.inputs.preview || 'false' }}
```
mrodm: This variables should be updated when using `cron` event since they will not exist, is that right ? 

Moreover, usually variables/fields from the `github.event` context is better to set them as environment variables to avoid any security issue. Reference: https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable

dev/requiresupdate/requiresupdate.go
```diff
@@ -0,0 +1,282 @@
...
+
+	if len(applied) == 0 && len(skipped) == 0 {
+		return nil, nil
+	}
+
+	// codeowners.PackageOwners treats its packageName argument as the
+	// package folder's basename (used elsewhere for xUnit-based ownership
+	// lookups), not the manifest's declared name — the two usually agree
+	// but aren't guaranteed to, e.g. under nested category directories.
+	res := resolveOwner(owners, filepath.Base(pkgPath), result.Codeowner)
```
mrodm: Related to these dicussions:
- https://github.com/elastic/integrations/pull/19217#discussion_r3542324832
- https://github.com/elastic/integrations/pull/19217#discussion_r3542358510

PR #19997 has already been merged and the method `ListPackageWithNames` is already available. I think it would be better to look for the package path based on the package name instead of the package folder name. Package manifest name is ensured to be unique that it is more relevant if there are packages in nested directories. 

As `PackageOwners` is based on the folder name, I'd say that probably it is better to duplicate that method to do the same, but based on the package manifest name (from the YAML).

This will help us to keep this code valid for packages based on the nested directories, independently from the xUnit issue.

mrodm (commented):
Added a few more comments, specially about the `PackageOwners` logic.

I've created this other PR to fix some issues found while running the `backport_branch.sh` https://github.com/elastic/integrations/pull/20101

It would also be good to run a test like this one https://buildkite.com/elastic/integrations-backport/builds/307 to validate that this script still works as expected with the changes introduced here (just copying `dev/` folder). In order to test the current PR branch, it is needed to apply the changes in this commit (that should be reverted before merging) https://github.com/elastic/integrations/pull/20101/commits/87e0e3b5ab7a287e42e158def2c0b2f81b44a11b

.buildkite/pipeline.yml
```diff
@@ -165,6 +165,8 @@ steps:
         allow_failure: false
       - step: "check-buildkite-scripts"
         allow_failure: false
+    if: |
```
teresaromero: To be removed prior merge

.github/workflows/requires-update.yml
teresaromero: i've removed this step, alongside with the `--force-with-lease` flag on the git push on publishing, which has stayed as --force only. each branch is created and if the branch is still there, the changes are forced to be updated from main everytime it runs, to have clear changes. this is only run by automation, and this change was taking into consideration the branches could have been run locally and not updated.

dev/gitutil/git.go
teresaromero: I've added some additional files too that i think might belong here. WDYT? perhaps some more dev folder files should belong here too?

.github/workflows/requires-update.yml
teresaromero: i've captured this as an env https://github.com/elastic/integrations/pull/19217/changes#diff-aa005e86c6a8b23284ae46485e46a91aa1416f61c8993f86d113ff28dcc45421R65 with the condition of workflow_dispach, as this will not be available at the scheduled workflow

dev/requiresupdate/requiresupdate.go
teresaromero: Refactored in the latest commit: the original codeowners internals are fully restored (githubOwners, readGithubOwners, all private methods — unchanged from main).

 The only addition is a thin public layer:
 - Owners — a wrapper over the private struct
 - LoadOwners — parses the file once for repeated lookups across all packages
 - PackageOwnersByPath(pkgPath, dataStream) — delegates to the existing findOwnerForFile, which walks up the directory tree from the given path

Since requiresupdate already gets the exact filesystem path from ListPackagesWithNames (pkg.Path), there's no need to go through a name-based lookup at all — path-based is both simpler and strictly more correct for nested packages, since findOwnerForFile handles arbitrary nesting (not just one level deep).

dev/gitutil/git.go
mrodm: LGTM the added files :+1: 
Thanks!

.github/workflows/requires-update.yml
mrodm: TIL it could be added conditions in those `env` keys.

.github/workflows/requires-update.yml
mrodm: I think this is managed by dependabot, at least there is configuration for that:

https://github.com/elastic/integrations/blob/71e89b61bfbf47066ec4aeaf733d59c0cc46abbb/.github/dependabot.yml#L10-L17

mrodm (approved):
LGTM!
Before merging, revert the changes in the CI step to trigger the testing of all the required packages.

teresaromero:
tested on my fork: https://github.com/teresaromero/integrations/actions/workflows/requires-update.yml 
https://github.com/teresaromero/integrations/pull/1

mrodm (approved):

PR #19235, opened 2026-05-27
Scope Vale linting for integration content

theletterf:
## Summary
- Restrict the integrations Vale workflow to user-facing package docs and metadata.
- Include fallback `packages/*/docs/README.md` files only when a package lacks `_dev/build/docs/README.md`.
- Opt integrations into YAML linting through `.vale-overrides.ini`.

## Test plan
- `actionlint .github/workflows/vale-lint.yml`
- `ruby -e \"require 'yaml'; YAML.load_file('.github/workflows/vale-lint.yml')\"`
- Parsed `.vale-overrides.ini` with Python `ConfigParser`.

Fixes https://github.com/elastic/vale-rules/issues/144


.github/workflows/vale-lint.yml
```diff
@@ -31,12 +35,66 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Select files to lint
+        id: select-files
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          mapfile -t changed_files < <(
+            gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" \
```
mrodm: I think it would be better to set these two variables from the github context as environment variables to avoid security issues:
- github.repository
- github.event.pull_request.number

https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable

.vale-overrides.ini
mrodm, line 1: I guess that updating this file should not trigger testing all packages. WDYT ?

If so , please add an entry for this in these files or functions:
- https://github.com/elastic/integrations/blob/a9cfbd711fafda62db9b27f3362faa4afebf3554/.buildkite/pull-requests.json#L16-L37
- https://github.com/elastic/integrations/blob/a9cfbd711fafda62db9b27f3362faa4afebf3554/.buildkite/scripts/common.sh#L780-L806

theletterf:
@P1llus Please have a look.

.github/workflows/vale-lint.yml
```diff
@@ -31,12 +35,66 @@ jobs:
...
+          )
+
+          files_to_lint=()
+          for file in "${changed_files[@]}"; do
+            case "$file" in
+              packages/*/_dev/build/docs/README.md|\
+              packages/*/manifest.yml|\
+              packages/*/data_stream/*/fields/fields.yml|\
+              packages/*/data_stream/*/manifest.yml|\
+              packages/*/changelog.yml)
```
mrodm: Currently, there is no package in nested directories, but there could be and the CI scripts are ready for this.
So packages could be located at:
- `packages/nginx` , or
- `packages/<technology>/nginx`

Related to https://github.com/elastic/integrations/issues/17403

Could this be updated to manage this ? 

There is a mage target to list all packages, in case this could help for this matter:

```
mage listPackages
```

It would require to install go and mage into this workflow.

.github/workflows/vale-lint.yml
```diff
@@ -31,12 +35,66 @@ jobs:
...
           # Fail the workflow if Vale finds error-level issues
           # Set to 'true' to enforce style as a required check
           fail_on_error: false
+          vale-paths: |
+            packages/*/_dev/build/docs/README.md
+            packages/*/docs/README.md
+            packages/*/manifest.yml
+            packages/*/data_stream/*/fields/fields.yml
+            packages/*/data_stream/*/manifest.yml
+            packages/*/changelog.yml
```
mrodm: Same as in this other comment: https://github.com/elastic/integrations/pull/19235/changes#r3311421146

There could be in the future some packages located in nested directories.

theletterf:
@mrodm  Addressed the review feedback in `50c4401176`:

- Moved `github.repository` and `github.event.pull_request.number` into `GH_REPOSITORY` and `PR_NUMBER` env vars before using them in the shell script.
- Switched the Vale workflow trigger, selector, and `vale-paths` patterns from `packages/*/...` to `packages/**/...` so future nested package paths are handled.
- Added `.vale-overrides.ini` to the Buildkite non-package-only skip lists in `.buildkite/pull-requests.json` and `.buildkite/scripts/common.sh` to avoid package test fan-out when only that override changes.

Validation run locally:

- `actionlint .github/workflows/vale-lint.yml`.
- `ruby -e \"require 'yaml'; YAML.load_file('.github/workflows/vale-lint.yml')\" && jq empty .buildkite/pull-requests.json && bash -n .buildkite/scripts/common.sh`.
- Simulated file selection for both `packages/nginx/...` and `packages/observability/nginx/...` paths.

P1llus (approved):
Looks good from my standpoint :) seems like the build failure is still because some file is making it test all integrations

theletterf:
Follow-up update: Buildkite was still running package checks because the PR changed `.buildkite/scripts/common.sh` while trying to exempt `.vale-overrides.ini`. That made the PR fan out into package tests, and NetBox failed inside its system tests.

I changed the approach to avoid package fan-out entirely:

- Removed the repository-level `.vale-overrides.ini` from this PR.
- Removed the Buildkite config/script changes from this PR.
- Passed the YAML Vale opt-in through the `vale-overrides` input in the Vale workflow instead.

The PR diff is now only `.github/workflows/vale-lint.yml`, and Buildkite reports `No changes required CI. Skipped.`

mrodm (approved):
👍

PR #19244, opened 2026-05-27
[Kafka] Add support for Strimzi Kafka 4.1.1 logs

Alphayeeeet:
Please label as enhancement

Extended grok patterns to support Kafka 4.1.1 logs.
As test data, logs written by the Strimzi Operator on version 0.51.0 running Kafka 4.1.1 were used:
Therefore the grok pattern might differ for other Kafka operators depending on their default log pattern settings.

However as Strimzi is a well-known Kafka management operator, we would wish to include support for their default log pattern in the generic integration, as we think providing out of the box support is a benefit for everyone using this.

The support has been added in a backward compatible manner so no side effects to current deployments are expected.


Alphayeeeet:
@andrewkroh Could you provide an update, when obs-infraobs team is planning to review this PR? Thanks

stefans-elastic (commented):
Thanks for the contribution!

Here are my concerns:
- PR title is somewhat misleading, it says `Add support for Kafka 4.1.1 logs`  but it is rather about `Strimzi Kafka` (not `Kafka 4.1.1`).
- I'm not sure this integration should focus on Kafka hosting provider differences (Strimzi is just one of many such providers. Maintaining support of all of them on top of quite a vast range of Kafka versions could be difficult).

I don't think I can answer there questions by myself so asking @agithomas to take a look.

Alphayeeeet:
@stefans-elastic If a generic integration does not support different hosting providers, then this integration is not effectively usable for us. But I will change the PR title to include the Strimzi reference. However I don't see any downside of supporting different Kafka hosting providers if requirements pop up, as the ingest pipelines grok pattern is easilyto be extended to support those in a backward compatible manner (as it was done in this PR).

Alphayeeeet:
@stefans-elastic Please re-review or forward to appropriate team if possible.

Alphayeeeet:
@stefans-elastic Could you please provide an update on the review of this PR?

PR #19245, opened 2026-05-27
[o365] Keep numbers as strings through to Elasticsearch

chrisberkhout:
## Proposed commit message

```
[o365] Keep numbers as strings through to Elasticsearch

By using `decode_json_string_numbers` instead of `decode_json` in the
CEL program, numbers will arrive at the ingest pipeline as strings
matching how they were formatted in the API's JSON string.

In most cases those values can be passed straight through to
Elasticsearch to be indexed as `long` or as `keyword`, so most
conversion processing has been removed.

There are two types of exceptions to this, handled by a normalization
script early in the pipeline:
- The `o365audit` input will still provide numbers, so where values need
  to be compared, we need to ensure we have a consistently formatted
  string. That is the case for the `RecordType` field.
- Some fields arrive from the API as numbers formatted with scientific
  notation, but they are indexed as `keyword` and should not use
  scientific notation there. That is the case for the other fields
  procesed by the normalization script.

To use `decode_json_string_numbers`, the minimum agent versions needed
to be updated from "^8.18.0 || ^9.0.0" to "^8.19.0 || ^9.1.0".
```

## Checklist

## Related issues

- Relates #14428
- Relates #14289


efd6 (commented):
There are some cases here where the fields should be converted to numbers in the ingest pipeline as their semantics are numeric; `….ConditionsMatched.SensitiveInformation.{Confidence,Count,UniqueCount}`, `o365audit.ExchangeMetaData.{FileSize,RecipientCount}` and `file.size` for example (there are others).

chrisberkhout:
@efd6 

> There are some cases here where the fields should be converted to numbers in the ingest pipeline as their semantics are numeric; `….ConditionsMatched.SensitiveInformation.{Confidence,Count,UniqueCount}`, `o365audit.ExchangeMetaData.{FileSize,RecipientCount}` and `file.size` for example (there are others).

For those mapped as `long`, such as `o365.audit.ExchangeMetaData.{FileSize,RecipientCount}` and `file.size`, Elasticsearch can take the number as a string and will convert it correctly during indexing.

<details>
<summary>Expand for an example for the Kibana Dev Tools Console</summary>

```
DELETE o365-string-number-check

PUT o365-string-number-check
{
  "mappings": {
    "properties": {
      "file": {
        "properties": {
          "size": {
            "type": "long"
          }
        }
      },
      "o365": {
        "properties": {
          "audit": {
            "properties": {
              "ExchangeMetaData": {
                "properties": {
                  "FileSize": {
                    "type": "long"
                  },
                  "RecipientCount": {
                    "type": "long"
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}

POST o365-string-number-check/_doc/integer?refresh=true
{
  "file": {
    "size": "1000"
  },
  "o365": {
    "audit": {
      "ExchangeMetaData": {
        "FileSize": "13405",
        "RecipientCount": "2"
      }
    }
  }
}

POST o365-string-number-check/_doc/float-string?refresh=true
{
  "file": {
    "size": "1000.0"
  },
  "o365": {
    "audit": {
      "ExchangeMetaData": {
        "FileSize": "13405.0",
        "RecipientCount": "2.0"
      }
    }
  }
}

POST o365-string-number-check/_doc/scientific-string?refresh=true
{
  "file": {
    "size": "1e9"
  },
  "o365": {
    "audit": {
      "ExchangeMetaData": {
        "FileSize": "1.3405e4",
        "RecipientCount": "2e0"
      }
    }
  }
}

GET o365-string-number-check/_search
{
  "fields": [
    "file.size",
    "o365.audit.ExchangeMetaData.FileSize",
    "o365.audit.ExchangeMetaData.RecipientCount"
  ],
  "_source": true,
  "sort": ["_id"]
}
```

</details>

It's a different case for `o365.audit.PolicyDetails.Rules.ConditionsMatched.SensitiveInformation.{Confidence,Count,UniqueCount}`. Because `o365.audit.PolicyDetails` is mapped as `flattened`, leaf field values such as for `Confidence`, `Count` and `UniqueCount` are mapped by default (and in this case) as `keyword`. When the field value of a flattened field is requested, the response will provide the full JSON object from `_source` ([ref](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/flattened#search-fields-flattened)). It's possible to map sub-fields with known names to other types, but currently we're not doing that.

In summary, the indexed values are unchanged, but the `_source` document has string-typed values in it where it had number-typed values before. Do you think it's important to preserve the value types in the `_source` document?

efd6 (approved):
I've looked into my concern and your response and I agree. There are no Elastic SIEM rules that would be broken by this change. It *is* possible that this change could break users who have painless scripts that do numeric operations on `params._source` in their detection rules, or third-party consumers of `_source`, but those scripts and uses are already unsafe.

PR #17398, opened 2026-02-12
[lambda_logs] fix parsing failure when `message` is `string`

ghost:
## Proposed commit message

this PR fixes `lambda_logs` parsing failure when `aws.lambda.message` receives a `string` value instead of an `object`.

## Checklist

- [ ] I have reviewed [tips for building integrations](https://www.elastic.co/docs/extend/integrations/tips-for-building) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [ ] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [ ] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

## Author's Checklist

- [ ] 

## Related issues

- Closes #15375

PR #19250, opened 2026-05-27
[AWS][Lambda] Fix AWS Lambda data stream aws.lambda.message handling

Kavindu-Dodan:
## Proposed commit message

`aws.lambda.message` is defined as a `flattened` field. However, Lambda integration's  JSON pipeline ignore checks to validate this. This causes JSON pipeline to parse but Elastic indexing to fail with `expecting token of type [START_OBJECT] but found [VALUE_STRING] `

See `fields.yaml` of this integration for specific field definition (see extraction below),
```yaml
- name: aws.lambda
  type: group
  fields:
  - name: message
    type: flattened
```

If the payload contains a non-compliant content, then this fix preserve the original message at root level `message` field.

See test changes and newly added tests to understand the change better. 

Additionally, `message` section is now added in plain text mode as well.

## Checklist

### DIFF
--- a/packages/aws/data_stream/lambda_logs/_dev/test/pipeline/test-awslambda-json-with-message.log-expected.json
+++ b/packages/aws/data_stream/lambda_logs/_dev/test/pipeline/test-awslambda-json-with-message.log-expected.json
@@ -1,0 +1,41 @@
+{
+    "expected": [
+        {
+            "@timestamp": "2025-03-20T11:07:58.223Z",
+            "aws": {
+                "lambda": {
+                    "message": {
+                        "@timestamp": "2026-05-28T00:00:00.000Z",
+                        "ecs.version": "1.6.0",
+                        "invokedTrigger": "CloudWatchEvent",
+                        "log.level": "info",
+                        "message": "Lambda triggered",
+                        "otelcol.component.id": "awslambda",
+                        "otelcol.component.kind": "receiver",
+                        "otelcol.signal": "logs",
+                        "resource": {
+                            "service.instance.id": "id-123",
+                            "service.name": "service-name-redacted"
+                        }
+                    }
+                }
+            },
+            "cloud": {
+                "provider": "aws",
+                "service": {
+                    "name": "aws_lambda"
+                }
+            },
+            "ecs": {
+                "version": "8.11.0"
+            },
+            "event": {
+                "original": "{\n    \"log.level\": \"info\",\n    \"@timestamp\": \"2026-05-28T00:00:00.000Z\",\n    \"message\": \"Lambda triggered\",\n    \"resource\": {\n        \"service.instance.id\": \"id-123\",\n        \"service.name\": \"service-name-redacted\",\n        \"service.version\": \"\"\n    },\n    \"otelcol.component.id\": \"awslambda\",\n    \"otelcol.component.kind\": \"receiver\",\n    \"otelcol.signal\": \"logs\",\n    \"invokedTrigger\": \"CloudWatchEvent\",\n    \"ecs.version\": \"1.6.0\"\n}"
+            },
+            "message": "Lambda triggered",
+            "tags": [
+                "preserve_original_event"
+            ]
+        }
+    ]
+}
--- a/packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
+++ b/packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
@@ -339,6 +339,17 @@ processors:
     target_field: aws.lambda.message
     ignore_missing: true
     ignore_failure: true
+    if: "ctx.parsed?.message instanceof Map"
+
+- set:
+    field: message
+    copy_from: parsed.message
+    if: "ctx.parsed?.message != null && !(ctx.parsed.message instanceof Map)"
+
+- set:
+    field: message
+    copy_from: event.original
+    if: "ctx.message == null && ctx.aws?.lambda?.message == null"
 
 - rename:
     field: parsed.users
@@ -411,12 +422,7 @@ processors:
     target_field: aws.lambda.message
     ignore_missing: true
     ignore_failure: true
-    if: "ctx.parsed instanceof Map && !(ctx.parsed.containsKey('message') || ctx.parsed.containsKey('record') || ctx.parsed.containsKey('_aws') || ctx.parsed.containsKey('time') || ctx.parsed.containsKey('timestamp'))"
-
-- set:
-    field: aws.lambda.message
-    copy_from: event.original
-    if: "ctx.parsed == null"
+    if: "ctx.parsed instanceof Map && !(ctx.parsed.containsKey('record') || ctx.parsed.containsKey('_aws') || ctx.parsed.containsKey('time') || ctx.parsed.containsKey('timestamp'))"
 
 - remove:
     field: parsed
--- a/packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
+++ b/packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
@@ -422,7 +422,7 @@ processors:
     target_field: aws.lambda.message
     ignore_missing: true
     ignore_failure: true
-    if: "ctx.parsed instanceof Map && !(ctx.parsed.containsKey('record') || ctx.parsed.containsKey('_aws') || ctx.parsed.containsKey('time') || ctx.parsed.containsKey('timestamp'))"
+    if: "ctx.parsed instanceof Map && !(ctx.parsed.containsKey('record') || ctx.parsed.containsKey('_aws') || ctx.parsed.containsKey('time') )"
 
 - remove:
     field: parsed
--- a/packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
+++ b/packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
@@ -358,12 +358,6 @@ processors:
     ignore_failure: true
 
 - rename:
-    field: parsed.timestamp
-    target_field: "@timestamp"
-    ignore_missing: true
-    ignore_failure: true
-
-- rename:
     field: parsed.tracing.xray_trace_id
     target_field: aws.lambda.xray_trace_id
     ignore_missing: true
--- a/packages/aws/changelog.yml
+++ b/packages/aws/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "6.19.1"
+  changes:
+    - description: Fix aws.lambda.message flattened field handling.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/19250
 - version: "6.19.0"
   changes:
     - description: Use new `release` field for agentless deployment mode to establish as beta.
--- a/packages/aws/data_stream/lambda_logs/_dev/test/pipeline/test-awslambda-dotnet-json.log-expected.json
+++ b/packages/aws/data_stream/lambda_logs/_dev/test/pipeline/test-awslambda-dotnet-json.log-expected.json
@@ -4,7 +4,6 @@
             "@timestamp": "2025-03-20T11:07:58.223Z",
             "aws": {
                 "lambda": {
-                    "message": "{@users} have joined the group",
                     "request_id": "8f711428-7e55-46f9-ae88-2a65d4f85fc5",
                     "trace_id": "1-6408af34-50f56f5b5677a7d763973804",
                     "users": [
@@ -34,6 +33,7 @@
             "log": {
                 "level": "Information"
             },
+            "message": "{@users} have joined the group",
             "tags": [
                 "preserve_original_event"
             ]
@@ -42,7 +42,6 @@
             "@timestamp": "2025-03-20T11:07:58.223Z",
             "aws": {
                 "lambda": {
-                    "message": "{users} have joined the group",
                     "request_id": "8f711428-7e55-46f9-ae88-2a65d4f85fc5",
                     "trace_id": "1-6408af34-50f56f5b5677a7d763973804",
                     "users": [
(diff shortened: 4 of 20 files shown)


agithomas:
@Kavindu-Dodan, is this PR a duplicate of https://github.com/elastic/integrations/pull/17398?

cc @gpop63

packages/aws/data_stream/lambda_logs/_dev/test/pipeline/test-awslambda-json-with-message.log-expected.json
```diff
@@ -0,0 +1,41 @@
...
+                    "name": "aws_lambda"
+                }
+            },
+            "ecs": {
+                "version": "8.11.0"
+            },
+            "event": {
+                "original": "{\n    \"log.level\": \"info\",\n    \"@timestamp\": \"2026-05-28T00:00:00.000Z\",\n    \"message\": \"Lambda triggered\",\n    \"resource\": {\n        \"service.instance.id\": \"id-123\",\n        \"service.name\": \"service-name-redacted\",\n        \"service.version\": \"\"\n    },\n    \"otelcol.component.id\": \"awslambda\",\n    \"otelcol.component.kind\": \"receiver\",\n    \"otelcol.signal\": \"logs\",\n    \"invokedTrigger\": \"CloudWatchEvent\",\n    \"ecs.version\": \"1.6.0\"\n}"
+            },
+            "message": "Lambda triggered",
```
kcreddy, line 35: When `parsed.message` is a string (non-Map), there is duplication - i.e., `parsed.message` is stored as `message` and also as `aws.lambda.message.message`. Is this intended behaviour?

MichaelKatsoulis: I believe we should drop the `parsed.message` from parsed before the fallback rename so the string only lives at root message.

Then, my suggestion is to keep this PR and close https://github.com/elastic/integrations/pull/17398

kcreddy: cc: @agithomas

Kavindu-Dodan: I intentionally added this so there's always a root level `message` iff the original event contains a `message` section. I think this helps users as `message` is an important detail

In this example, user now see the Lambda triggered detail at root level. Similarly, any Lambda log is visible through `message` section. And `aws.lambda.message.message` preserves the original payload as is for any other need as a MAP.

markjandejong:
## Additional Root Cause: `parsed.timestamp` blocks the catchall rename

I've been debugging this same issue in production and found an additional bug beyond the string `message` problem that both this PR and #17398 are addressing.

### The deeper issue

The pipeline's **final catchall rename** (`rename parsed → aws.lambda.message`) has this condition:

```yaml
if: "ctx.parsed instanceof Map && !(ctx.parsed.containsKey('message') || ctx.parsed.containsKey('record') || ctx.parsed.containsKey('_aws') || ctx.parsed.containsKey('time') || ctx.parsed.containsKey('timestamp'))"
```

For Powertools-style logs (which have a `timestamp` field), the date processor extracts the value into `@timestamp` but **never removes `parsed.timestamp` from the map**. This means the condition `ctx.parsed.containsKey('timestamp')` is always true, so the entire `if` evaluates to false, and the catchall rename **never fires** for any Powertools log.

After the string `message` is handled, `parsed` still contains all the remaining structured fields (`cold_start`, `function_name`, `service`, `name`, `function_version`, etc.) — but they're orphaned and never make it into `aws.lambda.message`.

### The result

Instead of getting a rich flattened object like:
```json
{"log": "Creating Allow Policy", "name": "MyLogger", "function_version": "$LATEST", "cold_start": false, ...}
```

You get an incomplete object (or just a string at root `message`), losing all the structured context that Powertools provides.

### Proposed fix (3 changes)

**1. Remove `parsed.timestamp` after the date processor extracts it:**
```yaml
- date:
    if: "ctx.parsed?.timestamp != null"
    field: parsed.timestamp
    target_field: "@timestamp"
    formats: ["yyyy-MM-dd HH:mm:ss,SSSZ", "ISO8601"]
    ignore_failure: true

- remove:
    field: parsed.timestamp
    ignore_missing: true
    ignore_failure: true
```

**2. Rename `parsed.message` → `parsed.log` instead of copying to root `message`:**
```yaml
- rename:
    field: parsed.message
    target_field: parsed.log
    if: "ctx.parsed?.message != null && ctx.parsed?.message instanceof String"
    ignore_failure: true
```

This keeps the string value *inside* `parsed` so it flows into `aws.lambda.message.log` via the catchall rename — preserving all sibling fields alongside it.

**3. Remove `timestamp` from the catchall rename condition** (since `parsed.timestamp` is now removed earlier, the check is no longer needed):

Original:
```yaml
if: "ctx.parsed instanceof Map && !(ctx.parsed.containsKey('message') || ctx.parsed.containsKey('record') || ctx.parsed.containsKey('_aws') || ctx.parsed.containsKey('time') || ctx.parsed.containsKey('timestamp'))"
```

Fixed:
```yaml
if: "ctx.parsed instanceof Map && !(ctx.parsed.containsKey('message') || ctx.parsed.containsKey('record') || ctx.parsed.containsKey('_aws') || ctx.parsed.containsKey('time'))"
```

### Results

- **No data loss**: All Powertools structured fields (`cold_start`, `function_name`, `service`, `name`, `xray_trace_id`, etc.) end up in `aws.lambda.message` as a proper flattened object
- **No duplication**: The string `message` lives only at `aws.lambda.message.log`, not duplicated at root `message` AND `aws.lambda.message.message` (addresses @kcreddy's concern)
- **Consistent with `flattened` mapping**: `aws.lambda.message` always gets a Map, never a bare string
- **Aligns with @MichaelKatsoulis's suggestion**: We drop `parsed.message` from `parsed` before the fallback rename — but we preserve its content under a different key (`log`) rather than discarding it entirely

Happy to open a separate PR with these changes + tests if that's preferred, or contribute them here. You can see my version [HERE](https://github.com/markjandejong/elastic-integrations/blob/fix/lambda-logs/packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml)

Kavindu-Dodan:
@markjandejong really appreciate the detailed explanation and the solution for the issue. 

IMO this deserves a dedicated PR with a focused tests given this specific PR focuses on Lambda fix. Happy to review once it's open.

packages/aws/data_stream/lambda_logs/_dev/test/pipeline/test-awslambda-json-with-message.log-expected.json
Kavindu-Dodan: > I believe we should drop the parsed.message from parsed before the fallback rename so the string only lives at root message.

In one of the versions I had this. But it felt wrong to remove just one field from the original payload. This is why I used copy from set processor

markjandejong:
> @markjandejong really appreciate the detailed explanation and the solution for the issue.
> 
> IMO this deserves a dedicated PR with a focused tests given this specific PR focuses on Lambda fix. Happy to review once it's open.

@Kavindu-Dodan  A new PR will be a duplicate of this issue as they both address the exact same problem with regards to Lambda log ingestion of JSON payloads. I figure I would provide input of my findings in hopes to come to a unified resolution.

Kavindu-Dodan:
> > @markjandejong really appreciate the detailed explanation and the solution for the issue.
> > IMO this deserves a dedicated PR with a focused tests given this specific PR focuses on Lambda fix. Happy to review once it's open.
> 
> @Kavindu-Dodan A new PR will be a duplicate of this issue as they both address the exact same problem with regards to Lambda log ingestion of JSON payloads. I figure I would provide input of my findings in hopes to come to a unified resolution.

Thanks for insisting on this issue. I added the change you proposed through commit https://github.com/elastic/integrations/pull/19250/changes/67466a3d132e7ee0c343979af1ee5e1420aef43c

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
markjandejong: I believe this to be redundant as it gets extracted [at lines 10-15](https://github.com/Kavindu-Dodan/integrations/blob/67466a3d132e7ee0c343979af1ee5e1420aef43c/packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml#L10-L15)

markjandejong, line 16: parsed.timestamp should be removed after extraction here. 
```yaml
- remove:
    field: parsed.timestamp
    ignore_missing: true
    ignore_failure: true
```

markjandejong, line 286: I believe this to be redundant as it gets extracted at lines 10-15

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
```diff
@@ -411,12 +422,7 @@ processors:
     target_field: aws.lambda.message
     ignore_missing: true
     ignore_failure: true
-    if: "ctx.parsed instanceof Map && !(ctx.parsed.containsKey('message') || ctx.parsed.containsKey('record') || ctx.parsed.containsKey('_aws') || ctx.parsed.containsKey('time') || ctx.parsed.containsKey('timestamp'))"
-
-- set:
-    field: aws.lambda.message
-    copy_from: event.original
-    if: "ctx.parsed == null"
+    if: "ctx.parsed instanceof Map && !(ctx.parsed.containsKey('record') || ctx.parsed.containsKey('_aws') || ctx.parsed.containsKey('time') )"
```
markjandejong, line 419: In  order for this to work, the field `parsed.timstamp` needs to be removed. This should happen after the the field is extracted after line 15

```yaml
- date:
    if: "ctx.parsed?.timestamp != null"
    field: parsed.timestamp
    target_field: "@timestamp"
    formats: ["yyyy-MM-dd HH:mm:ss,SSSZ"]
    ignore_failure: true

- remove:
    field: parsed.timestamp
    ignore_missing: true
    ignore_failure: true
```

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
Kavindu-Dodan: Yes, thank you for noticing. I have removed this with https://github.com/elastic/integrations/pull/19250/changes/6f25baf9b8532791bdf22fc77690d76e3628a4cb

Kavindu-Dodan: No I do not think we should remove this. `aws.lambda.message` is expected to preserve the original message. See here - https://github.com/elastic/integrations/pull/19250/changes#diff-47c5bd851c7038676fd767648faf7ee6b269124328340a869ba2326cf56187e6R59-R63

packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
Kavindu-Dodan: See this remark - https://github.com/elastic/integrations/pull/19250#discussion_r3335273562

Kavindu-Dodan:
@MichaelKatsoulis @kcreddy @markjandejong  can I get another round of review 🙏

kcreddy (approved):
The duplication (root `message` + `aws.lambda.message.message`) won't cause indexing failures. It's a storage/clarity trade-off, not a correctness issue.

The core issue for the string-into-flattened-field bug should be fixed by this PR. Deferring to @elastic/obs-infraobs-integrations as the integration owners to make the final call.

MichaelKatsoulis (approved):

MichaelKatsoulis:
Hey @Kavindu-Dodan ,

while testing this I noticed `aws-lambda-plaintext.yml` does not populate the message field for any events (`REPORT, END, INIT_START`). Their grok patterns does extract fields like `aws.lambda.event_type` and `aws.lambda.metrics.*` but not message. The document lands with message blank in Discover.

`aws-lambda-json.yml` already handles this :

```
- set:
    field: message
    copy_from: event.original
    if: "ctx.message == null && ctx.aws?.lambda?.message == null"
```

Could we add the same fallback to the plaintext pipeline? 
One small thing I noticed: the `START` pattern has a trailing `%{GREEDYMULTILINE:message}` that matches no characters on a `START RequestId: ... Version: $LATEST` line, so the field comes out as an empty string rather than null. So I tested with this condition  to cover both:

```
- set:
    field: message
    copy_from: event.original
    if: "ctx.message == null || ctx.message == ''"
    ignore_failure: true
```

I believe it is worth adding it in this PR

Kavindu-Dodan:
> Hey @Kavindu-Dodan ,
> 
> while testing this I noticed `aws-lambda-plaintext.yml` does not populate the message field for any events (`REPORT, END, INIT_START`). Their grok patterns does extract fields like `aws.lambda.event_type` and `aws.lambda.metrics.*` but not message. The document lands with message blank in Discover.
> 
> `aws-lambda-json.yml` already handles this :
> 
> ```
> - set:
>     field: message
>     copy_from: event.original
>     if: "ctx.message == null && ctx.aws?.lambda?.message == null"
> ```
> 
> Could we add the same fallback to the plaintext pipeline? One small thing I noticed: the `START` pattern has a trailing `%{GREEDYMULTILINE:message}` that matches no characters on a `START RequestId: ... Version: $LATEST` line, so the field comes out as an empty string rather than null. So I tested with this condition to cover both:
> 
> ```
> - set:
>     field: message
>     copy_from: event.original
>     if: "ctx.message == null || ctx.message == ''"
>     ignore_failure: true
> ```
> 
> I believe it is worth adding it in this PR

Thanks @MichaelKatsoulis . Yes I think we can add this too

See https://github.com/elastic/integrations/pull/19250/changes/7c0f33cad1fa6d876b807f2b19c51d55259bb61e, this is done :) 


If no further remarks by end of the day, I will merge this PR.

Kavindu-Dodan:
@agithomas could you please have a final look 🙏

agithomas (approved):
LGTM!

PR #19259, opened 2026-05-28
[Thor] Include campaign_id in logs to enable filtering and correlation by scan campaign

SimonKoetting:
Add campaign_id to the events to let the customers better identify the origin of the scans.
The campaign ID helps identify and group scans belonging to the same assessment in THOR APT Scanner. Without it, correlating results can be difficult, especially in large environments where many scans and findings overlap. Adding the campaign ID makes it easier to track related runs, separate investigations, and analyze results more reliably.

_relationship between scans and campaigns: a scan campaign is essentially a set of scans that share the same settings. The actual usage can differ depending on the customer’s workflow, but campaigns allow users to group scans that should behave in the same way.
For example, a customer could use one campaign for quick scans to get a fast overview of a system, while using another campaign for deeper scans of selected systems. Having the campaign_id available in the logs makes it easier to filter, group, and correlate scan results by campaign, such as showing only results from the “deep dive” campaign._


efd6 (commented):
The change LGTM, but I think we need background for the commit message; there is no issue and the proposed commit message doesn't explain the need or provide any documentation links to support it. Can you add those please?

SimonKoetting:
@efd6 I just updated the description to include a bit more context, do you think this is enough?

efd6:
Do you have doc links?

efd6 (approved):
Proposed commit message:
```
nextron_thor: add campaign_id to events

THOR Cloud associates every scan with a campaign — a named group of
scans sharing the same configuration (e.g. "quick overview" vs "deep
dive"). The campaign_id is not present in the raw scan log lines but
is available as scan metadata via the THOR Cloud API (GET /v1/scan).

Extract campaign_id from the scan search response and inject it into
each log event during collection. This lets users filter, group, and
correlate scan results by the campaign that produced them.

ref: https://thor-microsoft-defender-guide.nextron-systems.com/en/latest/usage/thor-cloud.html
```

PR #19278, opened 2026-05-28
[aws] Add var_groups for credential type selection with Cloud Connector support

Omolola-Akinleye:
## Summary

We are bumping to a major version because we are introducing changes:
-  Updating   [hbs ](https://github.com/elastic/integrations/pull/18762/changes#diff-738d65e4d73f15fb8346d84bdc4c49a249161f589a57430927731016f61012f9)  support cloud connectors
-  Update Manifest to Utilize Package Spec Version feature called `var_groups` 
-  Hiding cloud connector var groups ux for AWS Services that don't support cloud connector.

Mirrors upstream PR #18762 (by @seanrathier) plus the 7.0.0 version bump.

### Manifest / contract

- `format_version`: 3.4.0 → 3.6.0 (enables `var_groups`)
- `version`: 6.14.2 → 7.0.0 (breaking-change)
- Kibana floor: `^8.19.4 || ^9.2.1` → `^9.4.0` (8.x line dropped; 9.x floor raised for var_groups UI + auth.aws runtime) according [format compatibility documentation](https://github.com/elastic/elastic-package/blob/main/docs/howto/format_version.md).
- New `var_groups: credential_type` selector with 6 options: `cloud_connectors`, `direct_access_key`, `temporary_access_key`, `assume_role`, `assume_role_external_id`, `shared_credentials`
- New credential vars: `assume_role_duration`, `assume_role_expiry_window`, `supports_cloud_connectors`
- `external_id` is now `secret: true`
- `hide_in_var_group_options` added to 13 inputs across awshealth / billing / config / dynamodb / elb / lambda / rds / sns / sqs / transitgateway / securityhub_cspm / inspector / guardduty (Cloud Connector path not yet validated for these)

Related: elastic/ingest-dev#8788
Source: #18762


packages/aws/data_stream/lambda_logs/elasticsearch/ingest_pipeline/aws-lambda-json.yml
```diff
@@ -134,301 +148,357 @@ processors:
...
     target_field: aws.lambda.message
     ignore_missing: true
     ignore_failure: true
     if: "ctx.parsed instanceof Map && !(ctx.parsed.containsKey('record') || ctx.parsed.containsKey('_aws') || ctx.parsed.containsKey('time') )"
 
+- set:
+    tag: set_aws_lambda_message_2
+    field: aws.lambda.message
+    copy_from: event.original
+    if: "ctx.parsed == null"
```
MichaelKatsoulis: This seems to create a bug and buildkite fails. Is this change intentional?
If yes I think it needs to be
```
if: "ctx.parsed == null && ctx.aws?.lambda?.message == null"
```
or else you are overwriting the `aws.lambda.message` field as `ctx.parsed` will be null because in the rename processor above the `parsed` field gets deleted.

Omolola-Akinleye: Got it ! Thanks for the feedback.

MichaelKatsoulis (approved):
LGTM!

Omolola-Akinleye:
AWS Guard Duty with Cloud Connector and var groups enabled
Pasted Graphic
Agent Based Var Groups
Pasted Graphic 1
Heallthy Integrations created with cloud connector
Amazon GuardDuty
Cloud connector listing supported integrations
Screenshot 2026-06-18 at 5 14 17 PM
AWS Guardduty Findings
image

Omolola-Akinleye:
Need to wait to merge PR until the next [integration dev forum on Wednesday](https://elastic.slack.com/archives/C010LT16CFR/p1781606104072319)

andrewkroh (commented):
I think we should separate the processing changes from the var groups changes. We can add tags to all of the processors independent of var_groups. If we do that change before we branch off the integration, then there's less of a delta between the 8.19 release branch and main, which will make future backporting of pipeline fixes easier.

PR #17968, opened 2026-03-23
[Cato Networks][Audit] Add cato networks audit data-stream

sharadcrest:
## Proposed commit message

This release includes audit data stream and associated dashboard.

Cato Networks fields are mapped to their corresponding ECS fields where possible.

Test samples were derived from live data samples, which were subsequently
sanitized.

## Checklist

- [X] I have reviewed [tips for building integrations](https://www.elastic.co/docs/extend/integrations/tips-for-building) and this pull request is aligned with them.
- [X] I have verified that all data streams collect metrics or logs.
- [X] I have added an entry to my package's `changelog.yml` file.
- [X] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [X] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

### To test the cato networks package:

- Clone integrations repo.
- Install elastic package locally.
- Start elastic stack using elastic-package.
- Move to integrations/packages/cato_networks directory.
- Run the following command to run tests.

> elastic-package test
```
--- Test results for package: cato_networks - START ---
╭───────────────┬─────────────┬───────────┬────────────────────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE       │ DATA STREAM │ TEST TYPE │ TEST NAME                                                              │ RESULT │ TIME ELAPSED │
├───────────────┼─────────────┼───────────┼────────────────────────────────────────────────────────────────────────┼────────┼──────────────┤
│ cato_networks │             │ asset     │ dashboard cato_networks-95cbb672-baae-4508-b188-24a51b4d36da is loaded │ PASS   │        970ns │
│ cato_networks │ audit       │ asset     │ index_template logs-cato_networks.audit is loaded                      │ PASS   │        197ns │
│ cato_networks │ audit       │ asset     │ ingest_pipeline logs-cato_networks.audit-0.1.0 is loaded               │ PASS   │        164ns │
╰───────────────┴─────────────┴───────────┴────────────────────────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: cato_networks - END   ---
Done
Run pipeline tests for the package
2026/03/23 15:55:22  INFO elastic-package v0.118.0 version-hash 816ceecf (build time: 2025-12-30T18:33:37+05:30)
[...]

PR #17775, opened 2026-03-12
[Cato Networks][Event] Add cato networks event data-stream

sharadcrest:
## Proposed commit message

The initial release includes event data stream and associated dashboard.

Cato Networks fields are mapped to their corresponding ECS fields where possible.

Test samples were derived from live data samples, which were subsequently
sanitized.

## Checklist

- [X] I have reviewed [tips for building integrations](https://www.elastic.co/docs/extend/integrations/tips-for-building) and this pull request is aligned with them.
- [X] I have verified that all data streams collect metrics or logs.
- [X] I have added an entry to my package's `changelog.yml` file.
- [X] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).
- [X] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices) 

### To test the cato networks package:

- Clone integrations repo.
- Install elastic package locally.
- Start elastic stack using elastic-package.
- Move to integrations/packages/cato_networks directory.
- Run the following command to run tests.

> elastic-package test
```
--- Test results for package: cato_networks - START ---
╭───────────────┬─────────────┬───────────┬────────────────────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE       │ DATA STREAM │ TEST TYPE │ TEST NAME                                                              │ RESULT │ TIME ELAPSED │
├───────────────┼─────────────┼───────────┼────────────────────────────────────────────────────────────────────────┼────────┼──────────────┤
│ cato_networks │             │ asset     │ dashboard cato_networks-9d363f59-caa0-4b84-af82-dd85d789a7dd is loaded │ PASS   │        1.9µs │
│ cato_networks │             │ asset     │ search cato_networks-7d5f3596-eb8a-43cc-be1c-66d80fb18853 is loaded    │ PASS   │        594ns │
│ cato_networks │             │ asset     │ search cato_networks-fa1fade5-b107-4f85-9d3a-2a2103ed3ab0 is loaded    │ PASS   │        122ns │
│ cato_networks │             │ asset     │ search cato_networks-fb73f336-127f-4d49-94dd-709c0da4c5d1 is loaded    │ PASS   │        120ns │
│ cato_networks │ event       │ asset     │ index_template logs-cato_networks.event is loaded                      │ PASS   │        161ns │
[...]

PR #19307, opened 2026-06-01
[Cato Networks] Initial release of Cato Networks with Audit and Event data streams

### LINKED ISSUE #18163: [New Integration] Cato Networks
## Description 

[Cato Networks](https://www.catonetworks.com/) delivers a cloud-native Secure Access Service Edge (SASE) platform: SD-WAN, firewall, SWG, ZTNA, and related controls for sites, remote users, and cloud resources. Security and IT teams use Cato for network visibility, policy enforcement, and detection across WAN and internet-bound traffic.

The Elastic integration should ingest Cato audit logs (administrative/account activity) and events (security and operational events) for alerting and correlation use cases including: **audit trail** for admin and configuration changes; **monitoring** of firewall, access, and threat-related events; **correlation** with endpoint and identity data in Elastic Security; **compliance** reporting and evidence retention; and **operational** troubleshooting of policy and connectivity issues tied to event timelines.

## Architecture

**Primary path (expected):** Pull-based collection via the **[Cato GraphQL API](https://api.catonetworks.com/documentation/)**, authenticated with API credentials scoped to the account. The public schema exposes feed-oriented queries aligned with SIEM use cases:

| Data type | GraphQL (per API reference) |
|--------|-----------------------------|
| Administrative / account audit | Query **`auditFeed`** — payload types include **`AccountAuditData`**, **`AuditFeed`**, per-record **`AuditRecord`**, with field metadata via **`AuditField`** / **`AuditFieldName`** |
| Security / operational events | Query **`eventsFeed`** — payload types include **`EventsFeedData`**, **`EventsFeedAccountRecords`**, per-record **`EventRecord`**, with field metadata via **`EventField`** / **`EventFieldName`** |

sharad-elastic:
## Proposed commit message

The initial release includes Audit and Event data streams and associated dashboard.

Cato Networks fields are mapped to their corresponding ECS fields where possible.

sanitized.

## Checklist

### To test the cato networks package:

- Clone integrations repo.
- Move to integrations/packages/cato_networks directory.

> elastic-package test
```
--- Test results for package: cato_networks - START ---
╭───────────────┬─────────────┬───────────┬────────────────────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE       │ DATA STREAM │ TEST TYPE │ TEST NAME                                                              │ RESULT │ TIME ELAPSED │
├───────────────┼─────────────┼───────────┼────────────────────────────────────────────────────────────────────────┼────────┼──────────────┤
│ cato_networks │             │ asset     │ dashboard cato_networks-95cbb672-baae-4508-b188-24a51b4d36da is loaded │ PASS   │      1.196µs │
│ cato_networks │             │ asset     │ dashboard cato_networks-9d363f59-caa0-4b84-af82-dd85d789a7dd is loaded │ PASS   │        185ns │
│ cato_networks │             │ asset     │ search cato_networks-7d5f3596-eb8a-43cc-be1c-66d80fb18853 is loaded    │ PASS   │        117ns │
│ cato_networks │             │ asset     │ search cato_networks-fa1fade5-b107-4f85-9d3a-2a2103ed3ab0 is loaded    │ PASS   │        131ns │
│ cato_networks │             │ asset     │ search cato_networks-fb73f336-127f-4d49-94dd-709c0da4c5d1 is loaded    │ PASS   │        140ns │
│ cato_networks │ audit       │ asset     │ index_template logs-cato_networks.audit is loaded                      │ PASS   │        200ns │
│ cato_networks │ audit       │ asset     │ ingest_pipeline logs-cato_networks.audit-0.1.0 is loaded               │ PASS   │        202ns │
│ cato_networks │ event       │ asset     │ index_template logs-cato_networks.event is loaded                      │ PASS   │        139ns │
│ cato_networks │ event       │ asset     │ ingest_pipeline logs-cato_networks.event-0.1.0 is loaded               │ PASS   │        155ns │
╰───────────────┴─────────────┴───────────┴────────────────────────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: cato_networks - END   ---
Done
2026/05/22 10:03:50  INFO elastic-package v0.118.0 version-hash 816ceecf (build time: 2025-12-30T18:33:37+05:30)
2026/05/22 10:03:50  INFO elastic-stack: 8.18.0
--- Test results for package: cato_networks - START ---
╭───────────────┬─────────────┬───────────┬───────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE       │ DATA STREAM │ TEST TYPE │ TEST NAME                                 │ RESULT │ TIME ELAPSED │
├───────────────┼─────────────┼───────────┼───────────────────────────────────────────┼────────┼──────────────┤
│ cato_networks │ audit       │ pipeline  │ (ingest pipeline warnings test-audit.log) │ PASS   │ 263.309613ms │
│ cato_networks │ audit       │ pipeline  │ test-audit.log                            │ PASS   │ 283.989688ms │
│ cato_networks │ event       │ pipeline  │ (ingest pipeline warnings test-event.log) │ PASS   │ 268.753777ms │
│ cato_networks │ event       │ pipeline  │ test-event.log                            │ PASS   │ 338.472947ms │
╰───────────────┴─────────────┴───────────┴───────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: cato_networks - END   ---
Done
2026/05/22 10:03:51  INFO elastic-package v0.118.0 version-hash 816ceecf (build time: 2025-12-30T18:33:37+05:30)
2026/05/22 10:03:51  INFO elastic-stack: 8.18.0
--- Test results for package: cato_networks - START ---
No test results
--- Test results for package: cato_networks - END   ---
Done
Run script tests for the package
PKG cato_networks
[no test files]
--- Test results for package: cato_networks - START ---
No test results
--- Test results for package: cato_networks - END   ---
Done
2026/05/22 10:03:51  INFO elastic-package v0.118.0 version-hash 816ceecf (build time: 2025-12-30T18:33:37+05:30)
--- Test results for package: cato_networks - START ---
╭───────────────┬─────────────┬───────────┬──────────────────────────┬────────┬──────────────╮
│ PACKAGE       │ DATA STREAM │ TEST TYPE │ TEST NAME                │ RESULT │ TIME ELAPSED │
├───────────────┼─────────────┼───────────┼──────────────────────────┼────────┼──────────────┤
│ cato_networks │ audit       │ static    │ Verify sample_event.json │ PASS   │ 175.670713ms │
│ cato_networks │ event       │ static    │ Verify sample_event.json │ PASS   │ 155.810352ms │
╰───────────────┴─────────────┴───────────┴──────────────────────────┴────────┴──────────────╯
--- Test results for package: cato_networks - END   ---
Done
2026/05/22 10:03:52  INFO elastic-package v0.118.0 version-hash 816ceecf (build time: 2025-12-30T18:33:37+05:30)
2026/05/22 10:03:52  INFO elastic-stack: 8.18.0
2026/05/22 10:03:52  INFO Installing package...
2026/05/22 10:03:52  INFO License text found in "/home/devuser/github/integrations/LICENSE.txt" will be included in package
2026/05/22 10:04:04  INFO Running test for data_stream "audit" with configuration 'default'
2026/05/22 10:04:13  INFO Setting up independent Elastic Agent...
2026/05/22 10:04:28  INFO Setting up service...
2026/05/22 10:05:10  INFO Validating test case...
2026/05/22 10:05:11  INFO Tearing down service...
2026/05/22 10:05:12  INFO Write container logs to file: /home/devuser/github/integrations/build/container-logs/cato_networks-1779424512158122738.log
2026/05/22 10:05:15  INFO Tearing down agent...
2026/05/22 10:05:15  INFO Write container logs to file: /home/devuser/github/integrations/build/container-logs/elastic-agent-1779424515521114845.log
2026/05/22 10:05:26  INFO Running test for data_stream "event" with configuration 'default'
2026/05/22 10:05:34  INFO Setting up independent Elastic Agent...
2026/05/22 10:05:48  INFO Setting up service...
2026/05/22 10:06:37  INFO Validating test case...
2026/05/22 10:06:38  INFO Tearing down service...
2026/05/22 10:06:39  INFO Write container logs to file: /home/devuser/github/integrations/build/container-logs/cato_networks-1779424599100967047.log
2026/05/22 10:06:41  INFO Tearing down agent...
2026/05/22 10:06:41  INFO Write container logs to file: /home/devuser/github/integrations/build/container-logs/elastic-agent-1779424601545014175.log
2026/05/22 10:06:51  INFO Uninstalling package...
--- Test results for package: cato_networks - START ---
╭───────────────┬─────────────┬───────────┬───────────┬────────┬─────────────────╮
│ PACKAGE       │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT │    TIME ELAPSED │
├───────────────┼─────────────┼───────────┼───────────┼────────┼─────────────────┤
│ cato_networks │ audit       │ system    │ default   │ PASS   │  1m7.176221491s │
│ cato_networks │ event       │ system    │ default   │ PASS   │ 1m12.539761063s │
╰───────────────┴─────────────┴───────────┴───────────┴────────┴─────────────────╯
--- Test results for package: cato_networks - END   ---
Done
```

## Related issues
- Closes #18163

## Screenshots
image
image

---
Note: This integration follows a phased development process where individual data streams were reviewed and merged into a feature branch through separate PRs:

Event data stream: https://github.com/elastic/integrations/pull/17775 
Audit data stream: https://github.com/elastic/integrations/pull/17968
Update changelog entry PR: https://github.com/elastic/integrations/pull/19158

All PR's have been reviewed and merged in this feature branch, which is now ready for integration into the main branch.


.github/CODEOWNERS
```diff
@@ -163,6 +163,7 @@
 /packages/carbonblack_edr @elastic/security-service-integrations @elastic/sit-crest-contractors
 /packages/cassandra @elastic/obs-infraobs-integrations
 /packages/cassandra_otel @elastic/obs-infraobs-integrations
+/packages/cato_networks @elastic/security-service-integrations @elastic/sit-crest-contractors
```
chrisberkhout: Is this the right owner? I think most integrations with network flow data are owned by @elastic/integration-experience at the moment

narph: @qcorporation , looks like network/dns related data ingested , can you confirm the transfer to your team?

sharad-elastic: confirmed with @P1llus, transferred ownership to @elastic/integration-experience team.

ShourieG:
@vera-review-bot review

packages/cato_networks/kibana/dashboard/cato_networks-95cbb672-baae-4508-b188-24a51b4d36da.json
```diff
@@ -0,0 +1,2120 @@
...
+                },
+                "gridData": {
+                    "h": 12,
+                    "i": "pie-05",
+                    "w": 12,
+                    "x": 48,
+                    "y": 0
+                },
+                "panelIndex": "pie-05",
+                "title": "Events by Name [Logs Cato Networks]",
```
efd6:
```suggestion
                "title": "Events by Name",
```

packages/cato_networks/kibana/dashboard/cato_networks-95cbb672-baae-4508-b188-24a51b4d36da.json
```diff
@@ -0,0 +1,2120 @@
...
+                },
+                "gridData": {
+                    "h": 12,
+                    "i": "pie-06",
+                    "w": 12,
+                    "x": 60,
+                    "y": 0
+                },
+                "panelIndex": "pie-06",
+                "title": "Events by Value [Logs Cato Networks]",
```
efd6:
```suggestion
                "title": "Events by Value",
```

packages/cato_networks/kibana/dashboard/cato_networks-9d363f59-caa0-4b84-af82-dd85d789a7dd.json
```diff
@@ -0,0 +1,3875 @@
...
+                },
+                "gridData": {
+                    "h": 18,
+                    "i": "d79ef6a7-a464-4cbc-8793-ac422bb13938",
+                    "w": 48,
+                    "x": 0,
+                    "y": 131
+                },
+                "panelIndex": "d79ef6a7-a464-4cbc-8793-ac422bb13938",
+                "title": "Threat Essential Details [Logs Cato Networks]",
```
efd6:
```suggestion
                "title": "Threat Essential Details",
```

packages/cato_networks/kibana/dashboard/cato_networks-9d363f59-caa0-4b84-af82-dd85d789a7dd.json
```diff
@@ -0,0 +1,3875 @@
...
+                },
+                "gridData": {
+                    "h": 17,
+                    "i": "f03f3bde-fb4c-4fef-ba36-c46854f82ce2",
+                    "w": 48,
+                    "x": 0,
+                    "y": 149
+                },
+                "panelIndex": "f03f3bde-fb4c-4fef-ba36-c46854f82ce2",
+                "title": "Event Essential Details [Logs Cato Networks]",
```
efd6:
```suggestion
                "title": "Event Essential Details",
```

packages/cato_networks/kibana/dashboard/cato_networks-9d363f59-caa0-4b84-af82-dd85d789a7dd.json
```diff
@@ -0,0 +1,3875 @@
...
+                },
+                "gridData": {
+                    "h": 17,
+                    "i": "26a29836-c557-4fc6-b366-c702a6e9b709",
+                    "w": 48,
+                    "x": 0,
+                    "y": 114
+                },
+                "panelIndex": "26a29836-c557-4fc6-b366-c702a6e9b709",
+                "title": "Application Essential Details [Logs Cato Networks]",
```
efd6:
```suggestion
                "title": "Application Essential Details",
```

packages/cato_networks/img/cato_networks-event.png
efd6, line 1: Can this dashboard be split in sensible way to reduce the visualisation count? It's right on the boundary of being too complex.

packages/cato_networks/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1104 @@
...
+      target_field: cato_networks.event.time
+      formats:
+        - UNIX_MS
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+      if: ctx.json?.time != null && ctx.json.time != ''
+  - set:
+      field: '@timestamp'
```
ilyannn, line 1020: **`@timestamp` has no fallback — events where `time` is missing or unparseable get ingest-time instead of event-time.**

`@timestamp` is only ever copied from `cato_networks.event.time`, which is parsed from `json.time` (epoch-ms, L881–896). `time_str` *is* parsed into `cato_networks.event.time_str` a few lines later (L897–907) but is then **removed at L1040** and never used as a fallback.

**Concrete example** from the test corpus — one sample event carries:
```json
{ "time_str": "2026-02-26T09:29:16Z", "event_type": "Security", ... }
```
If that record arrived **without** the `time` key (or with `"time": ""`), the guard `if: ctx.json?.time != null && ctx.json.time != ''` skips `date_time`, `cato_networks.event.time` is never created, the `set @timestamp` (copy_from + ignore_empty_value) is a no-op, and Elasticsearch stamps **ingestion time**. The correctly-parsed `time_str` value is discarded a few processors later.

The **audit** pipeline has the same pattern: `@timestamp` comes only from `creation_date` (L1523–1543, whose `on_failure` removes the field), while `insertion_date` is parsed (L1544+) but never wired as a fallback.

Suggested fix — add a fallback `set` immediately after the `time_str` parse:
```yaml
- set:
    field: '@timestamp'
    tag: set_@timestamp_from_event_time_str
    copy_from: cato_networks.event.time_str
    ignore_empty_value: true
    if: ctx.cato_networks?.event?.time == null
```
And the equivalent in the audit pipeline using `insertion_date` as a fallback for `creation_date`.

:robot: Comment by Claude Code/Opus 4.8 under my supervision

packages/cato_networks/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1104 @@
...
+      field: event.kind
+      tag: set_event_kind
+      value: event
+  - append:
+      field: event.category
+      tag: append_event_category_network
+      value: network
+      allow_duplicates: false
+  - append:
+      field: event.type
```
ilyannn: **High priority: `event.type` is hard-coded to `info` and `event.outcome` is never set, even though `action` carries allow/deny/success semantics.**

Every event indexes as `event.kind: event`, `event.category: [network]`, `event.type: [info]` regardless of what happened. But the sample and test data already show `action` values like `Block`, `Prompt`, and connectivity events with `Succeeded` — the allow/deny/outcome signal is present in the data and should reach ECS.

As-is, an analyst filtering on `event.type:(allowed|denied)` or `event.outcome:(success|failure)` matches **zero** Cato events — blocked/denied traffic is invisible to any ECS-based detection or dashboard.

Suggested direction:
- `action: Allow` → `event.type: [allowed]`
- `action: Block` / `Drop` → `event.type: [denied]`
- `action: Succeeded` → `event.outcome: success`; `action: Failed` → `event.outcome: failure`

(Exact mapping is yours — the point is that `action` should drive `event.type` and `event.outcome` rather than being flattened to a constant `info`.)

:robot: Comment by Claude Code/Opus 4.8 under my supervision

packages/cato_networks/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1104 @@
...
+      tag: convert_tls_inspection_to_boolean
+      target_field: cato_networks.event.tls_inspection
+      type: boolean
+      ignore_missing: true
+      on_failure:
+        - append:
+            field: error.message
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - rename:
+      field: json.traffic_direction
```
ilyannn, line 1052: **Question:** `json.traffic_direction` is renamed only to `cato_networks.event.traffic_direction` — was mapping it to ECS `network.direction` considered?

The sample data carries values like `OUTBOUND`, which maps cleanly to `network.direction: outbound`. Without it, Network visualisations and SIEM rules keyed on `network.direction:(inbound|outbound|internal|external)` won't surface Cato traffic at all.

If some Cato values (e.g. `WANBOUND`/`LANBOUND`) don't cleanly map onto the ECS allowed set, a note in the docs or a partial mapping for the common cases (`INBOUND`/`OUTBOUND`) would help. Is there a reason to keep direction vendor-namespaced only?

:robot: Comment by Claude Code/Opus 4.8 under my supervision

packages/cato_networks/data_stream/event/fields/ecs.yml
```diff
@@ -0,0 +1,12 @@
+- name: observer.product
+  external: ecs
+  type: constant_keyword
+  value: SASE Cloud
```
ilyannn: **Question:** is the `observer.product` difference between the two data streams intentional?

- `data_stream/event/fields/ecs.yml:4` → `value: SASE Cloud`
- `data_stream/audit/fields/ecs.yml:4` → `value: Cato SASE Cloud Platform`

`observer.vendor` (`Cato Networks`) and `observer.type` (`Network Security`) already match across both streams. If these refer to the same product, two distinct `observer.product` values will split any cross-dataset dashboard or correlation rule that keys off it. Worth aligning to a single value if so.

:robot: Comment by Claude Code/Opus 4.8 under my supervision

ilyannn (commented):
Automated review of the Cato Networks initial release (findings surfaced by static analysis of the ingest pipelines, CEL inputs, and field definitions).

packages/cato_networks/data_stream/event/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,134 @@
...
+      {
+        "Header": {
+          "Content-Type": ["application/json"],
+          "x-api-key": [state.api_key],
+        },
+        "Body": {
+          "query": state.query,
+          "variables": {
+            "accountIDs": state.account_ids,
+            ?"marker": state.?cursor.marker,
```
haetamoudi, line 61:
```suggestion
            ?"marker": state.?cursor.?marker,
```
If `cursor` exists but `marker` is absent (e.g. after a fresh cursor write without pagination), this will throw a runtime error

packages/cato_networks/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1104 @@
...
+  - set:
+      field: event.kind
+      tag: set_event_kind
+      value: event
+  - append:
+      field: event.category
+      tag: append_event_category_network
+      value: network
+      allow_duplicates: false
+  - append:
```
haetamoudi: every other `append` processor have `allow_duplicates: false`. Is it on purpose not to set it here too?

packages/cato_networks/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1104 @@
...
+      tag: append_cato_networks_event_mitre_attack_subtechniques_into_threat_technique_subtechnique_name
+      value: '{{{cato_networks.event.mitre_attack_subtechniques}}}'
+      allow_duplicates: false
+      if: ctx.cato_networks?.event?.mitre_attack_subtechniques != null
+  - rename:
+      field: json.mitre_attack_tactics
+      tag: rename_mitre_attack_tactics
+      target_field: cato_networks.event.mitre_attack_tactics
+      ignore_missing: true
+  - append:
```
haetamoudi, line 623: what the pipeline will produce
```
"threat": {
    "tactic": {
        "name": ["Command and Control (TA0011)"]
    }
}
``` 
However here is what ECS expects https://www.elastic.co/guide/en/ecs/current/ecs-threat.html
```
"threat": {
    "tactic": {
        "name": ["Command and Control"],
        "id": ["TA0011"]
    }
}
```

We could parse the value with something like 
```
- dissect:
    field: cato_networks.event.mitre_attack_tactics
    tag: dissect_mitre_tactic
    pattern: "%{_tmp_tactic_name} (%{_tmp_tactic_id})"
    ignore_failure: true
- append:
    field: threat.tactic.name
    value: '{{{_tmp_tactic_name}}}'
    if: ctx._tmp_tactic_name != null
- append:
    field: threat.tactic.id
    value: '{{{_tmp_tactic_id}}}'
    if: ctx._tmp_tactic_id != null
```

packages/cato_networks/data_stream/event/agent/stream/cel.yml.hbs
efd6: That's not the case in CEL; once an access has been made optional, all subsequent accesses are optional.

packages/cato_networks/data_stream/event/elasticsearch/ingest_pipeline/default.yml
sharad-elastic: No longer applicable

sharad-elastic:
@ShourieG @ilyannn @efd6 All comments has been addressed. can you please re-review?

packages/cato_networks/_dev/build/docs/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Agentless integrations allow you to collect data without having to manage Elastic Agent in your cloud. They make manual agent deployment unnecessary, so you can focus on your data instead of the agent that collects it. For more information, refer to [Agentless integrations](https://www.elastic.co/guide/en/serverless/current/security-agentless-integrations.html) and the [Agentless integrations FAQ](https://www.elastic.co/guide/en/serverless/current/agentless-integration-troubleshooting.html).
+
+Agentless deployments are only supported in Elastic Serverless and Elastic Cloud environments. This functionality is in beta and is subject to change. Beta features are not subject to the support SLA of official GA features.
+
+### Agent-based installation
+
+Elastic Agent must be installed. For more details, check the Elastic Agent [installation instructions](docs-content://reference/fleet/install-elastic-agents.md). You can install only one Elastic Agent per host.
+
+### configure
```
haetamoudi: nitpick
```suggestion
### Configure
```

packages/cato_networks/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,149 @@
...
+            "Header": {
+              "Content-Type": ["application/json"],
+              "x-api-key": [state.api_key],
+            },
+            "Body": {
+              "query": state.query,
+              "variables": {
+                "timeFrame": "utc.{" + window_from + "--" + window_to + "}",
+                "accountIDs": state.account_ids,
+                ?"marker": state.?cursor.?marker,
```
efd6:
```suggestion
                ?"marker": state.?cursor.marker,
```
The explanation for this is in https://pkg.go.dev/github.com/google/cel-go/cel#hdr-Syntax_Changes-OptionalTypes under "Field Selection":

> The optional syntax in field selection is denoted as `obj.?field`. In other words, if a field is set, return `optional.of(obj.field)`, else `optional.none()`. **The optional field selection is viral in the sense that after the first optional selection all subsequent selections or indices are treated as optional**, i.e. the following expressions are equivalent:
>
> ```
> obj.?field.subfield
> obj.?field.?subfield
> ```

packages/cato_networks/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,149 @@
...
+                  {
+                    "last_timestamp": body.data.auditFeed.?to.optMap(v, timestamp(v).format("2006-01-02/15:04:05")).orValue(window_to),
+                  },
+                "want_more": body.data.auditFeed.?hasMore.orValue(false),
+              }
+            :
+              {
+                "events": {
+                  "error": {
+                    "message": "POST " + state.url.trim_right("/") + "/api/v1/graphql2" + ": " + body.?errors[0].?message.orValue("unknown error"),
```
efd6:
```suggestion
                    "message": "POST " + state.url.trim_right("/") + "/api/v1/graphql2" + ": " + body.?errors[0].message.orValue("unknown error"),
```

packages/cato_networks/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,149 @@
...
+                "events": (body.data.auditFeed.?accounts.orValue(null) != null) ?
+                  body.data.auditFeed.accounts.map(account, (account.?records.orValue(null) != null) ?
+                    account.records.map(record,
+                      {
+                        "message": record.fieldsMap.encode_json(),
+                      }
+                    )
+                  :
+                    []
+                  ).flatten()
```
efd6:
```suggestion
                  body.data.auditFeed.accounts.map(account, account.?records.orValue(null) != null,
                    account.records.map(record,
                      {
                        "message": record.fieldsMap.encode_json(),
                      }
                    )
                  ).flatten()
```
(untested)

sharad-elastic:
@efd6  I’ve addressed the review comments. Could you please re-review when you get a chance?

Also, could you please add the `changelog-link-check:skip` label so the build can pass? Unfortunately, I don’t have the required permissions to add it myself.

efd6 (approved):
LGTM, but not a codeowner.

packages/cato_networks/manifest.yml
```diff
@@ -0,0 +1,108 @@
...
+description: Collect logs from Cato Networks with Elastic Agent.
+type: integration
+categories:
+  - security
+  - network_security
+  - vpn_security
+  - firewall_security
+conditions:
+  kibana:
+    version: ^8.18.0 || ^9.0.0
```
ilyannn: Admittedly I have not created integrations for a while, but the [current guidance we have](https://github.com/elastic/integration-skills/blob/main/skills/create-integration/references/create-workflow.md#phase-2-scaffold-the-package) (as of 2 months ago) is to use

```
format_version: "3.4.2"
conditions.kibana.version: "^8.19.0 || ^9.1.0"
```

Since some processors (Fable highlighted geo enrichment) might work differently, I'd recommend we test specifically on a lower version if we have to support it for some reason.

packages/cato_networks/data_stream/event/elasticsearch/ingest_pipeline/default.yml
ilyannn: Looks like this was implemented

packages/cato_networks/data_stream/audit/_dev/test/pipeline/test-audit.log-expected.json
```diff
@@ -0,0 +1,1253 @@
...
+                "preserve_original_event"
+            ]
+        },
+        {
+            "@timestamp": "2025-12-12T03:21:49.000Z",
+            "cato_networks": {
+                "audit": {
+                    "account_id": "11111",
+                    "admin": "System User #40276",
+                    "admin_id": "40276",
```
ilyannn: I've noticed that this ID is not added to `related.user` – I recommend we do that. 

Also I've asked Fable for more suggestions and here they are:

---

  Clear ECS gaps

  1. Audit: the acting admin is never mapped to user.* (biggest gap).
  Every audit record has admin ("System User #31363") and admin_id, but the pipeline keeps both only under
  cato_networks.audit.* and appends just the ID to related.user (audit/.../default.yml:146-150). The
  expected output confirms audit docs contain no user.* fields at all — for a config-change audit log, the
  actor is the single most queryable thing. Should be:
  - admin → user.name, admin_id → user.id
  - both appended to related.user (today only the ID is)

  2. Event: device_id → host.id.
  A stable endpoint identifier (699b0f1b352cc37b4a5e4d4d), renamed to custom only
  (event/.../default.yml:327). host.id is exactly this, and it makes the host entity join work across
  events when device names collide.

  3. Event: threat_reference → event.reference.
  It's a vendor KB URL (https://support.catonetworks.com/hc/...), custom-only today (line 914).
  event.reference is defined precisely for this.

  4. Event: request_size / response_size → http.request.bytes / http.response.bytes.
  Converted to long but kept custom (line 708). These records are HTTP transactions (they carry
  http_request_method/http_response_code, already mapped), so the byte counters belong in http.*.

  5. risk_level → host.risk.static_level looks like a mis-mapping.
  Line 732-734 copies the IPS/DNS-Protection detection's risk level onto host risk. But risk_level
  qualifies the threat/event, not the endpoint — a host that visits one risky domain isn't a "Medium-risk
  host". The conventional home is event.severity (numeric, e.g. Low/Medium/High → 21/47/73 like other
  security integrations) with the string kept in the custom field. Worth reconsidering rather than adding.

  More for related.*

  - related.hosts should include the domains seen on the event. Today it only carries the device name +
  lowercased host.name (lines 218-233). The fixtures are full of host identifiers that never land there:
  domain_name → url.domain and dns_query → dns.question.name. related.hosts is defined as "all hostnames or
  other host identifiers seen on your event", and adding them is what makes a "was this domain seen
  anywhere" query work across the package.
  - src_site_name → related.user when src_is_site_or_vpn == "SDP User". Fixture evidence: for SDP users the
  "site" is the person — record 1 has src_site_name: "John Doe", record 3 "jane.doe@example.com". A
  conditional append (guarded on the SDP User discriminator so real site names don't pollute it) would
  capture identity that's currently dropped.
  - related.ip is already complete (all 7 IP fields feed it), and ad_name does reach related.user — no
  action there.

  Weaker / judgment-call candidates

  - account_id/account_name → organization.id/organization.name — it's the Cato tenant; useful in
  multi-account setups.
  - Static observer.vendor: Cato Networks / observer.product / observer.type, and pop_name → observer.name
  — the PoP is the enforcement point; other firewall/SASE packages set these.
  - public_ip → possibly source.nat.ip (the user's public egress IP). Semantics need confirming against
  Cato docs before mapping; it's already in related.ip, which covers the search use case.
  - Audit insertion_date → event.created (vendor ingestion time; currently only a @timestamp fallback,
  line 1562).

packages/cato_networks/data_stream/audit/_dev/test/pipeline/test-audit.log-expected.json
```diff
@@ -0,0 +1,1253 @@
...
+                                            "possible_platform": [
+                                                "X1600_LTE",
+                                                "X1600_5G"
+                                            ],
+                                            "value": "X1600_LTE"
+                                        },
+                                        {
+                                            "automatically_managed": false,
+                                            "links": [
+                                                "lan",
```
ilyannn, line 476: Is it desired that we produce duplicate strings here?

ilyannn (changes requested):

ilyannn:
Please don't forget to run the linting and formatting steps that currently fail the build pipeline

packages/cato_networks/changelog.yml
```diff
@@ -0,0 +1,12 @@
+# later versions go on top
+- version: '0.1.0'
+  changes:
+    - description: Initial release of package.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/19307
+    - description: Add support for audit data-stream.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/17968
```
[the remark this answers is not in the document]
ilyannn: i agree with this comment – there is no need for intermediate PRs in the changelog

packages/cato_networks/data_stream/event/fields/ecs.yml
```diff
@@ -0,0 +1,12 @@
...
+- name: observer.type
+  external: ecs
+  type: constant_keyword
```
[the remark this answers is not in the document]
ilyannn: This has been fixed

packages/cato_networks/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,147 @@
...
+                // hasMore=false: advance last_timestamp to response "to", clear paging state via clobber
+                "cursor": body.data.auditFeed.?hasMore.orValue(false) ?
+                  {
+                    ?"marker": body.data.auditFeed.?marker,
+                    "last_timestamp": window_from,
+                    "window_to": window_to,
+                  }
+                :
+                  {
+                    "last_timestamp": body.data.auditFeed.?to.optMap(v, timestamp(v).format("2006-01-02/15:04:05")).orValue(window_to),
```
ilyannn: **Does the Cato API echo `to` back in RFC3339, or in the custom `2006-01-02/15:04:05` layout used for requests?**

`timestamp(v)` only parses RFC3339. This program formats all *outgoing* request timestamps as `2006-01-02/15:04:05` (see `window_from`/`window_to` above). If `auditFeed.to` comes back in that same non-RFC3339 shape, `timestamp(v)` would raise a parse error, and `.orValue(window_to)` only covers the optional-absent case — not a runtime parse error — so the program would abort on the last page of every window instead of falling back to `window_to`.

Could you confirm what format `to` is actually returned in? If it's non-RFC3339, something like this would guard it:
```
"last_timestamp": body.data.auditFeed.?to.optMap(v,
  has(try(timestamp(v))) ? timestamp(v).format("2006-01-02/15:04:05") : window_to
).orValue(window_to),
```

ilyannn (commented):
:robot: Issues found by Fable, reviewed by me and posted by Sonnet 5

packages/cato_networks/data_stream/audit/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,14 @@
...
+  url: http://{{Hostname}}:{{Port}}
+  api_key: xxxx
+data_stream:
+  vars:
+    account_ids:
+      - 12121
+      - 67676
+    interval: 5m
+    preserve_original_event: true
+assert:
```
ilyannn, line 14: **Missing `wait_for_data_timeout` in system test config**

This config doesn't set `wait_for_data_timeout` (same for the event data stream's counterpart). It should be set to `1m`.
```yaml
wait_for_data_timeout: 1m
```

packages/cato_networks/data_stream/event/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,14 @@
...
+  url: http://{{Hostname}}:{{Port}}
+  api_key: xxxx
+data_stream:
+  vars:
+    account_ids:
+      - 12345
+      - 67890
+    interval: 30s
+    preserve_original_event: true
+assert:
```
ilyannn, line 14: **Missing `wait_for_data_timeout` in system test config**

This config doesn't set `wait_for_data_timeout` (same for the audit data stream's counterpart). It should be set to `1m`.
```yaml
wait_for_data_timeout: 1m
```

packages/cato_networks/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1162 @@
...
+      field: message
+      tag: remove_message
+      ignore_missing: true
+      description: The `message` field is no longer required if the document has an `event.original` field.
+      if: ctx.event?.original != null
+  - json:
+      field: event.original
+      tag: json_event_original
+      target_field: json
+  - fingerprint:
```
ilyannn, line 38: **Can `json.time` / `json.event_id` ever be absent — and if so, is routing that event into `on_failure` intended?**

This `fingerprint` has no `ignore_missing` and no `on_failure`. Later in this same pipeline, `json.time` is treated as optional (guarded by `if: ctx.json?.time != null && ctx.json.time != ''` with a `json.time_str` fallback), and `rename_event_id` uses `ignore_missing: true`. If an event without `time` or `event_id` is possible, it would currently throw here, get routed to `on_failure`, and be mislabeled `event.kind: pipeline_error` — losing deterministic `_id` de-duplication in the process. All four pipeline test fixtures happen to carry both fields, so this path isn't exercised.

If missing values are possible, something like this would avoid the failure path:
```yaml
  - fingerprint:
      fields:
        - json.event_id
        - json.time
      tag: fingerprint_event_id_time
      target_field: _id
      ignore_missing: true
      on_failure:
        - append:
            field: error.message
```

packages/cato_networks/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1162 @@
...
+      fields:
+        - json.event_id
+        - json.time
+      tag: fingerprint_event_id_time
+      target_field: _id
+  - set:
```
ilyannn, line 49: **Is `event.category: [network]` / `event.kind: event` intentional for IPS and DNS Protection detections, or should these get `intrusion_detection` / `alert` treatment?**

This is hardcoded for every document, but the fixtures show these are security detections — `event_type: Security` with `event_sub_type` IPS / DNS Protection, populated `threat_name`/`threat_type`, and `mitre_attack_tactics` (which the pipeline does map into `threat.tactic.*`/`threat.technique.*` further down). The expected test output confirms all four sample events get `kind=event`, `category=[network]` only, including the IPS and DNS Protection ones.

Wondering if that's a deliberate simplification, or if these detections should also carry `intrusion_detection` and `event.kind: alert`, e.g.:
```yaml
  - append:
      field: event.category
      tag: append_event_category_intrusion_detection
      value: intrusion_detection
      allow_duplicates: false
      if: >-
        ctx.cato_networks?.event?.threat_name != null ||
        ['IPS', 'DNS Protection'].contains(ctx.cato_networks?.event?.event_sub_type)
  - set:
      field: event.kind
      tag: set_event_kind_alert
      value: alert
      if: ctx.cato_networks?.event?.threat_name != null
```

packages/cato_networks/data_stream/event/_dev/test/pipeline/test-event.log
```diff
@@ -0,0 +1,4 @@
...26c06d1d9cdc","event_sub_type":"Internet Firewall","event_type":"Security","flow_id":"6427317971344338944","full_path_url":"https://sec-tunnel.com/","host_ip":"1.128.0.0","host_mac":"7c:1e:52:8f:4f:96","http_request_method":"GET","internalId":"8826c06d1d9cdc","internal_id":"8826c06d1d9cdc","ip_protocol":"TCP","network_rule":"Akira Impossible Travel","os_type":"OS_WINDOWS","os_version":"11","pop_name":"Miami","prompt_action":"","public_ip":"216.160.83.56","qos_priority":"10","request_size":"0","rule_id":"8512269566766760473","rule_name":"Prompt for Risky Categories","server_ip":"175.16.199.0","src_country":"United States","src_country_code":"US","src_ip":"1.128.0.0","src_is_site_or_vpn":"SDP User","src_isp_ip":"81.2.69.192","src_port":"60593","src_site_id":"1000033","src_site_name":"John Doe","static_host":"false","tcp_acceleration":"true","time":"1772098156792","time_str":"2026-02-26T09:29:16Z","tls_inspection":"true","traffic_direction":"OUTBOUND","url":"https://sec-tunnel.com/","user_agent":"Python-urllib/3.10","user_id":"1000033","user_name":"John Doe","user_reference_id":"2905856774","visible_device_id":"7c-1e-52-8f-4f-96","vpn_user_email":"john.doe@acmecorp46.onmicrosoft.com"}
```
ilyannn: **Pipeline fixture never exercises the Connectivity (ApiKey login) event shape**

All four records here are `event_type: Security`. The docker mock (`_dev/deploy/docker/files/config.yml`) and the system test also return `event_type: Connectivity` "ApiKey login successfully" records with a different field set (no IPs/URLs; has `authentication_type`, `login_type`, `api_name`). That shape is never covered at the pipeline-test level, so its parsing/categorization is unverified.

Suggest adding one anonymized Connectivity/login record and regenerating:
```bash
elastic-package test pipeline --data-streams event --generate
```

packages/cato_networks/kibana/search/cato_networks-fa1fade5-b107-4f85-9d3a-2a2103ed3ab0.json
```diff
@@ -0,0 +1,76 @@
...
+            }
+        },
+        "sort": [
+            [
+                "@timestamp",
+                "desc"
+            ]
+        ],
+        "timeRestore": false,
+        "title": "Event Essential details [Logs Cato Networks]",
```
ilyannn: **Orphaned saved-search asset — not referenced by any dashboard**

None of the three shipped `kibana/search/*.json` objects (this one, `cato_networks-7d5f3596-eb8a-43cc-be1c-66d80fb18853.json`, `cato_networks-fb73f336-127f-4d49-94dd-709c0da4c5d1.json`) is referenced by any dashboard — all four dashboards have zero `search`-type references, and none of these IDs appear in `kibana/dashboard/*.json`. The dashboards instead embed equivalent searches by value ("Event Essential Details", "Application Essential Details", "Threat Essential Details" panels). Result: dead assets duplicating the by-value panels, maintained in two places.

Pick one model — either wire the dashboard panels to reference these saved searches (add a `references` entry + `panelRefName` on the panel), or delete the three unused `kibana/search/*.json` files if by-value embedding is intentional.

packages/cato_networks/kibana/dashboard/cato_networks-95cbb672-baae-4508-b188-24a51b4d36da.json
```diff
@@ -0,0 +1,1909 @@
...
+                                                        "orderDirection": "desc",
+                                                        "otherBucket": true,
+                                                        "parentFormat": {
+                                                            "id": "terms"
+                                                        },
+                                                        "secondaryFields": [],
+                                                        "size": 5
+                                                    },
+                                                    "scale": "ordinal",
+                                                    "sourceField": "cato_networks.audit.change.after.country_raw"
```
ilyannn, line 803: **"Host Geo Country" choropleth is bound to a raw country name, not an ISO code — will likely render empty**

This panel (titled "Host Geo Country", line 925) binds its region map to `cato_networks.audit.change.after.country_raw` — "Raw country name configured for the site or entity," i.e. the country of a *site config change*, not a host's geolocation. Two issues: (a) the title misrepresents the field, and (b) the working choropleths on the Event dashboard join on `source.geo.country_iso_code`/`destination.geo.country_iso_code` (ISO codes) — feeding raw country *names* into an identically-configured choropleth means the EMS region join won't match, so the map likely renders empty for most/all rows. The field is also only populated on site-config-change audit events, so it's sparse by nature.

Suggest binding to an ISO-code geo field consistent with the event dashboard (adding one in the pipeline if audit events carry geo), and renaming the panel, e.g. "Changed Site Country." If no ISO field exists for audit, a terms table/bar on `country_raw` would be more reliable than a region map.

packages/cato_networks/_dev/build/docs/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+1. In the top search bar in Kibana, search for **Integrations**.
+2. In the search bar, type **Cato Networks**.
+3. Select the **Cato Networks** integration from the search results.
+4. Select **Add Cato Networks** to add the integration.
+5. Enable and configure only the collection methods which you will use.
+
+    * To **Collect Cato Networks logs via API**, you'll need to:
+
+        - Configure **URL**.
```
ilyannn: **Setup steps never state the actual API base URL value**

The README explains how to obtain the API Key and Account ID (lines 43-52) but never tells the user what to enter for the required **URL** field here — the `url` var is `required: true` with only "Base URL of the Cato Networks API" as its description.
```markdown
- **URL**: the base URL of the Cato Networks API (for example, `https://api.catonetworks.com`).
```

packages/cato_networks/data_stream/event/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1162 @@
+---
+description: Pipeline for processing event logs.
+processors:
+  - set:
+      field: ecs.version
+      tag: set_ecs_version
+      value: 9.3.0
+  - terminate:
```
ilyannn, line 8: **This `terminate` processor looks unreachable given how the CEL programs shape errors**

Both CEL programs (`audit/agent/stream/cel.yml.hbs`, `event/agent/stream/cel.yml.hbs`) return errors as a single object — `"events": {"error": {...}}` — which is retry semantics: the input logs the error and does *not* publish a document to the pipeline. But this `terminate` processor is keyed to `ctx.error?.message != null && ctx.message == null && ctx.event?.original == null`, which only matches an *array-shape* error document that actually reaches the pipeline. Since object-shape errors never reach the pipeline, this processor (and its counterpart in the audit pipeline) never fires — the two halves disagree on the error model.

If object-shape/retry is the intended model for these transient errors (rate limit, 5xx, null feed), the simplest fix is removing the unreachable processor from both pipelines. Alternatively, if error documents should be indexed, switch the CEL error branches to array shape (`"events": [{"error": {...}}]`).

efd6: This is not true; in the case that the `events` field holds an object, the cursor is dropped and the event is specifically logged if it holds an `error` field, but it is still published.

https://github.com/elastic/beats/blob/69c5d2b37992300ada506ec95b35e1859dfaaf3c/x-pack/filebeat/input/cel/input.go#L694-L724

packages/cato_networks/data_stream/audit/agent/stream/cel.yml.hbs
muskan-agarwal26: Implemented the guard, but using the try(...).as(...) idiom rather than the literal snippet:

```
"last_timestamp": body.data.auditFeed.?to.optMap(v,
  try(timestamp(v), "error").as(t, !has(t.error) ? t.format("2006-01-02/15:04:05") : window_to)
).orValue(window_to),
```
Two reasons for the deviation: has(try(timestamp(v))) doesn't compile, since has() only accepts a field selection (e.g. has(t.error)), not a bare expression; and the suggested form calls timestamp(v) a second time in the true branch, which is unguarded and would still throw on a malformed value. Binding the try result once with .as(t, ...) evaluates the parse a single time and reuses it. This matches the existing idiom in the repo (e.g. packages/cyera), and I verified it with the audit system test.

packages/cato_networks/data_stream/audit/_dev/test/pipeline/test-audit.log-expected.json
muskan-agarwal26: Those repeated `lan` entries are the verbatim `links` values from the vendor's per-port socket-interface config, preserved under the custom `cato_networks.audit.change.*` object. Each physical port configured as a LAN link contributes its own `lan` entry, so the duplication reflects the source record faithfully; we don't de-duplicate custom passthrough fields.

packages/cato_networks/data_stream/event/elasticsearch/ingest_pipeline/default.yml
ilyannn: You're right and I'll be the first to admit that I don't know much about CEL :smile:

ilyannn (approved):
LGTM

packages/cato_networks/data_stream/audit/agent/stream/cel.yml.hbs
efd6: Using an `as` binding is the correct approach here, but not for the reason of compilation. It's true that `has(try(timestamp(v)))` doesn't compile, but this is because it's in the wrong form, lacking a target field parameter. If that's added,
```
-- src.cel --
has(try(timestamp("i am not a timestamp"), "error").error)
-- out.json --
true
```
it does work.

The alternative is to use `is_error`,
```
-- src.cel --
is_error(timestamp("i am not a timestamp"))
-- out.json --
true
```
which could be used in place of `has` in the original suggestion.

However, this requires the timestamp to be parsed twice. The alternative (and slightly more concise) approach to the response is
```
-- src.cel --
timestamp("i am not a timestamp").as(t,
	!is_error(t) ?
		t.format("2006-01-02/15:04:05")
	:
		"alternative"
)
-- out.json --
"alternative"
```

jamiehynds:
@ilyannn can you please go ahead with the merge if all reviews are now complete? Thanks!

packages/cato_networks/data_stream/audit/agent/stream/cel.yml.hbs
efd6: Can we have a follow-up to fix this. Using the pattern will cause coding agents to see it as a good example.

PR #19320, opened 2026-06-01
[o365] Remove redundant filtering logic for non-cursor state

chrisberkhout:
## Proposed commit message

```
[o365] Remove redundant filtering logic for non-cursor state

#17540 added logic to filter out data in `state.cursor.todo_links` and
`state.cursor.todo_content` that isn't related to the currently
configured types. That's good. This PR adds an item for that in logic
overview comment at the top.

#17540 added similar filtering of `state.todo_types`, but that is
redundant because each config change or other restart will clear that
data already. This PR reverts that part and adds a comment to clarify.
```

## Checklist

## Related issues

- Relates #17540

### DIFF
--- a/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
+++ b/packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
@@ -92,6 +92,7 @@ redact:
 # - Rewrite legacy cursor data if present.
 # - Set a stopping point if absent.
 # - Generate the list of content types to process if empty.
+# - Filter stale cursor data.
 #
 # A check done on every evaluation:
 # - Ensure there is an active content type.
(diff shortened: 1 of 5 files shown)


packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -137,19 +138,16 @@ program: |-
...
-              (filtered.size() > 0) ?
-                filtered
-              :
-                configured_types
-            )
-          )
+          (existing_todo_types.size() > 0) ?
+            existing_todo_types
+          :
+            state.base.content_types.split(",").map(t, t.trim_space())
```
efd6: Do we want to make this more robust to user input like `type1, , type2` (extra commas, with spaces). The previous code does this incidentally.

chrisberkhout: Yes, that's good. Added.

packages/o365/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -137,19 +138,16 @@ program: |-
...
-              (filtered.size() > 0) ?
-                filtered
-              :
-                configured_types
-            )
-          )
+          (existing_todo_types.size() > 0) ?
+            existing_todo_types
+          :
+            state.base.content_types.split(",").map(t, t.trim_space()).filter(t, size(t) > 0)
```
efd6:
```suggestion
            state.base.content_types.split(",").map(t, t.trim_space() != "", t.trim_space())
```
rather than making two arrays, and using "" instead of size so that it's clear we're dealing with strings.

Also, will need to regenerate the policy tests.

chrisberkhout: Done.

efd6 (approved):

PR #19345, opened 2026-06-03
[openai] Fix Usage Overview dashboard panels erroring when only a subset of collections is enabled

### LINKED ISSUE #19340: [OpenAI]: Dashboard [Metrics OpenAI] OpenAI Usage Overview does not correctly calculate total tokens usage
### Integration Name

OpenAI [packages/openai]

### Dataset Name

openai.audio_speeches  openai.audio_transcriptions  openai.code_interpreter_sessions  openai.completions  openai.embeddings  openai.images  openai.moderations  openai.vector_stores

### Integration Version

2.0.0

### Agent Version

9.4.2

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.4.2

Debian GNU/Linux 12 (bookworm)

### Software/API Version

_No response_

### Error Message

Dashboard for token usage is showing an error for missing fields

### Event Original

_No response_

### What did you do?

I have enabled Completions only and not the other collections.

Image

### What did you see?

An error for tokens calculations

Since the collections are user driven and personally I have only enabled completions, it should disregard the lack of the other metrics and sum the metrics available

### Anything else?

_No response_

shmsr:
## Proposed commit message

The `[Metrics OpenAI] OpenAI Usage Overview` dashboard aggregates usage across several metric data streams:
- **Token panels** (Total tokens, Token usage by type, Token usage by model) sum fields from `completions` + `embeddings` + `moderations`.
- **Images generated** uses `images.images` / `images.size`.
- **Transcribed seconds** uses `audio_transcriptions.seconds`.
- **Characters synthesized** uses `audio_speeches.characters`.

OpenAI collections are individually toggleable. When a data stream is disabled it has no index, so its fields are absent from the `logs-*` data view, and Lens hard-errors the entire panel with a "missing field" error instead of showing the available data. The reporter enabled only Completions and saw this on the token panels; the same class of bug also affects the Images and Audio sections.

Declare the 11 dashboard-referenced usage fields in **every** metric data stream's `fields.yml` (each stream's own fields plus its siblings'), so `field_caps` always resolves them whenever any single collection is enabled. The sibling fields are mapped but only populated by their owning stream (documented inline).

Declaring a sibling data stream's field namespace inside another data stream's `fields.yml` is an established convention in this repo used by ~16 other packages. Concrete examples:

- **Kubernetes** — `data_stream/container/fields/base-fields.yml` declares `kubernetes.node.*` and `kubernetes.pod.*` (sibling data streams), so shared orchestrator dimensions resolve across metricsets.
- **Elasticsearch (Stack Monitoring)** — `data_stream/audit/fields/package-fields.yml` declares `elasticsearch.index.*`, `elasticsearch.node.*`, `elasticsearch.shard.*`.
- **Kafka** — `data_stream/broker/fields/package-fields.yml` declares `kafka.partition.*` and `kafka.topic.*`.

In most of those, the cross-declared fields are shared dimensions that are also populated; here they stay empty in the non-owning streams and exist purely so the shared data view resolves them — the same field-declaration mechanism, validated identically.

**Trade-off:** The embeddings/moderations/images/audio data streams' exported-field tables now list sibling usage fields they don't populate. This is cosmetic and invisible to dashboard users; it's the cost of keeping the change non-breaking. The alternative — unifying usage under `openai.base.*` — would rename fields, blanking pre-upgrade history in the panels and breaking custom queries/alerts.

**Non-breaking:** no field renames; the dashboard, ingest pipelines, and sample events are unchanged. Historical data keeps rendering and existing customer queries/alerts keep working. Bumped to `2.2.0` (enhancement).

## Checklist

## Author's Checklist

- [ ] Verify token/image/audio panels render without errors when only a subset of collections is enabled
- [ ] `elastic-package check` passes (lint + build + readme)

1. Enable only the `completions` data stream in the OpenAI integration and ingest some completions data.
2. Open the `[Metrics OpenAI] OpenAI Usage Overview` dashboard.
3. Confirm the token panels (Total tokens, Token usage by type, Token usage by model) render without "missing field" errors.
4. Optionally repeat with other single-stream configurations (images-only, audio-only) to verify the other panel groups.
5. Run `elastic-package check` to confirm lint + build + readme pass.

## Related issues

- Closes #19340

## Screenshots

Reproduced and verified on a live stack with **completions-only** data:

| | Before (`2.0.0`) | After (`2.2.0`) |
|---|---|---|
| Fields resolved by `field_caps` on `logs-*` | **5/11** (panels error) | **11/11** (panels render) |


packages/openai/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.2.0"
+  changes:
+    - description: Map usage fields across the metric data streams (completions, embeddings, moderations, images, audio_transcriptions and audio_speeches) so the OpenAI Usage Overview dashboard panels render correctly when only a subset of the usage collections is enabled.
+      type: enhancement
```
stefans-elastic, line 5: it says here `enhancement` but PR is labeled as `bug`

stefans-elastic: also if this ends up being changed to `bug` it would also required to change `version` - bump patch version instead of minor

stefans-elastic (approved):
approving but with a few minor comments

packages/openai/changelog.yml
shmsr: It's not actually a bug now (my bad that I labeled it like that), this is just how it currently works, since fields are kept on their own data streams. However, to support our dashboard feature where users can toggle subsets of data streams, treating this as an enhancement makes more sense

mykola-elastic (approved):

shmsr:
@ishleenk17 / @muthu-mps / etc could you please also take a look; need a code owner approval

lalit-satapathy (approved):
Code owner approval

PR #19347, opened 2026-06-03
[openai] Add rate limits data stream with headroom dashboard panel and alert

stefans-elastic:
## Proposed commit message

openai: add rate_limits stream and rate-limit headroom view

  This change adds an openai.rate_limits data stream and a rate-limit
  headroom view that compares OpenAI's configured per-project, per-model
  rate limits against actual Usage API consumption, so operators can see
  how close each project and model is to being throttled with HTTP 429
  responses. The stream provides the limit side of that comparison; the
  dashboard panels and alert join it to the existing usage streams.

   - The stream uses a CEL work-list: it pages the organization projects
     list, then drains each project's rate-limit endpoint, paginating both
     with after, last_id and has_more. To collect on every interval rather
     than every other one, the program never emits an empty event batch
     mid-chain (the filebeat CEL input ends a periodic run as soon as an
     evaluation yields zero events, before want_more is checked), so
     project-listing and the first rate-limit page are folded into a single
     emitting step. Each record is annotated with its project_id, name and
     status, is stamped with collected_at, and includes an optional limit
     field only when the API returns it, so sparse records never write null
     into the long-typed mappings. A failed call emits an error event and
     the run continues with the rest of the work list, and the admin token
     is redacted.

   - The OpenAI dashboard gains two ES|QL tables. "Rate limit headroom"
     joins limits to usage per project_id and model and reports peak
     one-minute used, limit and utilization for requests (RPM), tokens
     (TPM) and images (IPM). "Rate limit headroom - by model (org-wide)"
     rolls the same metrics up by model across projects; because OpenAI
     enforces limits per project, the summed per-project limit is a
     synthetic aggregate capacity and is labelled as indicative, not an
     exact throttle boundary. The model join normalizes dated model
     snapshots (e.g. omni-moderation-2024-09-26, gpt-image-1-2025-04-23) to
     Usage API reports per dated snapshot while rate_limits lists the base
     name; an exact-string join silently dropped those rows (e.g.
     gpt-image-1) from the headroom view. All six usage datasets
     (completions, embeddings, moderations, images, audio_speeches,
     audio_transcriptions) are added to the dashboard's global
     event.dataset filter so RPM is counted for audio and image models too.

   - A prebuilt "[OpenAI] Rate limit headroom low" .es-query rule fires at
     or above 80% peak TPM utilization, grouped by project_id::model. Its
     saved-object id and filename use the same "low" wording as the
     rule name and dashboard panel.

   - Request and image usage are normalized into single
     openai.base.usage_tokens and openai.base.usage_images fields,
     populated by the relevant usage pipelines. The panels and alert
     reference these normalized fields rather than per-stream columns, so
     the ES|QL resolves even before any usage data exists and keeps working
     when only some usage streams have data.

   - Each usage stream gains a finalization_grace setting (default 0s).
     OpenAI does not finalize a per-minute usage bucket when it ends; its
     counts keep rising for minutes afterward. At 0s (the default) every
     readable bucket is ingested immediately, favouring low latency: the
     most recently written minute may carry a lower count than the API
     eventually reports. Operators who need finalized counts can raise the
     grace period (e.g. 15m); the stream then holds back any bucket whose
     end time is younger than the grace period and re-fetches it on a
     later poll. The grace value is persisted in CEL state across polls.
     A residual limitation remains at any setting: OpenAI can revise
     buckets upward beyond any fixed grace window, and each bucket is
     ingested once, so a small undercount can persist for the busiest
     minutes. This does not affect the alert's ability to flag over-limit
     conditions.

  Utilization is computed with TO_DOUBLE because the used and limit values
  are longs; plain division truncated to zero until usage exceeded the
  limit, which also left the alert threshold ineffective. Headroom rows are
  ordered so that models with usage appear first, ranked by highest TPM
  utilization, with unused models sorted to the bottom.

  The peak one-minute calculation depends on the usage streams running with
  bucket_width set to 1m, which is the default. The package is bumped to
  2.2.0 and format_version to 3.5.7 for alerting_rule_template support.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


stefans-elastic:
## Test Results

### MOCK path (system test)

| Test scenario | Expected result | Actual result | Notes |
|---|---|---|---|
| `elastic-package test system -d rate_limits` runs to completion | Exit 0, PASS | ✅ PASS in 40.7 s | Validated against mock service (`proj_test_a/b`) |
| Mock doc for `gpt-4o-mini-2024-07-18` has correct field values | Fields match `mock_t_api.json` exactly | ✅ Exact match | `max_tokens_per_1_minute`, `max_requests_per_1_minute` etc. verified |
| Optional fields absent in API response are absent in ES (never `0`/null) | Fields missing from API must be absent in ES doc | ✅ Absent as expected | `max_images_per_1_minute` absent for text models |

---

### LIVE limits reconciliation (read-only)

| Test scenario | Expected result | Actual result | Notes |
|---|---|---|---|
| All live `rate_limits` docs are stable (MIN = MAX per project/model) | 0 value mismatches | ✅ 411/414 rows stable | 3 rows have MIN ≠ MAX: legitimate `gpt-realtime-2` TPM upgrade (40k → 250k) mid-window |
| `rate_limits` polls at configured interval | New docs every 5 minutes | ✅ 414 docs in 36 s chain at 14:08Z, then 14:14Z, 14:19Z, 14:24Z | Previously every other interval (10 min); now correct 5-min cadence |

---

### Data generation (all 6 usage streams, all 3 projects)

| Test scenario | Expected result | Actual result | Notes |
|---|---|---|---|
| **Default** project: all 6 streams produce requests | ≥1 doc per stream | ✅ All 6 streams populated | moderations: 1189 ok / 317 rate-limited; images: 4/5 ok (1× HTTP 503) |
| **xAi** project: all 6 streams produce requests | ≥1 doc per stream | ✅ All 6 streams populated | moderations: 1059 ok / 1424 rate-limited (project has lower RPM limit) |
| **dAi** project: all 6 streams produce requests | ≥1 doc per stream | ✅ All 6 streams populated | Initially rate-limited; retested after limits reset — 296 ok / 1056 rate-limited |
| moderations saturates RPM limit (>100% utilization) | >100% utilization visible in headroom panel | ✅ Default 186% RPM, xAi 176%, dAi ~59% RPM | dAi has higher RPM limit so saturation is less extreme |

---

### LIVE usage reconciliation (ES vs OpenAI usage API)

| Test scenario | Expected result | Actual result | Notes |
|---|---|---|---|
| **Default** — per-minute ES buckets equal usage API | ES = API per bucket, 0 mismatches | ✅ ES = API (part of 2183-request set, 20 buckets) | Finalization grace 15 min waited before checking |
| **xAi** — per-minute ES buckets equal usage API | ES = API per bucket, 0 mismatches | ✅ ES = API (part of 2183-request set, 20 buckets) | — |
| **dAi** — per-minute ES buckets equal usage API | ES = API per bucket, 0 mismatches | ✅ 10/10 buckets matched; ES = API = 322 requests | Burst window 14:54–14:59Z; all 5 streams reconciled |
| `gpt-image-1` usage joins via snapshot normalization | Non-zero `tpm_used` for gpt-image-1 in headroom panel | ✅ `tpm_used=864`, `tpm_limit=100000`, utilization shown | Bills as tokens under `completions`; `images` stream stays empty (expected) |
| Audio models (tts-1, whisper-1) show RPM utilization | RPM rows visible for audio models | ✅ Both models appear in per-project headroom panel |  |
| Utilization >100% renders correctly | Panel shows >100%, not clamped | ✅ omni-moderation 186.2% RPM Default, 176.9% TPM xAi | `TO_DOUBLE()` division prevents integer truncation |

---

## Overall Verdict

| Section | Result |
|---|---|
| MOCK system test | ✅ PASS |
| LIVE limits reconciliation | ✅ PASS |
| Data generation (all 6 streams × 3 projects) | ✅ PASS |
| Usage reconciliation (30 buckets, 3 projects) | ✅ PASS — ES = API = 2505 requests |

mykola-elastic (approved):

shmsr (commented):
Reviewed the rate_limits collector, the finalization-grace change across the 6 usage streams, the ES|QL panels + alert, and the ingest pipelines/fields/docs. Overall this is in great shape — the every-interval worklist design, finalization-grace bucket math, both-sided model normalization, stabilize-then-sum org rollup, and canonical `event.original` handling all check out and are well tested. The inline comments in the CEL programs are excellent and made review much easier.

One item was worth fixing before merge and is now pushed in 41baac6 (see the follow-up comment): the usage-stream error-return object dropped `cursor`, so a transient non-200 rewound the window to `initial_interval` on the next poll and re-emitted already-ingested buckets (no dedup in the usage pipeline → double count). Fixed by re-emitting `?"cursor": state.?cursor` in the error path of all 6 usage streams.

Everything below is **non-blocking** follow-up — none of it needs to hold the merge.

### CEL — clarifying comments on load-bearing logic

- **Multi-project-per-page cursor advance** — `data_stream/rate_limits/agent/stream/cel.yml.hbs` (L73–119). The LOAD step pages the project list but drains only `worklist[next]` (the *first* new project) before advancing `project_after = pbody.last_id` (L115) past the whole page. This is correct **only** because the next LOAD is gated on `next >= size(worklist)` (L50), so any other new projects from that page are drained by DRAIN steps before the list is paged again. If a future change made LOAD page the list more eagerly, those interior projects would be skipped silently. The state machine is already well documented (L27–33); a one-line note at L114–115 stating that invariant ("`project_after` advances a full page ahead of `next`; interior projects are drained from the worklist before the next page is requested") would protect it.

- **Ascending-bucket-order assumption** — `data_stream/completions/agent/stream/cel.yml.hbs` (L56) and the 5 sibling usage streams. `last_bucket_start` is recomputed per page, and in a `has_more` chain the *last page wins*; the held cursor is `min(start_time)` of still-finalizing buckets. That's correct only if the still-finalizing (most recent) buckets land on the final page — i.e. the API returns buckets ascending by time. OpenAI does, so it works, but the correctness hinges on it and it isn't asserted. Worth a one-line comment.

### System test coverage

- The config (`data_stream/rate_limits/_dev/test/system/test-default-config.yml`, `batch_size: 1`, `hit_count: 3`) exercises both project-list and per-project rate-limit pagination well, but the empty-page → keep-alive sentinel path (`{"skip": true}`, L113/L154/L225) — the mechanism the every-interval design hinges on — is only verified at the pipeline-test level (that the sentinel is dropped), not that the agent keeps the `want_more` chain alive when a project returns `data: []`. Consider adding a mock project that returns an empty rate-limit list so that path is covered end-to-end.

### ES|QL / alert

- The alert's `WHERE event.dataset IN ("openai.rate_limits", "openai.completions", "openai.embeddings", "openai.moderations")` deliberately excludes audio/images because it is TPM-only and only those three streams populate `openai.base.usage_tokens`. Correct as-is — a one-line comment noting the exclusion is intentional would stop a future editor from "fixing" it by adding audio/images and silently changing the denominator.
- The peak-per-minute math assumes `bucket_width: 1m`. The alert investigation guide documents this; consider adding the same one-line caveat to the panel 22/23 descriptions.

### Ingest pipelines

- **`usage_tokens` style divergence is justified, not a bug.** completions (`default.yml:54–65`) uses a painless `script` because it sums four sources (`input_tokens + output_tokens + input_audio_tokens + output_audio_tokens`); embeddings/moderations (`:54–57`) use `set … copy_from` because each has a single token source. Both land in `openai.base.usage_tokens`. A one-line comment on the single-source streams ("single token source, no summation needed") would prevent a "this one's missing the sum" misread. (Also confirmed completions correctly does **not** add `input_cached_tokens` — already counted inside `input_tokens`, so no double-count.)
- `data_stream/rate_limits/elasticsearch/ingest_pipeline/default.yml` (L44–51): the `collected_at` → `@timestamp` date processor uses `ignore_failure: true`. Risk is near-zero since `collected_at` is CEL-generated `now.format(RFC3339)`, but a malformed value would silently fall back to ingest-time `@timestamp` with no error tag. Dropping `ignore_failure` (so the pipeline's `on_failure` tags it) would make it observable if it ever happened.

shmsr:
Pushed 41baac6 with the `cursor`-on-error fix I flagged above.

**What it fixes:** the six usage streams' error-return object re-emitted `access_token`/`finalization_grace`/`initial_interval`/`initial_start_time` but not `cursor`. Since the CEL input replaces persisted state with the returned map, a single non-200 dropped `cursor`; the next poll's fresh-cycle reset then recomputed `initial_start_time` from the now-absent cursor and fell back to `now - initial_interval`, rewinding the window and re-emitting already-finalized buckets. With no dedup in the usage pipeline that double-counts — undercutting the accuracy the finalization-grace work is meant to protect.

**Change:** re-emit `?"cursor": state.?cursor` in the error path of all six usage streams. The optional-entry operator preserves `cursor` whenever it exists and omits it only on a first-run error before any cursor was written, where the `initial_interval` fallback is the correct behaviour.

packages/openai/docs/README.md
```diff
@@ -88,21 +91,78 @@ Example: For 100 API calls to a particular model per hour:
 
 > For optimal results with historical data, use 1-day bucket widths for long periods (15+ days), 1-hour for medium periods (1-15 days), and 1-minute only for the most recent 24 hours of data.
 
+### Finalization grace period
+
+OpenAI's Usage API does not finalize a per-minute bucket the moment it ends — a bucket's token and request counts keep climbing for several minutes afterward as late usage is accounted for. To avoid ingesting these partial, still-changing counts, each usage data stream waits a configurable **finalization grace period** before treating a bucket as final.
```
shmsr: This says each usage data stream waits for `finalization_grace`, and the bullets below say the latency applies to every usage stream. But `code_interpreter_sessions` and `vector_stores` still use the older one-bucket skip logic, have no `finalization_grace` setting, and still drop `cursor` on non-200 responses. Either apply the same finalization/cursor-preservation behavior to those two streams as well, or scope this docs section to the six usage streams participating in headroom.

packages/openai/data_stream/completions/agent/stream/cel.yml.hbs
```diff
@@ -48,17 +49,18 @@ program: |
       }).do_request().as(resp, resp.StatusCode == 200 ?
           bytes(resp.Body).decode_json().as(body,
             {
-              "events": body.data.filter(bucket,
-                    !(body.has_more == false && bucket.start_time >= body.data.map(bucket, bucket.start_time).max()))
+              "events": body.data.filter(bucket, bucket.end_time <= int(timestamp(now - duration(state.finalization_grace))))
```
shmsr, line 59: This can emit zero events while `body.has_more` is true if the returned page contains only buckets still inside `finalization_grace`. The new `rate_limits` stream has a dropped keep-alive sentinel because Filebeat CEL ends the cycle on zero emitted events before honoring `want_more`; the same stall can happen here and in the five sibling usage streams. Consider mirroring the sentinel/drop pattern when `events` would be empty but pagination needs to continue.

packages/openai/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "2.2.0"
+  changes:
+    - description: Add rate_limits data stream, a rate limit headroom dashboard panel, and a prebuilt rate limit headroom alert.
```
shmsr, line 4: The changelog only mentions the new `rate_limits` stream/dashboard/alert. This PR also changes existing usage-stream behavior by adding a default `finalization_grace: 15m`, which delays when usage data appears and affects headroom panels/alert timing. Please call that out in the release notes so upgrading users are not surprised by the new latency trade-off.

shmsr (commented):
Reviewed the current head again, including the existing PR comments. The earlier CI/review items look resolved or non-blocking; I left the actionable items that still look worth addressing before merge.

packages/openai/docs/README.md
stefans-elastic: Scoped this section to the six token/request usage streams that have a grace setting — `completions`, `embeddings`, `moderations`, `images`, `audio_speeches`, `audio_transcriptions` — and named `code_interpreter_sessions`/`vector_stores` as the streams that report sessions/bytes, do not feed headroom, and have no grace. Updated both `docs/README.md` and the `_dev/build/docs/README.md` template. Also applied the cursor-on-error preservation (`?"cursor": state.?cursor`) to those two streams so a transient non-200 no longer rewinds their window. Done in c7bad7af5a.

packages/openai/data_stream/completions/agent/stream/cel.yml.hbs
stefans-elastic: Added the keep-alive sentinel to all six usage CEL programs and a matching `drop_finalization_sentinel` processor to their ingest pipelines, mirroring the `rate_limits` pattern: when `events` filters to empty but `body.has_more` is true, emit `[{"message": {"skip": true}.encode_json()}]` instead of zero events. This is only reachable when the grace spans more than one page of buckets (~2h at the 1m default page size of 60), never at the 15m default, but the chain is now robust against large grace / misconfig. Done in c7bad7af5a.

packages/openai/changelog.yml
stefans-elastic: Added a 2.2.0 changelog entry calling out the new default `finalization_grace: 15m` on the usage streams and the ~15m latency it adds to usage data, the headroom panels, and the alert, with a pointer to lower it (or set `0s`) under Advanced options. Done in c7bad7af5a.

shmsr (approved):

PR #19391, opened 2026-06-05
Expose http timeout setting in Kubernetes package for kube state metrics

MichaelKatsoulis:
## Proposed commit message

In large kubernetes cluster, fetching metrics from kube state metrics can exceed the default http timeout of 10s.
This error leads to leader nodes failures that are unrecoverable. 
With this PR we expose the timeout option for KSM datasets that controls the http tiemout.
This option is availble in beats and standalone agent. 

## Checklist

## Screenshots
Screenshot 2026-06-05 at 12 02 42 PM

### DIFF
--- a/packages/kubernetes/manifest.yml
+++ b/packages/kubernetes/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.1.2
 name: kubernetes
 title: Kubernetes
-version: 1.83.0
+version: 1.83.1
 description: Collect logs and metrics from Kubernetes clusters with Elastic Agent.
 type: integration
 categories:
@@ -102,6 +102,14 @@ policy_templates:
             multi: false
             required: false
             show_user: false
+          - name: timeout
+            type: text
+            title: Timeout
+            description: HTTP request timeout. Raise on large clusters if scrapes exceed 10s and you see `context deadline exceeded` errors.
+            multi: false
+            required: false
+            show_user: false
+            default: 10s
     icons:
       - src: /img/logo_kubernetes.svg
         title: Logo Kubernetes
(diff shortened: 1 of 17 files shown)


packages/kubernetes/manifest.yml
```diff
@@ -102,6 +102,14 @@ policy_templates:
             multi: false
             required: false
             show_user: false
+          - name: timeout
+            type: text
+            title: Timeout
+            description: HTTP request timeout. Raise on large clusters if scrapes exceed 10s and you see `context deadline exceeded` errors.
```
gizas: `Valid time units are ns, us, ms, s, m, h.`

This is what we have in other timeouts in the integration

This points to this https://github.com/elastic/beats/blob/5f41561f849f0c899682cb5909f1a96dac43b59b/metricbeat/mb/mb.go#L349 beats which validates there.
In case of `s` is missing it treats this as seconds.

packages/kubernetes/manifest.yml
```diff
@@ -102,6 +102,14 @@ policy_templates:
...
+            multi: false
+            required: false
+            show_user: false
+            default: 10s
```
gizas, line 112: qq: why in the screenshot shows 30s?
Did u change the value? just to verify that we dont miss sth

gizas (approved):

packages/kubernetes/manifest.yml
MichaelKatsoulis: yes!

packages/kubernetes/manifest.yml
MichaelKatsoulis: Done!

PR #19401, opened 2026-06-05
microsoft_exchange_online_message_trace: fix pagination with single-quote-escaped skiptoken

kcreddy:
## Proposed commit message
```
microsoft_exchange_online_message_trace: fix pagination with single-quote-escaped skiptoken

The Microsoft Graph beta messageTraces endpoint wraps recipient
addresses containing apostrophes in single quotes with doubled
internal apostrophes (PowerShell string-literal escaping) inside
the base64-encoded skiptoken JSON. Its own backend then rejects the
token with "Invalid starting recipient address", causing an
infinite 400-error retry loop.

Extend the existing skiptoken sanitisation (which already handles
double-JSON-encoded values) to also detect and clean single-quote-
wrapped values: strip the outer quotes and replace '' with '.

The system test adds a fourth pagination page exercising this case.
```

## Checklist


packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
```diff
@@ -58,6 +59,8 @@ program: |-
                           {
                             k: (v.has_prefix('"') && v.has_suffix('"')) ?
                               v.decode_json().as(dv, is_error(dv) ? v : dv)
+                            : (v.has_prefix("'") && v.has_suffix("'")) ?
```
efd6: It's not like I don't trust them, but can we also check that the length is at least 2?

efd6 (approved):
nit, then LGTM

packages/microsoft_exchange_online_message_trace/data_stream/log/agent/stream/cel.yml.hbs
kcreddy: Addressed in [051250d](https://github.com/elastic/integrations/pull/19401/commits/051250d8a2d232aaba079779b23db7826121e224)

efd6 (approved):

PR #19405, opened 2026-06-05
[aws] Declare required permissions with provider_permissions

jeniawhite:
> **Status: Draft, blocked on #19828.** `provider_permissions` requires package-spec `format_version` 3.6.4, which per [format_version.md](https://github.com/elastic/elastic-package/blob/main/docs/howto/format_version.md) requires Kibana `^9.4.0+` (see [Andrew's comment](https://github.com/elastic/integrations/pull/19405#discussion_r3439845185)). Rather than bump the major version independently, this branch depends on #19828 (which does that major bump for an unrelated var_groups feature) — its commit is cherry-picked in temporarily so this PR lints/builds standalone. **The diff currently includes #19828's changes as a result.** Once #19828 merges, this branch will be rebased onto `main` and that content will drop out of the diff automatically.

## Proposed commit message

Adopt the new `provider_permissions` field (package-spec, `format_version` 3.6.4) so the AWS integration declares the IAM permissions it actually needs in a machine-readable way, instead of leaving them in prose docs and expecting users to attach a broad managed policy (e.g. `SecurityAudit`/`ReadOnlyAccess`).

### `aws` integration

- Bump `format_version` 3.4.0 → 3.6.4 and package `version` to `7.1.0` (on top of #19828's `7.0.0` major bump, which raises the Kibana/Agent requirement to `^9.4.0` — required for `format_version` 3.6.4); add a `changelog.yml` entry.
- Declare permissions across three layers, placed so that the effective (accumulated-and-deduplicated) permission set for each enabled data stream is minimal and contains no cross-layer duplication:
- **package** — `sts:GetCallerIdentity` only; the single call every credentialed AWS collector makes.
- **input** (50 inputs) — the permissions tied to each collection mechanism: `aws/metrics` → `ec2:DescribeRegions`, `cloudwatch:ListMetrics`, `cloudwatch:GetMetricData`, `tag:GetResources`; `aws-s3` → `s3:GetObject`/ `s3:ListBucket` + SQS notification actions; `aws-cloudwatch` → `logs:DescribeLogGroups`/`logs:FilterLogEvents`; `cel`/`httpjson` → the AWS Config / Inspector / GuardDuty API read calls.
- **data_stream** (26 streams) — every service/resource permission a stream needs for metric enrichment or API collection (e.g. `ec2:DescribeInstances` on `ec2_metrics`, `rds:Describe*` on `rds`, and the disjoint Security Hub `GetFindings` vs `GetInsights`/`ListInsights`/`GetInsightResults` reads).
- **policy_template** — intentionally unused: no permission is common to all of a template's data streams without over-granting a sibling, so placing anything here would either duplicate a lower layer or grant a stream more than it needs.

---

Permission requirements currently exist only in documentation, so they drift over time and cannot be consumed by tooling. Declaring them with `provider_permissions` versions them alongside the package and lets tooling generate appropriately scoped infrastructure-as-code rather than a static kitchen-sink policy.

The placement follows the spec's accumulate-and-deduplicate semantics together with least privilege: each permission is declared exactly once, at the broadest layer whose scope matches everything that needs it. This keeps minimal grants when only some data streams are enabled. Concretely:
- a logs-only deployment never receives `cloudwatch:*` (those live on the `aws/metrics` input);
- GuardDuty collected via S3/SQS receives only S3/SQS transport permissions - not the GuardDuty API reads, which sit on the `httpjson` input rather than the `guardduty` data stream;

No data streams, ingest pipelines, dashboards, or collection behavior change; this is metadata only.

## Checklist

- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-command).

## Author's Checklist

- [ ] `provider_permissions` are placed at the correct layer (package / input / data_stream) with no cross-layer duplication.
- [ ] IAM action names and the per-action descriptions are accurate for each service.
- [ ] Effective (aggregated) permission set per data stream is least-privilege.

## Related issues

- Depends on #19828 — do not merge before it; see status note at the top.
- Requires elastic/package-spec#1180
- Relates elastic/security-team#16737
- Relates elastic/ingest-dev#8142

### DIFF
--- a/packages/aws/manifest.yml
+++ b/packages/aws/manifest.yml
@@ -1,7 +1,7 @@
-format_version: 3.6.0
+format_version: 3.6.4
 name: aws
 title: AWS
-version: 7.0.0
+version: 7.1.0
 description: Collect logs and metrics from Amazon Web Services (AWS) with Elastic Agent.
 type: integration
 categories:
@@ -25,6 +25,34 @@ conditions:
   # auth.aws support in CEL and HTTPJSON inputs requires Elastic Agent 9.4.0+.
   agent:
     version: "^9.4.0"
+provider_permissions:
+  - provider: aws
+    description: >-
+      Baseline permissions required by every AWS data stream. Metric data streams
+      read CloudWatch metrics, and all data streams resolve the account identity,
+      enumerate regions, and collect resource tags for enrichment.
+    roles:
+      - name: SecurityAudit
+        id: arn:aws:iam::aws:policy/SecurityAudit
+        description: >-
+          AWS managed policy granting read-only access to configuration and
+          metadata across services. Covers the describe/list calls used for
+          resource discovery and tag enrichment across the integration.
+    permissions:
+      - name: sts:GetCallerIdentity
+        description: Resolve the AWS account ID associated with the configured credentials.
+      - name: iam:ListAccountAliases
+        description: Resolve the human-readable account alias used to enrich events.
+      - name: ec2:DescribeRegions
+        description: Discover the set of regions to collect from when none are configured.
+      - name: tag:GetResources
+        description: Collect resource tags (resourcegroupstaggingapi) for metric and log enrichment.
+      - name: cloudwatch:ListMetrics
+        description: Enumerate the CloudWatch metrics published by the monitored services.
+      - name: cloudwatch:GetMetricData
+        description: Read CloudWatch metric values for all metric data streams.
+      - name: cloudwatch:GetMetricStatistics
+        description: Read aggregated CloudWatch metric statistics for metric data streams.
 screenshots:
   - src: /img/metricbeat-aws-overview.png
     title: metricbeat aws overview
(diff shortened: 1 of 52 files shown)


packages/aws/manifest.yml
```diff
@@ -22,6 +22,12 @@ conditions:
     subscription: basic
   kibana:
     version: "^8.19.4 || ^9.2.1"
```
andrewkroh: With the package-spec move 3.4->3.6 the package becomes incompatible with 8.19 as per [format_version.md](https://github.com/elastic/elastic-package/blob/main/docs/howto/format_version.md). I think we should bump the major version of the package so that we can continue maintenance of the 6.x version of the package be able to support format_version 3.4 and Elastic Stack 8.19 on a branch until Jan 2027 (~6 months). Or merge this after #19828 which does the major bump.

andrewkroh (commented):
The permission sets look good to me, but I only did a few spot checks on the ones I know offhand. It would be good if the SMEs for the individual inputs used could take a closer look.

packages/aws/manifest.yml
jeniawhite: Thanks for catching this. Went with the second option: this PR now depends on #19828 rather than bumping the major version independently.

Rebased on top of #19828's commit (cherry-picked onto current `main` for now, since that PR is still open) — package version is `7.1.0`, `format_version` 3.6.4, `kibana`/`agent` `^9.4.0`, consistent with the compatibility table in [format_version.md](https://github.com/elastic/elastic-package/blob/main/docs/howto/format_version.md).

Marked this PR as Draft and blocked on #19828 in the description. The diff currently includes #19828's changes as a side effect of the cherry-pick — once that PR merges, I'll rebase this branch onto `main` and that content will drop out of the diff on its own.

PR #19408, opened 2026-06-05
[new integration] Backstage audit & application logs

Niceplace:
## What

Adds a new **Backstage** integration package that collects application and audit logs, with a focus on audit events emitted by Backstage's built-in `auditor` service (available since `@backstage/backend-plugin-api` 1.2.0). Audit events cover catalog operations, user/plugin activity, and severity-classified events, and normalizes them to ECS.

Fixes: https://github.com/elastic/enhancements/issues/28055

## Why

Backstage's audit log is the primary signal for tracking who did what on a developer platform (catalog access, plugin actions, admin operations). This integration lets teams ingest that data into the Elastic Stack for security monitoring, compliance, and investigating plugin behavior, without a custom Filebeat/Logstash pipeline.

## What's included

- **Package scaffolding**: `backstage` integration package (`manifest.yml`, icon, docs, changelog) targeting Kibana `^8.19.0 || ^9.1.0`, categorized as observability/security/web.
- **`audit_logs` data stream**: filestream input that tails Backstage's JSON audit log lines (one JSON object per `<plugin>.<eventId>` message), with configurable paths, exclude patterns, and optional multiline/ndjson parsers for pretty-printed JSON.
- **Ingest pipeline**: parses each line, drops non-audit lines (missing `isAuditEvent`), extracts user agent, maps event outcome/severity/action to ECS (`event.action`, `event.provider`, `event.severity`, `event.outcome`, etc.), and preserves `backstage.*` custom fields for plugin-specific metadata (e.g. catalog entity-fetch filters).
- **Field definitions**: `fields/backstage.yml`, plus standard `ecs.yml`/`agent.yml`/`beats.yml`/`base-fields.yml`.
- **Kibana dashboard**: a basic dashboard surfacing audit activity by user, plugin, and outcome (locally tested).
- **Tests**: pipeline test fixture + expected output (anonymized), common pipeline test config, and a system test config with a sample Docker Compose deployment and sample log file.
- **Docs**: package `README.md` covering overview, compatibility, data collected, setup/validation steps, and troubleshooting (multiline JSON, dropped non-audit lines).

## How to test

1. `elastic-package test pipeline` in `packages/backstage/data_stream/audit_logs` — validates the ingest pipeline against the anonymized fixture.
2. `elastic-package stack up` + `elastic-package test system` — spins up the sample Docker Compose deployment and verifies events land with `data_stream.dataset: "backstage.audit_logs"` and a populated `event.action`.
3. In Kibana Discover, confirm parsed fields (`event.provider`, `event.severity`, `backstage.*`) and check the bundled dashboard renders.

## Related issues

- Closes elastic/enhancements#28055

Note: As of 2026-07-21 I've successfully tested this integration with over 2 175 000 log entries pulled from our production instance of backstage.  Around 50 documents had some data quality issues related to  max length of the `url` field which was increased in https://github.com/elastic/integrations/pull/19408/changes/d2b339349972d0ec8380bc3f379a81e0dd774877.
This puts us in a good position quality-wise for the tech preview release.


packages/backstage/data_stream/logs/elasticsearch/ingest_pipeline/audit.yml
```diff
@@ -0,0 +1,349 @@
...
+      ignore_missing: true
+      tag: foreach_meta_query_filter
+      description: >-
+        Iterates over each filter string in the Backstage catalog query and splits `key=value` pairs
+        into structured fields under `backstage.query.filter`. Backstage catalog filters follow the
+        format `kind=Component,metadata.name=foo`.
+      processor:
+        kv:
+          field: _ingest._value
+          target_field: backstage.query.filter
```
P1llus: I might be missing something here, but if this is a list, you are looping over that list and overwriting the backstage.query.filter field for each item in the list, its not appending to it

packages/backstage/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,349 @@
...
+  - remove:
+      field: _tmp
+      ignore_missing: true
+      tag: remove_empty_tmp
+      description: Removes the empty temporary pipeline container.
+      if: ctx._tmp != null && ctx._tmp.isEmpty()
+  - remove:
+      field: json
+      ignore_missing: true
+      tag: remove_empty_json
```
P1llus: I think there is 2 minor things here:
1. You do not need to remove all children of an object before removing the parent. A single "remove json" which is what you have here at the end would be enough
2. A remove processor takes a list of fields in a single processor, you do not need to define one for each of them.

As a small tip, the easiest way to go about this, and you most likely will see it in most security integrations, is that we have a single remove processor near the end that handles all the cleanup. So it would remove things like "json" and "_tmp", and the only times you actually need to add a remove processor in the middle of the pipeline is when you want to do some sort of logic, for example when you need to remove something so other processors can add a value to it. But in almost all cases you tend to just have 1 remove processor at the end.

packages/backstage/data_stream/logs/fields/fields.yml
```diff
@@ -0,0 +1,23 @@
+- name: backstage
+  type: group
+  fields:
+    - name: severity_level
+      type: keyword
+      description: The severity level of the audit log registered by Backstage.
+    - name: meta
```
P1llus, line 19: Not exactly a wrong field type, just make sure that you understand the concept of `flattened` and what it does.
It is useful to be able to define a top field (like backstage.meta here) where all fields under it is either totally unknown, or of dynamic values (like a key being a hostname of a device).

It does come with some restrictions though, in terms of querying, aggregations and so on, so just making sure that this field is more the "catch everything we didn't map to a ECS field OR a `backstage.*` field).

All I am saying is that only store fields under backstage.meta that you do not know exist, or is unique to every user of the product, do not store fields under here that you know about, those should be renamed into proper `backstage.*` fields

packages/backstage/data_stream/logs/fields/backstage.yml
```diff
@@ -0,0 +1,23 @@
...
+      type: group
+      fields:
+        - name: fields
+          type: keyword
+          description: The entity fields requested by the catalog query.
+        - name: type
+          type: keyword
+          description: The type of catalog query that was issued (e.g. "all").
+        - name: filter
+          type: flattened
```
P1llus: Another one here, are you sure we need it? Can it not just be a string?

packages/backstage/data_stream/logs/fields/ecs.yml
```diff
@@ -0,0 +1,38 @@
+- name: event.kind
+  external: ecs
+- name: event.category
+  external: ecs
+- name: event.type
```
P1llus, line 7: Often the system/pipeline tests will only flag ECS fields missing from the list if they are in sample data, did you check the ingest pipeline and the ECS fields you use as destinations for your processors?

There are also processors that populates multiple destination fields (like `user_agent` and `uri_parts`)

packages/backstage/data_stream/audit_logs/manifest.yml
```diff
@@ -0,0 +1,125 @@
...
+          #file_identity.path: ''
+          #file_identity.inode_marker.path: ''
+          #file_identity.fingerprint: ~
+          #rotation.external.strategy.copytruncate.suffix_regex: '\.\d$'
+          #rotation.external.strategy.copytruncate.dateformat: '-20060102'
+          #include_lines: ['sometext']
+          #exclude_lines: ['^DBG']
+          #buffer_size: 16384
+          #message_max_bytes: 1048576
+elasticsearch:
```
P1llus: This part can go away, I believe its something that the default elastic-package template adds which is not really necessary anymore and causes confusion

packages/backstage/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "0.0.1"
```
P1llus: We usually start on 0.1.0 unless its a very experimental input of some sort, and filestream is quite a stable ingestion point

packages/backstage/LICENSE.txt
```diff
@@ -0,0 +1,93 @@
+Elastic License 2.0
```
P1llus: This file can be removed, its also one of those added by elastic-package. Its injected automatically on build.

packages/backstage/manifest.yml
```diff
@@ -0,0 +1,31 @@
+format_version: 3.6.2
```
P1llus: Unfortunately as long as we are going to support `version: "^8.19.0 || ^9.1.0"` there is a maximum format version supported of `3.4.2` I believe, might also be related to why you are getting an issue in build. I would recommend re-generating your system test with a new sample after that is changed, as I believe it causes some problems with geo fields and so.

packages/backstage/data_stream/audit_logs/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,14 @@
...
+  vars:
+    preserve_original_event: true
+    paths:
+      - "{{SERVICE_LOGS_DIR}}/*.log"
+    custom: |
+      prospector.scanner.fingerprint.enabled: false
+      prospector.scanner.fingerprint.offset: 0
+      prospector.scanner.fingerprint.length: 10
+agent:
+  user: root
```
P1llus: This should not be needed at all, this might be the result of some local issue? Try to take a look at any other filestream system test config and while some have specific settings needed for them to work, you do not have these in your .yml.hbs for the production configuration so I assume its related to some failing test locally?

packages/backstage/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,349 @@
...
+      description: The `message` field is no longer required if the document has an `event.original` field.
+      if: ctx.event?.original != null
+  - json:
+      field: event.original
+      target_field: json
+      tag: parse_event_original_json
+      description: Parses the JSON-encoded `event.original` into the temporary `json` object for field extraction.
+      if: ctx.event?.original != null
+  - drop:
+      if: ctx.json?.isAuditEvent == null
```
P1llus: Is this because there are non-audit events in the logs as well? Its a bit confusing.

P1llus (commented):
Just did a quick browse through the changes, nothing big stands out here and wanted to also take the opportunity to help out with some pointers. Happy to help or discuss any of them further 👍

packages/backstage/LICENSE.txt
Niceplace: Noted ! Will remove, thanks for flagging :D !

packages/backstage/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
Niceplace: Correct ! Audit event logs contain a `isAuditEvent: true` but "normal" logs don't event have the key at all😅 Do you think this warrants a clearer description ?

packages/backstage/data_stream/logs/elasticsearch/ingest_pipeline/audit.yml
Niceplace: Uh-oh ! Good catch, will fix that

packages/backstage/data_stream/logs/fields/fields.yml
Niceplace: Yeah, I'm taking a bit of a gamble here, from the [docs](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/flattened), this was the main reason I thought using `flattened` was a good idea but part of me thinks I'm over optimizing this

> This data type can be useful for indexing objects with a large or unknown number of unique keys.

At the moment we don't know all the possibilities for the content of `meta` yet and if we end up getting scenarios that are unsupported, at least in the meantime the data is going to be available somewhere and we can inspect its shape.

> All I am saying is that only store fields under backstage.meta that you do not know exist, or is unique to every user of the product, do not store fields under here that you know about, those should be renamed into proper backstage.* fields

Am I correct in understanding that I should I remove successfully mapped fields from `backstage.meta` and only keep the rest in the flattened field ?

packages/backstage/data_stream/logs/fields/ecs.yml
Niceplace: I'm not 100% sure I understand your question here, are you saying I should double check and ensure all ECS fields being mapped to in the ingest pipeline should be documented here ? That's my understanding but I could be wrong.

From what I understand, [this build configuration](https://github.com/elastic/integrations/pull/19408/changes#diff-eef51f0d7f2268aee9fc3aaf5f2840635d181ea4625cfc5b42d89adf3df1bad4R3) should be enough to support ECS mappings from a functional aspect. However, having this file with `external: ecs` mentions ensures that the ECS fields being used in this integration will end up in the generated documentation. Is this your understanding as well or should I take a different approach ?

packages/backstage/data_stream/audit_logs/manifest.yml
Niceplace: From my understanding of reading about [subobjects](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/subobjects) I'm a bit confused about what is the best approach here. If I remove this block, this will switch back to the default value of `true` for `subobjects`.  

Given our input data is already formatted in JSON I'm not expecting any conflicts with nested keys. Let's discuss this one tomorrow :D ! I'm wondering if this warrants an update to elastic-package but I don't want to go on a tangent and end up too far.

packages/backstage/data_stream/audit_logs/_dev/test/system/test-default-config.yml
Niceplace: The initial assumption here was that the logs would be created with a `root` user. Since filebeat reads those via a bind-mount volume (that's the case for my local tests, could be different in K8S) it would encounter permission issues if it doesn't use `root` to read the logs.

After some investigation though, I'm a bit confused because our Backstage setup is a bit special and lives on a different cluster than where the logs will be analyzed.  In that sense, our installation of Backstage and the way we collect audit logs differs from how we expect our customers to use it. 

I have to investigate this a bit more because I'm not getting mixed signals if `root` is necessary or not here.

packages/backstage/manifest.yml
Niceplace: Can you help me understand the impacts of downgrading that version ? I'm not against it I just want to understand what to check after doing it. 

As for the build issues, it was flagged as a compatibility issue with the `append` processor that was not supported in `^8.19.0` and fixed by https://github.com/elastic/integrations/pull/19408/changes/5fa20a78be41d1e93821a80444a60e43afa6e177. I had a few other formatting/test/README incompatibility issues after that but the tests are now all ✅ 

Let's discuss this one tomorrow

packages/backstage/data_stream/audit_logs/manifest.yml
Niceplace: After discussing, the reason is that it's set as a default on the elasticsearch level and setting it here makes no sense.

packages/backstage/data_stream/audit_logs/_dev/test/system/test-default-config.yml
Niceplace: After discussion, fingerprint.enabled and fingerprint.offset are definitely not needed, let's re-validate fingerprint.length if it's really needed. Compare with existing integrations that use filebeat

packages/backstage/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
Niceplace: The concern here is that our customer might be interested in the rest of the logs, what happens then ?
We should approach this from a "all logs" perspective use `event.Action` to help us determine what type of  log it is (audit event v.s. normal log).

- Rename the audit log data stream to just `logs`
- Keep ignoring/dropping non-audit logs, easier to add support for these in the future with a second pipeline
- Because the data_stream name will change, check the dashboards

packages/backstage/data_stream/audit_logs/_dev/test/system/test-default-config.yml
Niceplace: After double checking, the `fingerprint.enabled: false` pattern is used in at least two integrations

- https://github.com/elastic/integrations/blob/96400ccea7056b462c1baad1c2cbffe7fb961bf4/packages/proxysg/data_stream/log/_dev/test/system/test-logfile-config.yml#L11
- https://github.com/elastic/integrations/blob/96400ccea7056b462c1baad1c2cbffe7fb961bf4/packages/syslog_router/data_stream/log/_dev/test/system/test-filestream-config.yml#L13 

And I confirmed that the system tests fail without this option enabled, so I think we have to keep it here.

packages/backstage/data_stream/logs/elasticsearch/ingest_pipeline/audit.yml
Niceplace: I opted for something simpler, a backstage vendor field of type `text` which handles multi value as well, getting rid of the overwrite problem. I figured since we don't have a real use case for this right now, let's scale back the approach and add further extraction based on user feedback later on.

packages/backstage/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
Niceplace: @P1llus Done ! Changed to a [single remove processor at the end](https://github.com/elastic/integrations/pull/19408/changes#diff-1eca0380f21fa35f3a45c2bb69b680dc58f2c315596607e6efca8bb94151740cR364). 

There are still two explicit [remove processors](https://github.com/elastic/integrations/pull/19408/changes#diff-1eca0380f21fa35f3a45c2bb69b680dc58f2c315596607e6efca8bb94151740cR343-R352) that cleanup before we serialize leftover data in `backstage.meta`. We covered the use cases we know so far but with future usage by our customers, it's inevitable that we will run into new fields in the future and they will land there in the meantime.

Let me know if you have some concerns with that approach

packages/backstage/data_stream/logs/fields/fields.yml
Niceplace: I ended up changing  `backstage.meta` from `flattened` to `text`.  If we have specific customer requests about contents within `meta` we can add specific processing later. 

After analyzing a large batch of accumulated log entries from our backstage's production instance, I also changed the handling so known `json.meta.*` keys are extracted into dedicated `backstage.*` fields.

We use `backstage.meta` as a fallback for residual metadata we do not explicitly model yet. So it is no longer acting as a catch-all for fields we already understand.

packages/backstage/data_stream/logs/fields/backstage.yml
Niceplace: Changed to [text](https://github.com/elastic/integrations/pull/19408/changes#diff-1351e4881a9f53c83f4539291ab663b4a2612f7a85196447e7adcf7cade3c93aR74)

packages/backstage/data_stream/audit_logs/elasticsearch/ingest_pipeline/default.yml
Niceplace: After our discussion, I re-worked the pipelines to have a [default](https://github.com/elastic/integrations/pull/19408/changes#diff-123288eed0e295ad961ef43ef73e50d2e298209c64d935c3644ce6cc61fe24e8) ingest pipeline that dispatches to an [audit](https://github.com/elastic/integrations/pull/19408/changes#diff-1eca0380f21fa35f3a45c2bb69b680dc58f2c315596607e6efca8bb94151740c) ingest pipeline when `isAuditEvent: true` is found. 

The structure now le aves a much cleaner path to support other Backstage log types later without having to redesign the stream.

packages/backstage/data_stream/logs/fields/ecs.yml
Niceplace: Note that we discussed this in person and the outcome was that explicit ecs field documentation is necessary. I also had some missing ECS fields populated indirectly by the `user_agent` processor that I added as well.

As for the `uri_parts` hack that I was doing to supply a mock `http://localhost` so we could parse just the path + query params, my approach was wrong, `uri_parts` can successfully deal with an incomplete URL, which makes for a leaner ingest pipeline.

packages/backstage/manifest.yml
Niceplace: As discussed in person the minimum version here should be `3.4.2` to ensure compatibility with older Kibana/Elasticsearch versions.

packages/backstage/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,49 @@
...
+      if: ctx.json == null && ctx.event?.original != null
+      description: >-
+        Falls back to parsing `event.original` when the agent already moved the raw
+        line there (e.g. via the `preserve_original_event` tag) instead of `message`.
+  - drop:
+      if: ctx.json?.isAuditEvent == null
+      tag: drop_if_not_audit_event
+      description: >-
+        Drops messages that do not contain the isAuditEvent: true key/value pair
+        Non-audit logs do not contain that key at all and are not currently supported
```
ilyannn: Are non-audit logs guaranteed to be in NDJSON format? If no, we need to drop the lines that fail the `json` processor above (right now they are passed on as pipeline errors).

integration-skills-link
```diff
@@ -0,0 +1 @@
+/Users/sbeaulieu/workspace/work/integration-skills
```
ilyannn: This symlink should be removed; FYI you can use local `.git/info/exclude` or machine-wide git exclude file to avoid including unwanted files in the future.

packages/backstage/manifest.yml
```diff
@@ -0,0 +1,31 @@
+format_version: 3.4.2
+name: backstage
+title: "Backstage"
+version: 0.1.0
+source:
+  license: "Elastic-2.0"
+description: "Parses and analyzes logs from spotify backstage for security analysis"
```
ilyannn: is "spotify backstage" the real terminology? README seemed much more clear

packages/backstage/data_stream/logs/manifest.yml
```diff
@@ -0,0 +1,121 @@
...
+    template_path: filestream.yml.hbs
+    vars:
+      - name: paths
+        type: text
+        title: Paths
+        multi: true
+        required: true
+        show_user: true
+        default:
+          - /var/log/*.log
```
ilyannn: Created by :robot: Cursor/GPT-5.6 Sol under my supervision.

The default `/var/log/*.log` scans every top-level host log even though this stream only supports Backstage audit JSON. This increases harvester work and feeds unrelated records into the router. Please use a Backstage-specific default while keeping **Paths** user-configurable.

```suggestion
          - /var/log/backstage/*.log
```

packages/backstage/data_stream/logs/manifest.yml
```diff
@@ -0,0 +1,121 @@
...
+        multi: false
+        default: false
+      - name: tags
+        type: text
+        title: Tags
+        multi: true
+        required: true
+        show_user: false
+        default:
+          - backstage
```
ilyannn: Created by :robot: Cursor/GPT-5.6 Sol under my supervision.

The hidden default tags omit `forwarded`, so the existing `publisher_pipeline.disable_host` condition never activates. The generated sample consequently identifies the Elastic Agent container in `host.*`, not the Backstage source, which is misleading for the documented mounted container/pod log model.

```suggestion
          - forwarded
          - backstage-logs
```

packages/backstage/data_stream/logs/elasticsearch/ingest_pipeline/audit.yml
```diff
@@ -0,0 +1,417 @@
...
+      tag: set_event_severity_0716ffd2
+      field: event.severity
+      value: 5
+      description: Sets `event.severity` to 5 when the audit log's severity level is `medium`.
+      if: 'ctx.backstage?.severity_level == "medium"'
+  - set:
+      tag: set_event_severity_f5f4e875
+      field: event.severity
+      value: 7
+      description: Sets `event.severity` to 7 when the audit log's severity level is `high`.
```
ilyannn, line 76: Created by :robot: Cursor/GPT-5.6 Sol under my supervision.

The pipeline fixtures cover `low` and `medium`, but none contains `severityLevel: high`, so this `event.severity: 7` branch can regress without failing CI. Please add a synthetic high-severity audit record and regenerate the expected pipeline output.

packages/backstage/data_stream/logs/sample_event.json
```diff
@@ -0,0 +1,128 @@
...
+        "version": "9.3.3"
+    },
+    "backstage": {
+        "query": {
+            "fields": [
+                "metadata",
+                "kind",
+                "spec.profile"
+            ],
+            "filter": "kind=group,relations.hasMember=user:default/guest",
```
ilyannn: Created by :robot: Cursor/GPT-5.6 Sol under my supervision.

This generated sample is stale relative to the current pipeline: `test-audit.log-expected.json` preserves `backstage.query.filter` as an array, while the sample and rendered docs still show a scalar. Please regenerate the system-test sample and rebuild the docs. The current shape should be:

```json
"filter": [
  "kind=group,relations.hasMember=user:default/guest"
]
```

ilyannn (commented):
Six medium-severity findings from the Backstage integration review.

packages/backstage/kibana/dashboard/backstage-7db6b702-707f-4cde-ad45-578916b3f4ae.json
```diff
@@ -0,0 +1,1676 @@
...
+                },
+                "gridData": {
+                    "h": 8,
+                    "i": "69f32df1-c1bd-4e9a-b56e-d1c9808450d0",
+                    "w": 12,
+                    "x": 0,
+                    "y": 0
+                },
+                "panelIndex": "69f32df1-c1bd-4e9a-b56e-d1c9808450d0",
+                "title": "Total Requests",
```
ilyannn: Created by :robot: Cursor/GPT-5.6 Sol under my supervision.

This panel counts raw audit documents, not requests. Backstage emits separate initiated and terminal records for one operation; the pipeline fixture's identical task pair demonstrates that a completed operation is counted twice here. Please either rename all request-count panels to **Audit Events** or scope operation-count panels to terminal records (`event.type: end`).

```suggestion
                "title": "Total Audit Events",
```

packages/backstage/kibana/dashboard/backstage-7db6b702-707f-4cde-ad45-578916b3f4ae.json
```diff
@@ -0,0 +1,1676 @@
...
+                    "order": 4,
+                    "type": "optionsListControl",
+                    "width": "medium"
+                }
+            }
+        },
+        "description": "Overview of Backstage audit log activity: request volume by event ID and plugin, actor activity, severity distribution, and failures. Backstage audit events carry no session identifier, so the actor-vs-event heatmaps approximate a user's journey rather than a true session replay. Outcome (event.outcome: success / failure / unknown) reflects the audited operation's result.",
+        "kibanaSavedObjectMeta": {
+            "searchSourceJSON": {
+                "filter": [],
```
ilyannn: Created by :robot: Cursor/GPT-5.6 Sol under my supervision.

The Lens panels are scoped individually, but the dashboard has no global dataset filter. All five controls use `logs-*`, so their option lists can include values from unrelated integrations. Add a global `data_stream.dataset: backstage.logs` phrase filter; the controls already have `ignoreFilters: false` and will inherit it.

```suggestion
                "filter": [
                    {
                        "$state": { "store": "appState" },
                        "meta": {
                            "field": "data_stream.dataset",
                            "key": "data_stream.dataset",
                            "negate": false,
                            "params": { "query": "backstage.logs" },
                            "type": "phrase"
                        },
                        "query": {
                            "match_phrase": {
                                "data_stream.dataset": "backstage.logs"
                            }
                        }
                    }
                ],
```

packages/backstage/manifest.yml
Niceplace: Good catch, will standardize against README

packages/backstage/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
Niceplace: > Are non-audit logs guaranteed to be in NDJSON format?

Guaranteed ? For our customers, not quite, but close. For us yes. Backstage's default logging configuration is NDJSON and the audit logs inherit the root logger so it would be surprising to see otherwise.

I would prefer to keep JSON parsing failures as pipeline errors. A spike in these failures would indicate an abnormal or unsupported logging configuration and should remain observable so operators can alert on it.

More importantly, when parsing fails, we cannot determine whether the line was an ordinary non-audit log or a malformed/truncated audit event. Dropping all parse failures could therefore silently discard audit-relevant data.

Given this, do you have any strong objection to keeping these as pipeline errors?

packages/backstage/data_stream/logs/elasticsearch/ingest_pipeline/audit.yml
Niceplace: Added a event log with high severity in the pipeline tests

packages/backstage/kibana/dashboard/backstage-7db6b702-707f-4cde-ad45-578916b3f4ae.json
Niceplace: Good catch ! Change the nomenclature from requests -> events

packages/backstage/kibana/dashboard/backstage-7db6b702-707f-4cde-ad45-578916b3f4ae.json
Niceplace: Nice ! I Added a global filter for backstage.logs.

packages/backstage/data_stream/logs/fields/fields.yml
```diff
@@ -0,0 +1,73 @@
...
+        - name: fields
+          type: keyword
+          description: The entity fields requested by the catalog query.
+        - name: type
+          type: keyword
+          description: The type of catalog query that was issued (e.g. "all").
+        - name: filter
+          type: text
+          normalize:
+            - array
```
Niceplace, line 80: @ilyannn Can you double check this one ? This is the alleged fix for https://github.com/elastic/integrations/pull/19408#discussion_r3590943906, it sent me down quite deep in a rabbit hole and I didn't expect elastic-package to be the root cause here.

The summary of the findings is. It seems legit but I'd like some more eyes on it in case I'm doing something wrong.

```
The ingest pipeline was correct—the array was never converted to a scalar during ingestion.

The scalar appeared only while `elastic-package` generated `sample_event.json`:

1. The pipeline emitted `backstage.query.filter` as an array.
2. Because the data stream uses synthetic source, `elastic-package` generated the sample from Elasticsearch’s `fields` response instead of `_source`.
3. Elasticsearch’s `fields` API returns every field as an array.
4. To reconstruct natural JSON, `elastic-package` collapses single-element arrays into scalars unless the field definition explicitly declares `normalize: [array]`.
5. `backstage.query.filter` lacked that metadata, so `["kind=group"]` became `"kind=group"` in the generated sample and documentation. Arrays with multiple values remained arrays, making the behavior especially confusing.

The fix was:

```yaml
- name: filter
  type: text
  normalize:
    - array
```

`normalize: [array]` tells `elastic-package` that the field must remain an array even when it contains only one value. It does not change the Elasticsearch mapping—Elasticsearch has no separate array field type. This behavior is intentional and remains present in `elastic-package` v0.125.1.

packages/backstage/data_stream/logs/elasticsearch/ingest_pipeline/default.yml
ilyannn: It sounds like they are expected to be in the NDJSON format and deviations are an error – sounds good to me.

packages/backstage/data_stream/logs/fields/fields.yml
ilyannn: I've checked the docs, this sounds good - also this is mentioned in `.agents/skills/review-integration/references/transform-guide.md`:

>  `normalize: [array]` for array output fields

ilyannn (approved):
LGTM

packages/backstage/data_stream/logs/_dev/test/pipeline/test-non-audit.log-expected.json
```diff
@@ -0,0 +1,131 @@
+{
+    "expected": [
+        {
+            "backstage": {
+                "log": {
+                    "level": "info",
+                    "message": "Listening on port 7007",
+                    "service": "backstage",
+                    "timestamp": "2026-05-22T20:29:50.000Z"
```
ilyannn: It looks like we should still use standard fields like `@timestamp`!

Niceplace: Yeah ! So it seems I need to do some minimal processing for non-audit logs to ensure ECS is properly mapped, I added a pipeline to handle common cases before dispatching to specific pipelines

packages/backstage/_dev/deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+version: "2.3"
+services:
+  backstage-audit:
+    image: alpine
+    hostname: backstage-audit
+    # environment:
```
vinit-chauhan: Let's remove this if not needed

packages/backstage/_dev/build/docs/README.md
```diff
@@ -0,0 +1,115 @@
+# Backstage Integration for Elastic
```
vinit-chauhan, line 3:
```suggestion
{{- generatedHeader }}

# Backstage Integration for Elastic
```

packages/backstage/data_stream/logs/fields/fields.yml
vinit-chauhan, line 1: nit: This file name is valid as per spec. But most of the integrations if not all, use file name `fields.yml` for the package field mappings.

packages/backstage/data_stream/logs/elasticsearch/ingest_pipeline/common.yml
```diff
@@ -0,0 +1,175 @@
...
+      description: Maps the Backstage trace identifier to ECS trace.id.
+  - rename:
+      field: json.span_id
+      target_field: span.id
+      ignore_missing: true
+      tag: rename_common_span_id
+      description: Maps the Backstage span identifier to ECS span.id.
+  - rename:
+      field: json.request.url
+      target_field: _backstage.common.url
```
vinit-chauhan: nit: my personal preference would be to store temporary data in `_tmp` to make it obvious and be uniform with other integrations.

vinit-chauhan (approved):
This looks great! I've left a couple of minor, optional suggestions. Feel free to address them if you like, but it's entirely up to you.

packages/backstage/data_stream/logs/elasticsearch/ingest_pipeline/common.yml
Niceplace: @vinit-chauhan Just to make sure I understand, is your comment strictly scoped to the request URL parsing, or _all_ usage of `_backstage` as a temporary store until it's re-mapped somewhere else later in the pipeline ?

packages/backstage/data_stream/logs/fields/fields.yml
Niceplace: I prefer consistency :D Will update

packages/backstage/data_stream/logs/elasticsearch/ingest_pipeline/common.yml
vinit-chauhan: I meant all `_backstage.*` -> `_tmp.*`. Since we are using that field as a _temporary_ staging space to hold values until we map it to actual ECS/Vendor field and removing the entire thing at the end of the pipeline (L154).

qcorporation (approved):

PR #19491, opened 2026-06-11
ti_abusech: fix deleted_at pipeline error after upgrade to v4

### LINKED ISSUE #19490: ti_abusech: URL pipeline fails on upgrade from 3.x — missing deleted_at guard
When the ti_abusech package is upgraded from 3.x to 4.0.0+, the URL data stream's ingest pipeline produces hard errors on every document:

```
Processor date with tag date_deleted_at in pipeline logs-ti_abusech.url-4.0.0
failed with message field [deleted_at] not present as part of path [abusech.url.deleted_at]
```

### Root cause

The 4.0.0 pipeline is not upgrade-safe. On upgrade, the index template immediately routes documents through the new pipeline, but the agent may still be running the old (3.x) rendered configuration. The old config injects `_conf.interval` into documents; the new config injects `_conf.ioc_expiration_duration`. Until the agent receives and applies the re-rendered policy, every document arrives without `_conf.ioc_expiration_duration`.

The Painless script that computes `abusech.url.deleted_at` has a guard on `_conf.ioc_expiration_duration` being present, so it is skipped when the field is missing. The subsequent `date` processor runs unconditionally and fails because `deleted_at` was never computed.

The old pipeline used `_conf.interval` (copied to `labels.interval`) to drive expiration. The new pipeline removed that path but does not fall back to it.

### Evidence

Documents from affected agents show `_conf.interval` (set by the 3.x template) but no `_conf.ioc_expiration_duration` (added in 4.0.0). The `event.kind` is `pipeline_error` and error counts in the tens of thousands confirm the condition is persistent, not a brief race.

### Suggested fix

1. Add a guard to the `date` processor (`if: ctx.abusech?.url?.deleted_at != null`) so a missing value does not produce a hard error.
2. Fall back to a default 90-day expiration when `_conf.ioc_expiration_duration` is absent, so documents from pre-4.0.0 agents still get a usable `deleted_at`.

efd6:
## Proposed commit message
```
ti_abusech: fix deleted_at pipeline error after upgrade to v4

The date processor for deleted_at runs unconditionally but the
Painless script that computes the value is guarded by
_conf.ioc_expiration_duration being present. When upgrading from
an earlier version the agent may still be running the old template
which does not inject this field, causing a hard error on every
document.

Widen the script condition to always compute deleted_at and fall
back to a 90-day expiration when ioc_expiration_duration is absent.
Guard the date processor so a missing value is not a fatal error.
Apply the same fix to all six data streams and add pipeline tests
that simulate the upgrade scenario.

Fixes #19490
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #19490


navnit-elastic (approved):
LGTM

PR #19496, opened 2026-06-11
[akamai] - Replace CEL input with OpenTelemetry `akamai_siem` receiver

ShourieG:
## Type of change
- Enhancement
- Breaking Change
- Documentation

## Proposed commit message
```
akamai: replace CEL input with OpenTelemetry akamai_siem receiver

• siem_otel data stream: native OTel receiver replaces the CEL input
  as the API collection method. otelcol now backs the "Collect Akamai
  SIEM logs via API" policy-template option beside GCS. Breaking
  change → package 4.0.0. Requires stack 9.5.0; format_version
  3.6.0 (new SVR00006/SVR00009 checks excluded — legacy pipeline).

• Separate data stream because Fleet gives any otelcol-containing
  data stream a .otel-suffixed template pattern, which would orphan
  logs-akamai.siem-* for existing users.

• Routing: data_stream.dataset var + bodymap docs → events land in
  logs-akamai.siem-<ns>, existing pipeline reused.
  dynamic_dataset/namespace widen the API key to logs-*-* (Fleet
  scopes otelcol grants to .otel datasets that bodymap never writes
  to); the changelog documents the widened grant.

• Agent template: bodymap serializes only the body, so a transform
  writes tags (and the preserve_original_event tag) into the body map
  and seeds @timestamp from the record time — the pipeline only sets
  it when httpMessage.start is present (500s omit it), so those
  events were otherwise rejected by the data stream. String vars are
  quoted (all-digit values ≠ YAML ints); persist_cursor (on by
  default) declares a file_storage extension and sets
  storage: file_storage on the receiver so the poll cursor survives
  agent restarts.

• Tests: policy tests pin the compiled config — logs-*-* grant,
  storage wiring, tags/preserve_original_event rendering, cursor off.
  System test runs against a SIEM API emulator but is skipped pending
  elastic-package#2109 (asserting on a redirected dataset);
  initial_interval is capped at 12h — the receiver rejects larger
  values rather than clamping.

• Breaking-change changelog + README: existing API policies must be
  reconfigured after upgrade, within 12 hours — the SIEM API only
  serves the previous 12h of events; SSL options must be rewritten in
  OTel tls schema. README reorganised around two collection methods
  (API/GCS); cursor persistence ("Persist Cursor" option) not
  guaranteed across stack upgrades and breaking changes.
```

## Checklist

## NOTE:

Upgrade path was tested locally before going for a breaking-change

Steps: 
 - Version `3.2.0 (CEL Input)` was configured with akamai mock server and ingestion started.
 - Initially a version `3.3.0` was made with CEL being replaced by otel receiver and the input was enabled by default. 
    In this scenario, while upgrading we saw the following error :-
        
Screenshot 2026-07-24 at 5 09 13 PM

 This is because the separate data stream approach ( adopted due to fleet limitations) changed the fleet computed attribute paths and they were no longer configured.
 
 - Finally we created a breaking change with version `4.0.0` and kept the input disabled by default and the upgrade  happened without any errors. 
 - After reconfiguring the updated integration, ingestion started as usual but with much higher throughput.

 
## Author's Checklist

- [ ] 

## Related issues

- Relates https://github.com/elastic/opentelemetry-collector-components/pull/1119

### All Inputs

Screenshot 2026-07-24 at 6 31 37 PM

### Receiver Config

Screenshot 2026-07-24 at 6 32 36 PM


ShourieG:
@jamiehynds, wanted to get your opinion on this change from an user perspective.

With the current limitations imposed by Fleet, having an otel receiver config within an existing data stream breaks it for existing users since fleet renames the index template by appending an `.otel` suffix to the template name. Because of this I needed to create a separate data stream for the otel receiver and then route the data to the existing data stream to avoid ingest pipeline duplication.

The `user` sees this as a new input type in addition to the existing `cel` and `gcs` inputs as shown in the screenshots. Upon configuring and using it the data lands in the original `akamai.siem` data stream. The issue is now we have almost two similar inputs with almost identical configurations,  the `CEL API` based Input and the `OpenTelemetry` receiver that also uses the api underneath but with custom `Go` code that makes it way more stable and performant compared to the `CEL` input. We went the OTEL route since Akamai pull api has a lot of issues and the CEL implementation suffers a lot from these limitations with various reported and escalated in the past.

The issue now is about user perception. We end up having two similar looking inputs from the outside `CEL API` and `OpenTelemetry receiver` but the latter is much more stable an performant. 

Do we keep both of these or phase out the older `CEL` implementation slowly or what pitch do we make to users to it's less confusing on their side.

cc: @narph

jamiehynds:
Collection transport - I'm assuming none of the managed OTel paths apply, since we're still hitting the same Akamai SIEM API and Akamai isn't pushing to us. mOTLP (the Managed OTLP Endpoint) is out - that's a destination you send to, not something that polls, so Akamai can't push to a receiver like that; ECF doesn't fit for the same reason. We need a durable collector to pull Akamai's API on an interval. Right?

Agentless — Today agentless is limited to cel/httpjson, so I assume the gap is that the agentless runtime doesn't yet run the EDOT Collector / otel input type? Worth flagging: the CEL Akamai input can likely run agentless today, so if we make the OTel receiver the recommended path but it forces a self-managed agent, that's a regression for those users. What would it take to get the otel input supported in agentless? 

Packaging — would an OTel input package (akamai_input_otel) be a better fit than bolting the receiver into the existing data stream? My assumption is no: Akamai isn't OTel-native, we're only using the Collector to hit their API, and input packages store data OTel-native rather than ECS in logs-akamai.siem-* - so we'd lose pipeline/dashboard/detection-rule continuity for existing users and need a migration path, whereas your current design deliberately routes back through the existing ECS pipeline to keep all that working. 

cc @qcorporation @andrewkroh as this may relate to ingest hub onboarding, and it's the first example I'm aware of using an OTel collector but mapping back to ECS

qcorporation:
> Collection transport - I'm assuming none of the managed OTel paths apply, since we're still hitting the same Akamai SIEM API and Akamai isn't pushing to us. mOTLP (the Managed OTLP Endpoint) is out - that's a destination you send to, not something that polls, so Akamai can't push to a receiver like that; ECF doesn't fit for the same reason. We need a durable collector to pull Akamai's API on an interval. Right?
> 
> Agentless — Today agentless is limited to cel/httpjson, so I assume the gap is that the agentless runtime doesn't yet run the EDOT Collector / otel input type? Worth flagging: the CEL Akamai input can likely run agentless today, so if we make the OTel receiver the recommended path but it forces a self-managed agent, that's a regression for those users. What would it take to get the otel input supported in agentless?
> 
> Packaging — would an OTel input package (akamai_input_otel) be a better fit than bolting the receiver into the existing data stream? My assumption is no: Akamai isn't OTel-native, we're only using the Collector to hit their API, and input packages store data OTel-native rather than ECS in logs-akamai.siem-* - so we'd lose pipeline/dashboard/detection-rule continuity for existing users and need a migration path, whereas your current design deliberately routes back through the existing ECS pipeline to keep all that working.
> 
> cc @qcorporation @andrewkroh as this may relate to ingest hub onboarding, and it's the first example I'm aware of using an OTel collector but mapping back to ECS

Really interesting conversation cc.ing @nimarezainia as well

ShourieG:
> Collection transport - I'm assuming none of the managed OTel paths apply, since we're still hitting the same Akamai SIEM API and Akamai isn't pushing to us. mOTLP (the Managed OTLP Endpoint) is out - that's a destination you send to, not something that polls, so Akamai can't push to a receiver like that; ECF doesn't fit for the same reason. We need a durable collector to pull Akamai's API on an interval. Right?
> 
> Agentless — Today agentless is limited to cel/httpjson, so I assume the gap is that the agentless runtime doesn't yet run the EDOT Collector / otel input type? Worth flagging: the CEL Akamai input can likely run agentless today, so if we make the OTel receiver the recommended path but it forces a self-managed agent, that's a regression for those users. What would it take to get the otel input supported in agentless?
> 
> Packaging — would an OTel input package (akamai_input_otel) be a better fit than bolting the receiver into the existing data stream? My assumption is no: Akamai isn't OTel-native, we're only using the Collector to hit their API, and input packages store data OTel-native rather than ECS in logs-akamai.siem-* - so we'd lose pipeline/dashboard/detection-rule continuity for existing users and need a migration path, whereas your current design deliberately routes back through the existing ECS pipeline to keep all that working.
> 
> cc @qcorporation @andrewkroh as this may relate to ingest hub onboarding, and it's the first example I'm aware of using an OTel collector but mapping back to ECS

@jamiehynds,

1. Yes currently Akamai otel receiver also uses a polling logic and it has local cursor state.

2. I'm working on a fix to make sure we can enable Agentless with the akamai receiver [here](https://github.com/elastic/beats/issues/50223). Once done this thould help other otel receivers to also work in agentless mode. 

3. Making a separate package will just end up duplicating the ingest pipelines without any real benefit since we do not output actual otel format atm.

andrewkroh:
My take is this shouldn't be an additional input. The receiver should replace the CEL input with an invisible migration, so we have one API collection mechanism and one code path.

- For invisible migration we use Fleet's `migrate_from` feature. The CEL config maps over almost one-to-one (we need to align the names for `api_host` → `endpoint`, `interval` → `poll_interval`, `initial_interval` → `initial_lookback`, but the rest are identical), so an existing CEL user shouldn't have to touch anything on upgrade.
- The `.otel` suffix is a blocker. And we need to solve this with the Fleet team in Kibana. It's applied in EPM purely because the input type is `otelcol`. Our receiver emits ECS (bodymap), not semconv, so the data belongs in `logs-akamai.siem-*` as it does today. I'd rather get Fleet to let an `otelcol` input write to a classic non-`.otel` dataset than ship any workaround.
- Losing the cursor on cutover is low impact. Akamai's offset token is very short-lived, and the existing pipeline already fingerprints `event.original` into `_id` in the same index, so re-fetched data is deduplicated. But only if the `event.original` values are exactly the same, so we need to check this.
- The akamai package doesn't support agentless today so there are no agentless CEL users to regress. Therefore we are not blocked on the missing OTel storage extension that would let the receiver persist state to ES.

I would like the same package, replaced input, same dataset.

ShourieG:
> My take is this shouldn't be an additional input. The receiver should replace the CEL input with an invisible migration, so we have one API collection mechanism and one code path.
> 
> * For invisible migration we use Fleet's `migrate_from` feature. The CEL config maps over almost one-to-one (we need to align the names for `api_host` → `endpoint`, `interval` → `poll_interval`, `initial_interval` → `initial_lookback`, but the rest are identical), so an existing CEL user shouldn't have to touch anything on upgrade.
> * The `.otel` suffix is a blocker. And we need to solve this with the Fleet team in Kibana. It's applied in EPM purely because the input type is `otelcol`. Our receiver emits ECS (bodymap), not semconv, so the data belongs in `logs-akamai.siem-*` as it does today. I'd rather get Fleet to let an `otelcol` input write to a classic non-`.otel` dataset than ship any workaround.
> * Losing the cursor on cutover is low impact. Akamai's offset token is very short-lived, and the existing pipeline already fingerprints `event.original` into `_id` in the same index, so re-fetched data is deduplicated. But only if the `event.original` values are exactly the same, so we need to check this.
> * The akamai package doesn't support agentless today so there are no agentless CEL users to regress. Therefore we are not blocked on the missing OTel storage extension that would let the receiver persist state to ES.
> 
> I would like the same package, replaced input, same dataset.

@andrewkroh, but what about the state ? Currently the CEL state is not compatible with otel receiver, so it will be a breaking change for existing users on older stacks and then  upgrading to 9.5 ?

nimarezainia:
> Agentless — Today agentless is limited to cel/httpjson, so I assume the gap is that the agentless runtime doesn't yet run the EDOT Collector / otel input type? Worth flagging: the CEL Akamai input can likely run agentless today, so if we make the OTel receiver the recommended path but it forces a self-managed agent, that's a regression for those users. What would it take to get the otel input supported in agentless?

@jamiehynds  we are no longer bound by cel/httpjson only inputs, observabilty team has already added aws-metrics (as an example) and worked on supporting many other integrations.

@ishleenk17 and @lalit-satapathy can comment on the otel input and agentless deployment as their teams have been adapting many otel packages to agentless as well.

andrewkroh:
>> Losing the cursor on cutover is low impact. Akamai's offset token is very short-lived, and the existing pipeline already fingerprints event.original into _id in the same index, so re-fetched data is deduplicated. But only if the event.original values are exactly the same, so we need to check this.

> @andrewkroh, but what about the state ? Currently the CEL state is not compatible with otel receiver, so it will be a breaking change for existing users on older stacks and then upgrading to 9.5 ?

@ShourieG Isn't the pagination offset token only valid for 2 minutes? This would mean that cursor reset is a very normal occurrence, so a change-over that does not retain state will be indistinguishable from other events like an agent software upgrade that causes more than 2 minutes of downtime.

ShourieG:
> >> Losing the cursor on cutover is low impact. Akamai's offset token is very short-lived, and the existing pipeline already fingerprints event.original into _id in the same index, so re-fetched data is deduplicated. But only if the event.original values are exactly the same, so we need to check this.
> 
> 
> 
> > @andrewkroh, but what about the state ? Currently the CEL state is not compatible with otel receiver, so it will be a breaking change for existing users on older stacks and then upgrading to 9.5 ?
> 
> 
> 
> @ShourieG Isn't the pagination offset token only valid for 2 minutes? This would mean that cursor reset is a very normal occurrence, so a change-over that does not retain state will be indistinguishable from other events like an agent software upgrade that causes more than 2 minutes of downtime.

Yes this is true, token validity is indeed 2mins atm. But then I think we should explicitly call it out in some upgrade note/description etc, that this will be a breaking change, and increment to version 4.0.0. This is just to preserve clarity with the end user on what's happening.

I have just one concern with out right replacing, that is, is not better to do a bit of real-world testing with end users/customers before completely removing the CEL input ? That way we could potentially identify any issues that might exist. The CEL input is GA and has been for a while. The new input has not been stress tested under real circumstances yet.

ishleenk17:
> > Agentless — Today agentless is limited to cel/httpjson, so I assume the gap is that the agentless runtime doesn't yet run the EDOT Collector / otel input type? Worth flagging: the CEL Akamai input can likely run agentless today, so if we make the OTel receiver the recommended path but it forces a self-managed agent, that's a regression for those users. What would it take to get the otel input supported in agentless?
> 
> @jamiehynds we are no longer bound by cel/httpjson only inputs, observabilty team has already added aws-metrics (as an example) and worked on supporting many other integrations.
> 
> @ishleenk17 and @lalit-satapathy can comment on the otel input and agentless deployment as their teams have been adapting many otel packages to agentless as well.

Is the question if OTel receivers can be used in agentless mode ?

narph:
The reason we developed a dedicated Akamai input in OTel was to address known limitations with the Akamai SIEM API and the existing CEL-based implementation. This issue has affected not only us but also other vendors and competitors that have integrated with the API.

Adding both options would likely create confusion for users without really solving the underlying problem. We'd still be offering an ingestion method that we know has limitations and isn't the preferred approach.

We should stay consistent with how we've handled this in the past and replace the existing input rather than add another option. Since the upgrade impact and the effort required to switch inputs are minimal, replacing it gives users a simpler experience and steers them toward the more reliable solution.

ShourieG:
@narph @andrewkroh, Did an analysis with the results below, the main friction of replacing without breaking changes is that the ssl/tls configs differ a bit in beats vs otel.

---

## Akamai SIEM: TLS config mapping (Beats `ssl:` → OTel `tls:`)

When moving the Akamai SIEM collection from the **CEL input** (`data_stream/siem`) to the
**`akamai_siem` OTel receiver** (`data_stream/siem_otel`), the transport-security config
has to be re-expressed. Both configure the **same protocol** — TLS for the outbound HTTPS
connection to the Akamai API. "SSL" is just legacy naming.

Only the **config schema** differs:

| | Beats (CEL input) | OTel (`akamai_siem` receiver) |
|---|---|---|
| Config block name | `ssl:` | `tls:` |
| Underlying protocol | TLS | TLS (identical) |
| Source schema | [libbeat SSL common config][beats-ssl-docs] | [`configtls.ClientConfig`][otel-configtls-pkg] v1.60.0 (squashed via [`confighttp`][otel-confighttp-pkg]) |

The bytes on the wire are the same. The migration cost is **translating one YAML schema
into the other**, not changing the encryption.

## Beats `ssl:` (current CEL input)

```yaml
ssl:
  certificate_authorities:                 # LIST of files or inline PEM
    - /etc/agent/certs/akamai-ca.pem
    - /etc/agent/certs/corp-root.pem
  certificate: /etc/agent/certs/client.pem
  key: /etc/agent/certs/client-key.pem
  key_passphrase: "${KEY_PASS}"
  verification_mode: certificate           # full | strict | certificate | none
  supported_protocols: [TLSv1.2, TLSv1.3]  # explicit list
  cipher_suites:
    - ECDHE-RSA-AES-256-GCM-SHA384         # OpenSSL-style names
  curve_types: [P-256]
  ca_trusted_fingerprint: "9f8b...c2"      # pin by sha256 fingerprint
```

## OTel `tls:` (akamai_siem receiver)

[`confighttp.ClientConfig`][otel-confighttp-pkg] squashes [`configtls.ClientConfig`][otel-configtls-pkg]
at the receiver root, so this block sits directly under the receiver next to `endpoint:`.
The keys below are defined in [`configtls.Config`][otel-configtls-config-pkg] (shared) plus the
client-only fields in [`configtls.ClientConfig`][otel-configtls-pkg].

```yaml
tls:
  ca_file: /etc/agent/certs/akamai-bundle.pem   # the two CAs above concatenated into ONE file
  cert_file: /etc/agent/certs/client.pem
  key_file: /etc/agent/certs/client-key.pem
  # key_passphrase:        -> no equivalent; key must be decrypted first
  insecure_skip_verify: false                   # only `verification_mode: none` maps here (-> true)
  min_version: "1.2"                            # note "1.2", not "TLSv1.2"
  max_version: "1.3"
  cipher_suites:
    - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384     # IANA/Go names, NOT the Beats names
  curve_preferences: [P-256]
  include_system_ca_certs_pool: true            # see behavioral note below
  # ca_trusted_fingerprint: -> no equivalent
```

## Name changes (mechanical renames)

| Beats `ssl:` | OTel `tls:` | Note |
|---|---|---|
| `certificate_authorities` (list) | `ca_file` / `ca_pem` (single) | list → single; concat multiple CAs into one PEM |
| `certificate` | `cert_file` / `cert_pem` | rename only |
| `key` | `key_file` / `key_pem` | rename only |
| `curve_types` | `curve_preferences` | rename (value naming also differs) |
| `cipher_suites` | `cipher_suites` | same key, **different value vocabulary** (OpenSSL → IANA/Go names) |

## Lossy conversions (need human judgment)

| Beats `ssl:` | OTel `tls:` | Why lossy |
|---|---|---|
| `verification_mode: none` | `insecure_skip_verify: true` | enum → bool |
| `verification_mode: full` / `strict` | `insecure_skip_verify: false` (default) | enum → bool |
| `supported_protocols: [..]` | `min_version` / `max_version` | list → range; can't express non-contiguous sets (e.g. 1.1 + 1.3 but not 1.2); version string format differs |

## No OTel equivalent

| Beats `ssl:` | Status |
|---|---|
| `key_passphrase` | ❌ no field — passphrase-protected keys won't load; decrypt the key first |
| `verification_mode: certificate` | ❌ no equivalent — "verify chain but skip hostname" can't be expressed (OTel is verify-all or verify-nothing) |
| `ca_trusted_fingerprint` / `ca_sha256` | ❌ no fingerprint/sha256 CA pinning in `configtls` |
| `renegotiation` | ❌ no equivalent |

## Behavioral changes: system CA pool default flips

In `configtls`, specifying `ca_file`/`ca_pem` means you trust **only** that CA — the system
root pool is excluded unless you set `include_system_ca_certs_pool: true`. A Beats user who
added a private CA generally still has system roots available, so a naïve port can silently
break TLS to anything signed by a public CA.

---

## References

OTel side (resolved module versions in this repo):

- **`configtls.ClientConfig`** — the `tls:` schema (client-only fields: `insecure`,
  `insecure_skip_verify`, `server_name_override`), v1.60.0:
  [pkg.go.dev][otel-configtls-pkg] · [source][otel-configtls-src]
- **`configtls.Config`** — the shared/squashed TLS fields (`ca_file`, `ca_pem`, `cert_file`,
  `cert_pem`, `key_file`, `key_pem`, `min_version`, `max_version`, `cipher_suites`,
  `curve_preferences`, `include_system_ca_certs_pool`, `reload_interval`, …), v1.60.0:
  [pkg.go.dev][otel-configtls-config-pkg]
- **`confighttp.ClientConfig`** — owns the `tls` field and applies it via `ToClient`
  (`cc.TLS.LoadTLSConfig` → `transport.TLSClientConfig`), v0.154.0:
  [pkg.go.dev][otel-confighttp-pkg] · [`ToClient`][otel-confighttp-toclient] · [source][otel-confighttp-src]

In-repo wiring (this receiver):

- Squash that exposes `tls:` at the receiver root — [`config.go`][receiver-config] (`HTTP confighttp.ClientConfig` with `,squash`)
- Where the client (incl. TLS transport) is built — [`receiver.go`][receiver-toclient] (`r.cfg.HTTP.ToClient(...)`)

Beats side:

- **libbeat `tlscommon`** — the `ssl:` schema source: [config.go][beats-tlscommon-src]
- **SSL common config reference** (what the CEL `ssl` var description links to): [elastic.co][beats-ssl-docs]

[otel-configtls-pkg]: https://pkg.go.dev/go.opentelemetry.io/collector/config/configtls@v1.60.0#ClientConfig
[otel-configtls-config-pkg]: https://pkg.go.dev/go.opentelemetry.io/collector/config/configtls@v1.60.0#Config
[otel-configtls-src]: https://github.com/open-telemetry/opentelemetry-collector/blob/config/configtls/v1.60.0/config/configtls/configtls.go
[otel-confighttp-pkg]: https://pkg.go.dev/go.opentelemetry.io/collector/config/confighttp@v0.154.0#ClientConfig
[otel-confighttp-toclient]: https://pkg.go.dev/go.opentelemetry.io/collector/config/confighttp@v0.154.0#ClientConfig.ToClient
[otel-confighttp-src]: https://github.com/open-telemetry/opentelemetry-collector/blob/config/confighttp/v0.154.0/config/confighttp/client.go
[receiver-config]: ../../receiver/akamaisiemreceiver/config.go
[receiver-toclient]: ../../receiver/akamaisiemreceiver/receiver.go
[beats-tlscommon-src]: https://github.com/elastic/beats/blob/main/libbeat/common/transport/tlscommon/config.go
[beats-ssl-docs]: https://www.elastic.co/guide/en/beats/filebeat/current/configuration-ssl.html#ssl-common-config

ShourieG:
@narph @andrewkroh, I've added a key by key explanation for the hard gaps (missing keys) for the ssl block and their possible impact.

---

## Otel vs Beats (TLS Gaps)

| Beats `ssl:` | OTel `tls:` | Impact | Outcome |
|---|---|---|---|
| `key_passphrase` | nothing | can't use an encrypted private key | won't start |
| `verification_mode: certificate` | nothing (`server_name_override` helps a bit) | can't "check the cert but skip the hostname" | won't start |
| `ca_trusted_fingerprint` / `ca_sha256` | nothing | must supply the full CA file instead | won't start |
| `renegotiation` | nothing | handshake may fail with old proxies | won't start |
| (system CAs trusted by default) | `include_system_ca_certs_pool` (off by default) | adding a CA quietly drops public trust | starts anyway |

---

## Examples to explain each scenario

### 1. `key_passphrase` — encrypted private key

You use a client certificate to get through a proxy (mTLS), and your security
rules say the private key must be stored encrypted. Beats unlocks it with a
passphrase; the receiver can't.

```yaml
# Beats — works
ssl: { certificate: client.pem, key: client.key, key_passphrase: "${PASS}" }

# OTel — no passphrase setting; the key has to be unencrypted on disk
tls: { cert_file: client.pem, key_file: client.key }   # encrypted key -> fails to load
```

### 2. `verification_mode: certificate` — check the cert, skip the hostname

You go through a proxy whose certificate is valid (signed by a CA you trust) but
the hostname doesn't match. Beats can still check the certificate and ignore the
hostname. The receiver only does all-or-nothing.

```yaml
# Beats — check the cert, allow the hostname mismatch
ssl: { verification_mode: certificate }

# OTel — two blunt options:
tls: { insecure_skip_verify: true }             # turns off all checking (weaker)
tls: { server_name_override: "real-cert-name" } # works if you know the name on the cert
```

### 3. `ca_trusted_fingerprint` / `ca_sha256` — trust a CA by fingerprint

You trust your proxy's internal CA by its fingerprint instead of copying the whole
certificate file. The receiver has no fingerprint option — you give it the actual
certificate.

```yaml
# Beats — paste the fingerprint
ssl: { ca_trusted_fingerprint: "9f8b…c2" }

# OTel — point at the real certificate file
tls: { ca_file: /etc/agent/proxy-ca.pem }
```

### 4. `renegotiation` — TLS renegotiation

An old proxy renegotiates the TLS connection partway through (some do this for
client-cert auth). Beats lets you allow it; the receiver doesn't have the setting.
This rarely matters — TLS 1.3 dropped renegotiation.

```yaml
# Beats
ssl: { renegotiation: freely }

# OTel — no setting; Go's defaults apply
```

### 5. `include_system_ca_certs_pool` 

When you add your proxy's private CA, the receiver then trusts only that CA.
Anything signed by a normal public CA stops working — and nothing warns you,
because the config is valid.

```yaml
# Beats — adding a CA doesn't drop public trust
ssl: { certificate_authorities: [proxy-ca.pem] }

# OTel — turn public CAs back on, or the direct Akamai connection breaks
tls:
  ca_file: proxy-ca.pem
  include_system_ca_certs_pool: true   # leave this out and the Akamai connection fails
```

narph:
This looks like a functionality degradation, but I wonder how often these options are actually used. @andrewkroh, do you think we can proceed with this? It could break existing setups.

andrewkroh:
None of those missing TLS config options seem like blockers because the input is talking to a SaaS API with a CA cert issued by a widely trusted CA. If this is widely shared TLS config in the otel, then we should log issue(s) to close that gap between what the agent offers because it could be important for future use-cases.

ShourieG:
@vera-review-bot review

packages/akamai/data_stream/siem_otel/manifest.yml
```diff
@@ -0,0 +1,152 @@
+title: Akamai SIEM Logs
```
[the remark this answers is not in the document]
ShourieG: currently not possible due to routing and how elastic_package handles routing.

packages/akamai/data_stream/siem_otel/manifest.yml
```diff
@@ -0,0 +1,152 @@
...
+  # the grant to logs-*-* (same as OTel input packages) so writes to
+  # logs-akamai.siem-<namespace> are authorized.
+  dynamic_dataset: true
+  dynamic_namespace: true
+streams:
+  - input: otelcol
+    template_path: otelcol.yml.hbs
+    title: Akamai SIEM logs
+    description: Collect Akamai logs via the SIEM API
+    enabled: false
```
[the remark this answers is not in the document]
ShourieG: The disabling is intentional, doc is updated.

packages/akamai/data_stream/siem_otel/_dev/test/system/test-emulator-config.yml
```diff
@@ -0,0 +1,30 @@
+wait_for_data_timeout: 1m
+input: otelcol
+service: akamai-siem-emulator
+# The akamai_siem receiver redirects all events to the akamai.siem dataset
+# (see the data_stream.dataset var), so hits land in logs-akamai.siem-* rather
+# than logs-akamai.siem_otel-*. elastic-package cannot yet assert on a
+# redirected dataset, so this test is skipped until that support lands.
+skip:
```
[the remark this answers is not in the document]
ShourieG: policy test has been added.

packages/akamai/data_stream/siem_otel/manifest.yml
```diff
@@ -0,0 +1,152 @@
...
+      # collection modes present the same configuration surface. They render
+      # into the receiver's native config keys in otelcol.yml.hbs.
+      - name: api_host
+        type: text
+        title: API Host
+        description: API Hostname in the form of http(s)://akzz-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX.luna.akamaiapis.net without path
+        multi: false
+        required: true
+        show_user: true
+        default: https://akzz-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX.luna.akamaiapis.net
```
chrisberkhout, line 30: It thought this URL was a bit much, but I see something similar in the Akamai docs.
[There](https://techdocs.akamai.com/iam-api/v1/reference/get-started) it's `https://akzz-XXXXXXXXXXXXXXXX-XXXXXXXXXXXXXXXX.luna.akamaiapis.net`, which has an extra dash.

Is `http://` really an option? If not we can say `https://...` rahter than `http(s)://...`.

I'd leave that default value out since it's not valid. The description is enough and it's good to force the user to enter something.

packages/akamai/data_stream/siem_otel/manifest.yml
```diff
@@ -0,0 +1,152 @@
...
+        description: Dataset used by Fleet to route events. Defaults to akamai.siem so events are processed by the existing Akamai SIEM ingest pipeline and stored in logs-akamai.siem-<namespace>. Do not change unless you know what you are doing.
+        multi: false
+        required: true
+        show_user: false
+        default: akamai.siem
+      - name: enable_storage
+        type: bool
+        title: Enable Cursor Persistence
+        description: >-
+          Declares a file_storage OpenTelemetry extension and references it from the akamai_siem receiver so the receiver persists its poll cursor across agent restarts. Enabled by default. Requires Elastic Stack 9.5.0 or later, where Fleet rewrites the renamed storage-extension reference into the receiver configuration. When disabled, the receiver re-fetches the Initial Interval window after a restart; replayed events are deduplicated by the ingest pipeline's event.original fingerprint within the same backing index.
```
chrisberkhout, line 148: This option would be better called "Persist Cursor" / `persist_cursor`, and then we can enable or disable that, rather than having "enable" in the name.

packages/akamai/data_stream/siem_otel/agent/stream/otelcol.yml.hbs
```diff
@@ -0,0 +1,53 @@
...
+  # logs, so tags must be written into the body map: in bodymap mode the
+  # Elasticsearch exporter serializes only the body into the document (resource
+  # and log-record attributes are dropped), so a top-level `tags` array on the
+  # document must live under body["tags"].
+  transform/tags:
+    error_mode: ignore
+    log_statements:
+      - context: log
+        statements:
+          - set(log.body["tags"], [{{#if preserve_original_event}}"preserve_original_event"{{#if tags}}, {{/if}}{{/if}}{{#each tags}}"{{this}}"{{#unless @last}}, {{/unless}}{{/each}}])
```
chrisberkhout, line 39: This looks wrong. As is, tags are only populated if `preserve_original_event` is set.

This might benefit from being split over several lines and using the template whitespace trimming to get it back to one line.

Adding some tags to the policy test is the easy way to check it's working right.

packages/akamai/data_stream/siem_otel/fields/base-fields.yml
chrisberkhout, line 1: Do these need to be here if everything is redirected and nothing lands in a `siem_otel` data stream?

packages/akamai/_dev/build/docs/README.md
```diff
@@ -2,6 +2,8 @@
 
 The Akamai integration collects events from the Akamai API, specifically reading from the [Akamai SIEM API](https://techdocs.akamai.com/siem-integration/reference/api).
 
+> **Breaking change in 4.0.0**: The backend used to collect Akamai SIEM logs from the API has changed. This version requires Elastic Stack 9.5.0 or later. After upgrading, existing agent policies that collect Akamai SIEM logs via the API must be reconfigured for data collection to resume. If you previously set the advanced **SSL Configuration** option, it must be rewritten using the OpenTelemetry `tls` schema (`ca_file`/`ca_pem`, `cert_file`/`cert_pem`, `key_file`/`key_pem`, `insecure_skip_verify`, `min_version`, `max_version`) rather than the Beats keys (`certificate_authorities`, `verification_mode`, `supported_protocols`). Collection from Google Cloud Storage is unaffected.
```
chrisberkhout, line 5: Could mention that it's only possible to get 12 hours of historical data, so the new policy should be configured within that time to avoid data gaps.

chrisberkhout (changes requested):
For the commit message:
- Has this been manually tested against a live API?
- Should it be "OpenTelemetry `akamai_siem` receiver" rather than "OpenTelemetry `akamaisiemreceiver`"?

Question: Can you please say the reason why the new `siem_otel` collection can't be done directly in `data_stream/siem` if it's correct that it wouldn't work? Is it because of a fleet issue during upgrade? If it could be done it would avoid `data_steam/siem_otel` existing without being populated itself.

ShourieG:
> For the commit message:
> 
> * Has this been manually tested against a live API?
> * Should it be "OpenTelemetry `akamai_siem` receiver" rather than "OpenTelemetry `akamaisiemreceiver`"?
> 
> Question: Can you please say the reason why the new `siem_otel` collection can't be done directly in `data_stream/siem` if it's correct that it wouldn't work? Is it because of a fleet issue during upgrade? If it could be done it would avoid `data_steam/siem_otel` existing without being populated itself.

@chrisberkhout, 

This has been live tested against Andrew's mock Akamai API at https://proteus.sit.estc.dev/. There's a mock active, you can check it out. Getting access to a real env is tricky because of setup issues. 

We can rename to OpenTelemetry `akamai_siem` receiver, but kept it `akamaisiemreceiver` to align to the actual receiver name.

The reason we can't do `siem_otel` in the existing `siem` data stream is because fleet current renames the index template of the data stream by adding a `.otel` suffix to it, if it sees an otelcol type input. So `logs-akamai.siem` becomes `logs-akamai.siem.otel` with match pattern `logs-akamai.siem.otel-*`. Creating a separate data stream avoids this. We would still end up creating a `logs-akamai.siem.otel` template but nothing would ever write to it due to our custom receiver level routing.

packages/akamai/data_stream/siem_otel/agent/stream/otelcol.yml.hbs
ShourieG: The `{{#each tags}}` is actually outside the `{{#if preserve_original_event}}` block — the
nested `{{#if tags}}, {{/if}}` only emits the separating comma when both the
`preserve_original_event` tag and user tags are present. So tags render regardless of the
flag; the policy test output shows `["akamai-siem", "forwarded"]` with
`preserve_original_event` unset.

But I agree it's hard to read a one liner like this, so I'll make it multiline.

packages/akamai/data_stream/siem_otel/fields/base-fields.yml
ShourieG: Yea it's not required technically, but it's required by package-spec so removing it will make elastic-package/CI fail.

ShourieG:
@chrisberkhout, I've addressed all the issues.

packages/akamai/data_stream/siem_otel/fields/base-fields.yml
chrisberkhout: okay. Need to leave then, unless it something that can be skipped with an entry in `validation.yml`.

chrisberkhout:
> This has been live tested against Andrew's mock Akamai API at https://proteus.sit.estc.dev/. There's a mock active, you can check it out. Getting access to a real env is tricky because of setup issues.

Okay, that's fine. Could optionally note the mock-only testing in the end of the commit message.

> The reason we can't do `siem_otel` in the existing `siem` data stream is because fleet current renames the index template of the data stream by adding a `.otel` suffix to it, if it sees an otelcol type input. So `logs-akamai.siem` becomes `logs-akamai.siem.otel` with match pattern `logs-akamai.siem.otel-*`. Creating a separate data stream avoids this. We would still end up creating a `logs-akamai.siem.otel` template but nothing would ever write to it due to our custom receiver level routing.

Thanks, good explanation. That makes sense. So it's clear in the PR, but for future maintenance it would help to have a comment in the policy config, near the `data_stream.dataset` part, saying that the receiver's writes are redirected to `siem` and what Fleet does to the package's `siem_otel` DS.

> We can rename to OpenTelemetry `akamai_siem` receiver, but kept it `akamaisiemreceiver` to align to the actual receiver name.

Okay. I see `akamai_siem` in the README and policy config. So if `akamai_siem` is the Beats name, maybe that's the one we should prefer?

packages/akamai/data_stream/siem_otel/agent/stream/otelcol.yml.hbs
chrisberkhout: Thanks. Looks nice.

packages/akamai/_dev/build/docs/README.md
chrisberkhout: Don't see this part. Did you mean to commit something for that?

chrisberkhout (approved):
Thanks for the updates and explanations.

Remaining points worth a final look are only about doc/code comments:
- https://github.com/elastic/integrations/pull/19496#discussion_r3747690519
  (policy update required within max data range of 12h to avoid gaps)
- https://github.com/elastic/integrations/pull/19496#issuecomment-5237502985
  (testing in commit message, maintainer comment in policy, receiver naming)

Otherwise GTG

packages/akamai/data_stream/siem_otel/agent/stream/otelcol.yml.hbs
```diff
@@ -0,0 +1,62 @@
...
+  # logs, so tags must be written into the body map: in bodymap mode the
+  # Elasticsearch exporter serializes only the body into the document (resource
+  # and log-record attributes are dropped), so a top-level `tags` array on the
+  # document must live under body["tags"].
+  transform/tags:
+    error_mode: ignore
+    log_statements:
+      - context: log
+        statements:
+          - set(log.body["tags"], [
```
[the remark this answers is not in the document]
ShourieG: fixed in latest commit

packages/akamai/_dev/build/docs/README.md
ShourieG: addressed this.

PR #19503, opened 2026-06-12
[cloudflare_logpush] Fix ingestion failure for events collected via the HTTP Endpoint input

HW-Jeremy:
## Proposed commit message

```
cloudflare_logpush: fix ingestion failure for HTTP Endpoint input events

The 1.45.0 ingest pipeline rewrite (#18952, #19234) removed the
ignore_failure guard that previously existed on the json processor
parsing event.original (present up to 1.44.2).

With the http_endpoint input, the request body is parsed by the input
itself into the "json" field: there is no "message" field, so
event.original is never populated by the pipeline preamble. The
unguarded json processor then hard-fails with:

  Processor json with tag json_event_original_to_json_cac66847 in
  pipeline logs-cloudflare_logpush.firewall_event-1.46.0 failed with
  message: field [original] not present as part of path [event.original]

Because the processor has no on_failure handler, the failure aborts the
whole pipeline: every event delivered through the HTTP Endpoint input is
indexed unparsed (no ECS fields, no cloudflare_logpush.* fields,
@timestamp set to ingest time) with event.kind: pipeline_error.

Restore the guard by only running the json processor when
event.original is present, across all data streams.
```

## What happened

After upgrading the package from 1.44.x to 1.45.0+/1.46.x, 100% of events collected through the **HTTP Endpoint** input fail the ingest pipeline. Example failed document (firewall_event, Elastic Agent 8.19.15, `input.type: http_endpoint`): the full Cloudflare payload is present under `json.*` (parsed by the input), `message`/`event.original` are absent, and the document carries:

```
error.message: Processor json with tag json_event_original_to_json_cac66847 in pipeline
logs-cloudflare_logpush.firewall_event-1.46.0 failed with message:
field [original] not present as part of path [event.original]
event.kind: pipeline_error
```

Up to 1.44.2 the processor was:

```yaml
- json:
    field: event.original
    target_field: json
    ignore_failure: true
```

so the missing field was tolerated and the pipeline continued with the `json.*` fields provided by the input. The 1.45.0 rewrite dropped the guard.

## Checklist

Send a firewall_event document through the pipeline without `message`/`event.original` but with the payload already under `json` (as produced by the http_endpoint input): before this change the pipeline aborts with `pipeline_error`; after this change the event is parsed normally.

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/packages/cloudflare_logpush/changelog.yml
+++ b/packages/cloudflare_logpush/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.46.2"
+  changes:
+    - description: Fix ingestion failure for events collected via the HTTP Endpoint input by only parsing `event.original` when it is present. The input parses the request body into `json` itself, so `event.original` may be absent.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/1
 - version: "1.46.1"
   changes:
     - description: Use the ECS definition for email.attachments.
--- a/packages/cloudflare_logpush/changelog.yml
+++ b/packages/cloudflare_logpush/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Fix ingestion failure for events collected via the HTTP Endpoint input by only parsing `event.original` when it is present. The input parses the request body into `json` itself, so `event.original` may be absent.
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/1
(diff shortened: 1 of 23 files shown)


brijesh-elastic (approved):

efd6 (approved):
Thanks

PR #19570, opened 2026-06-17
[AWS Cloudwatch OTEL] Add configs per policy

muthu-mps:
- Enhancement

## Proposed commit message

This PR adds all necessary configs per policy

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/aws_cloudwatch_input_otel/manifest.yml
+++ b/packages/aws_cloudwatch_input_otel/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.5.0
 name: aws_cloudwatch_input_otel
 title: "AWS CloudWatch OpenTelemetry Input Package"
-version: 0.2.0
+version: 0.2.1
 source:
   license: "Elastic-2.0"
 description: "Collect AWS CloudWatch metrics for selected AWS services (EC2, RDS, SQS, ELB, Lambda, Fargate) using the OpenTelemetry Collector awscloudwatchmetrics receiver."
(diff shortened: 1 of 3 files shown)


ishleenk17:
@muthu-mps : Pls share screenshot of how the final UI looks like of the package

muthu-mps:
> @muthu-mps : Pls share screenshot of how the final UI looks like of the package

Screenshot 2026-06-17 at 1 07 18 PM

Screenshot 2026-06-17 at 1 07 30 PM

ishleenk17:
@muthu-mps : Are the AWS key id and other credentials optional ?
I think not all would be optional right ?

packages/aws_cloudwatch_input_otel/manifest.yml
```diff
@@ -181,13 +158,43 @@ policy_templates:
...
         default: 10m
         required: false
         show_user: true
+      - name: autodiscover_aggregation
+        type: text
+        multi: true
+        title: Aggregation Statistics
+        description: One or more CloudWatch statistics applied to every discovered metric. Valid values are `Sum`, `Average`, `Maximum`, `Minimum`, and `SampleCount`. `Average` suits gauge-style metrics; `Sum` suits counters. Each additional statistic multiplies GetMetricData cost.
+        default:
+          - Sum
```
ishleenk17, line 198: We are keeping this default as Sum for all the metric types @muthu-mps ?
@tommyers-elastic : Should we have Sum, Average, Maximum as default ? 
As we see that is what we are using for quite a lot in the [test configurations](https://github.com/elastic/cloud-scenario-generator/blob/main/collector/otelcol-cloudwatch-es.yaml).

muthu-mps: Adding the stats based on whats needed for each service which is decided based on the stats that is applied in standard aws integrations.

muthu-mps:
> @muthu-mps : Are the AWS key id and other credentials optional ? I think not all would be optional right ?

Good question —  no, you generally should not make `access_key_id/secret_access_key`

The SDK resolves credentials in priority order, any one of which is sufficient:

- Static keys (access_key_id + secret_access_key, optionally session_token)
- Shared credentials profile
- role_arn assume-role (often the most secure, especially cross-account with external_id)
- EC2 instance profile via IMDS

For comparison, the standard **aws** package keeps `access_key_id/secret_access_key` optional for exactly these reasons.

ishleenk17 (approved):
Looks good!.
Some open questions will be addressed in next PR.

packages/aws_cloudwatch_input_otel/manifest.yml
```diff
@@ -273,13 +377,43 @@ policy_templates:
...
-        show_user: true
+        show_user: false
+      - name: autodiscover_aggregation
+        type: text
+        multi: true
+        title: Aggregation Statistics
+        description: CloudWatch statistics applied to every discovered metric for this namespace.
+        default:
+          - Average
+        required: false
```
tommyers-elastic, line 415: this should probably be required: true, but show_user: false, right? without this the config will fail to render properly. i know it probably makes no functional difference, but i think it's worth being correct that this is definitely required.

muthu-mps: As per the receiver its optional, but yes we need to make it required:true to render the metrics appropriatly. Let me note it and update this in next PR.

PR #19603, opened 2026-06-17
 [Fortinet Fortiproxy] URL Field Parsing Errors 

### LINKED ISSUE #15111: [Fortinet Fortiproxy]: Fortinet Fortiproxy Integration does not correctly parse URL
### Integration Name

Fortinet FortiProxy [fortinet_fortiproxy]

### Dataset Name

_No response_

### Integration Version

1.2.1

### Agent Version

8.18.1

### Agent Output Type

logstash

### Elasticsearch Version

8.18.1

n/A

### Software/API Version

_No response_

### Error Message

Processor uri_parts with tag process_url in pipeline logs-fortinet_fortiproxy.log-1.2.1 failed with message: unable to parse URI [/css?family=Roboto:300,400,500,700|Google+Sans:400,500,700|Google+Sans+Text:400,500,700&lang=de]

### Event Original

"<190>date=2025-09-01 time=12:37:12 devname=\"testdevice\" devid=\"FPX412345678\" eventtime=1756730232283394873 tz=\"+0000\" logid=\"1059028704\" type=\"utm\" subtype=\"app-ctrl\" eventtype=\"signature\" level=\"information\" vd=\"root\" appid=15893 srcip=10.10.10.1 srccountry=\"Reserved\" dstip=142.250.203.106 dstcountry=\"United States\" srcport=64986 dstport=443 srcintf=\"port2\" srcintfrole=\"lan\" dstintf=\"port1\" dstintfrole=\"wan\" proto=6 service=\"HTTPS\" direction=\"outgoing\" policyid=5 poluuid=\"f742ad16-b894-51ef-1128-ff401f6ee4ef\" policytype=\"policy\" sessionid=1516513751 applist=\"app_Default\" action=\"pass\" appcat=\"Web.Client\" app=\"HTTP.BROWSER\" hostname=\"fonts.googleapis.com\" incidentserialno=628964575 url=\"/css?family=Roboto:300,400,500,700|Google+Sans:400,500,700|Google+Sans+Text:400,500,700&lang=de\" msg=\"Web.Client: HTTP.BROWSER\" apprisk=\"medium\" rawdataid=\"1/1\" rawdata=\"Method=GET|User-Agent=Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36|Referer=https://something.com/\""

### What did you do?

The integration is installed with defaults and works fine. The issue is in the content that Fortiproxy sends in its logs and the parsing thereof.

I have not been able to identify a clear pattern what exactly triggers the failure (some of the events that match the filter above do not raise the error).
Most likely it is something in the query if the "url" field sent by Fortiproxy only contains the query, many of the failed documents contained a lot of special characters in the query.

To get around the error and get more consistent populated fields I have added a bugfix pipeline before the URI parts processor, however this does not preserve the original content of the url field and could be improved upon.

```
[
  {
    "script": {
      "source": "if (ctx._fields_.url != null && ctx._fields_.url.startsWith('http')) { ctx.protocol_present = true; } else { ctx.protocol_present = false; }"
    }
  },
  {
    "set": {
      "field": "_fields_.url",
      "value": "{{network.protocol}}://{{url.domain}}{{_fields_.url}}",
      "if": "ctx.protocol_present == false"
    }
  },
  {
    "remove": {
      "field": "protocol_present"
    }
  }
]
```

### What did you see?

Due to inconsistencies in Fortinet Fortiproxies logs the URI parts processor fails or provides inconsistent content in the url subfields.

The logs differ in the following ways:
- sometimes the "url" field in the original message is a "proper" url, e.g. https://www.google.com
- sometimes the "url" field in the original message is just the query but there is a "hostname" field in the logs.

The issue concerns the latter.

The results are: 
- URI processor failing with an error, pipeline is not processed completely
  - ctx._fields_.type == "utm" && ctx._fields_.subtype == "app-ctrl" && ctx._fields_.eventtype
- Pipeline processes correctly but url.domain is not populated.

Consistent behavior across all documents sent by Fortiproxy and no failures when not using any custom pipelines or special settings.

robester0403:
## Proposed commit message

Fix URI parsing failures when url field contains only a path or query string.

## Checklist

## Related issues

Resolves: #15111 

## Screenshots

image


packages/fortinet_fortiproxy/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.4.2"
+  changes:
+    - description: Fix URI parsing failures when url field contains only a path or query string.
+      type: bugfix
+      link: https://github.com/elastic/integrations/issues/19603
```
ilyannn: I believe this should be the PR link

packages/fortinet_fortiproxy/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -576,11 +576,55 @@ processors:
       field: client.ip
       if: ctx._fields_.clientip != null
 
+  - script:
```
ilyannn: I don't think the Painless script is necessary as you can also use the `set` processors to reconstruct the URL:

```
- set: temp.scheme <- url.scheme / network.protocol
- set: temp.url <- fields.url        # if already http(s)://
- set: temp.url <- {{scheme}}://{{url.domain}}{{fields.url}}
...
```

packages/fortinet_fortiproxy/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -576,11 +576,55 @@ processors:
...
+        if (ctx._temp_ == null) { ctx._temp_ = [:]; }
+        String lc = raw.toLowerCase();
+        if (lc.startsWith('http://') || lc.startsWith('https://')) {
+          ctx._temp_.url = raw;
+          return;
+        }
+        String scheme;
+        if (ctx.url?.scheme != null) { scheme = ctx.url.scheme.toString().toLowerCase(); }
+        else if (['http','https'].contains(ctx.network?.protocol)) { scheme = ctx.network.protocol; }
+        else if (ctx.server?.port instanceof Number && ctx.server.port == 443) { scheme = 'https'; }
```
ilyannn: I'm wary of special-casing 443 as HTTPS may very well be using a different port. Note there are no instances of `ctx.server.port == 443` string in the current integrations codebase.

Perhaps the `scheme` will in fact be always present in these cases so we don't need to guess?

packages/fortinet_fortiproxy/changelog.yml
ilyannn: Right now it does resolve to the same `/pull/19603`, but it would be better to have it explicitly go the the PR.

packages/fortinet_fortiproxy/data_stream/log/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: Agreed, while `443` should be a safe guess, it's not always guaranteed to be https. Also, this wouldn't work at all for other port numbers.

Are there any other fields we can key off of to get the protocol/scheme? If not, then we shouldn't try to guess. `url.full` should be the full URL, but if we don't have the full URL in the event, or can't reliable build it, then we can't populate this field.

taylor-swanson (approved):

Issue #19552, opened 2026-06-12
[Entra ID] Entity Analytics Integration - Invalid signInActivity Property


TomonoriSoejima:
## Problem
Entity Analytics integration requests sign-in timestamp fields that don't exist as direct properties, causing Microsoft Graph API to reject the request.

**Error:** `Request_UnsupportedQuery: Invalid property: lastSignInDateTime`  
**Status:** Integration enters DEGRADED state daily at 08:57 UTC  
**Tenant:** f62aca8c-2ba2-417b-a0c1-ab3f1020ccc4

## Root Cause
The integration incorrectly requests three individual properties:
- `lastSignInDateTime`
- `lastNonInteractiveSignInDateTime`
- `lastSuccessfulSignInDateTime`

**Microsoft Graph API doesn't recognize these as direct user properties.** They're nested inside a parent object called `signInActivity`.

### Current Configuration (WRONG)
```yaml
select:
  users:
    - lastSignInDateTime                    # ❌ Not a direct property
    - lastNonInteractiveSignInDateTime      # ❌ Not a direct property
    - lastSuccessfulSignInDateTime          # ❌ Not a direct property
```

### Required Configuration (CORRECT)
```yaml
select:
  users:
    - signInActivity  # ✅ Request parent object, timestamps are nested inside
```

## Customer Confirmation
Customer tested Microsoft Graph API directly and confirmed the correct response structure:

```json
{
  "id": "c795fb42-e371-455c-b4a8-xxxxxxx",
  "displayName": "test",
  "signInActivity": {
    "lastSignInDateTime": "2026-06-09T01:49:24Z",
    "lastNonInteractiveSignInDateTime": "2026-06-09T01:49:25Z",
    "lastSuccessfulSignInDateTime": "2026-06-09T01:49:25Z"
  }
}
```

The timestamp fields exist **inside** `signInActivity`, not at the top level.

## Required Fix

1. Request `signInActivity` object (not individual timestamp fields)
2. Parse the nested JSON structure
3. Extract timestamp values from within the `signInActivity` object

## Impact
- Feature completely broken for customers who enable "User Sign-in Activity"
- No workaround available (disabling feature loses all sign-in data)
- Requires Azure AD Premium P1/P2 license + AuditLog.Read.All permission (customer has both)

## References
- [Microsoft Graph signInActivity Resource](https://learn.microsoft.com/en-us/graph/api/resources/signinactivity?view=graph-rest-1.0)

[...]

PR #19672, opened 2026-06-19
EntityAnalytics Entra ID: fix last activity support

chemamartinez:
## Proposed commit message

```
[Entra ID] Fix device last sign-in time and remove broken user sign-in activity option

Include approximateLastSignInDateTime in the Intune device select query so
the field is actually collected when Intune Managed Device Properties is enabled.

Remove the User Sign-in Activity (user_sign_in_activity) toggle and its
associated select.users: [signInActivity] configuration. The signInActivity
property is explicitly excluded from delta query responses by the Microsoft
Graph API; it is listed under "Properties not returned by default" and noted
as not supported in $select on delta endpoints:
https://learn.microsoft.com/en-us/graph/api/resources/users?view=graph-rest-1.0

Emitting it via select.users had no effect; the field was never populated.
Proper support requires a separate non-delta enrichment call in the agent,
which will be re-introduced once the beats-side implementation ships.
```

## Checklist

## Related issues

- Relates https://github.com/elastic/integrations/issues/19552

### DIFF
--- a/packages/entityanalytics_entra_id/changelog.yml
+++ b/packages/entityanalytics_entra_id/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.10.2"
+  changes:
+    - description: Include approximateLastSignInDateTime in the device selection when Intune Managed Device Properties is enabled.
+      type: enhancement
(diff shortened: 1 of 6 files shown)


ShourieG (approved):
LGTM

PR #19636, opened 2026-06-18
[Backport pipeline] Add PR-driven backport branch creation

mrodm:
## Proposed commit message

```
Adds an automated, PR-driven workflow to create backport branches from `.backports.yml`.
When a PR that adds new entries to `.backports.yml` is merged into `main`, the CI pipeline
detects the new entries and triggers the `integrations-backport` pipeline to create the
corresponding branches automatically. A comment is then posted back to the originating PR
reporting success or failure.
```

WHAT:
Adds an automated, PR-driven workflow to create backport branches from `.backports.yml`.
When a PR that adds new entries to `.backports.yml` is merged into `main`, the CI pipeline
detects the new entries and triggers the `integrations-backport` pipeline to create the
corresponding branches automatically. A comment is then posted back to the originating PR
reporting success or failure.

Key components:
- `.backports.yml` inventory — source of truth listing backport branches with their
  package, base version, base commit, and lifecycle metadata (`archived`, `maintained_until`).
- `trigger_backport.sh` / `trigger_backport_lib.sh` — diff the old and new inventory on each
  push to `main`, validate the new inventory via `mage ValidateBackportsInventory`, and
  emit Buildkite trigger steps for any newly-added active entries.
- `pipeline.yml` — adds `check-backports-inventory` (schema validation on PRs),
  `trigger-backport-dryrun` (dry-run on PRs), and `trigger-backport-create` (real creation
  on merge to main) steps.
- `pipeline.backport.yml` — updated to accept `BACKPORT_BRANCH_NAME` via `get`
  (passed by the trigger step), restricts UI access to the `ecosystem` team, adds a
  `notify-pr` step that posts a success/failure comment back to the originating PR.
- `notify_backport_pr.sh` — posts a per-run comment on the PR.
- `AddBackportEntry` mage target — developer tool to add a new entry to `.backports.yml`,
  resolving the base commit via `dev/scripts/get_release_commit.sh` and inserting the entry
  in sorted order (package ascending, version descending).
- `AddEntry` in `dev/backports/inventory.go` — validates the package name against the
  known packages set before writing, preventing branches that would fail `branchRE`.
- Shell tests in `test_trigger_backport.sh` — unit-tests `generate_trigger_pipeline` with
  a per-target mage mock (`MOCK_MAGE_VALIDATE_EXIT` / `MOCK_MAGE_CHECK_EXIT`).

WHY:
Previously, backport branches had to be created manually by running `backport_branch.sh`
[...]

PR #19684, opened 2026-06-22
[CI] Preserve required packages when removing others during backport branch creation

mrodm:
## Proposed commit message

```
When `REMOVE_OTHER_PACKAGES` is enabled during backport branch creation,
composable packages may depend on other packages declared under
`requires.input` and `requires.content` in their `manifest.yml`. Previously,
all packages except the target were removed unconditionally, which would break
composable packages at testing time since their required base packages would no
longer be present.
```

**WHAT:**
- Introduced `backport_branch_lib.sh`, a new sourceable library containing three pure helper functions extracted from `backport_branch.sh`:
  - `get_package_path` — resolves a package name to its directory path
  - `get_required_package_names` — reads `requires.input[].package` and `requires.content[].package` from a package manifest, outputting nothing if the section is absent (safe for non-composable packages)
  - `remove_other_packages` — updated to accept a variadic list of paths to keep instead of a single path
- Updated `backport_branch.sh` to source the lib and, before calling `remove_other_packages`, resolve all required package paths from the target package's manifest and include them in the keep list.
- Added `test_backport_branch.sh` with 23 unit tests covering all three functions, including a full end-to-end composable package scenario using `nginx_integration_otel` (from elastic/integrations#18994) as a fixture.
- Registered the new test file in `run_buildkite_scripts_tests.sh`.
- Added `backport_branch_lib.sh` and `test_backport_branch.sh` to `non_package_patterns` in `common.sh` so changes to these files trigger CI.

**WHY:**
Composable packages like `nginx_integration_otel` declare dependencies on input and content packages. Removing those dependencies during backport branch creation would cause tests to fail. The lib extraction also makes the helpers independently testable without needing to stub out the entire `backport_branch.sh` execution environment.

## Author's Checklist

- [x] `bash .buildkite/scripts/run_buildkite_scripts_tests.sh` passes (23 tests, 0 failed)
- [x] Revert debug change in `.buildkite/pipeline.yml` that disables integration tests (commit `0e41ee6313`)
- [x] ~Validate default value for REMOVE_OTHER_PACKAGES is `true`. https://buildkite.com/elastic/integrations-backport/builds/286/~
    - Moved changes to #19636

Run the unit test suite:
```bash
bash .buildkite/scripts/run_buildkite_scripts_tests.sh
```

## Related issues

- Relates https://github.com/elastic/integrations/issues/17403
- Relates https://github.com/elastic/elastic-package/issues/3277
- Relates #19636 

---


.buildkite/scripts/backport_branch.sh
```diff
@@ -246,7 +212,23 @@ updateBackportBranchContents() {
...
+    while IFS= read -r req_name; do
+      local req_path
+      req_path=$(get_package_path "${req_name}" || true)
+      if [[ -n "${req_path}" ]]; then
+        echo "Keeping required package: ${req_path} (required by ${PACKAGE_NAME})"
+        packages_to_keep+=("${req_path}")
+      else
+        echo "Warning: required package '${req_name}' not found in packages folder"
+      fi
+    done < <(get_required_package_names "${PACKAGE_PATH}")
```
mrodm, line 229: As integration packages could define some other packages as dependencies, and they could be use in testing stages, I think those packages should also be kept in the backport branch.

.buildkite/pipeline.backport.yml
```diff
@@ -55,7 +55,7 @@ steps:
           value: "true"
         - label: "False"
           value: "false"
-      default: "false"
+      default: "true"
```
mrodm: Changing this, by default all other packages in the backport branch will be deleted (codeowner lines and package folders).

.buildkite/pipeline.yml
```diff
@@ -148,6 +148,8 @@ steps:
         allow_failure: false
       - step: "check-buildkite-scripts"
         allow_failure: false
+    if: |
+      build.branch == "nonexisting"
```
mrodm: To be removed before merging.

teresaromero (approved):
i am not sure about the ngnix composable landing prod, but it might was well serve as example for testing 👍🏻

mrodm:
Most of the failures in this build https://buildkite.com/elastic/integrations/builds/44853 were related to a temporary network error accessing docker registry.

Examples:

```
Error: booting up the stack failed: running docker-compose failed: running command failed: running Docker Compose up command failed: exit status 18: Head "https://registry-1.docker.io/v2/tianon/true/manifests/multiarch": Get "https://auth.docker.io/token?scope=repository%3Atianon%2Ftrue%3Apull&service=registry.docker.io": net/http: request canceled 
```

```
could not setup service: could not boot up service using Docker Compose: running Docker Compose up command failed: exit status 18: Head "https://registry-1.docker.io/v2/library/cassandra/manifests/3.11.11": Get "https://auth.docker.io/token?scope=repository%3Alibrary%2Fcassandra%3Apull&service=registry.docker.io": net/http: request canceled (Client.Timeout exceeded while awaiting headers)</error>
```

After retrying the failed ones, there are just pending `cursor` and `netbox` that are also failing in main branch (known issues).

.buildkite/pipeline.backport.yml
mrodm: Remove these changes from this PR https://github.com/elastic/integrations/pull/19684
in favor of moving those to https://github.com/elastic/integrations/pull/19636

jsoriano (approved):

PR #19689, opened 2026-06-22
Add alerts to anthropic metrics integration

ishleenk17:
Add alerting rule template to anthropic metrics integration.


packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-cache-hit-rate-drop.json
```diff
@@ -0,0 +1,40 @@
...
+        "ruleTypeId": ".es-query",
+        "schedule": {
+            "interval": "5m"
+        },
+        "params": {
+            "searchType": "esqlQuery",
+            "timeWindowSize": 1,
+            "timeWindowUnit": "h",
+            "esqlQuery": {
+                "esql": "// Alert when cache hit ratio drops below a threshold.\n// Requires minimum token volume (10000) to avoid noisy alerts on low-traffic periods.\n// Adjust both the 10000 volume threshold and the 0.30 ratio threshold as needed.\nFROM metrics-anthropic_metrics.usage-*\n| WHERE anthropic.usage.uncached_input_tokens IS NOT NULL\n| STATS uncached = SUM(anthropic.usage.uncached_input_tokens), cached = SUM(anthropic.usage.cached_input_tokens), cache_creation = SUM(anthropic.usage.cache_creation_input_tokens)\n| EVAL total_input = uncached + cached + cache_creation\n| WHERE total_input > 10000\n| EVAL cache_hit_ratio = cached * 1.0 / total_input\n| WHERE cache_hit_ratio < 0.30"
```
muthu-mps: The stats has no grouping (`BY`). Can we consider grouping dimensions?

muthu-mps: Is `10000` a good threshold to consider for Anthropic models? Since single request is often in 1000s.

packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-cost-anomaly.json
```diff
@@ -0,0 +1,40 @@
...
+        "ruleTypeId": ".es-query",
+        "schedule": {
+            "interval": "1h"
+        },
+        "params": {
+            "searchType": "esqlQuery",
+            "timeWindowSize": 24,
+            "timeWindowUnit": "h",
+            "esqlQuery": {
+                "esql": "// Alert when daily cost exceeds a threshold (in cents).\n// The amount field is in cents — 5000 = $50.00.\n// Adjust the threshold in the final WHERE clause as needed.\nFROM metrics-anthropic_metrics.cost-*\n| WHERE anthropic.cost.amount IS NOT NULL\n| STATS daily_cost_cents = SUM(anthropic.cost.amount)\n| EVAL daily_cost_usd = ROUND(daily_cost_cents / 100.0, 2)\n| WHERE daily_cost_cents > 5000"
```
muthu-mps: The stats has no grouping (`BY`). Can we consider grouping dimensions?

muthu-mps: The query sums amount across all currencies and divides by 100 assuming cents (USD). There is a separate currency field, If org bills by a different currency then the conversion would be wrong. Consider validating whether the currency field has `USD`

packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-cache-hit-rate-drop.json
```diff
@@ -0,0 +1,40 @@
...
+            },
+            "groupBy": "row",
+            "timeField": "@timestamp"
+        },
+        "alertDelay": {
+            "active": 3
+        },
+        "artifacts": {
+            "dashboards": [
```
muthu-mps, line 30: artifacts.investigation_guide is missing, Add investigation_guide. Same applies for all the rules.

packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-monthly-budget-spend.json
```diff
@@ -0,0 +1,41 @@
...
+        "ruleTypeId": ".es-query",
+        "schedule": {
+            "interval": "6h"
+        },
+        "params": {
+            "searchType": "esqlQuery",
+            "timeWindowSize": 30,
+            "timeWindowUnit": "d",
+            "esqlQuery": {
+                "esql": "// Monthly budget spend limit alert.\n// Sums all cost over the rolling 30-day window.\n// Amount is in cents — 100000 = $1,000.00.\n// Adjust the budget threshold in the final WHERE clause.\nFROM metrics-anthropic_metrics.cost-*\n| WHERE anthropic.cost.amount IS NOT NULL\n| STATS monthly_cost_cents = SUM(anthropic.cost.amount)\n| EVAL monthly_cost_usd = ROUND(monthly_cost_cents / 100.0, 2)\n| WHERE monthly_cost_cents > 100000"
```
muthu-mps: The stats has no grouping (`BY`). Can we consider grouping dimensions?

muthu-mps: same as above for currency validation.

packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-monthly-budget-spend.json
```diff
@@ -0,0 +1,41 @@
+{
+    "id": "anthropic_metrics-monthly-budget-spend",
+    "type": "alerting_rule_template",
+    "attributes": {
+        "name": "[Anthropic] Monthly Budget Spend Limit",
+        "description": "Alerts when cumulative spend over the rolling 30-day window exceeds a configurable budget. Default threshold is $1,000 (100000 cents). Adjust the threshold in the ES|QL WHERE clause to match your organization's monthly budget.",
```
muthu-mps: The description says rolling **30-day window**, which user might be in mid of the month. When we also mention, **Organizations monthly budget** which may intend to **calendar month** which resets on 1st of every month. If we are performing `rolling 30 day`, then we need to update the description only to have rolling 30 day window.

packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-cache-hit-rate-drop.json
ishleenk17: will add `investigation_guide` to all alert templates.

packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-monthly-budget-spend.json
ishleenk17: Will add the same `WHERE anthropic.cost.currency == "USD"` guard here as well.

packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-cache-hit-rate-drop.json
ishleenk17: The alert is intentionally org-wide and not grouped by model. Low cache rate on one tiny model isn't actionable. 
Per-model analysis is already there in the dashboard. Hence no group by used.

ishleenk17: The 10,000 threshold is a noise filter to avoid false positives during low-traffic periods. 

If your org only processed 50 input tokens in the last hour (say, one small test request), and none were cached, the cache hit ratio would be 0%, technically "below 30%" but not meaningful. You don't want an alert firing because someone ran a single uncached request during a quiet period.

10,000 tokens is roughly one or two real conversations with Claude. It means "only evaluate cache efficiency when there's enough traffic to make the ratio statistically meaningful."
I have mentioned in the comment that this can be changed.
But main use is to avoid noise of the alert. I am ok to make it 50K, but for me there is no rationale on whats the right value. Its purely user dependent.

packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-cost-anomaly.json
ishleenk17: Intentionally not grouped. This is org wide spike. 
Per-Workspace Cost Spike alert already exists for dimension-based alerting. The two are complementary.

ishleenk17: Agreed. Will add WHERE currency == "USD" guard to both cost alerts.
For now it only supports USD.

packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-monthly-budget-spend.json
ishleenk17: Same rationale, this is org wide and across all models.

packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-monthly-budget-spend.json
ishleenk17: Will update the description

packages/anthropic_metrics/kibana/alerting_rule_template/anthropic_metrics-cache-hit-rate-drop.json
muthu-mps: **500,000** would be ideal for 5m interval.

ishleenk17: @muthu-mps : This number basically means after what number of tokens, should the alert be fired for not caching at a good ratio. 5L might be too high for that. 
Actually to be frank, this number is very much dependant on the way the models are used. 
Hence I have mentioned in the comment that user should change it as per what they feel to be the right number. 
Keeping that in mind, I think 50K should be good for now.

muthu-mps (approved):
LGTM!

PR #19715, opened 2026-06-23
[tenable_io] -  Preserve `severity_level` filter across export cycles

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
tenable_io: preserve severity_level filter across export cycles

The vulnerability data stream collects findings through the Tenable vulns
export API using the CEL input. The severity filter was set on the initial
CEL state and the first export request, but the state-rebuild blocks for
status polling, chunk fetching, cancellation, and error paths did not carry
severity_level forward. Once the first export completed, severity_level was
dropped from the persisted state, so every subsequent export omitted the
filter and Tenable returned all severities, including info. Info-severity
findings continued to be ingested even when the policy excluded them.

Carry severity_level through every state-rebuild block, matching how the
other persistent state fields are propagated. Strengthen the vulnerability
system test to exercise at least two export cycles against a mock that
rejects exports missing the severity filter, so the regression is caught.

Changing the mock to return body:[] instead of body:null removes cel EOF errors during tests
```

## Note
The updated system test mocks are derived from pre-existing system test mocks and do not contain any customer centric data.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/tenable_io/_dev/deploy/docker/files/config.yml
```diff
@@ -47,11 +47,15 @@ rules:
...
   - path: /vulns/export/xxx/chunks/2
     methods: ["GET"]
     responses:
       - status_code: 200
-        body: null
+        body: |
+          []
```
kcreddy, line 58: Looks like unrelated change. Is it important for this fix?

packages/tenable_io/data_stream/vulnerability/_dev/test/system/test-default-config.yml
```diff
@@ -10,6 +10,11 @@ vars:
   enable_request_tracer: true
 data_stream:
   vars:
+    # Short interval so the test exercises at least two export cycles within
+    # the test window. This guards against regressions where the
+    # severity_level filter is applied only on the first export and dropped
+    # from CEL state on subsequent cycles (see SDH-7298).
```
kcreddy: I don't think we should put SDH numbers in public repos?

packages/tenable_io/_dev/deploy/docker/files/config.yml
kcreddy: Could you add to commit message if relevant?

packages/tenable_io/data_stream/vulnerability/_dev/test/system/test-default-config.yml
ShourieG: Removed this this no worries

packages/tenable_io/_dev/deploy/docker/files/config.yml
ShourieG: This is what allows the cycles to complete without throwing an error, cleanly. It's nice to have. Updated the proposed commit message to include this info.

kcreddy (approved):
LGTM, thanks!

PR #19718, opened 2026-06-23
[Atlassian Cloud] Initial version of Atlassian Cloud integration

haetamoudi:
## Proposed commit message

[Atlassian Cloud] Initial version of Atlassian Cloud integration

Please explain:

The reason this is a separate integration is because it covers all the products, against a cloud only API, while existing integrations focuses on a specific product and API's that are mainly covered on both on-prem and cloud for that product alone.

This allows cloud users to collect audit logs for all their Atlassian products that is running in cloud from a single location.
## Checklist

## Screenshots

127 0 0 1_5601_app_dashboards (3)

127 0 0 1_5601_app_dashboards (2)


packages/atlassian_cloud/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,112 @@
...
+                    :
+                      from_millis,
+                  },
+                  "want_more": next != "",
+                })
+              )
+            )
+          :
+            {
+              "events": {
```
qcorporation, line 84: If we get something other than a 200 OK , should events be a list[] vs an object. We see similar patterns for example in the checkpoint_harmony_endpoint: https://github.com/elastic/integrations/blob/5bae74ade1eb2895e4b7ab87d599683c1bc237ba/packages/checkpoint_harmony_endpoint/data_stream/threatemulation/agent/stream/cel.yml.hbs#L100

packages/atlassian_cloud/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,112 @@
...
+                  "events": items.map(e, {"message": e.encode_json()}),
+                  "cursor": {
+                    ?"token": (next != "") ?
+                      optional.of(next)
+                    :
+                      optional.none(),
+                    "last_timestamp": (size(items) > 0) ?
+                      string(int(timestamp(items[size(items) - 1].attributes.processedAt)) * 1000)
+                    :
+                      from_millis,
```
qcorporation: if we see no data come in after the poll interval, e.g. item = 0 and next = "" the start interval is set to `from_millis` (the start of the window - not the end of the window aka now). If this happens again, the same thing will occur and the scan gap grows larger. 

The more correct implementation is 

```
    "cursor": {
      ?"token": ...,
      "last_timestamp": (size(items) > 0) ?
        string(int(timestamp(items[size(items) - 1].attributes.processedAt)) * 1000)
      :
        (next != "") ?
          from_millis          // mid-pagination: hold position
        :
          string(int(now) * 1000),   // advance to current time when no events returned
```

packages/atlassian_cloud/data_stream/audit/agent/stream/cel.yml.hbs
haetamoudi: The object shape `("events": {"error": {...}})` gives us retry semantics. When the agent sees an error as a single object, it discards the cursor and retries from the last successful position on the next poll cycle. This is for HTTP errors like 429 (rate limit) or 500 (server error), which are transient and should be retried rather than skipped over.

The array shape (`"events": [{"error": {...}}])` would be for advance semantics, The error gets indexed as a document and the cursor moves forward. That's if we want to log errors and keep going, but for this API, I think we should retry and not miss any events.

qcorporation (commented):
@haetamoudi The changes look good, can we change the manifest to have an agentless deployment? Should we defer or can we have it in the initial release?

qcorporation (approved):
Lgtm

packages/atlassian_cloud/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,435 @@
...
+          user_login_failed:
+            category: [authentication]
+            type: [info]
+            outcome: failure
+          org_settings_update:
+            category: [iam, configuration]
+            type: [change]
+            outcome: success
+        prefixes:
+          - prefix: user_login_failed
```
vinit-chauhan, line 275: Is there any reason to have these exact match phrases in prefix match too?

haetamoudi: The exact entries handle the base action name (e.g., `user_login_failed`), while the prefix entries catch suffixed variants from the API (e.g., `user_login_failed_mfa`, `user_login_failed_captcha`). The prefix `user_login_failed` is ordered before `user_login` so that failure variants don't accidentally match `user_login` and get `outcome: success`.

vinit-chauhan (approved):

PR #19719, opened 2026-06-23
[elastic_package_registry] Migrate to input-package approach

teresaromero:
## Proposed commit message

Migrates the `elastic_package_registry` integration from the legacy `prometheus/metrics` input to the `prometheus_input` input-package pattern (v1.0.1), fixes ingest pipeline crashes on missing start-time metrics, and updates dashboards and field mappings for stack 9.4.4.

**WHAT:**

The `metrics` data stream now depends on `prometheus_input` via `requires.input` in the package manifest and `streams[].package` in the data stream manifest (`format_version` bumped to 3.6.5). `stream.yml.hbs` was reduced to only render integration-owned values (`metrics_path`, `tags`); collector configuration (`metricsets`, `use_types`, `period`, `hosts`, `metrics_filters`, `processors`, etc.) is merged from the input package template. Integration-specific defaults remain declared on the data stream manifest — notably `rate_counters: false`, default `metrics_filters.exclude` (`^go_*`, `^promhttp_*`), and `hosts` marked required.

A `dataset: elastic_package_registry.metrics` override preserves the existing Fleet dataset name instead of falling back to the input package default.

The ingest pipeline no longer throws when `process_start_time_seconds` is null or missing: rename and `date` processors are guarded with `if` conditions. Pipeline regression tests cover both null and absent start-time cases.

Field mappings were updated to add `.rate` gauge variants alongside existing `.counter` fields, and counter field types were corrected from `long` to `double` to match Prometheus collector output.

The legacy metrics overview dashboard was migrated to updated Kibana Lens format (`formBased` datasources, panel version 8.9.0) for 9.4.4 compatibility.

Docs were regenerated against the current `elastic-package create` template (ECS reference bumped from `v8.3.1` to `v9.4.0` in `_dev/build/build.yml`) and now mention `prometheus_input`, since `{{ inputDocs }}` does not yet render for `streams[].package`-style integrations.

Test coverage was extended: policy test fixtures (default + overrides), pipeline regression tests, and expanded system test assertions (HTTP, storage, and indexer metrics). Local Docker deployment supports multiple Package Registry versions (`v1.11.0` / `v1.39.0` via `variants.yml`). A new `entrypoint.sh` drives synthetic registry API traffic after `SIGHUP`, so system tests can assert on metrics like `storage_requests_total` and indexer histograms that only populate under real traffic. `_dev/test/config.yml` declares `prometheus_input` as a test dependency for system and policy tests.

Kibana version constraint tightened to `^9.4.4` (drops 8.x). Package version bumped to `0.4.0`.

**WHY:**

The input-package approach removes duplicated stream/agent configuration by delegating it to `prometheus_input`, keeping this integration focused on EPR-specific vars, field mappings, and ingest logic. The `format_version` bump is required for `requires.input`. The pipeline fix addresses a real ingestion failure found during manual testing against a local Package Registry instance (see screenshot). Dashboard and field-mapping updates align the package with stack 9.4.4.

## Checklist

## Author's Checklist

- [ ] System tests pass against a local stack with `elastic-package test system`

```bash
elastic-package build
elastic-package test
```

To exercise both Package Registry versions:

```bash
elastic-package test system --variant v1.11.0
elastic-package test system --variant v1.39.0
```

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/7499

## Screenshots

Manual testing against a local Package Registry instance ingesting metrics.
EPR Metrics Overview


packages/elastic_package_registry/_dev/build/docs/README.md
```diff
@@ -1,53 +1,86 @@
...
-`http://localhost:9000/metrics` on your package registry instance.
+For more information on architectures that can be used for scaling this integration, check the [Ingest Architectures](https://www.elastic.co/docs/manage-data/ingest/ingest-reference-architectures) documentation.
+
+## Reference
+
+### Inputs used
+
+This integration uses the `prometheus_input` package. The following input is used:
+
+{{ inputDocs }}
```
teresaromero, line 68: opening a followup issue under elastic-package https://github.com/elastic/elastic-package/issues/3696 to address the missing content here on the rendered readme

packages/elastic_package_registry/validation.yml
```diff
@@ -0,0 +1,4 @@
+errors:
+  exclude_checks:
+    - SVR00006 # ingest pipeline processors missing required tag; pipeline pre-dates this requirement.
+    - SVR00009 # ingest pipeline on_failure handler must set event.kind to "pipeline_error"; pipeline pre-dates this requirement.
```
mrodm: Could the pipeline be updated with these tags ?

packages/elastic_package_registry/manifest.yml
```diff
@@ -1,14 +1,18 @@
...
-version: 0.3.1
+version: 0.4.0
 description: "Collect metrics from a Elastic Package Registry instance"
 type: integration
 categories:
   - elastic_stack
+requires:
+  input:
+    - package: prometheus_input
+      version: "1.0.0"
```
mrodm: There is a `1.0.1` version


```suggestion
      version: "1.0.1"
```

packages/elastic_package_registry/docs/README.md
```diff
@@ -176,82 +212,26 @@ An example event for `metrics` looks as following:
...
         "type": "elastic_package_registry"
     }
 }
 ```
 
+
+### Data streams using ILM policies
+
+
+
```
mrodm, line 298: As this section is empty, should we remove it ?

packages/elastic_package_registry/docs/README.md
```diff
@@ -77,78 +109,82 @@ You can verify that metrics endpoint is enabled by making an HTTP request to
...
     "agent": {
-        "ephemeral_id": "fb0962b4-3f3f-48d6-81d6-3df63366f744",
-        "id": "97cba3e2-ea7d-4d80-aa69-75752faa1576",
-        "name": "docker-fleet-agent",
+        "ephemeral_id": "477e2cb5-928f-4e3a-b884-233bd26b8873",
+        "id": "d01a1c9f-e0c3-4a3e-89bf-c33f5b9de0b3",
+        "name": "elastic-agent-53593",
         "type": "metricbeat",
-        "version": "8.0.0"
+        "version": "9.4.2"
```
mrodm: It looks like that these sample events were created with 9.4.2 instead of 9.4.0. Maybe there is no problem, but it would be good to use the minimum stack supported.

packages/elastic_package_registry/data_stream/metrics/manifest.yml
```diff
@@ -28,22 +28,13 @@ streams:
...
-        default: []
-      - name: processors
-        type: yaml
-        title: Processors
-        multi: false
-        required: false
+        title: Dataset name
+        description: Name of the dataset. Determines the index where data is written (metrics-<dataset>-*).
+        default: elastic_package_registry.metrics
+        required: true
```
mrodm: Is it required to set this dataset variable for now on?

Usually integration packages do not require to define the dataset.

mrodm: Just tested locally, and if it is not defined documents are ingested using another dataset `data_stream.dataset: prometheus`.

mrodm: In the nginx composiable PR (closed) the dataset was set out of the variables:

https://github.com/elastic/integrations/pull/18994/changes#diff-83e338f96ff6b89c27d5e359d1c61d94888309af9a8ce8b9b97af72ba0adda7bR3

mrodm (commented):
Not sure about the dataset being defined as a variable now.

In the nginx composable package (closed PR) it was set using the `dataset` field in the manifest.

packages/elastic_package_registry/docs/README.md
teresaromero: this is not on the template but it appears to be added when building the final readme. 

it was introduced here https://github.com/elastic/elastic-package/pull/3221

i wonder if this should only appear if there is actual payload on the content of the seccion cc @jrmolin @kcreddy 

i see some packages already published with this on their READMES (anthropic and cursor)

kcreddy: Looks like [`internal/docs/ilm.go`](https://github.com/elastic/elastic-package/blob/e77ff5fd1a8bfe46352cc9cddbdc06b6f38ea888/internal/docs/ilm.go#L113) is missing check on empty policies unlike [`internal/docs/transforms.go`](https://github.com/elastic/elastic-package/blob/e77ff5fd1a8bfe46352cc9cddbdc06b6f38ea888/internal/docs/transforms.go#L52-L54)

teresaromero: i will open an issue to address this thanks

packages/elastic_package_registry/validation.yml
teresaromero: yes! thanks for the heads up

packages/elastic_package_registry/data_stream/metrics/manifest.yml
teresaromero: i am investigating this point.

teresaromero: @mrodm  i'v opened a fix on elatic-package https://github.com/elastic/elastic-package/pull/3713 so the template uses the meta field of the dataset which is including the correct value... instead of showing any variable. I've tested with the package and will run an integration test to check it does not break current integrations. this is package-scoped so i would say is a safe change.

teresaromero: after re-reviewing the aproach i found out that the prometheus input was importing the variable and the template was actually using the variable; so when fleet merged the policy, the variable value was used at the policy.

there is no current option to "ignore" template content (which could be an option); also for integrations we dont want to open the variable option to users to override the dataset. This fixes are co-dependnat on each other;
- https://github.com/elastic/kibana/pull/275312
- https://github.com/elastic/elastic-package/pull/3719


I am drafting this PR until this gets resolved

teresaromero:
converting this intro draft until this are cleared https://github.com/elastic/integrations/pull/19719#discussion_r3490475041

packages/elastic_package_registry/_dev/deploy/variants.yml
```diff
@@ -0,0 +1,6 @@
+variants:
+  v1.11.0:
+    SERVICE_VERSION: v1.11.0
+  v1.38.0:
+    SERVICE_VERSION: v1.38.0
```
mrodm: Just to take advantage of this PR to set the latest EPR version published v1.39.0


```suggestion
  v1.39.0:
    SERVICE_VERSION: v1.39.0
```

packages/elastic_package_registry/data_stream/metrics/manifest.yml
```diff
@@ -1,10 +1,14 @@
 title: "Prometheus Metrics"
 type: metrics
+# dataset is the default form. this overrides the dataset set on the required input.
+# if not defined, fleet overrides the dataset with the default (packag_name.stream_type)
```
mrodm: Nit.
```suggestion
# if not defined, fleet overrides the dataset with the default (package_name.stream_type)
```

packages/elastic_package_registry/data_stream/metrics/fields/fields.yml
```diff
@@ -37,16 +37,28 @@
       description: Elastic Package Registry version.
       dimension: true
 - name: package_registry.storage_requests_total.counter
-  type: long
+  type: double
```
mrodm, line 40: Could these changes in mapping types cause conflicts if there is existing data ingested ? 

Should it be added a changelog entry of bugfix type for that too ?

packages/elastic_package_registry/changelog.yml
```diff
@@ -1,4 +1,10 @@
 # newer versions go on top
+
+- version: "0.4.0"
+  changes:
+    - description: Update package to use required input prometheus_input
+      type: enhancement
```
mrodm, line 6: Not sure what to do about this changelog type. As it is dropping support for users using stack < 9.4.4, should it be considered a breaking change ? 
What type are using other packages when they update the kibana version ?

packages/elastic_package_registry/data_stream/metrics/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -2,79 +2,104 @@
...
   - rename:
+      tag: rename_tmp_process_start_time_seconds_value_to_package_registry_start_time_seconds_b88d8934
       field: tmp.process_start_time_seconds.value
       target_field: package_registry.start_time_seconds
       ignore_missing: true
+  - remove:
+      tag: remove_null_package_registry_start_time_seconds_c1f2a9e0
+      field: package_registry.start_time_seconds
+      if: ctx.package_registry?.start_time_seconds == null
+      ignore_missing: true
```
mrodm: Probably, this step could be removed if the `if` condition is moved to the `rename` processor:

```yaml
  - rename:
      tag: rename_tmp_process_start_time_seconds_value_to_package_registry_start_time_seconds_b88d8934
      field: tmp.process_start_time_seconds.value
      target_field: package_registry.start_time_seconds
      ignore_missing: true
      if: ctx.tmp?.process_start_time_seconds?.value != null
``` 

The `if` added to the `date` processor, I think it could be kept just as safeguard.

packages/elastic_package_registry/data_stream/metrics/_dev/test/policy/test-default.yml
mrodm, line 1: Could you add another policy test adding some of the variables that were removed from the Agent template (metrics_filter.include, processors...)? 
That would ensure that this package version would support the same variables (it should).

packages/elastic_package_registry/data_stream/metrics/_dev/test/policy/test-default.expected
```diff
@@ -0,0 +1,43 @@
...
+          metrics_filters.exclude:
+            - ^go_*
+            - ^promhttp_*
+          metrics_filters.include: null
+          metrics_path: /metrics
+          metricsets:
+            - collector
+          password: null
+          period: 30s
+          rate_counters: false
```
mrodm, line 22: Why this value is `false` ? 

Default value should be true according to the 1.0.1 version of the `prometheus_input` 

https://github.com/elastic/integrations/blob/7601d01dccadc36768f43b5aca329832cbbcf805/packages/prometheus_input/manifest.yml#L58

At least, in the UI is set to `true`

Image

mrodm: Ok, I see why they appear.
They are defined statically in the `stream.yml.hbs`, but the UI shows those variables even if the template does not allow to change those values. It looks like they could be defined/updated but actually those policy changes will not do any effect.

Should Fleet take somehow that into account ? Would that be possible ?

Options from the package side that I could think of:
- we could add those variables in the datastream manifest with the default values
    - Disadvantage: those variables are not anymore in sync with the input package
- Or, update the template to use the same as the input package (removing those static values)
    - For this scenario, I cannot recall why it was decided those static values for use_types and rate_counters.
    - The value for `/metrics` was hardcoded because it is the URI for package-registry service.
    - I don't see that variable in the `prometheus_input` package
    - It is also used by other packages: [Example](https://github.com/elastic/integrations/blob/7601d01dccadc36768f43b5aca329832cbbcf805/packages/grafana/data_stream/metrics/agent/stream/stream.yml.hbs#L7)


Docs found for `prometheus` collector 

https://github.com/elastic/beats/blob/0f86bf11c6a8626521d2983e7aa092eb94ea8731/metricbeat/module/prometheus/collector/_meta/docs.md

mrodm (commented):
It would be good to check that the dashboards are still working with all these changes by performing different requests to the search and package endpoints for different packages. Those requests will allow us to ingest different data so dashboards show more data in the figures.

mrodm:
I think it would be good to update also the ecs reference used in `_dev/build/build.yml` to use one of the latest tags instead of `v8.3.1`.
Currently:

https://github.com/elastic/integrations/blob/7601d01dccadc36768f43b5aca329832cbbcf805/packages/elastic_package_registry/_dev/build/build.yml#L1-L3

I think it could be updated up to `v9.4.0` since the kibana constraint is now `^9.4.4`

https://github.com/elastic/ecs/tags

To be checked if that causes that some field mappings or descriptions are updated.

packages/elastic_package_registry/data_stream/metrics/_dev/test/policy/test-default.expected
teresaromero: >Why this value is false ?

The stream.yml.hbs are the "custom" values the original input had. When importing the prometheus_input some values are exposed as variables. Using the "local" template value was a way to override this value. 

I don't know how we could/should order the composable to remove some fields from the incoming input. There are other fields on the input that where not used/exposed in the original input. Ideally i think we should provide a way for the integration to opt-out variables or options when bundling. For now the solution i thought was to have a local template with overrides, but its true this opens the possibility of variables (incoming from input) on the UX that make no sense as the values are not used on the template per-se.


I am afraid this is an implementation gap we should address.

teresaromero: >I don't see that variable in the prometheus_input package

Yes, that is why i hardcoded and kept it as an override. Should we move this into a variable with default value? but still we need an override because the endpoint needs this, does it?

teresaromero:
>It would be good to check that the dashboards are still working with all these changes by performing different requests to the search and package endpoints for different packages. Those requests will allow us to ingest different data so dashboards show more data in the figures.

I've added the dashboard fields into the system test with default values, the test fixture performs some queries to ensure we get the fields. Moreover, i've run the test locally to check the dashboards and the data is populated too

teresaromero:
>To be checked if that causes that some field mappings or descriptions are updated.

Updated, this caused a description change on the readme, other than that after test i believe it works as expected

packages/elastic_package_registry/data_stream/metrics/_dev/test/policy/test-default.expected
teresaromero: After considering the options:
- for new fields that where hardcoded on the template, i am still adding them to the policy via the local template, but exposing the values through a variable with the preferred default value. This have show_user: false so it steps under the advanced config.
- for fields on the input template that where hardcoded on the integration, i am adding them on the integration datastream variables, changing the default value to the preferred (hardcoded) one

I find this solution convenient moving forward and it aligns with the fact that variables can be "promoted" on the integration overriding the source (input) value.

Going with the "remove variables from input" seems like a big decision which i am not sure its the required one, as inputs have a certain configuration that needs to be fulfilled.

packages/elastic_package_registry/data_stream/metrics/fields/fields.yml
teresaromero: I am adding a changelog entry for this. Related to the risk, after reading about it:

> For long → double (your actual change): low risk
> Counter values in practice are whole numbers (0, 27, etc.), as seen in sample data and pipeline tests. Those index fine into both long and double mappings.
> 
> Ingest failures are unlikely on the current write index, since integer counter values are compatible with an existing long mapping.
> After rollover, new indices will use double. Old indices still have long. Elasticsearch generally handles long/double coexistence in queries, though field capabilities may show both types across backing indices.
> Dashboards aggregating counters should continue to work.

I believe it will be a low risk.

packages/elastic_package_registry/changelog.yml
teresaromero: other packages usually use enhacement and upgrade the minor. i am keeping this update and adding a dedicated one for the stack version change (enhacement too) and a bugfix for the fields change.

eitherway this should be an update on the approach how to build the integration, but functionality is kept as is...

packages/elastic_package_registry/data_stream/metrics/manifest.yml
```diff
@@ -43,10 +47,31 @@ streams:
...
+        multi: false
+        show_user: false
+        default: false
+      - name: ssl
+        type: yaml
+        title: SSL Configuration
+        description: i.e. certificate_authorities, supported_protocols, verification_mode, key, passphrase etc.
+        multi: false
+        required: false
+        show_user: false
```
mrodm: `ssl` field was not used previously in the agent template, why is it needed now?

Should this be changed to be the `use_types` field instead that was set as a hardcoded `true` value previously in the agent template?

teresaromero: ssl field has a default value on the input that i am not sure is the one to use here too... so this is intended to overwrite the default value for ssl 

use_types was removed to inherit from input, as default value is true so it does not need an override

mrodm: The default value of the `ssl` variable in the `prometheus_input` package is just a YAML string that it is everything commented out.
https://github.com/elastic/integrations/blob/4acfa9aa42232cfd183d14e4e24e72aee5caa843/packages/prometheus_input/manifest.yml#L76-L101

So, IIUC then it is the same behaviour. It would be just added a some commented lines in the final agent template. Do you think it can be removed then ?

mrodm (approved):
LGTM
As a first package using the input-package dependencies approach.
It should be taken into account what it was highlighted here about the variables that could be set with hardcoded values in the agent template. They could appear in the UI as available to be modified, but those changes will not have any effect in the agent.

teresaromero:
>It should be taken into account what it was highlighted here about the variables that could be set with hardcoded values in the agent template. They could appear in the UI as available to be modified, but those changes will not have any effect in the agent.

I think we need to add a "how-to" guide on migration of packages into the composable aproach; i will open an issue for this

PR #19758, opened 2026-06-25
fix(otel-rum-dashboards): add selected option to `null` for ESQL controls

david-luna:
## Proposed commit message

Add `null` value to selected option for ESQL controls as default.

The OTEL RUM dashboard display 3 `esql_control`s which options depend on the data stored by the user in ES. These are
- the names of the services observed in the time frame
- the names of the browsers which accessed these services (depends on service names)
- the names of the operating systems which accessed these services (depends on service names)

Since is not possible to know such values beforehand. The default options of the control were set to an empty list. This made the panels fail but controls were able to load data and, once all controls had an option selected, panels started to show data.

In https://github.com/elastic/elasticsearch/pull/147748 the validation became more strict and since all queries from the dashboard components, even the ones that do not use esql vars,  are sending the variable values (empty lists) we got an error that we cannot recover from.

This change adds selected options to `null` for all controls which is allowed by ESQL.

## Checklist

## Author's Checklist

- [x] test it on stack version 9.5


trentm:
IIUC, longer term what you want for the dashboard here is "ANY" support: https://github.com/elastic/elasticsearch/issues/136735

packages/otel_rum_dashboards/changelog.yml
```diff
@@ -1,4 +1,10 @@
 # newer versions go on top
+- version: "0.0.5"
+  changes:
+    - description: Set ESQL controls selecte options to null
```
trentm: nit: Perhaps give the reason:
```suggestion
    - description: Set ESQL controls selected_options to null so it is interpreted as "nothing selected" in all ES versions.
```

trentm (approved):

Issue #19407, opened 2026-06-05
[New Integration] Claude Code and Cowork


jamiehynds:
### Description

Unlike the Claude Enterprise web product, **Cowork and Claude Code activity is explicitly excluded from the Compliance API** — confirmed across all plan tiers including Enterprise (as of May 2026, Cowork is excluded from Audit Logs, the Compliance API, and Data Exports; Claude Code is on a separate path entirely).

The supported route for session-level telemetry from these products is OpenTelemetry (OTel): both Claude Code and Cowork emit events via standard OTLP export. Elastic's InfoSec team has already built and documented a working pipeline for this (see the Security Labs post in references). This issue is about formalising that work as a native, installable integration rather than leaving it as a manual configuration exercise for every customer.

This is a meaningful security surface: agentic workflows in Claude Code and Cowork may involve shell execution, file-system access, MCP tool calls, and web browsing — all high-value signals for threat detection and compliance that are invisible to the Compliance API.

### What this data does (and doesn't) give you

OTel is the *only* telemetry path for these products, so the integration's value is gated by what's exported and at what verbosity:

**What the events expose:**
- **Tool execution** — every tool call and MCP server call (`tool_result`): `tool_name`, `tool_use_id`, `success`, `duration_ms`, `error_type`, and the permission `decision_source`.
- **Permission decisions** (`tool_decision`) — how each call was approved/rejected and the source (`config`, `hook`, `user_permanent`, `user_temporary`, `user_abort`, `user_reject`).
- **API usage & cost** (`api_request`) — `model`, `cost_usd`, token counts (input/output/cache), `duration_ms`, MCP attribution.
- **Session structure** — `session.id` + `event.sequence` (monotonic per-session counter) + `prompt.id` (links every event triggered by one prompt) enable full session reconstruction.
- **Identity/context** — `user.email`, `user.id`, `organization.id`, `terminal.type` on every event.

**What's redacted/gated by default (important):**
- **Prompt content** — redacted unless `OTEL_LOG_USER_PROMPTS=1`.
- **Tool parameters & input** (bash commands, MCP server/tool names, skill names, tool args) — redacted unless **`OTEL_LOG_TOOL_DETAILS=1`** (a *separate* gate from prompts). MCP server/tool visibility depends on this being set.
- **Raw API request/response bodies** — only with `OTEL_LOG_RAW_API_BODIES` (full conversation history; heavy).
[...]

PR #19765, opened 2026-06-26
claude_code: add Claude Code OTel integration

efd6:
## Proposed commit message
```
claude_code: add Claude Code OTel integration

Collect telemetry from Claude Code CLI sessions via OpenTelemetry.
The package includes an ingest pipeline that flattens OTel attributes
to root-level fields, enriches ECS fields (event.action, event.outcome,
related.user, related.hosts), and six Kibana dashboards covering
session timeline, security overview, cost/usage, MCP server access,
tool calls, and permission decisions.

Test samples were collected from live sessions and sanitised.
```

> [!NOTE]
> Apologies for the size. The OTel system tests need a mock server and the inputs to all pipeline tests are JSON, rather than lines.

Also, ref elastic/elastic-package#3711.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- For #19407

## Screenshots

Screenshot from 2026-06-26 13-36-17


packages/claude_code/data_stream/events/fields/fields.yml
```diff
@@ -0,0 +1,277 @@
...
+      description: Whether safe mode is active ("true" or "false").
+    - name: command_name
+      type: keyword
+      description: Slash command name from user_prompt events.
+    # Marketplace
+    - name: marketplace.name
+      type: keyword
+      description: Name of the marketplace entry.
+    # OTel instrumentation scope
+    - name: scope.name
```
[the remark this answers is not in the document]
efd6: This is half right. The `claude_code.events.scope` fields should be removed, but the `scope` fields are defined due to OTel.

packages/claude_code/data_stream/events/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,454 @@
...
+                handleMap(v);
+            } else if (v instanceof List) {
+                handleList(v);
+            }
+            return v == null || v == '' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0)
+            });
+        }
+        handleMap(ctx);
+
+  - set:
```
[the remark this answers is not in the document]
efd6: This is not correct. `@custom` calls are injected by fleet at install time.

packages/claude_code/_dev/build/docs/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+```bash
+export OTEL_RESOURCE_ATTRIBUTES="data_stream.dataset=claude_code.events"
+```
+
+### Option B: EDOT Collector
+
+Run the [Elastic Distribution of the OpenTelemetry Collector](https://github.com/elastic/elastic-agent) with an `otlp` receiver and an `elasticsearch` exporter. Configure the `data_stream.dataset` resource attribute as above. The collector routes events to `logs-claude_code.events.otel-*`.
+
+### Option C: Managed OTLP (mOTLP)
```
jamiehynds: @efd6 in the interest of giving users an opinionated "best" option, could mOTLP be positioned as the recommended approach, since it avoids users having to deploy/manage agents? Regarding the user flow for this option - do they need to install the integration assets in Fleet first (for the mappings/pipeline/dashboards) before pointing Claude Code at the mOTLP endpoint?

efd6: > do they need to install the integration assets in Fleet first (for the mappings/pipeline/dashboards) before pointing Claude Code at the mOTLP endpoint?

Yes, they do. Without this there would be no template, ingest pipeline or dashboard and so the data would go to a generic OTel data stream. I have updated the README to include this requirement and made mOTLP be the recommended option.

teresaromero (approved):

packages/claude_code/data_stream/events/manifest.yml
```diff
@@ -0,0 +1,17 @@
+title: Claude Code events
+type: logs
+streams:
+  - input: otelcol
```
kcreddy, line 4: I see all of the packages with `input: otelcol` are suffixed with `_otel`. Do you know if there a standard convention that we should also do it?

packages/claude_code/data_stream/events/fields/ecs.yml
```diff
@@ -0,0 +1,38 @@
+- external: ecs
```
kcreddy, line 1: I see in the manifest.yml, the `type: logs`. Since `ecs@mappings` is applicable for `logs-*` index pattern, I'm curious what pattern these end up in, and if we require explicit ECS definitions.

packages/claude_code/data_stream/events/sample_event.json
```diff
@@ -0,0 +1,101 @@
...
+    },
+    "host": {
+        "arch": "amd64",
+        "architecture": "amd64",
+        "os": {
+            "platform": "linux",
+            "version": "6.17.0-14-generic"
+        }
+    },
+    "observed_timestamp": "2026-06-30T02:43:31.836690337Z",
```
kcreddy, line 66: I don't see this field `observed_timestamp` inside the definitions, nor in ECS. Should we define it and add a `date` processor to the pipeline? or is it given that upstream always sends it in ISO 8601 format?

packages/claude_code/data_stream/events/sample_event.json
```diff
@@ -0,0 +1,101 @@
...
+    },
+    "scope": {
+        "name": "com.anthropic.claude_code.events",
+        "version": "2.1.175"
+    },
+    "service": {
+        "name": "claude-code",
+        "version": "2.1.175"
+    },
+    "tags": "preserve_original_event"
```
kcreddy, line 100: Shouldn't this be an array?

packages/claude_code/_dev/build/docs/README.md
```diff
@@ -0,0 +1,162 @@
...
+| `mcp_server_connection` | MCP server connection attempt (status, transport type). | `network` |
+| `hook_registered` | Hook registration (name, event type, matcher). | `process` |
+| `hook_execution_start` | Hook execution start. | `process` |
+| `hook_execution_complete` | Hook execution result (success/failure counts, duration). | `process` |
+| `plugin_loaded` | Plugin loaded (name, scope, paths). | `package` |
+| `skill_activated` | Skill activation (name, source, trigger). | `process` |
+
+## What do I need to use this integration?
+
+- An Elastic deployment running version 9.4.0 or later.
```
kcreddy, line 43: Do we need to call this out? It might get stale over time.

packages/claude_code/manifest.yml
```diff
@@ -0,0 +1,56 @@
+format_version: "3.6.0"
+name: claude_code
+title: Claude Code
+version: "0.1.0"
+source:
+  license: "Elastic-2.0"
+description: Collect Claude Code agentic telemetry via OpenTelemetry.
+type: integration
+categories:
+  - security
```
kcreddy, line 10: There is also another category: `opentelemetry`. Is it applicable here?

packages/claude_code/img/dashboard-permission-decisions.png
kcreddy, line 1: Since this is `9.4+`, maybe we could use collapsible sections to divide the dashboard. Suggestion not a blocker.

packages/claude_code/img/dashboard-cost-usage.png
kcreddy, line 1: I think the newer dashboard guidelines say the metrics should be centred instead of on the bottom-right.

Also another suggestion from dashboard guidelines:
> Omit the panel title since metrics already have their own title

packages/claude_code/data_stream/events/sample_event.json
efd6: It should, but this is what comes back. If there is another item added, it does come as an array. I suspect this is an otel thing.

packages/claude_code/data_stream/events/sample_event.json
efd6: This is an otel thing.

packages/claude_code/data_stream/events/manifest.yml
efd6: The existing _otel packages (httpcheck, sql_server_input_otel, mysql_input_otel, zipkin_input_otel) are all input packages (type: input) that have non-OTel counterparts. The suffix distinguishes the OTel variant, but that's not the case here.

packages/claude_code/_dev/build/docs/README.md
efd6: For the moment, I think it is more helpful than harmful.

packages/claude_code/data_stream/events/fields/ecs.yml
efd6: I think it's helpful here to indicate what fields we are populating.

kcreddy (approved):
LGTM

PR #19780, opened 2026-06-26
[Google Workspace] Reduce default lag time for data streams

### LINKED ISSUE #19072: [google_workspace]: Reduce default lag time and request interval for data streams
### Integration Name

Google Workspace [packages/google_workspace]

### Dataset Name

google_workspace.*

### Integration Version

3.3.1

### Agent Version

9.1.4

macOS

### User Goal

I'm on the TRaDE team, we create default detection rules for several data streams within this integration/package. We'd like the default lag times to be something closer to 3-5 min. This will allow our out of the box detections to run with a 5 min lookback window instead of the 130 min lookback window they currently use to account for the lag time. We should also consider reducing the default interval polling time as well. It's set to 15 min for some logs whereas other cloud integration like AWS Cloudtrail will poll for new S3 log objects every 1 min by default. 

### Existing Features

Currently some lag time defaults are set to 1-3 hours, even though Google docs shows that most of the logs we ingest are available [near real time](https://knowledge.workspace.google.com/admin/reports/data-retention-and-lag-times?visit_id=639147991739047986-1287413728&rd=1#lag_times). There does not seem to be a need for the default lag time to be this long. This is especially true for Filebeat where it seems all the log streams default to 2 hours. 

Image

### What did you see?

We reduced the lag times to 3 min for all the logs and we recieved expected data within 4-5 min for the log sources we tested. 

Image

The alerts logs also seem to come in within minutes, and as Security alerts these should certainly not have a 2 hour lag time when the data is available nearly immediately. 

Image

moxarth-rathod:
## Proposed commit message

```
google_workspace: reduce default lag time for select data streams

Reduce default lag time on login (3h to 3m), gcp (3h to 5m), user_accounts (1h to 10m), 
groups (2h to 30m), group_enterprise (2h to 30m), and calendar (2h to 30m) data streams to 
align with Google documented reporting delays. Leave alert interval, token lag time, and 
gmail interval unchanged.

Existing Fleet policies retain their configured values; only new policy receive the updated defaults.
```

## Checklist

## Related issues

- Closes #19072

### DIFF
--- a/packages/google_workspace/changelog.yml
+++ b/packages/google_workspace/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.5.0"
+  changes:
+    - description: Reduce default lag times for login, gcp, user_accounts, groups, group_enterprise, and calendar data streams to align with Google documented reporting delays.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
 - version: "3.4.1"
   changes:
     - description: Use the ECS definition for email.attachments.
--- a/packages/google_workspace/changelog.yml
+++ b/packages/google_workspace/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Reduce default lag times for login, gcp, user_accounts, groups, group_enterprise, and calendar data streams to align with Google documented reporting delays.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/1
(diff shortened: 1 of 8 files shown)


andrewkroh (approved):

imays11:
Does this mean udpating the integration will or will not update the default lag times in existing policies?

PR #19800, opened 2026-06-26
[buildkite/common] Extract non_package_patterns to a dedicated file with tests

mrodm:
## Proposed commit message

Extract the `non_package_patterns` list out of the hardcoded array inside
`is_pr_affected()` in `common.sh` and move it to a dedicated
`.buildkite/scripts/non_package_patterns.txt` file (one ERE pattern per line,
comments and blank lines ignored).

**WHY:**
Previously, any update to the patterns list required touching `common.sh`.
Because `common.sh` is not itself listed in the patterns, CI treats any change
to it as an infra change and runs tests for *all* packages — even when only
adding a new script entry to the list. With this change, the pattern list lives
in its own `.txt` file, which is self-listed, so editing it alone is classified
as a non-package change and does not trigger all-package test runs.

A secondary benefit is avoiding the bash `MAX_ARG_STRLEN` (~128 KB) limit.
The previous approach joined all patterns into a single `-E` regex argument
passed to `grep`. As the list grows, that argument could eventually exceed
the per-argument exec limit. Loading patterns via `grep -f` reads from a
file and is immune to this limit.

**WHAT:**
- Extract `pr_has_package_related_files()` as a standalone, testable function
  that reads file paths from stdin and returns 0 if any path is not covered
  by a non-package pattern (i.e. package tests should run).
- Replace the hardcoded array + regex join with:
  `grep -Evf <(grep -Ev '^[[:space:]]*(#|$)' "${SCRIPTS_BUILDKITE_PATH}/non_package_patterns.txt")`
- Add `.buildkite/scripts/non_package_patterns.txt` (self-listed so edits to
  the file itself don't trigger all-package test runs).
- Add `test_non_package_patterns.sh` with 53 unit tests covering:
  - All known non-package paths (must NOT trigger tests)
  - Infra-triggering paths (must trigger tests)
  - Mixed changesets
  - `^` anchoring edge cases (e.g. `vendor/docs/foo` must not match `^docs/`)
- Register the new test suite in `run_buildkite_scripts_tests.sh`.

## Author's Checklist

- [x] `test_non_package_patterns.sh` passes locally (`bash .buildkite/scripts/test_non_package_patterns.sh`)
- [x] All patterns from the old hardcoded array are present in `non_package_patterns.txt`
- [x] `non_package_patterns.txt` is self-listed so editing it alone is a non-package change
- [x] Test that updating .buildkite/pipeline.yml trigger all integration tests.
- [x] Test that updating just one package , trigger that specific integration test.

Run the unit tests directly from the repo root:

```bash
bash .buildkite/scripts/test_non_package_patterns.sh
# Expected: 53 passed, 0 failed
```

To manually verify the classification, source `common.sh` in a subshell
(it sets `set -euo pipefail`, so wrap each call in `if` to safely capture
the result):

```bash
# From repo root
source .buildkite/scripts/common.sh

# Should NOT trigger tests — path is covered by a pattern (expected: "non-package")
if echo "docs/foo.md" | pr_has_package_related_files; then
    echo "result: package change (unexpected)"
else
    echo "result: non-package change (expected)"
fi

# Should trigger tests — path is not covered (expected: "package change")
if echo ".buildkite/pipeline.yml" | pr_has_package_related_files; then
    echo "result: package change (expected)"
else
    echo "result: non-package change (unexpected)"
fi

# Mixed: one uncovered path among covered ones → should trigger tests (expected: "package change")
if printf '%s\n' "README.md" ".buildkite/pipeline.yml" | pr_has_package_related_files; then
    echo "result: package change (expected)"
else
    echo "result: non-package change (unexpected)"
fi

# The .txt file itself must be self-covered (expected: "non-package")
if echo ".buildkite/scripts/non_package_patterns.txt" | pr_has_package_related_files; then
    echo "result: package change (unexpected)"
else
    echo "result: non-package change (expected)"
fi
```

---

### DIFF
--- a/.buildkite/scripts/common.sh
+++ b/.buildkite/scripts/common.sh
@@ -887,7 +887,7 @@ teardown_test_package() {
 
 # list all directories that are packages from the root of the repository
 list_all_directories() {
-    mage -d "${WORKSPACE}" listPackages
+    mage -d "${WORKSPACE}" listPackages |grep -E '^packages/(elastic_package_registry|nginx)$'
 }
 
 check_package() {
--- a/.buildkite/scripts/common.sh
+++ b/.buildkite/scripts/common.sh
@@ -887,7 +887,7 @@ teardown_test_package() {
 
 # list all directories that are packages from the root of the repository
 list_all_directories() {
-    mage -d "${WORKSPACE}" listPackages |grep -E '^packages/(elastic_package_registry|nginx)$'
+    mage -d "${WORKSPACE}" listPackages
 }
 
 check_package() {
--- a/.buildkite/scripts/common.sh
+++ b/.buildkite/scripts/common.sh
@@ -704,55 +704,11 @@ is_logsdb_compatible() {
 # pr_has_package_related_files reads file paths from stdin (one per line) and returns 0 if
 # at least one path is not matched by any non-package pattern, meaning package tests should run.
 # Returns 1 if every path is covered by a non-package pattern (no package tests needed).
+# Patterns are loaded from non_package_patterns.txt (one ERE per line; # comments and blank lines ignored).
 # Avoid using grep -q: SIGPIPE under set -o pipefail can cause missed detections.
 # See: https://buildkite.com/elastic/integrations/builds/25606
 pr_has_package_related_files() {
-    local non_package_patterns=(
-        'packages/'
-        '\.agents/skills/'
-        '\.backports\.yml'
-        '\.buildkite/pipeline\.backport\.yml'
-        '\.buildkite/pipeline\.publish\.yml'
-        '\.buildkite/pipeline\.schedule-daily\.yml'
-        '\.buildkite/pipeline\.schedule-weekly\.yml'
-        '\.buildkite/pipeline\.serverless\.yml'
-        '\.buildkite/pull-requests\.json'
-        '\.buildkite/scripts/backport_branch\.sh'
-        '\.buildkite/scripts/backport_branch_lib\.sh'
-        '\.buildkite/scripts/check_backports_inventory\.sh'
-        '\.buildkite/scripts/notify_backport_pr\.sh'
-        '\.buildkite/scripts/trigger_backport\.sh'
-        '\.buildkite/scripts/trigger_backport_lib\.sh'
-        '\.buildkite/scripts/build_packages\.sh'
-        '\.buildkite/scripts/check_changelog_entries\.sh'
-        '\.buildkite/scripts/packages/.+\.sh'
-        '\.buildkite/scripts/requirements-ci-python-scripts\.txt'
-        '\.buildkite/scripts/run_buildkite_scripts_tests\.sh'
-        '\.buildkite/scripts/run_dev_scripts_tests\.sh'
-        '\.buildkite/scripts/test_backport_branch\.sh'
-        '\.buildkite/scripts/test_check_changelog_entries\.sh'
-        '\.buildkite/scripts/test_helpers\.sh'
-        '\.buildkite/scripts/test_non_package_patterns\.sh'
-        '\.buildkite/scripts/test_trigger_backport\.sh'
-        '\.github/dependabot\.yml'
-        '\.github/stale\.yml'
-        '\.github/workflows/'
-        '\.github/CODEOWNERS'
-        '\.github/ISSUE_TEMPLATE/'
-        '\.github/PULL_REQUEST_TEMPLATE\.md'
-        '\.gitignore'
-        '\.mergify\.yml'
-        'catalog-info\.yaml'
-        'dev/backports/'
-        'dev/scripts/'
-        'docs/'
-        'CODE_OF_CONDUCT\.md'
-        'CONTRIBUTING\.md'
-        'README\.md'
-    )
-    local non_package_regex
-    non_package_regex="^($(IFS='|'; echo "${non_package_patterns[*]}"))"
-    grep -E -v "${non_package_regex}" > /dev/null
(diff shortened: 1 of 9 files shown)


.buildkite/scripts/non_package_patterns.txt
```diff
@@ -0,0 +1,46 @@
...
+^\.gitignore
+^\.mergify\.yml
+^catalog-info\.yaml
+^dev/backports/
+^dev/scripts/
+^docs/
+^CODE_OF_CONDUCT\.md
+^CONTRIBUTING\.md
+^README\.md
+^.buildkite/scripts/common.sh
```
mrodm: To be reverted before merging.
It's here for testing purposes.

.buildkite/scripts/trigger_backport_lib.sh
```diff
@@ -48,11 +48,13 @@ backports_yml_changed() {
...
-    local dry_run="$3"        # "true" or "false" — passed to the triggered build
-    local pr_number="$4"      # merged PR number, may be empty
-    local pipeline_file="$5"  # output file; written only if new entries are found
+    local old_inventory="$1"    # path to the pre-change inventory
+    local new_inventory="$2"    # path to the post-change inventory
+    local dry_run="$3"          # "true" or "false" — passed to the triggered build
+    local pr_number="$4"        # merged PR number, may be empty
+    local pipeline_file="$5"    # output file; written only if new entries are found
+    local buildkite_pr="${6:-}" # BUILDKITE_PULL_REQUEST value; empty or "false" means not a PR build
+    local buildkite_commit="${7:-}" # BUILDKITE_COMMIT value; only used when buildkite_pr is set
```
mrodm, line 57: These changes related to `trigger_backport` were added to fix the execution of tests locally. Without this fix, tests were failing because `BUILDKITE_PULL_REQUEST` was not defined.

.buildkite/scripts/common.sh
```diff
@@ -701,6 +701,16 @@ is_logsdb_compatible() {
...
 
+# pr_has_package_related_files reads file paths from stdin (one per line) and returns 0 if
+# at least one path is not matched by any non-package pattern, meaning package tests should run.
+# Returns 1 if every path is covered by a non-package pattern (no package tests needed).
+# Patterns are loaded from non_package_patterns.txt (one ERE per line; # comments and blank lines ignored).
+# Avoid using grep -q: SIGPIPE under set -o pipefail can cause missed detections.
+# See: https://buildkite.com/elastic/integrations/builds/25606
+pr_has_package_related_files() {
+    grep -Evf <(grep -Ev '^[[:space:]]*(#|$)' "${SCRIPTS_BUILDKITE_PATH}/non_package_patterns.txt") > /dev/null
+}
```
mrodm, line 712: With this refactor, I want to avoid to trigger the testing of all packages if we want to include a new pattern to the list.

File `non_package_patterns.txt` will be updated but `common.sh` should not be needed to be updated, and therefore it will not trigger all the integration tests.

.buildkite/scripts/test_non_package_patterns.sh
```diff
@@ -0,0 +1,142 @@
...
+    ".buildkite/scripts/requirements-ci-python-scripts.txt"
+    ".buildkite/scripts/run_buildkite_scripts_tests.sh"
+    ".buildkite/scripts/run_dev_scripts_tests.sh"
+    ".buildkite/scripts/test_backport_branch.sh"
+    ".buildkite/scripts/test_check_changelog_entries.sh"
+    ".buildkite/scripts/test_helpers.sh"
+    ".buildkite/scripts/test_non_package_patterns.sh"
+    ".buildkite/scripts/test_trigger_backport.sh"
+    # TODO: revert — temporarily excluded in non_package_patterns.txt for debugging
+    ".buildkite/scripts/common.sh"
```
mrodm: To be reverted before merging

It requires to revert https://github.com/elastic/integrations/pull/19800/changes/20c3eeb36e687d8f4fe314e191bbc43b1eb145ee

.buildkite/scripts/test_non_package_patterns.sh
```diff
@@ -0,0 +1,142 @@
...
+# Paths NOT covered by non_package_patterns: must trigger tests
+# ---------------------------------------------------------------------------
+echo ""
+echo "--- pr_has_package_related_files: infra-file changes DO trigger tests"
+
+TRIGGERING_PATHS=(
+    ".buildkite/pipeline.yml"
+    "Makefile"
+    ".buildkite/scripts/build_packages_serverless.sh"
+)
```
mrodm, line 94: To be re-added before merging


```suggestion
TRIGGERING_PATHS=(
    ".buildkite/scripts/common.sh"
    ".buildkite/pipeline.yml"
    "Makefile"
    ".buildkite/scripts/build_packages_serverless.sh"
)
```

It requires to revert https://github.com/elastic/integrations/pull/19800/changes/20c3eeb36e687d8f4fe314e191bbc43b1eb145ee

teresaromero (approved):

.buildkite/scripts/test_non_package_patterns.sh
mrodm: Reverted in https://github.com/elastic/integrations/pull/19800/commits/40cd85ed152700774279c70c8a4c26559e0b6111

.buildkite/scripts/test_non_package_patterns.sh
mrodm: Reverted in https://github.com/elastic/integrations/pull/19800/commits/40cd85ed152700774279c70c8a4c26559e0b6111

.buildkite/scripts/non_package_patterns.txt
mrodm: Reverted in https://github.com/elastic/integrations/pull/19800/commits/2cd4e604bfcb758d37614561cd9ae6f5bcbbcac6

PR #19801, opened 2026-06-26
o365: fix Painless regex character-limit and Sender mapping errors

chemamartinez:
## Proposed commit message

```
o365: fix Painless regex character-limit and Sender mapping errors

Two production pipeline errors observed in `logs-o365.audit`:

1. `script_item_attachments` failed with a Painless regex character-limit
   error on long attachment filenames containing parentheses.
   The regex `/^(.+?)\.([^.]+)\s*\((\d+)b\)$/` was replaced
   with string operations (`lastIndexOf`, `substring`, `parseLong`)
   that have no backtracking cost and no character-count limit.

2. `document_parsing_exception` on `o365.audit.Sender` when Teams-originated
   ExchangeItem events deliver the field as a structured object
   `{UPN=..., MRI=..., DisplayName=...}` instead of a plain string. Fixed by
   setting `subobjects: false` with `object_type: keyword` on the field
   definition, following the same pattern used for `Parameters` and
   `ModifiedProperties`.

Includes regression test fixture `test-exchange-item-edge-cases.json` with
one event reproducing each bug.
```

## Checklist

### DIFF
--- a/packages/o365/changelog.yml
+++ b/packages/o365/changelog.yml
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "3.10.3"
+  changes:
+    - description: Fix `script_item_attachments` processor failing with a Painless regex character-limit error on long attachment filenames that contain parentheses. The regex is replaced with string operations that have no backtracking cost.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/99999
+    - description: "Set subobjects: false on o365.audit.Sender field to avoid document_parsing_exception when the field arrives as a structured object instead of a plain string."
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/99999
 - version: "3.10.2"
   changes:
     - description: Use the ECS definition for email.attachments.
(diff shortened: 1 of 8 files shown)


narph (approved):

packages/o365/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "3.10.3"
+  changes:
+    - description: Fix `script_item_attachments` processor failing with a Painless regex character-limit error on long attachment filenames that contain parentheses. The regex is replaced with string operations that have no backtracking cost.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/19801
+    - description: "Set subobjects: false on o365.audit.Sender field to avoid document_parsing_exception when the field arrives as a structured object instead of a plain string."
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/19801
```
ShourieG, line 9: can we have a single collapsed description since they are both bugfixes ?

chemamartinez: Both errors are different and unrelated, do you think it is a better practice to collapse the descriptions into just one?

ShourieG (approved):

PR #19803, opened 2026-06-26
[aws/config]: Skip service-linked rules and tolerate per-rule errors

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
aws/config: skip service-linked rules and tolerate per-rule errors

The AWS Config data stream listed every Config rule and then called
GetComplianceDetailsByConfigRule for each one. AWS owns service-linked
rules and blocks that call for them with an AccessDeniedException, which
aborted the whole collection cycle and degraded the input so no rules
were ingested.

Filter out service-linked rules in the DescribeConfigRules response by
dropping any rule with a populated CreatedBy field, which AWS sets only
for service-linked rules, so the blocked call is never attempted. As a
defence-in-depth measure, treat a per-rule GetComplianceDetailsByConfigRule
failure as non-fatal: emit the error as an advancing event and continue
with the remaining rules instead of stopping the cycle. Thread the
resource url through every output state branch so the program stays
self-contained across evaluations.

Add a service-linked rule and a matching AccessDeniedException fixture to
the system-test mock to confirm the rule is filtered and never queried,
and bump the package to 6.20.2.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/aws/data_stream/config/agent/stream/cel.yml.hbs
+++ b/packages/aws/data_stream/config/agent/stream/cel.yml.hbs
@@ -65,6 +65,7 @@ program: |
       {
         // State management: preserve existing state from previous iteration
         "worklist": state.worklist,
+        "url": state.url,
         "batch_size": state.batch_size,
         "has_next": state.has_next,
         "next": state.next,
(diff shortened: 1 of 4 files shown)


narph (approved):

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
```diff
@@ -188,10 +189,21 @@ program: |
           }
         ).do_request().as(resp, (resp.StatusCode == 200) ?
           // Response Processing: Successfully got config rules list
-          resp.Body.decode_json().as(body,
+          resp.Body.decode_json().as(raw_body,
+            // Skip AWS service-linked rules. AWS owns these rules and blocks
+            // GetComplianceDetailsByConfigRule for them with an AccessDeniedException
+            // ("An AWS service owns ServiceLinkedConfigRule"). They are identified by a
+            // populated CreatedBy field, which AWS sets only for service-linked rules.
```
kcreddy, line 196: > They are identified by a
            // populated CreatedBy field, which AWS sets only for service-linked rules.

It would be nice to add a documentation link to confirm this. Not a blocker though.

kcreddy (approved):
LGTM

packages/aws/data_stream/config/agent/stream/cel.yml.hbs
ShourieG: Will add it as a follow-up PR as doing it here would trigger another 2.5 hr wait for the aws tests.

ShourieG: Just as a ref: It's in this doc [here](https://docs.aws.amazon.com/config/latest/APIReference/API_ConfigRule.html)

PR #19805, opened 2026-06-26
CrowdStrike Adding Identity Protection IDP Parsing for FDR

jeff-bb:
## Proposed commit message

Create a pipeline (similar to Epp) for Idp events. The Falcon pipeline path has this, but the FDR pipline path does not. I used the existing Falcon Idp as the starting point with a few minor changes based on a combination of our real world data and the current version of the "Identity Protection detection summary fields" document from [CrowdStrike](https://docs.crowdstrike.com/r/en-US/g6p281ck/ae18b917)

I've created a test-fdr-idp-detection-summary.log file with a single sanitized event, but do not know how to test this to produce the expected output file.


efd6:
@jeff-bb To generate the test expectations, you will need to run `elastic-package test pipeline -g` in the package directory. `elastic-package` can be obtained from https://github.com/elastic/elastic-package/releases/

packages/crowdstrike/data_stream/fdr/_dev/test/pipeline/test-fdr-idp-detection-summary.log
navnit-elastic, line 1: The `default.yml` pipeline is configured to branch on `ctx.crowdstrike?.ExternalApiType == 'Event_IdpDetectionSummaryEvent'`, but this field is missing from the sample. Additionally, the timestamp format in the sample consists of ISO 8601 strings, whereas the pipeline expects the UNIX epoch format. Is this sample from the FDR feed?

sharad-elastic:
@jeff-bb, I wanted to check if there are any blockers with addressing the pending comments. Please let me know if you need any help with the comments or if there’s anything I can assist with to move the PR forward.

PR #19806, opened 2026-06-26
[kolide] Add new integration for Kolide Logs

vinit-chauhan:
## Proposed commit message

```
kolide: add new integration for Kolide Logs

Add a new Elastic integration for Kolide (by 1Password) Device Trust
that collects device-trust and endpoint-compliance logs and normalizes
them to the Elastic Common Schema (ECS).

The integration supports three collection methods that can be used
individually or combined:

  - Webhooks (HTTP endpoint): Kolide pushes events in near real time
    to a single signed endpoint; deliveries are verified with an
    HMAC-SHA256 signature.
  - REST API (CEL polling): the agent polls the Kolide REST API
    (version 2026-04-07) with cursor-based pagination for backfill
    and fuller resource records.
  - AWS S3 (Log Pipeline): the agent reads objects Kolide writes to a
    customer-owned bucket via SQS notifications or direct polling.

It ships six data streams:

  - webhook: single ingress that receives all webhook event types and
    routes each event to the correct data stream automatically.
  - auth: SSO authentication sessions (REST API and S3).
  - issues: device posture-check failures and resolutions (REST API).
  - device: device inventory and trust-status changes (REST API).
  - audit: administrative console audit events (REST API and S3).
  - device_check: full check-run results from the Log Pipeline (S3),
    covering passing, failing, inapplicable, and unknown runs.

Includes ingest pipelines with ECS mappings, field definitions,
pipeline and system tests, an overview dashboard, and documentation.
```

## Checklist

## Author's Checklist

- [ ] Review the `webhook` data stream routing logic that dispatches each event type to the correct destination data stream.
- [ ] Review `event.outcome` semantics for posture data (`device_check`, `issues`): `passing`/resolved maps to `success`, `failing`/open maps to `failure`, and `inapplicable`/`unknown` maps to `unknown`.
- [ ] Confirm the AWS S3 default Bucket List Prefixes include the `kolide/` parent (`kolide/auth_logs/`, `kolide/audit_logs/`, `kolide/check_runs/`).
- [ ] Confirm `device_check` host correlation: only `host.id` is set (no `host.name` in the payload); document the enrich-policy workaround.
- [ ] Verify the integration covers Kolide K2 (current Device Trust platform) and not the legacy open-source Kolide Fleet (osquery) product.

1. Check out this branch.
2. From `packages/kolide`, run the static checks and tests:

```bash
elastic-package check -v
elastic-package test pipeline -v
```

3. To exercise the integration end to end, bring up a local stack and run the system tests:

```bash
elastic-package test system -v
```

4. Optionally, validate manually in Kibana:
   - Add the **Kolide** integration from Management > Integrations.
   - For webhooks, enable the `webhook` data stream and provide the HMAC signing secret.
   - For the REST API, enable the data streams to poll (`auth`, `issues`, `device`, `audit`), set the API URL (`https://api.kolide.com`) and API key.
   - For AWS S3, provide AWS credentials and enable the `aws-s3` input on `auth`, `audit`, and/or `device_check` (defaults to the `kolide/...` prefixes).
   - In Discover, confirm documents arrive in the `logs-kolide.*` data streams, and open the **Kolide** dashboard.

## Related issues

- Closes https://github.com/elastic/integration-experience/issues/956

## Screenshots

image
image


packages/kolide/data_stream/audit/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,109 @@
...
+          state.url.trim_right("/") + "/audit_logs?" + (
+            (next_token != "") ?
+              {
+                "per_page": [string(int(state.batch_size))],
+                "cursor": [next_token],
+              }
+            :
+              {
+                "per_page": [string(int(state.batch_size))],
+                "query": ["timestamp>" + since],
```
ilyannn: **Possible data loss: CEL timestamp cursor uses a strict `>` boundary**

The next-poll query filters with a strict `timestamp>` against `since`, where `since` is the max timestamp of the previous batch (`cursor.last_timestamp`). Because the boundary record is *not* re-fetched, any audit record recorded at *exactly* the boundary timestamp after the poll completes is excluded forever (Kolide audit timestamps are 1-second precision, so multiple events per second are common).

**Reproduction (two polls):**
1. Poll 1: `since = now-24h`; API returns records at `T1 < T2 < T3`, `next_cursor` empty → cursor stores `last_timestamp = T3`.
2. Between polls a new action `R4` is recorded at exactly `T3`.
3. Poll 2: `since = T3`, query `timestamp>T3` → `R4` (`== T3`) is excluded and never re-queried → permanently lost.

**Fix:** use an inclusive boundary. The existing `event.id` fingerprint dedup (`default.yml` `fingerprint_event_id`) maps the re-fetched boundary records to the same `_id` and overwrites them, so no duplicates result — which is exactly what the `default.yml` comment ("CEL cursor uses an inclusive (>=) boundary and may re-fetch the last record") already says the design does.

```suggestion
                "query": ["timestamp>=" + since],
```

The same change applies to `packages/kolide/data_stream/auth/agent/stream/cel.yml.hbs:46`.

:robot: Comment by Claude Code under my supervision

packages/kolide/data_stream/audit/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,274 @@
...
+  - pipeline:
+      name: '{{ IngestPipeline "extended-mappings" }}'
+      tag: extended_mappings
+      ignore_missing_pipeline: true
+
+  # ECS categorization from the resolved event.action.
+  - pipeline:
+      name: '{{ IngestPipeline "categorize" }}'
+      tag: categorize
+      ignore_missing_pipeline: true
```
ilyannn, line 226: **Webhook `audit_log.recorded` events with an unrecognized description get no ECS categorization**

`categorize.yml` is gated on `if: ctx.event?.action != null`, and `extended-mappings.yml` leaves `event.action` unset when the description matches no prefix. The API path has a fallback that sets `event.action: audit_log` (lines 88–92, gated on `ctx.json?.event == null`), but the webhook path has none — `webhook.yml` only sets `event.action` for event types other than `audit_log.recorded`. So a webhook audit event whose description isn't classified ends up with **no `event.action`, `event.kind`, `event.category`, or `event.type`**.

(The `extended-mappings.yml` header comment, lines 9–10, even claims unmatched descriptions "keep the generic audit_log / webhook default" — but that default only exists on the API path, so it's inaccurate for webhooks.)

**Reproduction (pipeline test):** add a fixture under `data_stream/audit/_dev/test/pipeline/` with a webhook event whose description matches nothing:
```json
{
  "events": [
    {
      "event": "audit_log.recorded",
      "id": "01JA67B1DYJCKJ1J73T0F5UNCT",
      "timestamp": "2024-10-14T19:16:05Z",
      "data": {
        "actor_name": "Alice Johnson",
        "actor_type": "User",
        "actor_email": "alice.johnson@example.com",
        "ip_address": "81.2.69.142",
        "description": "Performed an unrecognized administrative action"
      }
    }
  ]
}
```
Run `elastic-package test pipeline --data-streams audit`: the output has no `event.action`/`kind`/`category`/`type` (vs. the existing `test-kolide-audit-webhook` fixture which categorizes to `check_enabled`).

**Fix:** add a path-agnostic fallback immediately before the `categorize` dispatch so both inputs behave identically (the generic `audit_log` entry already exists in `categorize.yml`). The existing API-only fallback at lines 88–92 then becomes redundant and can be removed.

```suggestion
  - set:
      field: event.action
      tag: set_event_action_fallback
      value: audit_log
      if: ctx.event?.action == null
  - pipeline:
      name: '{{ IngestPipeline "categorize" }}'
      tag: categorize
      ignore_missing_pipeline: true
```

:robot: Comment by Claude Code under my supervision

packages/kolide/data_stream/auth/fields/ecs.yml
```diff
@@ -0,0 +1,76 @@
...
+- name: source.geo.city_name
+  external: ecs
+- name: source.geo.continent_name
+  external: ecs
+- name: source.geo.country_iso_code
+  external: ecs
+- name: source.geo.country_name
+  external: ecs
+- name: source.geo.location
+  external: ecs
```
ilyannn: **CI failure: `source.geo.location` declaration causes static test failure in `kolide.auth`**

Build [#45159](https://buildkite.com/elastic/integrations/builds/45159#019f0602-f1c6-4993-aa19-548000e4d2a2) fails on:

```
static test: Verify sample_event.json in kolide.auth
test case failed: one or more errors found in document:
  [0] field "source.geo.location.coordinates" is undefined, could be a multifield
  [1] field "source.geo.location.type" is undefined, could be a multifield
```

**Root cause:** declaring `source.geo.location` with `external: ecs` pulls in the full ECS 9.x `geo_point` schema, which includes GeoJSON sub-properties (`.type`, `.coordinates`). The elastic-package static validator then requires those sub-properties to be explicitly declared — but they aren't, so the `auth` sample event (which has `"location": {"lat": …, "lon": …}`) fails validation.

**Fix:** replace the single parent entry with the two flat subfield declarations (same pattern used by `lyve_cloud` and `istio`, both of which pass static tests):

```suggestion
- name: source.geo.location.lat
  external: ecs
- name: source.geo.location.lon
  external: ecs
```

The same change is needed at `packages/kolide/data_stream/audit/fields/ecs.yml:45` — `audit`'s `sample_event.json` has empty `source.geo`, so it doesn't fail today, but it carries the same latent issue.

The `sample_event.json` `{"lat": …, "lon": …}` format in `auth` is correct and stays as-is.

:robot: Comment by Claude Code under my supervision

haetamoudi:
Thanks for this PR. Quick questions/comments:
1. Can we add agentless deployment in the manifest?
2. The four CEL data streams (auth, audit, device, issues) are missing the [recommended](https://github.com/elastic/integration-skills/blob/main/skills/ingest-pipelines/SKILL.md?plain=1#L69) `remove` + `terminate` opening processors before the `message → event.original` handling. Not a blocker, but worth adding for Agentless readiness

packages/kolide/data_stream/auth/elasticsearch/ingest_pipeline/s3.yml
```diff
@@ -0,0 +1,59 @@
...
+  - remove:
+      field: json.succeeded
+      tag: remove_succeeded
+      ignore_missing: true
+  # The nested per-event list on auth_session_summary is retained in
+  # event.original; drop the working copy to avoid mapping the variable array.
+  - remove:
+      field: json.authentication_events
+      tag: remove_authentication_events
+      ignore_missing: true
```
haetamoudi, line 59: Should we add a sub-pipeline on_failure handler to mirror the other sub-pipelines?

packages/kolide/data_stream/audit/elasticsearch/ingest_pipeline/s3.yml
```diff
@@ -0,0 +1,47 @@
...
+  - rename:
+      field: json.ip_address
+      target_field: source.ip
+      tag: rename_source_ip
+      ignore_missing: true
+      if: ctx.source?.ip == null && ctx.json?.ip_address != null && ctx.json.ip_address != ''
+  - remove:
+      field: json.ip_address
+      tag: remove_null_ip_address
+      ignore_missing: true
```
haetamoudi, line 47: Should we add a sub-pipeline on_failure handler to mirror the other sub-pipelines?

packages/kolide/data_stream/auth/fields/fields.yml
```diff
@@ -0,0 +1,47 @@
...
+    - name: okta
+      type: group
+      fields:
+        - name: app_name
+          type: keyword
+          description: Okta app name from the SAML request.
+        - name: app_instance_id
+          type: keyword
+          description: Okta app instance identifier from the SAML request.
+    - name: issues_displayed
```
haetamoudi, line 29: `issues_displayed` is declared as `type: group` but the API returns an array of objects and the test expected output (line 756) confirms the array shape. With `subobjects: false`, multiple issues will flatten to disconnected multi-value fields (we will lose which `id` belongs to which `title`). We could update it to either `type: nested`, `type: flattened`, or `type: object`. Also, sample_event.json:53 shows it as a flat object instead of an array, should be fixed to match.

packages/kolide/data_stream/audit/_dev/deploy/tf/files/2026-06-13T01:21:19Z.json
```diff
@@ -0,0 +1 @@
+{"type":"audit_log","timestamp":"2026-06-13T01:21:19Z","data":{"actor_name":"Noel Vasquez","actor_type":"ApiKey","actor_email":null,"ip_address":"2601:14d:4f00:37a0:80ed:dfd0:4335:8e73","description":"Quinn Park, os_version, hash, launchd, file, crontab, processes, etc_hosts, process_open_sockets"}}
```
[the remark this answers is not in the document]
vinit-chauhan: It was a dummy IP address, but I've changed it to proposed value just to be safe.

packages/kolide/data_stream/issues/agent/stream/cel.yml.hbs
```diff
@@ -0,0 +1,111 @@
...
+          state.url.trim_right("/") + "/issues?" + (
+            (next_token != "") ?
+              {
+                "per_page": [string(int(state.batch_size))],
+                "cursor": [next_token],
+              }
+            :
+              {
+                "per_page": [string(int(state.batch_size))],
+                "query": ["detected_at>" + since],
```
haetamoudi:
```suggestion
                "query": ["detected_at>=" + since],
```

(fixed in `auth` and `audit`)

haetamoudi (approved):

narph (approved):

ilyannn (approved):
LGTM

PR #19893, opened 2026-06-30
[Fortinet Fortigate] Prevent pipeline failures from oversized duration and non-IP address values

robester0403:
Two FortiGate log conditions currently fail the whole document in the `log` data stream:
- **`duration` overflow** — some FortiOS IPsec `tunnel-stats` events emit an unsigned 64-bit `duration` (e.g. `18446744073700579894`) that exceeds `long`, so the duration script throws and the entire event (incl. `advpnsc`, byte counts, etc.) is dropped.
- **Non-IP `source.ip` / `destination.ip`** — malformed values (e.g. `dstip=authid="..."`) cause `network_direction`/`geoip` to hard-fail and the field to be `_ignored` at index time.

## Proposed commit message

Prevent pipeline failures from oversized duration and non-IP address values

## Checklist

## Related issues

Partially Resolves: https://github.com/orgs/elastic/projects/1874/views/1?filterQuery=-is%3Apr+assignee%3A%40me&pane=issue&itemId=206368213&issue=elastic%7Cintegrations%7C19832

## Screenshots

Screenshot 2026-06-30 at 4 39 30 PM


packages/fortinet_fortigate/data_stream/log/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -405,10 +406,33 @@ processors:
       target_field: network.direction
       ignore_missing: true
       if: ctx.network?.direction == null
+  - grok:
```
andrewkroh: You should use a `convert` processor with `type: ip` to validate fields are IP addresses and will index in elasticsearch. The processor doesn't actually convert the field value or type (it's still a string), but you can attach an `on_failure` to remove invalid IPs.

andrewkroh (approved):

PR #19894, opened 2026-06-30
[Cisco Ise] Parse TACACS accounting start_time/stop_time with sub-second epoch precision

robester0403:
## Proposed commit message

Parse TACACS accounting start_time/stop_time with sub-second epoch precision

In the `log` data stream, Cisco ISE TACACS accounting emits `start_time`/`stop_time` AVPairs as **nanosecond** epoch values (e.g. `start_time=1782407978339676873`), but the pipeline parsed them as `UNIX` (seconds). The result overflowed to year `+292278994`, which Elasticsearch then rejected at index time — so `cisco_ise.log.avpair.start_time`/`stop_time` were silently `_ignored` and lost.

## Checklist

## Related issues

Partially Resolves: https://github.com/orgs/elastic/projects/1874/views/1?filterQuery=-is%3Apr+assignee%3A%40me&pane=issue&itemId=206368213&issue=elastic%7Cintegrations%7C19832 

## Screenshots

Screenshot 2026-06-30 at 5 31 49 PM


packages/cisco_ise/data_stream/log/elasticsearch/ingest_pipeline/pipeline_tacacs_accounting.yml
```diff
@@ -320,13 +342,19 @@ processors:
         - remove:
             tag: remove_cisco_ise_log_avpair_elapsed_time_bbb2e1da
             field: cisco_ise.log.avpair.elapsed_time
+  - gsub:
+      tag: gsub_cisco_ise_log_avpair_stop_time_58f81d66
+      field: cisco_ise.log.avpair.stop_time
+      pattern: '^(\d{13})\d+$'
+      replacement: '$1'
+      if: ctx.cisco_ise?.log?.avpair?.stop_time instanceof String && ctx.cisco_ise.log.avpair.stop_time.length() > 13
```
andrewkroh: `start_time` has a `!= "0"` guard. `stop_time` does not. Align them.

packages/cisco_ise/data_stream/log/elasticsearch/ingest_pipeline/pipeline_tacacs_accounting.yml
```diff
@@ -230,14 +230,20 @@ processors:
       target_field: cisco_ise.log.authen_method
       ignore_failure: true
       ignore_missing: true
+  - gsub:
+      tag: gsub_cisco_ise_log_avpair_start_time_18d7aae2
+      field: cisco_ise.log.avpair.start_time
+      pattern: '^(\d{13})\d+$'
```
andrewkroh: None of the new processors have a `description`. The digit-count magic needs to be explained.

andrewkroh (commented):
Consider using a script processor to centralize and document the approach ([prior art](https://github.com/elastic/integrations/blob/6b441edca050293e3375dad3dad65de6df1ed9d7/packages/aws_securityhub/data_stream/finding/elasticsearch/ingest_pipeline/default.yml#L41-L47)). It's probably more readable. Something like this then a single `date` processor that works on UNIX_MS. It eliminates the digit counting but uses a regex per doc.

```
- script:
    description: >-
      Normalize the epoch to milliseconds regardless of source unit (s / ms / µs / ns).
      Cisco ISE TACACS+ accounting emits nanoseconds; older data uses seconds.
    lang: painless
    if: ctx.cisco_ise?.log?.avpair?.start_time instanceof String
    source: |
      def s = ctx.cisco_ise.log.avpair.start_time;
      if (s == "0" || !(s ==~ /^\d+$/)) { return; }
      long v = Long.parseLong(s);
      if      (v >= 1000000000000000000L) { v /= 1000000L; } // ns -> ms
      else if (v >= 1000000000000000L)    { v /= 1000L; }    // µs -> ms
      else if (v <  10000000000L)         { v *= 1000L; }    // s  -> ms
      ctx.cisco_ise.log.avpair.start_time = v;
```

andrewkroh (approved):
LGTM.

PR #19030, opened 2026-05-15
[k8s otel] add ml_modules for k8s otel content package

tommyers-elastic:
# Kubernetes ML anomaly detection jobs

Adds two Kibana ML modules covering Kubernetes (OpenTelemetry) anomaly detection. Split into two modules so each maps to one OTel data stream and only lights up against the right data view in Kibana's ML module recognizer.

These jobs are deliberately scoped to detection that fixed-threshold alert rules cannot do: workload-wide drift, time-of-day pattern violations, per-reason event spikes. Per-replica, threshold-crossing events stay in the alert layer.

## `kubernetes_otel-metrics-ml`

Source: `metrics-kubeletstatsreceiver.otel-*`

- **`k8s_workload_memory_anomaly`** — `high_mean(k8s.pod.memory.working_set)` across three detectors (`by k8s.deployment.name`, `by k8s.statefulset.name`, `by k8s.daemonset.name`), each partitioned by `k8s.namespace.name`. 15m bucket span, 128mb model memory. Detects workloads whose pod memory has grown unusually relative to that workload's own baseline — catches slow-leak trajectories that affect every replica before any individual pod crosses a hard OOM limit. Workload identities survive pod recycling, which per-pod modelling does not. Influencers: deployment/statefulset/daemonset name, namespace.

- **`k8s_network_rx_anomaly`** / **`k8s_network_tx_anomaly`** — `max(bytes_per_bucket) partition k8s.namespace.name`. Aggregation-mode datafeeds compute the proper bytes/bucket rate from the cumulative `k8s.pod.network.io` counter via a `derivative` pipeline aggregation, with a `positive_only` `bucket_script` clamping the negative spikes that occur when a pod restart resets the counter. 5m sub-buckets feed the 15m model for finer derivative resolution. 32mb model memory each. Split into one job per direction (transmit / receive) because direction can't be a partition through the aggregation pipeline cleanly. Influencers: namespace, pod uid.

## `kubernetes_otel-events-ml`

Source: `logs-k8seventsreceiver.otel-*`

[...]

PR #19922, opened 2026-07-01
[aws_ecs_otel] Add ML anomaly detection module

JM-elastic:
## What

Adds a machine-learning anomaly-detection module (`kibana/ml_module/`) to the **aws_ecs_otel** integration, proposing anomaly detection as an addition alongside the integration's existing dashboards and alert rules. Modeled on the `kubernetes_otel` ML module ([#19030](https://github.com/elastic/integrations/pull/19030)).

## Why — complements the threshold alerts, doesn't duplicate them

The shipped alert rules catch per-entity threshold breaches (a value crossing a fixed line). These ML jobs model each metric **per entity against its own history**, catching the *drift* those miss — e.g. a slow memory leak across task recycling before any single task crosses a hard threshold. Each detector's description defers per-entity spikes to the alert rules, the same split `kubernetes_otel` uses. The detectors are drawn from the service's own signals and real failure modes — not tailored to any specific workflow.

## Jobs

- `aws_ecs_service_resource_anomaly` — per `ServiceName` (partition `ClusterName`): `high_mean` **MemoryUtilization** and **CPUUtilization**.

Datafeeds are **composite-aggregated** — required, because these `metrics-aws.*.otel-*` indices contain `aggregate_metric_double` fields that a plain (non-aggregating) ML datafeed cannot read.

## Validation

Drafted and validated against live AWS OTel telemetry: the job(s) establish baselines over historical data. The RDS connection-pool-exhaustion case was scored against a known injected incident and detected it on the correct entity (recall/precision/f1 = 1.0).

Methodology, tooling, and the scoring harness: https://github.com/elastic/aws_otel_ml_draft

## Notes for reviewers (@elastic/obs-infraobs-integrations)

- **Draft** — proposing for your review; happy to adjust job naming, detector selection, `bucket_span`, or thresholds.
- Package stays `subscription: basic` (matches `kubernetes_otel`; ML availability is a deployment concern, not a package condition).
- Entity fields use the **raw CloudWatch dimensions** present in the indexed documents (`ServiceName` + `ClusterName`), not the normalized fields the alert-rule `termField`s reference (those are not present in the documents).
- Open tuning item: ECS `MemoryUtilization` can be noisy on some services — a candidate for `bucket_span` tuning.

### DIFF
--- a/packages/aws_ecs_otel/changelog.yml
+++ b/packages/aws_ecs_otel/changelog.yml
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "0.10.0"
+  changes:
+    - description: Add ML anomaly detection module for ECS service CPU and memory utilization.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/19922
+    - description: Add ml_module to the Kibana asset tags.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/19922
 - version: "0.9.0"
   changes:
     - description: Add missing recommended alerts for CPU and memory reservation high.
--- a/packages/aws_ecs_otel/kibana/ml_module/aws_ecs_otel-metrics-ml.json
+++ b/packages/aws_ecs_otel/kibana/ml_module/aws_ecs_otel-metrics-ml.json
@@ -1,0 +1,182 @@
+{
+    "id": "aws_ecs_otel-metrics-ml",
+    "type": "ml-module",
+    "migrationVersion": {
+        "search": "7.9.3"
+    },
+    "references": [],
+    "attributes": {
+        "id": "aws_ecs_otel-metrics-ml",
+        "title": "AWS ECS service resources (OpenTelemetry)",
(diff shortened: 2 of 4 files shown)


jakubgalecki0 (commented):
Lets please make sure that we follow naming convention of other `aws_otel` assets i.e. 

> `[<servicename> OTel] Description` 

Right now we have following convention: `AWS ECS service resources (OpenTelemetry)`

Also add entry for `tags` since `package-spec` allows it to be tagged - https://github.com/elastic/package-spec/blob/main/spec/integration/kibana/tags.spec.yml#L34

```
  asset_types:
    - dashboard
    - alerting_rule_template
    - slo_template
    - ml_module <<< 
```

Other than that it looks good.

jakubgalecki0:
@JM-elastic should we also add `cloud.region` to aggregations as in other packages?

jakubgalecki0 (approved):

ishleenk17 (approved):

PR #19925, opened 2026-07-01
[aws_elb_metrics_otel] Add ML anomaly detection modules

JM-elastic:
## What

Adds a machine-learning anomaly-detection module (`kibana/ml_module/`) to the **aws_elb_metrics_otel** integration, proposing anomaly detection as an addition alongside the integration's existing dashboards, alert rules, and SLO templates. Modeled on the `kubernetes_otel` ML module ([#19030](https://github.com/elastic/integrations/pull/19030)).

## Why — complements the threshold alerts, doesn't duplicate them

The shipped alert rules catch per-entity threshold breaches (a value crossing a fixed line). These ML jobs model each metric **per entity against its own history**, catching the *drift* those miss — e.g. a backend slowing down, hosts flapping in and out of health, or an error/rejection rate rising below the fixed count threshold. Each detector's description defers per-entity spikes to the alert rules, the same split `kubernetes_otel` uses. The detectors are drawn from the service's own signals and real failure modes — not tailored to any specific workflow.

## Jobs

- `aws_elb_target_latency_anomaly` — per `TargetGroup` (partition `cloud.region`): `high_mean` **TargetResponseTime** and **UnHealthyHostCount**.
- `aws_elb_request_error_anomaly` — per `LoadBalancer` (partition `cloud.region`), the per-bucket **Sum**: `high_mean` **HTTPCode_ELB_5XX_Count, HTTPCode_Target_5XX_Count, RejectedConnectionCount**.

Datafeeds are **composite-aggregated** — required, because these `metrics-aws.*.otel-*` indices contain `aggregate_metric_double` fields that a plain (non-aggregating) ML datafeed cannot read.

## Validation

Structural draft. The other services in this set were validated against live AWS OTel telemetry (baselines established; RDS scored 1.0 against a known incident); this one awaits a live load balancer to confirm field shape and detection.

Methodology, tooling, and the scoring harness: https://github.com/elastic/aws_otel_ml_draft

## Notes for reviewers (@elastic/obs-infraobs-integrations)

- **Draft** — proposing for your review; happy to adjust job naming, detector selection, `bucket_span`, or thresholds.
- Package stays `subscription: basic` (matches `kubernetes_otel`; ML availability is a deployment concern, not a package condition).
- Entity fields use the **raw CloudWatch dimensions** present in the indexed documents (`LoadBalancer` / `TargetGroup`), not the normalized fields the alert-rule `termField`s reference (those are not present in the documents).
- Unlike the other services, these jobs were **not validated against live data** (no load balancer was present in the test environment) — they are structurally consistent with the validated services and should be reviewed with that in mind.

### DIFF
--- a/packages/aws_elb_metrics_otel/changelog.yml
+++ b/packages/aws_elb_metrics_otel/changelog.yml
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "0.10.0"
+  changes:
+    - description: Add ML anomaly detection modules for ALB target health (response time, unhealthy hosts) and request errors (5XX, rejected connections).
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/19925
+    - description: Add ml_module to the Kibana asset tags.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/19925
 - version: "0.9.0"
   changes:
     - description: Add tags to Kibana assets
--- a/packages/aws_elb_metrics_otel/kibana/ml_module/aws_elb_metrics_otel-activity-ml.json
+++ b/packages/aws_elb_metrics_otel/kibana/ml_module/aws_elb_metrics_otel-activity-ml.json
@@ -1,0 +1,195 @@
+{
+    "id": "aws_elb_metrics_otel-activity-ml",
+    "type": "ml-module",
+    "migrationVersion": {
+        "search": "7.9.3"
+    },
+    "references": [],
+    "attributes": {
+        "id": "aws_elb_metrics_otel-activity-ml",
+        "title": "AWS Application Load Balancer errors (OpenTelemetry)",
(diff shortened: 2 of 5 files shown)


jakubgalecki0 (commented):
Lets please make sure that we follow naming convention of other `aws_otel` assets i.e. 

> `[<servicename> OTel] Description` 

Right now we have following convention: `"AWS Application Load Balancer errors (OpenTelemetry)"`

Also add entry for `tags` since `package-spec` allows it to be tagged - https://github.com/elastic/package-spec/blob/main/spec/integration/kibana/tags.spec.yml#L34

```
  asset_types:
    - dashboard
    - alerting_rule_template
    - slo_template
    - ml_module <<< 
```

Other than that it looks good.

jakubgalecki0 (approved):

ishleenk17 (approved):

PR #19927, opened 2026-07-02
crowdstrike: stop overwriting event.action with generic values in fdr data stream

### LINKED ISSUE #19807: [CrowdStrike]: event.action, event.type, and message
### Integration Name

CrowdStrike [packages/crowdstrike]

### Dataset Name

_No response_

### Integration Version

3.26.0

### Agent Version

8.19.0

Oracle 9.7

### User Goal

event.action is being abused for what event.type is for, while message is being used to plug a gap created by this abuse.

**script_set_event_action_and_type_29345ceb**
event.type creation cover what is being asked for and folds right into existing event.type. "creation" isn't the action. There is a need to track what type of registry action was performed, and .type is correctly being used for that -- why clone to action? 

**set_event_action_526984f8**
Similar problem. If OS=Windows and *Written, then event.action is "creation"? No, append "creation" to event.type and leave *Written in event.action alone.

**set_event_action_735cfe72**
Similar problem. This appears to be specific to a driver loading, so maybe "device", "installation", and/or "creation" cover the need, but it is an event type, not an action. "load" is non descriptive for what could be multiple separate unrelated actions.

With those changes undone, then message doesn't need to exist. This is a non ECS compliant field that exists just so, what CrowdStrike calls, SimpleEventName has a home? message is already consumed/used as part of event.original, and co-opting it is wrong. Should I want to utilize a visualization or query against the SimpleEventName, I can't do that with the message field (text) and I can't do it consistently in event.action, with crowdstrike.name NOT being the SimpleEventName but full event name (ProcessRollup2 vs ProcessRollup2[Platform][Version])

### Existing Features

See above.

### What did you see?

See Above.

### Anything else?

_No response_

efd6:
## Proposed commit message
```
crowdstrike: stop overwriting event.action with generic values in fdr data stream

Three processors in the FDR ingest pipeline overwrote event.action
with generic ECS vocabulary (load, modification, creation, deletion,
open, query) instead of preserving the CrowdStrike event_simpleName.
This lost the specificity needed by detection rules which filter on
the original SimpleEventName values.

Remove the driver event.action="load" processor, the *Written
event.action="creation" processor, and the event.action assignment
from the registry RegOperationType script. The event.type assignments
from these processors were already correct and are retained.

Remove the processor that copied event_simpleName into message. The
crowdstrike.event_simpleName keyword field serves the same purpose
without abusing a match_only_text field.

Add creation to the MachOFileWritten event.type in the categorize
pipeline, closing the one gap the *Written processor was covering.
```

> [!NOTE]
> This is a counter proposal to #19808.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Fixes #19807
- Closes #19808

### DIFF
--- a/packages/crowdstrike/changelog.yml
+++ b/packages/crowdstrike/changelog.yml
@@ -1,4 +1,14 @@
 # newer versions go on top
+- version: "4.0.0"
+  changes:
+    - description: >-
+        Restore FDR event.action to the CrowdStrike event_simpleName value (e.g. DriverLoad, AsepKeyUpdate, NewExecutableWritten) instead of overwriting it with generic terms. Remove redundant message field copy from event_simpleName. Add missing creation type to MachOFileWritten events.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/19927
+    - description: >-
+        FDR event.action values have changed. Custom queries, dashboards, or rules that filter on the previous generic values (load, modification, creation, deletion, open, query) in event.action must be updated to use the CrowdStrike event_simpleName instead. The message field is no longer populated from event_simpleName; use crowdstrike.event_simpleName for the same data.
+      type: breaking-change
+      link: https://github.com/elastic/integrations/pull/19927
 - version: "3.26.1"
   changes:
     - description: Remove incorrect unlimited maximum executions documentation.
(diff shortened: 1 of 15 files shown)


navnit-elastic (approved):
Nit only - Regenerate the `sample_event.json` to remove `message` field for consistency.

efd6:
@navnit-elastic, for the event that was chosen, `event.action` does not get populated, but I've gone through all the generated output from the system test and it is properly generated in most of them, what it looks like is happening is that a host metadata document was chosen.

navnit-elastic (approved):
LGTM, Thanks!

andrewkroh (approved):

PR #19933, opened 2026-07-02
Disable unowned AI GitHub Actions workflows

kcreddy:
## Proposed commit message

```
ci: disable unowned AI GitHub Actions workflows

The sweep-* and trigger-* workflows wrap reusable workflows from
elastic/ai-github-actions. They were set up by a contributor who has
since left, and no team currently owns them.

Left running unowned, they generate noise that no one is triaging:
100+ AI-authored issues opened since March with no team labels, issues
filed against deprecated integrations, and issues created when the bots
themselves error out. Disable them to stop the noise until ownership is
decided — rather than deleting them, since deletion would lose the
config and history and a team may want to adopt and re-enable them.

Disable via a source-level `if: false` on the `run` job (not the
Actions UI toggle) so that re-enabling requires a reviewed PR and is
visible in git history — a silent flip-back-on isn't possible.

Disabled 16 workflows:
  sweep-dashboard-data-scope, sweep-field-mapping-conflicts,
  sweep-httpjson-pagination, sweep-ingest-pipeline-safety,
  sweep-pipeline-error-handling, trigger-breaking-change-detector,
  trigger-bug-hunter, trigger-docs-patrol,
  trigger-duplicate-issue-detector, trigger-issue-triage,
  trigger-mention-in-issue, trigger-mention-in-pr,
  trigger-newbie-contributor-patrol, trigger-pr-review,
  trigger-text-auditor, trigger-stale-issues

For the 12 single-job workflows, `if: false` plus a comment was added
directly to the `run` job.

For the 4 workflows that already had an `if:` activation condition
(trigger-issue-triage, trigger-mention-in-issue, trigger-mention-in-pr,
trigger-pr-review), YAML permits only one `if:` key per job, so the
original condition was moved into a comment above the job and replaced
with `if: false`. The comment preserves the exact original expression
so it can be restored verbatim on re-enable.

Excluded (intentionally not disabled):
  - trigger-package-tests-security-ml: actively maintained by the
    security-ml team (and uses elastic/ci-gh-actions, not
    elastic/ai-github-actions).
  - trigger-pr-actions-detective: has an extra step-level job
    (actions/github-script) before the reusable-workflow call, so it
    needs separate handling rather than a blanket if: false.

Note: workflows still trigger on their existing events/schedules but
the job is skipped, producing empty "skipped" runs. This is the
accepted trade-off for keeping the disabled state in source.
```

## Checklist

### DIFF
--- a/.github/workflows/sweep-dashboard-data-scope.yml
+++ b/.github/workflows/sweep-dashboard-data-scope.yml
@@ -13,6 +13,9 @@ permissions:
 
 jobs:
   run:
+    # Disabled 2026-07-02 — no current owner. Re-enable by removing this
+    # `if: false` line via a reviewed PR so the change is visible.
+    if: false
     uses: elastic/ai-github-actions/.github/workflows/gh-aw-code-quality-audit.lock.yml@v0
     with:
       title-prefix: "[dashboard-data-scope]"
--- a/.github/workflows/sweep-field-mapping-conflicts.yml
+++ b/.github/workflows/sweep-field-mapping-conflicts.yml
@@ -13,6 +13,9 @@ permissions:
 
 jobs:
   run:
+    # Disabled 2026-07-02 — no current owner. Re-enable by removing this
+    # `if: false` line via a reviewed PR so the change is visible.
+    if: false
     uses: elastic/ai-github-actions/.github/workflows/gh-aw-code-quality-audit.lock.yml@v0
     with:
       title-prefix: "[field-mapping-conflicts]"
--- a/.github/workflows/sweep-httpjson-pagination.yml
+++ b/.github/workflows/sweep-httpjson-pagination.yml
@@ -13,6 +13,9 @@ permissions:
 
 jobs:
   run:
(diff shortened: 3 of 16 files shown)


mrodm:
We (@elastic/ecosystem) did not add those agent workflows. Adding as a reviewers @v1v and @theletterf in case they know about these workflows.

fr4nc1sc0-r4m0n:
If the problem is the noisy issues that are being generated by the execution failures, I think this https://github.com/elastic/ai-github-actions/pull/1397 could help to reduce the noise. 
On the contrary, If you think that some of them are not giving any value, I'd prefer to remove them because we keep the workflows directory as clean as possible, we don't waste resources in non desired executions and if necessary, we could always recover them because they're in the git history.
WDYT?

teresaromero (approved):
i would rather delete them if no one use this

v1v (commented):
I'd suggest to disable the workflows instead of jobs, otherwise, they will run on events without doing much:
- https://github.com/elastic/integrations/actions/runs/29329000650

Image

PR #19954, opened 2026-07-03
[salesforce] Order EventLogFile queries by the cursor field (CreatedDate)

shmsr:
## What does this PR do?

Orders the **Apex**, **Login**, and **Logout** `EventLogFile` queries by `CreatedDate` (the cursor field) instead of `LogDate`, in both the `default` and `value` queries.

## Background: two unrelated timestamps

Every `EventLogFile` record has two datetime fields that mean different things:

- **`LogDate`** — the period the log *covers* (e.g. the start of the hour/day).
- **`CreatedDate`** — when Salesforce actually *generated* the file. This lags `LogDate`, and the lag is **variable** (minutes to days), so `CreatedDate` order does not follow `LogDate` order.

These data streams track the collection cursor on `CreatedDate` (`cursor.field: CreatedDate`) and resume with `WHERE CreatedDate > <cursor>`, but the queries sorted the results by `LogDate`:

```sql
-- before
... WHERE CreatedDate > [[ .cursor.event_log_file.last_event_time ]] AND (<event types>) ORDER BY LogDate ASC NULLS FIRST
```

## The bug, with an example

The input watermarks the cursor from the **last record it processes** in a page. When the page is ordered by `LogDate`, that last record is not necessarily the one with the greatest `CreatedDate`.

Consider two log files where the one covering the later period happened to be generated first:

| # | LogDate (period) | CreatedDate (generated) |
|---|------------------|--------------------------|
| 1 | `2026-06-22T00:00:00Z` | `2026-06-24T11:28:06Z` |
| 2 | `2026-06-23T00:00:00Z` | `2026-06-24T11:08:05Z` |

`ORDER BY LogDate ASC` returns them as `#1` then `#2`. The input processes `#1` (sets watermark `11:28:06`), then `#2` (overwrites watermark with `11:08:05`). So the stored `last_event_time` ends at **`2026-06-24T11:08:05Z`** — *earlier* than a record it already ingested (`11:28:06`).

On the next poll:

```sql
WHERE CreatedDate > 2026-06-24T11:08:05Z ...
```

…record #1 (`11:28:06`) matches again and is **re-collected**. The cursor effectively lags behind the data and re-fetches already-ingested files each poll.

This is not a contrived case — sorting real `EventLogFile` results by `LogDate` produces multiple such `CreatedDate` inversions per page whenever files are generated slightly out of period order (common for hourly logs).

## The fix

Order by the cursor field so the last record always carries the maximum `CreatedDate`:

```sql
-- after
... WHERE CreatedDate > [[ .cursor.event_log_file.last_event_time ]] AND (<event types>) ORDER BY CreatedDate ASC NULLS FIRST
```

With the example above, `ORDER BY CreatedDate ASC` returns `#2` then `#1`, so the watermark ends at `2026-06-24T11:28:06Z` (the true maximum) and the next poll (`CreatedDate > 11:28:06`) does not re-collect either file. This matches the already-correct `SetupAuditTrail` query, which orders by its cursor field.

## Why is it important?

Prevents repeated re-collection of already-ingested `EventLogFile` data and keeps the collection cursor moving strictly forward.

## Compatibility / upgrade safety

The change is limited to the `ORDER BY` clause. The `WHERE` filter and `cursor.field` are unchanged, so:

- Existing persisted cursors remain valid `CreatedDate` values and are interpreted identically.
- For any given cursor value, the **exact same set of records is returned** — only their order (and the resulting watermark) changes — so no records are skipped on upgrade. A previously lagging cursor simply re-fetches its small lag window once and then advances cleanly.

## Checklist

- [x] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this change adheres to the guidelines.
- [x] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md#bump-the-kibana-version-constraint).

- `elastic-package lint` in `packages/salesforce`.
- `elastic-package test system -v` for the `apex`, `login`, and `logout` data streams (the mock server query matchers in `_dev/deploy/docker/files/config.yml` are updated to match the new ordering).


packages/salesforce/data_stream/apex/agent/stream/salesforce.yml.hbs
```diff
@@ -29,8 +29,8 @@ event_monitoring_method:
...
-      value: SELECT Id,CreatedDate,LogDate,LogFile FROM EventLogFile WHERE {{#if log_file_interval}}Interval = '{{log_file_interval}}' AND {{/if}}CreatedDate > [[ .cursor.event_log_file.last_event_time ]] AND (EventType = 'ApexCallout' OR EventType = 'ApexExecution' OR EventType = 'ApexRestApi' OR EventType = 'ApexSoap' OR EventType = 'ApexTrigger' OR EventType = 'ExternalCustomApexCallout') ORDER BY LogDate ASC NULLS FIRST
+      default: SELECT Id,CreatedDate,LogDate,LogFile FROM EventLogFile WHERE {{#if initial_interval}}LogDate > [[ (formatTime (now.Add (parseDuration "-{{initial_interval}}")) "2006-01-02T15:04:05.000Z0700") ]] AND {{/if}}{{#if log_file_interval}}Interval = '{{log_file_interval}}' AND {{/if}}(EventType = 'ApexCallout' OR EventType = 'ApexExecution' OR EventType = 'ApexRestApi' OR EventType = 'ApexSoap' OR EventType = 'ApexTrigger' OR EventType = 'ExternalCustomApexCallout') ORDER BY CreatedDate ASC NULLS FIRST
```
ishleenk17, line 32: @shmsr : 

So the initial lookback window is bounded by LogDate, but once the cursor is established, all subsequent polls use CreatedDate. 

Is that right ?

ishleenk17 (approved):
1 question. 
Otherwise looks good!

packages/salesforce/data_stream/apex/agent/stream/salesforce.yml.hbs
shmsr: Yes!

`LogDate` is the start of the interval the events occurred in (the hour/day the file covers). `CreatedDate` is when Salesforce generated the file. So, CreatedDate is now being used to find new files, including late ones as files are not always generated in LogDate order, and a later job can regenerate a file for an earlier LogDate.

So the first poll (no cursor) depends on LogDate for initial_interval. After that we cursor on CreatedDate.

PR #18786, opened 2026-05-04
[verifier_otel] Make policy_id, policy_name, package_policy_id multi:true

Omolola-Akinleye:
## Summary

Implements [elastic/security-team#17149](https://github.com/elastic/security-team/issues/17149) (parent epic [#16847](https://github.com/elastic/security-team/issues/16847)).

Each verification log emitted by the OTel verifier must carry the  integration's `package_policy.id` so Fleet background task can aggregate verifier logs by package-policy instance. Today, `policy_id`, `policy_name`, and `package_policy_id` are `multi: false`, which prevents the package manifest from multiple package policy id values when needed. We need to associate multiple permissions by package policy id(integration).

This PR flips those three fields to `multi: true` in `manifest.yml`. Per-target distribution itself is delivered by Kibana, which creates one `verifier_otel` package policy per target integration; so `input.yml.hbs` and the test fixtures are intentionally unchanged.

## Changes

- **`packages/verifier_otel/manifest.yml`**
  - `policy_id`, `policy_name`, `package_policy_id` flipped to `multi: true`.
  - Descriptions updated to note positional alignment with `policy_templates`.
  - Version bumped `0.0.1` → `0.1.0`.
  - These vars stay `multi: false`: `package_name`, `package_title`, `package_version`, `identity_federation_id`, `identity_federation_name`, `verification_id`, `verification_type`, `provider`, `account_type`, all `credentials_*`.
- **`packages/verifier_otel/changelog.yml`** — `0.1.0` entry added.

## Test plan

- [x] `elastic-package lint` — passes.
- [x] `elastic-package check` — package builds (`verifier_otel-0.1.0.zip`).
- [x] `elastic-package test static` — passes.
- [ ] `elastic-package test policy` — to be confirmed in CI.

PR #19965, opened 2026-07-03
[verifier_otel] Make policy_id, policy_name, package_policy_id multi:true

seanrathier:
## Update
- Rebased #18786 onto current `main` to resolve merge conflicts (main had already shipped `verifier_otel` 0.1.0 for an unrelated change).
- Bumps `verifier_otel` to `0.2.0` and consolidates the changelog entries for this change under that version.
- Same functional change as #18786: makes `policy_id`, `policy_name`, and `package_policy_id` `multi: true`, and requires Kibana `^9.5.0`.

Supersedes #18786 due to a fork push permission issue — original author is @Omolola-Akinleye.

## Summary

Implements [elastic/security-team#17149](https://github.com/elastic/security-team/issues/17149) (parent epic [#16847](https://github.com/elastic/security-team/issues/16847)).

Each verification log emitted by the OTel verifier must carry the  integration's `package_policy.id` so Fleet background task can aggregate verifier logs by package-policy instance. Today, `policy_id`, `policy_name`, and `package_policy_id` are `multi: false`, which prevents the package manifest from multiple package policy id values when needed. We need to associate multiple permissions by package policy id(integration).

This PR flips those three fields to `multi: true` in `manifest.yml`. Per-target distribution itself is delivered by Kibana, which creates one `verifier_otel` package policy per target integration; so `input.yml.hbs` and the test fixtures are intentionally unchanged.

## Changes

- **`packages/verifier_otel/manifest.yml`**
  - `policy_id`, `policy_name`, `package_policy_id` flipped to `multi: true`.
  - Descriptions updated to note positional alignment with `policy_templates`.
  - Version bumped `0.1.0` → `0.2.0`.
  - These vars stay `multi: false`: `package_name`, `package_title`, `package_version`, `identity_federation_id`, `identity_federation_name`, `verification_id`, `verification_type`, `provider`, `account_type`, all `credentials_*`.
- **`packages/verifier_otel/changelog.yml`** — `0.2.0` entry added.

## Test plan

- [x] `elastic-package lint` — passes.
- [x] `elastic-package check` — package builds (`verifier_otel-0.1.0.zip`).
- [x] `elastic-package test static` — passes.
- [ ] `elastic-package test policy` — to be confirmed in CI.


packages/verifier_otel/changelog.yml
```diff
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "0.2.0"
+  changes:
+    - description: "Render one `policies[]` entry per target via parallel `multi: true` arrays (`policy_id`, `policy_name`, `policy_templates`, `package_policy_id`, `package_name`, `package_title`, `package_version`)."
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/19965
+    - description: "Require Kibana 9.5.0+ (`conditions.kibana.version: ^9.5.0`) so multi-target policy compilation is available at install time."
+      type: breaking-change
```
bhapas, line 8: Why is this a `breaking-change`? Since we are bumping the min Kibana version to 9.5 it should compile on upgrade to 9.5 isn't it?

seanrathier: When I took over this PR I questioned why it was a breaking change also until I saw that we are changing these from text fields to `multi` set to `true`.

We need to do a [version check on this package](https://github.com/elastic/kibana/pull/276100/changes#diff-97d64f70a79cbf9f86c64afb207c4e085e6236ade1a7d80d911d2f5ce8e6692cR61) in Kibana to handle the arrays.

https://github.com/elastic/integrations/pull/19965/changes#diff-83d9dd23ca61f735253f325eb8e7753bcc37599c9ad44d5c5663a07adc733e62R154

bhapas: Makes sense.. Thanks for the answer.

PR #20001, opened 2026-07-06
mimecast: apply grace period to threat intel feed end timestamp

efd6:
## Proposed commit message
```
mimecast: apply grace period to threat intel feed end timestamp

The Mimecast /api/ttp/threat-intel/get-feed endpoint rejects requests
where the end timestamp is at or beyond its server's current time,
returning err_validation_enddate_in_the_future. Because the integration
was sending end=now(), any clock skew between the agent host and
Mimecast's servers could trigger persistent failures.

Subtract a 1-minute grace period from now when computing the end
timestamp for both the CEL and httpjson input types. The CEL programs
hold the grace value in a state field (end_grace) so it can be made
user-configurable later if needed. The cursor fallback is also
adjusted to match the collection window so events in the grace period
are not skipped on the next cycle.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/mimecast/changelog.yml
+++ b/packages/mimecast/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.4.4"
+  changes:
+    - description: Apply grace period to threat intel feed end timestamp to avoid Mimecast API rejecting requests due to clock skew.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/20001
 - version: "3.4.3"
   changes:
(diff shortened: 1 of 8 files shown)


navnit-elastic (approved):
When a user configures an interval shorter than 1m (for example `30s`), `start` is more recent than `end`, producing an invalid time window that Mimecast is likely to reject. worth a manifest note, since the default `5m` interval won’t hit it in practice.

navnit-elastic (approved):
LGTM, Thanks!

PR #20005, opened 2026-07-07
[splunk/alert] -  Keep non-numeric `proto` as ECS field `network.transport`

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
splunk/alert: keep non-numeric proto as network.transport

The convert_proto_to_long processor forced splunk.alert.proto to a
long. Palo Alto Networks and other CEF sources send a transport name
such as "tcp" in this field, so the conversion failed, appended to
error.message, tagged the event as pipeline_error, and dropped the
value.

Handle the non-numeric case instead of failing: copy the value to ECS
network.transport (lowercased) and remove splunk.alert.proto without
recording an error. Numeric values still convert to long, so the field
mapping is unchanged and the fix is not a breaking change.

Replace the transport lowercase step with a script that converts
between the IANA protocol number and the transport name in both
directions, so a named proto (tcp) also yields network.iana_number
and a numeric proto (6) also yields network.transport. The lookup
tables live in the processor params, and the script is guarded to run
only when a protocol value is present. Numeric proto is still kept in
splunk.alert.proto, so the change is additive and not breaking.

Add a synthetic pipeline regression test covering both a protocol name
and a numeric protocol number.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/splunk/data_stream/alert/elasticsearch/ingest_pipeline/default.yml
efd6, line 1: Is it worth doing an IANA protocol number look-up if the convert succeeds? This would bring parity to the two forms.

ShourieG: But this would mean we would need a reverse lookup when convert fails right ?

efd6: No, not really. The field that would be populated by the look-up is an ECS field, so it has a greater value.

ShourieG:
@efd6, updated with IANA look-up

efd6 (approved):

Issue #13729, opened 2025-04-30
[Azure Frontdoor]: Add support for health probe log


navnit-elastic:
### Integration Name

Azure Frontdoor [azure_frontdoor]

### Dataset Name

health_probe

### Integration Version

2.2.0

### Agent Version

0.110.2

### OS Version and Architecture

Ubuntu 20.04 LTS (x84_64)

### User Goal

Add support for [Health Probe log](https://learn.microsoft.com/en-us/azure/frontdoor/monitor-front-door?pivots=front-door-standard-premium#health-probe-log) through the Azure Frontdoor integration.
- Relates: #7017 

### Existing Features

Currently, the integration only supports Access Logs and Web Application Firewall (WAF) Logs and does not support Health Probe Logs.

### What did you see?

The integration does not support for collecting Health Probe log.

navnit-elastic:
Support for health probe log added in https://github.com/elastic/integrations/pull/20016.

PR #20016, opened 2026-07-07
[Azure frontdoor] Add Health Probe dataset

herrBez:
## Proposed commit message

Add first support for Frontdoor health probe logs 

## Checklist

## Author's Checklist

- [ ] Double check the documentation
- [ ] Decide whether to activate the dataset by default or not (currently it's active by default)

- Configure an Azure Frontdoor and export the logs to an Eventhub
- Configure the integration
- Verify that Health Check logs are collected

## Related issues

- https://github.com/elastic/integrations/issues/13729

## Screenshots

image


packages/azure_frontdoor/data_stream/health_probe/_dev/test/pipeline/test-common-config.yml
```diff
@@ -0,0 +1,6 @@
+dynamic_fields:
+  "event.ingested": ".*"
```
efd6: `event.ingested` is not being set in the pipeline tests, so we don't need this.

packages/azure_frontdoor/data_stream/health_probe/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,197 @@
...
+      field: message
+      target_field: event.original
+      ignore_missing: true
+      if: 'ctx.event?.original == null'
+  - set:
+      field: cloud.provider
+      value: azure
+  - drop:
+      description: Drop if invalid json
+      if: 'ctx.event?.original != null && ctx.event.original.contains(''"records"'')'
```
efd6:
```suggestion
      if: ctx.event?.original != null && ctx.event.original.contains('"records"')
```

packages/azure_frontdoor/data_stream/health_probe/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,197 @@
...
+      field: event.category
+      value: [network]
+  - set:
+      field: event.type
+      value: [connection]
+  - rename:
+      field: message
+      target_field: event.original
+      ignore_missing: true
+      if: 'ctx.event?.original == null'
```
efd6:
```suggestion
      if: ctx.event?.original == null
```

packages/azure_frontdoor/data_stream/health_probe/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,197 @@
...
+  - set:
+      field: event.kind
+      value: pipeline_error
+  - append:
+      field: tags
+      value: preserve_original_event
+      allow_duplicates: false
+  - append:
+      field: error.message
+      value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
```
efd6:
```suggestion
  - set:
      field: event.kind
      tag: set_pipeline_error_into_event_kind
      value: pipeline_error
      if: ctx.error?.message != null
  - append:
      field: tags
      tag: append_tags
      value: preserve_original_event
      allow_duplicates: false
      if: ctx.error?.message != null
on_failure:
  - set:
      field: event.kind
      value: pipeline_error
  - append:
      field: tags
      value: preserve_original_event
      allow_duplicates: false
  - append:
      field: error.message
      value: >-
        Processor '{{{ _ingest.on_failure_processor_type }}}'
        {{{#_ingest.on_failure_processor_tag}}}with tag '{{{ _ingest.on_failure_processor_tag }}}'
        {{{/_ingest.on_failure_processor_tag}}}failed with message '{{{ _ingest.on_failure_message }}}'
```

packages/azure_frontdoor/data_stream/health_probe/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,197 @@
...
+  - remove:
+      field:
+        - azure.frontdoor.health_probe.time
+        - azure.frontdoor.health_probe.properties
+      ignore_missing: true
+  - geoip:
+      field: destination.ip
+      target_field: destination.geo
+      ignore_missing: true
+  - geoip:
```
efd6, line 194: Always use an `if` condition for field presence for `geoip` processors.

efd6 (commented):
Can you update the README?

packages/azure_frontdoor/data_stream/health_probe/_dev/test/pipeline/test-common-config.yml
herrBez: Thanks. Yes, you are right. I copied from the other datasets, but it's probably a legacy think. Let me fix it :)

packages/azure_frontdoor/data_stream/health_probe/elasticsearch/ingest_pipeline/default.yml
herrBez: Is ignore_missing not sufficient? I copied it from the other datasets and for example the apache integration. I don't mind adding the condition. Let me do that.

packages/azure_frontdoor/data_stream/health_probe/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,213 @@
...
+            value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - set:
+      field: url.full
+      copy_from: url.original
+      ignore_empty_value: true
+  - rename:
+      field: azure.frontdoor.health_probe.properties.originName
+      target_field: azure.frontdoor.health_probe.origin_name
+      ignore_missing: true
+  - rename:
```
navnit-elastic: Azure [documents](https://learn.microsoft.com/en-us/azure/frontdoor/monitor-front-door?pivots=front-door-standard-premium#health-probe-log) `originIP` as `IP:port`, so values with a port suffix are invalid for ECS `ip` and break GeoIP/ASN enrichment. Should we split IP and port in such cases?
```
  - grok:
      field: destination.ip
      patterns:
        - '%{IP:destination.ip}:%{NUMBER:destination.port}'
      ignore_missing: true
      if: ctx.destination?.ip != null && ctx.destination.ip.contains(':')
      on_failure:
        - append:
            tag: append_error_message_e18f0b9e
            field: error.message
```

herrBez: Hey, thanks for the feedback, this documeantion https://learn.microsoft.com/en-us/azure/frontdoor/monitor-front-door?pivots=front-door-standard-premium#health-probe-log and my tests seem to confirm that originIP is an IP. 

May I ask you where you found the information?

herrBez: Sorry, I cannot even read the examples of the doc 🤣 (to be fair the table and the examples don't report the same information xD) Let me fix that

herrBez:
```
ctx.destination?.ip != null && ctx.destination.ip.contains(':')
```

Is it not a risky condition? What if the IP is IPv6?

navnit-elastic: Good catch. Yes, the condition is too broad and risky for IPv6.
A safer approach is to use `convert` first, then fall back to grok for host:port forms.

herrBez: OK, I tried my best :) Can you double check my proposal?

packages/azure_frontdoor/data_stream/health_probe/elasticsearch/ingest_pipeline/default.yml
efd6: Unfortunately not, there's a bunch of work that happens before the `ignore_missing` state is used, so there ir a measurable performance impact from running the processor at all. This may change in future, but for the moment, this is important.

packages/azure_frontdoor/data_stream/health_probe/elasticsearch/ingest_pipeline/default.yml
efd6: Take a look at https://github.com/elastic/integrations/blob/49864dbf209d4c40579f93f7a1e2fb900140a823/packages/cisco_duo/data_stream/auth/elasticsearch/ingest_pipeline/default.yml#L89-L101

navnit-elastic: Thanks @efd6.

@herrBez, Sorry for the back and forth. Rather than using grok inside the `convert` `on_failure` handler, we can:
1. Run grok with explicit anchored patterns (`IPv4`, `IPv4:port`, `[IPv6]:port`, etc.)
2. Validate with `convert` after extraction (`remove` the field on failure)

That avoids the need for `contains(':')`, while handling `IP:port`.

herrBez: Can you double check the new version (I was on PTO Friday 😓 )

efd6 (approved):

navnit-elastic (approved):
LGTM, Thanks!

herrBez:
Can you please proceed with the merge? I don't have the rights to do so 😢 (as I used to have)

PR #20017, opened 2026-07-07
[Osquery_manager] Add pack_name and query_name fields for scheduled queries

marc-gr:
## What does this PR do?

Adds two keyword fields to the `osquery_manager` package data streams:

- `pack_name` — human-readable pack name for scheduled queries that belong to a pack.
- `query_name` — the query name (the pack's `queries` config map key) for scheduled queries that belong to a pack.

Both fields are added to:
- `data_stream/result` (`primary-fields.yml`) — osquery result documents.
- `data_stream/action_responses` (`action-responses.yml`) — synthetic scheduled response documents.

The legacy pack dashboards and saved searches (Compliance pack, OSSEC rootkit pack) previously matched scheduled queries by `action_id` only. They are updated to match `action_id` **OR** (`pack_name` **AND** `query_name`) so they keep working for both legacy documents and documents produced after the beats change.

Package version bumped `1.30.0` → `1.30.1` with `bugfix` changelog entries.

## Why is it important?

Osquerybeat now emits `pack_name` and `query_name` on scheduled pack query result and response documents (companion change on the beats side), and no longer sets `action_id` for scheduled queries. The Osquery Manager dashboards group and label scheduled query results by pack and query name; without mappings for these fields and updated filters, the dashboards cannot render those results correctly.

## Checklist

- [ ] I have verified that Kibana version constraints are current according to the [Kibana version support policy](https://github.com/elastic/integrations/blob/main/docs/kibana_version_support_policy.md).

1. Build the package (`elastic-package build`) and boot the stack (`elastic-package stack up -d`).
2. Install the package (`elastic-package install`).
3. Ingest scheduled pack query documents in both shapes into `logs-osquery_manager.result-default`:
   - legacy: `action_id: "pack_<pack>_<query>"`
   - new: `pack_name`, `query_name`, `schedule_id` (no `action_id`), with realistic `osquery.*` fields per query.
4. Open the two dashboards and the five saved searches and confirm both legacy and new documents are matched.

## Verification performed

Verified against a local `elastic-package` stack (Elasticsearch/Kibana `9.5.0-SNAPSHOT`) with package `1.30.1` installed. Realistic per-query documents were ingested in both shapes (legacy `action_id` and new `pack_name`+`query_name` with `action_id` absent), then the dashboards and saved searches were driven headlessly with Playwright.

Saved searches (Discover) — each returns both shapes with the expected columns populated:

| Saved search | Docs (legacy + new) | Filter type |
| --- | --- | --- |
| Mounts | 6 (3 + 3) | custom `bool.should` |
| DEB packages installed | 6 (3 + 3) | custom `bool.should` |
| OSSEC Rootkits | 6 (3 + 3) | custom `bool.should` |
| OS versions | 4 (2 + 2) | KQL OR |
| Kernel integrations | 4 (2 + 2) | KQL OR |

Dashboards — **0 "No results" panels and 0 panel errors** on both:
- **OSSEC rootkit pack** — the "hosts infected" metric counts both legacy and new hosts; the embedded search shows legacy rows (`action_id = pack_ossec-rootkit`) and new rows (`action_id = (null)`, matched via `pack_name`).
- **Compliance pack** — the Mounts-by-type donut, OS-versions donut, "Live Kernel integrations" gauge, and the DEB packages / Mounts tables all populate from both document shapes.

New documents show `action_id = (null)` in Discover yet still appear, confirming they match on `pack_name` + `query_name`, while legacy documents continue to match on `action_id`.

## Related issues

-

### DIFF
--- a/packages/osquery_manager/data_stream/result/fields/primary-fields.yml
+++ b/packages/osquery_manager/data_stream/result/fields/primary-fields.yml
@@ -9,9 +9,11 @@
   description: Pack identifier for scheduled queries that belong to a pack.
 - name: pack_name
   type: keyword
+  ignore_above: 1024
   description: Human-readable pack name for scheduled queries that belong to a pack.
 - name: query_name
   type: keyword
+  ignore_above: 1024
   description: Query name for scheduled queries. For pack queries this is the pack's queries config map key; for top-level scheduled queries it is the schedule config map key. Absent for live queries.
 - name: response_id
   type: keyword
--- a/packages/osquery_manager/data_stream/result/fields/primary-fields.yml
+++ b/packages/osquery_manager/data_stream/result/fields/primary-fields.yml
@@ -12,7 +12,7 @@
   description: Human-readable pack name for scheduled queries that belong to a pack.
 - name: query_name
   type: keyword
-  description: Query name (the pack's queries config map key) for scheduled queries that belong to a pack.
+  description: Query name for scheduled queries. For pack queries this is the pack's queries config map key; for top-level scheduled queries it is the schedule config map key. Absent for live queries.
 - name: response_id
   type: keyword
   description: Query response id. Present only for scheduled queries.
--- a/packages/osquery_manager/data_stream/result/fields/primary-fields.yml
+++ b/packages/osquery_manager/data_stream/result/fields/primary-fields.yml
@@ -7,6 +7,12 @@
 - name: pack_id
   type: keyword
   description: Pack identifier for scheduled queries that belong to a pack.
+- name: pack_name
+  type: keyword
+  description: Human-readable pack name for scheduled queries that belong to a pack.
+- name: query_name
+  type: keyword
+  description: Query name (the pack's queries config map key) for scheduled queries that belong to a pack.
 - name: response_id
   type: keyword
   description: Query response id. Present only for scheduled queries.
--- a/packages/osquery_manager/manifest.yml
+++ b/packages/osquery_manager/manifest.yml
@@ -1,7 +1,7 @@
 format_version: 3.5.8
 name: osquery_manager
 title: Osquery Manager
-version: 1.30.0
+version: 1.30.1
 description: Deploy Osquery with Elastic Agent, then run and schedule queries in Kibana
 type: integration
 categories:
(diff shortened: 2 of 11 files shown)


packages/osquery_manager/data_stream/result/fields/primary-fields.yml
```diff
@@ -7,6 +7,12 @@
 - name: pack_id
   type: keyword
   description: Pack identifier for scheduled queries that belong to a pack.
+- name: pack_name
```
tomsonpl, line 10: Small consistency check: here in the `result` data stream the new fields are `keyword` without `ignore_above`, while the same two fields in `data_stream/action_responses/fields/action-responses.yml` set `ignore_above: 1024`. Is that difference intentional, or would it be worth aligning the two definitions?

packages/osquery_manager/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 3.5.8
 name: osquery_manager
 title: Osquery Manager
-version: 1.30.0
+version: 1.30.1
```
tomsonpl, line 4: The checklist notes the Kibana version constraints weren't verified. Since these fields only populate once the companion Kibana (elastic/kibana#276690) and osquerybeat (elastic/beats#51781) changes ship, is the current `conditions.kibana.version` still appropriate for `1.30.1`, or should it be bumped to align with the stack version that starts emitting `pack_name` / `query_name`?

tomsonpl (approved):
Reviewed alongside the Kibana (elastic/kibana#276690) and osquerybeat (elastic/beats#51781) PRs. The `pack_name` / `query_name` mappings and the `action_id` OR (`pack_name` AND `query_name`) filter migration match the fields osquerybeat now emits, and keeping the legacy `action_id` clause means old documents still resolve — that looks like the right backwards-compatible approach to me. Two small things I wanted to confirm, inline.

packages/osquery_manager/manifest.yml
marc-gr: No bump needed — the change is backward compatible. Mappings are additive and the dashboards still match legacy docs via the action_id OR-branch, so the package works on stacks without the beats/Kibana changes; the new fields just start populating once those ship. Bumping would only block older stacks from a package that works fine for them.

PR #20018, opened 2026-07-07
[ti_misp] Fix daily_refetch never re-triggering after the first poll

kcreddy:
## Proposed commit message
```
[ti_misp] Fix daily_refetch never re-triggering after the first poll

The daily_refetch feature relies on a `last_daily_refetch` cursor value
that should persist across polls and only advance once 24h has elapsed.
It never did: httpjson evaluates cursor value templates with an empty
cursor context (updateCursor() builds an emptyTransformContext(), so
`.cursor` is `{}`), which means the `last_daily_refetch` template could
never read its own previous value. It fell through to writing "now" on
every poll, so the 24h gate (last_daily_refetch + 24h < now) was pushed
into the future each cycle and the re-fetch window never rewound to
initial_interval. Decay scores were never re-ingested.

Fix it entirely within the package. The request transform runs with the
real loaded cursor, so it decides whether this poll is a re-fetch and
writes the resulting marker into url.params.daily_refetch_at. That value
round-trips into .last_response.url.params, which IS available when the
cursor value template is evaluated, so the cursor field now reads the
marker back from there instead of from its own (empty) prior state. The
marker is carried across pagination and falls back to "now" so the
cursor is never stored empty.

The extra url param is sent to MISP but ignored: restSearch reads its
filters from the POST JSON body, not the query string. This mirrors the
existing url.params.timestamp workaround.

Verified with elastic-package lint/format and the threat_attributes
system tests (including daily-refetch). The 24h re-trigger itself is not
covered by an automated test.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


packages/ti_misp/data_stream/threat_attributes/fields/fields.yml
```diff
@@ -294,3 +294,4 @@
           type: flattened
           description: >
             List of attributes of the object in which the attribute is attached.
+
```
kcreddy, line 297: whitespace change due to `elastic-package build`

efd6 (approved):
unfortunate

PR #20029, opened 2026-07-08
[akamai] Add Support for Identity Fields from SIEM Events

### LINKED ISSUE #17728: [ Akamai SIEM ] Identity fields are not being parsed.
Hello,

There are a couple of fields that are returned by the Akamai API that are not being parsed, like the `identity` fields.

From their [example](https://techdocs.akamai.com/siem-integration/reference/api), we have these fields:

```
  "identity": {
    "ja4": "t13d201100_2b729b4bf6f3_9e7b989ebec8",
    "tlsFingerprintV2": "46008b1582967146",
    "tlsFingerprintV3": "3~fe38c35477967146"
  }
```

They are not being parsed.

To be able to parse those fields the user needs to parse the `event.original` field again in the custom pipeline into a temporary filter and extract those fields since the temporary field is removed before the custom ingest pipeline is called.

I have a custom pipeline parsing those fields into :

- `akamai.siem.identity.ja4`
- `akamai.siem.identity.tls_fingerprint_v2`
- `akamai.siem.identity.tls_fingerprint_v3`

But they should be added in the default ingest pipeline.

mohitjha-elastic:
## Proposed commit message

```
akamai: add support for identity fields in SIEM events

This change adds support for ingesting identity fields from Akamai SIEM events.
The new fields are parsed and mapped to the appropriate schema to provide additional
identity context for the events.
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/akamai directory.

> elastic-package test -v

## Related Issues

- Closes https://github.com/elastic/integrations/issues/17728

### DIFF
--- a/packages/akamai/changelog.yml
+++ b/packages/akamai/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Add support for identity fields from Akamai SIEM events.
       type: enhancement
-      link: https://github.com/elastic/integrations/pull/1
+      link: https://github.com/elastic/integrations/pull/20029
 - version: "3.1.1"
   changes:
     - description: Fix the issue of populating HTTP message headers.
--- a/packages/akamai/changelog.yml
+++ b/packages/akamai/changelog.yml
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "3.2.0"
+  changes:
+    - description: Add support for identity fields from Akamai SIEM events.
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/1
 - version: "3.1.1"
   changes:
     - description: Fix the issue of populating HTTP message headers.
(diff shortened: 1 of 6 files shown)


efd6 (approved):
I'm wondering why we're not putting the `json` processor's result directly into `akamai.siem` directly; this would mean that if the API changes the fields that it provides, we just naturally follow. This is fine, but we'd need to do less work in future if we just followed the API.

mohitjha-elastic:
> I'm wondering why we're not putting the `json` processor's result directly into `akamai.siem` directly; this would mean that if the API changes the fields that it provides, we just naturally follow. This is fine, but we'd need to do less work in future if we just followed the API.

I agree that populating the JSON processor output directly would reduce the maintenance effort, as any new fields introduced by the API would be ingested automatically without requiring additional mapping changes.

The trade-off is only that we would have less control over the schema and field mappings. However, I think the reduced maintenance and automatic adoption of new fields outweigh those concerns.

PR #20042, opened 2026-07-08
ThreatQ Integration | Fixes Issue Pulling Sources & Attributes

zach-threatq:
**Bug Fix**

## Proposed commit message

fix: issue returning sources & attributes from ThreatQ

ThreatQ recently updated the default API response from the Threat Library to exclude `attributes` and `sources` by default. The original API response can be maintained by adding the `fields` URL parameter. In addition to the default fields (`*`), we can ask for the `attributes` and `sources` fields, which will restore full functionality of the integration & mappings. This commit also specifically asks for attribute names and values to de-clutter the API response.

## Checklist

## Author's Checklist

_N/A_

Using a ThreatQ instance on the latest version, ensure that the list of `sources` and `attributes` properly get parsed into an Elastic index record with the default integration configuration.

## Related issues

_N/A_

## Screenshots

_N/A_


andrewkroh:
Is this change backward compatible with earlier versions of ThreatQ? Let's say a user is connected to an earlier version, then will the new query param on the GET request work the same?

> The original API response can be maintained by adding the fields URL parameter.

Can you point me at the documentation for the API that this integration is using? I was looking at https://helpcenter.threatq.com/assets/HTML/api/ThreatQ_REST_API_Reference_v6.16.0.html and didn't see the API endpoint.

zach-threatq:
> Is this change backward compatible with earlier versions of ThreatQ? Let's say a user is connected to an earlier version, then will the new query param on the GET request work the same?
> 
> > The original API response can be maintained by adding the fields URL parameter.
> 
> Can you point me at the documentation for the API that this integration is using? I was looking at https://helpcenter.threatq.com/assets/HTML/api/ThreatQ_REST_API_Reference_v6.16.0.html and didn't see the API endpoint.

The `fields` parameter has existed in the Threat Library API for quite some time, dating back to 2023 (ThreatQ v5.23.0). So this should be backwards compatible with most current versions of ThreatQ. I made the `fields` list configurable just in case there needs to be some tweaks to it for older versions or newer versions, but from what I understand, it should be compatible with any ThreatQ version >= v5.23.0.

As far as documentation goes, I don't believe there are public docs outlining the endpoints in full detail, just high level.

andrewkroh:
> As far as documentation goes, I don't believe there are public docs outlining the endpoints in full detail, just high level.

Even if it's high level info, can you point me toward where I would find a doc for `/api/indicators/query/hash/{{data_collection_id}}` if any exists. Thanks.

packages/ti_threatq/data_stream/threat/manifest.yml
```diff
@@ -95,6 +95,17 @@ streams:
...
+        type: text
+        title: Threat Library Fields
+        multi: false
+        required: true
+        show_user: true
+        default: "*,sources,attributes,attributes.name,attributes.value"
+        description: >-
+          Comma-separated list of fields to retrieve from the Threat Library.
+          An asterisk (*) can be used to retrieve all default fields. In ThreatQ v6.12 and later,
+          sources & attributes are no longer part of the default fields and must be explicitly requested.
```
andrewkroh:
```suggestion
          sources and attributes are no longer part of the default fields and must be explicitly requested.
```

packages/ti_threatq/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.39.2"
+  changes:
+    - description: Restores `sources` & `attributes` fields after recent updates to the ThreatQ Platform's API.
```
andrewkroh:
```suggestion
    - description: Restores `sources` and `attributes` fields after recent updates to the ThreatQ Platform's API.
```

jamiehynds:
@zach-threatq are you ok to proceed with the suggestions made by @andrewkroh and also provide the documentation he requested [here](https://github.com/elastic/integrations/pull/20042#issuecomment-4918700231). We have a few customers awaiting the integration update to support your latest API, so keen to get this PR merged. Thanks!

zach-threatq:
> @zach-threatq are you ok to proceed with the suggestions made by @andrewkroh and also provide the documentation he requested [here](https://github.com/elastic/integrations/pull/20042#issuecomment-4918700231). We have a few customers awaiting the integration update to support your latest API, so keen to get this PR merged. Thanks!

Yes, sorry, just been busy on my end and haven't had a chance to look back at this. I'll check it out and make the changes!

andrewkroh (approved):

andrewkroh:
Tests are failing with:

`error: could not add data stream config to policy: could not create package policy (req {"name":"ti_threatq-threat-73707","description":"","namespace":"73707","policy_id":"424faf96-d64a-4e67-a61e-7fd26f37f9c9","package":{"name":"ti_threatq","version":"1.39.2"},"inputs":{"ti_threatq-httpjson":{"enabled":true,"streams":{"ti_threatq.threat":{"enabled":true,"vars":{"client_id":"test","client_secret":"test","data_collection_id":"someid","enable_request_tracer":true,"host":"http://svc-threatq:8080","interval":"10m","ioc_expiration_duration":"5d","preserve_original_event":true,"token_url":"http://svc-threatq:8080/api/token"}}}}},"force":false}); API status code = 400; response body = {"statusCode":400,"error":"Bad Request","message":"Package policy is invalid: inputs.httpjson.streams.ti_threatq.threat.vars.threat_library_fields: Strings starting with special YAML characters like * or & need to be enclosed in double quotes."}`

andrewkroh:
This is a known Fleet issue: https://github.com/elastic/kibana/issues/184972

Fleet rejects text vars whose value starts with `*` or `&` at package-policy create time. Quoting the value in the UI does not work, because the quotes become part of the stored string.

Workaround: reorder the default so it does not start with `*`:

```yaml
default: "sources,attributes,attributes.name,attributes.value,*"
```

instead of:

```yaml
default: "*,sources,attributes,attributes.name,attributes.value"
```

packages/ti_threatq/data_stream/threat/manifest.yml
```diff
@@ -95,6 +95,17 @@ streams:
         show_user: false
         title: Page size
         description: Maximum number of records to pull in one request. The maximum supported page size value is 1000.
+      - name: threat_library_fields
+        type: text
+        title: Threat Library Fields
+        multi: false
+        required: true
+        show_user: true
+        default: "*,sources,attributes,attributes.name,attributes.value"
```
andrewkroh:
```suggestion
        default: "sources,attributes,attributes.name,attributes.value,*"
```

PR #20052, opened 2026-07-09
[qualys_vmdr] record missing KB entries in knowledge_base.status instead of error.message

### LINKED ISSUE #19961: [qualys_vmdr] "QID not found in the Knowledge Base" wrongly marks documents as pipeline errors
## Integration

Qualys VMDR — `asset_host_detection` data stream.

## Description

When a vulnerability detection's QID is not present in the returned Knowledge Base data, the ingest pipeline appends a message to `error.message`:

[`packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml#L1674-L1678`](https://github.com/elastic/integrations/blob/66eec1b731443eda5490e724452b49c81031cc98/packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml#L1674-L1678)

```yaml
  # Handle KNOWLEDGE_BASE
  - append:
      field: error.message
      if: ctx.json?.KNOWLEDGE_BASE == null || ctx.json.KNOWLEDGE_BASE.size() == 0
      value: 'Qualys QID not found in the Knowledge Base'
```

Because the message is written to `error.message`, it then triggers the pipeline's global error-classification logic, which sets `event.kind: pipeline_error` and tags the document `preserve_original_event`:

[`packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml#L1879-L1889`](https://github.com/elastic/integrations/blob/66eec1b731443eda5490e724452b49c81031cc98/packages/qualys_vmdr/data_stream/asset_host_detection/elasticsearch/ingest_pipeline/default.yml#L1879-L1889)

```yaml
  - set:
      field: event.kind
      tag: set_event_kind_2
      value: pipeline_error
  - append:
      field: tags
      tag: append_tags_preserve_original_event_1
      allow_duplicates: false
```

A QID that isn't in the Knowledge Base is a **normal data condition**, not a document-processing failure. Treating it as `pipeline_error` is misleading and creates alert noise.

## Impact

This is especially bad when the Qualys API is unavailable or under maintenance. In that state the KB fetch returns partial or empty data, so most detections have no matching KB entry. The CEL program sets an empty `KNOWLEDGE_BASE` object per detection, which means the majority of ingested documents get:

- `error.message = "Qualys QID not found in the Knowledge Base"`
- `event.kind: pipeline_error`
- a `preserve_original_event` tag

Any monitoring or detection logic that keys off `error.message` or `event.kind: pipeline_error` (e.g. integration/data-collection health checks) then fires on a large share of documents, even though nothing actually failed to process.

## Expected behaviour

A missing KB entry should not classify a document as a pipeline error. The diagnostic should be recorded in a package-specific field rather than the ECS `error.message` field, so it doesn't feed into error/pipeline_error-based alerting.

`asset_host_detection` already declares a field for this purpose that is currently unpopulated in this data stream:

[`packages/qualys_vmdr/data_stream/asset_host_detection/fields/fields.yml#L466-L467`](https://github.com/elastic/integrations/blob/66eec1b731443eda5490e724452b49c81031cc98/packages/qualys_vmdr/data_stream/asset_host_detection/fields/fields.yml#L466-L467)

```yaml
        - name: error
          type: keyword
```

(`qualys_vmdr.asset_host_detection.knowledge_base.error`). The sibling `knowledge_base` data stream already uses `knowledge_base.error` for an analogous "no data" diagnostic.

## Proposed fix

Write the message to `qualys_vmdr.asset_host_detection.knowledge_base.error` instead of `error.message`, so a missing KB entry:

- is still recorded for troubleshooting, but
- no longer sets `event.kind: pipeline_error` or the `preserve_original_event` tag.

Genuine KB decode/API failures are handled separately in the CEL error path and would continue to populate `error.message` as expected.

## Notes / open questions

- Field naming: `knowledge_base.error` still reads like an error. A more neutral field (e.g. a status keyword or boolean flag) may signal intent better, at the cost of adding a new mapping. Reusing the existing `knowledge_base.error` is the lower-churn option.
- This is a behaviour change for anyone alerting on the current `error.message` text or on `event.kind: pipeline_error` for these documents; it needs a changelog entry and version bump.

moxarth-rathod:
## Proposed commit message

```
qualys_vmdr: record missing KB entries in knowledge_base.status instead of error.message

This change updates the asset_host_detection ingest pipeline so that missing Knowledge Base entries for a QID are recorded in qualys_vmdr.asset_host_detection.knowledge_base.status instead of error.message. With this change, a missing KB entry no longer sets event.kind: pipeline_error or the preserve_original_event tag, while still being recorded for troubleshooting purposes.
```

## Checklist

## Related issues

- Closes #19961


kcreddy:
@moxarth-rathod can you address the CI failure due to PR number mismatch?

ShourieG (approved):
LGTM

PR #20068, opened 2026-07-09
o365: fix document_parsing_exception on Sender field

chemamartinez:
## Proposed commit message

```
o365: fix document_parsing_exception on Sender field

o365.audit.Sender is polymorphic across workloads: Exchange records
report it as a plain email address, while Microsoft Teams
impersonation-detection records report it as a structured identity
object ({UPN, MRI, DisplayName, OrganizationId, UserObjectId}).

The 3.10.3 fix addressed the Teams shape by mapping Sender as
`type: object` with `subobjects: false`. That resolved ingestion for
Teams events, but Elasticsearch only supports a single static type
per field path across a data stream. Since the Exchange string shape
is the dominant case across almost every other workload, locking the
field to `object` caused it to reject the plain-string form instead,
reproducing the same class of document_parsing_exception in the
opposite direction.

This change reverts o365.audit.Sender to `keyword` and adds an ingest
pipeline processor that renames Sender to a new SenderEntity field
whenever it arrives as a Map, before the document reaches the
mapping. This mirrors the existing pattern already used for
Parameters, ModifiedProperties, and ExtendedProperties, where a field
that can arrive as either a scalar or an object is normalized to one
consistent shape ahead of indexing rather than relying on the mapping
to absorb both shapes. Sender stays a keyword for every event that
reports it as a string, and the structured Teams identity object is
preserved in full under SenderEntity.
```

## Checklist

### DIFF
--- a/packages/o365/changelog.yml
+++ b/packages/o365/changelog.yml
@@ -5,7 +5,7 @@
         Fix `document_parsing_exception` on `o365.audit.Sender` by reverting it to `keyword`
         and moving structured Sender objects to the new `o365.audit.SenderEntity` field.
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/99999
+      link: https://github.com/elastic/integrations/pull/20068
 - version: "3.10.3"
   changes:
     - description: Fix `script_item_attachments` processor failing with a Painless regex character-limit error on long attachment filenames that contain parentheses. The regex is replaced with string operations that have no backtracking cost.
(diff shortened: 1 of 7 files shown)


efd6 (commented):
This will need this patch to allow the upgrade test to pass.

[0001-o365-make-upgrade-script-test-robust-to-index-roll-o.patch](https://github.com/user-attachments/files/29867694/0001-o365-make-upgrade-script-test-robust-to-index-roll-o.patch)

efd6 (approved):

PR #20083, opened 2026-07-10
[Cisco ftd] Change type for filesize from integer to long

robester0403:
## Proposed commit message

Convert FileSize to long instead of integer so large file sizes above the 32-bit integer limit no longer fail

## Checklist

## Related issues

Resolves: https://github.com/elastic/sdh-beats/issues/7358

## Screenshots

image


packages/cisco_ftd/data_stream/log/fields/fields.yml
```diff
@@ -231,7 +231,7 @@
         - name: file_sha256
           type: keyword
         - name: file_size
-          type: integer
+          type: long
```
Niceplace, line 234: Have we considered using `scaled_float` ? apparently it's more efficient: https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/number#_which_type_should_i_use

I could not find proof that changing a type from integer to long is a breaking change, what I would consider is how this will impact querying legacy data though, if that is needed at all.

Example: If we query file size with float values as arguments, is it possible that we miss some entries that were originally stored as integers ?

Niceplace: Seems like there's precedent and it was not considered a breaking change: https://github.com/elastic/integrations/pull/15673#issuecomment-3416210312

The TL;DR is that `cross-family` field change would be a breaking change, but not if we stay within the same family.

The [integration docs](https://github.com/elastic/integrations/blob/main/docs/extend/_publish_an_integration.md) also confirm that in a way by not explicitely listing same-family changes as breaking.

Niceplace (approved):
One clarified concern re: same family field change, nothing to do here.

My question about using `scaled_float` still remains but I'll leave the decision of implementing it up to you, otherwise LGTM !

packages/cisco_ftd/data_stream/log/fields/fields.yml
robester0403: Yes it should be change compatible from integer to long, thanks for the double check!

PR #20098, opened 2026-07-13
[gitlab] Normalize custom_message field to handle both string and object values

### LINKED ISSUE #19819: [Gitlab]: Normalize `gitlab.audit.details.custom_message` when GitLab emits string or object values to prevent Elasticsearch mapping conflicts
### Integration Name

GitLab [packages/gitlab]

### Dataset Name

`gitlab.audit` (`logs-gitlab.audit-default`)

### Integration Version

2.6.0

### Agent Version

Not applicable: ingestion is via **Logstash** (`elastic_integration` filter with pipeline `logs-gitlab.audit-2.6.0`), not Elastic Agent/Fleet.  

N/A. GitLab source version: **18.11.0-ee**.

### User Goal

Ingest GitLab audit logs into Elasticsearch via the GitLab integration (Logstash `elastic_integration` filter → `logs-gitlab.audit-default`) without `document_parsing_exception` errors or data loss when `details.custom_message` arrives in different shapes depending on event type.

Desired native behavior (aligned with the confirmed Logstash workaround):
- Leave `details.custom_message` unchanged when it is already a **string**
- When it is an **object** (e.g. `repository_git_operation` with `protocol` and `action`):
  - Copy `protocol` → `network.protocol`
  - Copy `action` → `event.action`
  - Replace `details.custom_message` with a single string (e.g. `git-upload-pack via http`)

### Existing Features

The official GitLab integration field set does not define `gitlab.audit.details.custom_message`. Because the field is absent from the integration mapping, Elasticsearch maps it dynamically on first write. GitLab emits inconsistent types for the same field path:

- **String** for authentication and similar events (e.g. LDAP login, job-token auth)
- **JSON object** for repository git operations (e.g. `{"protocol":"http","action":"git-upload-pack"}`)

Elasticsearch allows only one mapping type per field path. Whichever shape is indexed first wins; documents with the opposite type are rejected. Failure volume depends on event mix, not a fixed count (e.g. half string / half object can cause roughly half of documents to fail).

Users should not need a custom Logstash Ruby filter after `elastic_integration` to normalize this field before indexing.

### What did you see?

**Logstash error:**
```
[2026-05-06T10:31:58,714][WARN ][logstash.outputs.elasticsearch] Could not index event to Elasticsearch.
{:status=>400, :response=>{"index"=>{"status"=>400, "error"=>{"type"=>"document_parsing_exception",
"reason"=>"[1:453] object mapping for [gitlab.audit.details.custom_message] tried to parse field [custom_message] as object, but found a concrete value"}}}}
```

**String `custom_message` (authentication-style event):**
```json
{
  "details": {
    "event_name": "authenticated_with_ldap",
    "custom_message": "Signed in with LDAPMAIN authentication",
    ...
  },
  "event_type": "authenticated_with_ldap"
}
```

**Object `custom_message` (`repository_git_operation`):**
```json
{
  "details": {
    "event_name": "repository_git_operation",
    "custom_message": {
      "protocol": "http",
      "action": "git-upload-pack"
    },
    ...
  },
  "event_type": "repository_git_operation"
}
```

**Mapping:** `gitlab.audit.details.custom_message` is mapped as a string in one backing index and as an object in another, confirming the dynamic mapping conflict.

**Data flow:** GitLab → Logstash (HTTP/UDP input) → `elastic_integration` filter (`logs-gitlab.audit-2.6.0`) → Elasticsearch data stream `logs-gitlab.audit-default`. Managed ingest pipeline `logs-gitlab.audit-2.6.0` had zero document count because processing happens in Logstash.

### Anything else?

- **Confirmed workaround (Logstash Ruby filter after `elastic_integration`):** Normalizes object `custom_message` to a string and promotes `protocol` / `action` to ECS-aligned fields. User confirmed it works after also applying `event.set("[gitlab][audit][details]", details)` to persist mutated `details` back on the event.
- **KCS article:** [support.elastic.co/knowledge/c9e2baaa](https://support.elastic.co/knowledge/c9e2baaa)
- **Suggested integration implementation:** Add explicit field handling (and/or ingest pipeline logic) in the GitLab integration so `gitlab.audit.details.custom_message` is always indexed as a single scalar type (keyword/text), with structured git-operation sub-fields mapped to dedicated fields such as `network.protocol` and `event.action`, matching the Ruby workaround behavior.

brijesh-elastic:
## Proposed commit message

```
gitlab: normalize custom_message field to handle both string and object values
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/gitlab directory.

> elastic-package test -v

## Related issues

- Closes #19819


packages/gitlab/data_stream/audit/fields/fields.yml
```diff
@@ -43,3 +43,15 @@
...
+          type: group
+          fields:
+            - name: custom_message_object
+              type: group
+              fields:
+                - name: protocol
+                  type: keyword
+                - name: action
+                  type: keyword
+            - name: custom_message_string
```
efd6, line 56: Do we know that its fields are limited to this set? The issue says 'JSON object for repository git operations (**e.g.** `{"protocol":"http","action":"git-upload-pack"}`)'.

efd6 (commented):
I think we either need to fully enumerate the possible fields, or have a different mapping for the object case.

packages/gitlab/data_stream/audit/fields/fields.yml
brijesh-elastic: We don't know the full list of fields since the [documentation](https://docs.gitlab.com/user/compliance/audit_event_schema/#audit-event-schema) doesn't mention them. However, in the example section, there are two more fields. Should I map those as well?

efd6: The alternative is to used flattened or an object.

brijesh-elastic: If we use the `flattened` type, we won't be able to create visualizations on the subfields, and all of these fields look important.
On the other hand, if we map it as an `object`, we would have to explicitly define the types for all sub-objects (using `object_type` key). Since we don't know their data types in advance, we can't definitively decide if they should be `keyword`, `long`, etc.

Therefore, I’ve left it as-is for now. Elasticsearch will use dynamic mapping to handle incoming subfields after the JSON is parsed—automatically mapping `strings` as `keyword`, `integers` as `long`, and so on. According to https://www.elastic.co/docs/manage-data/data-store/mapping/dynamic-field-mapping

efd6: OK

efd6 (approved):

PR #20130, opened 2026-07-15
Supabase OpenTelemetry Integration Package

ishleenk17:
Supabase OpenTelemetry Integration (Composable package)

This package depends upon the prometheus_input_otel and the supabase_otel content package. 

Related Issue: https://github.com/elastic/obs-integration-team/issues/1153


packages/supabase/manifest.yml
```diff
@@ -0,0 +1,111 @@
...
+  - src: /img/supabase-logo.svg
+    title: Supabase logo
+    size: 256x256
+    type: image/svg+xml
+requires:
+  input:
+    - package: prometheus_input_otel
+      version: "0.1.1"
+  content:
+    - package: supabase_otel
```
teresaromero, line 32: when installing this with `elastic-package stack up` i get an error, this package is not available yet? also, i have not seen it locally.

teresaromero (commented):
I've tried to build the package and the bundle looks good. I've installed into a 9.4.4 stack and i see variables and options as expected. 

However, i encourage to have a good set of tests for the data streams so there is a clear expectation and good review that all is set as expected. Could you add policy tests and some default system tests with a basic mock service? If i recall correctly supabase can be set with a docker environment, so we should be able to check this out?

packages/supabase/manifest.yml
ishleenk17: YEs this package will be merged soon.
Also say if we update the version on content package with some change. I am hoping this package automatically undestands to always pick the latest pacakge. Or is it expected to change the version of the content packa here too.
Version should be a minimum version criteria imo. Else it should always pick the latest 
@teresaromero

ishleenk17:
@mrodm @teresaromero 
Since we are going with the workaround of enabling the TSDB index manually for now. 
I a unable to write proper tests. As I want to keep TSDB but the current dimention field (that I had to add to validate) is not semthing that is present in all the documents. 
And I don't want to put another field attributes.* as dimension field as if I do that would overwirete tha. mappings of the other child fields of attributes.*

Do you have any suggestion on how to go about it ?
Or else for now, I would skip the test part. Till the time we have the datatreams using TSDB by default.

mykola-elastic (approved):
Configuration was pretty straightforward to me and it worked. LGTM

teresaromero (approved):
This package depends on https://github.com/elastic/integrations/pull/20265 being on EPR; on install time. I guess it can be released but for prod use the content pack should be available. 

This is not surfacing here because there is no system test that requires package installation, just build. Composable packages require only inputs, content dependencies are downloaded by fleet.

PR #9701, opened 2024-04-24
[Azure] [Audit Logs] Update Azure Audit Logs pipeline with support for more fields

lucian-ioan:
## Proposed commit message

Update Azure Audit Logs pipeline with support for more fields.

## Checklist

- [ ] I have reviewed [tips for building integrations](https://github.com/elastic/integrations/blob/main/docs/tips_for_building_integrations.md) and this pull request is aligned with them.
- [ ] I have verified that all data streams collect metrics or logs.
- [ ] I have added an entry to my package's `changelog.yml` file.
- [ ] I have verified that Kibana version constraints are current according to [guidelines](https://github.com/elastic/elastic-package/blob/master/docs/howto/stack_version_support.md#when-to-update-the-condition).

## Author's Checklist

- [ ] 

## Related issues

- Closes [beats #37249](https://github.com/elastic/beats/issues/37249)

PR #20132, opened 2026-07-15
[Azure] Fix degraded auditlogs: raise keyword limits, fix missing field mappings, add pipeline tests

shmsr:
## Proposed commit message

```
azure: fix degraded documents in auditlogs by raising keyword limits

Azure Entra ID audit events can contain multi-kilobyte string values in several
`azure.auditlogs` fields, especially serialized object-state diffs under
`modified_properties.*.old_value` and `modified_properties.*.new_value`.

In v1.37.0 these values can exceed the effective `ignore_above: 1024` limit from
Fleet fallback/package-generated keyword mappings. When that happens,
Elasticsearch indexes the document but ignores the oversized keyword field, adds
the field name to `_ignored`, and Kibana surfaces the event as a degraded
document.
```

This change reduces those degraded documents by:

- raising `ignore_above` to `8191` for:
  - `azure.auditlogs.result_description`
  - `azure.auditlogs.properties.result_reason`
  - `azure.auditlogs.properties.result_description` _(newly declared — see below)_
  - `azure.auditlogs.properties.additional_details.value`
  - `azure.auditlogs.properties.target_resources.*.modified_properties.*.*`
- normalizing top-level `azure.auditlogs.resultDescription` to
  `azure.auditlogs.result_description` in the ingest pipeline
- replacing the per-leaf `modified_properties.*.{new_value,display_name,old_value}`
  declarations with a single `object_type: keyword` mapping on
  `modified_properties.*.*`
- fixing a pre-existing gap from #9701: `properties.resultDescription →
  properties.result_description` was renamed in the pipeline but the target field
  was never declared in `fields.yml`, leaving it unmapped since 2023
- adding pipeline test fixtures for both `resultDescription` renames

The `modified_properties` path is a doubly-nested wildcard path
(`target_resources.* -> modified_properties.*`). The previous per-leaf mapping
appears to hit the same Fleet/EPM shared-section mapping-generation issue
described in elastic/kibana#270746, where only the first declared leaf receives
the intended package dynamic template and later leaves can fall through to
Fleet's fallback `strings_as_keyword` template.

`ignore_above: 8191` is used instead of `32766` because `ignore_above` is
measured in characters while Lucene's hard term limit is 32,766 bytes. UTF-8
characters can use up to 4 bytes, so `32766 / 4 = 8191` is the largest safe
character limit that avoids hard document rejection for all UTF-8 input. This is
also the logsdb default for keyword fields in Elasticsearch 9.x.

This is not a field-type breaking change. The affected fields remain `keyword`,
so field capabilities remain compatible across old and new backing indices.
Existing documents that were already degraded are not retroactively repaired.
Existing backing indices also keep their current mappings unless their mappings
are explicitly updated; the package template change applies to new backing
indices created after upgrade.

## Checklist

- [x] I have verified that any added dashboard complies with Kibana's [Dashboard good practices](https://docs.elastic.dev/ux-guidelines/data-viz/dashboard-good-practices). No dashboards are added by this PR.

## Author's Checklist

- [x] Raised the affected auditlogs keyword fields to `ignore_above: 8191`.
- [x] Added ingest pipeline normalization for top-level `resultDescription`.
- [x] Reworked `modified_properties.*` leaf mappings into one `modified_properties.*.*` keyword object mapping.
- [x] Verified the fix on a 9.x logsdb data stream.
- [x] Added pipeline fixture coverage for top-level `resultDescription` rename.
- [x] Added pipeline fixture coverage for `properties.resultDescription` rename; declared missing `properties.result_description` field in `fields.yml`.


shmsr:
@elastic/obs-ds-hosted-services and @elastic/security-service-integrations could you PTAL?

ishleenk17 (approved):

zmoog (approved):

packages/azure/changelog.yml
```diff
@@ -1,3 +1,8 @@
+- version: "1.37.1"
+  changes:
+    - description: Reduce degraded documents in azure.auditlogs by raising the keyword ignore_above limit to 8191 on result_description, result_reason, additional_details.value, and modified_properties leaf fields; fix the ingest pipeline to normalize the top-level resultDescription field to snake_case so it is correctly mapped; declare the missing properties.result_description field mapping (gap since #9701). Mapping changes apply to new backing indices after upgrade or rollover.
```
shmsr:
```suggestion
    - description: Reduce degraded documents in azure.auditlogs by raising the keyword `ignore_above` limit to `8191` on `result_description`, `result_reason`, `additional_details.value`, and `modified_properties` leaf fields; fix the ingest pipeline to normalize the top-level `resultDescription` field to snake_case so it is correctly mapped; declare the missing `properties.result_description` field mapping as well.
```

Issue #18996, opened 2026-05-13
[AbuseCH] Add support for new Commercial API


jamiehynds:
## Summary

Our existing `ti_abusech` integration uses the free, community abuse.ch APIs (MalwareBazaar, URLhaus, ThreatFox). The abuse.ch–Spamhaus Alliance now offers a commercial API delivering the same datasets on enterprise infrastructure, with token-based authentication and commercial support. This issue tracks adding support for the commercial API as a supported API within our existing integration. 

## Background
The [abuse.ch commercial API](https://abusech.docs.spamhaus.com/) is provided via the Spamhaus platform and is positioned as the commercial-grade upgrade from the community APIs. It exposes the same core datasets but with:

- **Bearer token authentication** (replacing the unauthenticated or API-key community endpoints)
- Enterprise-grade reliability and scale for always-on intelligence ingestion
- An OpenAPI specification
- Commercial and technical support SLAs

The four datasets covered by the commercial API are the same as the current integration:

| Dataset | Description |
|---|---|
| MalwareBazaar | Malware sample hashes, metadata, family attribution, delivery method |
| URLhaus | Malicious URLs used for malware distribution |
| ThreatFox | IOCs (IPs, domains, URLs, hashes) with malware context |
| YARAify | YARA rule matches and scan results |

## Scope

- ✅ Polling the commercial REST API endpoints for the four datasets listed above
- ✅ Bearer token authentication (token obtained via POST to login endpoint, refreshed on expiry)
- ✅ Configurable API base URL to support the Spamhaus commercial endpoint

## Proposed Changes

1. **New authentication option** in the integration config — add a `Commercial API` mode alongside the existing community API mode, accepting a Bearer token credential
2. **Configurable base URL** — allow users to point the integration at the Spamhaus commercial endpoint rather than the community endpoints
3. **Token refresh handling** — the commercial API issues short-lived tokens; the integration should handle token expiry and re-authentication gracefully
4. **Documentation update** — add a setup section in the integration README explaining:
   - How to obtain a commercial API token via the [Spamhaus Customer Portal](https://portal.spamhaus.com)
   - Where to enter the Bearer token in the integration UI

## Notes
The commercial API docs are behind a customer portal login. Spamhaus have kindly agreed to provide access for the duration of integration development. 

## References
- Commercial API overview: https://www.spamhaus.com/data-access/abusech-api/
- API docs (requires portal access): https://abusech.docs.spamhaus.com/api-reference
[...]

PR #20134, opened 2026-07-15
[AbuseCH] Add Commercial API support for the ThreatFox data stream

moxarth-rathod:
## Proposed commit message

```
ti_abusech: add commercial API support for the threatfox data stream
```

## Checklist

- Clone integrations repo.
- Move to integrations/packages/ti_abusech directory.

> elastic-package test

## Related issues

- Relates #18996


packages/ti_abusech/data_stream/threatfox/_dev/test/pipeline/test-commercial-fields.log
```diff
@@ -0,0 +1,2 @@
+{"id":"841537","ioc":"wizzy.hopto.org","threat_type":"botnet_cc","threat_type_desc":"Indicator that identifies a botnet command&control server (C&C)","ioc_type":"domain","ioc_type_desc":"Domain that is used for botnet Command&control (C&C)","malware":"win.asyncrat","malware_printable":"AsyncRAT","malware_alias":null,"malware_malpedia":"https://malpedia.caad.fkie.fraunhofer.de/details/win.asyncrat","confidence_level":100,"first_seen":"2022-08-05T19:43:08Z","last_seen":null,"reference":"https://tria.ge/220805-w57pxsgae2","reporter":"test_reporter","tags":["asyncrat"],"sightings":3,"comment":"These domains are too bad!","is_compromised":false,"credits":[{"credits_from":"ThreatFox","credits_amount":5}],"malware_samples":[{"timestamp":"2022-01-10T19:17:22Z","malware_bazaar_url":"https://bazaar.abuse.ch/sample/b325c92fa540edeb89b95dbfd4400c1cb33599c66859a87aead820e568a2ebe7/","md5_hash":"5b7e82e051ade4b14d163eea2a17bf8b","sha256_hash":"b325c92fa540edeb89b95dbfd4400c1cb33599c66859a87aead820e568a2ebe7"}]}
```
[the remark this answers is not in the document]
efd6: Check this.

packages/ti_abusech/data_stream/threatfox/_dev/test/pipeline/test-commercial-fields.log-config.yml
```diff
@@ -0,0 +1,8 @@
+fields:
+  tags:
+    - preserve_original_event
+  _conf:
+    ioc_expiration_duration: 5d
+dynamic_fields:
+  event.ingested: ".*"
```
efd6:
```suggestion
```

packages/ti_abusech/data_stream/threatfox/_dev/test/pipeline/test-commercial-fields.log-config.yml
```diff
@@ -0,0 +1,8 @@
...
+  abusech.threatfox.deleted_at: ".*"
```
efd6: Is this needed?

packages/ti_abusech/data_stream/threatfox/_dev/test/policy/test-all.expected
```diff
@@ -28,52 +27,170 @@ inputs:
                 target: environment
           program: |
             state.with(
-              request("POST", state.url, '{ "query": "get_iocs", "days": ' + state.?cursor.days.orValue(state.initial_interval) + ' }'
-              ).with({
-                  "Header":{
+              state.?api_type.orValue("community") == "commercial" ?
+                (
+                  has(state.token) && has(state.token.access_token) && has(state.token.expires) && now() < timestamp(int(state.token.expires) - 60) ?
```
efd6:
```suggestion
                  has(state.?token.access_token) && has(state.?token.expires) && now() < timestamp(int(state.token.expires) - 60) ?
```
In CEL, boolean expression are not short-circuiting, so there is no point checking the parent field's existence.

packages/ti_abusech/data_stream/threatfox/_dev/test/policy/test-all.expected
```diff
@@ -28,52 +27,170 @@ inputs:
...
+                                string(auth_resp.Status) + " (" + string(auth_resp.StatusCode) + ")"
+                            ),
+                          },
+                        },
+                        "want_more": false,
+                      }
+                    )
+                ).as(token, !has(token.access_token) ? token :
+                  (
+                    has(state.cursor) && has(state.cursor.after) ?
```
efd6:
```suggestion
                    has(state.?cursor.after) ?
```

packages/ti_abusech/data_stream/threatfox/_dev/test/policy/test-all.expected
```diff
@@ -28,52 +27,170 @@ inputs:
...
+                    ).with({
+                      "Header": {
+                        "Authorization": ["Bearer " + token.access_token],
+                      },
+                    }).do_request().as(resp, resp.StatusCode == 200 ?
+                      resp.Body.decode_json().as(body, {
+                        "events": body.items.map(ind, {
+                          "message": ind.encode_json(),
+                        }),
+                        "want_more": has(body.meta) && has(body.meta.next_cursor),
```
efd6:
```suggestion
                        "want_more": has(body.?meta.next_cursor),
```

packages/ti_abusech/data_stream/threatfox/_dev/test/policy/test-all.expected
```diff
@@ -28,52 +27,170 @@ inputs:
...
+                        "cursor": (
+                          body.items.size() > 0 ?
+                            optional.of(
+                              (
+                                has(state.cursor) && has(state.cursor.first_seen) ?
```
efd6:
```suggestion
                                has(state.?cursor.first_seen) ?
```

packages/ti_abusech/data_stream/threatfox/_dev/test/policy/test-all.expected
```diff
@@ -28,52 +27,170 @@ inputs:
...
+                                  max(timestamp(state.cursor.first_seen), body.items.map(ind, timestamp(ind.first_seen)).max())
+                                :
+                                  body.items.map(ind, timestamp(ind.first_seen)).max()
+                              ).format(time_layout.RFC3339)
+                            )
+                          :
+                            state.?cursor.first_seen
+                        ).as(first_seen,
+                          has(body.meta) && has(body.meta.next_cursor) ?
```
efd6:
```suggestion
                          has(body.?meta.next_cursor) ?
```

packages/ti_abusech/data_stream/threatfox/_dev/test/policy/test-all.expected
```diff
@@ -28,52 +27,170 @@ inputs:
...
+                          "access_token": token.access_token,
+                          "expires": token.expires,
+                        },
+                      })
+                    :
+                      {
+                        "events": {
+                          "error": {
+                            "code": string(resp.StatusCode),
+                            "id": string(resp.Status),
```
efd6:
```suggestion
                            "id": resp.Status,
```

packages/ti_abusech/data_stream/threatfox/_dev/test/policy/test-all.expected
```diff
@@ -28,52 +27,170 @@ inputs:
...
+                            "message": "GET " + state.url.trim_right("/") + "/threatfox/v1/iocs: " + (
+                              size(resp.Body) != 0 ?
+                                string(resp.Body)
+                              :
+                                string(resp.Status) + " (" + string(resp.StatusCode) + ")"
```
efd6:
```suggestion
                                resp.Status + " (" + string(resp.StatusCode) + ")"
```

packages/ti_abusech/data_stream/threatfox/_dev/test/policy/test-all.expected
```diff
@@ -28,52 +27,170 @@ inputs:
...
-                      ),
+                }).as(req, req.do_request().as(resp, resp.StatusCode == 200 ?
+                  resp.Body.decode_json().as(body, {
+                    "events": (has(body.data) ?
+                      body.data.map(ind, {
+                        "message": ind.encode_json(),
+                      })
+                    :
+                      []
+                    ),
```
efd6:
```suggestion
                    "events": has(body.data) ?
                      body.data.map(ind, {"message": ind.encode_json()})
                    :
                      [],
```

packages/ti_abusech/data_stream/threatfox/_dev/test/policy/test-all.expected
```diff
@@ -28,52 +27,170 @@ inputs:
...
-                  "want_more": false,
-                }
-              ))
+                  })
+                :
+                  {
+                    "events": {
+                      "error": {
+                        "code": string(resp.StatusCode),
+                        "id": string(resp.Status),
```
efd6:
```suggestion
                        "id": resp.Status,
```

packages/ti_abusech/data_stream/threatfox/_dev/test/policy/test-all.expected
```diff
@@ -28,52 +27,170 @@ inputs:
...
+                        "message": "POST " + state.url.trim_right("/") + ":" + (
+                          size(resp.Body) != 0 ?
+                            string(resp.Body)
+                          :
+                            string(resp.Status) + ' (' + string(resp.StatusCode) + ')'
```
efd6:
```suggestion
                            resp.Status + ' (' + string(resp.StatusCode) + ')'
```

packages/ti_abusech/data_stream/threatfox/_dev/test/policy/test-all.expected
efd6, line 1: I should have written these on the the agent template file.

packages/ti_abusech/data_stream/threatfox/manifest.yml
```diff
@@ -8,11 +8,40 @@ streams:
...
+          URL for the ThreatFox API. For the Community API, use the full endpoint URL. For the Commercial API, use the API base URL (for example `https://api.spamhaus.com`).
         multi: false
         required: true
-        show_user: false
+        show_user: true
         default: https://threatfox-api.abuse.ch/api/v1/
+      - name: api_type
+        type: select
+        title: API Type
+        description: Select the abuse.ch API to use. Community API uses an Auth Key. Commercial API uses Spamhaus credentials and JWT authentication.
```
kcreddy: Add to README how to fetch credentials from commercial API.

packages/ti_abusech/data_stream/threatfox/manifest.yml
```diff
@@ -8,11 +8,40 @@ streams:
...
+        description: >-
+          URL for the ThreatFox API. For the Community API, use the full endpoint URL. For the Commercial API, use the API base URL (for example `https://api.spamhaus.com`).
         multi: false
         required: true
-        show_user: false
+        show_user: true
         default: https://threatfox-api.abuse.ch/api/v1/
+      - name: api_type
+        type: select
+        title: API Type
```
kcreddy: Are we trying to add the same options (`api_type`, `username`, `password`) for each new commercial data stream https://github.com/elastic/integrations/issues/18996? 

Can the same username / password be used for all new data streams, so users can configure it once, instead of 4-5 times?

kcreddy (commented):
Can you update/add benchmarks?

packages/ti_abusech/data_stream/threatfox/agent/stream/cel.yml.hbs
```diff
@@ -15,56 +15,191 @@ resource.timeout: {{http_client_timeout}}
...
+            },
+            "want_more": false,
+          }
+        :
+          (
+            has(state.?token.access_token) && has(state.?token.expires) && now() < timestamp(int(state.token.expires) - 60) ?
+              state.token
+            :
+              request(
+            "POST",
```
efd6: The indentation is wrong here.

packages/ti_abusech/data_stream/threatfox/manifest.yml
moxarth-rathod: Yes, done

packages/ti_abusech/data_stream/threatfox/agent/stream/cel.yml.hbs
moxarth-rathod: fixed

packages/ti_abusech/data_stream/threatfox/manifest.yml
```diff
@@ -8,10 +8,11 @@ streams:
       - name: url
         type: text
         title: URL
-        description: Base URL of the abuse.ch ThreatFox API to collect threat indicators.
+        description: >-
+          URL for the ThreatFox API. For the Community API, use the full endpoint URL. For the Commercial API, use the API base URL (for example `https://api.spamhaus.com`).
```
kcreddy: Let's keep the commercial vars shared at the `cel` input level rather than following the [bot's suggestion](https://github.com/elastic/integrations/pull/20134/changes#r3636519083) to push them back to the data stream. The bot is optimising for today's single commercial stream, but per #18996 we're adding MalwareBazaar, URLhaus, and YARAify next and it'll keep growing — if these live at the data-stream level, users re-enter the same Spamhaus username/password on every stream (4+ times today, more later). Since all six streams share the one `cel` input, input-level vars are the "configure once" home (that's already where `auth_key` lives).

Concretely, at the shared `cel` input in `packages/ti_abusech/manifest.yml`:

- `api_type` — `select` (community / commercial), `required: true`, `show_user: true`
- `auth_key` — `required: false`, `secret: true` (community)
- `username` — `required: false` (commercial)
- `password` — `required: false`, `secret: true` (commercial)

On the bot's point that `auth_key` lost `required: true`: we can get conditional validation back without making it unconditionally required. `required_vars` (available at our `format_version: 3.3.2`, at the input level) does exactly this — same pattern `elastic_security` uses. Add it to the `cel` input:

```yaml
required_vars:
  community:
    - name: api_type
      value: community
    - name: auth_key
  commercial:
    - name: api_type
      value: commercial
    - name: username
    - name: password
```

Now Fleet requires `auth_key` only when `api_type == community`, and `username`/`password` only when `commercial` — no unconditional `required`, no opaque runtime CEL failure. (Note: the vars referenced here must be `required: false` in their own definitions, or package-spec validation errors.)

On the URL: The community endpoints are already per-stream `show_user: false` and differ by host/path, so they must stay per-stream and hidden. For commercial, add one shared base URL at the input level, also `show_user: false` (adding a second _user-visible_ URL field would confuse users), defaulting to `https://api.spamhaus.com`, and have the template append the fixed per-dataset path, routing on `api_type`:

```handlebars
{{#if (eq api_type "commercial")}}
resource.url: {{commercial_url}}/api/v1/threatfox
{{else}}
resource.url: {{url}}
{{/if}}
```

That way users see no URL field in normal config for either mode — just `API Type` and the relevant credential. 

Update/add the tests accordingly.

ShourieG:
@vera-review-bot review

kcreddy (commented):
## Var scope decision for commercial API credentials — needs a call before more datasets land

cc @jamiehynds @moxarth-rathod @efd6

### Where we are

4.2.0 ([#20199](https://github.com/elastic/integrations/pull/20199), merged 2026-07-28, now published on EPR) shipped MalwareBazaar commercial support with `api_type`, `username` and `password` declared as **data-stream-level** vars on the `malwarebazaar` stream.

In the ThreatFox PR ([#20134](https://github.com/elastic/integrations/pull/20134)) I asked for those vars to move to the shared `cel` **input level**, so that users configure one Spamhaus credential pair for the whole integration instead of repeating it per dataset. With four commercial datasets in scope here, per-stream vars mean the same username/password entered 4+ times, and that number only grows. The PR has made that move and added conditional validation via input-level `required_vars`, which reads well.

The catch I missed when I asked for it: 4.2.0 is already released, so moving those vars is a **breaking change** for anyone who configured MalwareBazaar commercial on 4.2.0. Fleet does not carry var values across a scope change on its own.

### Why we can't just migrate the values automatically

Package-spec has exactly the right feature — variable-level `migrate_from` with `scope: stream` — but we cannot use it today:

- It is gated `before: 3.7.0` in the spec, so it requires `format_version: "3.7.0"`. The latest released spec is `v3.6.6` and the changelog head is `3.7.0-next`; no package in the repo uses 3.7.0 yet. (Unrelated bug for the spec owners: the changelog lists this feature under 3.6.3, which contradicts the gate.)
- Fleet's implementation ([elastic/kibana#268789](https://github.com/elastic/kibana/pull/268789)) is labelled `v9.5.0`. So even after the spec releases, using it would force `conditions.kibana.version: ^9.5.0`, well above this package's current `^9.1.0`.

### Options

1. **Move the vars now and own the break.** Ship as `5.0.0` with a `breaking-change` changelog entry and a Troubleshooting upgrade note telling 4.2.0 commercial MalwareBazaar users to re-enter API Type, Username and Password at the integration level. Precedent: this package's 4.0.0 entry did exactly this for the URLhaus endpoint change.
2. **Keep credentials per-stream.** No break today, but users configure the same Spamhaus credentials once per commercial dataset, and we carry duplicated var definitions across all four. Migrating later is strictly more expensive.
3. **Deprecate rather than remove for one cycle.** Keep the three vars on the `malwarebazaar` stream marked `deprecated: true`, have the MalwareBazaar template prefer the stream value when present, and remove them in the following major. Avoids the break, but means two sources of truth in the UI for a release, needs `format_version >= 3.6.0`, and needs Fleet-side confirmation that deprecated vars behave as we expect.
4. **Wait for spec 3.7.0 + Kibana 9.5.** Cleanest migration, but blocks the remaining commercial datasets and raises the stack floor to 9.5.

My recommendation is option 1. The blast radius is people who picked up a six-day-old feature, and it shrinks the longer we wait only in the sense that it grows the longer we defer the move.

### Next step

Whatever we choose, I'd like the upgrade path verified with an actual 4.2.0 to next-version upgrade test rather than reasoning about it, and the outcome recorded here so the remaining commercial datasets (URLhaus, ThreatFox, YARAify) follow one pattern.

efd6:
I agree that the break is likely the least worst.

packages/ti_abusech/data_stream/threatfox/_dev/test/system/test-auth-config.yml
```diff
@@ -5,6 +5,7 @@ data_stream:
     url: http://{{Hostname}}:{{Port}}/api/v1/
     preserve_original_event: true
     ioc_expiration_duration: 5d
+    api_type: community
```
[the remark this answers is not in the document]
kcreddy: @moxarth-rathod, please update this

packages/ti_abusech/_dev/build/docs/README.md
```diff
@@ -71,7 +85,8 @@ Elastic Agent must be installed. For more details, check the Elastic Agent [inst
 
     * To **Collect abuse.ch logs via API**, you'll need to:
 
-        - Configure **Auth Key**.
+        - Configure **Auth Key** for Community API datasets.
+        - For Commercial API collection, set **API Type** to **Commercial API**, set the **URL** to the commercial API base URL (for example `https://api.spamhaus.com`), and configure **Username** and **Password**.
```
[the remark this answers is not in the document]
kcreddy: @moxarth-rathod, update the instructions accordingly.

packages/ti_abusech/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "4.3.0"
+  changes:
+    - description: Add Commercial API support for the ThreatFox data stream.
```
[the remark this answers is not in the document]
kcreddy: @moxarth-rathod, we need breaking-change for this.

packages/ti_abusech/manifest.yml
```diff
@@ -9,7 +9,7 @@ categories:
   - threat_intel
 conditions:
   kibana:
-    version: "^8.19.0 || ^9.1.0"
+    version: "^9.1.0"
```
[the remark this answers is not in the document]
kcreddy: @moxarth-rathod, do we need to drop `8.19` support entirely to support `required_vars`?

kcreddy (commented):
@moxarth-rathod, can you move the other PRs of AbuseCH to draft until this design is finalised and merged.

Also can you check and comment on Vera bot reviews if required or not?

packages/ti_abusech/manifest.yml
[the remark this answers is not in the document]
moxarth-rathod: The fix is only available in 9.1.0 that's why we can remove 8.19 - https://github.com/elastic/kibana/pull/211863

packages/ti_abusech/changelog.yml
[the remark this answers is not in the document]
moxarth-rathod: Done

moxarth-rathod:
> @moxarth-rathod, can you move the other PRs of AbuseCH to draft until this design is finalised and merged.
> 
> Also can you check and comment on Vera bot reviews if required or not?

@kcreddy done

packages/ti_abusech/data_stream/malwarebazaar/_dev/test/policy/test-commercial.expected
```diff
@@ -0,0 +1,242 @@
...
+          interval: 10m
+          program: |
+            state.with(
+              state.?api_type.orValue("community") == "commercial" ?
+                (
+                  !has(state.username) || !has(state.password) ?
+                    {
+                      "events": {
+                        "error": {
+                          "message": "api_type is 'commercial' but username/password are not configured",
```
kcreddy, line 25: Add `error.id: configuration_error` and `error.code: 400`. See `url` data stream. This is also documented in README to quickly find the class of the error. Similarly in other placs.

packages/ti_abusech/manifest.yml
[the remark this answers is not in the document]
kcreddy: Thanks @moxarth-rathod, I verified this and your reading is right. The earlier [#207130](https://github.com/elastic/kibana/pull/207130) was backported to 8.18 but only covers stream-level.

We also really don't want to drop the support for `8.19` yet, so:

1. **Verify this claim**: Having `required_vars` isn't going to make package unusable. On 8.19 the field is simply ignored — no install error, collection unaffected. So if have to choose "8.19 gets no conditional validation" vs "8.19 gets no future releases of this package at all", including bugfixes for the five community datasets those users do run, I would choose the former.
2. **Add guards:** Because this PR also relaxes `auth_key` to `required: false` and leans entirely on `required_vars`, 8.19 users would end up with *no* credential validation — a regression from 4.2.0 where `auth_key` was unconditionally required. That's a stronger argument for your position than the bot's, but it's also addressable: the commercial branch already guards the missing-credentials case and surfaces it as an error event, and adding the equivalent guard for a missing `auth_key` in these data streams would make 8.19 degrade gracefully.

I'd lean towards keeping `^8.19.0 || ^9.1.0` plus the runtime guards, so 8.19 users keep receiving fixes. If we do decide to drop 8.19, it needs an explicit changelog entry saying so rather than an incidental manifest edit.
Also, I would go with `elastic_security`'s `^9.1.1` rather than `^9.1.0`, because of a likely bugfix that was made in `.1`.

@efd6 any other thoughts here: keep `^8.19.0 || ^9.1.0` and lean on runtime guards, or drop 8.19 so input-level `required_vars` validation actually applies? If we are really dropping 8.19, then we may have better argument to use `var_groups` option for better UI.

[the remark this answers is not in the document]
moxarth-rathod: Thanks. I've verified that `8.19.0` works as expected (minus the conditional validation) and added the config guards so missing credentials degrade gracefully.

We will not be dropping `8.19.0` support, so keeping ^8.19.0 || ^9.1.0 with the runtime guards is the way forward. Everything tested completely fine on 9.1.0 during verification, so I didn't see a need to bump it to 9.1.1.

packages/ti_abusech/elasticsearch/transform/latest_threatfox/fields/fields.yml
```diff
@@ -34,3 +34,53 @@
       type: keyword
       description: |
         The configured expiration duration.
+    - name: sightings
+      type: long
+      description: >
+        The number of sightings of the IOC.
+    - name: comment
+      type: keyword
```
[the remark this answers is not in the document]
kcreddy: @moxarth-rathod, can you fix this mismatch?

packages/ti_abusech/elasticsearch/transform/latest_threatfox/fields/fields.yml
```diff
@@ -34,3 +34,53 @@
       type: keyword
       description: |
         The configured expiration duration.
+    - name: sightings
```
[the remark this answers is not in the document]
kcreddy: @moxarth-rathod, this also needs a version bump so that new mappings are applied.

packages/ti_abusech/data_stream/threatfox/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -223,6 +230,118 @@ processors:
...
+                value: 'Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}'
+  - foreach:
+      tag: foreach_rename_malware_samples_time_stamp_to_timestamp
+      description: Normalize community API time_stamp to commercial timestamp field name.
+      field: abusech.threatfox.malware_samples
+      if: ctx.abusech?.threatfox?.malware_samples instanceof List
+      processor:
+        rename:
+          tag: rename_malware_samples_time_stamp_to_timestamp
+          field: _ingest._value.time_stamp
```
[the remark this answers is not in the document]
kcreddy: @moxarth-rathod, can you confirm this normalisation is deliberately skipped?

packages/ti_abusech/_dev/build/docs/README.md
```diff
@@ -94,6 +109,14 @@ Elastic Agent must be installed. For more details, check the Elastic Agent [inst
 
 ## Troubleshooting
 
+- **Upgrading to v5.0.0**: Version 5.0.0 moves **API Type**, **Username (Commercial)**, and **Password (Commercial)** from the MalwareBazaar data stream to the shared integration input. Fleet does not carry these values across that scope change. If you configured MalwareBazaar Commercial API on 4.2.0, re-enter the credentials after upgrade:
```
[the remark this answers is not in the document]
kcreddy: @moxarth-rathod, this is worth doing.

packages/ti_abusech/data_stream/malwarebazaar/_dev/test/policy/test-commercial.expected
```diff
@@ -0,0 +1,253 @@
...
+          interval: 10m
+          program: |
+            state.with(
+              state.?api_type.orValue("community") == "commercial" ?
+                (
+                  !has(state.username) || !has(state.password) ?
+                    {
+                      "events": {
+                        "error": {
+                          "code": 400,
```
kcreddy:
```suggestion
                          "code": "400",
```

This field is `keyword`

packages/ti_abusech/_dev/build/docs/README.md
```diff
@@ -35,9 +35,12 @@ This integration installs [Elastic latest transforms](https://www.elastic.co/doc
 
 ### From abuse.ch
 
-abuse.ch requires an `Auth Key` (API key) for request authentication. Any requests made without this key will be rejected by the abuse.ch APIs.
+Authentication depends on which API you use:
 
-#### Obtain `Auth Key`
+- **Community API**: requires an `Auth Key` (API key). Any requests made without this key will be rejected by the abuse.ch community APIs.
+- **Commercial API**: requires Spamhaus username and password credentials. The integration uses these to obtain a short-lived JWT for API requests.
+
```
kcreddy:
```suggestion
Which credentials you need depends on which datasets you enable, not on the integration as a whole:

- **ThreatFox threat indicators** (`threatfox`) and **MalwareBazaar payloads** (`malwarebazaar`) can use either API, selected with the shared **API Type** setting. The Community API requires an **Auth Key**. The Commercial API requires Spamhaus username and password credentials, which the integration exchanges for a short-lived JWT.
- **Malware URLs** (`url`) and **Malware payloads** (`malware`) always query the URLhaus Community API and require an **Auth Key** regardless of **API Type**. Requests without the key are rejected.
- **SSL Blacklisted Certificates** (`sslblacklist`) and **JA3 Fingerprints** (`ja3_fingerprints`) read the SSLBL feeds and use neither credential.

Because **API Type** applies to the whole integration, selecting **Commercial API** does not remove the need for an **Auth Key** unless you also disable **Malware URLs** and **Malware payloads**.

```

packages/ti_abusech/_dev/build/docs/README.md
```diff
@@ -71,7 +85,8 @@ Elastic Agent must be installed. For more details, check the Elastic Agent [inst
 
     * To **Collect abuse.ch logs via API**, you'll need to:
 
-        - Configure **Auth Key**.
+        - Configure **Auth Key** for Community API datasets.
+        - For Commercial API collection, set **API Type** to **Commercial API** and configure **Username** and **Password**.
```
kcreddy:
```suggestion
    - Configure **Auth Key (Community)**. The **Malware URLs** and **Malware payloads** datasets always need it, and **ThreatFox threat indicators** and **MalwareBazaar payloads** need it when **API Type** is **Community API**.
    - To use the Commercial API for **ThreatFox threat indicators** and **MalwareBazaar payloads**, set **API Type** to **Commercial API** and configure **Username (Commercial)** and **Password (Commercial)**.
```

kcreddy (commented):
@moxarth-rathod, check bot's comments: https://github.com/elastic/integrations/pull/20134#discussion_r3756916951, https://github.com/elastic/integrations/pull/20134#discussion_r3733757800, https://github.com/elastic/integrations/pull/20134#discussion_r3733757793

Also make following changes as well. These code changes helps users with clear errors in case they configure it wrongly. Also few improvements to README.


**1. Guard `url` when no Auth Key is set**

```diff
--- a/packages/ti_abusech/data_stream/url/agent/stream/cel.yml.hbs
+++ b/packages/ti_abusech/data_stream/url/agent/stream/cel.yml.hbs
@@ -46,5 +46,16 @@
         },
         "want_more": false,
       }
+    : !has(state.auth_key) ?
+      {
+        "events": {
+          "error": {
+            "code": "400",
+            "id": "configuration_error",
+            "message": "auth_key is not configured. The URLhaus API rejects unauthenticated requests. Generate an Auth Key in the abuse.ch authentication portal (https://auth.abuse.ch) and set the Auth Key (Community) setting.",
+          },
+        },
+        "want_more": false,
+      }
     :
     request("GET", state.url).with({
```

**2. Same guard for `malware`**

This stream has no configuration guards at all today, so the request needs wrapping.

```diff
--- a/packages/ti_abusech/data_stream/malware/agent/stream/cel.yml.hbs
+++ b/packages/ti_abusech/data_stream/malware/agent/stream/cel.yml.hbs
@@ -24,4 +24,16 @@
 program: |
   state.with(
+    !has(state.auth_key) ?
+      {
+        "events": {
+          "error": {
+            "code": "400",
+            "id": "configuration_error",
+            "message": "auth_key is not configured. The URLhaus payloads API rejects unauthenticated requests. Generate an Auth Key in the abuse.ch authentication portal (https://auth.abuse.ch) and set the Auth Key (Community) setting.",
+          },
+        },
+        "want_more": false,
+      }
+    :
     request("GET", state.url).with({
       "Header":{
         "Content-Type": ["application/json"],
```

Note that `malware` renders no `state:` block at all when `auth_key` is unset, so the guard has to work against an empty state map. I checked it with mito and it takes the error branch correctly. Once the guard is in, the `?"Auth-Key"` optional handling below it is unreachable-by-construction dead weight in both files, but simplifying that is a larger diff and can wait.

**3. Scope the `api_type` description**

```diff
--- a/packages/ti_abusech/manifest.yml
+++ b/packages/ti_abusech/manifest.yml
@@ -70,7 +70,8 @@
           - name: api_type
             type: select
             title: API Type
-            description: Select the abuse.ch API to use. Community API uses an Auth Key. Commercial API uses Spamhaus credentials and JWT authentication.
+            description: >-
+              Select the abuse.ch API to use for the ThreatFox and MalwareBazaar datasets. Community API uses an Auth Key. Commercial API uses Spamhaus credentials and JWT authentication. The Malware URLs and Malware payloads datasets always use the Community API and require an Auth Key regardless of this setting.
             multi: false
             required: false
```

One correction to the recommendation as written: `sslblacklist` and `ja3_fingerprints` do not send an `Auth-Key` header — both templates are a bare `get(state.url)` — so only `url` and `malware` need the key, and the description shouldn't claim otherwise.

**4. Explain it in the README** (also in the review comments below)

```diff
--- a/packages/ti_abusech/_dev/build/docs/README.md
+++ b/packages/ti_abusech/_dev/build/docs/README.md
@@ -36,8 +36,11 @@
 ### From abuse.ch
 
-Authentication depends on which API you use:
+Which credentials you need depends on which datasets you enable, not on the integration as a whole:
 
-- **Community API**: requires an `Auth Key` (API key). Any requests made without this key will be rejected by the abuse.ch community APIs.
-- **Commercial API**: requires Spamhaus username and password credentials. The integration uses these to obtain a short-lived JWT for API requests.
+- **ThreatFox threat indicators** (`threatfox`) and **MalwareBazaar payloads** (`malwarebazaar`) can use either API, selected with the shared **API Type** setting. The Community API requires an `Auth Key`. The Commercial API requires Spamhaus username and password credentials, which the integration exchanges for a short-lived JWT.
+- **Malware URLs** (`url`) and **Malware payloads** (`malware`) always query the URLhaus Community API and require an `Auth Key` regardless of **API Type**. Requests without the key are rejected.
+- **SSL Blacklisted Certificates** (`sslblacklist`) and **JA3 Fingerprints** (`ja3_fingerprints`) read the SSLBL feeds and use neither credential.
+
+Because **API Type** applies to the whole integration, selecting **Commercial API** does not remove the need for an `Auth Key` unless you also disable **Malware URLs** and **Malware payloads**.
 
 #### Obtain `Auth Key` (Community API)
```

The Setup step has the same blind spot — it currently reads as if Commercial API replaces the Auth Key:

```diff
--- a/packages/ti_abusech/_dev/build/docs/README.md
+++ b/packages/ti_abusech/_dev/build/docs/README.md
@@ -86,6 +86,6 @@
     * To **Collect abuse.ch logs via API**, you'll need to:
 
-        - Configure **Auth Key** for Community API datasets.
-        - For Commercial API collection, set **API Type** to **Commercial API** and configure **Username** and **Password**.
+        - Configure **Auth Key (Community)**. The **Malware URLs** and **Malware payloads** datasets always need it, and **ThreatFox threat indicators** and **MalwareBazaar payloads** need it when **API Type** is **Community API**.
+        - To use the Commercial API for **ThreatFox threat indicators** and **MalwareBazaar payloads**, set **API Type** to **Commercial API** and configure **Username (Commercial)** and **Password (Commercial)**.
         - Enable/Disable the required datasets.
         - For each dataset, adjust the integration configuration parameters if required, including the URL, Interval, etc. to enable data collection.
```

**Follow-ups after applying**

- `elastic-package test policy -g -d url,malware` to refresh the policy expectations, since the rendered programs change.
- `elastic-package build` to regenerate `docs/README.md` from `_dev/build/docs/README.md`.
- A changelog line for 5.0.0, something like "Emit a configuration error when the Auth Key is not set for the Malware URLs and Malware payloads datasets."
- Optional but consistent with what `url` already does: a `missing_auth_key.txt` script test alongside `deprecated_url.txt` and `interval_too_long.txt`, proving the guard fires.
- Update commit message accordingly

packages/ti_abusech/data_stream/threatfox/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
moxarth-rathod: Declining for now. time_stamp and timestamp are clearly the same field, and we need to merge them so the date processor can handle both API formats. malware_bazaar and malware_bazaar_url may represent the same value, but the APIs use different fieldnames, and I don't want to assume they're equivalent without vendor confirmation. Keeping both fields documented separately is the safer approach for this PR.

packages/ti_abusech/_dev/build/docs/README.md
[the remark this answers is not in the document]
moxarth-rathod: Thanks for pointing this out.

packages/ti_abusech/_dev/deploy/docker/files/config.yml
```diff
@@ -236,7 +250,7 @@ rules:
     request_headers:
       Content-Type: "application/x-www-form-urlencoded"
       Auth-Key: "test_auth_key"
-    query_params:
+    body:
```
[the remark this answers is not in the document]
efd6: `request_body`

packages/ti_abusech/_dev/deploy/docker/files/config.yml
```diff
@@ -21,25 +21,45 @@ rules:
...
                 "file_type_mime": "application/x-dosexec",
                 "first_seen": "2021-10-05T14:02:45Z",
                 "imphash": "f34d5f2d4577ed6d9ceec516c1f5a744",
-                "intelligence": {
-                  "clamav": ["Win.Trojan.RedLineStealer.UNOFFICIAL"],
-                  "downloads": 11,
-                  "uploads": 1
-                },
-                "magika": "pebin",
+                "intelligence": {"downloads": 11, "uploads": 1},
```
efd6: We could reduce diff noise by making these objects not be rendered on single lines.

packages/ti_abusech/_dev/deploy/docker/files/config.yml
```diff
@@ -11,7 +11,7 @@ rules:
             "token": "test-commercial-token",
             "expires": 4102444800
           }
-  - path: /malwarebazaar/v1/samples
+  - path: /threatfox/v1/iocs
```
efd6: Git is obviously struggling to understand that this is an addition in the position that it's in. Does it need to be exactly here? Can we help git out?

packages/ti_abusech/data_stream/malware/agent/stream/cel.yml.hbs
```diff
@@ -23,6 +23,18 @@ redact:
     - auth_key
 program: |
   state.with(
+    !has(state.auth_key) ?
```
efd6, line 26: Please fix up the indentation that's made wrong by this addition.

packages/ti_abusech/data_stream/malwarebazaar/agent/stream/cel.yml.hbs
```diff
@@ -40,151 +43,184 @@ program: |
           {
             "events": {
               "error": {
+                "code": 400,
```
efd6:
```suggestion
                "code": "400",
```

packages/ti_abusech/data_stream/malwarebazaar/agent/stream/cel.yml.hbs
```diff
@@ -40,151 +43,184 @@ program: |
...
+                          "since": since,
+                          "after": body.meta.next_cursor,
+                          ?"first_seen": first_seen,
+                        }
+                      :
+                        {
+                          "since": since,
+                          ?"first_seen": first_seen,
+                        }
+                    ),
```
efd6:
```suggestion
                    ).as(first_seen,
                      {
                        "since": since,
                        ?"after": body.?meta.next_cursor,
                        ?"first_seen": first_seen,
                      }
                    ),
```
This means that it is possible to express this all as
```
                      {
                        "since": since,
                        ?"after": body.?meta.next_cursor,
                        ?"first_seen": (
                          body.items.size() > 0 ?
                            optional.of(
                              (
                                has(state.?cursor.first_seen) ?
                                  max(timestamp(state.cursor.first_seen), body.items.map(ind, timestamp(ind.first_seen)).max())
                                :
                                  body.items.map(ind, timestamp(ind.first_seen)).max()
                              ).format(time_layout.RFC3339)
                            )
                          :
                            state.?cursor.first_seen
                        ),
                      }
```

packages/ti_abusech/data_stream/malwarebazaar/agent/stream/cel.yml.hbs
```diff
@@ -40,151 +43,184 @@ program: |
...
-              },
-            })
+                )
+              )
+            )
+          )
+      )
+      :
+      !has(state.auth_key) ?
+        {
```
efd6: Please fix the indentation.

packages/ti_abusech/data_stream/threatfox/agent/stream/cel.yml.hbs
```diff
@@ -15,58 +15,228 @@ resource.timeout: {{http_client_timeout}}
...
+                            "since": since,
+                            "after": body.meta.next_cursor,
+                            ?"first_seen": first_seen,
+                          }
+                        :
+                          {
+                            "since": since,
+                            ?"first_seen": first_seen,
+                          }
+                      ),
```
efd6: Same here.

packages/ti_abusech/data_stream/threatfox/agent/stream/cel.yml.hbs
```diff
@@ -15,58 +15,228 @@ resource.timeout: {{http_client_timeout}}
...
+        )
+    ).as(since,
+      state.with(
+        (state.?api_type.orValue("community") == "commercial") ?
+          (
+            (!has(state.username) || !has(state.password)) ?
+              {
+                "events": {
+                  "error": {
+                    "code": 400,
```
efd6:
```suggestion
                    "code": "400",
```

packages/ti_abusech/data_stream/url/agent/stream/cel.yml.hbs
```diff
@@ -39,13 +39,24 @@ program: |
       {
         "events": {
           "error": {
-            "code": "400",
+            "code": 400,
```
efd6:
```suggestion
            "code": "400",
```

packages/ti_abusech/data_stream/url/agent/stream/cel.yml.hbs
```diff
@@ -28,7 +28,7 @@ program: |
...
```
efd6:
```suggestion
            "code": "400",
```

packages/ti_abusech/data_stream/url/agent/stream/cel.yml.hbs
```diff
@@ -39,13 +39,24 @@ program: |
...
+          "error": {
+            "code": 400,
+            "id": "configuration_error",
+            "message": "auth_key is not configured. The URLhaus API rejects unauthenticated requests. Generate an Auth Key in the abuse.ch authentication portal (https://auth.abuse.ch) and set the Auth Key (Community) setting.",
+          },
+        },
+        "want_more": false,
+      }
     :
     request("GET", state.url).with({
```
efd6: Please fix the indentation.

packages/ti_abusech/data_stream/url/agent/stream/cel.yml.hbs
```diff
@@ -39,13 +39,24 @@ program: |
...
             "message": "the polling interval (" + state.interval_value + ") exceeds the 3-day window of the URLhaus recent API. Reduce the interval to 72h or less to avoid missing data.",
           },
         },
         "want_more": false,
       }
+    : !has(state.auth_key) ?
+      {
+        "events": {
+          "error": {
+            "code": 400,
```
efd6:
```suggestion
            "code": "400",
```

packages/ti_abusech/data_stream/malwarebazaar/agent/stream/cel.yml.hbs
efd6: Can you also apply the second part?

packages/ti_abusech/data_stream/threatfox/agent/stream/cel.yml.hbs
efd6: This refers to https://github.com/elastic/integrations/pull/20134#discussion_r3772115755.

packages/ti_abusech/data_stream/malware/agent/stream/cel.yml.hbs
```diff
@@ -23,42 +23,61 @@ redact:
...
-    ))
+    :
+      request("GET", state.url).with(
+        {
+          "Header": {
+            "Content-Type": ["application/json"],
+            ?"Auth-Key": has(state.auth_key) ?
+              optional.of([state.auth_key])
+            :
+              optional.none(),
```
efd6:
```suggestion
            ?"Auth-Key": state.?auth_key.optMap(v, [v]),
```

(check this; it may not be available on this stack version)

packages/ti_abusech/data_stream/malware/agent/stream/cel.yml.hbs
```diff
@@ -23,42 +23,61 @@ redact:
...
+          },
+        }
+      ).as(req,
+        req.do_request().as(resp, resp.StatusCode == 200 ?
```
efd6:
```suggestion
      ).do_request().as(resp, resp.StatusCode == 200 ?
```
and clean up. The binding was not necessary.

packages/ti_abusech/data_stream/malwarebazaar/agent/stream/cel.yml.hbs
```diff
@@ -40,159 +43,192 @@ program: |
...
+                            (
+                              has(state.?cursor.first_seen) ?
+                                max(timestamp(state.cursor.first_seen), body.items.map(ind, timestamp(ind.first_seen)).max())
+                              :
+                                body.items.map(ind, timestamp(ind.first_seen)).max()
+                            ).format(time_layout.RFC3339)
+                          )
+                        :
+                          state.?cursor.first_seen
+                      ).as(first_seen,
```
efd6: You don't need to bind this. It's only used once.

kcreddy (approved):
Thanks @moxarth-rathod. LGTM for my comments.
Please wait for @efd6 approval.

efd6 (approved):

PR #20142, opened 2026-07-15
[New Integration] TYCHON Quantum Command integration package

Audience2801:
## Proposed commit message

Add the TYCHON Quantum Command integration package for Elastic Agent and Fleet

This change introduces the initial `0.1.0` release of the `tychon_quantum_command` integration package aligned to the current TYCHON Quantum Command release.

WHAT:
- Adds the `tychon_pqc` logs data stream for TYCHON Quantum Command NDJSON/JSON output collected through the `filestream` input.
- Adds the ingest pipeline and field mappings required to normalize TYCHON source events into ECS-aligned and product-specific fields.
- Adds transform definitions that read from the namespace-safe source data stream pattern `logs-tychon_quantum_command.tychon_pqc-*` and publish analytical destination indices for TYCHON datasets including inventory, applications, certificates, ciphers, crypto libraries, keystores, VPN clients, IPsec tunnels, MACsec, installed apps, browser extensions, assets, archives, and system readiness.
- Adds packaged Kibana assets including dashboards, visualizations, searches, tags, and index patterns for the supported TYCHON reporting views.
- Adds package documentation, screenshots, validation configuration, sample events, and local tests to verify manifest assets, README generation, package readiness, stream configuration, changelog links, and Kibana saved object integrity.

WHY:
- TYCHON Quantum Command produces security-relevant cryptographic inventory and post-quantum readiness data that benefits from centralized ingestion, normalization, analytics, and visualization in Elastic.
- This package provides a supported ingestion path for TQC file output while preserving namespace-safe source indexing and producing dataset-focused transform destinations for dashboards and reporting.
- The package version is intentionally kept at `2.0.4` so the initial integration release matches the currently released TQC version.

## Checklist

## Author's Checklist

- [x] Verified `elastic-package check` completes successfully.
- [x] Verified `elastic-package build` produces `tychon_quantum_command-0.1.0.zip`.
- [x] Verified package docs use the generated README template flow.
- [x] Verified the root `sample_event.json` is a real event and not the default placeholder.
- [x] Verified transforms use the namespace-safe source pattern `logs-tychon_quantum_command.tychon_pqc-*`.
- [x] Verified local Python test suite passes.

1. Clone or update a local checkout of `elastic/integrations`.
2. Copy this package into `packages/tychon_quantum_command`.
3. From the repository root or package directory, run:
   ```bash
   elastic-package check
   elastic-package build


packages/tychon_quantum_command/kibana/visualization/tychon_quantum_command-01375d36-099a-405d-a513-b1a782190a1d.json
efd6, line 1: Please re-export all the dashboards using `elastic-package`. The direct API save leaves the object as a JSON-encoded object in a string instead of a JSON object.

packages/tychon_quantum_command/tests/__pycache__/test_changelog.cpython-314.pyc
efd6, line 1: Please remove the tests directory.

packages/tychon_quantum_command/_dev/build/docs/README.md
```diff
@@ -0,0 +1,1104 @@
...
+| kerberos.config_source |  | keyword |
+| kerberos.has_aes128 |  | boolean |
+| kerberos.has_aes256 |  | boolean |
+| kerberos.has_des |  | boolean |
+| kerberos.has_rc4 |  | boolean |
+| kerberos.is_quantum_safe |  | boolean |
+| kerberos.raw_bitmask |  | long |
+| kerberos.supported_types |  | keyword |
+| observer.domain |  | keyword |
+| observer.hostname |  | keyword |
```
efd6: These should not be included in the _dev/build/docs file, they should be templated into the final file in README.md by addition of the short code.

packages/tychon_quantum_command/kibana/visualization/tychon_quantum_command-01375d36-099a-405d-a513-b1a782190a1d.json
Audience2801: I re-exported the dashboards and dependencies with elastic-package However, the resulting files still contain Kibana JSON-string fields such as `panelsJSON`, `searchSourceJSON`, and index pattern fields like `fieldAttrs` and `fields`. Could you clarify the exact normalization/export workflow you expect for package-ready saved objects in this repo?

efd6: How did you do that? Did you follow the instruction [here](https://github.com/elastic/elastic-package/blob/main/docs/howto/create_new_package.md#export-package-dashboards)?

Audience2801: that is correct, used 'elastic-package export dashboards' command.  I provided them in latest commit.  Let me know if there's something I'm missing in this process.

efd6: What version of elastic-package are you using?

efd6: I think I see what's going on now; these are all visualisation that were previously saved by reference. The dashboards do not appear to actually use them at all, so it should be fine to delete the whole directory (note that the guidelines say that [visualisations should be saved by value, not reference](https://www.elastic.co/docs/extend/integrations/dashboard-guidelines)). Presumably at some point they were saved by reference and exported via the API and have just sat around as cruft since then.

This is also likely true of the kibana/index_pattern files.

packages/tychon_quantum_command/kibana/dashboard/tychon_quantum_command-02366033-dee4-4ea4-8458-7c8a8b58987e-cost-analysis.json
efd6, line 1: Please don't use vega visualisations; the complexity they bring is not required for the visualisations that are added. This applies to all five dashboards.

packages/tychon_quantum_command/kibana/dashboard/tychon_quantum_command-d0159137-9139-4a93-85bb-2ac9ddab3982-certificate-operations-dashboard.json
```diff
@@ -0,0 +1,653 @@
...
+                    "y": 91
+                },
+                "panelIndex": "3a5ae8e6-acd0-40f9-a6fb-488f44a55e32",
+                "panelRefName": "panel_3a5ae8e6-acd0-40f9-a6fb-488f44a55e32",
+                "title": "Certificate Inventory Detail",
+                "type": "search"
+            }
+        ],
+        "timeRestore": false,
+        "title": "[TYCHON Quantum Command]  Certificate Operations Dashboard",
```
efd6:
```suggestion
        "title": "[TYCHON Quantum Command] Certificate Operations Dashboard",
```

packages/tychon_quantum_command/kibana/dashboard/tychon_quantum_command-02366033-dee4-4ea4-8458-7c8a8b58987e-cost-analysis.json
Audience2801: We understand the concern about Vega complexity. In this package, Vega is being used deliberately because the dashboard panels are not simple static charts; they provide a consistent operational UI pattern across all five dashboards that Kibana Lens/aggregation panels do not reproduce cleanly.

The Vega panels are needed for:

- Consistent visual treatment across dashboards: the header tiles, tier badges, compact ranked lists, topology-style views, and dark operational styling are shared across the package and need to render predictably as a cohesive experience.
- Inline dashboard interactions: several panels use Vega click handlers with `kibanaAddFilter` / `kibanaRemoveFilter` so users can filter the dashboard directly from custom tiles, rows, and topology nodes.
- Custom conditional rendering: the panels encode domain-specific states such as PQC-ready, modern/classical/legacy risk, certificate expiry, and cost tiers using custom labels, colors, empty states, and compact summaries.
- Layout control: Lens is excellent for conventional charts, but these dashboards depend on precise typography, spacing, row composition, badges, and condensed information density to match the existing TYCHON Quantum Command dashboard set.
- Topology and selection views: the application connections and selection-detail panels are not straightforward metric/bar/table panels; they require custom Vega data transforms and mark composition to preserve fit and behavior.

We also reduced the saved-object footprint by embedding these visualizations by value and removing unused standalone visualization objects, so the package no longer introduces extra by-reference visualization dependencies. The Vega specs are self-contained in the dashboard panels and covered by package tests for references, query fields, and key panel behavior.

Replacing these with Lens at this stage would materially change the dashboard form and interaction model, and would likely reduce consistency across the package. We can continue simplifying individual Vega specs where possible, but removing Vega entirely would not preserve the current functionality, fit, or form.

packages/tychon_quantum_command/sample_event.json
efd6, line 1: Please delete this.

packages/tychon_quantum_command/LICENSE.txt
efd6, line 1: This is not necessary; it will be injected by the build system.

packages/tychon_quantum_command/manifest.yml
```diff
@@ -0,0 +1,55 @@
...
+policy_templates:
+  - name: tychon_quantum_command
+    title: Tychon Quantum Command
+    description: Collect TQC output files.
+    inputs:
+      - type: filestream
+        title: Collect TQC output
+        description: Collect NDJSON or JSON files generated by TQC.
+owner:
+  github: tychonllc
```
efd6:
```suggestion
  github: elastic/security-service-integrations
```
Please also add the relevant line to .github/CODEOWNERS.

packages/tychon_quantum_command/kibana/dashboard/tychon_quantum_command-02366033-dee4-4ea4-8458-7c8a8b58987e-cost-analysis.json
efd6: I believe there are bugs in the Vega, but it's impossible to properly review because of the way the code is rendered; this is a supporting point for why I'm pushing back against including it.
- either in 7/20 or 13/20 panels the order of the ternary used for the filter toggle is inverted.
- three panels are missing a `%timefield%`
- 35/36 panels fail to show an empty indicator when no data is returned
- the connections topology uses aggs without pagination

Audience2801: The Vega is included because these visualizations are not ad hoc additions for the package. They are part of the saved-object implementation of the official product dashboards, and the package work is adapting that existing dashboard implementation into a package-safe form. That required preserving the existing behavior, embedding the necessary Vega by value, and replacing environment-specific references and navigation with package-compatible equivalents. In other words, the Vega is already part of the dashboards users rely on; the package is carrying that implementation forward, not inventing a separate one.
I agree that Vega deserves a higher review bar because it is harder to diff and easier to hide bugs in than simpler visualizations. That said, excluding it here is not a practical option. These dashboards are the primary UI for the integration’s findings, and removing the Vega-backed panels would strip out the core operational views, drilldowns, and summary workflows that make the package usable. At that point, the package would technically install, but it would no longer deliver the dashboards users need to interpret the data, which would make the dashboard portion of the package effectively useless.
The right fix is to treat the Vega as reviewable package content and hold it to explicit validation requirements, not to omit the dashboards. I’ve already addressed the concrete issues that were raised: inverted filter-toggle logic, missing time bindings, missing empty-state indicators, and the topology panel’s bucket-limit disclosure. I also added regression coverage so these specific Vega failure modes are checked in-package going forward. Unless you have another suggestion for how to include the saved-object form of these official dashboards in the integration package without carrying the Vega they already depend on, I do not see a viable way to exclude it and still ship the dashboards in a usable form.

efd6: The connections topology still uses aggs without pagination.

Audience2801: We reviewed the pagination concern in detail. The underlying issue is that this panel’s topology query is a bounded composite aggregation, and true pagination would require repeated after_key requests. In this saved-object Vega implementation, that pagination cannot be handled safely within the panel itself without moving the retrieval logic outside the saved object. That becomes a more fundamental architectural change rather than a packaging adaptation.
More broadly, we are trying to avoid material deviation from the existing saved-object implementation of the official product dashboards. These dashboards are not ad hoc additions for the integration package; they are adapted from the official product dashboards, and preserving that structure matters for ongoing supportability, maintainability, and future alignment with the product line. Reworking this panel into a different retrieval architecture, or redesigning it into a different bounded-topology experience, would move it away from the current product implementation rather than simply packaging the existing dashboard behavior.
Given those constraints, we believe the most responsible approach here is to leave the panel aligned with the existing product dashboard behavior and make the limitation explicit rather than silent. The panel discloses that the topology may be truncated at the current bucket limit, so the behavior is visible and reviewable.
If there is a preferred alternative for including the saved-object form of these official dashboards in the integration package without carrying the Vega implementation they currently depend on, we are open to that suggestion. Absent that, keeping this panel aligned with the existing product dashboard is the least disruptive and most supportable option.

efd6: Thanks

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,627 @@
...
+
+  - date:
+      field: "@timestamp"
+      formats:
+        - ISO8601
+      ignore_failure: true
+
+  - set:
+      field: event.ingested
+      value: "{{_ingest.timestamp}}"
```
efd6, line 28:
```suggestion
      value: "{{{_ingest.timestamp}}}"
```

Please use triple-stache template syntax throughout; we don't need HTML escaping here.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
efd6, line 1: Add tags for all processors.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,627 @@
...
+        - ISO8601
+        - yyyy-MM-dd'T'HH:mm:ssXXX
+        - yyyy-MM-dd'T'HH:mm:ss.SSSXXX
+        - yyyy-MM-dd HH:mm:ss
+        - MMM dd HH:mm:ss yyyy 'GMT'
+      if: ctx?.certificate?.not_after != null
+      on_failure:
+        - append:
+            field: error.pipeline
+            value: "Failed to parse certificate.not_after: {{_ingest.on_failure_message}}"
```
efd6, line 473: Failures in these date and convert processors will mean that the type is not mappable, and so the resulting document will be rejected. If there is a failure, the field should be removed.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
efd6, line 1: The pipeline here puts values into non-ECS fields that are not `tychon_quantum_command.*` namespaced. Please ensure that public fields follow ECS.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,627 @@
...
+        - ISO8601
+        - yyyy-MM-dd'T'HH:mm:ssXXX
+        - yyyy-MM-dd'T'HH:mm:ss.SSSXXX
+        - yyyy-MM-dd HH:mm:ss
+        - MMM dd HH:mm:ss yyyy 'GMT'
+      if: ctx?.certificate?.not_before != null
+      on_failure:
+        - append:
+            field: error.pipeline
+            value: "Failed to parse certificate.not_before: {{_ingest.on_failure_message}}"
```
efd6, line 458: Please use the convention for error strings, for example [here](https://github.com/elastic/integrations/blob/5e50cc02d9a36caaa56773b6b6c93e54c5c84487/packages/abnormal_security/data_stream/ai_security_mailbox/elasticsearch/ingest_pipeline/default.yml#L32).

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,627 @@
...
+      value: "{{{ _ingest.on_failure_message }}}"
+  - set:
+      field: error.processor_type
+      value: "{{{ _ingest.on_failure_processor_type }}}"
+  - set:
+      field: error.processor_tag
+      value: "{{{ _ingest.on_failure_processor_tag }}}"
+  - set:
+      field: error.pipeline
+      value: "{{{ _ingest.on_failure_pipeline }}}"
```
efd6, line 627: Please use the conventional approach; example [here](https://github.com/elastic/integrations/blob/5e50cc02d9a36caaa56773b6b6c93e54c5c84487/packages/zeek/data_stream/capture_loss/elasticsearch/ingest_pipeline/default.yml#L55-L75).

efd6:
@vera-review-bot review

efd6 (commented):
Please add system and pipeline tests.

packages/tychon_quantum_command/elasticsearch/transform/applications/transform.yml
```diff
@@ -0,0 +1,21 @@
...
+latest:
+  unique_key:
+    - observer.id
+    - id
+  sort: "@timestamp"
+sync:
+  time:
+    field: "@timestamp"
+    delay: "60s"
+frequency: "1m"
```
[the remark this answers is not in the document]
Audience2801: Set frequency to 5m.

packages/tychon_quantum_command/changelog.yml
```diff
@@ -0,0 +1,5 @@
+- version: 2.0.4
```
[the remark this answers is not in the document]
Audience2801: set version to 0.1.0

packages/tychon_quantum_command/data_stream/tychon_pqc/sample_event.json
```diff
@@ -0,0 +1,37 @@
+{
+  "@timestamp": "2026-06-11T12:00:00Z",
+  "event": {
+    "kind": "event",
+    "category": ["configuration"],
+    "dataset": "tychon_quantum_command.tychon_pqc"
```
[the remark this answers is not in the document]
Audience2801: sample_event.json generated with "elastic-package test system --generate" command

packages/tychon_quantum_command/elasticsearch/transform/applications/transform.yml
```diff
@@ -0,0 +1,21 @@
...
+dest:
+  index: tychon-pqc-applications
+latest:
+  unique_key:
+    - observer.id
+    - id
+  sort: "@timestamp"
+sync:
+  time:
+    field: "@timestamp"
```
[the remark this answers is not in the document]
Audience2801: changed sync time field to "event.ingested".

packages/tychon_quantum_command/manifest.yml
```diff
@@ -0,0 +1,55 @@
...
+version: 2.0.4
+source:
+  license: "Elastic-2.0"
+description: "Fleet integration for Tychon Quantum Command output."
+type: integration
+categories:
+  - security
+  - os_system
+conditions:
+  kibana.version: "^8.11.0"
```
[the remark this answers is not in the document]
Audience2801: widened kibana.version conditions per recommendation.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,627 @@
...
+          currentDataset == 'tychon_quantum_command.tychon_pqc' ||
+          currentDataset == 'tychon_pqc' ||
+          (genericDataset != null && currentDataset == genericDataset) ||
+          (genericDataset != null && currentDataset instanceof List && currentDataset.contains(genericDataset))
+        );
+        if (shouldDeriveDataset) {
+          def dataset = 'tychon_quantum_command.tychon_pqc';
+          if (tychonIndex != null) {
+            if (tychonIndex == 'applications') {
+              dataset = 'tychon.application';
```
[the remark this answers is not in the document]
Audience2801: fixed event.dataset rewriting.  moved to vendor namespace.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,627 @@
...
+        if (ctx.quantum_readiness.cost_analysis instanceof Map) {
+          Map costAnalysis = ctx.quantum_readiness.cost_analysis;
+          if (costAnalysis.instance_count == null) {
+            costAnalysis.instance_count = 1;
+          }
+        }
+
+  - set:
+      field: x509.subject.common_name
+      value: "{{certificate.subject_common_name}}"
```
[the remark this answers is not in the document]
Audience2801: corrected double-brace mustache replaced with triple braces.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,627 @@
...
+      ignore_failure: true
+
+  - date:
+      field: "@timestamp"
+      formats:
+        - ISO8601
+      ignore_failure: true
+
+  - set:
+      field: event.ingested
```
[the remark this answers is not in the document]
Audience2801: corrected.  allowing Fleet pipeline to populate "event.ingested".

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,627 @@
...
+          ctx.event.dataset = dataset;
+        }
+
+  - set:
+      field: data_stream.type
+      value: logs
+
+  - set:
+      field: data_stream.dataset
+      value: tychon_pqc
```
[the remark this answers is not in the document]
Audience2801: corrected pipeline overwriting.

Audience2801:
> Please add system and pipeline tests.

system and pipeline tests were added.  Can we get an updated review following our recent commit?

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
efd6: This has not been resolved, just moved.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
efd6: This has not been resolved.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
efd6: This has not been resolved.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
efd6: This has not been resolved.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
efd6: This also asks for a pipeline test.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
efd6: This processor is still present.

packages/tychon_quantum_command/data_stream/tychon_pqc/sample_event.json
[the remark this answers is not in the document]
efd6: How was this done? There is no system test.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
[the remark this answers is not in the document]
efd6: Please provide details; there is too much movement in the follow-up commit to be able to properly review this.

efd6 (commented):
There are still outstanding concerns. This change is very large, so to facilitate review, please do not resolve concerns unilaterally. If you intend to apply a fix to a concern, either mark it with a 👍 (or if complex, state what will be done), and if you reject a concern, explain why.

The major concerns I have at this stage is the absence of any tests and the non-conventional approach to mapping vendor fields; they should be under `tychon_quantum_command.tychon_pqc` based on the package and data stream name If this is too long (I think this is probably the case), then the filesystem names should be changed and the JSON path to match.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
Audience2801: We addressed this in two parts.

First, the recoverable date/convert failures now remove the invalid mapped values instead of leaving them behind. The date failure handlers remove the original field plus derived copies such as x509.* and tychon.certificate_leaf_details/tychon.cipher_negotiation fields. The convert failure handlers remove invalid numeric values from tychon.application.port and tychon.application.port_quantum_grade.score.

Second, we normalized the recoverable failure messages to the conventional processor/type/tag/pipeline/message wording so the failure identifies the processor type, processor tag, pipeline, and underlying error.

One implementation note: these recoverable sanitization warnings are stored in error.pipeline rather than error.message because elastic-package treats populated error.message as an unexpected pipeline error for documents that were successfully sanitized and indexed. The fatal top-level on_failure path still uses error.message and we’ll handle that separately in the next cleanup step.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
Audience2801: We updated the fatal pipeline on_failure path to emit the conventional processor/type/tag/pipeline/message string in error.message. We also removed the custom tychon.pipeline.failed flag and the separate error.processor_type / error.processor_tag fields from the emitted event and data stream mappings.

The recoverable date/convert sanitization warnings from the previous step remain in error.pipeline, because elastic-package treats populated error.message as an unexpected pipeline error for documents that were successfully sanitized and indexed. Fatal pipeline failures still use error.message.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
Audience2801: We understand the concern that package-specific fields should normally align with the package/data stream namespace. In our case, `tychon.*` is the established TYCHON vendor namespace used across our data model, and we have not historically represented TYCHON data under `tychon_quantum_command.tychon_pqc.*`.

We would prefer to keep `tychon.*` for TYCHON-specific fields so this integration remains consistent with existing TYCHON field conventions. Renaming the package to `tychon` is not possible as a package with a that name, serving a different function, already exists.

packages/tychon_quantum_command/data_stream/tychon_pqc/_dev/test/pipeline/test-applications.log-expected.json
```diff
@@ -0,0 +1,136 @@
...
+                    "name": "scheduler.exe",
+                    "path": "c:\\program files\\fortinet\\forticlient\\scheduler.exe",
+                    "pid": 2892,
+                    "port_count": 0,
+                    "private_key_count": 0,
+                    "service_display_name": "FortiClient VPN Service Scheduler",
+                    "service_name": "FA_Scheduler",
+                    "source": "process"
+                },
+                "host": {},
```
efd6: Empty object and arrays should be removed in a clean-up script in the ingest pipeline.

Add the following processor near the end of the ingest pipeline:
```
  - script:
      tag: script_to_drop_null_values_7559f019
      lang: painless
      description: This script processor iterates over the whole document to remove fields with null values.
      source: |-
        void handleMap(Map map) {
          map.values().removeIf(v -> {
            if (v instanceof Map) {
              handleMap(v);
            } else if (v instanceof List) {
              handleList(v);
            }
            return v == null || v == '' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0)
          });
        }
        void handleList(List list) {
          list.removeIf(v -> {
            if (v instanceof Map) {
              handleMap(v);
            } else if (v instanceof List) {
              handleList(v);
            }
            return v == null || v == '' || (v instanceof Map && v.size() == 0) || (v instanceof List && v.size() == 0)
          });
        }
        handleMap(ctx);
```

packages/tychon_quantum_command/data_stream/tychon_pqc/_dev/test/pipeline/test-common-config.yml
```diff
@@ -0,0 +1,9 @@
+fields:
+  tags:
+    - preserve_original_event
+  data_stream.type: logs
+  data_stream.dataset: tychon_quantum_command.tychon_pqc
+  data_stream.namespace: default
```
efd6: We don't need these, they don't impact on pipeline tests.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,943 @@
...
+  - set:
+      tag: set_ecs_version
+      field: ecs.version
+      value: 8.17.0
+
+  - set:
+      tag: set_event_ingested
+      field: event.ingested
+      value: "{{{_ingest.timestamp}}}"
+      ignore_empty_value: true
```
efd6: Remove this and let the `logs@default-pipeline` handle it.

packages/tychon_quantum_command/data_stream/tychon_pqc/_dev/test/pipeline/test-common-config.yml
```diff
@@ -0,0 +1,9 @@
...
+dynamic_fields:
+  event.ingested: ".*"
+  error.pipeline: ".*"
```
efd6: These should not be required:
- `event.ingested` does not need to be set in pipeline tests
- `error.pipeline` will be constant for any given input

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
efd6: @jamiehynds Do Tychon have a special arrangement for working outside ECS conventions?

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,943 @@
...
+      tag: remove_message_2
+      field: message
+      ignore_missing: true
+      if: ctx.event != null && ctx.event.original != null
+
+on_failure:
+  - set:
+      tag: set_error_message
+      field: error.message
+      value: "Processor {{{_ingest.on_failure_processor_type}}} with tag {{{_ingest.on_failure_processor_tag}}} in pipeline {{{_ingest.on_failure_pipeline}}} failed with message: {{{_ingest.on_failure_message}}}"
```
efd6:
```suggestion
  - append:
      tag: append_tags_b6c08950
      field: tags
      value: preserve_original_event
      allow_duplicates: false
      if: ctx.error?.message != null
on_failure:
  - append:
      tag: append_error_message_417e65c9
      field: error.message
      value: |-
        Processor '{{{ _ingest.on_failure_processor_type }}}'
        {{{#_ingest.on_failure_processor_tag}}}with tag '{{{ _ingest.on_failure_processor_tag }}}'
        {{{/_ingest.on_failure_processor_tag}}}failed with message '{{{ _ingest.on_failure_message }}}'
  - set:
      field: event.kind
      tag: set_pipeline_error_to_event_kind_72b1902f
      value: pipeline_error
  - append:
      tag: append_tags_279d5a5c
      field: tags
      value: preserve_original_event
      allow_duplicates: false
```

packages/tychon_quantum_command/data_stream/tychon_pqc/fields/base-fields.yml
```diff
@@ -0,0 +1,2 @@
+- name: "@timestamp"
+  type: date
```
efd6:
```suggestion
  external: ecs
```

packages/tychon_quantum_command/data_stream/tychon_pqc/fields/ecs.yml
efd6, line 1: Use `external: ecs` for ECS fields.

packages/tychon_quantum_command/elasticsearch/transform/applications/fields/base-fields.yml
```diff
@@ -0,0 +1,19 @@
+- name: "@timestamp"
+  type: date
```
efd6:
```suggestion
  external: ecs
```

packages/tychon_quantum_command/elasticsearch/transform/applications/fields/base-fields.yml
```diff
@@ -0,0 +1,19 @@
...
+        - name: id
+          type: keyword
+        - name: dataset
+          type: keyword
+        - name: family
+          type: keyword
+    - name: index
+      type: keyword
+    - name: type
+      type: keyword
```
efd6: Move these to the fields.yml file.

packages/tychon_quantum_command/elasticsearch/transform/applications/fields/ecs.yml
efd6, line 1: Use `external: ecs` for ECS fields.

packages/tychon_quantum_command/elasticsearch/transform/archives/fields/base-fields.yml
```diff
@@ -0,0 +1,19 @@
+- name: "@timestamp"
+  type: date
```
efd6:
```suggestion
  external: ecs
```

packages/tychon_quantum_command/elasticsearch/transform/archives/fields/base-fields.yml
```diff
@@ -0,0 +1,19 @@
...
+
+- name: tychon
```
efd6: Move these to fields.yml.

packages/tychon_quantum_command/elasticsearch/transform/archives/fields/ecs.yml
efd6, line 1: Use `external: ecs` for ECS fields.

packages/tychon_quantum_command/elasticsearch/transform/assets/fields/base-fields.yml
```diff
@@ -0,0 +1,19 @@
+- name: "@timestamp"
+  type: date
```
efd6:
```suggestion
  external: ecs
```

packages/tychon_quantum_command/elasticsearch/transform/assets/fields/base-fields.yml
```diff
@@ -0,0 +1,19 @@
...
+
+- name: tychon
```
efd6: Please move these.

packages/tychon_quantum_command/elasticsearch/transform/assets/fields/ecs.yml
efd6, line 1: Use `external: ecs` for ECS fields.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
jamiehynds: @efd6 its been awhile but I think @chrisberkhout ran into this with the Tychon Agentless integration and can probably explain where we landed in the end.

dev/gitutil/git.go
chrisberkhout, line 1: This shouldn't be removed.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
chrisberkhout: The `tychon.*` namespace is used in the TYCHON Agentless (`tychon`) integration.
I don't foresee a problem with several Tychon integrations putting fields under the one `tychon.*` namespace.

But we should keep non-ECS field inside the the vendor prefixed namespace. Looking through some of the other comments and the expected pipeline test outputs, there are a lot of non-ECS top level fields. e.g. `kerberos`.

With all the transforms, it would be good to keep source and destination field definitions in sync. It's possible to use `elastic-package links` to replace a duplicate file with a link that points to the original. If different transform destinations have different subsets of field definitions, each could be in a separate `fields/*.yml` file, and e.g. `elasticsearch/transform/applications/fields/application.yml` could be a link to `data_stream/tychon_pqc/fields/application.yml`.

packages/tychon_quantum_command/data_stream/tychon_pqc/manifest.yml
chrisberkhout, line 1: Why is the data stream slug `tychon_pqc` when the title is "Tychon Quantum Command" (also used: "TQC"). Is it a typo?

packages/tychon_quantum_command/kibana/index_pattern/05179988-8b3d-4f26-9832-d83a1e769b2d.json
chrisberkhout, line 1: We usually don't ship index patterns with integration packages. Dashboards can generally select the fields they need without an index pattern object. That's what we did for the other Tychon integration.

dev/gitutil/git.go
teresaromero: +1

packages/tychon_quantum_command/manifest.yml
```diff
@@ -0,0 +1,55 @@
+format_version: 2.0.0
+name: tychon_quantum_command
+title: "TYCHON Quantum Command"
+version: 0.1.0
```
teresaromero, line 4: i am not sure i understand the pr description:

> This change introduces the initial 2.0.4 release of the tychon_quantum_command integration package aligned to the current TYCHON Quantum Command release.

is 2.0.4 the version of the package or the version of the service?

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,999 @@
...
+          if (ctx.quantum_ready != null) { ctx.tychon.quantum_ready.overall = ctx.quantum_ready; }
+          if (ctx.quantum_ready_cert != null) { ctx.tychon.quantum_ready.certificate = ctx.quantum_ready_cert; }
+          if (ctx.quantum_ready_cipher != null) { ctx.tychon.quantum_ready.cipher = ctx.quantum_ready_cipher; }
+          if (ctx.quantum_ready_kx != null) { ctx.tychon.quantum_ready.key_exchange = ctx.quantum_ready_kx; }
+          if (ctx.quantum_ready != null) {
+            if (ctx.tychon.application == null || !(ctx.tychon.application instanceof Map)) { ctx.tychon.application = [:]; }
+            ctx.tychon.application.quantum_ready = ctx.quantum_ready;
+          }
+        }
+        if (ctx.tychon.applicaiton != null) {
```
teresaromero:
```suggestion
        if (ctx.tychon.application != null) {
```
 is this a typo?

packages/tychon_quantum_command/validation.yml
```diff
@@ -0,0 +1,6 @@
+# validation.yml
+docs_structure_enforced:
+  enabled: true
+  version: 1
+errors:
+  exclude_checks: [ "SVR00002" ]
```
teresaromero, line 7: nit: there is usually a comment inline to reflect why excluding the code (meaning of the code)

teresaromero (dismissed):
Please, restore the git file. thanks!

dev/gitutil/git.go
Audience2801: restored in a3d5022d68.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
Audience2801: fixed in 72dd8aacd7

packages/tychon_quantum_command/validation.yml
Audience2801: documented in a3d5022d68

packages/tychon_quantum_command/manifest.yml
Audience2801: PR description updated to match package version: changed to 0.1.0.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
efd6: This has not been done; note the append prior to the `on_failure` was not accidental.

packages/tychon_quantum_command/elasticsearch/transform/macsec/fields/fields.yml
```diff
@@ -0,0 +1,261 @@
+- name: macsec
```
efd6, line 1: This ends up in root.
```suggestion
- name: tychon.macsec
```

packages/tychon_quantum_command/elasticsearch/transform/archives/fields/fields.yml
```diff
@@ -0,0 +1,253 @@
+- name: archive
```
efd6, line 1: This ends up in root.
```suggestion
- name: tychon.archive
```

packages/tychon_quantum_command/elasticsearch/transform/certificates/fields/fields.yml
```diff
@@ -0,0 +1,391 @@
...
+    type: boolean
+  - name: is_explicitly_configured
+    type: boolean
+  - name: is_quantum_safe
+    type: boolean
+  - name: raw_bitmask
+    type: long
+  - name: supported_types
+    type: keyword
+- name: hash.sha256_certificate
```
efd6, line 336: This is not an ECS field, placed under an ECS hierarchy.

packages/tychon_quantum_command/elasticsearch/transform/certificates/fields/fields.yml
```diff
@@ -0,0 +1,391 @@
...
+  type: keyword
+- name: observer.tychon_client_id
+  type: keyword
+- name: observer.windows_server_role
+  type: keyword
+- name: observer.software_version
+  type: keyword
+- name: x509.is_valid
+  type: boolean
```
efd6, line 345: Same here.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
efd6: There are quite a few non-conformant fields.

### Wrong top-level namespaces (not ECS, not `tychon.*`)

| Field | Reason |
|---|---|
| `macsec.*` (group) | `macsec` is not an ECS namespace; should be `tychon.macsec.*` |
| `archive.*` (group) | `archive` is not an ECS namespace; should be `tychon.archive.*` |

### Invalid sub-fields under ECS namespaces

| Field | Reason |
|---|---|
| `hash.sha256_certificate` | ECS `hash` only defines `cdhash`, `md5`, `sha1`, `sha256`, `sha384`, `sha512`, `ssdeep` |
| `x509.is_ca` | Not an ECS x509 field |
| `x509.is_valid` | Not an ECS x509 field |
| `x509.validity.not_after` | ECS has `x509.not_after` directly, not under `x509.validity.*` |
| `x509.validity.not_before` | Same |
| `file.permissions` | ECS has `file.mode`, not `file.permissions` |
| `file.group.name` | ECS defines `file.group` as a plain keyword — making it an object with `.name` conflicts |
| `file.owner.name` | Same — ECS has `file.owner` as a plain keyword |
| `host.cpu.cores` | ECS only defines `host.cpu.usage` |
| `host.memory.total` | Not in ECS |
| `observer.domain` | Not in ECS (ECS has `observer.hostname`, `observer.name`, not `observer.domain`) |
| `observer.os.build` | ECS has `observer.os.{family,full,kernel,name,platform,type,version}` but not `observer.os.build` |
| `observer.id` | Not in ECS (ECS has `observer.serial_number`, `observer.name`, but not `observer.id`) |
| `observer.software_version` | Not in ECS |
| `observer.tychon_client_id` | Not in ECS — vendor-specific field, should be `tychon.*` |
| `observer.windows_server_role` | Not in ECS — vendor-specific field, should be `tychon.*` |
| `process.path` | Not in ECS (use `process.executable`) |
| `server.protocol` | Not in ECS — ECS `server.*` has no `protocol` sub-field |

### Filestream boilerplate (same as tychon package, so given that they are there, I guess we have precedent and these could be allowed)

| Field | Reason |
|---|---|
| `input.type` | Not in ECS |
| `log.file.device_id` | Not in ECS |
| `log.file.fingerprint` | Not in ECS |
| `log.file.inode` | Not in ECS |

packages/tychon_quantum_command/img/TQC_Application_Detail.png
efd6, line 1: Rather than smearing the sensitive values, can you use an instance that does not include values you don't want seen?

Also a general comment on the use of dark mode; I believe this is an accessibility issue as the contrast and darkness make the images extremely difficult to get anything from.

packages/tychon_quantum_command/manifest.yml
```diff
@@ -0,0 +1,55 @@
...
+source:
+  license: "Elastic-2.0"
+description: "Fleet integration for Tychon Quantum Command output."
+type: integration
+categories:
+  - security
+  - os_system
+conditions:
+  kibana.version: ">=8.11.0 <10.0.0"
```
efd6, line 13: Is there a reason to reach this far back and forward?

packages/tychon_quantum_command/manifest.yml
efd6: Please address this.

efd6 (changes requested):

teresaromero (commented):
Thanks for reverting the changes on git

packages/tychon_quantum_command/manifest.yml
Audience2801: tightened range to match other existing integration.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1239 @@
...
+      ignore_missing: true
+      if: (ctx.tychon != null || ctx.containsKey('tychon.index')) && (ctx.event == null || ctx.event.original == null)
+
+  - remove:
+      tag: remove_message_2
+      field: message
+      ignore_missing: true
+      if: ctx.event != null && ctx.event.original != null
+
+on_failure:
```
efd6, line 1265: Please add the append to tags requested in https://github.com/elastic/integrations/pull/20142#discussion_r3816677012.

packages/tychon_quantum_command/data_stream/tychon_pqc/_dev/test/pipeline/test-malformed-json.json-expected.json
```diff
@@ -0,0 +1,30 @@
...
+                "dataset": "tychon_quantum_command.tychon_pqc",
+                "kind": "pipeline_error",
+                "original": "{ malformed-json"
+            },
+            "tags": [
+                "preserve_original_event"
+            ],
+            "tychon": {
+                "error": {
+                    "pipeline": [
```
efd6, line 22: Why are you putting this here?

packages/tychon_quantum_command/sample_event.json
efd6: This appears to have come back. Please deleted it.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,1239 @@
...
+            tag: initialize_tychon_json_message
+            field: tychon
+            value: {}
+            override: false
+        - set:
+            tag: preserve_malformed_json_message
+            field: event.original
+            copy_from: message
+            override: false
+            ignore_empty_value: true
```
efd6: Why are you doing this conditionally on failure? If a user wants to keep the original event they cannot under this model.

Audience2801: We moved the event.original copy before JSON parsing and gated it on the user’s preserve_original_event setting. This should preserve the raw message even when parsing succeeds or
    fails.

packages/tychon_quantum_command/data_stream/tychon_pqc/_dev/test/pipeline/test-malformed-json.json-expected.json
Audience2801: We removed the malformed-JSON tychon.error.pipeline output from the expected fixture. Parse failures now use ECS error.message; the vendor-specific error is no longer emitted for this case.

packages/tychon_quantum_command/data_stream/tychon_pqc/elasticsearch/ingest_pipeline/default.yml
Audience2801: We added preserve_original_event to the JSON parser’s on_failure path with duplicate protection. The top-level pipeline failure handler already had the same behavior.

packages/tychon_quantum_command/_dev/build/docs/README.md
```diff
@@ -0,0 +1,198 @@
...
+4. Elasticsearch transforms filter the combined source stream by `tychon.index` and publish current-state upsert/entity destination indices such as `tychon-pqc-inventory`, `tychon-pqc-certificates`, and `tychon-pqc-system-readiness`.
+5. Kibana dashboards and saved objects query both the raw and transformed views for operational analysis.
+
+## What data does this integration collect?
+
+The integration collects file-based TYCHON Quantum Command output in NDJSON or JSON form. The package is designed around one primary ingest stream and multiple analytical transforms.
+
+### Primary ingest stream
+
+* Data stream: `tychon_pqc`
```
efd6, line 30: Why is the slug suffixed with `pqc` rather than `qc`? From the product page I guessing that this is because the the name of the CISA document is "Post-Quantum" and the requirement in EO 14412 is for a PQC Migration Plan. Those are the only places that explain it. I would suggest that either PQC be introduced in the overview if it is important, or more likely, name the integration after the Tychon product; tychon_qc or something similar (though there is stutter since the package is already tychon_quantum_command, so the full path would be tychon_quantum_command.tychon_qc which is repetitive).

packages/tychon_quantum_command/data_stream/tychon_pqc/_dev/test/pipeline/test-events.json
```diff
@@ -0,0 +1,27 @@
+{
+  "events": [
+    {
+      "tychon": {
+        "raw": {
+          "@timestamp": "2026-07-06T17:03:00.055778Z",
+          "event.category": "process",
```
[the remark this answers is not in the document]
efd6: Please do this. Alternative to a script may be to use `dot_expander`.

packages/tychon_quantum_command/data_stream/tychon_pqc/_dev/test/pipeline/test-invalid-typed-fields.log
```diff
@@ -0,0 +1,4 @@
+{"@timestamp":"2026-07-06T17:03:00.055778Z","event.category":"network","event.dataset":"tychon.inventory","event.kind":"event","event.type":"info","id":"invalid-typed-fields-test-record-1","observer.id":"00000000-0000-4000-8000-000000000002","observer.hostname":"test-host-01","observer.domain":"example.test","certificate.not_before":"not-a-date","certificate.not_after":"also-not-a-date","cipher.last_seen":"not-a-date-either","tychon.application.port":"not-a-port","tychon.application.port_quantum_grade.score":"not-a-score","tychon.index":"inventory","tychon.type":"inventory"}
+{"@timestamp":"2026-07-06T17:03:01.055778Z","event.category":"process","event.dataset":"tychon.app_certificate","event.kind":"event","event.type":"info","id":"valid-certificate-test-record-1","host.ip":"192.0.2.10","observer.id":"00000000-0000-4000-8000-000000000002","certificate.issuer_common_name":"Example Test CA","certificate.subject_common_name":"app.example.test","certificate.not_before":"2025-01-01T00:00:00Z","certificate.not_after":"2030-01-01T00:00:00Z","certificate.signature_algorithm":"SHA256-RSA","tychon.index":"certificates","tychon.type":"certificate"}
```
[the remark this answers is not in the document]
efd6: Please add this.

packages/tychon_quantum_command/data_stream/tychon_pqc/_dev/test/pipeline/test-malformed-json.json-expected.json
efd6: This is still here.

ShourieG:
@vera-review-bot ignore

PR #19756, opened 2026-06-25
Rename Agentless to Elastic Managed Integration

michalpristas:
This pull request standardizes terminology across several integrations by replacing references to "Agentless" with "Elastic Managed" in documentation, manifests, and ingest pipeline descriptions. The updates clarify the deployment mode and ensure consistency in how Elastic Managed integrations are presented to users.

**Terminology and Documentation Updates:**

* All references to "Agentless" have been replaced with "Elastic Managed" in integration documentation and user guides for `1password`, `abnormal_security`, `admin_by_request_epm`, `airlock_digital`, and `agentless_hello_world` packages. This includes section headings, feature descriptions, and installation instructions. 

**Manifest and Metadata Changes:**

* The `agentless_hello_world` integration manifest and related documentation have been updated to use "Elastic Managed Hello World" as the title and throughout the description, replacing all instances of "Agentless Hello World." 

**Ingest Pipeline Descriptions:**

* Ingest pipeline YAML files for `agentless_hello_world` have updated their descriptions and processor comments to refer to "Elastic Managed" instead of "Agentless," ensuring alignment with the new terminology. 

These changes improve clarity for users and maintain consistency as Elastic transitions the naming of this integration type.

PR #20182, opened 2026-07-16
Bump patch version for each packages affected by Agentless renaming

michalpristas:
Bump version for each packages affected by Agentless renaming

Packages had documentation changes to reflect the change in naming
from "Agentless" to "Elastic Managed" in #19756. This change bumps the
version to publish those changes.


rubenruizdegauna (approved):
LGTM

MichaelKatsoulis (approved):

packages/1password/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.35.1"
```
efd6: Please bump minor for enhancements ([ref](https://github.com/elastic/integrations/wiki/Fleet-Package-Code-Review-Comments#versioning-changes)), or make the type be "bugfix".

Also, the description says that the documentation is updated, but I don't see any changes except for the manifest version and the changelog entry. Can you clarify this?

mohitjha-elastic: I think this is a follow-up PR for this [change](https://github.com/elastic/integrations/pull/19756/).

Changelog entry should not be required for this change since it's only a terminology update in the documentation. This is part of a feature rename that users will encounter throughout Kibana and the [documentation](https://www.elastic.co/docs/manage-data/ingest/managed-integrations/managed-integrations), and there isn't any functional or behavioral change to the integration. Adding a changelog entry for every integration may add unnecessary noise.

Happy to hear others' thoughts on this.

michalpristas: i'm ok with skipping changelog entry, wanted to be transparent why version bumped though

efd6: Ah, OK. That's confusing; that change should have had the bump since it's necessary to get the documentation rendered. I guess we need to do this one then, though the description seem wrong and the link is also sort of wrong (though I don't see a way to get a "right" one). Rather than "Rename Agentless to Elastic Managed integrations.", I'd suggest "Publish Agentless to Elastic Managed integrations name change." or similar, and then choose either bugfix (because the publication was missed) and use a patch bump, or enhancement (because we could claim that this was intentional) and use a minor bump.

I think it would be helpful to include a link to that PR in the commit message when the PR is merged.

michalpristas: i had it before automatic review just complained. i can revert to that i agree it's more informational

efd6: I've turned off the link validation, but that's not quite what I meant. I think that the link *should* match this PR, particularly given the status as an enhancement and the description that's now being used. What I was suggesting was to make the commit message be something like
```
Bump version for each packages affected by Agentless renaming

Packages had documentation changes to reflect the change in naming
from "Agentless" to "Elastic Managed" in #19756. This change bumps the
version to publish those changes.
```
and revert the link field to `https://github.com/elastic/integrations/pull/20182`.

mohitjha-elastic (approved):

packages/1password/changelog.yml
michalpristas: sorry, misunderstood

efd6 (approved):

ishleenk17 (approved):

P1llus (approved):

PR #20183, opened 2026-07-16
[checkpoint]: fix handling of different user formats

taylor-swanson:
## Proposed commit message

- Add grok processors to better handle different formats of the user fields in checkpoint logs. Added support for full name with email, email, and name only.

## Checklist

```
cd packages/checkpoint
elastic-package test
```


packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -255,23 +255,36 @@ processors:
...
-      field: user.name
-      copy_from: checkpoint.user
-      if: ctx.user?.name == null
-      ignore_empty_value: true
+  - grok:
+      tag: grok_src_user_name
+      field: checkpoint.src_user_name
+      if: ctx.checkpoint?.src_user_name != null
+      patterns:
+        - "^%{DATA:source.user.full_name} \\(%{EMAIL:source.user.email}\\)$"
```
Niceplace: Question: It seems like the raw log example shown in the original issue show a trailing space after `src_user_name` and `user`. Given the current shape of the regex, I believe it could fail parsing because the patterns expects to match the full line without a trailing space.

Should we be concerned about that or is every field trimmed before it gets here ?

Niceplace: I'll try to add that trailing space in the examples and run the test suites again, see if it works

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -255,23 +255,36 @@ processors:
...
+        - "^%{EMAIL:source.user.email}$"
+        - "^%{DATA:source.user.name}$"
+      pattern_definitions:
+        EMAIL: "%{DATA}@%{DATA}"
+  - grok:
+      tag: grok_administrator
+      field: checkpoint.administrator
+      if: ctx.checkpoint?.administrator != null && ctx.source?.user?.name == null
```
ilyannn: The `src_user_name` grok above leaves `source.user.name` unset for email-only and full-name-with-email values while setting `source.user.email` and optionally `source.user.full_name`. Because this fallback checks only `source.user.name`, an event containing both `src_user_name` and `administrator` runs both groks and can overwrite or combine two source identities. Please also guard `source.user.email` and `source.user.full_name`, and add a combined-fields fixture, so `src_user_name` retains precedence.

This comment has been drafted by 🤖 Cursor/GPT-5.6 Sol under my supervision

packages/checkpoint/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.47.1"
+  changes:
+    - description: Improve handling of different user formats in logs.
+      type: bugfix
```
ilyannn: I would call it enhancement though – for a start it has new fields.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: I think it gets trimmed before hand, but let me double check...

taylor-swanson: Yeah it looks like the `kv` processors up top trim extra whitespace.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -255,23 +255,36 @@ processors:
...
+  - grok:
+      tag: grok_user
+      field: checkpoint.user
+      if: ctx.checkpoint?.user != null && ctx.user?.name == null
+      patterns:
+        - "^%{DATA:user.full_name} \\(%{EMAIL:user.email}\\)$"
+        - "^%{EMAIL:user.email}$"
+        - "^%{DATA:user.name}$"
+      pattern_definitions:
+        EMAIL: "%{DATA}@%{DATA}"
```
ilyannn: It looks like the standard `EMAILADDRESS` would have also worked?

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: I don't think that `src_user_name` and `administrator` will appear in the same log (this pipeline handles many different variations of logs). We would have a bigger problem if that were to occur, but as of now, I don't think we've ever had an issue filed on this. Regardless, it's a good check to add, but I don't think I'll add a test case for it, since I don't think it's a legitimate scenario.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: I added whitespace in the original log for the pipeline tests.

packages/checkpoint/data_stream/firewall/elasticsearch/ingest_pipeline/default.yml
taylor-swanson: I thought about it originally, but went with a more basic grok. `EMAILADDRESS` does work, so I went ahead and switched to that. If for some reason the grok doesn't match, we still fall back to `user.name` which I suppose is acceptable.

packages/checkpoint/changelog.yml
taylor-swanson: Yeah I was on the fence about it. Made it an enhancement.

Niceplace (approved):

ilyannn (approved):
LGTM

PR #20187, opened 2026-07-17
[ti_opencti] Fix flaky policy test due to Fleet js-yaml→yaml timestamp serialization

juliaElastic:
## Summary

Fixes the flaky `ti_opencti` indicator policy test (`test-multi-values.yml`) that began failing on 8.19.19-SNAPSHOT after the Fleet js-yaml→yaml migration (Kibana PR #274902) was backported to the 8.19 branch.

## Root cause

The test input timestamps were written without explicit milliseconds (e.g. `"2024-01-01T00:00:00Z"`). When Handlebars renders these into the agent template, they appear as unquoted YAML scalars:

```yaml
created_after: 2024-01-01T00:00:00Z
```

Fleet's `compileTemplate()` then parses this compiled YAML back into a JS object:

| Fleet version | Parse call | Unquoted timestamp result |
|---|---|---|
| ≤ 8.19.18 | `js-yaml.load()` (YAML 1.1) | **Date object** — `Date.toISOString()` emits `.000Z` |
| ≥ 8.19.19 | `yaml.parse()` (YAML 1.2 core, default) | **string** — no milliseconds added |

The expected output file was generated with the ≥8.19.19 behavior (which adds `.000Z`). The new code preserves the input string as-is, so the output no longer matches.

## Fix

Specify milliseconds explicitly in the test input:

```yaml
# before
created_after: "2024-01-01T00:00:00Z"
# after
created_after: "2024-01-01T00:00:00.000Z"
```

With `.000Z` in the input, both code paths produce the same semantic value (`"2024-01-01T00:00:00.000Z"`). The quoting style differs (unquoted on ≤8.19.18, double-quoted on ≥8.19.19) but elastic-package's policy test comparison normalises YAML before comparing, so both pass against the existing `.expected` file.

No changes to the `.expected` file or to Kibana are required.

## Testing

- Passes locally against 8.19.18 (js-yaml path)
- Fixes the failure reported in https://github.com/elastic/integrations/issues/20163

### DIFF
--- a/packages/ti_opencti/data_stream/indicator/_dev/test/policy/test-multi-values.yml
+++ b/packages/ti_opencti/data_stream/indicator/_dev/test/policy/test-multi-values.yml
@@ -27,8 +27,8 @@ data_stream:
     marking_ids:
       - marking-definition--613f2e26-407d-48c7-9eca-b8e91df99dc9
       - marking-definition--34098fce-860f-48ae-8e50-ebd3cc5e41da
-    valid_from_start: "2024-01-01T00:00:00Z"
-    valid_until_end: "2024-12-31T23:59:59Z"
+    valid_from_start: "2024-01-01T00:00:00.000Z"
+    valid_until_end: "2024-12-31T23:59:59.000Z"
     confidence_min: 50
-    created_after: "2024-01-01T00:00:00Z"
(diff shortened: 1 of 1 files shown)


efd6 (approved):
Thank you

PR #20190, opened 2026-07-17
Crowdstrike: use inclusive >= cursor boundary in vulnerability and alert CEL programs

chemamartinez:
## Proposed commit message

```
[crowdstrike]: use inclusive >= cursor boundary in vulnerability and alert CEL programs

Strict > on updated_timestamp permanently skips records that share a
boundary timestamp when a page or error boundary falls mid-group. Switch
to >= so same-timestamp records are always re-queried.

Re-fetched boundary records are absorbed without creating duplicates:
both pipelines run fingerprint processors that include updated_timestamp
so events that were already ingested are dropped.
```

## Checklist

### DIFF
--- a/packages/crowdstrike/changelog.yml
+++ b/packages/crowdstrike/changelog.yml
@@ -3,7 +3,7 @@
   changes:
     - description: Use an inclusive lower bound (>=) on the updated_timestamp cursor filter in the vulnerability and alert data streams to prevent records sharing a boundary timestamp from being permanently skipped at page or error boundaries.
       type: bugfix
-      link: https://github.com/elastic/integrations/pull/99999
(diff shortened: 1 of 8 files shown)


chemamartinez:
@vera-review-bot review

efd6 (approved):
Please update the commit message to fix the text, "so a re-fetch is an in-place overwrite, not a duplicate." There is no overwrite here,

navnit-elastic (approved):
`host` data stream has the same exclusive time window issue, we can follow-up on that separately.

chemamartinez:
> `host` data stream has the same exclusive time window issue, we can follow-up on that separately.

Thanks for pointing that out, I'll handle that case in a separate PR.

PR #20211, opened 2026-07-19
sentinel_one: fix duplicate event ingestion in threat_event and application_risk

efd6:
## Proposed commit message
```
sentinel_one: fix duplicate event ingestion in threat_event and application_risk

Add a time watermark (updatedAt__gte) to the threat_event CEL program
so each interval only fetches threats updated since the last completed
walk, rather than re-walking the entire catalog. The high-water mark
is tracked in cursor.last_timestamp and seeded from a new
initial_interval manifest variable on the first run. The default poll
interval drops from 24h to 5m since the watermark eliminates full
re-walks.

Add fingerprint processors to both the threat_event and
application_risk ingest pipelines so that re-emitted documents with
the same id are rejected within a backing index rather than
accumulating as duplicates.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/sentinel_one/data_stream/application_risk/elasticsearch/ingest_pipeline/default.yml
+++ b/packages/sentinel_one/data_stream/application_risk/elasticsearch/ingest_pipeline/default.yml
@@ -26,6 +26,13 @@ processors:
       field: event.original
       tag: json_event_original
       target_field: json
+  - fingerprint:
+      tag: fingerprint_application_risk_dedup
+      fields:
+        - json.id
+        - json.status
+      target_field: _id
+      ignore_missing: true
   - set:
       field: event.kind
       tag: set_event_kind
--- a/packages/sentinel_one/changelog.yml
+++ b/packages/sentinel_one/changelog.yml
@@ -1,4 +1,12 @@
 # newer versions go on top
+- version: "2.11.1"
+  changes:
+    - description: Add time watermark and fingerprint deduplication to the threat event data stream to prevent duplicate ingestion.
+      type: bugfix
+      link: https://github.com/elastic/integrations/pull/20211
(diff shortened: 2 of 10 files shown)


packages/sentinel_one/data_stream/application_risk/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -26,6 +26,12 @@ processors:
       field: event.original
       tag: json_event_original
       target_field: json
+  - fingerprint:
+      tag: fingerprint_application_risk_dedup
+      fields:
+        - json.id
```
mohitjha-elastic, line 32: The fingerprint currently uses only `json.id`. Although this is an `event.kind: state` stream, it still performs a full re-walk on every interval. Fields such as `lastScanDate`, `daysDetected`, and `status` can change for the same `id`, and using a stable `_id` causes those updated documents to be rejected as duplicates.

Would it be safer to include a changing field (for example, json.lastScanDate, and optionally status or daysDetected) in the fingerprint?

efd6: That would break the deduplication. This is certainly true for `lastScanDate` and `daysDetected`. It would be less problematic for the `status` field since there are only two states. This is a trade off that we've had to deal with in the past; we can either have the first state, or we can have the all the data and use a latest transform to remove all but the last. Potentially we could use `status`; this would give us at most two entries which could be differentiated by that field.

mohitjha-elastic: Thanks!

mohitjha-elastic (approved):

PR #20240, opened 2026-07-20
Support dynamic IaC for integrations

jeniawhite:
## Proposed commit message

We would like to support dynamic IaC generation for integrations, this change will allow integrations to declare IaC template manifests that will be used to create dynamic IaC for them to be deployed in a frictionless manner.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


seanrathier (commented):
Looks good to me!

My only thoughts were if it could be possible to to have all the IAC files in one JSON.  Granted, AWS is an extreme case,  but I am wondering if that was considered?

Side note while reviewing this,  would it be trouble for us to cache the IAC JSON in the IAC Provisioner?
This could eliminate the round trips to the EPR for known integrations with IAC.

We would need a refresh mechanism,  maybe check the version?

PR #20250, opened 2026-07-21
[ti_google_threat_intelligence] - Recover ioc_stream from persistent API 400 errors

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
ti_google_threat_intelligence: recover ioc_stream from persistent API 400 errors

The ioc_stream CEL program could never make progress past a collection
window that the API persistently rejects with HTTP 400 (for example, a
window containing an IOC whose URL identifier exceeds the API's
internal 500-byte key limit). On error the program left the stale
next-page cursor in state, so every subsequent poll replayed the same
failing request with the dead pagination cursor attached, and the
window start was never advanced. The input stayed degraded until the
Filebeat registry was deleted by hand.

On errors the program now discards the in-flight page cursor so
retries start from a clean first page. HTTP 400 responses additionally
maintain a consecutive-failure counter persisted in the cursor; after
three consecutive 400s for the same window the program advances the
cursor past the window and resumes collection, surfacing the retries
and the skip as error events in the data stream. Other error classes
keep the existing retry-forever behaviour.

Add a system test that drives the input through the failing-window
cycle against the mock API and asserts recovery.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-


ShourieG:
Ignore, something went wrong. Somehow the files got untracked in my local worktree and they got pushed with the commit.

PR #20254, opened 2026-07-21
[ti_google_threat_intelligence] - Recover ioc_stream from persistent API 400 errors

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
ti_google_threat_intelligence: recover ioc_stream from persistent API 400 errors

The ioc_stream CEL program could never make progress past a collection
window that the API persistently rejects with HTTP 400 (for example, a
window containing an IOC whose URL identifier exceeds the API's
internal 500-byte key limit). On error the program left the stale
next-page cursor in state, so every subsequent poll replayed the same
failing request with the dead pagination cursor attached, and the
window start was never advanced. The input stayed degraded until the
Filebeat registry was deleted by hand.

On errors the program now discards the in-flight page cursor so
retries start from a clean first page. HTTP 400 responses additionally
maintain a consecutive-failure counter persisted in the cursor; after
three consecutive 400s for the same window the program advances the
cursor past the window and resumes collection, surfacing the retries
and the skip as error events in the data stream. Other error classes
keep the existing retry-forever behaviour.

Add a system test that drives the input through the failing-window
cycle against the mock API and asserts recovery.
```

## Checklist

## Author's Checklist

- [ ] 

## Related issues

-

### DIFF
--- a/packages/ti_google_threat_intelligence/data_stream/ioc_stream/fields/ecs.yml
+++ b/packages/ti_google_threat_intelligence/data_stream/ioc_stream/fields/ecs.yml
@@ -2,3 +2,11 @@
 # https://github.com/elastic/elasticsearch/issues/150902.
 - name: threat.enrichments
   external: ecs
+# Error fields emitted by the CEL input when a collection window is skipped
+# after repeated HTTP 400 responses from the API.
+- name: error.code
+  external: ecs
+- name: error.id
+  external: ecs
+- name: error.message
+  external: ecs
--- a/packages/ti_google_threat_intelligence/data_stream/ioc_stream/fields/ecs.yml
+++ b/packages/ti_google_threat_intelligence/data_stream/ioc_stream/fields/ecs.yml
@@ -2,11 +2,3 @@
 # https://github.com/elastic/elasticsearch/issues/150902.
 - name: threat.enrichments
   external: ecs
-# Error fields emitted by the CEL input when a collection window is skipped
-# after repeated HTTP 400 responses from the API.
-- name: error.code
-  external: ecs
-- name: error.id
-  external: ecs
-- name: error.message
-  external: ecs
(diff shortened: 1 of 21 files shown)


packages/ti_google_threat_intelligence/manifest.yml
```diff
@@ -3,7 +3,7 @@ name: ti_google_threat_intelligence
 title: Google Threat Intelligence
 # This version must match the User-Agent version used in CEL code.
 # Remember to update the User-Agent in CEL code when changing this version.
-version: "1.1.1"
+version: "1.1.2"
```
kcreddy, line 6: Can you also update the `User-Agent` version as noted?

packages/ti_google_threat_intelligence/data_stream/ioc_stream/fields/ecs.yml
```diff
@@ -2,3 +2,11 @@
...
 - name: threat.enrichments
   external: ecs
+# Error fields emitted by the CEL input when a collection window is skipped
+# after repeated HTTP 400 responses from the API.
+- name: error.code
+  external: ecs
+- name: error.id
+  external: ecs
+- name: error.message
+  external: ecs
```
kcreddy: The fields should be covered by `ecs@mappings` template.

kcreddy (approved):
Nits only. LGTM

Also, on the skip branch (`"cursor": fails >= 3 ? {"last_timestamp": interval.end_time, ...}`): worth being explicit in the commit/comment about the true extent of the loss. Because the request filter is open-ended on `start_time` only and `end_time` is `now` at cycle start, skipping to `interval.end_time` discards everything on page 2+ of the window between `start_time` and `now`, not just the single poisoned record. If page 1 itself 400s, the whole `[start_time, now]` span goes. The "bounded data loss" phrasing undersells this a bit.

packages/ti_google_threat_intelligence/manifest.yml
ShourieG: updated this in all data streams since they were out of sync.

PR #20282, opened 2026-07-22
[island_browser] - Fix device pipeline parsing of multi-NIC internalIpAddress

### LINKED ISSUE #18159: [Island Browser] device ingest pipeline fails to parse internalIpAddress for devices with 3+ NICs
### Integration Name

Island Browser [island_browser]

### Dataset Name

island_browser.device

### Integration Version

1.0.0

### Agent Version

9.2.6

### Agent Output Type

elasticsearch

### Elasticsearch Version

9.2.6

Linux ECK

### Software/API Version

v1, `/api/external/v1/devices`

### Error Message

```
Processor convert with tag convert_internal_ip_address_to_ip in pipeline logs-island_browser.device-1.0.0 failed with message: '192.168.111.1 | Wi-Fi/192.168.1.164' is not an IP string literal.
```

### Event Original

```json
{
  "island_browser": {
    "device": {
      "internal_ip": "192.168.241.1 | VMware Network Adapter VMnet1/192.168.111.1 | Wi-Fi/192.168.1.164",
      "internal_ip_address": "192.168.111.1 | Wi-Fi/192.168.1.164"
    }
  },
  "error": {
    "message": [
      "Processor convert with tag convert_internal_ip_address_to_ip in pipeline logs-island_browser.device-1.0.0 failed with message: '192.168.111.1 | Wi-Fi/192.168.1.164' is not an IP string literal."
    ]
  },
  "event": { "kind": "pipeline_error" },
  "tags": ["preserve_original_event"]
}
```

### What did you do?

Deployed `island_browser` v1.0.0 with all 5 streams enabled (CEL input, 5m interval). The `device` stream polls `/api/external/v1/devices` which returns `internalIpAddress` as `IP | AdapterName/IP | AdapterName/...`.

### What did you see?

Several device documents have `event.kind: pipeline_error`. The `dissect` pattern `%{_temp1}/%{island_browser.device.internal_ip_address}` splits on the first `/` only. For 3+ NICs the dissected value still contains `/` separators, failing the subsequent `convert` to IP type. Devices with fewer NICs work correctly.

Reproducible without an agent

```
POST _ingest/pipeline/logs-island_browser.device-1.0.0/_simulate
{"docs":[{"_source":{"message":"{\"internalIpAddress\": \"192.168.241.1 | VMware Network Adapter VMnet1/192.168.111.1 | Wi-Fi/192.168.1.164\"}"}}]}
```

All IPs extracted from multi-NIC values

from the above example,`192.168.241.1`, `192.168.111.1`, and `192.168.1.164` parsed as valid IPs and added to `related.ip`.

### Anything else?

_No response_

ShourieG:
## Type of change
- Bug

## Proposed commit message
```
island_browser: fix device pipeline parsing of multi-NIC internalIpAddress

The device ingest pipeline used a single dissect on the first "/" to
extract an IP from internalIpAddress. When the Island API returned a
"/"-separated list of "<ip> | <adapter>" pairs for hosts with 3+ NICs,
the leftover separators failed the subsequent convert-to-ip and produced
event.kind: pipeline_error.

Strip the adapter labels with gsub, split the cleaned list into
individual addresses, and append each one to related.ip. The single-IP
path is unchanged. Adds multi-NIC, two-NIC, and single-NIC-with-adapter
cases to the device pipeline test.
```

## Note

Test data was generated using LLMs, it does not contain any customer data.

## Checklist

## Author's Checklist

- [ ] 

## Related issues

- Closes https://github.com/elastic/integrations/issues/18159


packages/island_browser/data_stream/device/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -261,28 +261,51 @@ processors:
...
+        The Island API can return internalIpAddress as a "/"-separated list of
+        "<ip> | <adapter>" pairs (one per NIC). Strip the adapter labels so only
+        the IP addresses remain.
+  - split:
+      field: _temp1
+      target_field: island_browser.device.internal_ip_address
+      tag: split_internal_ip_address
+      separator: '\s*/\s*'
+      ignore_missing: true
+      if: ctx._temp1 != null
```
mohitjha-elastic: `ignore_missing` already handles this case, so I think this `if` condition may not be necessary.

packages/island_browser/data_stream/device/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -261,28 +261,51 @@ processors:
...
       field: island_browser.device.internal_ip
-      tag: dissect_internalIpAddress
-      pattern: '%{_temp1}/%{island_browser.device.internal_ip_address}'
-      ignore_failure: true
+      target_field: _temp1
+      tag: gsub_strip_internal_ip_adapters
+      pattern: '\s*\|\s*[^/]*'
+      replacement: ''
+      ignore_missing: true
+      if: ctx.island_browser?.device?.internal_ip != null
```
mohitjha-elastic: https://github.com/elastic/integrations/pull/20282/changes#r3629960452

packages/island_browser/changelog.yml
```diff
@@ -1,4 +1,9 @@
 # newer versions go on top
+- version: "1.1.2"
+  changes:
+    - description: Fix device ingest pipeline to parse `internalIpAddress` values that contain multiple NIC entries.
+      type: bugfix
+      link: https://github.com/elastic/integrations/issues/20282
```
mohitjha-elastic:
```suggestion
      link: https://github.com/elastic/integrations/pull/20282
```

packages/island_browser/data_stream/device/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -261,28 +261,51 @@ processors:
...
+      ignore_missing: true
+      if: ctx._temp1 != null
+      description: Split the cleaned "/"-separated list into individual IP addresses.
   - convert:
       field: island_browser.device.internal_ip_address
       tag: convert_internal_ip_address_to_ip
       type: ip
       ignore_missing: true
-      if: ctx.island_browser?.device?.internal_ip_address != ''
+      if: ctx.island_browser?.device?.internal_ip_address != null
```
mohitjha-elastic: https://github.com/elastic/integrations/pull/20282/changes#r3629960452

packages/island_browser/data_stream/device/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -261,28 +261,51 @@ processors:
...
+      replacement: ''
+      ignore_missing: true
+      if: ctx.island_browser?.device?.internal_ip != null
+      description: >-
+        The Island API can return internalIpAddress as a "/"-separated list of
+        "<ip> | <adapter>" pairs (one per NIC). Strip the adapter labels so only
+        the IP addresses remain.
+  - split:
+      field: _temp1
+      target_field: island_browser.device.internal_ip_address
```
mohitjha-elastic, line 277: can we have foreach to convert all these `internal_ip_address` to `type: ip`?

packages/island_browser/data_stream/device/_dev/test/pipeline/test-device.log-expected.json
```diff
@@ -964,6 +964,276 @@
...
+            },
+            "island_browser": {
+                "device": {
+                    "browser_name": "Island",
+                    "email": "test.one@example.com",
+                    "external_ip_address": "203.0.113.202",
+                    "id": "device-test-singlenic-adapter",
+                    "internal_ip": "192.0.2.12 | Ethernet0",
+                    "internal_ip_address": [
+                        "192.0.2.12"
```
mohitjha-elastic: For a single IP value, I think we should use a `string` instead of a `list`, similar to this [case](https://github.com/elastic/integrations/blob/main/packages/island_browser/data_stream/device/_dev/test/pipeline/test-device.log-expected.json#L90).

We can handle this after the `split` processor by converting the value to a string when the list contains only a single item.

mohitjha-elastic (changes requested):

packages/island_browser/data_stream/device/elasticsearch/ingest_pipeline/default.yml
ShourieG: The convert processor just below this already converts every element when the field is an array, a foreach here would be redundant.

ShourieG:
@mohitjha-elastic, I've addressed all your comments.

packages/island_browser/data_stream/device/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -279,18 +278,26 @@ processors:
...
+  - script:
+      lang: painless
+      tag: script_flatten_single_internal_ip
+      description: >-
+        Represent a single internal IP as a scalar rather than a single-element
+        list, so devices with one NIC match the shape used for devices that
+        report a single IP directly.
+      if: ctx.island_browser?.device?.internal_ip_address instanceof List && ctx.island_browser.device.internal_ip_address.size() == 1
+      source: >-
+        ctx.island_browser.device.internal_ip_address = ctx.island_browser.device.internal_ip_address.get(0);
```
ShourieG, line 300: The script is resistant against edge-case values like `192.0.2.10/` compared to a split + set that will produce a list in this case.

packages/island_browser/data_stream/device/_dev/test/pipeline/test-device.log-expected.json
```diff
@@ -964,6 +964,274 @@
...
+                    "browser_name": "Island",
+                    "email": "test.three@example.com",
+                    "external_ip_address": "203.0.113.200",
+                    "id": "device-test-multinic-3",
+                    "internal_ip": "192.0.2.10 | Ethernet0/198.51.100.20 | Wi-Fi/203.0.113.30",
+                    "internal_ip_address": [
+                        "192.0.2.10",
+                        "198.51.100.20",
+                        "203.0.113.30"
+                    ],
```
ShourieG, line 1017: @mohitjha-elastic, example of convert processor working on multiple ips

packages/island_browser/data_stream/device/elasticsearch/ingest_pipeline/default.yml
mohitjha-elastic: I thought these processors would work only when the field contains a single item. If the field contains a list of items, we would typically use a `foreach` processor to iterate over the list and convert each element to the appropriate type, which is the primary purpose of `foreach`. This is also the pattern I observed that we have generally followed across integrations. But in the convert processor [documentation](https://www.elastic.co/docs/reference/ingest-processor/convert-processor), I can see that all members of an array field can also be converted hence this clears the confusion here.

mohitjha-elastic (approved):

